Issue #5916, opened 2020-05-29
Response Selector evaluation fails when testing on a test split


tczekajlo:
Rasa Version     : 1.11.0a1
Rasa SDK Version : 1.10.0
Rasa X Version   : None
Python Version   : 3.7.7 (default, May  5 2020, 13:33:17) 
Operating System : Linux-4.15.0-1044-gke-x86_64-with-debian-buster-sid

**Issue**:
NLU test fails with dataset `Carbon Bot` and configuration file `Sparse + DIET + ResponseSelector(T2T)`.

**Error (including full traceback)**:
```
2020-05-29 09:24:09 INFO     rasa.nlu.test  - Response selection evaluation results:
2020-05-29 09:24:09 INFO     rasa.nlu.test  - Response Selection Evaluation: Only considering those 151 examples that have a defined response out of 515 examples.
/home/github/.cache/pypoetry/virtualenvs/rasa-6QvRaJbr-py3.7/lib/python3.7/site-packages/sklearn/metrics/_classification.py:1272: UndefinedMetricWarning: Precision and F-score are ill-defined and being set to 0.0 in labels with no predicted samples. Use `zero_division` parameter to control this behavior.
  _warn_prf(average, modifier, msg_start, len(result))
/home/github/.cache/pypoetry/virtualenvs/rasa-6QvRaJbr-py3.7/lib/python3.7/site-packages/sklearn/metrics/_classification.py:1272: UndefinedMetricWarning: Precision is ill-defined and being set to 0.0 in labels with no predicted samples. Use `zero_division` parameter to control this behavior.
  _warn_prf(average, modifier, msg_start, len(result))
2020-05-29 09:24:09 INFO     rasa.nlu.test  - Classification report saved to results/public/carbon-bot/en/cvf_diet_responset2t.yml/response_selection_report.json.
2020-05-29 09:24:09 INFO     rasa.nlu.test  - Incorrect response predictions saved to results/public/carbon-bot/en/cvf_diet_responset2t.yml/response_selection_errors.json.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/github/rasa/__main__.py", line 108, in main
    cmdline_arguments.func(cmdline_arguments)
  File "/home/github/rasa/cli/test.py", line 153, in run_nlu_test
    test_nlu(model_path, nlu_data, output, vars(args))
  File "/home/github/rasa/test.py", line 174, in test_nlu
[...]

PR #6013, opened 2020-06-12
Fix bug during response evaluation

tabergma:
**Proposed changes**:
- fixes https://github.com/RasaHQ/rasa/issues/5916

If the test data is missing a response that is present in the training data and during testing this response is predicted, the evaluation failed. This is not fixed.


dakshvar22:
@tabergma Tried reproducing the functionality, attaching a screenshot - 
response_selection_confusion_matrix

I am thinking if we should refactor our train test split function for response selector data to always include all responses both in train and test splits. That way, if a user uses `rasa data split`, they'll never run into this problem. What do you think?

tabergma:
Was thinking the same, but I guess we should create a separate issue for that. Would nevertheless keep this "fix" as it also might happen, that users split the data on their own and miss some labels in the test set. What do you think?

dakshvar22:
Makes sense, do we need a changelog entry for it?

tabergma:
I don't think a changelog entry is needed as the feature is not yet released. Will create a separate issue for the train/test split.

Issue #6015, opened 2020-06-12
RulePolicy: Implement new Form featurization


wochinge:
The happy path of a form should represented by a single `Form` event in order to simplify featurization:

```
[ ACTION_LISTEN, 
  intent_which_triggers_form,
  FormEvent <- this could be multiple user messages, Actions, etc
  SlotSetEvents
  ACTION_LISTEN
  next_intent
]
```

Unhappy path:

```
[ ACTION_LISTEN, 
  intent_which_triggers_form,
  FormEvent <- this could be multiple user messages, Actions, etc
  ACTION_LISTEN
  SomeUnhappyIntent
  FormEvent
] 
```

The necessary logic should be implemented in `DialogueStateTracker.applied_events()`

PR #6029, opened 2020-06-16
New Form featurization

wochinge:
**Proposed changes**:
- adapt form featurization so that the happy path is represented by one action
- fix https://github.com/RasaHQ/rasa/issues/6015


rasa/core/policies/rule_policy.py
```diff
@@ -153,6 +153,26 @@ def predict_action_probabilities(
         last_action_was_rejection = active_form_name and tracker.events[
             -1
         ] == ActionExecutionRejected(active_form_name)
+        should_predict_form = (
```
wochinge, line 156: reverted these changes due to some failing tests. Will drop it later again

rasa/core/trackers.py
```diff
@@ -398,7 +401,9 @@ def generate_all_prior_trackers(
     def applied_events(self) -> List[Event]:
```
wochinge, line 401: @Ghostvv Could you have a short look please? I also added some tests to `test_trackers` where you can see how the `applied_events()` vs the raw tracker

rasa/core/trackers.py
```diff
@@ -420,11 +432,67 @@ def undo_till_previous(event_type, done_events):
...
+                    event.action_name, applied_events
+                )
+            ):
+                self.undo_till_previous_loop_execution(
+                    event.action_name, applied_events
+                )
+            elif isinstance(event, ActionExecutionRejected):
+                # Undo previously executed action
+                undo_till_previous(ActionExecuted, applied_events)
+                applied_events.append(event)
```
Ghostvv, line 449: why we didn't do it before? how it was handled?

Ghostvv: I think the `ActionExecuted` event is not added in this case, so you revert the wrong action

rasa/core/trackers.py
```diff
@@ -420,11 +432,67 @@ def undo_till_previous(event_type, done_events):
...
             else:
                 applied_events.append(event)
 
         return applied_events
 
+    def _upcoming_rejection(self, index: int) -> bool:
+        return next(
+            (
+                isinstance(e, ActionExecutionRejected)
```
Ghostvv, line 458: I think it is a good idea to check that ActionExecutionRejected corresponds to the active loop name at that time, should be the name of ActionExecuted for which you look for upcoming rejection

rasa/core/trackers.py
```diff
@@ -420,11 +432,67 @@ def undo_till_previous(event_type, done_events):
...
+                for e in list(self.events)[index + 1 :]
+                if isinstance(e, (ActionExecutionRejected, ActionExecuted))
```
Ghostvv, line 460: will it terminate `for` after first `if`, or generate the whole list? we care only for the first `if`

rasa/core/trackers.py
```diff
@@ -420,11 +432,67 @@ def undo_till_previous(event_type, done_events):
                 # the `action_listen` action right before it).
                 undo_till_previous(UserUttered, applied_events)
                 undo_till_previous(ActionExecuted, applied_events)
+            elif (
+                isinstance(event, ActionExecuted)
+                and event.action_name in form_names
+                and not self._upcoming_rejection(idx)
```
Ghostvv, line 438: the action that got immediately rejected is not added to the `tracker.events` (unless you changed that, so you're looking for rejection for different action

rasa/core/trackers.py
```diff
@@ -420,11 +432,67 @@ def undo_till_previous(event_type, done_events):
...
+                and not self._first_loop_execution_or_previous_rejection(
```
Ghostvv, line 439: if you account for above, I don't think there is a need for this check

rasa/core/trackers.py
Ghostvv: left comments

rasa/core/trackers.py
wochinge: Ah, I didn't know that. Will fix it 👍

rasa/core/trackers.py
wochinge: it's a generator so it'll go for the first `if`

rasa/core/policies/rule_policy.py
wochinge: kept the revert for laziness as I don't want to adapt the tests now 😆

rasa/core/trackers.py
```diff
@@ -420,11 +432,61 @@ def undo_till_previous(event_type, done_events):
...
+            if (
+                isinstance(event, ActionExecuted)
+                and event.action_name == loop_action_name
+            ):
+                return False
+
+        return True
+
+    @staticmethod
+    def undo_till_previous_loop_execution(
```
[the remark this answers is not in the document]
wochinge: Good point. I'll check what the best practices are and make it consistent then.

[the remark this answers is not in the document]
wochinge: I'll make both private class methods since there is no need to protect this function from the rest of the world and they are no closures (https://stackabuse.com/python-nested-functions)

rasa/core/policies/rule_policy.py
Ghostvv: I'm updating this

wochinge: yep. Reverted my changes in this file so that it's the same as in the current `rules` branch

rasa/core/trackers.py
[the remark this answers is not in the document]
Ghostvv:
```suggestion
    def _undo_till_previous_loop_execution(
```
I think this should be private

PR #6030, opened 2020-06-16
RulePolicy: scope dispatches per form

### LINKED ISSUE #6017: RulePolicy: scope dispatches per form
**Current State**
As part of the `RulePolicy` we moved the `FormAction` from the Rasa SDK to Rasa Open Source. 
Whenever we request a slot, we currently dispatch an utterance to ask for this slot (see [here](https://github.com/RasaHQ/rasa/blob/da148ae98cbd0c03c8c5a48161622f700e15ec22/rasa/core/actions/forms.py#L476)). The utterance follows the template `utter_ask_<slot name>`.
The problem is that if a slot is part of multiple forms, we always use the same utterance to ask for the slot although the slot might have slightly different meanings in the form.

**Proposed Changes**
* Check if there is a `utter_ask_<form_name>_<slot_name>` in the `domain`. If yes, dispatch it. If no, dispatch `utter_ask_<slot_name>`

alwx:
**Proposed changes**:
- fix #6017

### DIFF
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -462,17 +462,24 @@ class FormAction(LoopAction):
         # no more required slots to fill
         return [SlotSet(REQUESTED_SLOT, None)]
 
-    @staticmethod
+    def _name_of_utterance(self, domain: Domain, slot_name: Text) -> Text:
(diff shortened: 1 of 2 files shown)


alwx:
It seems simple, so I am not entirely sure if I took everything into consideration. @wochinge could you please check and let me know?

Issue #6032, opened 2020-06-16
Warning about intents not in domain is shown when sending a message to /model/parse while running rasa with an NLU-only model


chkoss:
**Issue**:
When sending a message to /model/parse while running rasa with an NLU-only model, the following warning is shown for the recognised intent:
`UserWarning: Interpreter parsed an intent 'greet' which is not defined in the domain. Please make sure all intents are listed in the domain.
  More info at https://rasa.com/docs/rasa/core/domains/`

**Command or request that led to error**:
```
rasa train nlu
rasa run --enable-api -m models/nlu-20200616-181712.tar.gz
curl localhost:5005/model/parse -d '{"text":"hi"}
```

muximuxi:
same error

PR #6034, opened 2020-06-16
fix log_unseen_features warning for nlu-only models

erohmensing:
**Proposed changes**:
- exit function early if no domain is provided (if you train with `rasa train nlu`), since the point is to check against the domain
- fix https://github.com/RasaHQ/rasa/issues/6032

### DIFF
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -420,34 +420,40 @@ class MessageProcessor:
             logger.debug(f"Current slot values: \n{slot_values}")
 
     def _log_unseen_features(self, parse_data: Dict[Text, Any]) -> None:
-        """Check if the NLU interpreter picks up intents or entities that aren't
-        recognized."""
+        """Logs unrecognized features from the NLU parse data.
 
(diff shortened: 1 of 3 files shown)


rasa/core/processor.py
```diff
@@ -420,17 +420,23 @@ def _log_slots(tracker) -> None:
             logger.debug(f"Current slot values: \n{slot_values}")
 
     def _log_unseen_features(self, parse_data: Dict[Text, Any]) -> None:
-        """Check if the NLU interpreter picks up intents or entities that aren't
-        recognized."""
+        """Logs unrecognized features from the NLU parse data.
```
chkoss, line 423: "Log" is a little weird, because nothing is actually logged. Maybe we should rename it to `_check_for_unseen_features` and adjust the doc string accordingly.

erohmensing: oh true, it doesn't log but actually raises warnings! I think this is leftover from when we used `logger.warning`, good catch :)

PR #5977, opened 2020-06-09
Changed duckling extractor to return entity values as string

davidezanella:
Signed-off-by: Davide <davide.zanella10@gmail.com>

**Proposed changes**:
- let duckling return always string values, in order to allow entities extracted by it to be tested. With the actual rasa version in the following part of the test story, the entities cannot be tested correctly beacuse of the different types of the values:
```
* start_cooking: let's cook the [second]{"entity": "ordinal", "value": "2"} [one]{"entity": "number", "value": "1"}   <!-- predicted: start_cooking: let's cook the [second]{"entity": "ordinal", "value": 2} [one]{"entity": "number", "value": 1} -->
```
With the changes made, all the values are returned as a string and can be tested in the `conversation_tests.md` file, otherwise, they are reported as wrong predictions while testing.

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6042, opened 2020-06-18
Convert entity values from Duckling to strings for testing

tabergma:
**Proposed changes**:
In our tests all entity values are of type string. However, values from duckling can be of any type. In order to correctly compare predicted and target entities, we need to convert the entity values from duckling to strings during our evaluation.

related to https://github.com/RasaHQ/rasa/pull/5977

### DIFF
--- a/rasa/core/test.py
+++ b/rasa/core/test.py
@@ -6,6 +6,14 @@ from collections import defaultdict, namedtuple
 from typing import Any, Dict, List, Optional, Text, Tuple, Union
 
 import rasa.utils.io as io_utils
+from rasa.nlu.constants import (
+    EXTRACTOR,
+    ENTITY_ATTRIBUTE_VALUE,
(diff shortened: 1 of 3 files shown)


rasa/core/test.py
```diff
@@ -215,9 +223,22 @@ def _clean_entity_results(
...
+        ):
             if k in set(r):
+                if (
+                    k == ENTITY_ATTRIBUTE_VALUE
+                    and EXTRACTOR in set(r)
+                    and r[EXTRACTOR] == "DucklingHTTPExtractor"
+                ):
+                    # convert values from duckling to strings for evaluation as
+                    # target values are all of type string
+                    r[k] = str(r[k])
```
erohmensing, line 241: Re the conversation we had in slack, i'm not sure if it makes more sense to change it for all extractors (e.g. if someone created a custom RegexExtractor that pulled a number in `int` instead of str)

PR #6045, opened 2020-06-19
Separate sequence and sentence features

tabergma:
**Proposed changes**:
- Separate sequence and sentence features.
- DIET process sequence and sentence features independently. The features are concatenated just before the transformer.
- This allows us to have different kind of features for the sequence and the sentence. 
- `LexicalSyntacticFeaturizer` does not produce sentence features anymore.
- `ConveRTFeaturizer` does not duplicate the sequence features anymore as they don't need to fit the size of the sentence features anymore. 

closes https://github.com/RasaHQ/research/issues/79


docs/api/custom-nlu-components.rst
```diff
@@ -42,9 +42,10 @@ methods that you should implement:
     which is needed further down the pipeline.
 
 .. note::
-    If you create a custom featurizer you should return a sequence of features.
-    E.g. your featurizer should return a matrix of size (number-of-tokens x feature-dimension).
-    The feature vector of the ``__CLS__`` token should contain features for the complete message.
+    If you create a custom featurizer you can return two different kind of features: sequence features and sentence
+    features. The sequence features are a matrix of size ``(number-of-tokens - 1 x feature-dimension)``, e.g.
+    the matrix contains a feature vector for every token in the sequence except the ``__CLS__`` token.
```
Ghostvv, line 47: I don't think we need `__CLS__` token

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1079,21 +1189,27 @@ def _ordered_tag_specs(
...
                 raise InvalidConfigError(
                     f"No label features specified. "
                     f"Cannot train '{self.__class__.__name__}' model."
                 )
             if (
                 self.config[SHARE_HIDDEN_LAYERS]
-                and self.data_signature[TEXT_FEATURES]
-                != self.data_signature[LABEL_FEATURES]
+                and self.data_signature[TEXT_SENTENCE_FEATURES]
+                != self.data_signature[LABEL_SENTENCE_FEATURES]
```
Ghostvv, line 1212: why we check only sentence features here?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1203,25 +1319,36 @@ def _prepare_sparse_dense_layers(
...
+            self._tf_layers[
+                f"dense_input_dropout.{name}_{feature_type}"
+            ] = tf.keras.layers.Dropout(rate=self.config[DROP_RATE])
+            self._prepare_sparse_dense_layers(
+                self.data_signature[f"{name}_{feature_type}_features"],
+                f"{name}_{feature_type}",
+                self.config[REGULARIZATION_CONSTANT],
+                self.config[DENSE_DIMENSION][name],
+            )
+            self._tf_layers[f"ffnn.{name}_{feature_type}"] = layers.Ffnn(
```
Ghostvv, line 1345: I think the name of the layer `ffnn` is confusing here, because it clashes with ffnn before transformer

docs/api/custom-nlu-components.rst
tabergma: removed the CLS token

rasa/nlu/classifiers/diet_classifier.py
tabergma: Should not clash: We first have `ffnn.text_sequence` and `ffnn.text_sentence` and then right before the transformer we have `ffnn.text`. Do you think we should add something to the name?

rasa/nlu/classifiers/diet_classifier.py
tabergma: good point, will add a check for sequence features as well

rasa/nlu/classifiers/diet_classifier.py
Ghostvv: I meant confusing from a reader of the code point of view, I would call it `concat_layer` and use a `Dense` layer instead of `Ffnn` one, so that the name explicitly state its purpose

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -446,50 +468,86 @@ def _extract_features(
...
+                dense_sentence_features.append(_dense_sentence)
+
+        sparse_sequence_features = np.array(sparse_sequence_features)
+        sparse_sentence_features = np.array(sparse_sentence_features)
+        dense_sequence_features = np.array(dense_sequence_features)
+        dense_sentence_features = np.array(dense_sentence_features)
 
-        return [sparse_features, dense_features]
+        return (
+            [sparse_sequence_features, dense_sequence_features],
```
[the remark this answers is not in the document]
tabergma: not sure if that makes things easier, would leave it to the refactoring issue around rasa model data

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1513,7 +1750,18 @@ def batch_loss(
     ) -> tf.Tensor:
         tf_batch_data = self.batch_to_model_data_format(batch_in, self.data_signature)
 
-        sequence_lengths = self._get_sequence_lengths(tf_batch_data[TEXT_SEQ_LENGTH][0])
+        mask_sequence_text = self._get_mask_for(tf_batch_data, TEXT_SEQUENCE_LENGTH)
+        mask_sentence_text = self._get_mask_for(tf_batch_data, TEXT_SENTENCE_LENGTH)
```
Ghostvv, line 1754: why do we need a mask for sentence? I thought it always of length 1

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1513,7 +1750,18 @@ def batch_loss(
...
+
+        if TEXT_SEQUENCE_LENGTH in tf_batch_data:
+            sequence_lengths = self._get_sequence_lengths(
+                tf_batch_data[TEXT_SEQUENCE_LENGTH][0]
+            )
+            sequence_lengths += 1  # add cls token
```
Ghostvv, line 1760: I think it should be smth like
```suggestion
            sequence_lengths += 1  # add sentence features
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1513,7 +1750,18 @@ def batch_loss(
...
+        else:
+            sequence_lengths = self._get_sequence_lengths(
+                tf_batch_data[TEXT_SENTENCE_LENGTH][0]
+            )
```
Ghostvv, line 1764: should this simply be `1`, then we add `sequence_lengths` if `TEXT_SEQUENCE_LENGTH` exists

Ghostvv: same for labels

rasa/nlu/classifiers/diet_classifier.py
tabergma: it is, we can also remove it

rasa/nlu/classifiers/diet_classifier.py
tabergma: unfortunately that does not work for response selector, setting the `sequence_lengths` to 1 works for DIET but for the Reponse Selector it causes a share mismatch `Dimensions of inputs should match: shape[0] = [1,1] vs. shape[1] = [64,1]`. So we need to know the batch dimension, but we cannot get that during `batch_loss`.

Ghostvv: who raises this error?

tabergma: solved it, was using `tensor.shape[0]`, which is not working in non-eager mode, switched to `tf.shape(tensor)[0]`.

tabergma:
@Ghostvv Update the PR, looks fine from my side now. Currently testing one more time a couple of configurations locally.

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1203,25 +1315,36 @@ def _prepare_sparse_dense_layers(
...
-            rate=self.config[DROP_RATE]
-        )
-        self._prepare_sparse_dense_layers(
-            self.data_signature[f"{name}_features"],
-            name,
-            self.config[REGULARIZATION_CONSTANT],
-            self.config[DENSE_DIMENSION][name],
-        )
-        self._tf_layers[f"ffnn.{name}"] = layers.Ffnn(
+        self._tf_layers[f"dense_layer.{name}"] = layers.Ffnn(
```
Ghostvv, line 1318: sorry, I'd keep `ffnn` here. I only meant to rename concat layer
```suggestion
        self._tf_layers[f"ffnn.{name}"] = layers.Ffnn(
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1340,35 +1469,120 @@ def _features_as_seq_ids(
...
+            sentence_features,
+            sequence_mask,
+            text_mask,
+            name,
+            sparse_dropout,
+            dense_dropout,
         )
         x = tf.reduce_sum(x, axis=1)  # convert to bag-of-words
-        return self._tf_layers[f"ffnn.{name}"](x, self._training)
+        return self._tf_layers[f"dense_layer.{name}"](x, self._training)
```
Ghostvv, line 1579:
```suggestion
        return self._tf_layers[f"ffnn.{name}"](x, self._training)
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -586,24 +673,42 @@ def _create_model_data(
...
-        if label_attribute and model_data.feature_not_exist(LABEL_FEATURES):
+        if (
+            label_attribute
+            and model_data.feature_not_exist(LABEL_SENTENCE_FEATURES)
+            and model_data.feature_not_exist(LABEL_SEQUENCE_FEATURES)
+        ):
             # no label features are present, get default features from _label_data
             model_data.add_features(
-                LABEL_FEATURES, self._use_default_label_features(label_ids)
+                LABEL_SEQUENCE_FEATURES, self._use_default_label_features(label_ids)
```
Ghostvv, line 711: shouldn't the default features be sentence features?
```suggestion
                LABEL_SENTENCE_FEATURES, self._use_default_label_features(label_ids)
```

PR #6046, opened 2020-06-19
Update JWT documentation

federicotdn:
**Proposed changes**:
- Update documentation detailing how to use JWT token authentication.


docs/user-guide/configuring-http-api.rst
```diff
@@ -220,6 +221,16 @@ Your requests should have set a proper JWT header:
...
+The following is an example payload for a JWT token:
+
+.. code-block:: json
+
+      {
+          "user": {
+              "username": "<sender_id>",
+              "role": "user"
+          }
+      }
```
akelad, line 233: Potentially dumb question, but where are they sending this payload?

federicotdn: Only the Rasa Open Source server is processing it, and it only applies for endpoints enabled with `--enable-api`. No relationship with input channels.

akelad: but does Rasa Open Source have an endpoint to generate a JWT token with this payload? Or do they need to create this elsewhere (since from what I see they need to create the JWT token to send as a header like in the block above)

akelad: (e.g. Rasa X does have an endpoint)

federicotdn: The person running the server must create the tokens themselves, and then make sure that whoever is making the request has included the token.

federicotdn: @akelad Do you think this is ready to merge?

akelad: sorry, just getting to this now. I think we should make it clear that people need to generate this themselves, externally from Rasa. Then we're good to merge

PR #6055, opened 2020-06-23
[FEATURE] modify conversation id of server to accomodate viber id

IzidoroBaltazar:
**Proposed changes**:
- Change conversation id format of server api to accommodate Viber channel


sara-tagger:

degiz:
Hey @IzidoroBaltazar 

Could you please give example/point me to the viber doc with their conversation id formats?

IzidoroBaltazar:
Here is general documentation.
https://developers.viber.com/docs/api/rest-bot-api/#get-user-details

From documentation you can see that `=` is used in user_id. We also know that user_id contains `/`, but documentation does not mention that. However we found multiple user_ids that contain `/` which caused problems with current server setup.

I have aggregated few problematic user_ids I see these characters:
```
+/1032547698=ACBEDGFIHKJMLONQPSRUTWVYXZacbedgfihkjmlonqpsrutwvyxz
```
Slash is usually somewhere in the string.

degiz:
@IzidoroBaltazar 

I see, thanks.
The case with `+/1032547698=ACBEDGFIHKJMLONQPSRUTWVYXZacbedgfihkjmlonqpsrutwvyxz` is covered if you use `path` type for the parameter, check [here](https://sanic.readthedocs.io/en/latest/sanic/routing.html)

Would it work for you? Seems easier to maintain compared to the regex.

Could also write an explicit test for this and call it something like `test_post_conversation_id_with_slash`?

IzidoroBaltazar:
I have replaced `regex` with Sanic `path` it's working fine. I have modified the test `test_push_multiple_events`. This test is using problematic `customer_id` characters.

degiz:
> I have modified the test `test_push_multiple_events`

Exactly, we're trying to have one test per problem. So if `test_push_multiple_events` fails in the future bc of special symbols, it will be confusing. That's the reason I'm asking to add an explicit `test_post_conversation_id_with_special_symbols`.

IzidoroBaltazar:
Why is this pull request not merging? Can someone have a look?

PR #6066, opened 2020-06-25
pass interpreter to policies during training and inference

wochinge:
**Proposed changes**:
- pass `interpreter` to Core policies for training
- pass `interpreter` to Core policies during prediction
- implement steps 2 and 3 of the implementation proposal for end-to-end support
- change interface method `Policy::predict_action_probabilities` to accept an `interpreter` argument (+ `kwargs` to make these changes easier in the future)
- change interface method `PolicyEnsemble::probabilities_using_best_policy` to accept an `interpreter` argument  (+ `kwargs` to make these changes easier in the future)


rasa/train.py
```diff
@@ -348,6 +378,7 @@ async def _train_core_with_validated_data(
     train_path: Optional[Text] = None,
     fixed_model_name: Optional[Text] = None,
     additional_arguments: Optional[Dict] = None,
+    interpreter: Optional[Interpreter] = None,
```
chkoss, line 381: optional: looks nicer if additional arguments come after all other arguments
```suggestion
    interpreter: Optional[Interpreter] = None,
    additional_arguments: Optional[Dict] = None,
```

tests/core/test_processor.py
```diff
@@ -663,3 +666,61 @@ async def test_should_predict_another_action(
...
     )
+
+
+def test_get_next_action_probabilities_passes_interpreter_to_policies(
+    monkeypatch: MonkeyPatch,
+):
+    policy = TEDPolicy()
+    test_interpreter = Mock()
+
+    # Some old custom policies might not have the new `interpreter` argument
```
chkoss, line 677:
```suggestion
```

=> That's in the next test

tests/test_train.py
```diff
@@ -1,14 +1,29 @@
...
+from unittest.mock import Mock
 
 import pytest
 from _pytest.capture import CaptureFixture
 from _pytest.monkeypatch import MonkeyPatch
 
 import rasa.model
+import rasa.core
+from rasa import model
+from rasa.constants import DEFAULT_NLU_SUBDIRECTORY_NAME
```
chkoss, line 16:
```suggestion
```

tests/test_train.py
```diff
@@ -1,14 +1,29 @@
...
+from rasa.core.interpreter import RasaNLUInterpreter
+from rasa.nlu.model import Interpreter
```
chkoss, line 18:
```suggestion
```

rasa/core/policies/ensemble.py
```diff
@@ -466,6 +479,34 @@ def _best_policy_prediction(
 
         return self._pick_best_policy(predictions)
 
+    @staticmethod
+    def _get_prediction(
+        policy: Policy,
+        tracker: DialogueStateTracker,
+        domain: Domain,
+        interpreter: NaturalLanguageInterpreter,
+        priority: int,
```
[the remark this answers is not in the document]
wochinge: good catch 👍  🙈

rasa/train.py
wochinge: I don't want to change the order in case somebody is passing everything as arguments (not as key word argument) and the changed order might break things for them. `train_core` is somewhat part of the public Python interface)

rasa/train.py
```diff
@@ -222,17 +229,37 @@ async def _do_training(
...
+    if fingerprint_comparison_result.should_retrain_nlu():
+        model_path = await _train_nlu_with_validated_data(
+            file_importer,
+            output=output_path,
+            train_path=train_path,
+            fixed_model_name=fixed_model_name,
+            persist_nlu_training_data=persist_nlu_training_data,
+            additional_arguments=nlu_additional_arguments,
+        )
+        interpreter_path = os.path.join(model_path, DEFAULT_NLU_SUBDIRECTORY_NAME)
```
evgeniiaraz, line 247: I think `model_path` can be a `*.tar.gz` path. In this case `os.path.join()` will not provide a proper path, I think.

evgeniiaraz: (And I have a feeling unpacking it here, just after saving, is a bit of a stupid solution)

rasa/core/processor.py
```diff
@@ -803,5 +802,5 @@ def _get_next_action_probabilities(
                 )
 
         return self.policy_ensemble.probabilities_using_best_policy(
-            tracker, self.domain
+            tracker, self.domain, self.interpreter
```
evgeniiaraz, line 805: I have a doubt here which I would be very happy to be wrong :) 
I think, with current setting, the self.interpreter will always be `RegexInterpreter`. it will be changed only if we're using the current e2e testing (I mean, this e2e: https://rasa.com/docs/rasa/user-guide/testing-your-assistant/).

evgeniiaraz: I don't think I can make a suggestion there but in `rasa/test.py` in `test_core` function removing the check for use_e2e should do the trick :)

rasa/train.py
wochinge: it can return a `tar.gz`, but only when NLU was trained on its own: https://github.com/RasaHQ/rasa/blob/bd8cc1e633e604ca505ab0c857fe2bd2b932d8fc/rasa/train.py#L543

rasa/core/processor.py
wochinge: Haha, now I was bit confused myself 😆 The `interpreter` in processor is the same one which is already used for parsing the incoming NLU messages. So `self.interpreter` is definitely not the `RegexInterpreter`, but the one which is loaded by `Agent`

rasa/core/policies/ensemble.py
```diff
@@ -117,11 +118,14 @@ def train(
...
         domain: Domain,
+        interpreter: NaturalLanguageInterpreter,
         **kwargs: Any,
     ) -> None:
         if training_trackers:
             for policy in self.policies:
-                policy.train(training_trackers, domain, **kwargs)
+                policy.train(
+                    training_trackers, domain, interpreter=interpreter, **kwargs
+                )
```
chkoss, line 128:
```suggestion
                number_of_arguments_in_rasa_1_0 = 3
                arguments = common_utils.arguments_of(policy.predict_action_probabilities)
                if (
                    len(arguments) > number_of_arguments_in_rasa_1_0
                    and "interpreter" in arguments
                ):
                    policy.train(
                        training_trackers, domain, interpreter=interpreter, **kwargs
                    )
                else:
                    common_utils.raise_warning(
                        "The function `train` of "
                        "the `Policy` interface was changed to support "
                        "additional parameters. Please make sure to "
                        "adapt your custom `Policy` implementation.",
                        category=DeprecationWarning,
                    )
                    policy.train(
                        training_trackers, domain, **kwargs
                    )
```

wochinge: we don't have to change `train` as the `Policy` interface already includes the `**kwargs` argument. Custom polices will  hence work out of the box.

chkoss: Ah, of course!

rasa/core/processor.py
evgeniiaraz: Aaa 🤯
I am sorry to bother you with this but I need help finding flaws in my thinking :) 
So, e.g., when we do `rasa test core`, we get to `test_core` function [here](https://github.com/RasaHQ/rasa/blob/a706b30f86d343af26afb719e0447201eb9ff4e5/rasa/test.py#L146-L160). If `use_e2e` flag is false, `_interpreter` is `RegexInterpreter`  and that is passed to `Agent.load`.  And within `Agent.load` the new interpreter will not be loaded as we have passed the RegexInterpreter. 
(I really want to see where I'm wrong 🙈)

wochinge: You're right! I didn't think of testing - only of `rasa run`

wochinge: fixed and added a test

evgeniiaraz:
Yayyy! 💥
there might be some little changes depending on what YAML story reader does with its interpreter but with respect to the policies it seems like all done! 🎉

Issue #5905, opened 2020-05-27
Token_pattern is incorrectly applied on incoming text during inference


dakshvar22:
**Rasa version**: 1.10.0

**Python version**: 3.6.5

**Operating system** (windows, osx, ...): MacOS

**Issue**: The original problem is explained [here](https://github.com/RasaHQ/rasa/pull/5038#issue-359224769). It is not a quick fix [because](https://github.com/RasaHQ/rasa/pull/5038#issuecomment-631984638) since we use sequence in our downstream models, we cannot alter the length of the sequence in a featurizer. Hence, `token_pattern` should be added as a parameter inside a tokenizer and removed from countvectorsfeaturizer.

PR #6073, opened 2020-06-26
Move token_pattern to tokenizers

tabergma:
**Proposed changes**:
Remove option ``token_pattern`` from ``CountVectorsFeaturizer``.
Instead all tokenizers now have the option ``token_pattern``.
If a regular expression is set, all detected tokens of the tokenizer are further split into tokens according to the
regular expression.

closes https://github.com/RasaHQ/rasa/issues/5905


rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -105,6 +108,44 @@ def _split_intent(self, message: Message):
...
+        Args:
+            tokens: list of tokens to split
+
+        Returns:
+            List of tokens.
+        """
+        if not self.token_pattern:
+            return tokens
+
+        token_pattern = re.compile(self.token_pattern)
```
dakshvar22, line 123: Let's move this to the constructor because it's a one time activity and also a costly operation.

changelog/5905.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Remove option ``token_pattern`` from ``CountVectorsFeaturizer``.
+Instead all tokenizers now have the option ``token_pattern``.
+If a regular expression is set, all detected tokens of the tokenizer are further split into tokens according to the
+regular expression.
```
dakshvar22, line 4: This is not always correct now right? Because for tokenizer like spacy we first use spacy's tokenizer and then apply the token pattern, whereas for convert and language model tokenizers we first apply the token pattern and then use the model's tokenizer. I am a bit confused about how to phrase this without making it sound complex.

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -82,7 +84,8 @@ def tokenize(self, message: Message, attribute: Text) -> List[Token]:
 
         return tokens_out
 
-    def _clean_tokens(self, tokens: List[bytes]):
+    @staticmethod
+    def _clean_tokens(tokens: List[bytes]):
```
dakshvar22, line 88:
```suggestion
    def _clean_tokens(tokens: List[bytes]) -> List[Text]:
```

changelog/5905.bugfix.rst
tabergma: right... let me think about it

PR #6091, opened 2020-06-29
prototype implementation of commandline based deploy

tmbo:
### Prototype: Deploying an assistant to the cloud using a rasa-CLI command 
Goal: Make it as easy as possible to get a local assistant running in the cloud
(ideally with rasa-x deployed alongside it). No installation of anything locally but `rasa` (so no docker, helm, ... needed locally).

#### Approach:
- use gcloud, because I know it and it has a lot of services to play with
- it does not seem to be possible without ANY user interaction, the user needs to
  manually create the gcloud project and create a service key
- uses terraform to create a new cluster and setup the necessary hardware. after that is finished, cloudbuild is used to build any needed docker containers (e.g. for custom actions in the project). in the end, the project is deployed using our rasa-x helm charts.

#### How to try it out:
You will need to create a fresh google cloud project and a service account with the OWNER role in that project (that is just my laziness, you actually need a lot fewer permissions, but God knows which 🤷 ). The
script will walk you through the creation of these.

- checkout this branch `git checkout rasa-deploy`
- create a project with `rasa init` (I needed to move the default location of actions, so better to create a new project)
- navigate in to your project and run `rasa train` (better if you have a model)
- run `rasa deploy`

#### Warning!

This is really an optimal-case implementation: if you don't play by the rules it will
break 🔥😅. PLEASE DO NOT RUN THIS IN A GOOGLE PROJECT THAT
YOU WOULD NOT FEEL COMFORTABLE DELETING. It is a prototype and things in that project
might go wrong which would be a mess to clean up manually if you still need that project.

In case you play by the rules you might get lucky and end up with a
proper deployment after about 13 minutes
(should be faster on future runs on the same project, about 3-4 min).


tczekajlo:
An idea, wasn't it be simpler to move terraform to cloudbuild? Then we don't need to have installed terraform. 

Btw, a service account and a key for the account can be created by terraform as well.

tmbo:
mhm not sure what you mean 🤔  terraform is already running as part of one of the cloudbuild jobs, the user doesn't need to install it locally (nor do they need to install helm / docker)

tczekajlo:
> mhm not sure what you mean 🤔 terraform is already running as part of one of the cloudbuild jobs, the user doesn't need to install it locally (nor do they need to install helm / docker)

Ah, cool, "uses terraform to create a new cluster and setup the necessary hardware" - I understood it that we run in on a local machine/user side 🤦

rasa/__main__.py
```diff
@@ -23,6 +23,8 @@
     train,
     visualize,
     x,
+    export,
```
tmbo, line 26:
```suggestion
```

tmbo:
@chandrikas I guess this one can be closed?

chandrikas:
Yes, I think it's ok to be closed :-)

tmbo:
😂 hi there - hope you are doing well and stay as amazing as you have always been! all the best 👋

PR #6097, opened 2020-06-30
fix breaking tests in rules branch

wochinge:
**Proposed changes**:
- FormPolicy **and** RulePolicy satisfy requirement if forms given in domain
- Rasa default intents (e.g. `restart`) overrule RulePolicy and forms
- fix form emulation in end to end tests for RulePolicy
- fix some tiny bugs
- fix form featurization and the failing `TestFormPolicy` tests


wochinge:
@Ghostvv I started fixing some of the failing tests. 

Memoization is currently broken for unhappy paths due to the new form featurization. The problem is that the `ActionRejected` is missing in the training stories for Memoization and hence `applied_events` assumes everything is still part of the `happy path` which in turn leads to wrong states. 

Solution I could imagine:
- fix this in the data reader by injecting `ActionExecutedRejected` events whenever an intent **after** executing a form isn't prefixed with `form: ` 
- let users add the rejection event manually (not ideal since this would require changes on their side)
- add a flag to `applied_events` to turn off / on the new featurization for old policies.

Ghostvv:
I think vice versa. Since all the happy path of a form is hidden now, training stories always represent unhappy path, so it should be fine for memo policy. Our change of form featurization doesn't influence training stories

wochinge:
> I think vice versa. Training stories always represent unhappy path, so it should be fine for memo policy. Our change of form featurization doesn't influence training stories 

@Ghostvv But the test for the unhappy paths is currently broken and it's definitely caused by the change in `applied_events`

Ghostvv:
> But the test for the unhappy paths is currently broken
could you please point me to the test?

wochinge:
Omg, new things started to fail 🤯  I'll quickly fix and push again so that we have the actual error.

wochinge:
@Ghostvv Just pushed the fixes. It should fail on `TestFormPolicy`

rasa/core/featurizers.py
```diff
@@ -563,7 +563,7 @@ def __init__(
     ) -> None:
 
         super().__init__(state_featurizer, use_intent_probabilities)
-        self.max_history = max_history
+        self.max_history = max_history or self.MAX_HISTORY_DEFAULT
```
Ghostvv, line 566: this will not allow set it to `None`, which we need for `RulePolicy`

rasa/core/trackers.py
```diff
@@ -384,30 +385,48 @@ def _undo_till_previous(event_type: Type[Event], done_events: List[Event]) -> No
...
-                isinstance(event, ActionExecutionRejected)
-                and event.action_name == loop_action_name
-            ):
+            form_was_not_active = isinstance(event, Form) and event.name is None
+            if form_was_not_active:
                 return True
 
-            # Form might have run before but was deactivated and is now running again.
-            if isinstance(event, Form) and event.name is None:
+            if DialogueStateTracker._is_unhappy_path(
```
Ghostvv, line 398: do we need this method to be static? I think `self. _is_unhappy_path` is easier

rasa/core/featurizers.py
wochinge: We explicitly set it to `None` in the `RulePolicy` by passing in the initialized featurizer (as opposed to `None` as before)

rasa/core/policies/rule_policy.py
```diff
@@ -1,20 +1,37 @@
...
+    ACTION_LISTEN_NAME,
+    ACTION_RESTART_NAME,
+    ACTION_BACK_NAME,
+    ACTION_SESSION_START_NAME,
+)
 
 logger = logging.getLogger(__name__)
 
+# These are Rasa Open Source default actions and overrule everything at any time.
+DEFAULT_ACTION_MAPPINGS = {
```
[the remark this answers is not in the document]
wochinge: True, but that's up to the users, correct? We don't have to write the rules for that.

rasa/core/trackers.py
```diff
@@ -384,30 +385,48 @@ def _undo_till_previous(event_type: Type[Event], done_events: List[Event]) -> No
...
+            form_ran_before = event.action_name == loop_action_name
+            if form_ran_before:
                 return False
 
+            next_action_in_the_future = event.action_name
+
         return True
 
+    @staticmethod
+    def _is_unhappy_path(
```
Ghostvv, line 415: I think it might be correct, but the logic is confusing. The name suggests that you check the whole path, while this method checks event by event

rasa/core/trackers.py
```diff
@@ -384,30 +385,48 @@ def _undo_till_previous(event_type: Type[Event], done_events: List[Event]) -> No
...
     ) -> bool:
+        next_action_in_the_future: Optional[Text] = None
+
         for event in reversed(applied_events):
-            # Form was rejected previously.
-            if (
-                isinstance(event, ActionExecutionRejected)
-                and event.action_name == loop_action_name
-            ):
+            form_was_not_active = isinstance(event, Form) and event.name is None
```
Ghostvv, line 394: I think the name doesn't represent what this bool actually means

rasa/core/featurizers.py
Ghostvv: hm, I think resetting to None is less intuitive. But it's not super important, e2e will change that

Ghostvv: in e2e it is None by default

rasa/core/policies/rule_policy.py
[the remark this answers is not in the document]
Ghostvv: oh, I wanted to say we shouldn't forget to update the docs there, somehow "docs" disappeared during typing 😅

rasa/core/featurizers.py
wochinge: Ok 👍 I basically just changed it back to the `master` version and then did the "hack" in the RulePolicy to pass in `None`. Will leave it like that then for now.

rasa/core/trackers.py
wochinge: `might be correct` :-D very convincing 😆

rasa/core/trackers.py
```diff
@@ -383,31 +384,53 @@ def _undo_till_previous(event_type: Type[Event], done_events: List[Event]) -> No
             if isinstance(e, event_type):
                 break
 
-    @staticmethod
-    def _first_loop_execution_or_previous_rejection(
-        loop_action_name: Text, applied_events: List[Event]
+    def _first_loop_execution_or_unhappy_path(
```
wochinge, line 387: @Ghostvv please re-check

rasa/core/test.py
```diff
@@ -326,12 +314,22 @@ def _collect_action_executed_predictions(
...
+                processor.domain, partial_tracker, predicted
+            )
+        ):
+            # Wrong action was predicted,
+            # but it might be Ok if form action is rejected.
+            _emulate_form_rejection(partial_tracker)
             # try again
             action, policy, confidence = processor.predict_next_action(partial_tracker)
+
+            # Even if the prediction is also wrong, we don't have to undo the emulation
```
wochinge, line 330: @Ghostvv please re-check

wochinge:
Will quickly implement the mutual exclusion check in here as well

wochinge:
added the mutual exclusion check

rasa/core/trackers.py
```diff
@@ -383,31 +384,53 @@ def _undo_till_previous(event_type: Type[Event], done_events: List[Event]) -> No
...
+        next_action: Optional[Text] = None
+
         for event in reversed(applied_events):
-            # Form was rejected previously.
-            if (
-                isinstance(event, ActionExecutionRejected)
-                and event.action_name == loop_action_name
-            ):
+            # If there was a form deactivation we know that the form is running for
+            # the first time.
```
Ghostvv, line 394: Rather smth along the lines:
We should look for previous form actions only until previous form deactivation event, because it means that further there was a different form.

PR #6103, opened 2020-07-01
Prepare release 1.10.4

tabergma:
**Proposed changes**:
Prepare release 1.10.4

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -21,7 +21,6 @@ env:
   # needed to fix issues with boto during testing:
   # https://github.com/travis-ci/travis-ci/issues/7940
   BOTO_CONFIG: /dev/null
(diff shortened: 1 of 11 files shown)


tabergma:
@wochinge I updated the dependencies via `poetry update` as the `pyproject.tmol` changed, but now installation fails. What is the best way to handle this?

wochinge:
Quick fix: Revert the changes to the lock file as not `poetry update` is necessary (only when dependency version changed). However, this will bite the next one then so we have to fix it anyway. Currently looking into it.

wochinge:
@tabergma Maybe this lines are not longer needed with GitHub actions: https://github.com/RasaHQ/rasa/blob/40f6491b650dbbea3b5ac350d1754261c540dccc/.github/workflows/continous-integration.yml#L24

PR #6117, opened 2020-07-01
rename `user: ` key to `intent: ` key

wochinge:
**Proposed changes**:
- rename the `user: ` key to `intent: ` to reflect the latest changes in the yml training data format (`user` will be used for e2e messages)
- warn if intent is prefixed with `/` as this is not longer needed
- drop some assumptions about e2e


rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -161,17 +156,16 @@ def _parse_raw_user_utterance(self, step: Dict[Text, Any]) -> Optional[UserUtter
...
 
-        if not user_utterance.startswith(INTENT_MESSAGE_PREFIX):
+        if user_utterance.startswith(INTENT_MESSAGE_PREFIX):
             common_utils.raise_warning(
                 f"Issue found in '{self.source_name}': \n"
-                f"User intent '{user_utterance}' should start with "
-                f"'{INTENT_MESSAGE_PREFIX}'. "
-                f"This story step will be skipped.",
+                f"User intent '{user_utterance}' starts with "
+                f"'{INTENT_MESSAGE_PREFIX}'. This is not required.",
```
wochinge, line 163: should we phrase that differently?

degiz: The phrasing is fine. I'm just thinking what if users will accidentally confuse `intent` key with a `user` one?
Should we check if that's a single word after stripping the spaces?

tests/core/training/story_reader/test_yaml_story_reader.py
```diff
@@ -93,18 +93,18 @@ async def test_is_yaml_file():
...
     with pytest.warns(UserWarning):
-        await training.load_data(
+        tracker = await training.load_data(
             yaml_file,
             default_domain,
             use_story_concatenation=False,
             tracker_limit=1000,
             remove_duplicates=False,
         )
+        assert tracker[0].latest_message == UserUttered("simple", {"name": "simple"})
```
degiz, line 107: 🚀

wochinge:
Couldn't answer to the comment, so answering here

> The phrasing is fine. I'm just thinking what if users will accidentally confuse intent key with a user one?
Should we check if that's a single word after stripping the spaces?

Good point! I think that's not required because unlike us users will not know that there was ever a `user` key to specify intents. And I think the paraphrasing / e2e part will be more advanced which means that users would learn about the `user` key when they already know about `intent` (so much for the theory 😄 )

degiz:
@wochinge I would at least print a warning for this situation - there's a non-zero chance that it happens, and the users will have no clue what's wrong.

wochinge:
> I would at least print a warning for this situation - there's a non-zero chance that it happens, and the users will have no clue what's wrong.

Yes, but what if the message is something like `Hi`? I think validating against the domain und then not just issue a warning, but actually fail would be great, wdyt? The check would be

```
if key == "user"utterance in domain.intent_names():
   explode()
```

degiz:
> validating against the domain

That sounds great!!

wochinge:
I now understood what you mean 👍 🙈 I had the logic the other way around (which should also be done)

wochinge:
@degiz Could you please have another short look on the last commit? Thanks 🙏

rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -117,27 +134,40 @@ def _parse_step(self, step: Dict[Text, Any]) -> None:
...
+            logger.debug(
+                "Skipped validating if intent is in domain as domain " "is `None`."
+            )
+            return
+
+        if intent_name not in self.domain.intents:
+            common_utils.raise_warning(
+                f"Issue found in '{self.source_name}': \n"
+                f"Found intent '{intent_name}' in stories which is not part of the "
+                f"domain.",
```
degiz, line 155: Should we also mention that this utterance will be skipped then?

wochinge: I don't skip it. I just issue a warning for now.

degiz: AFAIK the same check will happen later and it will be skipped during the training anyway. But I'm fine with it.

wochinge: I tried to find it, but I think only `rasa data validate` will check, right?

wochinge: Markdown also just warns and continues

PR #6121, opened 2020-07-01
Fix the way training data is generated in rasa test nlu -P

### LINKED ISSUE #6120: Rasa test nlu produces wrong partitions on using the percentages argument
**Rasa version**: rasa==1.10.4

**Python version**: Python 3.6.9

**Operating system** (windows, osx, ...): Ubuntu 18

**Issue**:
Currently `rasa test nlu` allows specifying percentages of the training dataset to be excluded while training using the `-P` flag. This isn't currently done correctly.
E.g: If the percentages were `-P 50 25` then you would expect that the number of training samples would be 500 and 250 if the original training dataset size was 1000 samples.
BUT, the way the function currently work will generate number of training samples of 500 and 125 samples.

https://github.com/RasaHQ/rasa/blob/4830308fcc21ba3c82f296e510ba7544fb5166d4/rasa/nlu/test.py#L1848-L1865

```
rasa test nlu -c config.yml config_dup.yml -r 1 -p 50 25
```

I am willing to submit a PR to solve this so kindly assign the issue to me.

Thanks

AMR-KELEG:
Fixes #6120
The command used to exclude percentages from the previous partition
of the training dataset instead of excluding percentages from the
whole training dataset

**Proposed changes**:
- This PR will fix the bug in excluding parts of the training dataset in comparison mode.


sara-tagger:

AMR-KELEG:
Hi @b-quachtran 
Thanks for the review.
I have updated the file.
I couldn't really build Rasa from source yet so I failed to run `black` before creating the PR.
I have used the package as an external package (without building rasa) and seems like this did the job.

PR #6132, opened 2020-07-02
Add possibility to split the domain into separate files

### LINKED ISSUE #5984: New Training Data Format - Domain
Part of #5970

Domain part of the New Training Data Format.
The latest format examples are in the notion.

Features to cover compared to the existing format:
- Add possibility to split the domain into separate files
- `--domain` parameter should accept a path with multiple partial domain YAML files

degiz:
Closes #5984 

**Proposed changes**:
- `domain` file now doesn't have to be called `domain.yaml`


changelog/6132.feature.rst
```diff
@@ -0,0 +1 @@
+Add possibility to split the domain into separate files. Domain file now doesn't have to be called `domain.yaml` anymore.
```
federicotdn, line 1: Maybe we should explain how the user would use the feature. For example:
```suggestion
Added possibility to split the domain into separate files. All YAML files under the path specified with ``--domain`` will be scanned for domain information (e.g. intents, actions, etc) and then combined into a single domain.

The default value for ``--domain`` is still ``domain.yml``.
```

rasa/core/domain.py
```diff
@@ -133,8 +152,9 @@ def from_yaml(cls, yaml: Text) -> "Domain":
         return cls.from_dict(data)
 
     @classmethod
-    def from_dict(cls, data: Dict) -> "Domain":
-        utter_templates = cls.collect_templates(data.get("responses", {}))
+    def from_dict(cls, data: Dict) -> Optional["Domain"]:
+
```
federicotdn, line 156:
```suggestion
```

rasa/core/domain.py
```diff
@@ -133,8 +152,9 @@ def from_yaml(cls, yaml: Text) -> "Domain":
         return cls.from_dict(data)
 
     @classmethod
-    def from_dict(cls, data: Dict) -> "Domain":
-        utter_templates = cls.collect_templates(data.get("responses", {}))
+    def from_dict(cls, data: Dict) -> Optional["Domain"]:
```
federicotdn, line 155: Why is it now `Optional`? Looks like we're always returning `cls(...)`

rasa/core/domain.py
```diff
@@ -187,13 +207,12 @@ def _get_session_config(session_config: Dict) -> SessionConfig:
     @classmethod
     def from_directory(cls, path: Text) -> "Domain":
         """Loads and merges multiple domain files recursively from a directory tree."""
-        from rasa import data
 
         domain = Domain.empty()
         for root, _, files in os.walk(path, followlinks=True):
             for file in files:
                 full_path = os.path.join(root, file)
-                if data.is_domain_file(full_path):
```
federicotdn, line 196: Should we delete `data.is_domain_file` now? I only see one other use, in `multi_project.py`. If we update that use then we should be able to delete it (given that this goes into 2.0 it shouldn't be a problem)

rasa/core/domain.py
```diff
@@ -1074,6 +1093,27 @@ def is_empty(self) -> bool:
...
+
+        if not Path(filename).suffix in YAML_FILE_EXTENSIONS:
+            return False
+        try:
+            content = rasa.utils.io.read_yaml_file(filename)
+            if any(key in content for key in ALL_DOMAIN_KEYS):
+                return True
+        except YAMLError:
+            pass
+
```
federicotdn, line 1116:
```suggestion
return False
```

rasa/core/domain.py
degiz: that's a good point, I think I messed up.

rasa/core/domain.py
degiz: thanks for spotting that, yes, I'll remove the `data::is_domain_file()`.

degiz:
> FutureWarning: No tracker session configuration was found in the loaded domain. Domains without a session config will automatically receive a session expiration time of 60 minutes in Rasa version 2.0 if not configured otherwise.
  docs=DOCS_URL_DOMAINS + "#session-configuration",

I removed the warning and set the default session expiration time to 60 minutes as of now.

degiz:
@federicotdn update: used `DEFAULT_SESSION_EXPIRATION_TIME_IN_MINUTES` everywhere and still keep it as 0 now. This question needs more clarification and I don't want it to block the PR 😬

federicotdn:
:+1: 
Let me know if I should review again

rasa/cli/arguments/default_arguments.py
```diff
@@ -56,7 +56,9 @@ def add_domain_param(
         "--domain",
         type=str,
         default=DEFAULT_DOMAIN_PATH,
-        help="Domain specification (yml file).",
+        help="Domain specification. It can be a single 'yaml' file, or a directory "
+        "that contains several files with domain specification in it. The content "
+        "of these files will be read and merged together.",
```
federicotdn, line 61:
```suggestion
        help="Domain specification. This can be a single YAML file, or a directory "
        "that contains several files with domain specifications in it. The content "
        "of these files will be read and merged together.",
```

PR #6136, opened 2020-07-03
YAML rule and form support

ricwo:
**Proposed changes**:
- support for rules in YAML training data
- support for forms in YAML training data


changelog/6065.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for rule data in YAML format.
```
wochinge, line 1: should we add this or wait until we got docs and everything?

rasa/core/policies/ensemble.py
```diff
@@ -119,59 +119,82 @@ def _check_priorities(self) -> None:
                     docs=DOCS_URL_POLICIES,
                 )
 
-    # TODO: This is a temporary implementation for the RulePolicy prototype.
-    # Policies should be able to decide themselves whether to use rules or stories
-    # for training.
+    def _policy_ensemble_contains_policy_with_rules_support(self) -> bool:
+        """Determine whether the policy ensemble contains at least one policy
```
wochinge, line 123: optional: I'd either add a full docstring (including return type) or no docstring

rasa/core/policies/ensemble.py
```diff
@@ -119,59 +119,82 @@ def _check_priorities(self) -> None:
...
+        return any(tracker.is_rule_tracker for tracker in training_trackers)
 
     def _emit_rule_policy_warning(
-        self, rule_trackers: List[DialogueStateTracker]
+        self, training_trackers: List[DialogueStateTracker]
     ) -> None:
-        """Emit `UserWarning`s about missing `RulePolicy` training data."""
-        rule_policy_is_active = self._policy_ensemble_contains_rule_policy()
+        """Emit `UserWarning`s about missing rule-based data."""
+        rule_policy_is_active = (
```
wochinge, line 150: optional: 
```suggestion
        is_rules_consuming_policy_available = (
```

rasa/core/policies/ensemble.py
```diff
@@ -119,59 +119,82 @@ def _check_priorities(self) -> None:
...
+    ) -> List[DialogueStateTracker]:
+        """Return trackers to be trained on ML-based policies and `RulePolicy`.
+
+        Args:
+            training_trackers: Trackers to split.
+
+        Returns:
+            Trackers from ML-based training data and rule-based data.
+        """
+        supported_data = policy.supported_data()
```
wochinge, line 185: optional: I'm wondering if splitting the training data should actually be the responsibility of the `Ensemble` or rather some function / static method with `trackers` or even a method of `SupportData` 🤔

rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -1,22 +1,25 @@
 from pathlib import Path
-from typing import Dict, Text, List, Any, Optional, Union
+from typing import Dict, Text, List, Any, Optional, Union, Tuple
```
wochinge, line 2: unused
```suggestion
from typing import Dict, Text, List, Any, Optional, Union
```

data/test_yaml_stories/stories_and_rules.yml
```diff
@@ -0,0 +1,51 @@
+rules:
+- rule: rule 1
+  steps:
+  - form: loop_q_form
+  - slot: requested_slot
+    value: some_slot
+  - ...
+  - user: /inform
```
wochinge, line 8: I merged my other PR now, so it should be 
```suggestion
  - intent: inform
```

rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -25,154 +28,202 @@
 
 
 class YAMLStoryReader(StoryReader):
-    """Class that reads the core training data in a YAML format"""
+    """Class that reads Core training data and rule data in YAML format"""
```
wochinge, line 31: actually I think the prior version was already general enough, no?

wochinge, line 31:
```suggestion
    """Class that reads Core training data and rule data in YAML format."""
```

rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -25,154 +28,202 @@
...
+            if not isinstance(item, dict):
                 common_utils.raise_warning(
-                    f"Unexpected block found in '{self.source_name}': \n"
-                    f"{story_item}\n"
-                    f"Items under the '{KEY_STORIES}' key must be YAML dictionaries. "
-                    f"It will be skipped.",
-                    docs=DOCS_URL_STORIES,
+                    f"Unexpected block found in '{self.source_name}':\n"
+                    f"{item}\nItems under the "
+                    f"'{self._get_plural_item_title(is_rule_data)}' key must be YAML "
```
wochinge, line 74: what does the `get_plural` part refer to?

rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -25,154 +28,202 @@
...
+                    docs=self._get_docs_link(is_rule_data),
                 )
                 continue
 
-            if KEY_STORY_NAME in story_item.keys():
-                self._parse_plain_story(story_item)
+            if item_title in item.keys():
+                self._parse_plain_item(item, is_rule_data)
+
+    def _parse_plain_item(self, item: Dict[Text, Any], is_rule_data: bool) -> None:
```
wochinge, line 83: do we actually need to pass through the `is_rule_data` or can we take it from the current `step_builder`?

tests/core/test_ensemble.py
```diff
@@ -377,3 +380,64 @@ def test_from_dict_does_not_change_passed_dict_parameter():
...
+
+def test_rule_based_data_warnings_no_rule_trackers():
+    trackers = [DialogueStateTracker("some-id", slots=[], is_rule_tracker=False)]
+    policies = [RulePolicy()]
+    ensemble = SimplePolicyEnsemble(policies)
+
+    with pytest.warns(UserWarning) as record:
+        ensemble.train(trackers, Domain.empty(), RegexInterpreter())
+
+    assert (
```
wochinge, line 393: optional: I think it's enough to just test for a warning and not for the actual string

rasa/core/training/story_reader/yaml_story_reader.py
wochinge: optional: To be honest, I'm not a huge fan of the [flag argument](https://martinfowler.com/bliki/FlagArgument.html#:~:text=A%20flag%20argument%20is%20a,do%20this%3A%20regular%20and%20premium%20.). It works at this stage (that's why `optional` but makes things 
- complex: additional parameter everywhere + if else -> I think this one could also be solved by moving the flag to the StoryStepBuilder
- inflexible: It currently only makes sense because Rules and Stories are very similar. If you want to deal with unique features of one (e.g. currently you could mix a `story` into `rules` section  or in the future you might want to forbid e2e usage in rules). In that case I think it might make sense to have two classes which inherit from a common base class and can just override certain functionality (ala https://refactoring.guru/replace-conditional-with-polymorphism) . 

On the other hand we are very early stage it might make more sense to find a suited abstraction once the dust settled. Actually, as I'm writing this I'm slightly favoring in waiting 😄 However, still wanted to share my thoughts.

changelog/6065.feature.rst
ricwo: I think it's helpful for now, but we can always consolidate related changelogs later

data/test_yaml_stories/stories_and_rules.yml
ricwo: do you mean in the `rules` branch? doesn't seem to be updated yet

rasa/core/policies/ensemble.py
ricwo: 👍 forgot to add a return type

rasa/core/policies/ensemble.py
ricwo: 👍

rasa/core/training/story_reader/yaml_story_reader.py
ricwo: yes it was but I wanted to change `in a YAML` to `in YAML`. Are you saying I should remove the `and rule data` part again?

rasa/core/training/story_reader/yaml_story_reader.py
ricwo: It returns `stories` as opposed to `story`. Also `rules` instead of `rule`.

rasa/core/training/story_reader/yaml_story_reader.py
ricwo: In this case the flag mainly determines what docstring is used and what warning messages are emitted. IMO this doesn't justify two sets of methods, as it would make the class unnecessarily bloated

tests/core/test_ensemble.py
ricwo: compare this test with the one below - `train()` emits two different warning messages and these two tests are specifically about making sure that the correct one is emitted in each case

rasa/core/policies/ensemble.py
ricwo: 👍  good point, moved it to `SupportedData`

rasa/core/training/story_reader/yaml_story_reader.py
wochinge: >  Are you saying I should remove the and rule data part again?

Yep. `Core training data` sums it up nicely, doesn't it?

rasa/core/training/story_reader/yaml_story_reader.py
wochinge: optional: is the `plural` in the name actually important? Up to you 👍

ricwo: yes because there is another method `_get_item_title()` from which this needs to be distinguished

rasa/core/training/story_reader/yaml_story_reader.py
ricwo: 👍 updated

rasa/core/training/story_reader/yaml_story_reader.py
wochinge: it would just be having one class with the common behavior and having two subclasses which overwrite the docstring content / warning messages, no? 

What do we do about the fact that you can currently mix a `story` step in a `rules` section and vice versa?

ricwo: but we'd be overriding every `_parse_<X>` method?

> What do we do about the fact that you can currently mix a story step in a rules section and vice versa?

this is not an issue right? if we're in the `stories` section and encounter a `- rule:`, it will just be read as a rule

wochinge: > but we'd be overriding every _parse_<X> method?

Nope, as long as we use something like https://en.wikipedia.org/wiki/Template_method_pattern

> this is not an issue right? if we're in the stories section and encounter a - rule:, it will just be read as a rule

That's not ideal imo. But it should probably be taken care of by yaml schema validation. 

Let's keep it like that for now and see how things evolve 👍

PR #6147, opened 2020-07-06
Move option 'case_sensitive' from tokenizers to featurizers

### LINKED ISSUE #5759: Add option case_sensitive to featurizers
**Description of Problem**:
Currently, some of our tokenizers support the option `case_sensitive`. If the user decides to set this option to `False` all featurizer will use the lowercased tokens. This might not be ideal. For example, some of the features in the `LexicalSyntacticFeaturizer` do not work if all tokens are lowercased. 
It might be better to move the option `case_sensitive` to the featurizers itself. Each featurizer can be configured separately. 

**Overview of the Solution**:
Remove the option `case_sensitive` from all tokenizers.
Add the option `case_sensitive` to featurizers for which it makes sense.

tabergma:
**Proposed changes**:
- Remove the option from the ``WhitespaceTokenizer`` and ``ConveRTTokenizer``.
- Add option ``case_sensitive`` to the ``RegexFeaturizer``.

closes #5759 

### DIFF
--- a/rasa/nlu/tokenizers/whitespace_tokenizer.py
+++ b/rasa/nlu/tokenizers/whitespace_tokenizer.py
@@ -14,16 +14,12 @@ class WhitespaceTokenizer(Tokenizer):
         "intent_tokenization_flag": False,
         # Symbol on which intent should be split
         "intent_split_symbol": "_",
-        # Text will be tokenized with case sensitive as default
-        "case_sensitive": True,
     }
 
     def __init__(self, component_config: Dict[Text, Any] = None) -> None:
         """Construct a new tokenizer using the WhitespaceTokenizer framework."""
 
         super().__init__(component_config)
-
-        self.case_sensitive = self.component_config["case_sensitive"]
 
         self.emoji_pattern = self.get_emoji_regex()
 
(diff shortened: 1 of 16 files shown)


rasa/nlu/tokenizers/whitespace_tokenizer.py
```diff
@@ -14,17 +14,13 @@ class WhitespaceTokenizer(Tokenizer):
...
-        # Text will be tokenized with case sensitive as default
-        "case_sensitive": True,
     }
 
     def __init__(self, component_config: Dict[Text, Any] = None) -> None:
         """Construct a new tokenizer using the WhitespaceTokenizer framework."""
 
         super().__init__(component_config)
 
-        self.case_sensitive = self.component_config["case_sensitive"]
```
Ghostvv, line 26: should we add a deprecation warning if this option is present in the config to move it to featurizers?

tabergma: Not sure if it is needed as it goes into Rasa 2.0, but can add a quick warning, that this option is not present anymore.

PR #6159, opened 2020-07-08
Create PRINCIPLES.md

JEM-Mosig:
**Proposed changes**:
- Add the 'Principles for Building Ethical Conversational Assistants' document


PRINCIPLES.md
```diff
@@ -0,0 +1,34 @@
+# Principles for Building Ethical Conversational Assistants
```
lunelson, line 1: We don't need the `h1` tag in the document, because it is specified on the website side

JEM-Mosig: So I delete that line?

lunelson: Yes. I've added that title to the configuration in the website that will pull in the file

lunelson:
Closes https://github.com/RasaHQ/rasa-website/issues/761

lunelson:
@m-vdb this PR only contains one file, I'd like to go ahead and merge it even though there are failed checks, since they don't seem relevant to this change at all, wdyt?

PR #6163, opened 2020-07-08
Improve rasa data split nlu

### LINKED ISSUE #6024: rasa data split nlu: test data should contain at least one example for every label
**Description of Problem**:
When we split NLU data via `rasa data split nlu` it might happen that the test data does not contain any examples for some response labels. 
(related to https://github.com/RasaHQ/rasa/pull/6013)

**Overview of the Solution**:
Make sure to include at least one example per response label in the test data when splitting the data via `rasa data split nlu`.

tabergma:
**Proposed changes**:
`rasa data split nlu`: make sure that at least one example per intent and response is in the test data.

closes #6024 


rasa/nlu/training_data/training_data.py
```diff
@@ -168,7 +168,10 @@ def examples_per_intent(self) -> Dict[Text, int]:
     @lazy_property
     def examples_per_response(self) -> Dict[Text, int]:
         """Calculates the number of examples per response."""
-        return dict(Counter(self.responses))
```
degiz, line 171: Can we remove the `self.responses` completely now?

rasa/nlu/training_data/training_data.py
```diff
@@ -427,16 +430,40 @@ def split_nlu_examples(
...
+            n_train = int(_count * train_frac)
+            train.extend(_examples[:n_train])
+            test.extend(_examples[n_train:])
+
+        # to make sure we have at least one example per response and intent in the
+        # training/test data, we first go over the response examples and then go over
+        # intent examples
+
+        for response, count in self.examples_per_response.items():
+            examples = [
```
degiz, line 450: optional: we can extract that to `_get_responses_by_name()` or similar.

rasa/nlu/training_data/training_data.py
```diff
@@ -427,16 +430,40 @@ def split_nlu_examples(
...
+            ]
+            _split(examples, count)
+            training_examples = training_examples - set(examples)
+
+        for intent, count in self.examples_per_intent.items():
+            examples = [
+                e
+                for e in training_examples
+                if INTENT in e.data and e.data[INTENT] == intent
+            ]
```
degiz, line 463: optional: we can extract that to _get_intents_by_name() or similar.

rasa/nlu/training_data/training_data.py
```diff
@@ -427,16 +430,40 @@ def split_nlu_examples(
         self, train_frac: float, random_seed: Optional[int] = None
```
degiz, line 430: Since it's a public method, can we have a docstring?

rasa/nlu/training_data/training_data.py
tabergma: No, `self.responses` is still used.

rasa/nlu/training_data/training_data.py
tabergma: as we are not getting the actual responses, but the number of examples per response, I would either keep it or maybe rename to `number_of_examples_per_response`

PR #6168, opened 2020-07-08
preserve key order when dumping YAML

### LINKED ISSUE #6057: RulePolicy: SlotMappings in Domain are requested by alphabet
With the RulePolicy (see `rules` branch) you can add SlotMappings in the domain. When the `FormAction` within Rasa Open Source runs, it should ask for the slots in the same order as it's given in the domain. Currently the slots are requested by alphabet. It's probably due to the `sort` in `Domain.merge`.

ricwo:
**Proposed changes**:
- fixes #6057 


rasa/core/utils.py
```diff
@@ -201,10 +201,19 @@ def _dump_yaml(obj: Dict, output: Union[Text, Path, StringIO]) -> None:
     yaml_writer.dump(obj, output)
 
 
-def dump_obj_as_yaml_to_file(filename: Union[Text, Path], obj: Dict) -> None:
-    """Writes data (python dict) to the filename in yaml repr."""
+def dump_obj_as_yaml_to_file(
+    filename: Union[Text, Path], obj: Dict, should_preserve_key_order: bool = False
```
wochinge, line 205: should we not always preserve the order? 🤔 What's a good reason to not preserve the order?

wochinge, line 205: `obj` could also be of type list, str, float, bool (and probably more) . So maybe just use `Any`?

rasa/utils/io.py
```diff
@@ -199,7 +199,7 @@ def read_config_file(filename: Text) -> Dict[Text, Any]:
         )
 
 
-def read_yaml_file(filename: Text) -> Union[List[Any], Dict[Text, Any]]:
+def read_yaml_file(filename: Union[Text, Path]) -> Union[List[Any], Dict[Text, Any]]:
```
wochinge, line 202: 😻

rasa/core/utils.py
wochinge: performance?

rasa/utils/io.py
```diff
@@ -225,15 +225,69 @@ def unarchive(byte_array: bytes, directory: Text) -> Text:
...
+                    converted_list.append(element)
+            out[k] = converted_list
+        # everything else can just be copied
+        else:
+            out[k] = v
+
+    return out
+
+
+def fix_yaml_representer() -> None:
```
wochinge, line 260: optional: should this be `private`?

wochinge, line 260: optional: `fix` is a bit cryptic to me
```suggestion
def _enable_ordered_dict_yaml_dumping() -> None:
```

rasa/utils/io.py
```diff
@@ -225,15 +225,69 @@ def unarchive(byte_array: bytes, directory: Text) -> Text:
...
      Args:
         data: The data to write.
         filename: The path to the file which should be written.
+        should_preserve_key_order: Whether to preserve key order in `data`.
     """
-    with open(str(filename), "w", encoding=DEFAULT_ENCODING) as outfile:
-        yaml.dump(data, outfile, default_flow_style=False, allow_unicode=True)
+    if should_preserve_key_order:
+        fix_yaml_representer()
+        data = convert_to_ordered_dict(data)
```
wochinge, line 285: note that `data` could be a list and then your code breaks

tests/utils/test_io.py
```diff
@@ -332,3 +334,67 @@ def test_create_directory_for_file(tmp_path: Path):
...
+    "should_preserve_key_order, expected_keys",
+    [
+        (True, list(reversed(string.ascii_lowercase))),
+        (False, list(string.ascii_lowercase)),
+    ],
+)
+def test_dump_yaml_key_order(
+    tmp_path: Path, should_preserve_key_order: bool, expected_keys: List[Text]
+):
+    file = tmp_path / "dir" / "test.yml"
```
wochinge, line 349: why do we need the subdirectory `dir`?

rasa/core/utils.py
ricwo: yeah exactly - I've only set it to `True` in the domain because that's where it's been an issue

liuxun666:
this PR is still not merged into the official branch？ I have the same problem in version 2.0

Issue #6016, opened 2020-06-12
RulePolicy: Have one `validate_form` function instead of multiple ones


wochinge:
**Current State**
As part of the `RulePolicy` we moved the `FormAction` to Rasa Open Source. The form logic is now inside Rasa Open Source. The slot validation can be overwritten by adding custom actions with the syntax `validate_<slot name>` ([see this part of the code](https://github.com/RasaHQ/rasa/blob/da148ae98cbd0c03c8c5a48161622f700e15ec22/rasa/core/actions/forms.py#L390)) . The custom validation function currently receives a temporary tracker which means it's the current conversation + any new `SlotSet` events. 

**Proposed Changes**
- [x] have one function `action_validate_<form_name` to validate everything. `action_validate_<form_name` can also be used to apply custom slot mappings. This means whatever slots they set, they set.
- [x] Change the temporary tracker which we send to this action so that it is clear which slots were extracted previously. This means we send `[*old_tracker_events, ActionExecuted(<form name>), - - [ ] extracted slots)
- [x] Currently we return a `ActionExecutedRejection` event if no new potential slot values were found (see [here](https://github.com/RasaHQ/rasa/blob/da148ae98cbd0c03c8c5a48161622f700e15ec22/rasa/core/actions/forms.py#L436)). As users can now implement their custom slot mappings as part of the `action_validate_<form_name` we need to change the logic. We now should return a `ActionExecutedRejection` when
  * no new slots were extracted **and**
  * `action_validate_<form_name` did neither return a `SlotSet` event nor a bot response

wochinge:
@Ghostvv Just realized that this breaks if the user is adding more slots in the custom action (changing the type of mapping is fine, just an entirely new slot would break things). We can work around that by checking if the custom action returns `SlotSet(REQUESTED_SLOT, <some_slot_name>)`?

Ghostvv:
entire new slot would break rejection logic?

wochinge:
Because we `FormAction.request_next_slot` will check if all slots from the `SlotMapping` are set. In case they are all set, the form will be marked as done. The problem is that it could be the case that all slots from the slotmappings are filled, but the custom action wants to fill some others.

Ghostvv:
so the problems with custom slot mappings? I propose to create a custom separate action that is run every time to get custom slot mappings as a temporary solution

wochinge:
[...]

PR #6171, opened 2020-07-09
have one validate function for the whole form

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6016

### DIFF
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -18,7 +18,6 @@ logger = logging.getLogger(__name__)
 # TODO: Temporary implementation as part of the RulePolicy prototype
 # - add more tests
 # - simplify / refactor
-# - add proper docstrings
 
 
 class FormAction(LoopAction):
@@ -307,7 +306,7 @@ class FormAction(LoopAction):
             domain: The current model domain.
             output_channel: The output channel which can be used to send messages
                 to the user.
-            nlg: The `NaturalLanguageGenerator` to dispatch messages.
+            nlg:  `NaturalLanguageGenerator` to use for response generation.
 
         Returns:
             The validation events including potential bot messages and `SlotSet` events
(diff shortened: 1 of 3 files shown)


wochinge:
@Ghostvv I thought of something like this for custom slot mappings. 

I still have to work on making the extracted slot candidates clearly distinguishable for the custom `validate` function.

Ghostvv:
let's merge [this one](https://github.com/RasaHQ/rasa/pull/6172) into here. Could you add a test that validation is called on the trigger input even if there is different activation action in between

rasa/core/actions/forms.py
```diff
@@ -295,36 +294,56 @@ async def validate_slots(
...
+        _tracker = self._temporary_tracker(tracker, events, domain)
+        _action = RemoteAction(validate_name, self.action_endpoint)
+        validate_events = await _action.run(output_channel, nlg, _tracker, domain)
+
+        validated_slot_names = [
+            event.key for event in validate_events if isinstance(event, SlotSet)
+        ]
+
+        return validate_events + [
+            event for event in events if event.key not in validated_slot_names
```
Ghostvv, line 332: please add comment why we're doing this

rasa/core/actions/forms.py
```diff
@@ -370,22 +389,50 @@ async def request_next_slot(
...
+            (
+                slot
+                for slot in self.required_slots(domain)
+                if self._should_request_slot(tracker, slot)
+            ),
+            None,
+        )
+
     def _name_of_utterance(self, domain: Domain, slot_name: Text) -> Text:
         full_name = f"utter_ask_{self._form_name}_{slot_name}"
```
Ghostvv, line 437: if the name starts with `utter_` it is not a custom action, should we first look for `full_name = f"action_ask_{self._form_name}_{slot_name}"` and only `run` in this case?

Ghostvv:
> I still have to work on making the extracted slot candidates clearly distinguishable for the custom validate function.

this has to be part of the sdk, right?

wochinge:
> I still have to work on making the extracted slot candidates clearly distinguishable for the custom validate function.

Yes 👍 However, I added a `ActionExecuted(form_action)` to the temporary tracker which we send to the SDK so that we can simply take all `SlotSet` events after the `ActionExecuted(form_action)` event

rasa/core/actions/forms.py
wochinge: Let's do this after the alpha and see if users actually want a custom action to ask for the next slot.

Ghostvv: ok, let's remove `.run` there and add `TODO`

wochinge: The run is fine, what speaks against it? It will use `ActionUtterTemplate` and that's it?

Ghostvv: it just a bit counter intuitive (form how usually we call templates in custom actions), but I think it is actually fine

PR #6172, opened 2020-07-09
validate user input in activation

Ghostvv:
**Proposed changes**:
- make sure trigger inout get validated
- clean up unused method

### DIFF
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -190,7 +190,6 @@ class FormAction(LoopAction):
             value = value[0]
         return value
 
-    # noinspection PyUnusedLocal
     def extract_other_slots(
         self, tracker: DialogueStateTracker, domain: Domain
     ) -> Dict[Text, Any]:
@@ -241,7 +240,6 @@ class FormAction(LoopAction):
 
         return slot_values
 
-    # noinspection PyUnusedLocal
     def extract_requested_slot(
         self, tracker: "DialogueStateTracker", domain: Domain
     ) -> Dict[Text, Any]:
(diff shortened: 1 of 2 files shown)


rasa/core/actions/forms.py
```diff
@@ -498,13 +496,16 @@ async def _validate_if_required(
             - the form is called after `action_listen`
             - form validation was not cancelled
         """
-        if tracker.latest_action_name == "action_listen" and tracker.active_loop.get(
-            "validate", True
-        ):
-            logger.debug(f"Validating user input '{tracker.latest_message}'")
+        # no active_loop means that it is called during activation
```
wochinge, line 499: this function is not called during activation anyway

rasa/core/actions/forms.py
```diff
@@ -498,13 +496,16 @@ async def _validate_if_required(
...
+        need_validation = not tracker.active_loop or (
+            tracker.latest_action_name == "action_listen"
```
wochinge, line 501: let's use the constant
```suggestion
            tracker.latest_action_name == ACTION_LISTEN_NAME
```

rasa/core/actions/loops.py
```diff
@@ -43,6 +44,7 @@ async def is_activated(
 
     # default implementation checks if form active
     def _default_activation_events(self) -> List[Event]:
+        # TODO if this is in the loop action, probably it should not be `Form`
```
wochinge, line 47: You mean we should rename the event?

rasa/core/actions/forms.py
Ghostvv: it is called, because inside `LoopAction` we first call `activate`, than `do`. So effectively it is called during activation which is first call

rasa/core/actions/loops.py
Ghostvv: yes, and also think whether we want to use the same event name for forms and loops. So we need to think through how they interact with each other. Let's tackle it separately. I added `TODO` so that we don't forget

PR #6174, opened 2020-07-09
Solve TODO in "entity_is_desired"

tabergma:
**Proposed changes**:
only fill another slot if the slot mapping contains a role or group label to avoid problems like
> in carbon-bot, if you simply input Berlin, it extracts it as city, then it fills both travel_departure and travel_destination with Berlin


rasa/core/actions/forms.py
```diff
@@ -149,18 +149,24 @@ def entity_is_desired(
...
-        return slot_equals_entity  # or slot_fulfils_entity_mapping
+        # whether we can fill a slot with an entity (only if a role or a group label
+        # is set)
+        if (
+            requested_slot_mapping.get("role") is None
+            and requested_slot_mapping.get("group") is None
+        ):
+            slot_fulfils_entity_mapping = False
+        else:
+            matching_values = self.get_entity_value(
```
Ghostvv, line 161: could it be that different slots would have the same entity/role/group mapping: e.g. number of people and number of pizzas? The entity name would be number, and both of them could have groups

tabergma: Don't think so, at least cannot think of any example right now. If you mark both as numbers, you would most likely give them different roles.

Ghostvv: but there is no role for them

Ghostvv: both are just numbers, and if we order pizza for 2 groups, both of them could have different groups

Ghostvv: for example:
"I want 1[number, group: 1] margarita pizza and 2[number, group: 2] vegetarian pizzas"
"we'll be 2[number, group: 1] people at one table and 3[number, group: 2] people at another"

then this can override pizza and vice versa

tabergma: but you could modify this to be 
```
"I want 1[number, group: 1, role: pizza] margarita pizza and 2[number, group: 2, role: pizza] vegetarian pizzas"
"we'll be 2[number, group: 1, role: people] people at one table and 3[number, group: 2, role: people] people at another"
```

But yes, you could come up with such scenarios. But I think that is kind of unlikely. Users actually like the fact that if the bot ask "where do you travel to?" and the user answers "from Berlin to London" that both of them are already picked up and the corresponding slots are filled. This is only possible if we keep this.

Ghostvv: It seems to me that this is only about `role`. Shall we only do it only for roles? and not for groups?

Ghostvv: I think there is a safe check

Ghostvv: if we pass `requested_slot` name here, we can find slot mappings for other slots that differ by roles/groups from slot mapping for `requested_slot`.
However the problem with numbers might still appear, even without roles/group:
"I want 2 pizzas for 2 people"

tabergma: but isn't that in this case just a design decision of the user? e.g. how he labels his data and defines his slot mapping?

I think we could leave out groups here, but I am not sure if users already rely on it.

Ghostvv: but at least we can check that slot_mapping is the same as for requested slot except for role/group

Ghostvv: this way, I think, we make a possibility for mistake smaller

tabergma: Not sure what you mean. 

Before we introduced roles/groups, other slots were just filled if the slot name was equal to the entity name (this is still the case). I extended this check to "also fill a slot if the slot mapping fits". However, this has some problems as you pointed out. If you have two slots that can come from a combination of entity+role and just entity, both of them will be filled if just the entity is present. In order to avoid that, I changed the check to "also fill a slot if the slot mappings depends on a role or a group and fits".

rasa/core/actions/forms.py
```diff
@@ -577,3 +594,22 @@ async def is_done(
...
         logger.debug(f"Deactivating the form '{self.name()}'")
         return []
+
+    def _get_entity_type_of_slot_to_fill(
+        self, slot_to_fill: Text, domain: "Domain"
+    ) -> Optional[Text]:
+        if not slot_to_fill:
+            return None
+
+        mappings = self.get_mappings_for_slot(slot_to_fill, domain)
```
Ghostvv, line 604: I think we should filter out all not `from_entity` mappings here

PR #6177, opened 2020-07-09
Rules feedback

wochinge:
**Proposed changes**:
- translate rules example to yaml
- add deprecation warnings for old rule-like policies


wochinge:
@ricwo I get a lot warnings regarding `unsatisfied checkpoints` when training on rules data in yaml format. Any idea? I used the `examples/rules`

wochinge:
@Ghostvv Do you know these warnings?
Screenshot 2020-07-09 at 18 48 03

rasa/core/training/generator.py
```diff
@@ -727,6 +727,9 @@ def _issue_unused_checkpoint_notification(
         collected_start = set()
         collected_end = set()
         for step in self.story_graph.story_steps:
+            if step.is_rule:
```
wochinge, line 730: fix most of the warnings with this hack 👀

Ghostvv:
the error is the first warning, it thinks that all your stories start with some checkpoint. It should be fixed

rasa/core/training/generator.py
Ghostvv: but `md` format doesn't require this hack? I think there is smth wrong with yaml then

wochinge: no, both require this hack.

wochinge:
I'd simply disable the checkpoint warning for the `RulePolicy`, hm?

examples/rules/data/stories.md
```diff
@@ -1,109 +0,0 @@
-<!-- each story starting with `>>` will be perceived as independent rule -->
```
Ghostvv, line 1: do we still support `md` format? let's not remove that. E.g. for me it is much better format for stories

rasa/core/policies/fallback.py
```diff
@@ -57,6 +58,13 @@ def __init__(
         self.core_threshold = core_threshold
         self.fallback_action_name = fallback_action_name
 
+        common_utils.raise_warning(
+            f"'{self.__class__.__name__}' is deprecated and will be removed in version "
+            "3.0. It is recommended to use the 'RulePolicy' instead.",
```
Ghostvv, line 63: are we really going to drag old style up to version `3`?

wochinge: We can also just say "in the future"?

examples/rules/data/stories.md
wochinge: time for change 😆 I can't put both, yaml and md, in there as then everything would bee duplicate. I'm also not sure if we should introduce new features to the old training data format.

rasa/core/training/generator.py
Ghostvv: wait, isn't this hack disables the warning for all non rule stories? that's not a good idea

examples/rules/data/stories.md
Ghostvv: we should

Ghostvv: I'm planning to use md format

Ghostvv: let's just put in a different folder `stories_md` or smth

rasa/core/policies/fallback.py
Ghostvv: yes, let's do that

Issue #5957, opened 2020-06-04
Add a lookup entity extractor


tabergma:
**Description of Problem**:
Lookup tables are used to create additional features for our entity extractors. In order for the features (aka "the word is part of lookup table xyz, so it is likely that is has the entity xyz") to work, you need to have some examples in your training data with entires from the lookup table.

However, many users think that all entries in the lookup table are extracted as an entity right away. So, as soon as an entry from the lookup table is detected it is extracted as an entity. Many users on the forum struggling with this.

**Overview of the Solution**:
- Create a new component `LookupEntityExtractor` that takes a lookup table. It will extract exact matches of entries in the lookup table and marks them as certain entity. (What about plural/singular/etc. of words? Should we also consider those?)
- We should update the docs to clarify the difference between lookup tables and lookup entity extractor.

erohmensing:
@tabergma this is an interesting idea -- i definitely think users are confused about this. Do you think a `RegexEntityExtractor` would similarly be helpful? I think some users have implemented it themselves already

tabergma:
@erohmensing Good point 👍 `RegexFeaturizer` could cause a similar issue, might make sense to also add an extractor for it.

koaning:
@JiteshGaikwad if case you're interested, we'd be interested in a PR relating to your work [here](https://github.com/JiteshGaikwad/LookupEntityExtractor/).

JiteshGaikwad:
hey @koaning @tabergma @erohmensing  I have submitted the [PR](https://github.com/RasaHQ/rasa/pull/6183)

PR #6183, opened 2020-07-10
Create LookupEntityExtractor.py

JiteshGaikwad:
https://github.com/RasaHQ/rasa/issues/5957#issuecomment-654088366

Lookup Entity Extractor built for Rasa for extracting known entities from a list of examples without the need for training on every set of examples. The extractor is not machine-learned but uses text-matching logic for entity prediction. It doesn't discover any additional values other than the provided list of examples.

Lookup Entity Extractor is a good fit when you have list of exampless that are known and can be expanded as required.

[README.md](https://github.com/JiteshGaikwad/LookupEntityExtractor/blob/master/README.md)

**Proposed changes**:
- ...


sara-tagger:

rasa/nlu/extractors/LookupEntityExtractor.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @staticmethod
+    def _parse_all_entities(user_input: str, entities: list, file_path: list) -> List[Dict[Text, Any]]:
+        """
+        This method does the actual entity extraction work.\n
+        So here I am running the loop over the list of data in the text file\n
+        and check whether it exists in the user's message
+        """
+        results = []
+        for i in range(0, len(entities)):
```
koaning, line 103: You can replace `range(0, len(entities))` with `range(len(entities))`.

rasa/nlu/extractors/LookupEntityExtractor.py
```diff
@@ -0,0 +1,129 @@
...
+from rasa.nlu.extractors.extractor import EntityExtractor
+from rasa.nlu.model import Metadata
+from rasa.nlu.training_data import Message
+
+logger = logging.getLogger(__name__)
+
+
+class LookupEntityExtractor(EntityExtractor):
+    """
+    Searches for entities in the user's message from a list of examples.\n
```
koaning, line 20: I've noticed the `\n` character appear in the docstrings. Is there a reason for this?

rasa/nlu/extractors/LookupEntityExtractor.py
```diff
@@ -0,0 +1,129 @@
...
+        So here I am running the loop over the list of data in the text file\n
+        and check whether it exists in the user's message
+        """
+        results = []
+        for i in range(0, len(entities)):
+            f = open(file_path[i], "r")
+            examples = f.readlines()
+
+            for example in examples:
+                if example.lower().strip() in user_input.lower():
```
koaning, line 108: You seem to lower case everything here. I can certainly see that there is merit to this but I wonder if this is the behavior you'd want in all scenarios. It might be better to have this be a configuration. @tabergma you might also have an opinion on this.

rasa/nlu/extractors/LookupEntityExtractor.py
```diff
@@ -0,0 +1,129 @@
...
+                    start_index = user_input.lower().index(example.lower().strip())
+                    end_index = start_index + len(example.strip())
+                    temp = {}
```
koaning, line 111: This is personal preference, but it might be appropriate to declare the dictionary directly. 

```python
temp = {
  "entity": entities[i], 
  "start": start_index,
  "end": end_index,
  "value": user_input[start_index:end_index]
}
```

rasa/nlu/extractors/LookupEntityExtractor.py
```diff
@@ -0,0 +1,129 @@
...
+    @staticmethod
+    def _parse_all_entities(user_input: str, entities: list, file_path: list) -> List[Dict[Text, Any]]:
+        """
+        This method does the actual entity extraction work.\n
+        So here I am running the loop over the list of data in the text file\n
+        and check whether it exists in the user's message
+        """
+        results = []
+        for i in range(0, len(entities)):
+            f = open(file_path[i], "r")
```
koaning, line 104: It might be preferable to explicitly use a context manager here. Something like;

```python
with open(file_path[i], "r") as f:
    ...
```

Another alternative might be to use pathlib. 

```python
import pathlib
pathlib.Path(file_path[i]).read_text()
```

For details on why there's a link to [stackoverflow](https://stackoverflow.com/questions/7395542/is-explicitly-closing-files-important).

koaning:
Hi @JiteshGaikwad thanks for the PR! I've given it a first order review, mostly some style things for now. Feel free to discuss them if you're unsure if my suggestion is valid. 

One thing that is currently missing is a set of unit tests though. Have you thought about how you'd want to unit test this component? Feel free to ask for help if this is unfamiliar territory.

rasa/nlu/extractors/LookupEntityExtractor.py
JiteshGaikwad: If in the Lookup file, if there is any value which is lower case or user query has both the things, so to avoid the casing issue, I preferred lower case everything here. let me know if this behavior increases the code processing time.

rasa/nlu/extractors/LookupEntityExtractor.py
JiteshGaikwad: I have added it so that whenever I see the class definition it would look properly formatted

image

Let me know if I need to remove the formatting with "\n"

rasa/nlu/extractors/LookupEntityExtractor.py
JiteshGaikwad: Updated the code to declare the dictionary directly.

rasa/nlu/extractors/LookupEntityExtractor.py
JiteshGaikwad: @koaning how about `f.close()` won't it close the file connection?

JiteshGaikwad:
> Hi @JiteshGaikwad thanks for the PR! I've given it a first-order review, mostly some style things for now. Feel free to discuss them if you're unsure if my suggestion is valid.
> 
> One thing that is currently missing is a set of unit tests though. Have you thought about how you'd want to unit test this component? Feel free to ask for help if this is unfamiliar territory.

@koaning 
I haven't used the unit testing yet for my previous work 😅 ya which is unfamiliar territory for me.

Thanks

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,129 @@
...
+                    pass
+                else:
+                    self.component_config["files_path"] = None
+                    print("File path", file_path)
+                    common_utils.raise_warning(
+                        f"Can't extract Lookup Entities, File \"{file_path}\" does not exist."
+                    )
+                    break
+
+    def add_extractor_name(
```
tabergma, line 63: You don't need this method here. It is already defined in the parent class.

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,129 @@
...
+        for entity in entities:
+            entity["extractor"] = self.name
+        return entities
+
+    @staticmethod
+    def _parse_entities(self, text: Text) -> List[Dict[Text, Any]]:
+        """
+        pass the user input to  extract the entities
+
+        """
```
tabergma, line 78:
```suggestion
        """Extract entities from the user input."""
```

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Searches for entities in the user's message from a list of examples.\n
+    Required Parameters: \n
+    @entities -> list \n
+    @files_path -> list
+    """
+
+    defaults = {
+        "entities": None,
+        "files_path": None
```
tabergma, line 28: Can you please add comments what those values are for? Thanks.

tabergma: If I understand it correctly `entities` and `files_path` are both lists. So the first file path lists all values of the first entity type listed in `entities`. Maybe it would be simple to store this as a dict. For example
```
- name: LookupEntityExtractor
   lookup:
     city: /some/path/city.txt
     person: /some/other/path/person.txt
```
What do you think?

tabergma: If you store them as dict, it would also simplify the checks in `__init__` quite a bit.

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,129 @@
...
+                    temp = {
+                        "entity": entities[i], 
+                        "start": start_index,
+                        "end": end_index,
+                        "value": user_input[start_index:end_index]
+                    }
+                    results.append(temp)
+        return results
+
+    def process(self, message, **kwargs):
```
tabergma, line 119:
```suggestion
    def process(self, message: Message, **kwargs: Any) -> None:
```

rasa/nlu/extractors/lookup_entity_extractor.py
JiteshGaikwad: @tabergma thanks for the review, I will make those changes

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+                # if the lookup dictionary is empty after filtering
+                # the lookup entities, assign it None value
+                self.component_config["lookup"] = None
+        else:
+            common_utils.raise_warning(
+                "Can't extract Lookup Entities,\
+                Please configure the lookup entities in the config.yml."
+            )
+
+    def add_entities_if_lookup_entity(self, entity, file_path) -> None:
```
tabergma, line 53:
```suggestion
    def _validate_lookup_entry(self, entity: Text, file_path: Text) -> None:
```

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+        if file_path is not None:
+            if os.path.isfile(file_path):
+                pass
+            else:
+                # remove the entity from the lookup dictionary,
+                # if the file path doesn't exist
+                self.component_config["lookup"].pop(entity)
+                common_utils.raise_warning(
+                    f"can't extract lookup entity: '{entity}',\
+                    make sure the provided file '{file_path}' exists.")
```
tabergma, line 63:
```suggestion
            if not os.path.isfile(file_path):
                # remove the entity from the lookup dictionary,
                # if the file path doesn't exist
                self.component_config["lookup"].pop(entity)
                common_utils.raise_warning(
                    f"The file path '{file_path}' for entity '{entity}' does not exist. Please provide a valid file path.")
```

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            # remove the entity from the lookup dictionary,
+            # if the file path is not provided
+            self.component_config["lookup"].pop(entity)
+            common_utils.raise_warning(
+                f"can't extract '{entity}' entity,\
+                please provide the example file.")
```
tabergma, line 70:
```suggestion
                common_utils.raise_warning(
                    f"No file path for entity '{entity}' was given. Please provide a valid file path.")
```

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+
+            if not bool(component_config["lookup"]):
+                # if the lookup dictionary is empty after filtering
+                # the lookup entities, assign it None value
+                self.component_config["lookup"] = None
+        else:
+            common_utils.raise_warning(
+                "Can't extract Lookup Entities,\
+                Please configure the lookup entities in the config.yml."
+            )
```
tabergma, line 51: I would actually raise a `ValueError` here. If no lookup dictionary is configured, the component cannot be used at all. This should result in a failure.

tabergma: Same for the other cases above. E.g. if all entries in the lookup dictionary are invalid, so that nothing is left, raise a `ValueError`.

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+            # remove the entity from the lookup dictionary,
+            # if the file path is not provided
+            self.component_config["lookup"].pop(entity)
+            common_utils.raise_warning(
+                f"can't extract '{entity}' entity,\
+                please provide the example file.")
+
+    def _parse_entities(self, user_input: Text) -> List[Dict[Text, Any]]:
+        """Extract entities from the user input."""
+        if self.component_config["lookup"] is not None:
```
tabergma, line 74: If we raise a `ValueError` in case the lookup dict is not set, we can also remove this check as we know that it exists and that it is valid.

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+            for entity, file_path in list(
+                    self.component_config["lookup"].items()):
```
tabergma, line 76:
```suggestion
            for entity, file_path in self.component_config["lookup"].items():
```
I don't think you need to wrap this in a list.

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+                results = self._parse_all_entities(
+                    user_input, entity, file_path)
+                return results
+        else:
+            return []
+
+    @staticmethod
+    def _parse_all_entities(
+        user_input: str, entity: list, file_path: list
+    ) -> List[Dict[Text, Any]]:
```
tabergma, line 86:
```suggestion
    def _extract_entities(user_input: Text, entity: Text, file_path: Text) -> List[Dict[Text, Any]]:
```

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+        This method does the actual entity extraction work.
+        So here I am running the loop over the list of data in the text file
+        and check whether it exists in the user's message
+        """
+        results = []
+        with open(file_path, "r") as f:
+            examples = f.readlines()
+
+            for example in examples:
+                example = example.lower().strip()
```
tabergma, line 97: Let's add an option `lowercase`. We can set its default value to `True`. But I think it might happen that users need to consider the casing when using this component.

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        This method does the actual entity extraction work.
+        So here I am running the loop over the list of data in the text file
+        and check whether it exists in the user's message
+        """
+        results = []
+        with open(file_path, "r") as f:
+            examples = f.readlines()
+
+            for example in examples:
```
tabergma, line 96: Let's read the file line by line to avoid memory issues.
```
        with open(file_path, "r") as file:
            for example in file:
```

rasa/nlu/extractors/LookupEntityExtractor.py
koaning: Every editor is bound to render it differently, but it's probably best to stick to the PEP guidelines here which is to remove the `\n`.

rasa/nlu/extractors/LookupEntityExtractor.py
koaning: I'll resolve this conversation since tanja replied with a more indepth correspondence below.

koaning:
@tabergma as far as unit tests go, I'm wondering what items we'd really like to test. From the top of my head I'll list a few here so we can discuss then. 

1. Make sure we properly add the entities.
2. Make sure the lowercase setting works as expected.
3. Make sure we don't remove previous entities? 

I'm also thinking about documentation. Do we want to add an entry to the .rst documents right away? I can help out with this if it is appreciated.

tabergma:
@koaning The tests you suggested sound good. 
> Make sure we don't remove previous entities?

This is a good point. We should definitely not overwrite existing entities.

Regarding documentation: We should not merge without at least an entry in the `component.rst` page. There are some other pages, which we could update, but this page is the most important one to update, I would say.

JiteshGaikwad:
@tabergma I have added the suggested changes, also I have added the checks for the casing.
@koaning need help in understanding point 1 and 3 below:
> 1. Make sure we properly add the entities.
> 2. Make sure the lowercase setting works as expected.
> 3. Make sure we don't remove previous entities?

Thanks.

tabergma:
Thanks again @JiteshGaikwad. We added test and documentation. Here is the link to the final PR: https://github.com/RasaHQ/rasa/pull/6214/files

JiteshGaikwad:
Thanks @tabergma  and @koaning

PR #6191, opened 2020-07-13
Update 'convert_predictions_into_entities'

tabergma:
**Proposed changes**:
If two entities are separated by a comma, they are currently detected as one entity. Example:
`please add buffalo, ranch, mustard, barbeque sauces` results in `{'entity': 'sauces', 'start': 11, 'end': 43, 'value': 'buffalo, ranch, mustard, barbeque', 'extractor': 'DIETClassifier'}`

This PR fixes this. Entities separated by a comma will be listed as individual entities.

Related to [forum post](https://forum.rasa.com/t/unable-to-classify-multiple-examples-of-the-same-entity-please-help/28851/7)


rasa/nlu/extractors/extractor.py
```diff
@@ -164,13 +166,28 @@ def convert_predictions_into_entities(
                 )
                 entities.append(entity)
             else:
-                entities[-1][ENTITY_ATTRIBUTE_END] = token.end
-                if confidences is not None:
-                    self._update_confidence_values(entities, confidences, idx)
+                if last_token_end == token.start - 1:
```
ricwo, line 169: could you add a comment on this condition?

rasa/nlu/extractors/extractor.py
```diff
@@ -164,13 +166,28 @@ def convert_predictions_into_entities(
...
+                    entities[-1][ENTITY_ATTRIBUTE_END] = token.end
+                    if confidences is not None:
+                        self._update_confidence_values(entities, confidences, idx)
+                else:
```
ricwo, line 173: can this `if/else` be indented to the left, so that we have
```python
if new_tag_found:
    ...
elif last_token_end == token.start - 1:
    ...
else:
    ...
```

rasa/nlu/extractors/extractor.py
ricwo, line 169: could you add a comment on this condition?

rasa/nlu/extractors/extractor.py
```diff
@@ -164,13 +166,28 @@ def convert_predictions_into_entities(
...
+                    # create a new entity as the two entities are separated by a comma
+                    # or some other symbol
```
ricwo, line 175: what about spaces `" "`?

tabergma: If you have multiple spaces between an two tokens of the same entity type, we treat them as separate entities. Do you think this could be a problem?

ricwo: no that seems fine, I just wasn't sure what would happen. but your comment in the code is very clear now!

PR #6192, opened 2020-07-13
Update the tag patter in the docs

degiz:
**Proposed changes**:

- The docs can now point to the `2.0.0a1` `2.0.0rc1` version

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -353,7 +353,7 @@ scv_whitelist_branches = (re.compile("^master$"),)
 scv_grm_exclude = ("README.md", ".gitignore", ".nojekyll", "CNAME")
 scv_whitelist_tags = (
     re.compile(r"^[2-9]+\.\d+\.\d+$"),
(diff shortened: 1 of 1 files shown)


tmbo:
mhm I think we usually use the `2.0.0a1` and `2.0.0rc1` format (as that is the python default)

degiz:
Shouldn't it be the semver style?

tmbo:
yes imo it should be semver, but the python community decided it would be a great idea to go a different route when handling pre-release versions...

docs/conf.py
```diff
@@ -353,6 +353,7 @@
 scv_grm_exclude = ("README.md", ".gitignore", ".nojekyll", "CNAME")
 scv_whitelist_tags = (
     re.compile(r"^[2-9]+\.\d+\.\d+$"),
+    re.compile(r"^2\.[0-9]+\.[0-9]?(\w+)?$"),
```
tmbo, line 356:
```suggestion
    re.compile(r"^2\.[0-9]+\.[0-9]+(\w+)?$"),
```

PR #6204, opened 2020-07-14
Changes to NLU to correctly process E2E data

evgeniiaraz:
**Proposed changes**:
- support for MD for E2E
- changed the name from `MESSAGE_ACTION_NAME` to `ACTION_NAME`
- made `additional_e2e_action_names` into a `set()`
- adding `attribute` to `process ` functions of NLU modules
- adding a `parse_message` function to `nlu.model.Interpreter`
- getting rid of `text` attribute of `Message`; instead text will be stored in `.data`


rasa/importers/importer.py
```diff
@@ -328,17 +328,20 @@ async def get_stories(
     async def get_config(self) -> Dict:
         return await self._importer.get_config()
 
-    async def get_nlu_data(self, language: Optional[Text] = "en", only_nlu: bool = False) -> TrainingData:
+    async def get_nlu_data(
+        self, language: Optional[Text] = "en", only_nlu: bool = False
+    ) -> TrainingData:
```
Ghostvv, line 333: let's do this cosmetic changes in a previous PR

rasa/importers/importer.py
```diff
@@ -376,7 +379,13 @@ def _messages_from_user_utterance(event: UserUttered) -> Message:
 
 def _messages_from_action(event: ActionExecuted) -> Message:
     # we need to store the action text twice to be able to differentiate between user and bot text in NLU processing
-    return Message(event.e2e_text or "", data={MESSAGE_ACTION_NAME: event.action_name, ACTION_TEXT: event.e2e_text or ""})
+    return Message(
```
Ghostvv, line 382: cosmetic changes belong to previous PR

rasa/nlu/constants.py
```diff
@@ -39,19 +39,26 @@
...
 
-MESSAGE_ATTRIBUTES = [TEXT, INTENT, RESPONSE]
-DENSE_FEATURIZABLE_ATTRIBUTES = [TEXT, RESPONSE]
+MESSAGE_ATTRIBUTES = [TEXT, INTENT, RESPONSE, MESSAGE_ACTION_NAME, ACTION_TEXT]
+DENSE_FEATURIZABLE_ATTRIBUTES = [TEXT, RESPONSE, ACTION_TEXT]
 
 LANGUAGE_MODEL_DOCS = {
     TEXT: "text_language_model_doc",
     RESPONSE: "response_language_model_doc",
+    ACTION_TEXT: "action_text_model_doc",
```
Ghostvv, line 48: `BOT_TEXT` ?

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -224,8 +226,10 @@ def _get_message_tokens_by_attribute(
         """Get text tokens of an attribute of a message"""
         if message.get(TOKENS_NAMES[attribute]):
             return [t.lemma for t in message.get(TOKENS_NAMES[attribute])]
-
-        return message.get(attribute).split()
+        if not attribute == MESSAGE_ACTION_NAME or attribute == INTENT:
```
Ghostvv, line 229: why this check?

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -434,7 +438,7 @@ def _create_features(
 
             sequence_features.append(seq_vec.tocoo())
 
-            if attribute in [TEXT, RESPONSE]:
+            if attribute in [TEXT, RESPONSE, ACTION_TEXT]:
```
Ghostvv, line 441: I thought we wanted to use response key for action text

wochinge:
I'm holding off with this one till we merged the other one. Just re-request review then please.

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -166,20 +168,24 @@ def _create_sparse_features(self, message: Message) -> None:
         features."""
         import scipy.sparse
 
-        tokens = message.get(TOKENS_NAMES[TEXT])
+        # TODO: do I get it right that these features will still exist only for user text?
+        tokens = None
+        if not message.get(MESSAGE_ACTION_NAME):
```
Ghostvv, line 173: yes, but why do we need to check for action_text, doesn't it go to separate message attribute?

rasa/nlu/model.py
```diff
@@ -162,6 +165,15 @@ def _build_pipeline(
 
         return pipeline
 
+    def filter_data(self, training_data: TrainingData) -> TrainingData:
```
Ghostvv, line 168: what are we filtering here?

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -87,17 +94,37 @@ def train(
                 ):
                     if attribute == INTENT:
                         tokens = self._split_intent(example)
+                    elif attribute == MESSAGE_ACTION_NAME:
+                        # when we have text action, the text will be stored both in `action_name` and in `text`;
```
Ghostvv, line 98: can we change that and set `action_name` to `None`? similar with user intents?

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -87,17 +94,37 @@ def train(
...
+                        # we need to check if the text is empty to make sure that it is actually the `action_name`
+                        if not example.get(ACTION_TEXT):
+                            tokens = self._split_action(example)
```
Ghostvv, line 101: what is the difference between split action?

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -108,6 +135,17 @@ def _split_intent(self, message: Message):
 
         return self._convert_words_to_tokens(words, text)
 
+    def _split_action(self, message: Message) -> List[Token]:
```
Ghostvv, line 138: let's have two generic `_split_text` and `_split_label`

rasa/utils/tensorflow/crf.py
```diff
@@ -13,8 +13,10 @@
 class CrfDecodeForwardRnnCell(tf.keras.layers.AbstractRNNCell):
     """Computes the forward decoding in a linear-chain CRF."""
 
+    # TODO: the way transition params are initialised in Windows doesn't pass the type check for TensorLike.
+    # I'll look into it later.
     @typechecked
-    def __init__(self, transition_params: TensorLike, **kwargs) -> None:
+    def __init__(self, transition_params, **kwargs) -> None:
```
Ghostvv, line 19: @tabergma could you please take a look at that?

rasa/nlu/model.py
evgeniiaraz: we're filtering out bot data

evgeniiaraz:
@Ghostvv what do you think about yesterdays changes with `NAME_TOKENS`?

rasa/nlu/constants.py
```diff
@@ -39,19 +39,28 @@
...
+    RESPONSE: "response_spacy_doc",
+    ACTION_TEXT: "action_text_spacy_doc",
 }
-SPACY_DOCS = {TEXT: "text_spacy_doc", RESPONSE: "response_spacy_doc"}
 
 TOKENS_NAMES = {
     TEXT: "text_tokens",
     INTENT: "intent_tokens",
     RESPONSE: "response_tokens",
+    MESSAGE_ACTION_NAME: "action_name_tokens",
```
Ghostvv, line 62:
```suggestion
    ACTION_NAME: "action_name_tokens",
```

rasa/nlu/constants.py
```diff
@@ -39,19 +39,28 @@
 
 NUMBER_OF_SUB_TOKENS = "number_of_sub_tokens"
 
-MESSAGE_ATTRIBUTES = [TEXT, INTENT, RESPONSE]
-DENSE_FEATURIZABLE_ATTRIBUTES = [TEXT, RESPONSE]
+MESSAGE_ATTRIBUTES = [TEXT, INTENT, RESPONSE, MESSAGE_ACTION_NAME, ACTION_TEXT]
```
Ghostvv, line 42:
```suggestion
MESSAGE_ATTRIBUTES = [TEXT, INTENT, RESPONSE, ACTION_NAME, ACTION_TEXT]
```

evgeniiaraz: Done with changes to `MESSAGE_ACTION_NAME`! I just changed the `message_action_name` everywhere in one go

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -119,7 +119,8 @@ def _create_feature_to_idx_dict(
         all_features = []
         for example in training_data.training_examples:
             tokens = example.get(TOKENS_NAMES[TEXT])
-            all_features.append(self._tokens_to_features(tokens))
+            if tokens:
```
Ghostvv, line 122: why do we need this? we didn't have it before

rasa/nlu/constants.py
```diff
@@ -39,19 +39,28 @@
 
 NUMBER_OF_SUB_TOKENS = "number_of_sub_tokens"
 
-MESSAGE_ATTRIBUTES = [TEXT, INTENT, RESPONSE]
-DENSE_FEATURIZABLE_ATTRIBUTES = [TEXT, RESPONSE]
+MESSAGE_ATTRIBUTES = [TEXT, INTENT, RESPONSE, ACTION_NAME, ACTION_TEXT]
+# the dense featurizable attributes are essentially text attributes
+DENSE_FEATURIZABLE_ATTRIBUTES = [TEXT, RESPONSE, ACTION_TEXT]
+NAME_ATTRIBUTES = [INTENT, ACTION_NAME]
```
Ghostvv, line 45: this system seem to be working good, the name of the constant should be different maybe

rasa/importers/importer.py
```diff
@@ -7,7 +7,7 @@
 from rasa.core.events import ActionExecuted, UserUttered, Event
 from rasa.core.interpreter import RegexInterpreter, NaturalLanguageInterpreter
 from rasa.core.training.structures import StoryGraph
-from rasa.nlu.constants import MESSAGE_ACTION_NAME, MESSAGE_INTENT_NAME, ACTION_TEXT
+from rasa.nlu.constants import ACTION_NAME, MESSAGE_INTENT_NAME, ACTION_TEXT
```
Ghostvv, line 10: let's also rename intent constant as well
```suggestion
from rasa.nlu.constants import ACTION_NAME, INTENT_NAME, ACTION_TEXT
```

rasa/utils/tensorflow/crf.py
tabergma: Do we get any error message for this? Are all requirements up-to-date? 
`TensorLike` covers the following types
```
TensorLike = Union[
    List[Union[Number, list]],
    tuple,
    Number,
    np.ndarray,
    tf.Tensor,
    tf.SparseTensor,
    tf.Variable,
]
```
In `crf.py` I guess all parameters defined as `TensorLike` could also be directly set to `tf.Tensor` if that would help.

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
evgeniiaraz: because now we can have empty text at the featurization level (e.g., `Message("", {action_name: "action_listen"})`)

rasa/utils/tensorflow/crf.py
evgeniiaraz: I think my laptop became friendlier to me after the holidays and everything started working... 🤷‍♀️

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
evgeniiaraz: added a comment about it

evgeniiaraz:
I also did a little performance test on Sara (the way it is in the model regression tests) - and the results are the same (up to a random seed the same).

data/test_yaml_stories/stories_hybrid_e2e.yml
```diff
@@ -1,3 +1,4 @@
+
```
Ghostvv, line 1:
```suggestion
```

rasa/core/training/story_reader/markdown_story_reader.py
```diff
@@ -29,11 +29,14 @@ class MarkdownStoryReader(StoryReader):
...
-    async def read_from_file(self, filename: Text,) -> List[StoryStep]:
+    async def read_from_file(self, filename: Text) -> List[StoryStep]:
         """Given a md file reads the contained stories."""
 
         try:
-            with open(filename, "r", encoding=io_utils.DEFAULT_ENCODING) as f:
+            # TODO: Hack by Genie for Windows
+            with open(
+                filename, "r", encoding=io_utils.DEFAULT_ENCODING, errors="ignore"
+            ) as f:
```
Ghostvv, line 39: is it a general solution? current master doesn't work on Windows?

rasa/core/training/story_reader/story_reader.py
```diff
@@ -42,12 +42,24 @@ def _new_story_part(self, name: Text, source_name: Text):
...
-    def _add_event(self, event_name: Text, parameters: Dict[Text, Any]) -> None:
+    def _add_event(
+        self, event_name: Text, parameters: Dict[Text, Any], is_e2e: bool = False
+    ) -> None:
 
         # add 'name' only if event is not a SlotSet,
         # because there might be a slot with slot_key='name'
         if "name" not in parameters and event_name != SlotSet.type_name:
-            parameters["name"] = event_name
+            # TODO: Hack by Genie as part of the prototype
```
Ghostvv, line 52: is it also related only to markdown support?

rasa/importers/importer.py
```diff
@@ -368,19 +374,21 @@ def _message_from_conversation_event(event: Event) -> Optional[Message]:
...
 def _messages_from_user_utterance(event: UserUttered) -> Message:
-    return Message(event.text, data={MESSAGE_INTENT_NAME: event.intent_name})
+    return Message(event.text, data={INTENT_NAME: event.intent_name})
 
 
 def _messages_from_action(event: ActionExecuted) -> Message:
-    return Message(event.e2e_text or "", data={MESSAGE_ACTION_NAME: event.action_name})
+    return Message(
+        "", data={ACTION_NAME: event.action_name, ACTION_TEXT: event.e2e_text or "",},
+    )
```
Ghostvv, line 383: there is `Message.build` method that creates message with text, intents and entities, let's create a similar `build_for_actions` or smth that would create message from action_name and action_text

Ghostvv: also I think `Message.text` doesn't make sense anymore

tests/nlu/tokenizers/test_whitespace_tokenizer.py
```diff
@@ -152,3 +163,47 @@ def test_whitespace_training(supervised_embeddings_config):
...
+    tk.process(message, INTENT)
+    tokens_intent = message.get(TOKENS_NAMES[INTENT])
+    tk.process(message, TEXT)
+    tokens_text = message.get(TOKENS_NAMES[TEXT])
+    assert [t.text for t in tokens_intent] == expected_tokens_intent
+    assert [t.text for t in tokens_text] == expected_tokens_text
+
+    message = Message(
+        "Where are you going?",
+        {ACTION_NAME: "Where are you going?", ACTION_TEXT: "Where are you going?",},
```
Ghostvv, line 199:
```suggestion
        {ACTION_NAME: None, ACTION_TEXT: "Where are you going?"},
```

tests/nlu/tokenizers/test_whitespace_tokenizer.py
```diff
@@ -137,9 +143,14 @@ def test_whitespace_training(supervised_embeddings_config):
                 ],
             },
         ),
+        Message("action_restart", {"action_name": "action_restart"},),
+        Message(
+            "Where are you going?",
+            {ACTION_NAME: "Where are you going?", ACTION_TEXT: "Where are you going?",},
```
Ghostvv, line 149:
```suggestion
            {ACTION_NAME: None, ACTION_TEXT: "Where are you going?"},
```

Ghostvv:
looks good, left a couple of comments. I think it makes sense to refactor slightly `Message` object, get rid of `self.text` (use `self.data[TEXT]` instead) and create build helper for actions

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -81,21 +87,28 @@ def train(
...
             for attribute in MESSAGE_ATTRIBUTES:
-                if example.get(attribute):
-                    if attribute == INTENT:
-                        tokens = self._split_intent(example)
+                if (
+                    example.get(attribute) is not None
+                    and not example.get(attribute) == ""
+                ):
+                    if attribute in [INTENT, ACTION_NAME]:
+                        tokens = self._split_name(example, attribute)
```
tabergma, line 95: Minor: Rename to `_split_label`

rasa/core/training/story_reader/markdown_story_reader.py
evgeniiaraz: Tbh, I assume it is my laptop being stupid about things... I'm sure if it didn;t work people would complain a mln times!

evgeniiaraz:
@Ghostvv just to make sure -- are you saying `Message` shouldn't have `text` property at all? 

@tabergma while I'll be in the `Message`, I was trying to see previously how it is best to return `features` in `as_dict()`. Everything I came up with was unbearably ugly :( e.g., `intent_sparse_sequence, intent_dense_sequence` etc. 
Do you have any thoughts on how to make it beautiful?

rasa/nlu/tokenizers/tokenizer.py
evgeniiaraz: I was trying my best to avoid the word `label` because label is also the Y in DIET and TED...

tabergma:
@evgeniiaraz You should implement the method `as_dict` in the class `Features`. And then you should make sure to call this method for all features when `message.as_dict` is called.

evgeniiaraz:
@tabergma oh, are you saying that for every feature it should return a dictionary which looks like {`attribute`: `TEXT`, `is_sparse`: True, `sequence`: True, `numbers`: [0.0.0]}? rather than a flat a dictionary it will be a list of dictionaries for all features, right?

Ghostvv:
> are you saying Message shouldn't have text property at all

yes, since there could be 2 different texts

tests/nlu/tokenizers/test_whitespace_tokenizer.py
evgeniiaraz: I would like to leave the examples as is -- in reality we will not have the text loaded into action name but I wanted to test that both attributes will be processed correctly.

rasa/core/training/story_reader/story_reader.py
evgeniiaraz: No, I think this will work for both Markdown and Yaml. what it does in reality is clean up of the action text. (i'm not sure this is that much of a hack...)

rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -343,7 +342,7 @@ def _parse_bot_message(self, step: Dict[Text, Any]) -> None:
             )
             return
 
-        self._add_event(bot_message, {"e2e_text": bot_message})
+        self._add_event(bot_message, {"e2e_text": bot_message}, is_e2e = True)
```
evgeniiaraz, line 345: @wochinge Correct me if I'm wrong... I think the `is_e2e` tag was supposed to be here. 
There was a test and it worked because `ActionExecuted.__eq__()` only compared action names and not whether there was `e2e_text` and,  whether it was the same. (updated that as well)

wochinge: The `is_e2e` should no longer be needed. We only had that for the hacky markdown format

rasa/core/training/story_reader/story_reader.py
wochinge: It's indeed only related to markdown support and should be removed

rasa/core/training/story_reader/yaml_story_reader.py
evgeniiaraz: How will it know to set `e2e_text` in `ActionExecuted` if we don't pass `is_e2e` to it?

wochinge: yaml handles this here https://github.com/RasaHQ/rasa/blob/dfecdc7a6bff0bed20c54cd650860350667d6d80/rasa/core/training/story_reader/yaml_story_reader.py#L333

evgeniiaraz: ah! my bad! 
then I'll change `markdown` to work the same way and change it here so that the `action_name` is empty
and remove the `is_e2e` fluff we had before :)

wochinge: why do we keep the markdown stuff in the first place? Can't we just omit it's implementation?

evgeniiaraz: I mean we already have all of it done, so, why not?

wochinge: because 
1) it's not tested
2) we didn't consult anybody else how the format should look
3) Markdown is very likely gonna be dropped

wochinge: If you do a revert commit we can still re-add it later in case we need it, but for now I'd simply drop it.

evgeniiaraz: With respect to tests, I added one similar to the yaml one. 
I totally agree with 2 and 3!
The problem for me is that it means I (just like our users) have to manually rewrite all my stories to yaml to test anything I do with e2e now... a little bit of a pain 😅

wochinge: > The problem for me is that it means I (just like the rasa users) have to manually rewrite all my stories to yaml to test anything I do with e2e now... a little bit of a pain 😅

I feel you, but I am afraid there is no way around it as we have to do it anyway sooner or later

rasa/core/events/__init__.py
```diff
@@ -1010,11 +1011,27 @@ def __str__(self) -> Text:
...
+            if hasattr(self, "e2e_text") and hasattr(other, "e2e_text"):
+                return (
+                    self.action_name == other.action_name
+                    and self.e2e_text == other.e2e_text
+                )
+            # if both of them don't have the attribute, only compare action name
+            if self._check_for_e2e_text() == other._check_for_e2e_text():
+                return self.action_name == other.action_name
+            else:
+                return False
```
wochinge, line 1034: why do we need this?

rasa/core/training/story_reader/markdown_story_reader.py
```diff
@@ -172,6 +181,13 @@ def _parse_event_line(line):
             )
             return "", {}
 
+    @staticmethod
+    def _parse_bot_message_e2e(line):
```
wochinge, line 185: types

rasa/importers/importer.py
```diff
@@ -301,13 +302,17 @@ async def _get_domain_with_e2e_actions(self) -> Domain:
...
-                event.action_name
-                for event in story_step.events
-                if isinstance(event, ActionExecuted) and event.e2e_text
-            ]
+            additional_e2e_action_names.update(
+                [
+                    event.e2e_text
+                    for event in story_step.events
+                    if isinstance(event, ActionExecuted) and event.e2e_text
+                ]
```
wochinge, line 312:
```suggestion
                {
                    event.e2e_text
                    for event in story_step.events
                    if isinstance(event, ActionExecuted) and event.e2e_text
                }
```

rasa/nlu/classifiers/keyword_intent_classifier.py
```diff
@@ -48,20 +48,20 @@ def train(
         duplicate_examples = set()
         for ex in training_data.training_examples:
             if (
-                ex.text in self.intent_keyword_map.keys()
-                and ex.get(INTENT) != self.intent_keyword_map[ex.text]
+                ex.get(TEXT) in self.intent_keyword_map.keys()
```
wochinge, line 51: I think it would be better in the future to such a changes in a separate PR to keep the surface of the PR smaller and more concise

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -167,19 +168,22 @@ def _create_sparse_features(self, message: Message) -> None:
         import scipy.sparse
 
         tokens = message.get(TOKENS_NAMES[TEXT])
+        # this check is required because there might be training data examples without TEXT,
+        # e.g., `Message("", {action_name: "action_listen"})`
+        if tokens:
```
wochinge, line 173: optional: use a guard clause and do something like:

```
if not tokens:
    return

... all your code

File: rasa/nlu/training_data/message.py [hunk 30]
```diff
@@ -84,16 +82,16 @@ def as_dict(self, only_output_properties=False) -> dict:
         # Message object in markdown format
         d = {key: value for key, value in d.items() if value is not None}
```
wochinge, line 83:
```suggestion
        return {key: value for key, value in d.items() if value is not None}
```

rasa/nlu/training_data/message.py
```diff
@@ -84,16 +82,16 @@ def as_dict(self, only_output_properties=False) -> dict:
...
 
-        return dict(d, text=self.text)
+        return dict(d)
```
wochinge, line 85:
```suggestion
```

rasa/nlu/training_data/message.py
```diff
@@ -114,6 +112,16 @@ def build(
             data[ENTITIES] = entities
         return cls(text, data, **kwargs)
 
+    @classmethod
+    def build_from_action(
+        cls,
+        action_text: Optional[Text] = "",
+        action_name: Optional[Text] = "",
+        **kwargs,
```
wochinge, line 120: types please

rasa/nlu/training_data/message.py
```diff
@@ -24,16 +26,16 @@
 class Message:
     def __init__(
         self,
-        text: Text,
+        text: Optional[Text] = "",
```
wochinge, line 29: can it actually be  `None`?

tests/nlu/featurizers/test_count_vectors_featurizer.py
```diff
@@ -397,3 +404,102 @@ def test_count_vectors_featurizer_train():
...
+@pytest.mark.parametrize(
+    "sentence, action_name, action_text, action_name_features, response_features",
+    [
+        ("hello", "greet", None, [[1]], None),
+        ("hello", "greet", "hi", [[1]], [[1]]),
+        ("hello", "", "hi", None, [[1]]),
+    ],
+)
+def test_count_vector_featurizer_action_attribute_featurization(
+    sentence, action_name, action_text, action_name_features, response_features
```
wochinge, line 418: types please

wochinge: same for the others

rasa/nlu/training_data/message.py
wochinge: and a docstring please

rasa/nlu/model.py
```diff
@@ -385,3 +388,11 @@ def parse(
...
         output.update(message.as_dict(only_output_properties=only_output_properties))
         return output
+
+    def parse_message(self, message: Message, attribute: Text = TEXT) -> Message:
+        for component in self.pipeline:
+            if isinstance(component, Featurizer) or isinstance(
+                component, tokenizer.Tokenizer
+            ):
+                component.process(message, attribute, **self.context)
+        return message
```
wochinge, line 398: this is unused according to Pycharm. Can we remove it?

tests/nlu/featurizers/test_count_vectors_featurizer.py
evgeniiaraz: None of the featurizer tests have the types; should I add types in all of them? or do it in a separate PR later?

rasa/core/events/__init__.py
evgeniiaraz: I've simplified it a bit; in essence now when there is `e2e_text` in both `ActionExecuted`'s, we need to check it is equal to say that the `ActionExecuted`'s are equal

rasa/core/training/story_reader/story_reader.py
evgeniiaraz: done and removed!

rasa/nlu/training_data/message.py
evgeniiaraz: No, because it breaks reading from `json`, tries it before :)

rasa/nlu/model.py
evgeniiaraz: Pycharm is right but it is used in core featurization. I tried to add all the NLU changes here.

tests/nlu/featurizers/test_count_vectors_featurizer.py
wochinge: Depends on your time :-) For now I'd just add them for the ones you created

rasa/nlu/model.py
wochinge: Ok 👍 Sure it shouldn't be a method of `NaturalLanguageInterpreter` then?

rasa/nlu/training_data/message.py
wochinge: The docstring should also describe the parameters please (see https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e)

rasa/nlu/model.py
```diff
@@ -385,3 +388,11 @@ def parse(
         output = self.default_output_attributes()
         output.update(message.as_dict(only_output_properties=only_output_properties))
         return output
+
+    def parse_message(self, message: Message, attribute: Text = TEXT) -> Message:
+        for component in self.pipeline:
+            if isinstance(component, Featurizer) or isinstance(
+                component, tokenizer.Tokenizer
+            ):
```
wochinge, line 396:
```suggestion
            if isinstance(component, (Featurizer, tokenizer.Tokenizer)):
```

rasa/core/events/__init__.py
wochinge: where do we need the comparison?

rasa/nlu/model.py
evgeniiaraz: It calls this `Interpreter` inside it; that's why this function is needed here. I'll add a docstring though

rasa/nlu/training_data/message.py
Ghostvv: @evgeniiaraz I thought we agreed to remove `Message.text`

rasa/nlu/training_data/message.py
```diff
@@ -24,16 +26,16 @@
...
         output_properties: Optional[Set] = None,
         time: Optional[Text] = None,
         features: Optional[List["Features"]] = None,
         **kwargs,
     ) -> None:
-        self.text = text
         self.time = time
-        self.data = data if data else {}
+        self.data = data.copy() if data else {}
+        self.data.update({TEXT: text})
```
Ghostvv, line 38: why do we need that?

rasa/nlu/training_data/message.py
evgeniiaraz: `Message.text`  as a property is gone, so, you can't do `self.text` or anything. 
In the `__init__` I left it in because in many(many many many) places `Message` is not created through `build ` but through direct `Message(text, bla bla)`.

rasa/nlu/training_data/message.py
evgeniiaraz: For line 38: Partially, because of the `init` problem I described above. And because we need to make sure all the passed input is stored in the data. 
For line 37: because `text` as a definitive property is gone. and now imagine the case where I want to initialise 2 messages as follows:
```
data = {intent: my_intent}
texts = ['hello', 'hi']
message1 = Message(texts[0], data)
message2 = Message(texts[1], data)
```
then I want to change the intent in message1:
`message1.data["intent"] = "another_intent"`
 (if we don't make a copy) this will change the data of message2 as well because they came from the same `data` variable initially.

rasa/nlu/training_data/message.py
Ghostvv: @evgeniiaraz could you give examples of these places? How hard is it to update these places with passing `data{TEXT: text}`. keeping `text` in `init` seems to me as a bad solution

rasa/nlu/training_data/message.py
Ghostvv: let's update the places with passing `data` instead of text during initialization

rasa/core/events/__init__.py
evgeniiaraz: off the top of my head, in many tests; but I cant guarantee its not used anywhere else

rasa/nlu/training_data/message.py
evgeniiaraz: Done!

tests/nlu/featurizers/test_count_vectors_featurizer.py
evgeniiaraz: Done!

rasa/core/events/__init__.py
wochinge: If it's just for the tests then we shouldn't add it to the production code.

evgeniiaraz: Do you mean that the `__eq__` function should not be tehre at all?

rasa/nlu/training_data/message.py
evgeniiaraz: Done in #6394

rasa/nlu/training_data/message.py
evgeniiaraz: Done in #6394

m-vdb:
@wochinge @evgeniiaraz @Ghostvv I was about to ping you on this 🔔  to see if there is a good chance to make it for the 2.0.0a2 release tomorrow. Can you let us know if there are any roadblocks that prevent it to get merged in time? 🙏

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -218,8 +218,6 @@ def _get_message_tokens_by_attribute(
         """Get text tokens of an attribute of a message"""
         if message.get(TOKENS_NAMES[attribute]):
             return [t.lemma for t in message.get(TOKENS_NAMES[attribute])]
-        if attribute not in [INTENT, ACTION_NAME]:
-            return message.get(attribute).split()
```
Ghostvv: why do we need this?

Ghostvv:
@m-vdb we need a review for it

Ghostvv:
@evgeniiaraz as far as I understand, this is not a breaking PR and is independent of other PRs, so it can be directly against master

rasa/core/events/__init__.py
wochinge: In general I'd say: don't add production code which isn't required in production (e.g. only for tests). So I'd simply revert your changes. As the whole `__eq__` is probably only used in tests it's weird anyway :-D Imo it's also ok to keep it then. I'd just remove the `hasattr` then as it will always have the attribute

wochinge:
I'll give it another review 👍

wochinge:
@Ghostvv But there is other stuff in the `e2e` branch which needs to get into `master`, too

rasa/core/events/__init__.py
```diff
@@ -1021,7 +1022,14 @@ def __eq__(self, other) -> bool:
         if not isinstance(other, ActionExecuted):
             return False
         else:
-            return self.action_name == other.action_name
+            # if both have e2e_text compare both action name and e2e text
+            if hasattr(self, "e2e_text") and hasattr(other, "e2e_text"):
```
wochinge, line 1026: as said in the comment: It's ok to keep. I'd remove the `hasattr` calls though as they are not required and add tests for the method

rasa/core/training/story_reader/markdown_story_reader.py
```diff
@@ -173,7 +182,14 @@ def _parse_event_line(line):
             )
             return "", {}
 
-    async def _add_user_messages(self, messages, line_num):
+    @staticmethod
+    def _parse_bot_message_e2e(line: Text):
```
wochinge, line 186: can you please add the return type?

rasa/core/training/story_reader/markdown_story_reader.py
```diff
@@ -154,7 +163,7 @@ def _clean_up_line(line: Text) -> Text:
         return re.sub(r"<!--.*?-->", "", line).strip()
 
     @staticmethod
-    def _parse_event_line(line):
+    def _parse_event_line(line: Text):
```
wochinge, line 166: return type please

rasa/nlu/model.py
```diff
@@ -261,7 +264,11 @@ class Interpreter:
     # that will be returned by `parse`
     @staticmethod
     def default_output_attributes() -> Dict[Text, Any]:
-        return {"intent": {INTENT_NAME_KEY: None, "confidence": 0.0}, "entities": []}
+        return {
+            "text": "",
+            "intent": {INTENT_NAME_KEY: None, "confidence": 0.0},
+            "entities": [],
```
wochinge, line 270: let's use the constants from `rasa.nlu.constants` here

rasa/nlu/training_data/message.py
wochinge, line 38:
```suggestion
        self.data[TEXT] = text
```

rasa/nlu/training_data/message.py
```diff
@@ -42,22 +44,18 @@ def __init__(
             self.output_properties = output_properties
         else:
             self.output_properties = set()
+        self.output_properties.add(TEXT)
```
wochinge, line 47: Why can't we do it as for the other output properties?

rasa/nlu/training_data/message.py
```diff
@@ -103,6 +99,15 @@ def build(
         entities: List[Dict[Text, Any]] = None,
         **kwargs,
     ) -> "Message":
+        """
+        Builds a Message from UserUttered data
```
wochinge, line 103:
```suggestion
        Build a Message from `UserUttered` data.
```

rasa/nlu/training_data/message.py
```diff
@@ -114,6 +119,24 @@ def build(
             data[ENTITIES] = entities
         return cls(text, data, **kwargs)
 
+    @classmethod
+    def build_from_action(
+        cls,
+        action_text: Optional[Text] = "",
+        action_name: Optional[Text] = "",
+        **kwargs,
```
wochinge, line 127: Please add they type for `**kwargs` as well

rasa/nlu/training_data/message.py
```diff
@@ -114,6 +119,24 @@ def build(
...
+    ) -> "Message":
+        """
+        Builds a Message from ActionExecuted data
```
wochinge, line 130:
```suggestion
        Build a `Message` from `ActionExecuted` data.
```

rasa/nlu/training_data/message.py
evgeniiaraz: I'll leave it as is for now, because when we merge the `removal of text` PR these lines will be gone anyway.

rasa/nlu/training_data/message.py
evgeniiaraz: Because we want it to always include `text` in the output but if no output properties were passed explicitly it won't.  Previously, it was included because `text` was added separately from `self.text` .

rasa/nlu/model.py
```diff
@@ -261,7 +264,11 @@ class Interpreter:
     # that will be returned by `parse`
     @staticmethod
     def default_output_attributes() -> Dict[Text, Any]:
-        return {"intent": {INTENT_NAME_KEY: None, "confidence": 0.0}, "entities": []}
+        return {
+            TEXT: "",
+            INTENT: {INTENT_NAME_KEY: None, "confidence": 0.0},
```
wochinge, line 269:
```suggestion
            INTENT: {INTENT_NAME_KEY: None, INTENT_CONFIDENCE_KEY: 0.0},
```

rasa/nlu/model.py
```diff
@@ -12,12 +12,15 @@
     IntentClassifier,
 )
 from rasa.nlu.components import Component, ComponentBuilder  # pytype: disable=pyi-error
+from rasa.nlu.featurizers.featurizer import Featurizer  # pytype: disable=pyi-error
+from rasa.nlu.tokenizers import tokenizer  # pytype: disable=pyi-error
 from rasa.nlu.config import RasaNLUModelConfig, component_config_from_pipeline
 from rasa.nlu.extractors.extractor import EntityExtractor  # pytype: disable=pyi-error
-from rasa.nlu.constants import INTENT_NAME_KEY
+from rasa.nlu.constants import INTENT_NAME_KEY, TEXT, INTENT, ENTITIES
```
wochinge, line 19:
```suggestion
from rasa.nlu.constants import INTENT_NAME_KEY, TEXT, INTENT, ENTITIES, INTENT_CONFIDENCE_KEY
```

Issue #6166, opened 2020-07-08
New docs: Setup build of docs in CI


m-vdb:
⬆️ Relates to https://github.com/RasaHQ/growth/issues/1436

We should add a new workflow to test the new docs and build them upon release:
- current docs are tested [here](https://github.com/RasaHQ/rasa/blob/master/.github/workflows/continous-integration.yml#L103-L108). We can add a new step to this workflow to test the new docs build
- the docs are built and released in [this separate workflow](https://github.com/RasaHQ/rasa/blob/3918d9a30f8e26057ca75eb05d1307218def7844/.github/workflows/documentation.yml#L1-L63). We should add a new job in this workflow, that will take care of building the new docs.

💡 the current building system works as follows:
- code is checked out, sphinx builds the doc files
- we push the doc files to a `docs` branch on the same repo
- Netlify reacts (need details) to this and pulls the changes

We could mirror this and create `newdocs` branch. That way we do not override the existing docs and can still build the new docs. The Netflify part is not handled in this ticket.

lunelson:
@m-vdb we are at 9 out of 9 members on our Netlify "team", but some of these can probably be removed and you could be added; then I could walk you through the configuration

![](https://user-images.githubusercontent.com/1242864/87169261-fbfe7e80-c2cf-11ea-99c2-08db51a9de35.png)

m-vdb:
let's wait and see if Tom wants to handle that part. Otherwise we'll figure it out (maybe Rajesh doesn't need access?)

tmbo:
added you on netlify

tmbo:
> The Netflify part is not handled in this ticket.
I guess it is now? :D

tmbo:
Regarding the ticket:
@m-vdb before you go ahead and add a build step for the new docs we (@lunelson) should align on how it should look like (let me know if that already happened):
- all of the old docs should stay where they are for now (no change to the `docs` branch on rasahq/rasa and no change to its netlify config)
- new docs will be pushed to similar setup as we had before:
   - build docs in github ci on push to release branches (releasing docs based on master is fine for now, but won't work when we start working on 2.1)
   - built artefacts are pushed to rasahq/rasa@documentation (feel free to choose a different name other than `documentation`, can't use `docs` though - see above)
   - netlify deploys that branch

m-vdb:
yes this is exactly what we discussed 👍

lunelson:
[...]

Issue #6202, opened 2020-07-14
New docs: add support for program-output directive


m-vdb:
⬆️ Relates to RasaHQ/growth#1436

With sphinx, we had a directive named `.. program-output::` that would make it possible to run a command and display it's output. We need this with docusaurus. One way of doing it is by using variables (see the discussion on [this commit](https://github.com/RasaHQ/rasa/commit/01376059d1d062c600ead167d7db87e7858ae8b7#r40589934)).

## Implementation proposal

What I think we can do (similarly to what we did with `variables.json`):
1. declare a syntax for this directive
2. write a custom pre-processor that generates all command outputs and saves them inside the `docs/sources` folder (for instance)
3. write a custom remark plugin that looks for these output inside the `docs/sources` folder. (see `remark-sources` plugin for reference)

That way it would work both in dev and in production (where there are multiple versions). The key is to keep things under the `docs` folder.

Issue #6199, opened 2020-07-13
New docs: include source files


m-vdb:
⬆️ Relates to RasaHQ/growth#1436

We need to support the previous `.. literalinclude::` directives. We can use `remark-sources` and manually replace (it cannot be done by the rst-to-markdown converter because sphinx loads the source file long before it's handed to the converter).

💡 We are dropping support for "parts" of a file (for instance see `:start-after:` or `:pyobject:` options). This will require some documentation work.

## Implementation proposal

What I think we can do (similarly to what we did with `variables.json`):
1. keep the links as is in the `mdx` files
2. write a custom pre-processor that copies all "source" files inside the `docs/sources` folder (for instance)
3. write a custom remark plugin that looks for these files inside the `docs/sources` folder (or the original path). We could fork `remark-sources` and do that.

That way it would work both in dev and in production (where there are multiple versions). The key is to keep things under the `docs` folder.

tmbo:
I think we can reduce this to the minimum and really only include complete files. In most locations, this should be fine (or alternatively even add a simpler example to the docs). I'd really only go with complete includes for now and rather rework the docs to avoid using line or object based includes (e.g. most of them are in the events section, which doesn't make sense anyways - since users are using events from rasa-sdk and not from rasa open source)

m-vdb:
yes, fine by me. @erohmensing do you think that what @tmbo is easily doable? We might need your input to rework the content :)

erohmensing:
Sure, we can probably handle that. In fact it might help us make our code a little cleaner, and will make sure that the user always has the necessary imports to use any code referenced. 

Looks like the `Forms` and `Reminders and External Events` page will need a rework to manage referencing the content, and the `Events` page should get a rework for other reasons (for right now it might just make sense to take these includes out and deal with the rest of that page later)

Other than that, would be helpful for the `Custom Connectors` page if `RestInput` got its own module instead of being part of `channels.py`  (which I think would make sense in any case, as it's not intuitive to find in the codebase)

If that sounds good I'm happy to take on the content side of these changes -- if you assign me to a PR you make to do the backend changes for the existing includes, I can update these pages on there 👍

m-vdb:
great, thanks @erohmensing !

PR #6211, opened 2020-07-15
New docs ci build

m-vdb:
**Proposed changes**:
Fixes https://github.com/RasaHQ/rasa/issues/6166
Here is how it works inside the "Documentation" workflow:
1. When building for master or a tag, run the workflow
2. Create a temp directory and clone the [`documentation` branch](https://github.com/RasaHQ/rasa/tree/documentation) in it
3. Copy the `docs/` folder (together with the sources required by the docs) inside this temp folder
4. In case of a new version (not on master basically), cut a new version of the doc (cf. [Docusaurus versioning](https://v2.docusaurus.io/docs/versioning))
5. If the git directory is "dirty", commit the changes and push the `documentation` branch
6. Let `Netflify` handles the build: https://rasahq-docsv2-rasa.netlify.app/


.github/workflows/continous-integration.yml
```diff
@@ -89,24 +101,11 @@ jobs:
...
-    # FIXME: https://github.com/RasaHQ/rasa/issues/6166
-    # - name: Test Docs 📃
-    #   run: |
-    #     make install-full
-    #     poetry run python -c "from scripts import release; release.generate_changelog('major.minor.patch')"
-    #     export SPHINXBUILD=$(poetry run which sphinx-build)
-    #     cd docs && make SPHINXBUILD=$SPHINXBUILD SPHINXOPTS="-W --keep-going -A html_theme=rasabaster" html && make SPHINXBUILD=$SPHINXBUILD linkcheck
+    - name: Test Docs 📃
+      run: |
+        make install-full
```
tmbo, line 106: is that needed, did we not run this in L90 already?

.github/workflows/documentation.yml
```diff
@@ -45,20 +54,30 @@ jobs:
...
-
-    #     git config --global user.email "builds@github-ci.com"
-    #     git config --global user.name "GitHub CI"
-    #     git remote set-url --push origin "git@github.com:${{github.repository}}"
-
-    #     poetry run sphinx-versioning push docs docs . -- -b dirhtml -A html_theme=rasabaster
+      run: make install install-docs
+
+    - name: Build Docs variables 🧶
+      run: cd $DOCS_FOLDER && yarn run variables
```
tmbo, line 68: should we do this in the ci workflow as well?

.github/workflows/continous-integration.yml
m-vdb: ah yes that's true, copy-paste 🤷

.github/workflows/documentation.yml
m-vdb: yep, good catch

newdocs/package.json
```diff
@@ -5,8 +5,7 @@
   "scripts": {
     "start": "yarn run variables && docusaurus start",
     "build": "yarn run variables && docusaurus build",
-    "swizzle": "docusaurus swizzle",
-    "deploy": "docusaurus deploy",
+    "new-version": "docusaurus docs:version"
```
tmbo, line 8:
```suggestion
    "new-version": "docusaurus docs:version",
```

scripts/push_docs_to_branch.sh
```diff
@@ -0,0 +1,30 @@
...
+mkdir $TMP_DOCS_FOLDER
+git clone --depth=1 --branch $DOCS_BRANCH git@github.com:$GITHUB_REPOSITORY.git $TMP_DOCS_FOLDER
+
+echo "Updating the docs..."
+cp $DOCS_FOLDER/* $TMP_DOCS_FOLDER/
+cd $TMP_DOCS_FOLDER
+
+if [ -z "$NEW_VERSION" ]
+then
+    echo "Skipping generation of new version."
```
tmbo, line 19: in this case, is there actually anything we change? 🤔

tmbo: as in: I don't see anything we'd need to push in the lines after this `if`

scripts/push_docs_to_branch.sh
```diff
@@ -0,0 +1,30 @@
...
+else
+    echo "Generating docs for new version $NEW_VERSION..."
+    yarn run new-version $NEW_VERSION
```
tmbo, line 22: this doesn't actually build the docs does it? it only creates a version, so not sure what we are pushing to git in this case either

scripts/push_docs_to_branch.sh
m-vdb: there might be changes under the `docs/` folder or the source files imported by the docs (the former `literalinclude` directive). That's why we copy them

scripts/push_docs_to_branch.sh
m-vdb: yes exactly, it creates a new version. Then Netflify will run the build command by itself.

tmbo: wait, so does the `documentation` branch contain rendered HTML pages or `mdx` pages?

m-vdb: mdx pages. Otherwise we would have to version minified JS and HTML :/

m-vdb: I've updated the PR desc, it should be clearer :)

scripts/push_docs_to_branch.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -Eeuo pipefail
+
+RASA_FOLDER=rasa
+TESTS_FOLDER=tests
+DATA_FOLDER=data
+TMP_DOCS_FOLDER=tmp-documentation
+TODAY=`date "+%Y%m%d"`
+# we build new versions only for majors
+PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.0\\.0$"
```
m-vdb, line 11: we can control which version we want to push from this regex

newdocs/README.md
```diff
@@ -26,8 +26,4 @@ This command generates static content into the `build` directory and can be serv
 
 ### Deployment
```
m-vdb, line 27: I was thinking about adding the Netlify deploy status badge to this README or the repo's README, WDYT?

.github/workflows/documentation.yml
```diff
@@ -45,20 +53,29 @@ jobs:
...
+      run: |
+        eval "$(ssh-agent -s)"; touch docs_key; chmod 0600 docs_key
+        echo "$GITHUB_DOCS_KEY" > "docs_key"
+        ssh-add docs_key
+
+        git config --global user.email "builds@github-ci.com"
+        git config --global user.name "GitHub CI"
+        git remote set-url --push origin "git@github.com:${{github.repository}}"
+
+        ./scripts/push_docs_to_branch.sh
```
m-vdb, line 81: I still need to run this in the context of GH actions

m-vdb: 7a9fb8532d9

m-vdb: I ran the workflow for these use cases (and iteratively fixing the bugs):
- push to master
- push to version but not major
- push to major version

For each, the Netlify deployment worked properly. The current latest commit in the [`documentation` branch](https://github.com/RasaHQ/rasa/tree/documentation) is the one with a new version, that way we see how it looks like in terms of folder structure / website experience. I'll cleanup the branch history after the review

m-vdb: "next" version is available at this URL: https://rasahq-docsv2-rasa.netlify.app/next/
Screenshot 2020-07-16 at 09 21 27

Default version is the latest version pushed (2.0.0 in this test case):
Screenshot 2020-07-16 at 09 21 08

tmbo: regarding versions: 
- versions don't show up on the docs (i.e. there is no way to select a version). is there some configuration missing, or did the docosaurus docs site do something special?
Bildschirmfoto 2020-07-16 um 10 09 45
- would be good to have the version banner that docosaurus uses as well, but looks like they added that to their custom theme, so seems it needs a bit more effort and we should add it to the list of things to do later 
https://github.com/facebook/docusaurus/blob/5ccd24cc1f07efbd4e7c3fe68937e357c0e44855/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx#L126

Bildschirmfoto 2020-07-16 um 10 19 14

newdocs/README.md
tmbo: sure :+1:

Makefile
```diff
@@ -49,6 +49,9 @@ install-mitie:
 install-full: install install-mitie
 	poetry install -E full
 
+install-docs:
+	cd newdocs/ && yarn install
+
```
tmbo, line 54: let's also update the other make targets that work with the docs, e.g. https://github.com/RasaHQ/rasa/pull/6211/files#diff-b67911656ef5d18c4ae36cb6741b7965R86-R90

scripts/push_docs_to_branch.sh
```diff
@@ -0,0 +1,55 @@
...
+    NEW_VERSION=${GITHUB_REF/refs\/tags\//}
+fi
+
+# clone the $DOCS_BRANCH in a temp directory
+git clone --depth=1 --branch=$DOCS_BRANCH git@github.com:$GITHUB_REPOSITORY.git $TMP_DOCS_FOLDER
+
+echo "Updating the docs..."
+cp -R $DOCS_FOLDER/* $TMP_DOCS_FOLDER/docs
+# we need these because they are imported by docs files
+cp -R $RASA_FOLDER/* $TMP_DOCS_FOLDER/$RASA_FOLDER
```
tmbo, line 31: is there a reason not to copy over everything? this seems a bit brittle as it will break without errors if someone includes a file that is not located in e.g. `data/`

scripts/push_docs_to_branch.sh
```diff
@@ -0,0 +1,55 @@
...
+RASA_FOLDER=rasa
+TESTS_FOLDER=tests
+DATA_FOLDER=data
+TMP_DOCS_FOLDER=tmp-documentation
+TODAY=`date "+%Y%m%d"`
+# we build new versions only for majors
+PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.0\\.0$"
+MASTER_REF=refs/heads/master
+
+[[ ! $GITHUB_REF =~ $PATTERN_FOR_NEW_VERSION ]] \
```
tmbo, line 14: I am not sure this works as we'd want it to:
- we want to have separate documentation pages for the separate minor versions (e.g. one for 2.0 and another one for 2.1 and one for master/next)
- on a push to the release branch e.g. `2.1.x`, we need to update the `2.1` documentation (same for master, but that already works)

.github/workflows/continous-integration.yml
```diff
@@ -89,24 +101,12 @@ jobs:
...
-    #   run: |
-    #     make install-full
-    #     poetry run python -c "from scripts import release; release.generate_changelog('major.minor.patch')"
-    #     export SPHINXBUILD=$(poetry run which sphinx-build)
-    #     cd docs && make SPHINXBUILD=$SPHINXBUILD SPHINXOPTS="-W --keep-going -A html_theme=rasabaster" html && make SPHINXBUILD=$SPHINXBUILD linkcheck
+    - name: Test Docs 📃
+      run: |
+        poetry run python -c "from scripts import release; release.generate_changelog('major.minor.patch')"
+        cd newdocs
+        poetry run yarn variables
```
tmbo, line 108: should we add a make command, `make docs`, that does `yarn install && yarn variables && yarn build` - also for local usage

scripts/push_docs_to_branch.sh
```diff
@@ -0,0 +1,55 @@
...
+cp -R $DATA_FOLDER/* $TMP_DOCS_FOLDER/$DATA_FOLDER
+cp -R $TESTS_FOLDER/* $TMP_DOCS_FOLDER/$TESTS_FOLDER
+
+cd $TMP_DOCS_FOLDER
+
+if [ ! -z "$NEW_VERSION" ]
+then
+    echo "Generating docs for new version $NEW_VERSION..."
+    cd docs
+    yarn run new-version $NEW_VERSION
```
tmbo, line 41: I think there is another issue hidden here: it doesn't seem like docosaurus "scopes variables". This means that e.g. `{variables.release}` will get resolved to the same number for all versions.

.github/workflows/documentation.yml
m-vdb: Yes I think both are custom made. It's doable via [the theme](https://v2.docusaurus.io/docs/theme-classic#navbar-dropdown), and I guess we would only display the versions we want to display in this dropdown. Here is how Docusaurus [does it on their website](https://github.com/facebook/docusaurus/blob/v2.0.0-alpha.58/website/docusaurus.config.js#L113-L134).

.github/workflows/continous-integration.yml
m-vdb: the build isn't so useful for local usage, but sure, I think it makes sense (especially since running `poetry run yarn` is not really intuitive)

scripts/push_docs_to_branch.sh
m-vdb: 😱 wow this one, I didn't see it coming. I assumed that the `src` folder was moved inside each docs' version.

scripts/push_docs_to_branch.sh
m-vdb: - I thought we discussed we wanted the docs only for majors; will update the regex and it will take care of it
- I totally missed this part of the workflow :/

scripts/push_docs_to_branch.sh
m-vdb: yeah I agree. At first I thought that we'd rather keep the `documentation` branch as "clean" as possible and not include everything (code, tests, scripts, etc...) but it seems that it will bring up more issues than solving them. I was on the fence, thanks for your input!

scripts/push_docs_to_branch.sh
m-vdb: @tmbo something I am wondering: instead of listening to a push on a release branch + finding out the version "manually", why not listening to the push of the patch tag (which is done from the release branch), and modify the script to do a special case for this one? It would simplify the implementation I think and we wouldn't lose any functionality.

m-vdb: 456caef223d

.github/workflows/documentation.yml
m-vdb: version dropdown added in 6fab2bfd17f and I updated the desc of https://github.com/RasaHQ/growth/issues/1501 to handle the version banner

scripts/push_docs_to_branch.sh
m-vdb: I think I found a solution with c47348ed22b but it will require some testing

m-vdb: aaaand, it works -> https://rasahq-docsv2-rasa.netlify.app/

tmbo: nice :+1: we prob need to do the same thing if we want to add e.g. the output of `rasa --help` to the docs

newdocs/docs/api/tracker-stores.mdx
```diff
@@ -3,7 +3,7 @@ id: tracker-stores
 sidebar_label: Tracker Stores
 title: Tracker Stores
 ---
-import variables from '@site/src/variables';
+import variables from '@site/docs/variables.json';
```
tmbo, line 6: why do we use the `@site` syntax here (while we explicitly don't for the others?)

scripts/push_docs_to_branch.sh
m-vdb: yes I think so too

newdocs/docs/api/tracker-stores.mdx
m-vdb: I omitted this one 😬

tmbo:
One open question I have: do you think we can handle the versioning properly for the dynamic content we have? (code snippets loaded from source files, `rasa --help` output) because we also need to circumvent this dynamic content, basically we need to handle them before we copy things into the versioned folders

m-vdb:
@tmbo I don't think there is a way to do this out of the box. What I think we can do (similarly to what I did with `variables.json`) for the snippets loaded from files:
1. keep the links as is in the `mdx` files
2. write a custom pre-processor that copies all "source" files inside the `docs/sources` folder (for instance)
3. write a custom remark plugin that looks for these files inside the `docs/sources` folder (or the original path). We could fork `remark-sources` and do that.

For the command outputs, it would work similarly I think.

That way it would work both in dev and in production. It wouldn't hinder the experience of writing the docs IMO. The key is to keep things under the `docs` folder. So doable, but not free :)

tmbo:
Ok sounds good - do we have an issue for that? 

I’ll do another review and then we can merge this one 🙌

m-vdb:
@tmbo yes we do: https://github.com/RasaHQ/rasa/issues/6199 and https://github.com/RasaHQ/rasa/issues/6202

I'll update the descriptions with my comment

m-vdb:
fixing the failing build on master with https://github.com/RasaHQ/rasa/pull/6251

PR #6212, opened 2020-07-15
update docs root page

akelad:
change the docs root page to point to 1.10.8, rather than the alpha. ideally this should just pick the latest 1.10.x release but i'm not sure how to make it do that in this conf file. 

Also I'm like 95% sure these are the correct configurations here, but @lunelson is there a way to preview what this is gonna look like before publishing?


degiz:
@akelad Locally yo can run `poetry run sphinx-versioning build docs . -b dirhtml -A html_theme=rasabaster`

akelad:
yeah i did that but it doesn't show me the versions in the sidebar 🤔

akelad:
i am trying for a second time now

degiz:
DId you check the `./html_theme=rasabaster/index.html` ? I generated the log yesterday and can see the tags on the left side.

lunelson:
@akelad I'm guessing you and @degiz know more about this than I do, I'm not at all familiar with the sphinx build process :grimacing:

akelad:
ahhh @degiz ok yes i think i looked at the wrong index.html 🙃

akelad:
confirmed that it works - landing page is now 1.10.8. This message for 2.0.0a1 is somewhat untrue, but fine imo
image

akelad:
yes good point, will do.

Issue #5377, opened 2020-03-06
CRFEntityExtractor splits one entity into two


tabergma:
**Rasa version**: 1.8.0
**Python version**: 3.7
**Operating system** (windows, osx, ...): OSX

**Issue**:
If the `CRFEntityExtractor` is trained with `BILOU_flag` set to `False` entities consisting of two words are shown as individual entities and are not combined. 

In [this function](https://github.com/RasaHQ/rasa/blob/1c627fe2dbf92d41cfa3a685b0b66e5d1b4367dc/rasa/nlu/extractors/crf_entity_extractor.py#L323), we do not combine entities that have the same type and occur directly after each other. 

Forum thread: https://forum.rasa.com/t/lookup-table-not-working-for-entities-with-multiple-words/25305

**Command to reproduce bug**:
To reproduce the error, train the `restaurantbot` example with the config below.
```
rasa train nlu
```
If you execute afterwards `rasa shell nlu` and enter `north american food`, you will get something like:
```
  "entities": [
    {
      "start": 0,
      "end": 5,
      "value": "north",
      "entity": "cuisine",
      "confidence": 0.8311922393651153,
      "extractor": "CRFEntityExtractor"
    },
    {
      "start": 6,
      "end": 14,
      "value": "american",
      "entity": "cuisine",
      "confidence": 0.9248010334484645,
      "extractor": "CRFEntityExtractor"
    }
  ],
```

**Content of configuration file (config.yml)** (if relevant):
```yml
language: "en"

pipeline:
- name: "SpacyNLP"
- name: "SpacyTokenizer"
- name: "SpacyFeaturizer"
- name: "RegexFeaturizer"
- name: "CRFEntityExtractor"
  BILOU_flag: False
- name: "EntitySynonymMapper"
- name: "SklearnIntentClassifier"
```

HuppeJ:
Hello,

I had the same problem, but I solved it by adding nlu data with multiple words:

```
[…] [Costa Rica](country) […]
[…] [Dominican Republic](country) […]
[…] [Central African Republic](country) […]
[…] […] [Papua New Guinea](country) […]
[…] [Papua New Guinea](country) […]
...
```

After looking at the code, if I'm not mistaken, this shouldn't be a bug.

tabergma:
We refactored the way how we create entities during the last couple of months. This issue is solved on master.

iszainab:
[...]

PR #6214, opened 2020-07-15
Add RegexEntityExtractor

tabergma:
**Proposed changes**:
- closes https://github.com/RasaHQ/rasa/issues/5957
- Add an entity extractor that extracts entities based on a lookup table/regexes

original PR https://github.com/RasaHQ/rasa/pull/6183


AMR-KELEG:
O/
Will this PR solve this bug https://github.com/RasaHQ/rasa/issues/5377?

tabergma:
@AMR-KELEG This PR is not related to #5377. However, I just added a comment to the issue as it is solved already on master. 

`LookupEntityExtractor` directly extracts entities using a lookup table. It is not a feature extractor as the `RegexFeaturizer` that converts lookup tables into features that can then be used by our entity extractors.

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,134 @@
...
+    def _validate_lookup_entry(self, entity: Text, file_path: Text) -> None:
+        if file_path is None:
+            # remove the entity from the lookup dictionary,
+            # if the file path is not provided
+            self.lookup.pop(entity)
+            message = (
+                f"No file path for entity '{entity}' was given. "
+                f"Please provide a valid file path."
+            )
+            common_utils.raise_warning(message)
```
koaning, line 76: Is raising a warning here enough or might it be better just to raise an error?

rasa/nlu/extractors/lookup_entity_extractor.py
```diff
@@ -0,0 +1,134 @@
...
+            entities += self._extract_entities(user_message, entity, file_path)
+
+        return entities
+
+    def _extract_entities(
+        self, user_message: Text, entity: Text, file_path: Text
+    ) -> List[Dict[Text, Any]]:
+        """Extract entities of the given type from the given user message."""
+        entities = []
+        _user_message = user_message
```
koaning, line 102: might `copy_user_message` be a better variable name?

tests/nlu/extractors/test_lookup_entity_extractor.py
```diff
@@ -0,0 +1,186 @@
...
+        {"lookup": lookup_table, "lowercase": lowercase}
+    )
+    entity_extractor.process(message)
+
+    entities = message.get(ENTITIES)
+    assert entities == expected_entities
+
+
+def test_do_not_overwrite_any_entities():
+    message = Message("Max lives in Berlin.")
```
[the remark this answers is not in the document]
koaning: The lookup table might also still be generating features now that I think of it ... meaning it may not just be spaCy but potentially also the CRF entity detector that picks up similar entities.

tabergma:
@koaning I briefly discussed this new extractor with Vova and he mentioned that we should reuse the lookup definition from the training data (see [docs](https://rasa.com/docs/rasa/nlu/training-data-format/#lookup-tables)). I agree with him. What do you think? If you agree I would update the PR once again.

koaning:
I don't see any strong reasons not to do it, but I do wonder. This way we might generate both features and labels based on the lookup definition. Is that still intended behavior?

tabergma:
I think that is fine. We can add a warning if both are used inside the pipeline.

koaning:
Mhm actually. We should check one thing here. From the documentation found [here](https://rasa.com/docs/rasa/nlu/training-data-format/#lookup-tables);

> When lookup tables are supplied in training data, the contents are combined into a large, case-insensitive regex pattern that looks for exact matches in the training examples. These regexes match over multiple tokens, so lettuce wrap would match get me a lettuce wrap ASAP as [0 0 0 1 1 0]. These regexes are processed identically to the regular regex patterns directly specified in the training data.

Does the current implementation allow for general regexes to be passed in the lookup file? Is that something we'd want or is that maybe a different feature like `RegexEntityExtractor`?

tabergma:
My current plan is to combine them in a `RegexEntityExtractor`. So similar to the `RegexFeaturizer` which converts lookup tables and regexes defined in the training data into features, the `RegexEntityExtractor` would use the lookup tables and regexes in the training data to directly extract entitites.

ValeKnappich:
> My current plan is to combine them in a `RegexEntityExtractor`. So similar to the `RegexFeaturizer` which converts lookup tables and regexes defined in the training data into features, the `RegexEntityExtractor` would use the lookup tables and regexes in the training data to directly extract entitites.

That sounds great. I'm currently having trouble with the RegexFeaturizer and thought about using a custom component RegexEntityExtractor. When do you think this will be included in a release?

tests/nlu/extractors/test_lookup_entity_extractor.py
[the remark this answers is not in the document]
tabergma: We don't overwrite any entities, we just add them to the list of already extracted entities. It might happen that if you have multiple entity extractors in the pipeline, the same entity is extracted multiple times. But this should not cause any issue.

tabergma:
@ValeKnappich The implementation is almost ready, so we are going to merge the feature soon into master. The next release we are planning is the 2.0 release which should happen somewhen this quarter. We will release some release candidates beforehand, I think the next one is planned for mid August.

tabergma:
@Ghostvv Just adding you as a reviewer to take a second look at the code as @koaning isn't that familiar with the codebase yet.

rasa/nlu/components.py
```diff
@@ -102,6 +102,32 @@ def validate_only_one_tokenizer_is_used(pipeline: List["Component"]) -> None:
...
+    regex_featurizer_used = False
+    regex_entity_extractor_used = False
+    for component in pipeline:
+        if isinstance(component, RegexFeaturizer):
+            regex_featurizer_used = True
+        elif isinstance(component, RegexEntityExtractor):
+            regex_entity_extractor_used = True
+
+    if regex_featurizer_used and regex_entity_extractor_used:
+        raise InvalidConfigError(
```
Ghostvv, line 125: I would not raise here, but rather make a debug statement. I could imagine a situation where it useful to have both

tabergma: Just had a brief discussion with @koaning about it. Don't you think it might be confusing for users as the lookup tables/regexes are used differently/have a different meaning depending on the component? Also in the docs we point our that you can also use regexes for intent classification to indicate an utterance follows a certain schema and should most likely be classified as intent xyz. Those regexes should not be there if you use the `RegexEntityExtractor`.

changelog/5957.feature.rst
```diff
@@ -0,0 +1,4 @@
+Add new entity extractor ``RegexEntityExtractor``. The entity extractor extracts entities using the lookup tables
+and regexes defined in the training data. You can either use the ``RegexEntityExtractor`` or the ``RegexFeaturizer``
+in your pipeline. If you use both components an error will be thrown.
```
Ghostvv, line 3: let's remove the error. Why not use both components?

docs/nlu/components.rst
```diff
@@ -951,6 +956,36 @@ EntitySynonymMapper
...
+        It is not possible to do both at the same time.
+
+:Configuration:
+    By default we convert the entries in the lookup file and the user message to lowercase before we check if there
+    is a match. You can disable this behavior by setting ``lowercase: False``.
+
+    .. code-block:: yaml
+
+        pipeline:
+        - name: LookupEntityExtractor
```
Ghostvv, line 985:
```suggestion
        - name: RegexEntityExtractor
```

docs/nlu/components.rst
```diff
@@ -951,6 +956,36 @@ EntitySynonymMapper
...
+    This component extract entities using the lookup tables and regexes defined in the training data.
+    The component checks if the user message contains an entry of one of the lookup tables or matches one of the
+    regexes. If a match is found, the value is extracted as entity.
+
+    .. note::
+        You cannot use the :ref:`RegexFeaturizer` and the ``RegexEntityExtractor`` in the same pipeline. You need to
+        decide whether you want to use the lookup tables and regexes to create features or to directly extract entities.
+        It is not possible to do both at the same time.
+
+:Configuration:
```
Ghostvv, line 978: I'd add a config to use only lookup or only regex or both. E.g. I can imagine when users want to use only lookup tables

changelog/5957.feature.rst
Ghostvv: e.g. regexentityextractor with only lookup (see comment below) and actual regexes as features

rasa/nlu/components.py
tabergma: Not sure what you mean. Currently, `RegexFeaturizer` produces sentence and sequence features because of this. Do you suggest to remove the sentence features and update the docs to just mention that those features are good for entities?

docs/nlu/components.rst
tabergma: Ah ok, so you would do something like this?
```
- name: RegexEntityExtractor
  use_lookup_tables: True
  use_regex_features: False
  lowercase: True
```
Would you then also add the same to the `RegexFeaturizer`?

Ghostvv: yes, we could do that, but I think it's fine for featurizer to use all

rasa/nlu/components.py
Ghostvv: No, I meant in this case we only sequence regexes for entities

Ghostvv: then intent regexes will not influence `RegexEntityExtractor`

docs/nlu/components.rst
tabergma: Ok. Vincent mentioned a similar point but we weren't sure if the config would become even more complicated with this addition. But I guess we can add this.

rasa/nlu/components.py
tabergma: Currently we cannot distinguish between intent and entity regexes. But maybe I guess we could filter out only those regexes in the `RegexEntityExtractor` which have a name that is in the list of entities. Will think about something.

docs/nlu/components.rst
Ghostvv: since we have defaults, this config becomes entirely optional

rasa/nlu/extractors/regex_entity_extractor.py
```diff
@@ -0,0 +1,132 @@
...
+    """Searches for entities in the user's message using the lookup tables and regexes
+    defined in the training data."""
+
+    defaults = {
+        # text will be processed with case insensitive as default
+        "case_sensitive": False,
+        # use lookup tables to extract entities
+        "use_lookup_tables": True,
+        # use regex features to extract entities
+        "use_regex_features": True,
```
Ghostvv, line 35:
```suggestion
        # use regexes to extract entities
        "use_regexes": True,
```

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
```diff
@@ -33,8 +33,12 @@ def required_components(cls) -> List[Type[Component]]:
...
 
     defaults = {
-        # Text will be processed with case sensitive as default
-        "case_sensitive": True
+        # text will be processed with case sensitive as default
+        "case_sensitive": True,
+        # use lookup tables to generate features
+        "use_lookup_tables": True,
+        # use regex features to generate features
+        "use_regex_features": True,
```
Ghostvv, line 41:
```suggestion
        # use regexes to generate features
        "use_regexes": True,
```

data/test_config/embedding_random_seed.yaml
```diff
@@ -1,3 +1,5 @@
 policies:
 - name: TEDPolicy
   random_seed: 42
+  validation_split: 0
+  max_history: 5
```
Ghostvv, line 5: let's remove the whole ymal. It's now the same as ted yaml

tabergma: Just wanted to test if it is related to the config. Seems like that `max_history` makes some kind of difference... The option `validation_split` does not exist anymore, if I am not mistaken. Strange. Any idea why this could be? But can remove the file from the test for now.

tabergma: Just wanted to test if it is related to the config. Seems like that max_history makes some kind of difference... The option validation_split does not exist anymore, if I am not mistaken. Strange. Any idea why this could be? But can remove the file from the test for now.

tabergma: ok, it is still failing... will remove the file

Ghostvv: that's very strange, we have 2 identical configs, but one is failing and the other is not

tabergma: is there anything that we store between the two runs? a cache or something that is not cleared?

Ghostvv: I don't know

PR #6229, opened 2020-07-16
moodbot with rules & yml files

ArjaanBuijk:
**Proposed changes**:
- Use rules for greet, goodbye & bot_challenge
- Leave the happy & sad paths as stories


akelad:
@ArjaanBuijk thanks for submitting that :D The stories also need to be converted to the yaml format, and so does the nlu training data - could you do that as well? not sure if any other changes are needed e.g. in the domain file, but it's good to take a look at the `rules` folder in the examples

examples/moodbot/data/stories.yml
```diff
@@ -0,0 +1,23 @@
+stories:
+
+- story: happy path
+  steps:
+  - intent: mood_great 
+  - action: utter_happy
+
+- story: sad path 1
+  steps:
+  - intent: mood_unhappy
```
akelad, line 10: i actually wonder if this is the correct split here 🤔  because "mood_unhappy" is the direct response to the bots greeting (it says "hi how are you" or something)

ArjaanBuijk: @akelad ,

I restored the greet->utter_greet as starting point of all the stories, and removed the corresponding rule.

akelad:
@ArjaanBuijk do you have time to fix the tests?

tmbo:
I based my PR in #6323 on this and I think I fixed the tests (we'll see if CI passes, fingers crossed). so might make sense to close this one and just merge the other one.

akelad:
sounds fine to me

PR #6195, opened 2020-07-13
rule policy documentation

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6184

Open Todos
- [x] add rules to glossary
- [x] finish migration guide or decide to wait for migration
- [x] add links from stories to rules
- [x] describe how to submit form via rule

**Required code todos** (things which are described in the docs, but don't work yet as they are not yet implemented):
- [x] Core fallback for rules https://github.com/RasaHQ/rasa/issues/6208
- [x] re-adding `ambiguity_threshold` https://github.com/RasaHQ/rasa/issues/6244
- [ ] custom actions to ask for slot https://github.com/RasaHQ/rasa/issues/6185

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [x] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6230, opened 2020-07-16
Concert Bot Rules

b-quachtran:
**Proposed changes**:
- Added RulePolicy to config, removed fallback & mapping policies
- Implemented chitchat & nlu_fallback intents as rules


akelad:
@b-quachtran thanks for submitting this - the same goes here as in Arjaans PR, the stories should also be converted to yaml and so should the nlu training data (if there is any)

akelad:
@b-quachtran what about the stories for searching venues/concerts?

b-quachtran:
@akelad Oops sorry .gitignore was set to ignore changes in the concertbot/data folder -_-

examples/concertbot/data/nlu.yaml
```diff
@@ -0,0 +1,39 @@
+nlu:
+  - intent: greet
+    examples: |
+      hi
```
akelad, line 4: can you add the `-` prefix here? Like in Arjaans PR: https://github.com/RasaHQ/rasa/pull/6229/files#diff-ed593145bcc162429c6f89a64f2fcc0dR5

examples/concertbot/data/rules.yaml
```diff
@@ -0,0 +1,27 @@
+rules:
+  - rule: fallback
+    steps:
+      - ...
+      - intent: nlu_fallback
```
akelad, line 5: did you test this works? I personally have no idea how the fallback works in the new setup. Doesn't the threshold need to be defined somewhere?

b-quachtran: I implemented it based on Tobias' doc https://github.com/RasaHQ/rasa/pull/6195/files#diff-9bbb964215f342154b3f976a61d47d2cR43. Just updated the config file and it's working now

PR #6242, opened 2020-07-20
Windows support and testing

alwx:
**Proposed changes**:
- fixes the existing tests and make it possible to test Rasa on Windows


pyproject.toml
```diff
@@ -201,4 +202,4 @@ extras = [ "tls", "srv",]
 git = "https://github.com/RasaHQ/sphinxcontrib-versioning.git"
 
 [tool.poetry.dev-dependencies.sphinx_rtd_theme]
-git = "https://github.com/RasaHQ/sphinx_rtd_theme.git"
+git = "https://github.com/RasaHQ/sphinx_rtd_theme.git"
```
wochinge, line 205:
```suggestion
git = "https://github.com/RasaHQ/sphinx_rtd_theme.git"

```

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+markers =
+    win32: mark a test as a Windows test.
+    darwin: mark a test as a MacOS test.
+    linux: mark a test as a Linux test.
```
wochinge, line 5:
```suggestion
    linux: mark a test as a Linux test.

```

tests/conftest.py
```diff
@@ -271,6 +273,14 @@ def random_user_uttered_event(timestamp: Optional[float] = None) -> UserUttered:
     )
 
 
+def pytest_runtest_setup(item):
+    supported_platforms = PLATFORMS.intersection(
+        mark.name for mark in item.iter_markers()
+    )
+    if supported_platforms and sys.platform not in supported_platforms:
+        pytest.skip("cannot run on platform {}".format(sys.platform))
```
wochinge, line 281:
```suggestion
        pytest.skip(f"cannot run on platform {sys.platform}")
```

tests/conftest.py
```diff
@@ -271,6 +273,14 @@ def random_user_uttered_event(timestamp: Optional[float] = None) -> UserUttered:
     )
 
 
+def pytest_runtest_setup(item):
```
wochinge, line 276: please add the type

tests/conftest.py
```diff
@@ -40,6 +41,7 @@
     INCORRECT_NLU_DATA,
 )
 
+PLATFORMS = set("darwin linux win32".split())
```
wochinge, line 44: why not
```suggestion
PLATFORMS = {"darwin", "linux", "win32"}
```

tests/core/test_data.py
```diff
@@ -158,7 +159,7 @@ def test_find_nlu_files_with_different_formats(test_input, expected):
     examples_dir = "data/examples"
     data_dir = os.path.join(examples_dir, test_input)
     core_files, nlu_files = data.get_core_nlu_files([data_dir])
-    assert nlu_files == expected
+    assert [Path(f) for f in nlu_files] == [Path(f) for f in expected]
```
wochinge, line 162: why did we need to change that?

tests/nlu/featurizers/test_convert_featurizer.py
```diff
@@ -83,6 +87,8 @@ def test_convert_featurizer_train(component_builder):
         ("ńöñàśçií", "ńöñàśçií"),
     ],
 )
+@pytest.mark.linux
+@pytest.mark.darwin
```
wochinge, line 91: optional: Having the `linux` and `darwin` marker implicitly means that the tests don't run on Windows. I wonder if turning it around (e.g. having `@pytest.mark.skip_on_windows`) would make this more explicit

tests/nlu/test_train.py
```diff
@@ -90,6 +90,8 @@ def test_all_components_are_in_at_least_one_test_pipeline():
 
 
 @pytest.mark.parametrize("language, pipeline", pipelines_for_tests())
+@pytest.mark.linux
+@pytest.mark.darwin
```
wochinge, line 94: some of these pipelines should work on Windows. Can we be more selective instead of skipping to run them all?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -118,6 +118,8 @@ def test_check_labels_features_exist(messages, expected):
         ],
     ],
 )
+@pytest.mark.linux
+@pytest.mark.darwin
 async def test_train_persist_load_with_different_settings(
```
wochinge, line 123: wouldn't the last pipeline also work on Windows?

tests/core/test_agent.py
```diff
@@ -52,7 +51,7 @@ async def model(request: Request) -> StreamingHTTPResponse:
 
 @pytest.fixture()
 def model_server(
-    loop: Loop, sanic_client: Callable, trained_moodbot_path: Text
+    loop: asyncio.AbstractEventLoop, sanic_client: Callable, trained_moodbot_path: Text
```
wochinge, line 54: 🚀

tests/cli/test_cli.py
```diff
@@ -30,6 +31,8 @@ def test_data_convert_help(run: Callable[..., RunResult]):
         assert output.outlines[i] == line
 
 
+@pytest.mark.linux
+@pytest.mark.darwin
 def test_version_print_lines(run: Callable[..., RunResult]):
```
wochinge, line 36: why does it not work on Windows? If the output is formatted differently, should we add a Windows only test then?

rasa/core/test.py
```diff
@@ -542,7 +542,7 @@ def _log_stories(
             f.write("\n\n")
 
 
-async def test(
+async def evaluate_stories(
```
wochinge, line 545: Having a function `test` in the module `test` was definitely a bad decision back then 🙈 However, we can't just change it as it's somewhat part of the public API of Rasa Open Source. Do we need that for the tests to pass on Windows? Might also make sense to create an issue to handle this separately

rasa/core/events/__init__.py
```diff
@@ -1272,6 +1272,13 @@ class SessionStarted(Event):
 
     type_name = "session_started"
 
+    def __init__(
+        self,
+        timestamp: Optional[float] = None,
+        metadata: Optional[Dict[Text, Any]] = None,
+    ) -> None:
+        super().__init__(timestamp, metadata)
+
```
wochinge, line 1281: this is not required, is it?
```suggestion
```

rasa/core/__init__.py
```diff
@@ -3,7 +3,6 @@
 import rasa
 
 from rasa.core.train import train
-from rasa.core.test import test
```
wochinge, line 6: please see my comment in `rasa/core/test.py`

tests/cli/test_cli.py
ricwo: +1 is there any way this can be checked on windows?

tests/cli/test_cli.py
```diff
@@ -30,6 +31,8 @@ def test_data_convert_help(run: Callable[..., RunResult]):
         assert output.outlines[i] == line
 
 
+@pytest.mark.linux
+@pytest.mark.darwin
```
ricwo, line 35: we could also leave this out and add an `xfail` on windows:
https://docs.pytest.org/en/latest/skipping.html#condition-parameter

tests/conftest.py
```diff
@@ -271,6 +273,14 @@ def random_user_uttered_event(timestamp: Optional[float] = None) -> UserUttered:
     )
 
 
+def pytest_runtest_setup(item: pytest.Item) -> None:
+    supported_platforms = PLATFORMS.intersection(
+        mark.name for mark in item.iter_markers()
+    )
+    if supported_platforms and sys.platform not in supported_platforms:
+        pytest.skip(f"cannot run on platform {sys.platform}")
```
ricwo, line 281: as @wochinge suggested, could we switch the decorators around and mark them as `skip_on_windows`?

tests/core/test_data.py
ricwo: is it because the `relative_normpath` function changed?

tests/core/test_ensemble.py
```diff
@@ -132,8 +132,8 @@ def test_policy_priority():
 def test_fallback_mapping_restart():
     domain = Domain.load("data/test_domains/default.yml")
     events = [
-        ActionExecuted(ACTION_DEFAULT_FALLBACK_NAME),
-        utilities.user_uttered(USER_INTENT_RESTART, 1),
+        ActionExecuted(ACTION_DEFAULT_FALLBACK_NAME, timestamp=1),
+        utilities.user_uttered(USER_INTENT_RESTART, 1, timestamp=2),
```
ricwo, line 136: is this required for windows tests to pass or just a general fix?

tests/nlu/test_train.py
```diff
@@ -59,6 +59,24 @@ def pipelines_for_tests():
...
+            "zh",
+            as_pipeline(
+                "MitieNLP", "JiebaTokenizer", "MitieFeaturizer", "MitieEntityExtractor"
+            ),
+        ),
+        ("fallback", as_pipeline("KeywordIntentClassifier", "FallbackClassifier")),
+    ]
+
+
+def pipelines_for_non_windows_tests():
```
ricwo, line 72: return type is missing

tests/test_server.py
```diff
@@ -207,9 +207,13 @@ def send_request() -> None:
     train_request.terminate()
 
 
-# due to unknown reasons this test can not be run in pycharm, it
-# results in segfaults...will skip in that case - test will still get run on CI
+# Due to unknown reasons this test can not be run in pycharm, it
+# results in segfaults...will skip in that case - test will still get run on CI.
+# It also doesn't run on Windows because of Process-related calls and an attempt
```
ricwo, line 212: should we create an issue and reference it here in the comment?

tests/conftest.py
alwx: it will make it much harder to use these decorators with `pytest.mark.parametrize`. For example, this is what platform-specific decorators allow us to do:
```
@pytest.mark.parametrize(
    "_config",
    [
        {"pipeline": [{"name": "WhitespaceTokenizer"}, {"name": "SpacyFeaturizer"}]},
        pytest.param(
            {
                "pipeline": [
                    {"name": "WhitespaceTokenizer"},
                    {"name": "ConveRTFeaturizer"},
                ]
            },
            marks=[pytest.mark.linux, pytest.mark.darwin],
        ),
        pytest.param(
            {
                "pipeline": [
                    {"name": "ConveRTTokenizer"},
                    {"name": "LanguageModelFeaturizer"},
                ]
            },
            marks=[pytest.mark.linux, pytest.mark.darwin],
        ),
    ],
)
```
There is no way to do the same thing if you mark tests with `skip_on_windows`. It can be something like `unix-only` but I personally find it less clear than just specifying platforms directly.

tests/core/test_data.py
alwx: Exactly. Because it doesn't return string anymore, but `Path` instead.

tests/core/test_ensemble.py
alwx: It's required. For some reason the timestamps on Windows are the same thus the order in this test cannot be guaranteed.

tests/test_server.py
alwx: done!

tests/conftest.py
ricwo: but we're only using `marks=` for two tests right?
1. `test_missing_required_component()`: both versions of this parametrized test are marked `[pytest.mark.linux, pytest.mark.darwin]`, so we could just mark the whole test as `mark.skip_on_windows`
2. `test_train_persist_load_with_different_settings()` - one part of this test runs on windows, the other one doesn't. is it possible that the `skip_on_windows` marker would still work even inside the `marks=` kwarg? if not, we could still just break up this test into two, and not parametrize (one for linux, one for windows)

PR #6251, opened 2020-07-21
Fix typing in Message class

m-vdb:
**Proposed changes**:
- Fix typing in `Message` class. It made my builds fail here: https://github.com/RasaHQ/rasa/pull/6211/checks?check_run_id=890593396

### DIFF
--- a/rasa/nlu/training_data/message.py
+++ b/rasa/nlu/training_data/message.py
@@ -4,6 +4,7 @@ import numpy as np
 import scipy.sparse
 import typing
 
+from rasa.exceptions import RasaException
 from rasa.nlu.constants import (
     ENTITIES,
     INTENT,
@@ -126,14 +127,17 @@ class Message:
     @staticmethod
     def separate_intent_response_key(
         original_intent: Text,
-    ) -> Tuple[Optional[Text], Optional[Text]]:
+    ) -> Tuple[Text, Optional[Text]]:
 
         split_title = original_intent.split(RESPONSE_IDENTIFIER_DELIMITER)
         if len(split_title) == 2:
             return split_title[0], split_title[1]
         elif len(split_title) == 1:
             return split_title[0], None
-        return None, None
+
+        raise RasaException(
+            f"Intent name is invalid, it should not contain '{RESPONSE_IDENTIFIER_DELIMITER}'."
+        )
 
     def get_sparse_features(
         self, attribute: Text, featurizers: Optional[List[Text]] = None
--- a/rasa/nlu/training_data/message.py
+++ b/rasa/nlu/training_data/message.py
@@ -96,7 +96,11 @@ class Message:
 
     @classmethod
     def build(
-        cls, text: Text, intent: Optional[Text] = None, entities=None, **kwargs
+        cls,
+        text: Text,
+        intent: Optional[Text] = None,
+        entities: List[Dict[Text, Any]] = None,
+        **kwargs,
     ) -> "Message":
         data = {}
         if intent:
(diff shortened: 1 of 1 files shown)


rasa/nlu/training_data/message.py
```diff
@@ -95,11 +95,14 @@ def __hash__(self) -> int:
         return hash((self.text, str(ordered(self.data))))
 
     @classmethod
-    def build(cls, text, intent=None, entities=None, **kwargs) -> "Message":
+    def build(
+        cls, text: Text, intent: Optional[Text] = None, entities=None, **kwargs
```
dakshvar22, line 99: Should we add the type for `entities` as well?

rasa/nlu/training_data/message.py
```diff
@@ -117,13 +120,16 @@ def get_combined_intent_response_key(self) -> Text:
...
+    def separate_intent_response_key(
+        original_intent: Text,
+    ) -> Tuple[Optional[Text], Optional[Text]]:
 
         split_title = original_intent.split(RESPONSE_IDENTIFIER_DELIMITER)
         if len(split_title) == 2:
             return split_title[0], split_title[1]
         elif len(split_title) == 1:
             return split_title[0], None
+        return None, None
```
dakshvar22, line 132: We should actually raise an exception here. This will only happen if the intent name has more than one `/`(`RESPONSE_IDENTIFIER_DELIMITER`) which is not a valid intent name.

m-vdb: happy to raise here. What type of exception should I raise?

rasa/nlu/training_data/message.py
m-vdb: yes sure, tried to look it up but couldn't. Is this a list of strings? or something else?

rasa/nlu/training_data/message.py
dakshvar22: Maybe a ValueError, right? bc the value of `original_intent` is invalid. In the description we can explain the above mentioned reason.

rasa/nlu/training_data/message.py
dakshvar22: Not completely sure. @tabergma can you help here?

rasa/nlu/training_data/message.py
m-vdb: ah yeah ok, I thought we had custom exception classes. Will use the right standard one then

dakshvar22: Ohh right, `RasaException` in `rasa/exceptions.py`. None of the other custom exception descriptions fit the case here.

rasa/nlu/training_data/message.py
tabergma: entities should be a `List[Dict[Text, Any]]`

rasa/nlu/training_data/message.py
```diff
@@ -117,14 +125,20 @@ def get_combined_intent_response_key(self) -> Text:
...
+    ) -> Tuple[Text, Optional[Text]]:
 
         split_title = original_intent.split(RESPONSE_IDENTIFIER_DELIMITER)
         if len(split_title) == 2:
             return split_title[0], split_title[1]
         elif len(split_title) == 1:
             return split_title[0], None
 
+        raise RasaException(
+            f"Intent name is invalid, it should not contain '{RESPONSE_IDENTIFIER_DELIMITER}'."
```
dakshvar22, line 139:
```suggestion
            f"Intent name is invalid, it cannot contain more than one '{RESPONSE_IDENTIFIER_DELIMITER}'."
```

PR #6256, opened 2020-07-21
Added a schema and a writer for the 2.0 NLU Parser

### LINKED ISSUE #6252: Add schema and writer for 2.0 NLU data
Part of #6129 

- Add YAML schema for the 2.0 YAML NLU data
- Add a NLU writer (will be used to convert MD to the 2.0 YAML)

degiz:
Closes #6252

**Proposed changes**:
- Added a schema and a writer for the 2.0 NLU format
- Extracted common parts of writers for both the MD and YAML
- Fixed a bug with synonyms not being extracted from the entities

### DIFF
--- a/rasa/core/training/interactive.py
+++ b/rasa/core/training/interactive.py
@@ -6,14 +6,15 @@ import textwrap
 import uuid
 from functools import partial
 from multiprocessing import Process
+from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Text, Tuple, Union, Set
 
 import numpy as np
 from aiohttp import ClientError
 from colorclass import Color
 
+from rasa.nlu.training_data.loading import MARKDOWN, RASA, RASA_YAML
 from rasa.nlu.constants import INTENT_NAME_KEY
-from rasa.nlu.training_data.loading import MARKDOWN, RASA
 from sanic import Sanic, response
 from sanic.exceptions import NotFound
 from terminaltables import AsciiTable, SingleTable
@@ -831,7 +832,9 @@ def _write_nlu_to_file(export_nlu_path: Text, events: List[Dict[Text, Any]]) -> 
     # need to guess the format of the file before opening it to avoid a read
     # in a write
     nlu_format = _get_nlu_target_format(export_nlu_path)
-    if nlu_format == MARKDOWN:
+    if nlu_format == RASA_YAML:
+        stringified_training_data = nlu_data.nlu_as_yaml()
+    elif nlu_format == MARKDOWN:
         stringified_training_data = nlu_data.nlu_as_markdown()
     else:
         stringified_training_data = nlu_data.nlu_as_json()
@@ -840,13 +843,21 @@ def _write_nlu_to_file(export_nlu_path: Text, events: List[Dict[Text, Any]]) -> 
 
 
 def _get_nlu_target_format(export_path: Text) -> Text:
+    from rasa.data import (
+        YAML_FILE_EXTENSIONS,
+        MARKDOWN_FILE_EXTENSION,
+        JSON_FILE_EXTENSION,
+    )
+
     guessed_format = loading.guess_format(export_path)
 
-    if guessed_format not in {MARKDOWN, RASA}:
-        if export_path.endswith(".json"):
+    if guessed_format not in {MARKDOWN, RASA, RASA_YAML}:
+        if export_path.endswith(JSON_FILE_EXTENSION):
             guessed_format = RASA
-        else:
+        elif export_path.endswith(MARKDOWN_FILE_EXTENSION):
             guessed_format = MARKDOWN
+        elif Path(export_path).suffix in YAML_FILE_EXTENSIONS:
+            guessed_format = RASA_YAML
 
     return guessed_format
 
@@ -1110,7 +1121,7 @@ async def _validate_action(
 
 def _as_md_message(parse_data: Dict[Text, Any]) -> Text:
     """Display the parse data of a message in markdown format."""
-    from rasa.nlu.training_data.formats import MarkdownWriter
+    from rasa.nlu.training_data.formats.readerwriter import TrainingDataWriter
 
     if parse_data.get("text", "").startswith(INTENT_MESSAGE_PREFIX):
         return parse_data["text"]
@@ -1118,7 +1129,7 @@ def _as_md_message(parse_data: Dict[Text, Any]) -> Text:
     if not parse_data.get("entities"):
         parse_data["entities"] = []
 
-    return MarkdownWriter.generate_message_md(parse_data)
+    return TrainingDataWriter.generate_message(parse_data)
 
 
 def _validate_user_regex(latest_message: Dict[Text, Any], intents: List[Text]) -> bool:
--- a/rasa/nlu/training_data/formats/rasa_yaml.py
+++ b/rasa/nlu/training_data/formats/rasa_yaml.py
@@ -1,13 +1,31 @@
 import logging
+from collections import OrderedDict
 from pathlib import Path
-from typing import Text, Any, List, Dict, Tuple, TYPE_CHECKING, Union, Iterator
+from typing import (
+    Text,
+    Any,
+    List,
+    Dict,
+    Tuple,
+    TYPE_CHECKING,
+    Union,
+    Iterator,
+    Optional,
+)
 
-from ruamel.yaml import YAMLError
+from rasa.utils.validation import validate_yaml_schema, InvalidYamlFileError
+from ruamel.yaml import YAMLError, StringIO
 
 import rasa.utils.io as io_utils
-from rasa.constants import DOCS_URL_TRAINING_DATA_NLU
+from rasa.constants import (
+    DOCS_URL_TRAINING_DATA_NLU,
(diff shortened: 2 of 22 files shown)


degiz:
@federicotdn tagging you just in case, but the PR is not ready yet 😬

degiz:
@tmbo @federicotdn asking review from either one of you, both are not necessary 😄

rasa/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -142,7 +170,7 @@ def _parse_training_examples(
         if isinstance(examples, list):
             example_strings = [
                 # pytype: disable=attribute-error
-                example.get(KEY_INTENT_TEXT, "")
+                example.get(KEY_INTENT_TEXT, "").strip("\n\r ")
```
tmbo, line 173: why do we need the strip?

rasa/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -29,6 +47,8 @@
 
 MULTILINE_TRAINING_EXAMPLE_LEADING_SYMBOL = "-"
 
+NLU_SCHEMA_FILE = "nlu/schemas/nlu.yml"
```
tmbo, line 50: would it make sense in the future to have a "rasa" schema file combining core, nlu, responses and domain (given that you can combine their contents in files)?

rasa/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -322,3 +350,108 @@ def is_yaml_nlu_file(filename: Text) -> bool:
...
         return False
+
+
+class RasaYAMLWriter(TrainingDataWriter):
+    """Writes training data into a file in a YAML format."""
+
+    def dumps(self, training_data: "TrainingData") -> Text:
+        """Turns TrainingData into a string."""
+        stream = StringIO()
+        self.dump(stream, training_data)
```
tmbo, line 361: mhm `stream` isn't a filename (and not a `Text`) as stated in the below signature

rasa/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -322,3 +350,108 @@ def is_yaml_nlu_file(filename: Text) -> bool:
...
+
+        result = OrderedDict()
+        result[KEY_TRAINING_DATA_FORMAT_VERSION] = DoubleQuotedScalarString(
+            LATEST_TRAINING_DATA_FORMAT_VERSION
+        )
+        result[KEY_NLU] = nlu_items
+
+        io_utils.write_yaml(result, filename, True)
+
+    @staticmethod
```
tmbo, line 388: should this rather be a classmethod, e.g. we could call`cls.process_training_examples_by_key`

rasa/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -322,3 +350,108 @@ def is_yaml_nlu_file(filename: Text) -> bool:
...
+        key_examples: Text,
+        example_extraction_predicate=lambda x: x,
+    ) -> List[OrderedDict]:
+        from ruamel.yaml.scalarstring import LiteralScalarString
+
+        result = []
+        for entity_key, examples in training_examples.items():
+
+            converted_examples = []
+            converted_examples += [
```
tmbo, line 445:
```suggestion
            converted_examples = [
```

rasa/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -322,3 +350,108 @@ def is_yaml_nlu_file(filename: Text) -> bool:
...
+                TrainingDataWriter.generate_list_item(
+                    example_extraction_predicate(example).strip("\n\r ")
```
tmbo, line 447: what if an example contains new lines?

rasa/core/training/interactive.py
```diff
@@ -840,13 +843,21 @@ def _write_nlu_to_file(export_nlu_path: Text, events: List[Dict[Text, Any]]) ->
...
+        JSON_FILE_EXTENSION,
+    )
+
     guessed_format = loading.guess_format(export_path)
 
-    if guessed_format not in {MARKDOWN, RASA}:
-        if export_path.endswith(".json"):
+    if guessed_format not in {MARKDOWN, RASA, RASA_YAML}:
+        if export_path.endswith(JSON_FILE_EXTENSION):
             guessed_format = RASA
```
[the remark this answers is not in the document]
degiz: I don't think so, that's excplicitly the "extension" of the file as  opposed to the "format". The same way we have `YAML_FILE_EXTENSIONS`

rasa/nlu/training_data/formats/rasa_yaml.py
degiz: In testing phase I faced several problems with leading/trailing spaces and `\n` in the training examples, and found this as a safe solution to always take only "text" from the training examples/name of the intents.
At least I haven't seen the trailing white space as a vital part of a training data.

rasa/nlu/training_data/formats/rasa_yaml.py
degiz: Yes, I'm also thinking about that.. Will definitely do in the future.

rasa/nlu/training_data/formats/rasa_yaml.py
degiz: True! Fixed that.

rasa/nlu/training_data/formats/rasa_yaml.py
degiz: makes sense.

rasa/nlu/training_data/formats/rasa_yaml.py
degiz: That's a good question! The current implementation of reader/writer is done with an assumption that every line should start with `-` and be a separate training example, e.g. in the reader we're using the `splitlines()` all over.
I guess this has to be noted in the docs and possibly be changed before the release.

rasa/nlu/training_data/formats/readerwriter.py
```diff
@@ -33,6 +46,87 @@ def dumps(self, training_data: "TrainingData") -> Text:
...
+
+        entity_dict = {
+            ENTITY_ATTRIBUTE_TYPE: entity_type,
+            ENTITY_ATTRIBUTE_ROLE: entity_role,
+            ENTITY_ATTRIBUTE_GROUP: entity_group,
+            ENTITY_ATTRIBUTE_VALUE: entity_value,
+        }
+        entity_dict = {k: v for k, v in entity_dict.items() if v is not None}
+
+        return f"[{entity_text}]{json.dumps(entity_dict)}"
```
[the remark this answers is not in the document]
degiz: that's a very good point, thanks!

rasa/nlu/training_data/formats/rasa_yaml.py
tmbo: that sounds good, I think in the future we can make the splitting more advanced, and e.g. only treat lines starting with `-` as a new example. until then we might be able to escape the newlines? (using a literal `\n`)

rasa/nlu/training_data/formats/rasa_yaml.py
tmbo: makes sense :+1:

rasa/nlu/training_data/formats/rasa_yaml.py
degiz: Yes, the `\n` works for now!

rasa/utils/io.py
```diff
@@ -12,18 +12,24 @@
...
+from ruamel.yaml import RoundTripRepresenter
 
 from rasa.constants import ENV_LOG_LEVEL, DEFAULT_LOG_LEVEL, YAML_VERSION
 
 if TYPE_CHECKING:
     from prompt_toolkit.validation import Validator
 
 DEFAULT_ENCODING = "utf-8"
+ESCAPE_DCT = {"\b": "\\b", "\f": "\\f", "\n": "\\n", "\r": "\\r", "\t": "\\t"}
+ESCAPE = re.compile(r"[\b\f\n\r\t]")
```
tmbo, line 29: can we use the dict in L28 to create the regex?

tests/nlu/training_data/formats/test_rasa_yaml.py
```diff
@@ -1,43 +1,109 @@
...
-    f"   metadata:\n"
-    f"   examples:\n"
-    f"    - text: |\n"
-    f"        how much CO2 will that use?\n"
-    f"      metadata:\n"
-    f"        sentiment: positive\n"
-    f"    - text: |\n"
-    f'        how much carbon will a one way flight from [new york]{{"entity": "city", "role": "from"}} to california produce?'
+MINIMAL_VALID_EXAMPLE = (
+    f"version: '{LATEST_TRAINING_DATA_FORMAT_VERSION}'\n" f"nlu:\n" f"stories:"
```
tmbo, line 58:
```suggestion
    f"version: '{LATEST_TRAINING_DATA_FORMAT_VERSION}'\n nlu:\n stories:"
```

PR #6258, opened 2020-07-22
New docs include literals

m-vdb:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/6199

**Doc files to rework**
- [x] Forms
- [x] Reminders and External Events
- [x] Events
- [x] Custom Connectors (move `RestInput` in its own module)


m-vdb:
@erohmensing here is my PR that makes it possible to include source files into the docs. I added a checklist in the PR description to reference the documents that need update. On my end, I think there's nothing else to be done. Would you mind taking this over and updating the `.mdx` files? Happy to answer your questions if any 😄

rasa/core/channels/channel.py
```diff
@@ -356,147 +360,3 @@ async def send_custom_json(
...
-        await self.messages.put(message)  # pytype: disable=bad-return-type
-
-
-class RestInput(InputChannel):
-    """A custom http input channel.
-
-    This implementation is the basis for a custom implementation of a chat
-    frontend. You can customize this to send messages to Rasa Core and
-    retrieve responses from the agent."""
-
```
erohmensing, line 388: Should we make a deprecation warning'd class here that imports the other one and tells the user to change their imports? We had done that with many modules we'd moved before, so that custom connectors don't break on upgrade. 

If not, we should add the change to the migration guide

rasa/core/channels/rest.py
```diff
@@ -0,0 +1,163 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class RestInput(InputChannel):
+    """A custom http input channel.
+
+    This implementation is the basis for a custom implementation of a chat
+    frontend. You can customize this to send messages to Rasa Core and
+    retrieve responses from the agent."""
```
erohmensing, line 27:
```suggestion
    frontend. You can customize this to send messages to Rasa and
    retrieve responses from the assistant."""
```

rasa/core/channels/rest.py
```diff
@@ -0,0 +1,163 @@
...
+                except CancelledError:
+                    logger.error(
+                        "Message handling timed out for "
+                        "user message '{}'.".format(text)
+                    )
+                except Exception:
+                    logger.exception(
+                        "An exception occured while handling "
+                        "user message '{}'.".format(text)
+                    )
```
erohmensing, line 139:
```suggestion
                except CancelledError:
                    logger.error(
                        f"Message handling timed out for "
                        f"user message '{text}'."
                    )
                except Exception:
                    logger.exception(
                        f"An exception occured while handling "
                        f"user message '{text}'."
                    )
```

tests/core/test_channels.py
```diff
@@ -869,7 +869,7 @@ async def test_slackbot_send_text():
 
 @pytest.mark.filterwarnings("ignore:unclosed.*:ResourceWarning")
 def test_channel_inheritance():
-    from rasa.core.channels.channel import RestInput
+    from rasa.core.channels import RestInput
     from rasa.core.channels.rasa_chat import RasaChatInput
```
erohmensing, line 873: why do we the import from `channels` instead of `channels.rest`? Maybe this should be consistent among the way we import channels in the tests

m-vdb:
@erohmensing can I fix the merge conflicts? it seems you are done 🤗

erohmensing:
I haven't pushed the mdx changes yet... but its hard for me to know what should go in because the diffs are strange, maybe due to merge conflicts. Maybe you want to fix those and then i'll run it?

m-vdb:
oups I thought you would modify the mdx files directly. Because you will have to resolve all the conflicts :/ It might be easier to simply update manually the mdx with your changes :/

erohmensing:
alright go for it 💪

newdocs/docs/core/forms.mdx
```diff
@@ -417,7 +343,7 @@ This is a great starting point, but
 you should give your bot to people to test **as soon as possible**. One of the guiding principles
 behind Rasa Core is:
 
-> Learning from real conversations is more important than designing hypothetical ones
+Learning from real conversations is more important than designing hypothetical ones
```
erohmensing, line 346:
```suggestion
> Learning from real conversations is more important than designing hypothetical ones
```

rasa/core/channels/channel.py
m-vdb: Hum, since we're launching 2.0, I don't know if it really make sense to introduce it. And the class _was_ and _is still_ importable `from rasa.core.channels import ...`

tests/core/test_channels.py
m-vdb: the way I see this public API is that if you want to use a channel, you should import from the `rasa.core.channels` modules. You do not really need to know how the `channels` module is organised when you're consuming it.

rasa/core/channels/channel.py
erohmensing: you make a good point!

PR #6235, opened 2020-07-17
Port formbot to rules/yaml format

akelad:
**Proposed changes**:
- ...

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6263, opened 2020-07-22
add capability to rule policy to memorize negative rules from rule snippets and stories

Ghostvv:
**Proposed changes**:
- refactor rules and memo to accomodate learning negative rules from stories
- contains updates needed for https://github.com/RasaHQ/rasa/pull/6235


Ghostvv:
@wochinge I need to add tests, but the code can be reviewed

rasa/core/policies/rule_policy.py
```diff
@@ -47,13 +52,14 @@ def supported_data() -> SupportedData:
...
         """
-        return SupportedData.RULE_DATA
+        return SupportedData.ML_AND_RULE_DATA
 
     def __init__(
         self,
         featurizer: Optional[TrackerFeaturizer] = None,
         priority: int = FORM_POLICY_PRIORITY,
         lookup: Optional[Dict] = None,
+        negative_lookup: Optional[Dict] = None,
```
wochinge, line 62: please add a docstring which describes the old and new parameters

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +80,154 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
+        indicator = PREV_PREFIX + RULE_SNIPPET_ACTION_NAME
+        if indicator not in action_before_listen.keys():
+            return [{indicator: 1}, action_before_listen, states[-1]]
+        return [action_before_listen, states[-1]]
+
+    @staticmethod
+    def _clean_feature_keys(lookup: Optional[Dict]) -> Optional[Dict]:
+        # remove action_listens that were added after conditions
+        updated_lookup = lookup.copy()
+        for key in lookup.keys():
```
wochinge, line 131: When you also get the `predicted_action` right away here you don't need to retrieve it again in `lookup[key]` (line 131). 
```suggestion
        for states, predicted_action in lookup.items():
```

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +80,154 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
+            # Delete rules if there is no prior action or if it would predict
+            # the `...` action
+            if PREV_PREFIX not in key or lookup[key] == RULE_SNIPPET_ACTION_NAME:
+                del updated_lookup[key]
+            elif RULE_SNIPPET_ACTION_NAME in key:
+                # If the previous action is `...` -> remove any specific state
```
wochinge, line 137: "specific state" refers just refers to the previous action, right?

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +80,154 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
+
+                del updated_lookup[key]
+
+        return updated_lookup
+
+    def _create_negative_lookup_from_states(
+        self,
+        trackers_as_states: List[List[Dict]],
+        trackers_as_actions: List[List[Text]],
+    ) -> Dict[Text, Text]:
```
wochinge, line 156: can you please add a docstring which describes what this function is used for?

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +80,154 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
+        negative_lookup = {}
+        for states, actions in zip(trackers_as_states, trackers_as_actions):
+            active_form = self._get_active_form_name(states[-1])
+            # even if there are two identical feature keys
+            # their form will be the same
+            # because of `active_form_...` feature
+            if active_form:
+                # leave only last 2 dialogue turns
```
wochinge: why only the last 2? Let's add the explanation as comment.

rasa/core/policies/rule_policy.py
```diff
@@ -123,83 +277,25 @@ def _rule_is_good(
...
-    def predict_action_probabilities(
-        self,
-        tracker: DialogueStateTracker,
-        domain: Domain,
-        interpreter: NaturalLanguageInterpreter = RegexInterpreter(),
-        **kwargs: Any,
-    ) -> List[float]:
-        """Predicts the next action the bot should take after seeing the tracker.
+    @staticmethod
+    def _predict_default_action(tracker: DialogueStateTracker) -> Optional[Text]:
```
wochinge, line 283: optional: the name of the function sounds like it returns a bool
```suggestion
    def _applicable_default_action(tracker: DialogueStateTracker) -> Optional[Text]:
```

rasa/core/policies/rule_policy.py
```diff
@@ -214,85 +310,125 @@ def predict_action_probabilities(
...
+        return recalled
 
+    def predict_action_probabilities(
+        self,
+        tracker: DialogueStateTracker,
+        domain: Domain,
+        interpreter: NaturalLanguageInterpreter = RegexInterpreter(),
+        **kwargs: Any,
+    ) -> List[float]:
+        """Predicts the next action the bot should take after seeing the tracker.
```
wochinge, line 404: imo you don't need the docstring as interface `Policy` has the docstring and that's enough. If you decide to keep it then it should follow the google format for docstrings.

rasa/core/policies/rule_policy.py
```diff
@@ -214,85 +310,125 @@ def predict_action_probabilities(
...
-    ):
-        return None
+        Returns the list of probabilities for the next actions.
+        If memorized action was found returns 1 for its index,
+        else returns 0 for all actions.
+        """
+        result = self._default_predictions(domain)
 
-    return DEFAULT_ACTION_MAPPINGS.get(tracker.latest_message.intent.get("name"))
+        if not self.is_enabled:
```
wochinge, line 412: can we throw out `toggle`  and `is_enabled`?

rasa/core/policies/rule_policy.py
```diff
@@ -214,85 +310,125 @@ def predict_action_probabilities(
...
+                tracker.update(FormValidation(False))
+
+        if recalled is not None:
+            logger.debug(f"There is a rule for the next action '{recalled}'.")
+        else:
+            logger.debug("There is no applicable rule.")
+
+        return recalled
 
+    def predict_action_probabilities(
```
wochinge, line 397: This function looks beautiful now! Great job!

rasa/core/policies/rule_policy.py
```diff
@@ -214,85 +310,125 @@ def predict_action_probabilities(
...
+        if predicted_default_action_name:
+            return self._prediction_result(
+                predicted_default_action_name, tracker, domain
+            )
+
+        # A form has priority over any other rule.
+        # The rules or any other prediction will be applied only if a form was rejected.
+        # If we are in a form, and the form didn't run previously or rejected, we can
+        # simply force predict the form.
+        predicted_form_action_name = self._predict_form_action(tracker)
```
wochinge, line 428: it's a bit counter intuitive that this can also be `action_listen`

wochinge: should we rename to something like `_action_from_form_happy_path`?

rasa/core/policies/rule_policy.py
```diff
@@ -214,85 +310,125 @@ def predict_action_probabilities(
...
+                    # negative rules don't contain a key that corresponds to
+                    # the fact that active_form shouldn't be predicted
+                    logger.debug(
+                        f"Predicted form '{active_form_name}' by overwriting "
+                        f"'{ACTION_LISTEN_NAME}' predicted by general rule."
+                    )
+                    return active_form_name
+
+                # do not predict anything
+                recalled = None
```
wochinge, line 380: should this be this?
```suggestion
                recalled = self._default_predictions(domain)
```

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +80,154 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
+            # because of `active_form_...` feature
+            if active_form:
+                # leave only last 2 dialogue turns
+                states = self._modified_states(states[-2:])
+                feature_key = self._create_feature_key(states)
+
+                if (
+                    # form is predicted after action_listen in unhappy path,
+                    # therefore no validation is needed
+                    self._prev_action_listen_in_state(states[-1])
```
wochinge, line 170: That's only the case when the form rejected because `tracker.applied_events()` filters out the happy paths, correcT?

rasa/core/policies/rule_policy.py
```diff
@@ -214,85 +310,125 @@ def predict_action_probabilities(
...
+    def _predict_rule_action(
+        self, tracker: DialogueStateTracker, domain: Domain
+    ) -> Optional[Text]:
         tracker_as_states = self.featurizer.prediction_states([tracker], domain)
         states = tracker_as_states[0]
 
         logger.debug(f"Current tracker state: {states}")
 
+        possible_keys = set(self.lookup[RULES].keys())
+        negative_keys = set(self.lookup[NEGATIVE_RULES].keys())
```
wochinge, line 335: should we only collect the negative keys if a form is active? Might be a good idea performance wise.

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +80,154 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
 
+    @staticmethod
+    def _get_active_form_name(state: Dict[Text, float]) -> Optional[Text]:
+        found_forms = [
+            state_name[len(ACTIVE_FORM_PREFIX) :]
+            for state_name, prob in state.items()
+            if ACTIVE_FORM_PREFIX in state_name
+            and state_name != ACTIVE_FORM_PREFIX + "None"
+            and prob > 0
+        ]
```
wochinge, line 91: let's use a generator as we don't have to process every state once we have found a form
```suggestion
                found_forms = next(
            (
                state_name[len(ACTIVE_FORM_PREFIX) :]
                for state_name, prob in state.items()
                if ACTIVE_FORM_PREFIX in state_name
                and state_name != ACTIVE_FORM_PREFIX + "None"
                and prob > 0
            ),
            None,
        )
```

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +80,154 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
+        return any(
+            PREV_PREFIX + ACTION_LISTEN_NAME in state_name and prob > 0
+            for state_name, prob in state.items()
+        )
+
+    @staticmethod
+    def _modified_states(
+        states: List[Dict[Text, float]]
+    ) -> List[Optional[Dict[Text, float]]]:
+        """Modify the states to
```
wochinge, line 106: Please add the docstring in the googledoc format which we use elsewhere as well

rasa/core/policies/rule_policy.py
```diff
@@ -34,6 +34,11 @@
     USER_INTENT_SESSION_START: ACTION_SESSION_START_NAME,
 }
 
+RULES = "rules"
+NEGATIVE_RULES = "negative_rules"
+NO_VALIDATION = "no_validation"
```
wochinge, line 39: to make it clearer:
```suggestion
FORM_EXECUTION_AFTER_REJECTION = "no_validation"
```

rasa/core/policies/rule_policy.py
```diff
@@ -34,6 +34,11 @@
     USER_INTENT_SESSION_START: ACTION_SESSION_START_NAME,
 }
 
+RULES = "rules"
+NEGATIVE_RULES = "negative_rules"
```
wochinge, line 38: negative rules seems unintuitive to me, how about
```suggestion
FORM_UNHAPPY_RULES = "negative_rules"
```

rasa/core/policies/rule_policy.py
```diff
@@ -34,6 +34,11 @@
...
+NO_VALIDATION = "no_validation"
+NO_ACTIVE_FORM = "no_active_form"
```
wochinge, line 40: The name is somewhat confusing, how about
```suggestion
DO_NOT_PREDICT_FORM_ACTION = "no_active_form"
```

rasa/core/policies/rule_policy.py
Ghostvv: no recalled is the name of the action

rasa/core/policies/rule_policy.py
Ghostvv: I think we can, they are here for very long time and I don't know where they are used 🙈

rasa/core/policies/rule_policy.py
wochinge: oh dear - I was a bit confused by jumping back and forth in the code

rasa/core/policies/rule_policy.py
Ghostvv: no it refers to whole dialogue turn represented as state

rasa/core/policies/rule_policy.py
Ghostvv: this is during training so all the actions inside form are unhappy paths

rasa/core/policies/rule_policy.py
Ghostvv: hm, I don't see how `FORM_EXECUTION_AFTER_REJECTION` is related to no validation

Ghostvv: renamed to `DO_NOT_VALIDATE_FORM`

rasa/core/policies/rule_policy.py
Ghostvv: forgot to remove `negative_lookup`
added docstring to memoization policy, since the params are the same

Ghostvv:
Some parts of RulePolicy is still not refactored because they depend on e2e states changes

rasa/core/policies/rule_policy.py
```diff
@@ -74,6 +79,183 @@ def _create_feature_key(self, states: List[Dict]) -> Text:
...
+                del updated_lookup[feature_key]
+            elif RULE_SNIPPET_ACTION_NAME in feature_key:
+                # If the previous action is `...` -> remove any specific state
+                # requirements for that state (anything can match this state)
+                new_feature_key = re.sub(
+                    rf".*{PREV_PREFIX}\.\.\.[^|]*", "", feature_key
+                )
+
+                if new_feature_key:
+                    if new_feature_key.startswith("|"):
```
tabergma, line 152: What does `|` mean again? Does it make sense to replace it with a constant?

rasa/core/policies/rule_policy.py
```diff
@@ -214,85 +353,108 @@ def predict_action_probabilities(
...
 
-        for i, state in enumerate(reversed(states)):
-            possible_keys = set(
-                filter(lambda _key: self._rule_is_good(_key, i, state), possible_keys)
-            )
-
-        if possible_keys:
-            # TODO rethink that
-            key = max(possible_keys, key=len)
+        rule_keys = self._get_possible_kes(self.lookup[RULES], states)
```
tabergma, line 375: type: `_get_possible_keys`

rasa/core/policies/rule_policy.py
Ghostvv: not yet, this will be the part of next refactor after e2e

Ghostvv:
@tabergma is it good to go, giving the fact that the way rules are stored will be refactored later?

tabergma:
@Ghostvv Code wise it looks good. As I am not 100% into the topic it is hard to judge whether functionality is as expected. But I guess Tobias already took a look at that. Do you want me to approve?

Ghostvv:
I added tests to check functionality, I think they cover the problems, it supposed to solve.
@akelad could you please test that it does what we discussed and approve if it is

PR #6272, opened 2020-07-24
Use output_channel of externally triggered intent as input_channel

pheel:
**Proposed changes**:
- Currently any intent pushed to the tracker via `/conversations/<conversation_id>/trigger_intent` results in a user event with input channel `None`. This means `latest` output channel is effectively lost until a "real" user message comes in from a channel. This sets input channel using the output channel determined from the request.


sara-tagger:

federicotdn:
It's a bit strange to be using an output channel to populate the latest input channel field of the tracker, given that they are not the same thing. Maybe we could update the tracker in a way where the latest input channel property is not lost? Maye @JEM-Mosig has an opinion on this.

pheel:
@federicotdn, @b-quachtran, I could make `rasa.core.trackers.DialogueStateTracker.get_latest_input_channel` skip over messages with null output channel, and it would have the same effect of `latest` not losing its meaning. Sounds like a less invasive change. Should I go for it?

pheel:
Ah, I see this was cloned at https://github.com/RasaHQ/rasa/pull/6472.

federicotdn:
@pheel 
> I could make rasa.core.trackers.DialogueStateTracker.get_latest_input_channel skip over messages with null output channel, and it would have the same effect of latest not losing its meaning. Sounds like a less invasive change. Should I go for it?

I think that makes a lot of sense, given the name of the method `get_latest_input_channel`. The thing is, that with that change we would be effectively modifying the value the method returns in different contexts (on trackers with trailing UserUttered events with no input channel). The PR I opened was a test to see if it made sense to instead make the change more specific to the externally triggered intents feature (note that I did not change `get_latest_input_channel`). Apart from this, I still think we should not mix input and output channels.

pheel:
Yeah, I'll close this PR in favor of https://github.com/RasaHQ/rasa/pull/6472 then.

PR #6276, opened 2020-07-24
API improvements

pheel:
**Proposed changes**:
- When Rasa is run with `--enable-api`, exceptions incurred as a result of a request are redirected to the HTTP response. This is nice, but it's useful to also have them show up in ones favorite stream handler via the Python logger.
- When attempting to load an outdated model, Rasa terminates with an `UnsupportedModelError` or `UnsupportedDialogueModelError`. This is problematic for users that have access to the `/train` route, but not the file system from which models are retrieved. This allows Rasa to boot so that such users can retrain.


sara-tagger:

changelog/6276.improvement.rst
```diff
@@ -0,0 +1 @@
+Allow Rasa to boot even when UnsupportedDialogueModelError or UnsupportedModelError occurs. Forward HTTP Error responses to standard log output.
```
wochinge, line 1: We recently changed our docs system 🙈  The changelog entries now have to be in `md` format.

wochinge, line 1:
```suggestion
Allow Rasa to boot even when `UnsupportedDialogueModelError` or `UnsupportedModelError` occurs. Forward HTTP Error responses to standard log output.
```

rasa/core/interpreter.py
```diff
@@ -37,7 +37,13 @@ def create(
         if isinstance(obj, NaturalLanguageInterpreter):
             return obj
         elif isinstance(obj, str) and os.path.exists(obj):
-            return RasaNLUInterpreter(model_directory=obj)
+            from rasa.nlu.model import UnsupportedModelError
```
wochinge, line 40: What do you think of adding the exception handler [here](https://github.com/RasaHQ/rasa/blob/cd1500ec5c1cdb9b1dc87a7e91edb408966c296e/rasa/core/run.py#L249) ? That way the change is really isolated to the startup and we handle the error as one, and not separate for Core and NLU?

Let's also add a test for it 👍

pheel: Good idea, since these model-loading fatal errors only occur at boot. I have since also gotten `zlib.error` exceptions (from corrupt tarballs). So I think a fully general exception catching block is in order? cf. https://github.com/RasaHQ/rasa/blob/d52d50625676cc679ec83dcd4e663429f7ce2d73/rasa/server.py#L315-L322 Will submit revision with test. 🚀

tests/core/test_run.py
```diff
@@ -40,3 +42,38 @@ def test_create_single_input_channels_by_class_wo_credentials():
 
     assert len(channels) == 1
     assert channels[0].name() == "rest"
+
+
+async def test_load_agent_on_start_with_good_model_file(
+    trained_rasa_model, rasa_server, loop,
```
wochinge, line 48: You'll probably have to run the black formatting on this then
```suggestion
    trained_rasa_model: Text, rasa_server: Sanic, loop: AbstractEventLoop,
```

tests/core/test_run.py
```diff
@@ -40,3 +42,38 @@ def test_create_single_input_channels_by_class_wo_credentials():
...
+    assert isinstance(agent.policy_ensemble, policies.PolicyEnsemble)
+    assert isinstance(agent.domain, domain.Domain)
+
+
+async def test_load_agent_on_start_with_bad_model_file(
+    tmpdir, rasa_server, loop,
+):
+    from pathlib import Path
+
+    fake_model = Path(tmpdir) / "fake_model.tar.gz"
```
wochinge, line 64:
```suggestion
    tmp_path: Path, rasa_server: Sanic, loop: AbstractEventLoop,
):
    fake_model = tmp_path / "fake_model.tar.gz"
```

changelog/6276.improvement.rst
```diff
@@ -0,0 +1 @@
+Allow Rasa to boot when model loading exception occurs. Forward HTTP Error responses to standard log output.
```
wochinge, line 1: can you please change the type of the file to `.md` due to our new docsystem? Thanks 🙏

pheel: My bad, missed the consequence on extension!

rasa/core/run.py
```diff
@@ -224,12 +224,11 @@ async def load_agent_on_start(
     endpoints: AvailableEndpoints,
     remote_storage: Optional[Text],
     app: Sanic,
-    loop: Text,
```
wochinge, line 227: I think we can't remove this as `Sanic` passes the loop when running this function. Pretty sure that this would lead to an error otherwise

pheel: :sweat_smile: Adding it back...

wochinge (approved):
Thanks a lot!

PR #6297, opened 2020-07-29
Initial docs for 2.0 Training data format

degiz:
**Proposed changes**:
- Added initial docs for the NLU and stories

### DIFF
--- a/docs/core/stories.rst
+++ b/docs/core/stories.rst
@@ -28,7 +28,42 @@ This is a guide to the story data format.
 Format
 ------
 
+With Rasa Open Source 2.0 release we've introduced a YAML format as unified and the most extendable way to manage the training data.
+With YAML users are free to distribute training data among any number of YAML files.
+
+Rasa Open Source YAML format file should specify the `version` key in order to be parsed correctly:
+
+- Training data files with no "version" specified will be considered to have the "latest" format version.
+- Training data files with a version greater that currently used Rasa Open Source will be skipped.
+
+
 Here's an example of a dialogue in the Rasa story format:
+
+In YAML format:
+.. code-block:: yaml
+
+    version: "2.0"
+    stories:
+    - story: greet + location/price + cuisine + num people  # name of the story - just for debugging
+      steps:
+      - intent: greet
+      - action: action_ask_howcanhelp
+      - intent: inform
+        entities:
+        - location: "rome"
+        - price: "cheap"  # user utterance
+      - action: action_on_it
+      - action: action_ask_cuisine
+      - intent: inform
+        entities:
+        - cuisine: "spanish"
+      - action: action_ask_numpeople  # action that the bot should execute
+      - intent: inform
+        entities:
+        - people: "six"
+      - action: action_ack_dosearch
+
+Or in Markdown format:
 
 .. code-block:: story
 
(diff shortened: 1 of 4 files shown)


degiz:
@akelad @tmbo tagging both of you just in case 😄

docs/core/stories.rst
```diff
@@ -28,8 +28,43 @@ This is a guide to the story data format.
 Format
 ------
 
+With Rasa Open Source 2.0 release we've introduced a YAML format as unified and the most extendable way to manage the training data.
```
tmbo, line 31: Let's try to target novice users more, maybe something like:

```
Format
-------
The training data uses the YAML format. You can split the training data among any number of YAML files (just make sure to mention all of them when you run the training).  

.. note ::

    The new training data format was introduced with Rasa Open Source 2.0. If you want to update from an older 
    training data format, you can find more information about the upgrade in our migration guide.


```
let's also:
- remove the note that is right before the `Format` and incorporate the info here
- I'd vote for not calling this `the Rasa Core dialogue system` (paragraph before format)

docs/nlu/training-data-format.rst
```diff
@@ -15,10 +15,100 @@ Data Formats
 ~~~~~~~~~~~~
 
 
-You can provide training data as Markdown or as JSON, as a single file or as a directory containing multiple files.
-Note that Markdown is usually easier to work with.
+You can provide training data as YAML, Markdown or as JSON, as a single file or as a directory containing multiple files.
 
 
+YAML Format
```
tmbo, line 21: should we really split this up? @akelad I think it might make more sense to have a single page that contains all the information about the training data format, ala:
1. general structure (it is yaml, multi file, what are the keys)
2. story training data
3. intent training data
...

docs/core/stories.rst
```diff
@@ -28,8 +28,43 @@ This is a guide to the story data format.
...
+      - intent: inform
+        entities:
+        - cuisine: "spanish"
+      - action: action_ask_numpeople  # action that the bot should execute
+      - intent: inform
+        entities:
+        - people: "six"
+      - action: action_ack_dosearch
+
+Or in Markdown format:
```
tmbo, line 66: I think the conclusion was to remove the documentation about the markdown format completely: it will still be available in the 1.10.x docs and we want to drive users to use yaml instead of having the choice to use either one of them

tmbo: I'd rather add a note ala "there is also a deprecated file format, you can find more information about that here: legacy-docs.rasa.com"

docs/core/stories.rst
akelad: +1 would call it `dialogue manager` or `dialogue management` or similar

docs/core/stories.rst
akelad: agreed, though not sure where we point them to at the moment because right now it's not at `legacy-docs.rasa.com` but on the same page just a different version

docs/nlu/training-data-format.rst
melindaloubser1: I'd vote for one page.

akelad:
@melindaloubser1 are we closing this in favour of yours?

melindaloubser1:
Yes, if that's ok

akelad:
closing in favour of #6357

PR #6298, opened 2020-07-29
Interactive code cells

### LINKED ISSUE #6169: New docs: interactive code cells
⬆️ Relates to https://github.com/RasaHQ/growth/issues/1436

With our new doc system, we'll need to support interactive code cells that we introduced in [this pull request](https://github.com/RasaHQ/rasa/pull/6095) (the prototyper).

cc @alwx

m-vdb:
**Proposed changes**:
- Fixes #6169
- Use a custom theme inspired by the standard `theme-live-codeblock`. The standard theme is not usable as is because it's not configurable.

**Checklist**
- [x] set tracking id only on production build (use a default one otherwise)
- [x] integrate `.. chatblock::` inside the doc
- [x] implement spinner state for buttons
- [x] fix display of initial screen of chat (weird font)


m-vdb:
@alwx here is my exploration for setting up live codeblocks. I've summarized the approach in the PR desc and listed the things that are left to do. LMK if you want to contribute

docs/src/utils.js
```diff
@@ -0,0 +1,8 @@
+const uuidv4 = () => ([1e7]+-1e3+-4e3+-8e3+-1e11).replace(
+    /[018]/g,
+    c => (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
+);
+
+const isProductionBuild = () => typeof window.__REACT_ERROR_OVERLAY_GLOBAL_HOOK__ === 'undefined';
```
m-vdb, line 6: not sure it's the legit way of checking this, but it seems to work \o/

docs/src/components/prototyper/prototyper.jsx
```diff
@@ -0,0 +1,155 @@
...
+const Prototyper = ({children, startPrototyperApi, trainModelApi, chatBlockSelector, chatBlockScriptUrl}) => {
+    const [trackingId, setTrackingId] = React.useState(null);
+    const [hasStarted, setHasStarted] = React.useState(false);
+    const [projectDownloadUrl, setProjectDownloadUrl] = React.useState(null);
+    const [trainingData, setTrainingData] = React.useState({});
+    const [pollingIntervalId, setPollingIntervalId] = React.useState(null);
+    const [hasTrained, setHasTrained] = React.useState(false);
+    const [isTraining, setIsTraining] = React.useState(false);
+
+    // FIXME: once we can use `rasa-ui` outside of `rasa-x`, we can remove this
```
m-vdb, line 24: Really not proud of this, but couldn't find a better way to do this. The `<script>` tags are not read in the MDX files 🤷 . It'll be cleaned up anyway when we make `rasa-ui` available outside of `rasa-x`

docs/docs/user-guide/prototype-an-assistant.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+[NLU](../glossary.mdx#rasa-nlu) data provides examples of what users might say to your assistant and what they mean by it.
+[Intent](../glossary.mdx#intent)  refers to what the user means with a specific message.
+Your assistant can only learn to recognize intents for which it has seen examples, so we need to provide some data.
+
+Add examples for your user-goal specific intents in the format shown below.
+You can delete the `subscribe` and `inform` intents if you’re not using them;
+you can also add or change examples for any of the other intents.
+
+```yaml live noResult name=nlu
```
m-vdb, line 41: using components here instead of the ` ```yaml live ...` syntax would have been more convenient to pass props, but I couldn't make it work properly.

docs/docs/user-guide/prototype-an-assistant.mdx
```diff
@@ -0,0 +1,181 @@
...
+This page explains the basics of building an assistant with Rasa Open Source and shows the structure of a Rasa project.
+You can test it out right here without installing anything. You can also install
+[install Rasa Open Source](./installation.mdx) and follow along in your command line.
+
+<Prototyper
+  startPrototyperApi="https://trainer-service.prototyping.rasa.com/startPrototyping"
+  trainModelApi="https://trainer-service.prototyping.rasa.com/trainings"
+  chatBlockSelector="#rasa-chat-block"
+  chatBlockScriptUrl={useBaseUrl("js/rasa-chatblock.min.js")}
+>
```
alwx, line 20: Love how easy it is to embed a component!

docs/src/utils.js
alwx: It seems like this statement actually checks for React WebTools to be not available. Seems a bit hacky, but we can't be use `process.env.NODE_ENV` here, right?

m-vdb: I don't think `process.env.NODE_ENV` works in the browser :/

m-vdb:
> Looks good! Error handling is missing, but it was not there in the original code as well.

The good thing is that it should be much easier to add now!

m-vdb:
@alwx @tmbo the prototyper can be tested here: https://rasa.com/docs/rasa/next/user-guide/prototype-an-assistant

PR #6323, opened 2020-08-03
Response generator responses with multimedia elements

tmbo:
**Proposed changes**:
- added yaml support for responses. example of a `responses.yml` with the format as it is currently implemented:
```yaml
responses:
  chitchat/ask_weather:
  - text: Where do you want to check the weather?
    buttons:
    - title: Current location
      payload: here
    - title: Other place
      payload: other_location

  chitchat/ask_name:
  - text: my name is Sara, Rasa's documentation bot!
    image: "https://i.imgur.com/nGF1K8f.jpg"

```
- responses model only gets trained on the `text` part of the first `response`.
- allows users to use multimedia elements (images, buttons, ...) in response templates
- the format (and the parser) is the same as we use for utter templates in the domain
- this includes the changes to port the moodbot demo to the new file format (needed an example...)

- [x] handle case where the `response.text` is empty -> use the intent name instead


tmbo:
@dakshvar22 are there any concerns regarding the data format?

rasa/core/training/story_reader/markdown_story_reader.py
```diff
@@ -233,7 +233,7 @@ def is_markdown_story_file(file_path: Text) -> bool:
         """
         suffix = PurePath(file_path).suffix
 
-        if suffix and suffix != MARKDOWN_FILE_EXTENSION:
+        if suffix not in MARKDOWN_FILE_EXTENSIONS:
```
tmbo, line 236: I think we should discuss if we keep the original behaviour: I changed it to keep it uniform across readers. The difference between the versions is that the original one would happily read files without an extension as story files, which seems arbitrary. Nevertheless, it is breaking backwards compatibility in an odd way, so kind of betting that no one relies on this odd behaviour 🤔

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1098,7 +1104,7 @@ def _load_model(
             data_signature=model_data_example.get_signature(),
             label_data=label_data,
             entity_tag_specs=entity_tag_specs,
-            config=meta,
+            config=copy.deepcopy(meta),
```
tmbo, line 1107: this fix was needed as this config gets passed to tensorflow which in place convert all values of the dictionary to tensorflow types (e.g. it will replace an ordinary python string with a tensorflow specific string). to avoid tf modifying this original dictionary we need to pass in a copy...

dakshvar22:
@tmbo  Data format looks good. Just two comments(maybe you have thought of these already):
1. In contrast to usual response templates, there cannot be multiple `text` attributes for one retrieval intent. For example, this would be invalid:
```
responses:
  chitchat/ask_weather:
  - text: Where do you want to check the weather?
    buttons:
    - title: Current location
      payload: here
    - title: Other place
      payload: other_location
 - text: Do you want to know the weather for today?
    buttons:
    - title: Today
      payload: here
    - title: Whole week
      payload: other_location
```
This is primarily because the response selector can only be trained with one ground truth label. Supporting multiple possible responses should be left out of scope for now IMO.

2. I am not sure of this but the `text` attribute of any response template is optional, right? For example, is this valid? - 
```
responses:
  chitchat/ask_weather:
  - buttons:
    - title: Current location
      payload: here
    - title: Other place
      payload: other_location
```
If it is valid, then we should use the full name of the retrieval intent(for e.g. - `chitchat/ask_weather`) as the proxy for the `text` attribute(just for model training)

tmbo:
> In contrast to usual response templates, there cannot be multiple text attributes for one retrieval intent. For example, this would be invalid:

just because the ML model can't be trained with multiple of these, it doesn't mean this needs to be invalid. e.g. we could do the same as we do for utterance templates and just select a random one.

tmbo:
> I am not sure of this but the text attribute of any response template is optional, right? For example, is this valid? -

Yes I think that is valid. I like the solution, let's use the intent name in that case

dakshvar22:
> just because the ML model can't be trained with multiple of these, it doesn't mean this needs to be invalid. e.g. we could do the same as we do for utterance templates and just select a random one.
Yes, we can do that but which one would the model be trained on? It has to be just one and taking a random one for training, doesn't seem like a good option. If we always take the first response as the training response would it introduce any confusion/problems in the UX?(I am not very clear on this myself) For example, model would need to be retrained if the first response was swapped with the last.

tmbo:
well, we can train the model on either the first, all of them or any of them: whatever you prefer.

I think from a UX perspective it is easier to explain that the format is exactly the same as the one for utterance templates. The model gets retrained on any change to the responses at the moment, I think from the training perspective it is pretty consistent.

rasa/nlu/training_data/training_data.py
```diff
@@ -232,27 +236,21 @@ def sort_regex_features(self) -> None:
...
-                    )
+            if example.get(RESPONSE_KEY_ATTRIBUTE) is None:
+                continue
+
+            # look for corresponding bot utterance
+            story_lookup_intent = example.get_combined_intent_response_key()
+            assistant_utterances = self.responses.get(story_lookup_intent, [])
+            if assistant_utterances:
+                # selecting only first assistant utterance for now
+                example.set(RESPONSE, assistant_utterances[0].get("text"))
```
wochinge, line 253: Can we re-use an existing constant like `nlu.constants.TEXT` or define a new one?

rasa/nlu/training_data/training_data.py
```diff
@@ -285,13 +295,21 @@ def nlu_as_yaml(self) -> Text:
...
     def persist_nlu(self, filename: Text = DEFAULT_TRAINING_DATA_OUTPUT_PATH):
 
-        if filename.endswith("json"):
+        if Path(filename).suffix in JSON_FILE_EXTENSIONS:
             rasa.nlu.utils.write_to_file(filename, self.nlu_as_json(indent=2))
-        elif filename.endswith("md"):
+        elif Path(filename).suffix in MARKDOWN_FILE_EXTENSIONS:
+            rasa.nlu.utils.write_to_file(filename, self.nlu_as_markdown())
+        elif Path(filename).suffix in YAML_FILE_EXTENSIONS:
             rasa.nlu.utils.write_to_file(filename, self.nlu_as_markdown())
```
wochinge, line 313: I'd also add a test which checks the correct persisting behavior 
```suggestion
            rasa.nlu.utils.write_to_file(filename, self.nlu_as_yaml())
```

rasa/nlu/training_data/training_data.py
```diff
@@ -285,13 +295,21 @@ def nlu_as_yaml(self) -> Text:
...
-        return RasaYAMLWriter().dumps(self)
+        # avoid dumping NLG data (responses). this is a workaround until we
+        # can remove the distinction between nlu & nlg when converting to a string
+        # (so until after we remove markdown support)
+        no_responses_training_data = copy.copy(self)
+        no_responses_training_data.responses = {}
+
+        return RasaYAMLWriter().dumps(no_responses_training_data)
 
     def persist_nlu(self, filename: Text = DEFAULT_TRAINING_DATA_OUTPUT_PATH):
```
wochinge, line 306:
```suggestion
    def persist_nlu(self, filename: Text = DEFAULT_TRAINING_DATA_OUTPUT_PATH) -> None:
```

rasa/nlu/training_data/training_data.py
dakshvar22, line 253: @tmbo I haven't read through the changes in the PR, but i am assuming this line means that we always take the text of the first response for model training in case there are multiple?

rasa/nlu/training_data/training_data.py
```diff
@@ -384,6 +415,16 @@ def validate(self) -> None:
                     f"this the training may fail."
                 )
 
+        # emit warnings for response intents without a response template
+        for example in self.training_examples:
+            if example.get(RESPONSE_KEY_ATTRIBUTE) and not example.get(RESPONSE):
+                raise_warning(
+                    f"Your training data contains an example '{example.text[:20]}...' "
```
wochinge, line 422: optional: use a constant for `20`

rasa/nlu/training_data/training_data.py
```diff
@@ -394,44 +435,38 @@ def train_test_split(
...
-    def build_nlg_stories_from_examples(examples) -> Dict[Text, list]:
+    def _needed_responses_for_examples(
+        self, examples: List[Message]
+    ) -> Dict[Text, List[Dict[Text, Any]]]:
+        """Get all responses used in any of the examples."""
 
-        nlg_stories = {}
+        responses = {}
         for ex in examples:
             if ex.get(RESPONSE_KEY_ATTRIBUTE) and ex.get(RESPONSE):
```
wochinge, line 466: can no longer happen due to the new validation, can it?
```suggestion
            if ex.get(RESPONSE_KEY_ATTRIBUTE):
```

rasa/nlu/training_data/training_data.py
```diff
@@ -394,44 +435,38 @@ def train_test_split(
...
-        train_nlg_stories = self.build_nlg_stories_from_examples(train)
-        test_nlg_stories = self.build_nlg_stories_from_examples(test)
-        return test_nlg_stories, train_nlg_stories
-
-    @staticmethod
-    def build_nlg_stories_from_examples(examples) -> Dict[Text, list]:
+    def _needed_responses_for_examples(
+        self, examples: List[Message]
+    ) -> Dict[Text, List[Dict[Text, Any]]]:
+        """Get all responses used in any of the examples."""
```
wochinge, line 462: optional: Let's either add a full docstring according to our Python conventions or None

tests/nlu/training_data/test_training_data.py
```diff
@@ -122,12 +122,12 @@ def test_demo_data(files):
     td = training_data_from_paths(files, language="en")
```
wochinge, line 122: do we have a test for the new validation of `responses` in `TrainingData`?

tests/nlu/selectors/test_selectors.py
```diff
@@ -38,7 +38,6 @@ def test_train_selector(pipeline, component_builder, tmpdir):
     td = load_data("data/examples/rasa/demo-rasa.md")
     td_responses = load_data("data/examples/rasa/demo-rasa-responses.md")
     td = td.merge(td_responses)
-    td.fill_response_phrases()
```
wochinge, line 41: awesome moving this to the constructor 🚀

rasa/nlu/training_data/training_data.py
tmbo: yes, the machine learning part is the same as it was before

dakshvar22: Okay, do you think we should still run the model regression tests before merging it?
The benchmarking datasets aren't in yaml format, nor do they contain any multimedia elements though.

rasa/nlu/training_data/training_data.py
tmbo: I think it's fine to keep the 20: it is readable and just for logging. since it is very obvious what is happening, naming this random constant will not increase readability.

tests/nlu/training_data/test_training_data.py
tmbo: isn't that what https://github.com/RasaHQ/rasa/pull/6323/files#diff-91cfb497152a74c0719d07aaae3db53dR304 does? there is nothing more to test apart from the key being present and not empty, the validation isn't more specific.

rasa/nlu/training_data/training_data.py
tmbo: don't think that is necessary since I didn't touch the ML parts, the input data is still exactly the same (tested in unit tests).

rasa/server.py
```diff
@@ -835,6 +855,13 @@ async def evaluate_intents(request: Request) -> HTTPResponse:
             "evaluate your model.",
         )
 
+        if request.headers.get("Content-type") == YAML_CONTENT_TYPE:
+            test_data = _training_payload_from_yaml(request).get("training_files")
+        else:
+            test_data = rasa.utils.io.create_temporary_file(
+                request.body, mode="w+b", suffix=".md"
+            )
+
```
wochinge, line 864: optional: extract to function and de-duplicate as it's the same as above

rasa/server.py
```diff
@@ -811,14 +825,20 @@ async def evaluate_stories(request: Request) -> HTTPResponse:
             "evaluate your model.",
         )
 
-        stories = rasa.utils.io.create_temporary_file(request.body, mode="w+b")
+        if request.headers.get("Content-type") == YAML_CONTENT_TYPE:
+            test_data = _training_payload_from_yaml(request).get("training_files")
```
wochinge, line 829: 🤠

rasa/server.py
```diff
@@ -1177,14 +1203,12 @@ def _model_output_directory(save_to_default_model_directory: bool) -> Text:
 
 
 def _validate_yaml_training_payload(yaml_text: Text) -> None:
-    # TODO: Use NLU schema validation once it's merged
     try:
-        rasa.utils.io.read_yaml(yaml_text)
-    except Exception:
+        RasaYAMLReader.validate(yaml_text)
```
wochinge, line 1207: 🙏

tests/test_server.py
```diff
@@ -399,16 +401,23 @@ def test_train_nlu_success(
     default_nlu_data: Text,
     default_domain_path: Text,
 ):
-    with ExitStack() as stack:
-        domain_file = stack.enter_context(open(default_domain_path))
-        config_file = stack.enter_context(open(default_stack_config))
-        nlu_file = stack.enter_context(open(default_nlu_data))
+    domain_data = rasa_utils.io.read_yaml_file(default_domain_path)
```
wochinge, line 404: How about using `parametrize` or adding another test to make sure it still works with markdown?

rasa/nlu/training_data/training_data.py
```diff
@@ -272,6 +271,18 @@ def nlg_as_markdown(self) -> Text:
 
         return NLGMarkdownWriter().dumps(self)
 
+    def nlg_as_yaml(self) -> Text:
+        """Generates yaml representation of the response phrases (NLG) of
+        TrainingData."""
```
wochinge, line 276: let's also describe the return type in the docstring

rasa/nlu/selectors/response_selector.py
```diff
@@ -323,6 +326,14 @@ def preprocess_train_data(self, training_data: TrainingData) -> RasaModelData:
 
         return model_data
 
+    def _full_response(self, label: Dict[Text, Any]) -> Optional[Dict[Text, Any]]:
+        """Given a label return the full response based on the labels id."""
```
wochinge, line 330: let's either add a full docstring or none

rasa/nlu/selectors/response_selector.py
```diff
@@ -323,6 +326,14 @@ def preprocess_train_data(self, training_data: TrainingData) -> RasaModelData:
...
+        for key, responses in self.responses.items():
+            for response in responses:
+                if hash(response.get("text", "")) == label.get("id"):
```
wochinge, line 333: let's use the existing constant

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -880,6 +881,7 @@ def _predict_label(
         # if X contains all zeros do not predict some label
         if label_ids.size > 0:
             label = {
+                "id": hash(self.index_label_id_mapping[label_ids[0]]),
```
wochinge, line 884: for my understanding: That's the hash of the intent name then, correct?

rasa/nlu/selectors/response_selector.py
wochinge: uff. the `text` of a `response` is the `RESPONSE` of a `Message`, correct? Maybe makes sense to add a comment

tests/nlu/training_data/formats/test_rasa_yaml.py
```diff
@@ -280,3 +282,97 @@ def test_minimal_yaml_nlu_file(tmp_path):
...
+
+    with pytest.raises(InvalidDomain):
+        reader.reads(responses_yml)
+
+
+def test_nlg_multimedia_load_dump_roundtrip():
+    responses_yml = textwrap.dedent(
+        """
+      responses:
+        chitchat/ask_weather:
```
wochinge, line 360: optional: add another response to make sure the order isn't changed when dumping

docs/docs/chitchat-faqs.mdx
```diff
@@ -26,65 +26,70 @@ In this section we're going to cover the following topics:
...
+```yaml title="config.yml"
 policies:
 - name: MemoizationPolicy
   max_history: 1
 - name: MappingPolicy
 ```
 
-:::note
-The MappingPolicy is there because it handles the logic of the `/restart` intent,
+:::note MappingPolicy
```
wochinge, line 62: should probably be part of different PR: they should use the RulePolicy instead of the MappingPolicy

rasa/nlu/schemas/nlu.yml
```diff
@@ -46,6 +46,10 @@ mapping:
         lookup:
           type: "str"
         examples: *examples_anchor
+  responses:
+    type: "map"
```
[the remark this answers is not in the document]
tmbo: I don't think that is possible since the keys are the intent names, e.g. `chitchat/ask_name` and they would need to be known to further validate.

I think in the future we should consider moving to strict yaml instead of normal yaml + validation

rasa/nlu/training_data/lookup_tables_parser.py
```diff
@@ -1,8 +1,8 @@
-from typing import Text, List, Dict
+from typing import Any, Text, List, Dict
 
 
 def add_item_to_lookup_tables(
-    title: Text, item: Text, existing_lookup_tables: List[Dict[Text, List[Text]]]
+    title: Text, item: Text, existing_lookup_tables: List[Dict[Text, Any]]
```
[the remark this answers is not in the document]
tmbo: this failed in my pytype, but I can change it back

rasa/core/training/story_reader/markdown_story_reader.py
wochinge: Should we add this to the changelog? `Training data form in markdown format` has to have the file suffix `.md` from now on`?

rasa/nlu/classifiers/diet_classifier.py
tmbo: nope hash of the text...but this will change again once #6392 gets merged...

rasa/nlu/selectors/response_selector.py
tmbo: yes, that is whats trained on. but as added in the other comment, this will change in the pr from daksh

rasa/nlu/selectors/response_selector.py
tmbo: I'll do so, but I am not sure this is a usefull practice: I am all with you that it is necessary for public methods. For private methods, I don't see why it is better to have no comment on a method than to add just a comment without args (assuming a helpful comment).

rasa/nlu/schemas/nlu.yml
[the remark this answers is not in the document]
wochinge: I think it should work using https://pykwalify.readthedocs.io/en/unstable/validation-rules.html#matching-rule 🤔

rasa/nlu/training_data/lookup_tables_parser.py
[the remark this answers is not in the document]
wochinge: what did pytype say?

docs/docs/chitchat-faqs.mdx
tmbo: yes, but that is separate

rasa/nlu/classifiers/diet_classifier.py
wochinge: That's within the `DietClassifier` so I think it's the intent, no? Anyway, not important if get's changed by Daksh's PR

rasa/nlu/selectors/response_selector.py
wochinge: I think that's something which we should discuss as part of the coding conventions. I'd argue that 
- describing parameters + return type isn't that much of an effort
- then the comment is even more helpful
- it's more consistent

rasa/nlu/schemas/nlu.yml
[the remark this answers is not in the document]
tmbo: ah good point :+1: will add that in a separate PR as it needs to be added for the domain as well (I just copied it from there)

tmbo:
thanks a lot @wochinge fur the super quick and thorough review :rocket:

tests/test_server.py
tmbo: I took a closer look and there are already a couple of other methods in there that test the same endpoint with markdown files. since this actually trains a model and takes time, I'd rather avoid adding more training runs. what do you think?

wochinge: Yes, but then I'd maybe rename the tests to make it clear that this testing the MD support. Otherwise somebody might come, adapt the test, and suddenly all MD testing is gone.

Another quick way would be to extract the payload extraction to a separate function and test that using `parametrize`.

rasa/nlu/selectors/response_selector.py
tmbo: mhm, right thinking about it I agree :+1:

wochinge:
@tmbo During the review I noticed the close coupling of training data objects (`Domain`, `TrainingData`) with the different `Writer` classes.

What do think about having a pendant to `TrainingDataImporter` to 
- support other file format implementations (e.g. implementations of the community, adapters to other frameworks)
- clear separation of concerns / classes don't do multiple things at the same time
- easier testability
- DRYer (each class is doing some sort of `if endswith "json" elif "endswith md"` currently)

tmbo:
yes @wochinge I think that is a good idea :+1:

PR #6325, opened 2020-08-03
useBaseUrl() hook for redoc

m-vdb:
**Proposed changes**:
- Fix static url for redoc

### DIFF
--- a/docs/docs/api/action-server.mdx
+++ b/docs/docs/api/action-server.mdx
@@ -6,6 +6,7 @@ description: Check out the API docs for open source chatbot framework Rasa's act
 hide_table_of_contents: true
 ---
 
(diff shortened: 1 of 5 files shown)


lunelson:
Haven't pulled it in and tested it but it looks like the right approach. Wondering about this line, and whether you can't use `import()` Promise syntax instead of `require()`, so that it remains tree-shakeable?

https://github.com/RasaHQ/rasa/blob/47445b3aab6fc41802780cef7987a770a34eb5df/docs/src/components/redoc.jsx#L11

m-vdb:
@lunelson I managed to make it work, and I tested locally with a production build, too. Can you review this PR?

Issue #6303, opened 2020-07-30
command: `rasa init` still used old data format and policy in Rasa v2.0.0a1


howl-anderson:
**Rasa version**: 2.0.0a1

**Issue**:

command: `rasa init` still used old data format and policy in Rasa v2.0.0a1

update the content created by rasa init from 1.x to 2.x will help lots of people who want to know what's new in 2.0

sara-tagger:
Thanks for the issue, @b-quachtran will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

akelad:
@TyDunn is prod gonna handle this? CSE has been taking care of updating demo bots, and we overlooked this one I guess so we can tackle it (should be a v quick job)

@howl-anderson thanks a lot for pointing that out!

TyDunn:
@akelad Not in the current sprint, so I think it would be best if y'all take it on

PR #6329, opened 2020-08-03
Rasa init to 20

ArjaanBuijk:
**Proposed changes**:
- Update the `rasa init` bot to 2.0 (fixes https://github.com/RasaHQ/rasa/issues/6303)

- [x] Replaced the files of the bot
- [x] Updated the default configuration of Rasa to 2.0


ArjaanBuijk:
@chkoss ,
I am asking for your review because I updated the default config to 2.0 as part of migrating the `rasa init` bot.

chkoss:
@ArjaanBuijk Will see if I get to this today, otherwise I'll review it on Thursday.

akelad:
@chkoss while we talked about this something came up - why do we actually have this as a separate project and don't just pull it from the moodbot example directory? I think they're exactly the same bots right?

chkoss:
@akelad I think it makes sense to keep it separate. While (currently) containing basically the same bot, one is supposed to be a finished mini example project, the other one is supposed to be the starting point for a new project (including some example data so you see what it should look like).

At the moment, the projects differ in e.g.
- `config.yml` (moodbot: some configuration, init: mostly empty with comments about how to use the suggested config feature)
- `actions.py` (moodbot: no such file as the bot doesn't have actions, init: an empty file with comments on how to use custom actions)
- `endpoints.yml`
- `credentials.yml`
- There are more nlu examples for moodbot
- The domain file for moodbot has two buttons for `utter_greet` while the one for init just has text

What could make sense is to pull `nlu`, `stories` and `domain` from the moodbot example.

But then again, moodbot might get extended at some point (e.g. adding even more nlu examples and stories), whereas that doesn't make much sense for the init project which isn't supposed to be "good at its job" but only to demonstrate what the project files should look like.

akelad:
@chkoss hm, do you think we should hard code a version? we'll have to manually update that every time. should this even be that aggressive a warning for the user?

akelad:
i applied your suggested changes, and also removed the dots from the rules since that's merged now - can you double check all the format is definitely up to date?

akelad:
ok will you create an issue for that? 🙏

PR #6343, opened 2020-08-05
Generate the source code reference

m-vdb:
## Proposed changes
- Add a `docusaurus` renderer here: https://github.com/RasaHQ/pydoc-markdown/pull/1
- This renderer autogenerates the docs + a sidebar. We can include this sidebar in our sidebar conf and magic happens
Screenshot 2020-08-05 at 15 50 40

## Checklist
- [ ] some docs will need revamping, either do it on this PR or in a separate PR with CSE


Makefile
```diff
@@ -58,6 +58,8 @@ install-full-windows: install install-mitie
 	poetry install -E spacy -E transformers -E jieba
 
 install-docs:
+	# FIXME: poetry doesn't support installing from a git subdirectory
+	poetry run pip install git+https://github.com/RasaHQ/pydoc-markdown.git@23d825f#subdirectory=pydoc-markdown
```
m-vdb, line 62: really not idea but not sure there is a workaround

docs/sidebars.js
```diff
@@ -103,7 +103,10 @@ module.exports = {
       type: 'category',
       label: 'Reference',
       collapsed: false,
-      items: ['glossary'],
+      items: [
+        'glossary',
+        require("./docs/reference/sidebar.json"),
```
m-vdb, line 108: this is generated by the script

Makefile
tmbo: can we just publish and install from our pypi.rasa.com?

m-vdb: happy to do it; can you share credentials in 1password?

tmbo:
merged since it already passed tests

Issue #6208, opened 2020-07-15
RulePolicy should predict Core fallback action if no rule was found


wochinge:
With the RulePolicy you can now write rules to handle low NLU confidence.
However, as we deprecate the FallbackPolicies we now have no policy which predicts a fallback action in case the Core confidence is low.

**Proposed Steps**
- [ ] The `RulePolicy` should predict a fallback action `x` with a confidence `y` in case no matching rule was applied
- [ ] the name `x` of the fallback action can be configured in the policy config (`config.yml`)
- [ ] the confidence threshold `y` for the fallback can be configured in the policy config
- [ ] add a config value to enable / disable this behavior

```yaml
polices:
- name: RulePolicy
  core_threshold: 0.5
  core_fallback_action_name: action_default_fallback
  enable_fallback_prediction: True/False
```

PR #6345, opened 2020-08-05
predict core fallback if no rule found

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6208

```yaml
policies:
  - name: RulePolicy
    # Confidence of the prediction if no rule matched d de-facto threshold for a
    # core fallback.
    core_fallback_threshold: 0.3
    # Name of the action which should be predicted if no rule matched.
    core_fallback_action_name: "action_default_fallback"
    # If `True` `core_fallback_action_name` is predicted in case no rule matched.
    enable_fallback_prediction: True

```
- [ ] updated the documentation (will happen in https://github.com/RasaHQ/rasa/pull/6195/files)


wochinge:
@melindaloubser1 @b-quachtran Can you please give some feedback on the parameter naming?

melindaloubser1:
Sounds intuitive to me 👍

tests/core/test_agent.py
```diff
@@ -292,6 +292,63 @@ def test_two_stage_fallback_without_deny_suggestion(
...
+    ],
+)
+def test_rule_policy_without_fallback_action_present(
+    domain: Dict[Text, Any], policy_config: Dict[Text, Any]
+):
+    with pytest.raises(InvalidDomain):
+        Agent(
+            domain=Domain.from_dict(domain),
+            policies=PolicyEnsemble.from_dict(policy_config),
+        )
```
chkoss, line 315: To catch wrong error message formatting like above 😉 
```suggestion
    with pytest.raises(InvalidDomain) as execinfo:
        Agent(
            domain=Domain.from_dict(domain),
            policies=PolicyEnsemble.from_dict(policy_config),
        )
    assert "fallback action 'my_fallback' must be present" in str(execinfo.value)
```

tests/core/policies/test_rule_policy.py
```diff
@@ -1,4 +1,4 @@
-from typing import List, Text
+from typing import List, Text, Optional
```
chkoss, line 1:
```suggestion
from typing import List, Text
```

rasa/core/policies/rule_policy.py
```diff
@@ -23,8 +23,12 @@
     ACTION_BACK_NAME,
     ACTION_SESSION_START_NAME,
     RULE_SNIPPET_ACTION_NAME,
+    ACTION_DEFAULT_FALLBACK_NAME,
 )
 
+if TYPE_CHECKING:
+    from rasa.core.policies.ensemble import PolicyEnsemble
```
chkoss, line 30: I think this should solve the "Check Types" CI issue. (The file with `PolicyEnsemble` imports `RulePolicy`, so we have an import cycle. But as the one here is only for type checking, it should be fine to simple disable.)
```suggestion
    from rasa.core.policies.ensemble import PolicyEnsemble # pytype: disable=pyi-error
```

wochinge:
> I think it would make sense to add another rule policy test that the fallback action is not wrongly predicted. So basically setting up a simple case where a rule applies and then

That's covered by all the other `RulePolicy` tests imo. They'd all break if the fallback would jump in and overwrite the actual prediction. Hence, I'd skip adding the test.

tests/core/test_agent.py
wochinge: I am personally not a huge fan of testing the content of a log message / exception message and prefer to test **that** the exception / log message is thrown / printed. The reason is that it makes the test very fragile to minor changes and the message content not really belongs to the functionality for me.

I also think that above are somewhat weak arguments and tried to find some best practices:
- https://softwareengineering.stackexchange.com/questions/384164/should-log-statements-be-tested
- https://stackoverflow.com/questions/11998713/should-we-unit-test-logging
- https://softwareengineering.stackexchange.com/questions/406079/testing-the-wording-of-an-exception-message
- https://thecodebuzz.com/unit-testing-exceptions-best-practices-net-core/ (this one is somewhat pro)

I'd say that it's somewhat optional to test it and as the correct formatting of the message is not critical, it's okay to just test **that** the exception happens.

Wdyt?

chkoss: Thanks for the resources!

Seems to me that there are two good ways:
- only test that it is thrown
- also test that the correct error is thrown (there are several cases where our code raises `InvalidDomain`, for different reasons with different messages)
If we want to test if it is the correct error, we can reduce the snippet to be only `"my_fallback"` to make it less fragile. The name of the fallback action should always be part of the message even if we decide to rephrase it.

Both options would be fine with me, feel free to pick one.
We should quickly go through the rest of the test file for consistency though. `test_form_without_form_policy`, `test_trigger_without_mapping_policy`, `test_two_stage_fallback_without_deny_suggestion` and `test_pull_model_with_invalid_domain` all test for message snippets.

wochinge: I agree. Testing the exception message now to see that the exception was caused by the RulePolicy 👍

PR #6357, opened 2020-08-07
2.0 training data format docs

melindaloubser1:
**Proposed changes**:
- Initial docs for 2.0 training data format.

**TODOS**:
- [x] complete the responses section - Edit: pointing to the domains page since the format is still relevant and the page also contains a better explanation of using responses
- [ ] clarify the entity format meaning in stories
- [ ] merge in new structure
- [ ] correct links when new structure merged


melindaloubser1:
Ope, forgot to add a form section. I'll do that now.

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+  - Hallo
+  - Hi
+
+- intent: faq/language
+  examples: |
+  - What language do you speak?
+  - Do you only handle english?
+
+e2e_tests:
+- user: |
```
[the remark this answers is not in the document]
melindaloubser1: I was basing this off the notion page, so I'm actually not sure. I think we would only need it if it was a multiline utterance? @wochinge

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+Note the `metadata` key inside the `examples` dictionary. It can contain arbitrary key-value data and the parser will not read its value. You can use it to store
+any information relevant to the example.
+
+### Entities
+
+Where applicable, entities are annotated in training examples using the
+syntax:
+
+```yaml
+[<entity-text>]{"entity": "<entity name>"}
```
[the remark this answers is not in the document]
melindaloubser1: Oh, I thought that was getting deprecated?

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+description: Description of the YAML format for training data
+---
+
+## Overview
+
+### YAML Format
+
+Rasa Open Source 2.0 uses [YAML](https://yaml.org/spec/1.2/spec.html) as
+a unified and extendable way to manage all kinds of training data,
+including NLU data, stories, and responses. With the YAML format, users
```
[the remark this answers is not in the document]
melindaloubser1: I'm adding a reference to the domain and leaving out anything that goes in the domain from this page, since it's already defined.

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+- [Lookup Tables](#lookup-tables)
+
+Here is a simple example of each item type:
+
+```yaml
+nlu:
+- intent: greet
+  examples: |
+  - hey
+  - hello
```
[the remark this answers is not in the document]
melindaloubser1: I'm adding one

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+  examples: |
+  - Peso
+  - Euro
+  - Dollar
+```
+
+
+### Training Examples
+
+Training examples are grouped by intent and listed under the
```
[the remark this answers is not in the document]
melindaloubser1: Yes they can! woot woot

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+  - text: |
+      hallo
+    metadata:
+      sentiment: neutral
+  - text: |
+      hey there!
+```
+
+Note the `metadata` key inside the `examples` dictionary. It can contain arbitrary key-value data and the parser will not read its value. You can use it to store
+any information relevant to the example.
```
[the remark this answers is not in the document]
melindaloubser1: For now I'm just explaining that it is recording sentiment as something one might be interested in besides intent & entiteis

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+available regex features.
+
+:::note
+
+Regex features only define entities when used in combination with the RegexEntityExtractor. Otherwise they  
+don't define entities nor intents! They simply provide patterns to
+help the classifier recognize entities and related intents. Hence, you
+still need to provide intent & entity examples as part of your
+training data!
+:::
```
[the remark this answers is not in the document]
melindaloubser1: I think it would be good if we could focus only on format here, it's a bit tricky with these components though because the format depends on which component you're using (name doesn't matter or does) and there isn't really a usage explanation elsewhere. For now let's just pare it down a bit?

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+- intent: check_balance
+  examples: |
+  - how much do I have on my [credit card account]{"entity": "account", "value": "credit"}
+  - how much do I owe on my [credit account]{"entity": "account", "value": "credit"}
+```
+
+To use the synonyms defined in your training data, you need to make sure the
+pipeline contains the `EntitySynonymMapper` component (see components). You
+should define synonyms when there are multiple ways users refer to the same
+thing.
```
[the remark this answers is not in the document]
melindaloubser1: I think rather best practices, this page is long enough already

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+
+Synonyms can also be definedin-line in your training examples by
+specifying the `value` of the entity:
+
+```yaml
+nlu:
+- intent: check_balance
+  examples: |
+  - how much do I have on my [credit card account]{"entity": "account", "value": "credit"}
+  - how much do I owe on my [credit account]{"entity": "account", "value": "credit"}
```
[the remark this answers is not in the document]
melindaloubser1: it does, sorry I thought it didn't

docs/docs/api/training-data-format.mdx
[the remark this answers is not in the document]
melindaloubser1: nope it isn't sorry

docs/docs/api/training-data-format.mdx
```diff
@@ -0,0 +1,637 @@
...
+
+- regex: zipcode
+  examples: |
+  - [0-9]{5}
+
+- lookup: additional_currencies
+  examples: |
+  - Peso
+  - Euro
+  - Dollar
```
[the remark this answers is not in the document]
melindaloubser1: Currency also should actually be duckling, i'll make it account type all the way through

melindaloubser1:
Left over review comments/questions from Akela:
[ ] For single examples, is the vertical bar necessary? e.g. 
```
  examples: |
  - [0-9]{5}
```
vs
```
  examples: [0-9]{5}
```
[ ] Are forms referred to as loops now? 
[ ] Desired vs. actual behaviour of versioning (if later than latest, is it ignored or will it use the latest version?)
[ ] Todo: Pull in actual latest version number

tmbo:
regarding 

>  For single examples, is the vertical bar necessary? e.g.
  examples: |
>   - [0-9]{5}

yes, let's keep it consistent - it is unlikely that people will only have one example (they really shouldn't ) and it keeps things simpler

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
+---
+id: training-data-format
+sidebar_label: Training Data Format
+title: Training Data Format
+description: Description of the YAML format for training data
+---
+
+## Overview
+
+### YAML Format
```
tmbo, line 10: I'd remove this headline and put the text under `overview`. I think we should either add a paragraph or note on old data formats with a link to their documentation

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+Rasa Open Source 2.0 uses [YAML](https://yaml.org/spec/1.2/spec.html) as
```
tmbo, line 12:
```suggestion
Rasa Open Source uses [YAML](https://yaml.org/spec/1.2/spec.html) as
```

I don't think we need to mention that if we add a note about the deprecated formats

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+a unified and extendable way to manage all training data,
+including NLU data, stories and rules. 
+
+With the YAML format, training data can be split over any number of YAML files,
+and every file can contain any kind of data.
```
tmbo, line 17: @degiz doing the multimedia response implementation I noticed that this is actually not quite true (we should make it work though). We do the "what is this file type containing" discovery (is it nlu, stories, ...) and the parsing afterwards. if your file contains stories and nlu data, only one of them will be read. are you aware of that?

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+- `version`
+- `stories`
+- `nlu`
+- `rules`
+- `e2e_tests`
+
+All YAML training data files should specify the `version` key in order
+to be parsed correctly. Training data files with no `version` specified will be assumed to
+be in the format of the latest version of Rasa Open Source. Training data files with a version greater than is currently
```
tmbo, line 41: the version is NOT specifying the rasa open source version, but the training data version

tmbo: I'd rather phrase it as (also using active voice)

"If you don't specify a `version` key in your training data file, Rasa will assume you are using the latest training data format specification."

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+rules:
+- rule: Greet user
+  steps:
+  - intent: greet
+  - action: utter_greet
+
+responses:
+  faq/language:
+  - text: |
```
tmbo, line 67: I think we should not use blocks for these but rather use simple strings. e.g.
`- text: "I can only do English at the moment"`

tmbo: it avoids people confusing this with the nlu syntax and doing odd things like adding a `-` in the beginning

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+  intent: greet
+- action: utter_greet
+- user: |
+    what language do you speak
+  intent: faq/language
+- action: respond_faq
+```
+
+
+## NLU Training Data
```
tmbo, line 98: should we rather call this something like "User Message Training Data"?

tmbo: I'd rather like to avoid the term NLU in most areas as the user likely doesn't know what it means

tmbo: even if we introduce NLU as a term in this section, we should avoid using these terms in the title of that section

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+NLU training data is defined under the `nlu` key. Items that can be added under this key are:
+
+- [Training examples](#training-examples) grouped by user intent,
+  optionally with annotated [entities](#entities)
+- [Synonyms](#synonyms)
+- [Regular Expression Features](#regular-expression-features)
+- [Lookup Tables](#lookup-tables)
+
+Here is a simple example of each item type:
```
tmbo, line 115: should we rather split this up and add the examples separately to the above list? I think that makes it easier to understand and map

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+```yaml
+nlu:
+- intent: greet
+  examples: |
+  - hey
+  - hi
+  - whats up
+```
+
+However, its also possible to use an extended format if you want to include
```
tmbo, line 155:
```suggestion
However, its also possible to use an extended format if you need to include
```

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+  - text: |
+      hi
+    metadata:
+      sentiment: neutral
+  - text: |
+      hey there!
+```
+
+The `metadata` key can contain arbitrary key-value data and the parser will not read its value. 
+Metadata can be useful for storing information relevant to an example, but not used by
```
tmbo, line 171: I don't think this is actually true, e.g. the sentiment in the above example would be used in an NLU component (we just don't have one by default).

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+  - text: |
+      hey there!
+```
+
+The `metadata` key can contain arbitrary key-value data and the parser will not read its value. 
+Metadata can be useful for storing information relevant to an example, but not used by 
+the NLU pipeline. In the example above, the sentiment of 
+the example is recorded as metadata.
+
+### Entities
```
tmbo, line 175: should we add a description what entities are here first?

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+Where applicable, [entities])(../glossary#entity) can be annotated in training examples using either the
+abbreviated syntax:
+
+```yaml
```
tmbo, line 180: I think we should discuss if we should really add both of them or if it would be better to just mention the more flexible one using `{}`

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+In a training example, this would look like:
+
+```yaml
+nlu:
+- intent: check_balance
+  examples: |
+  - how much do I have on my [savings]("account") account
+  - how much money is in my [checking]{"entity": "account"} account
+  - What's the balance on my [credit card account]{"entity":"account","value":"credit"}
```
tmbo, line 199: I think we should always keep this in mind and use this structure throughout our docs:
explanation first, then an example block.

The above has a structure that is more like
explanation, example with additional syntax, explanation of the additional syntax

This makes jumping through sections harder in my opinion. I'd rather remove the third example and introduce it separately in the paragraph that focuses on it.

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+following section. To understand what the labels `role` and `group` are
+for, see the section on (entity roles and groups)[nlu/entity-extraction#entities-roles-and-groups].
+
+### Synonyms
+
+Synonyms provide a way to normalize your training data by mapping an
+extracted entity to a value other than the literal text extracted.
+Synonyms can be defined in the format:
+
+```yaml
```
tmbo, line 222: let's rather use an actual example here

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+nlu:
+- intent: check_balance
+  examples: |
+  - how much do I have on my [credit card account]{"entity": "account", "value": "credit"}
+  - how much do I owe on my [credit account]{"entity": "account", "value": "credit"}
+```
+
+To use the synonyms defined in your training data, you need to make sure the
+pipeline contains the [`EntitySynonymMapper`](../nlu/components#entitysynonymmapper) component. You
```
tmbo, line 241: should mention the configuration, as in "pipeline in your configuration file"

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+**warning**
+
+You have to be careful when you add data to the lookup table.  
+For example, if there are false positives or other noise in the table,
+this can hurt performance. So make sure your lookup tables contain
+clean data.
+
+
+## Stories
```
tmbo, line 376: should we rather call this `Conversation training data (stories)`?

docs/docs/user-guide/training-data-format.mdx
degiz: If a file contains `nlu` and `stories` key, it should be accepted by both the NLU and Stories parser.
If that's not the case, I need to fix that 😬 I'll check that.

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: Is there info somewhere about training data versions?

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: Would stories & rules both fall under `Conversation training data`?

tmbo: mhm that is a good point. Maybe it should be something along the lines of 

## Conversation training data
... You can do stories that use machine learning to identify patterns ... You can use rules to implement deterministic outcomes ... (prob needs better phrasing, but just for illustration)
### Stories
...
### Rules
...

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: Ah right, I should actually remove the responses example entirely because at the top it says they're part of the domain

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: imo, if both are possible still, that should be mentioned. People coming from 1.x might try it anyway and be confused if they see it works but there's no docs for it.

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: Ok, I was wondering about that. Is this true then?
```
The `metadata` key can contain arbitrary key-value data. The parser will read it's value, but it will only be used if a component in the pipeline needs it. In the example above, the sentiment of 
the example is recorded as metadata. This could be used by a custom component in the pipeline for sentiment analysis.
```

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+With the YAML format, training data can be split over any number of YAML files,
+and every file can contain any kind of data. 
+The training data parser will read the top level keys in each file to decide
+what kind of data is in a section at training time.
+
+The [domain](../glossary#domain) uses
+the same YAML format as the training data and can also be split across
+multiple files or combined with training data in one file. The domain includes
+the definitions for [responses](../core/domains#responses) and [forms](../core/forms).
+See the [documentation for domains](../core/domains) to learn more.
```
erohmensing, line 23: @m-vdb i'm confused, do these links need to have `.mdx` in them or not? Running them locally with the `make install-docs` from `master` and the links including the anchors seem to render fine 🤔

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: I'm not sure about this. If users are looking at the training data format (which will be not at the top the way it is now), I think they've already seen NLU and know what it is. We also talk about the NLU model, so it makes it clear what model it's feeding.

melindaloubser1: It's true it should be defined first though

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: There's this at the top of the nlu section - should it move?
```
Training examples can also include **entities**. Entities are structured
pieces of information that can be extracted from a user's message.
```

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: 👍 switched it around

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: lmk if it makes sense now

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: added note; not sure where links should point

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+Then, if either of these phrases is extracted as an entity, it will be
+mapped to the **value** `credit`.
+
+:::note
+Synonym mapping only happens **after** entities have been extracted.
+That means that you need to provide examples of the variations 
+on a synonym in your training examples so
+that Rasa Open Source can learn to pick them up.
+:::
```
erohmensing, line 248: How about following the example: 

Synonym mapping only happens **after** entities have been extracted.
That means that your training examples should include the synonym examples (`credit card account` and `credit account`) so that the model will learn to recognize these as entities and replace them with `credit`.

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+### Regular Expression Features
+
+Regular expressions can be used in two different ways:
+
+1. They can be used to support intent classification and entity
+extraction when using the [RegexFeaturizer](../nlu/components#regexfeaturizer) component in the pipeline.
```
erohmensing, line 255:
```suggestion
extraction when using the [`RegexFeaturizer`](../nlu/components#regexfeaturizer) component in the pipeline.
```

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+2. They can be used to extract entities from a user message using the rule-based
+[`RegexEntityExtractor`](../nlu/components#regexentityextractor) component in the pipeline.
+
+For example, if your entity has a deterministic structure (like a
+zipcode or an email address), you can use a regular expression to ease
```
erohmensing, line 261:
```suggestion
zipcode or an email address), you can use a regular expression to improve
```

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+If you are using the regular expressions for the `RegexFeaturizer` the
+name of the regular expression does not matter. If does not define the
+entity nor the intent, it is a human readable description for you
+to remember what this regex is used for and is the title of the
+corresponding pattern feature.
+
+If you want to use the RegexFeaturizer you can also use the regex
+features to improve the intent classification performance, for example,
+by defining a greet clause:
```
erohmensing, line 285: Do you think it's worth it to split this into 2 parts: regex for entity extraction and regex for intent classification?

erohmensing: Then we can easily add info about L308-314, as it only applies to entity extraction and not to intent classification (as all intent classifiers use the features)

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+available regex features.
+
+:::note
+
+Regex features only define entities when used in combination with the `RegexEntityExtractor`. Otherwise they  
+don't define entities nor intents! They only provide patterns to
+help the classifier recognize entities and related intents. Hence, you
+still need to provide intent & entity examples as part of your
+training data!
+:::
```
erohmensing, line 313:
```suggestion
:::note

When using the `RegexFeaturizer`, a regular expression only provides a feature
that helps the model learn an association between intents/entities and inputs
that fit the regular expression. In order for the model to learn this association,
you must provide example inputs that include that regular expression! 
:::
```
Do you think this makes more sense? I'm trying to rephrase it as what to do when using features (versus when not using direct extraction)

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+If you are using lookup tables in combination with the RegexFeaturizer, there must be a few examples of matches  
+in your training data. Otherwise the model will not learn to use the
+lookup table match features.
+
+**warning**
+
+You have to be careful when you add data to the lookup table.  
+For example, if there are false positives or other noise in the table,
+this can hurt performance. So make sure your lookup tables contain
+clean data.
```
erohmensing, line 352: These aren't an actual note + warning

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+Lookup tables provide a convenient way to supply a list of entity
+examples. The format is as follows:
+
+```yaml
+- lookup: <lookup table name>
+  examples: |
+  - <an entity>
+  - <another entity>
+```
```
erohmensing, line 325: similar to synonyms, let's use a real example

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+The name of the lookup table is subject to the same constraints as the
+name of a regex feature.
+
+When lookup tables are supplied in training data, the contents are
+combined into a large regex pattern that looks for exact matches in the
+training examples. These regexes match over multiple tokens, so
+`lettuce wrap` would match `get me a lettuce wrap ASAP` as
+`[0 0 0 1 1 0]`.
```
erohmensing, line 334:
```suggestion
When you supply a lookup table in your training data, the contents of that table
are combined into one large regular expression. This regex is used to check
each training example to see if it contains matches for entries in the
lookup table.
```

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+The name of the lookup table is subject to the same constraints as the
+name of a regex feature.
+
+When lookup tables are supplied in training data, the contents are
+combined into a large regex pattern that looks for exact matches in the
+training examples. These regexes match over multiple tokens, so
+`lettuce wrap` would match `get me a lettuce wrap ASAP` as
+`[0 0 0 1 1 0]`. 
+
+These regexes are processed identically to the regular
```
erohmensing, line 336:
```suggestion
Lookup table regexes are processed identically to the
```

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+### Stories
+
+Stories are composed of:
+  
+  - name: The name is arbitrary and not used in training; 
+    you can use it as a human-readable reference for the story.
+  - metadata: arbitrary and optional, not used in training, 
+    you can use it to store relevant information about the story
+    like e.g. the author
+  - a list of ``steps``: The user messages and actions that make up the story
```
erohmensing, line 374: name: here is confusing because it makes me think that is the key. maybe we do it by keys? 

```
Stories include the following keys: 

  - `story` (required): The story's name. The name is not used in training; 
     use it as a human-readable reference for the story.
  - `metadata` (optional): Extra information about the story. Not used in training; 
     you can use it to store relevant information about the story, e.g. the author.
  - a list of `steps` (required): The user messages and actions that make up the story.

File: docs/docs/user-guide/training-data-format.mdx [hunk 37]
```diff
@@ -0,0 +1,659 @@
...
+  steps:
+  # list of steps
+  - intent: greet
+  - action: utter_greet
+```
+
+Each step can be one of the following:
+
+  - A [user message](#user-messages), represented by **intent** and **entities**.
+  - An [`or`](#or) statement, which includes two or more user messages under it
```
erohmensing, line 393:
```suggestion
  - An [or statement](#or), which includes two or more user messages under it
```
just to keep things consistent

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+Each file can contain one or more **keys** with corresponding training
+data. One file can contain multiple keys, as long as there is not more
+than one of a certain key in a single file. The available keys are:
+
+- `version`
+- `stories`
+- `nlu`
+- `rules`
+- `e2e_tests`
```
erohmensing, line 39: let's put these in the order they're explained. I'm wondering if it makes sense to not group stories and rules under `conversation training data`, and rather explain what each is at the beginning of its section? because they're on the same level as `nlu` and there's no reference to "conversation training data" here, maybe it would make sense? not sure though

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,659 @@
...
+e2e_tests:
+- user: |
+    hey
+  intent: greet
+- action: utter_greet
+- user: |
+    what language do you speak
+  intent: faq/language
+- action: respond_faq
+```
```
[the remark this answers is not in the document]
degiz: True, that can cause confusion. I can't think of a better solution atm though.

docs/docs/user-guide/training-data-format.mdx
erohmensing: Well lol it looks like we just moved away from that 😅

melindaloubser1: Just switched this to be one heading today @tmbo  suggestion - https://github.com/RasaHQ/rasa/pull/6357/commits/aba2e2e778c0ebc32bccfc5ddd2e89125dc38f4b

It kinda makes sense to group it because NLU is everything for an NLU model, and then conversational is everything for a dialogue management model.

melindaloubser1: Why can't I see your follow up comment from the files changed tab  🤔 anyway that's fine then.

docs/docs/user-guide/training-data-format.mdx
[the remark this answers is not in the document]
melindaloubser1: If it's a single string, can it just be provided as `- user: what language do you speak` ? Or does that also require the weird escaping?

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: the features also apply to entities though. How about 
```
#### Regular Expression Features

#### Regular Expressions for Rule-Based Entity Extraction
```
Or is that too long a subtitle?

melindaloubser1: Oh oh wait I see what you mean. Yes your version makes more sense.

docs/docs/user-guide/training-data-format.mdx
tmbo: mhm yes I think both will accept them, but I though only one of them will actually get called 🤔  - I'll close this comment though nothing we need to discuss here.

docs/docs/user-guide/training-data-format.mdx
tmbo: Well, so here is how it should work technically:
- we do have different training data specifications, each gets a version. first one is 2.0 - if at some point we need to change the training data schema, e.g. because we want to rename of a training data example, we'd increase the version number to 2.1. this is independent of the rasa open source version though. we should only change the version number if the change is not backwards compatible.
- the user specifies the specification their training data adheres to
- if the installed rasa open source version does not support the specified training data version, it should fail. if the user did not specify any format version, rasa open source should use the latest format it supports

tmbo: hopefully, we'll not need to make a lot of use of that version number

docs/docs/user-guide/training-data-format.mdx
tmbo: yes exactly, that makes more sense. metadata really is more an extension point for people that want to use their own or third party components that need additional data annotations

docs/docs/user-guide/training-data-format.mdx
tmbo: Let's keep it there and add some version of that here as well, maybe something along the lines of:
`Entities are structured pieces of information that can be extracted from a user's message. For the entity extraction to work, you need to either specify training data to train an ML model to recognize your entities or you need to define rules to extract entities based on a character pattern.`. not sure if it makes sense to include the ml / rules part, up to you

docs/docs/user-guide/training-data-format.mdx
tmbo: well, it will work but we'll throw a deprecation warning. if we mention both formats, we should at least differentiate them. e.g. adding a note that there is an additional format that is deprecated.

docs/docs/user-guide/training-data-format.mdx
tmbo: I think it should be according to the docusaurus docs: https://v2.docusaurus.io/docs/markdown-features/#referencing-other-documents

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: Did not know it'd throw a warning - should've checked. In that case makes sense to remove it.

tmbo: well, not sure it does already throw that warning. can you either check or create an issue + assign sasha to the issue?

docs/docs/user-guide/training-data-format.mdx
```diff
@@ -0,0 +1,655 @@
...
+
+
+Checkpoints can help simplify your training data and reduce redundancy in it,
+but **do not overuse them**. Using lots of checkpoints can quickly make your
+stories hard to understand. It makes sense to use them if a sequence of steps
+is repeated often in different stories, but stories without checkpoints
+are easier to read and write. 
+
+#### or
```
melindaloubser1, line 579:
```suggestion
#### OR statement
```

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: it does not throw a warning, created an issue https://github.com/RasaHQ/rasa/issues/6365

docs/docs/user-guide/training-data-format.mdx
melindaloubser1: oh wow did not even see that I misnamed the name. Thanks

docs/docs/user-guide/training-data-format.mdx
m-vdb: you don't have too, but it looks like it's preferred (the only benefit I see from now is that if we want to navigate the doc from github files, but in our case it wouldn't really work anyway).

tmbo:
@melindaloubser1 how are things going, anything still open or is this ready for a another review?

melindaloubser1:
Ready for another review. I don't know the answer to Ella's q about versions, if you do I'll clarify that

tmbo:
> Do you mean latest, or the version you have installed? I'd assume the latter as a version can't know about new versions?

It will use the latest data format supported by the currently installed rasa open source (which might not be the latest data format overall, as your rasa version might be outdated).

> Similar question about Training data files with a version greater... -- if you try to run version: 2.1 on rasa==2.0.0, it will still fail even if 2.1 is released, right?

Exactly. If you are trying to use a training data version that is not yet supported by your installation you likely need to update your installed rasa version.

tmbo:
@melindaloubser1 are these the questions you were referring to?

melindaloubser1:
Yes, thank you for clarifying! related question - where can users find information about these training data versions? I haven't come across a list yet

tmbo:
I don't think it makes sense to add a list until after we have multiple data versions that we support.

melindaloubser1:
Ok, I think a user wouldn't know what version to put right now though - should we include that "current training data version is 2.0" or something?

tmbo:
yes that is a good point 😅  - let's include that :+1:

melindaloubser1:
Added. It's a bit verbose but hopefully clear.

melindaloubser1:
Now that the new structure is in place, wdyt about splitting it up into overview-dialogue-nlu pages?

tmbo:
I'd rather want to merge this first, as I need it in e.g. the multimedia responses PR to add my changes. I think we can split it up in a separate PR (I think we can also discuss that in the channel or on an issue first)

tmbo:
@erohmensing do you want to give the PR another review?

docs/docs/training-data-format.mdx
```diff
@@ -1,355 +1,655 @@
...
+```yaml
+    steps:
+    - intent: celebrate_bot
+    - slot_was_set: feedback_value
+      value: positive
+    - action: utter_yay
 ```
 
+This means the story requires that the current value for the `feedback_value`
+slot be `positive` for the conversation to continue as specified.
```
erohmensing, line 529: We don't have to add slot steps for unfeaturized slots, right?

docs/docs/training-data-format.mdx
```diff
@@ -1,355 +1,655 @@
...
 :::note
-Composite entities are currently only supported by the [DIETClassifier](./components/intent-classifiers.mdx#dietclassifier) and [CRFEntityExtractor](./components/entity-extractors.mdx#crfentityextractor).
+Stories do not **set** slots. The slot must be set by an entity or custom action **before** the `slot_was_set` step.
+:::
+
+
+#### Checkpoints
+
+Checkpoints are ways to connect stories together. They can be either the first
+or the last step in a story. If they are the last step in a stories, that story
```
erohmensing, line 539:
```suggestion
or the last step in a story. If they are the last step in a story, that story
```

docs/docs/training-data-format.mdx
```diff
@@ -1,355 +1,655 @@
...
+  - intent: greet
+  - action: utter_greet_thanksgiving
+```
+
+
+Checkpoints can help simplify your training data and reduce redundancy in it,
+but **do not overuse them**. Using lots of checkpoints can quickly make your
+stories hard to understand. It makes sense to use them if a sequence of steps
+is repeated often in different stories, but stories without checkpoints
+are easier to read and write.
```
erohmensing, line 577: isn't this pretty much the use case for rules?

docs/docs/training-data-format.mdx
```diff
@@ -1,355 +1,655 @@
...
 :::
 
-## Normalizing Data
 
-<a aria-hidden="true" tabIndex="-1" className="anchor enhancedAnchor" id="entity-synonyms"></a>
+## Conversation Training Data
+
+Stories and rules are both representations of conversations between a user and a conversational
+assistant. They are used to train the dialogue management
+model. **Stories** are used to train a machine learning model to identify patterns
```
melindaloubser1, line 363: just a note -  the stories page hasn't been updated yet, it still shows markdown training data. We can link so long though
```suggestion
model. [**Stories**](stories.mdx) are used to train a machine learning model to identify patterns
```

docs/docs/training-data-format.mdx
melindaloubser1: It does seem like it. I'm actually not sure of a concrete example for checkpoints 🤔

docs/docs/training-data-format.mdx
```diff
@@ -1,355 +1,655 @@
...
+  steps:
+    - intent: <intent name> 
+      entities:
+      - <entity_name>: <entity value>
 ```
 
-:::note
-Please note that adding synonyms using the above format does not improve the model's classification of those entities.
-**Entities must be properly classified before they can be replaced with the synonym value.**
+Intent names can be prefixed with `/` or not; it makes no difference.
```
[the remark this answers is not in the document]
melindaloubser1: I'm taking this sentence out for now, I think it's unnecessary to mention and potentially confusing

docs/docs/training-data-format.mdx
```diff
@@ -1,355 +1,655 @@
...
-:::caution
-This feature is experimental.
-We introduce experimental features to get feedback from our community, so we encourage you to try it out!
-However, the functionality might be changed or removed in the future.
-If you have feedback (positive or negative) please share it with us on the [forum](https://forum.rasa.com).
+#### Actions
 
-:::
+All actions executed by the bot are specified with the `action:` key followed
+by the name of the action.
```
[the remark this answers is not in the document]
melindaloubser1: I'll add it for user message, slots, and checkpoints; for forms it's a bit complicated

docs/docs/training-data-format.mdx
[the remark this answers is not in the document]
erohmensing: Agreed 👍

docs/docs/training-data-format.mdx
melindaloubser1: No, but I think that's explained on the slots types part of the domain.

docs/docs/training-data-format.mdx
erohmensing: I guess the idea is if you have like a feedback form flow, you could use a checkpoint to make sure that after each of your use case flows, you go right into the feedback flow (without having to have the full feedback flow written down everywhere, making it easier if you change the feedback flow)? I feel like that's maybe something that'd be easier with checkpoints than rules... 🤔 

in other words maybe it's good for a sequence of steps that's repeated often, but started by the bot and not the user? but i also have no idea of rules and if you can accomplish this with them (why am i reviewing these docs again? lmao)

docs/docs/training-data-format.mdx
melindaloubser1: Format issue for slots though - I see two examples in notion:

```
    slot_was_set: 
    - context_scenario: holiday
    - holiday_name: thanksgiving
```
and
```
  - slot_was_set: feedback_value
    value: positive
```
In either case, what's the format for a text slot where value doesn't matter? Can it be just 
```
- slot_was_set: text_slot
```

docs/docs/training-data-format.mdx
melindaloubser1: That's true actually, like a sequence of follow up actions?

melindaloubser1: Hey I dunno what's up with rules yet either 😂

docs/docs/training-data-format.mdx
erohmensing: Mmm, no idea. @degiz ?

docs/docs/training-data-format.mdx
erohmensing: tenor

docs/docs/training-data-format.mdx
degiz: The parser won't read that:
```
slot_was_set: 
- context_scenario: holiday
- holiday_name: thanksgiving
```

The only way is to use atm is:
```
- slot_was_set: feedback_value
  value: positive
```
and you can omit the `value`!

melindaloubser1: Ok! I'll take it out of notion too then.

degiz: I see where the confusion comes from :( 
Checkpoints have the slots, and it follows that syntax:
```
- checkpoint: greet_checkpoint
    # This checkpoint should only apply if slot is set to the specified value
    slot_was_set: 
    - context_scenario: holiday
    - holiday_name: thanksgiving
```

melindaloubser1: oof. What if a checkpoint must be conditioned on a set slot with irrelevant value?

docs/docs/training-data-format.mdx
melindaloubser1: I think let's wait for clarification on rules & checkpoints and try again for a better explanation in the next round?

m-vdb:
@melindaloubser1 I wanted to check with you and see if we can wrap this up before the 2.0.0a2 release tomorrow. Are there any outstanding items that prevent the merge?

melindaloubser1:
I don't think so, all major points have been addressed, unless @erohmensing or @tmbo have anything final to add. Some outstanding stuff for next round, more so discussion points.

m-vdb:
great, so let's get this merged -- i'll let you add the `status:ready-to-merge` label

docs/docs/training-data-format.mdx
erohmensing: Sounds good to me

tmbo:
@melindaloubser1 was it intentional that you added this page? https://rasa.com/docs/rasa/next/nlu-training-data

melindaloubser1:
Yes, there's a discussion about it here https://rasa-hq.slack.com/archives/C016FLYV4QN/p1597319543009500

melindaloubser1:
We can remove most of it for now if you think that's better for the alpha, it's the concept content we didn't want to lose

PR #6358, opened 2020-08-07
check model is exits 

Dustyposa:
**Proposed changes**:
- Make sure model had been download when replace `agent`

### DIFF
--- a/rasa/server.py
+++ b/rasa/server.py
@@ -1030,11 +1030,13 @@ def create_app(
                     {"parameter": "model_server", "in": "body"},
                 )
 
-        app.agent = await _load_agent(
+        tmp_agent = await _load_agent(
             model_path, model_server, remote_storage, endpoints, app.agent.lock_store
         )
+        if tmp_agent.model_directory:
+            app.agent = tmp_agent
 
-        logger.debug(f"Successfully loaded model '{model_path}'.")
+            logger.debug(f"Successfully loaded model '{model_path}'.")
         return response.json(None, status=204)
 
     @app.delete("/model")


Dustyposa:
There is a bad problem when use the `/model` path to update the model.
When use:
```javascript
{
  "model_server": {
    "url": "https://hioashfioasf.tar.gz,
    "params": {},
    "headers": {},
    "basic_auth": {},
    "token": "string",
    "token_name": "string",
    "wait_time_between_pulls": 0
  }
}
```
If the url return `code 404`,  but the local model be replaced by a empty model.The server can't use old model.

rasa/server.py
```diff
@@ -1030,11 +1030,13 @@ async def load_model(request: Request) -> HTTPResponse:
                     {"parameter": "model_server", "in": "body"},
                 )
 
-        app.agent = await _load_agent(
+        tmp_agent = await _load_agent(
             model_path, model_server, remote_storage, endpoints, app.agent.lock_store
         )
+        if tmp_agent.model_directory:
```
Dustyposa, line 1036: I check the model in here, I can't make sure this position is the best position to do check. Have any advice?

sara-tagger:

erohmensing:
Hey @Dustyposa, to clarify, are you saying that the issue is: 
- If you try to replace the model from a URL that returns a 404, it will (incorrectly) say that it loaded the model and will replace it by an empty model 

Would the ideal behavior then be: 
- If you try to replace a model from a URL that returns a 404, it will not replace the model, but will return an error code indicating that the URL did not provide a model

Dustyposa:
> Hey @Dustyposa, to clarify, are you saying that the issue is:
> 
> * If you try to replace the model from a URL that returns a 404, it will (incorrectly) say that it loaded the model and will replace it by an empty model
> 
> Would the ideal behavior then be:
> 
> * If you try to replace a model from a URL that returns a 404, it will not replace the model, but will return an error code indicating that the URL did not provide a model

@erohmensing Yes.
I think shouldn't use an empty model when the model server return 404. 
I checked the result in path function, I can't make sure this position is the best position to avoid it.

tmbo:
I don't think this should be fixed in the server: the configuration you send to the server is valid, hence will be set and a 200 will be returned. The model will be set to an empty one, because fetching from the configured remote fails. This does seem to be expected.

Dustyposa:
> I don't think this should be fixed in the server: the configuration you send to the server is valid, hence will be set and a 200 will be returned. The model will be set to an empty one, because fetching from the configured remote fails. This does seem to be expected.

I think if you want to use an empty mode you that we can set a flag to do it.
First, we can't make sure the `server` which the model file in always work fine.
Second, the model file changed position is possible.
Most import, if users want to use an empty model,  it's use a empty flag like `to_be_empty = True` rather than use a `404` url ( and i need to check it and to confirm it it's a 404 path)

wochinge:
@tmbo Could you please follow up with @Dustyposa ?

tmbo:
The issue I have with the proposed fix is that it returns a 204 for an invalid configuration but then it doesn't actually update the agent on failure. The caller won't know that they are still using the old model. I agree, the current implementation isn't ideal either but at least the caller can be sure that we load the new model and fallback to an empty model on failure.

Issue #6338, opened 2020-08-04
Model regression tests - the CI fails if a configuration can't be found


tczekajlo:
The CI for model regression fails if a configuration can't be found. 

We should remove the step which fails if a comment with configuration can't be found and run matrix with tests only if `commend-id != ''`

PR #6375, opened 2020-08-11
Skip the model regression tests if a configuration doesn't exist

tczekajlo:
**Proposed changes**:
- Skip the model regression tests if a configuration doesn't exist https://github.com/RasaHQ/rasa/issues/6338
- Use the latest images for github runner
- Delete useless condition for the `remove_runner_gpu` job


tczekajlo:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]

```

.github/workflows/ci-model-regression.yml
```diff
@@ -87,6 +88,7 @@ jobs:
 
     - name: Update the comment with the configuration
       uses: peter-evans/create-or-update-comment@v1
+      if: steps.fc_config.outputs.comment-id != ''
```
wochinge, line 91: How about using https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error instead of all the the `if`s?

tczekajlo: Previously the workflow fails if a configuration couldn't be found, and it was ok, but a user saw that the workflow failed and it could be misleading for him. It's the main reason why I used `if`s, it helps to have green workflow and red only if something really bad happened 😄

tczekajlo:
> Why is it bad if the job fails / what's the advantage of skipping things?

@dakshvar22 Can you tell a few words what is your experience with running the model regression tests and why is bad if the workflow fails if a configuration comment can't be found?

dakshvar22:
So, the behaviour that I experienced was that as soon as I add the label `status:model-regression-tests`, the helper message like [this](https://github.com/RasaHQ/rasa/pull/6375#issuecomment-671858175) comes up. Post this the expected behaviour is that it should wait for the user to add a comment with the configuration to be run(only if no previous comment with such a configuration existed) but in my case it didn't wait for me to post the comment and the `CI - Model Regression / Reads tests configuration (pull_request)` step failed. As a user this was unexpected behaviour. Does that answer your question?

PR #6388, opened 2020-08-12
add links for legacy docs versions

m-vdb:
**Proposed changes**:
- Add links to v0 and v1 docs on home page
- Add navbar link to v1


m-vdb:
@lunelson i've configured the two domains in Netlify. It's the first time I do this, so let me know if I did anything wrong! On thing you mentioned on the issue was that I should setup a 301 redirect from legacy-docs to legacy-docs-v0. Where should I do this? At the moment, the netlify site is configured with both domains:
Screenshot 2020-08-12 at 09 57 45

lunelson:
#### Re: `legacy-docs-v0.rasa.com`

I've updated the Netlify project to set `legacy-docs-v0.rasa.com` is its primary domain. Some TODOs for this one:

- [ ] remove the search function from the top bar: these results are for Rasa-X and link to the main website
- [ ] set the Github link to link the correct repo/branch (right now it's just `rasahq`)
- [ ] create a `_redirects` file in the root of the deployment that will redirect the old subdomain to the new one, so that crawlers etc. will see that it's a permanent change (see below)

```sh
# /_redirects -- NB: http and https need separate rules if you don't force_ssl
http://legacy-docs.rasa.com/* https://legacy-docs-v0.rasa.com/blog/:splat 301!
https://legacy-docs.rasa.com/* https://legacy-docs-v0.rasa.com/blog/:splat 301!
```

#### Re: `legacy-docs-v1.rasa.com`

Some similar TODOs:

- [ ] set the Github link to the v1 branch of `rasahq/rasa`
- [ ] (as above) remove the search function from the top bar: these results are for Rasa-X and link to the main website
- [ ] remove all other top bar links, since they are now `404`s

Some question:
- is this move to a "legacy" subdomain also going to happen with the Rasa-X docs? or will they just update in place?
- could we rename the Netlify project for greater clarity? See below.

|old name|new name|
|--|--|
|priceless-jang-11b3ee|rasahq-docs-v0|
|rasahq-docs-rasa|rasahq-docs-rasa-v1|
|rasahq-docsv2-rasa|rasahq-docs-rasa-v2|

m-vdb:
thanks @lunelson !

About the TODOs: I'm not sure there is an easy way to do that:
- v0:
  - GH link seems like a feature, not a bug (it's currently like that on the v1 site)
  - I don't know if there is a way to customize the deployed site. Is there? I found [this repo](https://github.com/RasaHQ/legacy-docs-deployment) but it's rather empty. I would say that the value brought to users is not worth the investment.
- v1:
  - GH link seems like a feature, not a bug (it's currently like that)
  - I'm not sure I can update the search easily now, we can wait for 2.0 release I think (same for top bar links)


> is this move to a "legacy" subdomain also going to happen with the Rasa-X docs? or will they just update in place?

No, I think it will be in place, since we're not switching to a new major version in Rasa X.

> could we rename the Netlify project for greater clarity? See below.

sure, but it means we need to update the redirects in [`rasa-website/netlify.toml`](https://github.com/RasaHQ/rasa-website/blob/master/netlify.toml). I don't think we should do that now, but rather wait for the full release of the docs?

WDYT? I'd like @tmbo input on this too :)

tmbo:
sounds good :+1: and yes no need for legacy rasa x atm, docs won't change to much apart from styling

lunelson:
OK let's pin some of these tasks though, for post-2.0 release yes?

m-vdb:
yep, I'll create them @lunelson . In the meantime, could you approve the PR? 😄

PR #6401, opened 2020-08-13
removing attribute from process

evgeniiaraz:
**Proposed changes**:
- This is a cosmetic change of removing `attribute` from process and rather tokenizing/featurizing all the applicable features which are present in a `Message` 


evgeniiaraz:
/modeltest
```
include:
 - dataset: ["Sara", "Hermit", "Private 2"]
   config: ["Sparse + DIET(bow) + ResponseSelector(bow)", "ConveRT + DIET(bow) + ResponseSelector(bow)", "Sparse + Spacy + DIET(seq) + ResponseSelector(t2t)"]
```

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -212,21 +212,18 @@ def train(
...
-        **kwargs: Any,
+        self, message: Message, *, tf_hub_module: Any = None, **kwargs: Any,
     ) -> None:
-        if attribute not in DENSE_FEATURIZABLE_ATTRIBUTES:
-            return
 
-        sequence_features, sentence_features = self._compute_features(
-            [message], tf_hub_module, attribute=attribute
-        )
+        for attribute in DENSE_FEATURIZABLE_ATTRIBUTES:
```
wochinge, line 218: I think it could make sense to add function `get_featurizable_attributes` to the parent class `Featurizer` and then overwrite it in `DenseFeaturizer` and `SparseFeaturizer`. This way the solution is DRY

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -48,13 +48,11 @@ def _get_doc(self, message: Message, attribute: Text) -> Any:
...
 
-    def process(self, message: Message, attribute: Text = TEXT, **kwargs: Any) -> None:
+    def process(self, message: Message, **kwargs: Any) -> None:
         """Sets the dense features from the language model doc to the incoming
         message."""
-        if attribute not in DENSE_FEATURIZABLE_ATTRIBUTES:
-            return
-
-        self._set_lm_features(message, attribute)
+        for attribute in DENSE_FEATURIZABLE_ATTRIBUTES:
```
wochinge, line 54: do we need to check `if message.get(attribute)`?

rasa/nlu/featurizers/dense_featurizer/mitie_featurizer.py
```diff
@@ -71,17 +71,18 @@ def process_training_example(
...
-            return
-
-        mitie_feature_extractor = self._mitie_feature_extractor(**kwargs)
-        tokens = message.get(TOKENS_NAMES[attribute])
-        sequence_features, sentence_features = self.features_for_tokens(
-            tokens, mitie_feature_extractor
-        )
+    def process(self, message: Message, **kwargs: Any) -> None:
+        for attribute in DENSE_FEATURIZABLE_ATTRIBUTES:
+            mitie_feature_extractor = self._mitie_feature_extractor(**kwargs)
```
wochinge, line 76: this line should be outside the loop. No need to instantiate it every iteration.

rasa/nlu/featurizers/dense_featurizer/spacy_featurizer.py
```diff
@@ -60,11 +60,9 @@ def train(
     def get_doc(self, message: Message, attribute: Text) -> Any:
         return message.get(SPACY_DOCS[attribute])
 
-    def process(self, message: Message, attribute: Text = TEXT, **kwargs: Any) -> None:
-        if attribute not in DENSE_FEATURIZABLE_ATTRIBUTES:
-            return
-
-        self._set_spacy_features(message, attribute)
+    def process(self, message: Message, **kwargs: Any) -> None:
```
wochinge, line 63: @tabergma I have general question. Every tokenizer computes features and adds this features to the message. While the computation of the features and the setting might be individual for every tokenizer, the implementation of `process` should always be the same (probably also for `train` right?). Would it make sense from your perspective to introduce the 2 methods `get_features` and `set_features` in the `Featurizer` parent class?

wochinge: (wouldn't do it as part of this PR if it even makes sense)

rasa/nlu/model.py
```diff
@@ -406,5 +406,5 @@ def parse_message(self, message: Message, attribute: Text = TEXT) -> Message:
```
wochinge, line 406: the method name `parse_message` doesn't really make sense. Should we rename to something like `featurize_message`?

rasa/nlu/tokenizers/spacy_tokenizer.py
```diff
@@ -34,6 +34,8 @@ def get_doc(self, message: Message, attribute: Text) -> "Doc":
 
     def tokenize(self, message: Message, attribute: Text) -> List[Token]:
         doc = self.get_doc(message, attribute)
```
wochinge, line 36: the return type of `Doc` should be `Optional["Doc"]`, shouldn't it?

rasa/nlu/featurizers/dense_featurizer/spacy_featurizer.py
tabergma: I think you are mixing tokenizers and featurizers in your comment. We have this for the tokenizer. I don't think it makes sense for the featurizers as some of them, for example, need to generate a vocabulary first in train before setting features. Also not all featurizers are handling the same kind of text messages, e.g. text vs intent vs response vs action name. But maybe `set_features` makes sense. So I think we might be able to extract some useful helper methods to the super class but I would not implement `process` or `train` over there.

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
evgeniiaraz: I looked at it and it doesn't really make sense to have it for SparseFeaturizers, as each of them has a different set of featurize they go through. So it would only be for the `DenseFeaturizer` which boils down to the same thing, `get_featurizable_attributes` will  just return `DENSE_FEATURIZABLE_ATTRIBUTES`.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
evgeniiaraz: No, because it is done in `_set_lm_features`

rasa/nlu/tokenizers/spacy_tokenizer.py
evgeniiaraz: Yep, changed it!

rasa/nlu/model.py
evgeniiaraz: It makes sense, but I would change it later once we settle on how it is actually used within Core.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: I was wondering because the `convert_featurizer` does it as part of `process` 👍

PR #6403, opened 2020-08-13
Correct path for Poetry cached libraries on Windows

alwx:
**Proposed changes**:
- update CI to include the correct path to Poetry cached libraries on Windows


m-vdb:
@alwx i've re-run the build to see if the cache is picked up (because I didn't see it in the last build)

m-vdb:
It doesn't look like it's loading the cache :/

m-vdb (dismissed):
doesn't seem to work in CI :/

alwx:
It took a while to finish it, but now cache works for both Windows and Ubuntu!

pyproject.toml
```diff
@@ -95,10 +95,10 @@ pika = "~1.1.0"
 jsonschema = "~3.2"
 packaging = ">=20.0,<21.0"
 pytz = ">=2019.1,<2021.0"
-rasa-sdk = "^2.0.0a4"
+rasa-sdk = "^2.0.0a5"
```
wochinge, line 98: why do we need to change that?

tests/core/training/converters/test_story_markdown_to_yaml_converter.py
```diff
@@ -1,15 +1,23 @@
...
+from sanic.request import Request
 
 from rasa.core.training.converters.story_markdown_to_yaml_converter import (
     StoryMarkdownToYamlConverter,
 )
 from rasa.shared.constants import LATEST_TRAINING_DATA_FORMAT_VERSION
 
 
+@pytest.yield_fixture(scope="session")
+def event_loop(request: Request) -> Iterator[asyncio.AbstractEventLoop]:
```
wochinge, line 15: why is this needed? If the scope is `session` it should go into `conftest` imo. If this just for the module, then the scope should be `module`

.github/workflows/continous-integration.yml
```diff
@@ -139,12 +139,18 @@ jobs:
...
+
     - name: Load Poetry Cached Libraries ⬇
+      if: matrix.os == 'windows-latest'
       uses: actions/cache@v1
       with:
-        path: ~/.cache/pypoetry/virtualenvs
-        key: ${{ runner.os }}-poetry-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
-        restore-keys: ${{ runner.os }}-poetry-${{ matrix.python-version }}
+        path: .venv
+        key: ${{ runner.os }}-poetry-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}-4
```
wochinge, line 152: what is the `-4` for?

.github/workflows/continous-integration.yml
```diff
@@ -156,26 +162,28 @@ jobs:
 
     - name: Install Dependencies (Windows) 📦
       if: matrix.os == 'windows-latest'
+      # Restoring cache doesn't work properly on Windows due to symlinks.
+      # That's why we're manually removing symlinks here and everything that is inside `$spacy_data_dir` directory.
+      # More information: https://github.com/actions/cache/issues/120
       run: |
-        make install-full-windows
+        $spacy_data_dir = ".venv\lib\site-packages\spacy\data"
+        if (Test-Path $spacy_data_dir) {
```
wochinge, line 170: no clue about the Windows scripting, but it looks like we only delete symlinks within `$spacy_data_dir`? The comment suggests that we do that for all of `.venv`

.github/workflows/continous-integration.yml
```diff
@@ -139,12 +139,18 @@ jobs:
...
+        poetry config virtualenvs.in-project true
+
     - name: Load Poetry Cached Libraries ⬇
+      if: matrix.os == 'windows-latest'
       uses: actions/cache@v1
       with:
-        path: ~/.cache/pypoetry/virtualenvs
-        key: ${{ runner.os }}-poetry-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
-        restore-keys: ${{ runner.os }}-poetry-${{ matrix.python-version }}
+        path: .venv
```
wochinge, line 151: Is `.venv` the default for poetry? In their docs they say the default virtualenv path is `{cache-dir}\virtualenvs` 🤔

.github/workflows/continous-integration.yml
```diff
@@ -139,12 +139,18 @@ jobs:
       with:
         poetry-version: ${{ env.POETRY_VERSION }}
 
+    - name: Set up virtual environment (Windows)
+      if: matrix.os == 'windows-latest'
+      run: |
+        poetry config virtualenvs.in-project true
```
wochinge, line 145: Let's add a comment why we are doing this.

.github/workflows/continous-integration.yml
```diff
@@ -139,12 +139,18 @@ jobs:
       with:
```
wochinge, line 139: general question: Do we actually need to install `wget`? Apparently it's already included https://github.com/agda/agda/issues/4691

wochinge (commented):
Thanks for fixing this! My comments are mostly questions for my own understanding 😆

pyproject.toml
alwx: I updated sanic in the latest version of rasa-sdk, and the new version of sanic has better Windows support. We had a discussion about it while you were on vacation :)

.github/workflows/continous-integration.yml
alwx: 👍

.github/workflows/continous-integration.yml
alwx: Oh that's a mistake (just because I was testing cache, will update in a sec)

.github/workflows/continous-integration.yml
alwx: It's default when you use set `virtualenvs.in-project` to `true`. I will add a comment

.github/workflows/continous-integration.yml
alwx: Did that.

.github/workflows/continous-integration.yml
alwx: Might be true. It's there because I haven't found it on this list (https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md) and because it was in our `Makefile` before.
I will try removing it.

alwx: Apparently, no, it doesn't work: https://github.com/RasaHQ/rasa/pull/6403/checks?check_run_id=1123537533

data/test_config/config_empty_en_after_dumping_windows.yml
```diff
@@ -2,7 +2,7 @@ language: en
 pipeline:
 # # No configuration for the NLU pipeline was provided. The following default pipeline was used to train your model.
 # # If you'd like to customize it, uncomment and adjust the pipeline.
-# # See https://rasa.com/docs/rasa/model-configuration for more information.
+# # See https://rasa.com/docs/rasa/tuning-your-model for more information.
```
wochinge, line 5: 🚀

pyproject.toml
wochinge: I saw it but wouldn't it automatically pick the latest sdk version and latest sanic version (without changing the minimal alpha version)?

.github/workflows/continous-integration.yml
wochinge: thanks for trying it out 👍

wochinge (approved):
Can we re-run the workflow once? The last run didn't use the cache 😬

.github/workflows/continous-integration.yml
```diff
@@ -139,13 +139,34 @@ jobs:
...
+      if: matrix.os == 'windows-latest'
+      # Poetry on Windows cannot pick up the virtual environments directory properly,
+      # and it creates a new one every time the pipeline runs.
+      # We want to  allow Poetry to create a venv directory if it doesn't
+      # exist, but also pick it up if it's already there.
+      # This step solves this problem — it tells poetry to always use `.venv` directory inside
+      # the project itself, which also makes it easier for us to determine the correct directory
+      # that needs to be cached.
+      run: |
+        poetry config virtualenvs.in-project true
```
m-vdb, line 152: @alwx @wochinge wondering if we should do this for both ubuntu and windows, for the sake of keeping it simple 🤔

PR #6405, opened 2020-08-14
improved getting started docs

tmbo:
**Proposed changes**:
- remove leading `$ ` from bash code blocks (as it gets copied when you click the copy button which is very annoying). if we feel we need it, it should be implemented in the highlighter for bash as a displayed but not copied char
- made the docs landing page a little more useful with a pointer where to go next
- cleaned up the installation page
   - removed TOC (not necessary, we got a sidebar now)
   - added a hint to run `rasa init` after installation or use prototyping page
   - use groups for the tab sections (selecting "MacOS" once will show you macos for all tab groups)
   - always use `pip3` and `python3` as it is a best practice
   - made extra dependencies headlines more useful in sidebar

### DIFF
--- a/docs/docs/installation.mdx
+++ b/docs/docs/installation.mdx
@@ -9,39 +9,42 @@ import TabItem from '@theme/TabItem';
 
 ## Quick Installation
 
-You can install Rasa Open Source using pip (requires Python 3.6 or 3.7).
+You can install Rasa Open Source using pip (requires Python 3.6, 3.7 or 3.8).
 
+<!-- TODO: REMOVE VERSION NUMBER FOR RASA 2.0 RELEASE -->
 ```bash
-$ pip3 install rasa
+pip3 install rasa==2.0.0a2
 ```
 
-* Having trouble installing? Read our [step-by-step installation guide](./installation.mdx#installation-guide).
+You are now ready to go! So what's next?
+You can create a new project by running
 
-* You can also [build Rasa Open Source from source](./installation.mdx#build-from-source).
+```bash
+rasa init
+```
 
-* For advanced installation options such as building from source and installation instructions for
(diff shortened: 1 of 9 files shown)


docs/docs/installation.mdx
```diff
@@ -245,7 +246,7 @@ anywhere. If you want to use MITIE, you need to
 tell it where to find this file (in this example it was saved in the
 `data` folder of the project directory).
 
-:::caution
+:::caution Deprecation Warning
 Mitie support is likely to be deprecated in a future release.
```
erohmensing, line 250: Is this actually true? we keep saying so and then deciding to keep it around

docs/docs/messaging-and-voice-channels.mdx
```diff
@@ -12,7 +12,7 @@ in a `credentials.yml` file.
 An example file is created when you run `rasa init`, so it's easiest to edit that file
 and add your credentials there. Here is an example with Facebook credentials:
 
-```yaml
+```yaml title="credentials.yml"
```
erohmensing, line 15: Very cool, didn't know we could do this 😎

docs/docs/installation.mdx
tmbo: I mean you got a very valid point there, but as far as I remember there is no alternative for certain languages (was it Japanese or Cantonese 🤷 ). let's poll the channel.

docs/docs/messaging-and-voice-channels.mdx
tmbo: yes it's a lot nicer than our usual: "... into your `xyz.yaml` put: "

docs/docs/installation.mdx
tmbo: I'll remove it for now, until we have a plan to deprecate it.

PR #6419, opened 2020-08-14
Fix the EvaluationStore to match prediction and target entities

davidezanella:
**Proposed changes**:
- Fix the EvaluationStore serialise method: the prediction and target entities were not aligned. Previously, they were aligned only by adding 'None' at the end of the smaller list instead of looking where they should be placed.

Practical example:
```
target entities: [
   {'text': 'hi, how are you', 'start': 0, 'end': 2, entity: 'verb' },
   {'text': 'hi, how are you', 'start': 4, 'end': 7, entity: 'noun' }
]
predicted entities: [
   {'text': 'hi, how are you', 'start': 4, 'end': 7, entity: 'noun' }
]
```

Adding 'None' at the end, as done originally, leads to wrong metrics because even the right prediction was marked as wrong.


sara-tagger:

davidezanella:
@tabergma can you review this pr?

degiz:
Hey @dakshvar22 

Thanks a lot for the PR! Could you please also write a unit for the change? I believe you can use existing `tests/test_test.py` file.

degiz (changes requested):
Hey @davidezanella

Thanks a lot for the PR! Could you please also write a unit for the change? I believe you can use existing `tests/test_test.py` file.

tabergma:
@davidezanella Thanks for addressing this. I think your concern is valid and should be fixed, but I am not sure if your method fixes the complete problem. Currently you just iterating over the gold entities to see if there are matching predicted entities and you are adding `None` in case no matching entity could be found. That way we ensure that the size of the gold end predicted entities is the same. However, i think we also need to consider the case when we have more predicted entities than gold entities. For example, you have a sentence like "Tanja is currently in Munich, but she lives in Berlin" and the annotator made a mistake and just annotate "Tanja" as `person`, but forgot to tag "Munich" and "Berlin" as `city`. However, the model predicts all three of them. If I understand your code correctly, you would just consider the gold entity "Tanja" and you would ignore the two additional entities "Munich" and "Berlin", which is not correct as those are false positive predictions in this case. What do you think?

davidezanella:
Hey @tabergma, you are right! I've fixed the code and the tests. 
Let me know what do you think about it

degiz (commented):
@tabergma will review the code.

tabergma:
@davidezanella Sorry for the late reply, I was offline a couple of days. Will take a look at the PR today. Can you please resolve the conflicts in the meantime? Thanks.

rasa/core/test.py
```diff
@@ -105,26 +104,104 @@ def has_prediction_target_mismatch(self) -> bool:
 
     def serialise(self) -> Tuple[List[Text], List[Text]]:
```
tabergma, line 105: This method is now very long. Can you please split it up? We could extract the entity prediction calculation into a separate method.

rasa/core/test.py
```diff
@@ -105,26 +104,104 @@ def has_prediction_target_mismatch(self) -> bool:
...
         """Turn targets and predictions to lists of equal size for sklearn."""
+        # sklearn does not cope with lists of unequal size, nor None values
```
tabergma, line 107: This comments seems to be a bit unrelated. I would either add it to the doc string or remove it.

rasa/core/test.py
```diff
@@ -105,26 +104,104 @@ def has_prediction_target_mismatch(self) -> bool:
...
+            )
+            entity_predictions = sorted(
+                filter(lambda x: x.get("text") == text, self.entity_predictions),
+                key=lambda x: x.get("start"),
+            )
+
+            i_pred = 0
+            i_target = 0
+
+            def compare_entities():
```
tabergma, line 135: I would move this method out of the method `serialise` and make it private.

tabergma: Also can you please add some comments to the methods to make it even more clear what cases we check.

rasa/core/test.py
```diff
@@ -105,26 +104,104 @@ def has_prediction_target_mismatch(self) -> bool:
...
+                return 1 if target else -1
+
+            while i_pred < len(entity_predictions) or i_target < len(entity_targets):
+                cmp = compare_entities()
+                if cmp == -1:  # predicted comes first
+                    entity_predictions_fixed.append(
+                        TrainingDataWriter.generate_entity(
+                            entity_predictions[i_pred].get("text"),
+                            entity_predictions[i_pred],
+                        )
```
tabergma, line 169: This could be extracted into a helper method.

rasa/core/test.py
```diff
@@ -105,26 +104,104 @@ def has_prediction_target_mismatch(self) -> bool:
...
+            list(
+                set(
+                    [e.get("text") for e in self.entity_targets]
+                    + [e.get("text") for e in self.entity_predictions]
+                )
+            )
         )
+
+        entity_targets_fixed = []
+        entity_predictions_fixed = []
```
tabergma, line 119: Can we rename them to `aligned_entity_targets` and `aligned_entity_predictions`?

tabergma (changes requested):
Looks great 🚀  Added a few comments. Can you please also add a [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) entry? Thanks.

rasa/core/test.py
```diff
@@ -106,28 +105,108 @@ def has_prediction_target_mismatch(self) -> bool:
...
+            entity_targets = sorted(
+                filter(lambda x: x.get("text") == text, self.entity_targets),
+                key=lambda x: x.get("start"),
+            )
+            entity_predictions = sorted(
+                filter(lambda x: x.get("text") == text, self.entity_predictions),
+                key=lambda x: x.get("start"),
+            )
+
+            i_pred, i_target = 0, 0
```
tabergma, line 174: Minor: We try to not use abbreviations in names. I would rename this to `index_prediction` and `index_target`.

rasa/core/test.py
```diff
@@ -106,28 +105,108 @@ def has_prediction_target_mismatch(self) -> bool:
...
+
+            while i_pred < len(entity_predictions) or i_target < len(entity_targets):
+                cmp = self._compare_entities(
```
tabergma, line 177: Minor: I would rename this to `comparison_result`.

rasa/core/test.py
```diff
@@ -106,28 +105,108 @@ def has_prediction_target_mismatch(self) -> bool:
             or self.action_predictions != self.action_targets
         )
 
+    @staticmethod
+    def _compare_entities(
+        entity_predictions: List[Dict[Text, Any]],
+        entity_targets: List[Dict[Text, Any]],
+        i_pred: int,
```
tabergma, line 112: Minor: Rename to `index_prediction`.

rasa/core/test.py
```diff
@@ -106,28 +105,108 @@ def has_prediction_target_mismatch(self) -> bool:
...
+        i_target: int,
```
tabergma, line 113: Minor: Rename to `index_target`.

tabergma (approved):
Looks great 💯  Thanks for fixing this!

tabergma:
@degiz I think you need to update your review otherwise we are not able to merge this PR.

degiz (approved):

Issue #6409, opened 2020-08-14
Rename `Form` to `Loop` event


wochinge:
As part of the `RulePolicy` refactoring, we introduced the concept of a `Loop` which represents conversation parts which are
* unfeaturized to ML policies
* which are driven by hardcoded rules / `if` and `else`s 

In contrast `Form`s to request information from the user and fill required slots.

This means every `Form` is a `Loop`, but not every `Loop` is a `Form`. E.g. the `TwoStageFallback` was implemented using the `Loop` concept but it's not a Form.

- [x] Rename the `Form` event to `Loop` event
- [x] make sure that old `Form` events can still be retrieved from the tracker stores
- [x] adapt Rasa X frontend
- [ ] ~adapt Rasa X backend `active_form`~ not required (the tests which involve this are commented out anyway)
- [x] [enable Rasa SDK to parse `active_loop` events](https://github.com/RasaHQ/rasa-sdk/pull/246)

wochinge:
@mjaycub Do you know if there is any special handlings for Rasa Open Source `form` events in the frontend? `form` events are events like `user` or `bot`. Just a different type. I did a text search but couldn't find anything.

wochinge:
Found these lines in the frontend, but that seems easy to handle: 
https://github.com/RasaHQ/rasa-x/blob/d7bc3df310f3804bbb87e9196857917ac776b17f/src/rasa-frontend/src/state/messages/reducer.ts#L81

mjaycub:
@wochinge I can have a look tomorrow if you'd like, today is blocked up by IVC work

wochinge:
That would be great. It's not critical but needs to be done for 2.0 compatibility.

mjaycub:
@wochinge it doesn't look like we do anything special with `form` events, and we don't ever seem to use `active_form` at the moment. I think all we'll need to do is add a type for `LoopEvent` and make sure we update the `isEvent` check in `ConversationStoryViewer` ([here](https://github.com/RasaHQ/rasa-x/blob/6165b9e14854b419da4ef98646939588170de7e9/src/rasa-frontend/src/components/ConversationStoryViewer.tsx#L28)) to include a check for `loop`.

Will I open a PR with these changes?

wochinge:
> Will I open a PR with these changes?

I don't know. Will you? 😄  I'd hold off a bit - could be that also rename `FormValidation`. I'll give you a heads up 👍

mjaycub:
[...]

PR #6439, opened 2020-08-17
rename `Form` event to `ActiveLoop`

wochinge:
**Proposed changes**:
- fix the Rasa Open Source part of https://github.com/RasaHQ/rasa/issues/6409 (rasa-sdk needs still to be adapted)
- rename the `Form` event to `ActiveLoop` to reflect that it no longer only applies to forms (which are used for slot filling), but also to any other conversation parts which should be unfeaturized for Core policies


wochinge:
@akelad Do you have any concerns about this? I'll do a PR for rasa-sdk, too.

akelad:
hmm should the event itself be called `ActiveLoop` or just `Loop`? I would be for `Loop`, given that the `Loop` only gets activated once it gets triggered

wochinge:
Should we also rename `FormValidation`? It somewhat stays very form specific (though it's applied to all `Loops`).

erohmensing:
Since I don't know much about `Loop`s yet (I didn't know they could be anything but forms tbh) I think it makes sense if @akelad approves this one. 

> Rasa Open Source can handle old form events as to date. However, when dumping a tracker it will replace any form events with the new active_loop representation.

Will Rasa X still read both? Or will the event consumer also take into account that either could appear (when e.g. `rasa export`ing a tracker) -- or maybe that also includes a new tracker dump

wochinge:
> Will Rasa X still read both? Or will the event consumer also take into account that either could appear (when e.g. rasa exporting a tracker) -- or maybe that also includes a new tracker dump

It has to support reading both imo. The required changes for Rasa X are listed in the issue.

wochinge:
> hmm should the event itself be called ActiveLoop or just Loop? I would be for Loop, given that the Loop only gets activated once it gets triggered

I changed it to `ActiveLoop` so it's in line with our training data. There we also say `active_loop: null` to deactivate loops.

akelad:
i'll look at this tomorrow if that's ok. I need to look into the new implementation still, but in the old version the `Form` didn't just activate the form, it also asked for all the required slots - is that no longer the case?

wochinge:
> but in the old version the Form didn't just activate the form, it also asked for all the required slots - is that no longer the case?

The `FormAction` still asks for the slots. 
The `Form` / `Loop` **event** marks which form is active / not active. 

Feel free to give me a call then 👍

rasa/core/training/story_reader/yaml_story_reader.py
```diff
@@ -368,7 +368,7 @@ def _parse_action(self, step: Dict[Text, Any]) -> None:
         self._add_event(action_name, {})
 
     def _parse_form(self, form_name: Optional[Text]) -> None:
```
degiz, line 370:
```suggestion
    def _parse_active_loop(self, active_loop_name: Optional[Text]) -> None:
```

rasa/core/training/story_writer/yaml_story_writer.py
```diff
@@ -124,7 +124,9 @@ def story_contains_forms(story_step) -> bool:
             `True` if the `story_step` contains at least one form action,
```
degiz, line 124: Ok, this I'll fix as a separate commit, once this PR is merged 😄

rasa/core/events/__init__.py
```diff
@@ -1167,6 +1166,23 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
...
+    """
+
+    type_name = "form"
+
+    def as_dict(self) -> Dict[Text, Any]:
+        d = super().as_dict()
+        # Dump old `Form` events as `ActiveLoop` events instead of keeping the old
+        # event type.
+        d["event"] = ActiveLoop.type_name
+        return d
```
akelad, line 1183: should we have a deprecation warning for this event though?

rasa/core/events/__init__.py
```diff
@@ -1167,6 +1166,23 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
         tracker.change_form_to(self.name)
```
akelad, line 1166: should that be `change_active_loop_to` now?

rasa/core/events/__init__.py
```diff
@@ -1167,6 +1166,23 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
...
+    def as_dict(self) -> Dict[Text, Any]:
+        d = super().as_dict()
+        # Dump old `Form` events as `ActiveLoop` events instead of keeping the old
+        # event type.
+        d["event"] = ActiveLoop.type_name
+        return d
+
+
 class FormValidation(Event):
     """Event added by FormPolicy to notify form action
```
akelad, line 1187: we should remove reference to the formpolicy here for sure

akelad (commented):
ok the rename to ActiveLoop definitely makes sense to me now, thanks for explaining Tobias! I noticed a few references to forms still, and also the mention of the `FormPolicy`
As for whether `FormValidation` should be renamed - that's tricky. You say it's called by every loop right? Where is the distinction between a Loop and a Form generally speaking?

wochinge:
Thanks for your reviews! Will wrap it up later today.

> As for whether FormValidation should be renamed - that's tricky. You say it's called by every loop right? Where is the distinction between a Loop and a Form generally speaking?

I explained it a bit in https://github.com/RasaHQ/rasa/issues/6409.

As part of the `RulePolicy` refactoring, we introduced the concept of a `Loop` which represent conversation parts which are
* unfeaturized to ML policies
* which are driven by hardcoded rules / `if` and `else`s 

In contrast `Form`s to request information from the user and fill required slots.

This means every `Form` is a `Loop`, but not every `Loop` is a `Form`. E.g. the `TwoStageFallback` was implemented using the `Loop` concept but it's not a Form.

Does that make sense? 

Just renaming to `LoopValidation` doesn't really make sense because disabling validation doesn't really make sense for other actions which implement `Loop`. It's actually indicating that a `Loop` was triggered after handling an unhappy path, isn't it? So maybe something like `LoopUnhappyPath`?

wochinge:
If we do the re-name of `FormValidation`, I'd do it in a separate PR so this one doesn't get too huge / complex.

rasa/core/events/__init__.py
wochinge: I wouldn't because users can't do much about this without doing a database migrations. If there are old `Form` events in the database there would be a warning whenever we'd interact with them (e.g. loading the tracker or return it via endpoint).

rasa/core/events/__init__.py
wochinge: Kept it and added `RulePolicy` as this is still true

rasa/core/training/story_writer/yaml_story_writer.py
wochinge: ❤️

wochinge:
Should we also rename the `FORM_PREFIX`? 🙈 This change has far too many implications 🙈

rasa/core/events/__init__.py
wochinge: great catch!

akelad:
ok should we create an issue for FormValidation and discuss there? I already get the general gist of Forms vs Loops, but why FromValidation also runs on Loops is a separate question for me. 
As for `FORM_PREFIX`... for completeness we probably should also rename it 🙈

rasa/core/events/__init__.py
akelad: wait but i thought that `FormPolicy` is dead?

rasa/core/events/__init__.py
akelad: ok fair

rasa/core/events/__init__.py
wochinge: It's deprecated, not dead.

wochinge:
> ok should we create an issue for FormValidation and discuss there?

Sounds good 👍 

I'd also create a separate issue for `FORM_PREFIX` then, ok? Then we have nice small PRs. Can you please approve then?

PR #6447, opened 2020-08-18
Rasa CLI plugins

wochinge:
**Proposed changes**:
- support plugins for the CLI ala [kubernetes](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/)
- `rasa plugin`  will list all available plugins. 
- Demo Video: https://gfycat.com/somberhonoredgonolek
- implementation of https://www.notion.so/rasa/Rasa-CLI-plugins-c4beb36ca14d4a38b99bf2a0ec581857#3af7313ece7b4bebb571f49b004dc792

**How you can try it out**
1. Create an executable `rasa-{some-name}`
    E.g. 
    ```bash
    #!/bin/bash

   echo "Hi, I am a Rasa CLI plugin 👍🏻"
    ```
2. Make sure this executable is in your `PATH` (if the executable is in your current PATH: `export PATH=$PATH:$(pwd)`
3. Run `rasa {some-name}`

### DIFF
--- a/docs/docs/command-line-interface.mdx
+++ b/docs/docs/command-line-interface.mdx
@@ -393,3 +393,80 @@ The following arguments are available for `rasa x`:
 
 ```text [rasa x --help]
 ```
+
+## Command Line Plugins
+
+You can extend the Rasa Open Source command line interface with custom plugins.
+Plugins extending the `rasa` command line interface can implement new or custom
+behavior which is not included within Rasa Open Source by default.
+
+:::note
+It is currently not possible to overwrite existing `rasa` commands. For example it is
+not possible to use a plugin `rasa-train` to override the `rasa train` command.
+:::
+
+### Installing a Plugin
+
+A command line plugin is nothing more than an executable file, whose name begins with
+`rasa-`. Once it's installed, it is available on the command line via
+`rasa <plugin name>`.
+
+To make the plugin discoverable, it has to be in your `PATH` environment. You can see
+all potential directories in your `PATH` by running `echo $PATH`.
+Assuming that `/usr/local/bin` is in your `PATH`, make your plugin discoverable by
+moving it to there:
+
+```bash
+sudo mv ./rasa-hello  /usr/local/bin
+```
+
+Make sure that the executable name begins with `rasa-`.
+
+### Listing Available Plugins
+
+You can use `rasa plugins` to list command line plugins which are discoverable by
+Rasa Open Source. Rasa Open Source discovers command line plugins by searching your
+`PATH` environment for executables which begin with the prefix `rasa-`.
+
+### Writing a Plugin
+
+You can write a plugin in any programming language or script that allows you
+to write command-line commands. The command that your plugin provides, is
+determined based on the name of your plugin's executable. Everything after the first dash
+will be considered the plugin name. This means an executable `rasa-my-plugin` will
+be available via `rasa my-plugin`.
+
+The following snippet is the example of a plugin which prints `Hello world!` if you
+run `rasa hello` without additional parameters.
+
+```bash
+#!/bin/bash
+if [[ $# -eq 0 ]]
+then
+    echo "Hello world!"
+else
+    echo "Hello $1"
+fi
+```
+
+A plugin has to be executable. To make the script from above executable run:
+
+```bash
+chmod +x ./rasa-hello
+```
+
+Then add your plugin to your `PATH` as described
+[here](command-line-interface.mdx#installing-a-plugin).
+You can now invoke your plugin, by running:
+
+```bash
+rasa hello
+# Hello world!
+```
+
+All arguments which you pass to your command, will be passed to your plugin as they are:
+
+```bash
+rasa hello universe
+# Hello universe!
+```
--- a/rasa/__main__.py
+++ b/rasa/__main__.py
@@ -11,6 +11,7 @@ from rasa.cli import (
     data,
     export,
     interactive,
+    plugins,
     run,
     scaffold,
     shell,
@@ -68,6 +69,8 @@ def create_argument_parser() -> argparse.ArgumentParser:
     data.add_subparser(subparsers, parents=parent_parsers)
     export.add_subparser(subparsers, parents=parent_parsers)
     x.add_subparser(subparsers, parents=parent_parsers)
+
+    plugins.add_plugin_parser(parser, subparsers, parents=parent_parsers)
 
     return parser
 
--- a/rasa/cli/plugins.py
+++ b/rasa/cli/plugins.py
@@ -1,0 +1,111 @@
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+from typing import Text, List, Set
+
+import rasa.shared.utils.cli
+import rasa.shared.constants
+
+PLUGIN_PREFIX = "rasa-"
+
+
+# noinspection PyProtectedMember
+def add_plugin_parser(
+    parser: argparse.ArgumentParser,
+    subparsers: argparse._SubParsersAction,
+    parents: List[argparse.ArgumentParser],
+) -> None:
+    _register_plugin_fallback(parser)
+    _add_plugins_list_parser(subparsers, parents)
+
+
+def _register_plugin_fallback(parser: argparse.ArgumentParser) -> None:
+    original_error_fn = parser.error
+
+    def command_not_found(message: Text) -> None:
+        args = sys.argv
+
+        if len(args) < 2:
+            return original_error_fn(message)
+
+        plugin_path = shutil.which(_executable_name(args))
+        if plugin_path:
+            return _run_plugin(plugin_path, args)
+
+        return original_error_fn(message)
+
+    # This patches only the outermost parser but that's enough for our purposes
+    parser.error = command_not_found
+
+
+def _executable_name(args: List[Text]) -> Text:
+    return f"{PLUGIN_PREFIX}{args[1]}"
+
+
+def _run_plugin(plugin_path: Text, current_args: List[Text]) -> None:
+    current_args[1] = plugin_path
+    # Remove `rasa` executable and replace plugin name with its path
+    current_args = current_args[1:]
+    try:
+        process = subprocess.run(current_args)
+        sys.exit(process.returncode)
+    except Exception as e:
+        rasa.shared.utils.cli.print_error(
+            f"Something went wrong when executing the plugin "
+            f"'{current_args[0]}'. Please make sure it is a valid "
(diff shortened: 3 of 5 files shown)


docs/docs/command-line-interface.mdx
```diff
@@ -321,3 +321,78 @@ The following arguments are available for `rasa x`:
...
+It is currently not possible to overwrite existing `rasa` commands. For example it is
+not possible to use a plugin `rasa-train` to override the `rasa train` command.
+:::
+
+### Installing a Plugin
+
+A command line plugin is nothing more than an executable file, whose name begins with
+`rasa-`. Once it's installed, it is available on the command line via
+`rasa <plugin name>`.
+To install a command line plugin, move its executable somewhere on your `PATH`.
```
erohmensing, line 341: i'd tell the user either how to view their path, or how to add the directory with the executable to their path

docs/docs/command-line-interface.mdx
```diff
@@ -321,3 +321,78 @@ The following arguments are available for `rasa x`:
...
+
+You can use `rasa plugins` to list command line plugins which are discoverable by
+Rasa Open Source. Rasa Open Source discovers command line plugins by searching your
+`PATH` environment for executables which begin with the prefix `rasa-`.
+
+### Writing a Plugin
+
+You can write a plugin in any programming language or script that allows you
+to write command-line commands. The command which your plugin is providing, is
+determined based on the name of our plugin's executable. Everything after the first dash
```
erohmensing, line 354:
```suggestion
to write command-line commands. The command that your plugin provides is
determined based on the name of your plugin's executable. Everything after the first dash
```

docs/docs/command-line-interface.mdx
erohmensing: maybe by moving the

To make the plugin discoverable, it has to be in your `PATH` environment. You can see
all potential directories in your `PATH` by running `echo $PATH`.
Assuming that `/usr/local/bin` is in your `PATH`, make your plugin discoverable by
moving it to there:

```bash
sudo mv ./rasa-hello  /usr/local/bin
```

Stuff up here and linking from the other one like "install your plugin to make it discoverable". this would make sure users know how to install whether writing their own or using one someone else wrote

erohmensing (commented):
Seems like acool idea! Are we going to include examples in the initial PR?

rasa/cli/plugins.py
```diff
@@ -0,0 +1,105 @@
...
+    plugin_names = _plugin_executables()
+
+    if not plugin_names:
+        cli_utils.print_info("No command-line plugins were found.")
+    else:
+        formatted_plugins = "\n".join(
+            [f"- {executable}" for executable in plugin_names]
+        )
+        cli_utils.print_info(
+            f"The following command-line plugins were found:\n" f"{formatted_plugins}"
```
m-vdb, line 91:
```suggestion
            f"The following command-line plugins were found:\n{formatted_plugins}"
```

rasa/cli/plugins.py
```diff
@@ -0,0 +1,105 @@
...
+    parser: argparse.ArgumentParser,
+    subparsers: argparse._SubParsersAction,
+    parents: List[argparse.ArgumentParser],
+) -> None:
+    _register_plugin_fallback(parser)
+    _add_plugins_list_parser(subparsers, parents)
+
+
+def _register_plugin_fallback(parser: argparse.ArgumentParser) -> None:
+    current_error_fn = parser.error
```
m-vdb, line 25:
```suggestion
    original_error_fn = parser.error
```

rasa/__main__.py
```diff
@@ -62,6 +63,8 @@ def create_argument_parser() -> argparse.ArgumentParser:
     export.add_subparser(subparsers, parents=parent_parsers)
     x.add_subparser(subparsers, parents=parent_parsers)
 
+    plugins.add_plugin_parser(parser, subparsers, parents=parent_parsers)
```
m-vdb, line 66: do we also want to update the output of `rasa --help` to include information about plugins?

tests/cli/test_plugins.py
```diff
@@ -0,0 +1,91 @@
...
+def example_plugin() -> Text:
+    if sys.platform == "win32":
+        return f"""echo on
+echo "{EXPECTED_PLUGIN_OUTPUT}" """
+    else:
+        return f"""#!/bin/bash
+echo "{EXPECTED_PLUGIN_OUTPUT}" """
+
+
+def _install_plugin(
```
m-vdb, line 25: can we also add a test for a plugin that tried to override a rasa command?

m-vdb (changes requested):
super nice! what could be nice is to have a pypi package (or other) that is named `rasa-hello` and is an example of a plugin. That way you could reference it in the docs + it would give a boilerplate for user (they could fork it and play with it). But maybe not at this stage 🤷

docs/docs/command-line-interface.mdx
```diff
@@ -321,3 +321,78 @@ The following arguments are available for `rasa x`:
...
+
+### Installing a Plugin
+
+A command line plugin is nothing more than an executable file, whose name begins with
+`rasa-`. Once it's installed, it is available on the command line via
+`rasa <plugin name>`.
+To install a command line plugin, move its executable somewhere on your `PATH`.
+Make sure that the executable name begins with `rasa-`.
+
+### Listing Available Plugins
```
TyDunn, line 344: The part that is not clear to me. How can I take advantage of plugins created by other people? Where can find them?

m-vdb: I agree and I think it relates a bit to my comment:
> what could be nice is to have a pypi package (or other) that is named rasa-hello and is an example of a plugin. That way you could reference it in the docs + it would give a boilerplate for user (they could fork it and play with it). But maybe not at this stage 🤷

if we have a lightweight setup here:
- setup a `rasa-plugins` public monorepo
- create a `rasa-hello` plugin there
- welcome pull requests to the repo
- have a quick-and-dirty way to install the plugins (basically clone the repo + manipulate the path)

I think that could work easily and help with discovery / contribution. WDYT @TyDunn @wochinge ?

wochinge: @koaning How about using `rasa-lit` for this?

tmbo:
We can't guarantee backwards-compatibility with any plugins, as we might overwrite existing plugins when we add new commands. This got me thinking and it might make sense to run plugins using a separate command, e.g. `rasa external hello`

wochinge:
> We can't guarantee backwards-compatibility with any plugins, as we might overwrite existing plugins when we add new commands. This got me thinking and it might make sense to run plugins using a separate command, e.g. rasa external hello

Valid point 👍  I wouldn't move to a sub-command because
* when forcing it to use `rasa external` the plugins loose the ability to integrate themselves seamlessly and this might push users to define their own CLI command instead of integrating it into the Rasa CLI.
* somewhat stupid argument, but `kubectl` is doing fine without using a subcommand (you can check out some plugins of theirs [here](https://awesomeopensource.com/projects/kubectl-plugins) - they are quite often some product names)
* our CLI changes relatively rarely. I think the chances that we pick a command which matches an existing CLI plugin are very small. I'd also add a warning to the docs that users shouldn't pick to general names for their plugins to avoid that they will be occupied in the future. If a potentially new CLI command matches the name of a plugin I could also imagine that they are doing similar things in the first place (e.g. when moving something from an experimental plugin to the Rasa Open Source core) which seems okay to me then

rasa/__main__.py
wochinge: I added the link to the documentation of plugins to the output. `rasa --help` now looks as follows:

```bash
❯ rasa --help
usage: rasa [-h] [--version] {init,run,shell,train,interactive,telemetry,test,visualize,data,export,x,plugins} ...

Rasa command line interface. Rasa allows you to build your own conversational assistants 🤖. The 'rasa' command allows you to easily run most common commands like creating a new bot, training or evaluating models.

positional arguments:
  {init,run,shell,train,interactive,telemetry,test,visualize,data,export,x,plugins}
                        Rasa commands
    init                Creates a new project, with example training data, actions, and config files.
    run                 Starts a Rasa server with your trained model.
    shell               Loads your trained model and lets you talk to your assistant on the command line.
    train               Trains a Rasa model using your NLU data and stories.
    interactive         Starts an interactive learning session to create new training data for a Rasa model by chatting.
    telemetry           Configuration of Rasa Open Source telemetry reporting.
    test                Tests Rasa models using your test NLU data and stories.
    visualize           Visualize stories.
    data                Utils for the Rasa training files.
    export              Export conversations using an event broker.
    x                   Starts the Rasa X interface.
    plugins             List available command-line plugins. See 'https://rasa.com/docs/rasa/command-line-interface#command-line-plugins' for more information.

optional arguments:
  -h, --help            show this help message and exit
  --version             Print installed Rasa version
```

docs/docs/command-line-interface.mdx
wochinge: > How can I take advantage of plugins created by other people?

I think that's something they would need to document in their README as this might be different for each programming language. 

> Where can find them?

We can start listing some here as they evolve. A good start would be `rasa lit` in my opinion. 

> setup a rasa-plugins public monorepo

I don't think that one repository for all plugins makes sense. Take `rasa-lit` as example: It stands for itself and the fact that it can be used as CLI plugin for Rasa Open Source is only a nice to have for it. 

> have a quick-and-dirty way to install the plugins (basically clone the repo + manipulate the path)

That would be a simply `pip install` btw.

I can obviously create a quick `rasa-hello` repo, but I don't think it's necessary because
1) `rasa-lit` can serve as example
2) The target group for this is developers who are developing some tool in the rasa ecosystem. They don't need any boilerplate to start out, they just need to expose their existing project as executable so that Rasa Open Source can find it. The way this is done is different per build tool / programming language, but is usually very well documented.

wochinge:
Updated the PR and worked in the review comments.

m-vdb (commented):
LGTM @wochinge great initiative 🚀 I am ok with [your answer to Tom's comment](https://github.com/RasaHQ/rasa/pull/6447#issuecomment-700149902), @tmbo WDYT?

wochinge:
@tmbo ?

tmbo:
It isn't a strong no - I think I would sleep better though if it wouldn't silently fail if we overwrite someone's plugin, is there an easy way to check that?

wochinge:
> Is there an easy way to check that?

We'd have to check it whenever the CLI is invoked. A simply check would be to use `shutil.which` and check whether it returns something 🤔  Just checked it and it would add `0.003` milliseconds to the startup time, which seems fine. Will add a check then 👍🏻

Issue #6399, opened 2020-08-13
Add CI to lint data examples in docs


erohmensing:
We provide lots of examples of training data and domains in our documentation. It would be great to have CI that validates whether these examples are valid rasa format. We can use the existing rasa methods to validate this.

**Implementation:**

1. on the rendering side, docusaurus uses Prisme. And it looks possible to add a custom language (i’d suggesting “yaml-rasa”). So basically we would add a new language that has the same syntax as yaml.
2. on the CI side: it should be possible to detect codeblocks of a specific type. This is was we’re currently doing in our of our plugins. This way we can run tests that pass each codeblock of this type through the training data reader/linter.

erohmensing:
@m-vdb would step 1 be something that you or @lunelson do? I am happy to take on step 2. I've seen this [come up quite a lot](https://rasa-hq.slack.com/archives/C36SS4N8M/p1597329286049800) during the alpha, so would love to have it around

lunelson:
>would step 1 be something that you or @lunelson do?
@erohmensing yeah that sounds possible. Is the main concern here that you want to lint these things on build, or that you want to highlight them a different way, or that you just want to the code block to indicate that it's "rasa yaml" not just "yaml"?

erohmensing:
Main concert is linting things on build to make sure they're valid! I don't know one adds a custom language, but for these purposes it could be just a subclass of the yaml one that behaves exactly the same, but is rasa yaml, and therefore needs to be linted in the CI. 

highlighting them in a different way could also be cool but is likely unnecessary 😄

lunelson:
OK cool. The linting process sounds like something that could be added alongside the other build scripts that run in the `pre-build` step, as a node script that just extracts those codeblocks on the basis of regexes, wdyt @m-vdb ?

m-vdb:
yes that's what I had in mind. In terms of prio, it doesn't strike me as something we need to do now (as opposed to after 2.0 release). WDYT @erohmensing ?

erohmensing:
Mm, I think it'd be really good to have it for the 2.0 release. This would make sure that every example we give is actually valid, which is especially important when we're trying to tell users how to use a completely new training format for them.

PR #6449, opened 2020-08-19
Docs lint/test traning data

m-vdb:
**Proposed changes**:
- Fix https://github.com/RasaHQ/rasa/issues/6399
- Tech writer API: use `yaml-rasa` or `yml-rasa` code blocks in docs
- Logic: created a simple python test that loads every docs file, parses it for codeblocks and validates the training data using `RasaYAMLReader`


erohmensing:
bless you for working on this 🙏 i was looking into tagging all of them this morning -- happy to put in the manual work of doing that once this is ready

m-vdb:
@erohmensing it's ready! If you could flag all the training data that we want to validate in the docs, that'd be awesome. I've added a description to the PR so that it's easier to understand, but anyway it's not super hard to use. I've already added a CI step, and you can run the checks locally with `make test-docs`. It's highly probable that some use cases are not supported yet - bear with me!

m-vdb:
@erohmensing yes sure, will do 👍

erohmensing:
Otherwise I believe we are good! It can't seem to get some files for some reason, not sure what's up with that

m-vdb:
> It can't seem to get some files for some reason

@erohmensing which files?

erohmensing:
Oh wait, maybe someone updated them without updating them in the docs, looks like the autoconfig ones https://github.com/RasaHQ/rasa/runs/1014039899#step:13:113

m-vdb:
@erohmensing no it's a problem that has been introduced with this PR. Found the root cause, on it

.github/workflows/continous-integration.yml
```diff
@@ -101,11 +101,14 @@ jobs:
       # trying to run the `rasa` command once before the docs build.
       run: poetry run rasa --help
 
-    - name: Test Docs 📃
+    - name: Build Docs 📃
       run: |
         poetry run python -c "from scripts import release; release.generate_changelog('major.minor.patch')"
```
erohmensing, line 106: wdyt about adding this directly to `make docs` so its easier to `make docs test-docs` locally?

m-vdb: no we can't do that, it's used to generate the docs in the deployment pipeline

m-vdb: but we could definitely modify test-docs to do that automatically

erohmensing: ah makes sense 👍

erohmensing: wait, doesn't the changelog still need to be generated though 🤔

m-vdb: yes but it's not done like this. Look at the "major.minor.patch" argument :)

m-vdb: I updated the workflow here: c55905a (and added a helper). It should be easier to use locally now 👍

erohmensing (approved):
🥳

PR #6453, opened 2020-08-19
Cleanup deprecation warnings

m-vdb:
**Proposed changes**:
- cleanup some old deprecation warnings
- new API to raise a deprecation warning, that requires a `warn_until_version` argument.
([Original Slack thread](https://rasa-hq.slack.com/archives/C36SS4N8M/p1597154190078900))


m-vdb:
@tmbo @erohmensing this is a second pass to remove deprecation warnings and introduce a new API to raise deprecation warnings. WDYT?

rasa/core/tracker_store.py
```diff
@@ -186,6 +186,8 @@ def _deserialise_dialogue_from_pickle(
         sender_id: Text, serialised_tracker: bytes
     ) -> Dialogue:
 
+        # TODO: not sure if we should remove because it says "it will be deprecated in 2.0"
+        #       instead of removed.
         logger.warning(
```
erohmensing, line 191: This should be turned into one of your new warnings now, no?

rasa/nlu/training_data/formats/markdown.py
```diff
@@ -65,19 +63,6 @@ def reads(self, s: Text, **kwargs: Any) -> "TrainingData":
...
-                " Please use the following format: \n"
-                '[<entity-text>]{"entity": "<entity-type>", "value": '
-                '"<entity-synonym>"}.'
-                "\nYou can use the following command to update your training data file:"
-                "\nsed -i -E 's/\\[([^)]+)\\]\\(([^)]+):([^)]+)\\)/[\\1]{"
-                '"entity": "\\2", "value": "\\3"}/g\' nlu.md',
-                category=FutureWarning,
-                docs=DOCS_URL_TRAINING_DATA_NLU,
-            )
-
```
erohmensing, line 80: Have we tested how this one breaks if you provide the wrong format? Just wondering because this is one i'm almost certain people will run into (having seen the deprecation warning before but not removing it)

Also I don't see where we removed the implementation of the functionality still working during deprecation phase

rasa/utils/common.py
```diff
@@ -385,6 +385,20 @@ def formatwarning(
...
+) -> None:
+    """
+    Thin wrapper around `raise_warning()` to raise a deprecation warning. It requires
+    a version until which we'll warn, and after which the support for the feature will
+    be removed.
+    """
+    if warn_until_version not in message:
+        message = f"{message} (will be removed in {warn_until_version})"
+
+    raise_warning(message, DeprecationWarning, docs, **kwargs)
```
erohmensing, line 399: If we don't raise a `FutureWarning`, it won't be printed into the user's console.

erohmensing (commented):
Yes, this is what I had in mind!

rasa/core/tracker_store.py
m-vdb: yeah, I left it because of the `TODO:`, not sure if it's ok to remove it. If we do not remove it, I'll turn it into the new syntax

rasa/nlu/training_data/formats/markdown.py
m-vdb: I have no idea yet - but it looks like we dropped support for this for some time now

rasa/utils/common.py
m-vdb: arf, but why?

rasa/core/tracker_store.py
erohmensing: How i interpreted it: 

someone during 1.x decided to deprecate this. But instead of saying "this will be removed in 2.0" they said "this will be deprecated in 2.0" 

Someone upon 2.0 cleaned up the deprecation warnings. But they were unsure whether they could remove this functionality because it didn't say it would be removed in 2.0 

IMO we should deprecate this in 2.0, to be removed in 3.0. Which is what leaving the warning does -- was just saying that we should update it to your new format :)

rasa/nlu/training_data/formats/markdown.py
erohmensing: oh weird 👀

rasa/utils/common.py
erohmensing: https://docs.python.org/3/library/warnings.html#warning-categories

Deprecation warning: 

Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).
--


Furue warning: 

Base category for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.
--
 
Not sure if things are triggered by code in main? but would need to check. this was implemented because important messages werent getting to the user 🤔

rasa/core/tracker_store.py
m-vdb: ok, will got with this then

rasa/utils/common.py
m-vdb: ok will test

rasa/core/channels/mattermost.py
```diff
@@ -132,12 +132,12 @@ def from_credentials(cls, credentials: Optional[Dict[Text, Any]]) -> InputChanne
...
         # pytype: disable=attribute-error
         if credentials.get("pw") is not None or credentials.get("user") is not None:
-            raise_warning(
+            common_utils.raise_deprecation_warning(
                 "Mattermost recently switched to bot accounts. 'user' and 'pw' "
                 "should not be used anymore, you should rather convert your "
                 "account to a bot account and use a token. Password based "
                 "authentication will be removed in a future Rasa Open Source version.",
-                FutureWarning,
+                warn_until_version="3.0.0",
```
degiz, line 140: Should we extract that to a constant? Something like `NEXT_MAJOR_DEPRECATIONS` (assuming that `2.0` is out).
Then it will be super easy to find this things.

degiz (changes requested):
That's great! 🚀 

One note about `Removed support for string `responses` (use dictionaries instead).`
Let's include that in the `rasa/core/schemas/domain.yml`, then it will be a part of our standard domain verification!

rasa/core/channels/mattermost.py
m-vdb: I can see the pure engineering argument to define the constant, but I don't think it would help finding these things (simply looking for `warn_until_version` would yield the same results).

That being said, we could even go further and:
- define the constant
- make it a default argument to the function: I do not see a use case where we would need to raise the warning until the next next major (e.g. 4.0). WDYT?

rasa/core/schemas/domain.yml
```diff
@@ -27,6 +27,33 @@ mapping:
...
+            regex;(^buttons$):
+              type: "seq"
+              sequence:
+              - type: "map"
+                mapping:
+                  title:
+                    type: "str"
+                  payload:
+                    type: "str"
+
```
erohmensing, line 56: Can we enforce that they need to have at least one of `(custom, text)`?

m-vdb: I don't think we can, using the library we use. I've briefly looked at it and didn't find it. Will look again

rasa/core/channels/mattermost.py
m-vdb: a546cd93508 WDYT?

rasa/core/schemas/domain.yml
m-vdb: oh actually, this already validates that one of custom, text, image or buttons is present. Is that enough? You can see the tested config file [here](https://github.com/RasaHQ/rasa/pull/6453/files#diff-45293964fd59daf5bde79922380588cdR17)

rasa/core/channels/mattermost.py
degiz: Sounds good!

rasa/core/schemas/domain.yml
```diff
@@ -27,6 +27,33 @@ mapping:
   responses:
     type: "map"
     allowempty: True
+    matching-rule: 'any'
```
degiz, line 30: not necessary since there's only one regex below.

rasa/core/schemas/domain.yml
```diff
@@ -27,6 +27,33 @@ mapping:
...
+    mapping:
+      regex;(.+):
+        type: "seq"
+        required: True
+        sequence:
+        - type: "map"
+          allowempty: True
+          matching-rule: 'any'
+          mapping:
+            regex;(^text$):
```
degiz, line 40: is there a reason to use a regex here? I though that the keys `text`, `image`, `buttons`, `custom` are fixed 🤔
If we replace the regex with fixed values, then `matching-rule: 'any'` is not necessary as well.

rasa/core/schemas/domain.yml
m-vdb: yeah and it's the default anyway

rasa/core/schemas/domain.yml
m-vdb: no I think I got confused 🤷 removing

rasa/core/schemas/domain.yml
```diff
@@ -27,6 +27,39 @@ mapping:
   responses:
```
m-vdb, line 27: @degiz since I copy-pasted the definition of the `responses` key in both schemas, I thought about maybe defining the response schema in `rasa/utils/schemas/responses.yml` and [including](https://pykwalify.readthedocs.io/en/latest/partial-schemas.html#include) it in both domain and nlu. Have you thought about this?

degiz: The reason `responses` are in the nlu schema is to allow users to keep their training data in one file, but each individual reader would use it's own schema.
The approach is not perfect though 🤦 I'm still thinking how to do it properly and even have an items for that [here](https://github.com/orgs/RasaHQ/projects/30)

m-vdb: I'm not saying we should keep everything in one schema. I'm simply proposing to remove the schema definition duplication, by using the [inclusion feature of `pykwalify`](https://pykwalify.readthedocs.io/en/latest/partial-schemas.html#include), that allows us to import a schema definition inside another schema definition. Does that make sense?

m-vdb: @degiz check this out: dba1ebbe57a WDYT?

m-vdb:
@degiz I've implemented the changes we just talked about: keep `nlu.yml` the same as it doesn't need validation to go through; custom validation of "custom" and "text" keys inside of a response. Ready for review 🚀

rasa/core/schemas/domain.yml
```diff
@@ -27,6 +27,40 @@ mapping:
...
+              sequence:
+              - type: "map"
+                mapping:
+                  title:
+                    type: "str"
+                  payload:
+                    type: "str"
+            channel:
+              type: "str"
+            # allow custom keys
```
degiz, line 59: is it necessary if we already have a very flexible `custom` key?
Asking because users can make a type, like `imag` instead of `image` and that would pass the verification.

m-vdb: yeah I'd rather remove it too, but some tests failed because of this, for instance [this one](https://github.com/RasaHQ/rasa/blob/master/tests/nlu/training_data/formats/test_rasa_yaml.py#L311). Where can I find the definition of the format and what's officially supported?

degiz: clarification from tom, who implemented that:

> keys are exactly the same as for domain responses (utter templates)

So that's `text` `buttons` `image` `custom` `channel`.

degiz: I believe you can adjust the failing test and remove invalid examples.

m-vdb: yayyyyy

m-vdb:
@degiz last thing: `test_nlg_fails_on_empty_response()` is failing because of [the changes I made in `RasaYAMLReader`](https://github.com/RasaHQ/rasa/pull/6453/files#diff-86155ca59d79b070ac1783ef65833d5c). Should I remove the test? Why did we write it in the first place?

degiz:
@m-vdb I've just noticed that responses are parsed as a part of `NLU`, it used to be a `domain`!

@tmbo since you've implemented the recent changes to response selector, we now have `Domain::collect_templates()` and `RasaYAMLReader` both operating on `responses` YAML key, was it done intentionally? Could it have be just one of the files?

tmbo:
yes and no - it could have been a single file, but previously we required users to move the responses into their own file which made it easier to implement this in the same way. 

conceptually it should be possible to combine utter templates and response selector responses, we just need to make sure they end up in the right place

m-vdb:
@degiz from [the commit on this file](https://github.com/RasaHQ/rasa/commit/da11cf9cd95a2d24a76370d57d5f62fbf38d9ae5#diff-75686dc39a842033808ce0c6137c102b) it looks like it was intended. The commit also introduced the changes to [`schemas/nlu.yml`](https://github.com/RasaHQ/rasa/commit/da11cf9cd95a2d24a76370d57d5f62fbf38d9ae5#diff-90a2c30c42c614b474f32ac0230b79dbR49-R52).

It looks to me that I should re-revert my changes:
- create `rasa/utils/schemas.yml` file
- define a [partial schema](https://pykwalify.readthedocs.io/en/latest/partial-schemas.html#partial-schemas) for the `responses` key there
- reference this partial schema in both `domain.yml` and `nlu.yml`

Do you agree?

degiz:
@m-vdb I'm sorry things are so complicated 😄 
Anyway, this "split responses" topic is definitely not part of this commit, so yes, please proceed with you original change.

m-vdb:
no problemo - good learning opportunity for me! Will push my changes

degiz (approved):
🚀

PR #6458, opened 2020-08-20
improve the suggested pipeline

### LINKED ISSUE #6326: Improve the suggested pipelines
See feedback [in Notion](https://www.notion.so/rasa/Feedback-on-2-0-4574eaacc0764967ac0b8b82fbc72904#75a5a5747ffa462ba0b9ead6d45b355f).

Currently, doing `pip install rasa==2.0.0a1` and then `rasa init` fails because the suggested pipeline includes spaCy, which is not installed by default. Also, the suggested pipeline doesn't match what we recommend in [our docs](https://rasa.com/docs/rasa/nlu/choosing-a-pipeline/#how-to-choose-a-pipeline).

A potential solution would be to include conveRT in the default installation, except on Windows (where tensorflow-text isn't supported yet), and then suggest the following pipelines:

If `language: en` and not on Windows, suggest
```yaml
pipeline:
  - name: ConveRTTokenizer
  - name: ConveRTFeaturizer
  - name: RegexFeaturizer
  - name: LexicalSyntacticFeaturizer
  - name: CountVectorsFeaturizer
  - name: CountVectorsFeaturizer
    analyzer: "char_wb"
    min_ngram: 1
    max_ngram: 4
  - name: DIETClassifier
    epochs: 100
  - name: EntitySynonymMapper
  - name: ResponseSelector
    epochs: 100
```

else suggest
```yaml
pipeline:
  - name: WhitespaceTokenizer
  - name: RegexFeaturizer
  - name: LexicalSyntacticFeaturizer
  - name: CountVectorsFeaturizer
  - name: CountVectorsFeaturizer
    analyzer: "char_wb"
    min_ngram: 1
    max_ngram: 4
  - name: DIETClassifier
    epochs: 100
  - name: EntitySynonymMapper
  - name: ResponseSelector
    epochs: 100
```

chkoss:
Fix #6326 

**Proposed changes**:
- install `tensorflow-text` by default, so the `ConveRTFeaturizer` can be used (except on Windows)
- if on Windows or the `language` is not `en`, use an embedding-less pipeline
- if not on Windows and `language: en`, use a pipeline with `ConveRT`


poetry.lock
```diff
@@ -3065,7 +3065,6 @@ version = ">=3.7.4"
 [[package]]
 category = "main"
 description = "Backport of pathlib-compatible object wrapper for zip files"
-marker = "python_version < \"3.7\" and python_version != \"3.4\" or python_version < \"3.8\""
```
chkoss, line 3068: Not sure where this comes from... I only updated the section for `tensorflow-text` in the `pyproject.toml` and then ran `poetry update tensorflow-text`.

tests/dependencies.py
```diff
@@ -0,0 +1,12 @@
+import sys
+import pkg_resources
+
+
+def test_tensorflow_text_install():
```
chkoss, line 5: There is probably a better place for this to go, this isn't really a unit test so it shouldn't be in `tests`. I just added it here for now so I can see in the CI if it works on both Linux and on Windows.

chkoss, line 5: There is probably a better place for this to go, this isn't really a unit test so it shouldn't be in `tests`. I just added it here for now so I can see in the CI if it works on both Linux and on Windows.

tmbo:
@wochinge I guess @chkoss won't be able to wrap this up, are you taking over?

chkoss:
@tmbo I hopefully will finish this, it's on my to-do list for today/tomorrow.

tests/dependencies.py
chkoss: @wochinge When renaming `import_time.py` to `test_import_time.py` the test failed (average import time 0.9, while the test tries to assert <0.3). I undid the rename for now because it isn't related to this PR, but will create a new issue for this.

pyproject.toml
```diff
@@ -168,7 +167,7 @@ optional = true
 
 [tool.poetry.dependencies.tensorflow-text]
 version = "~2.3"
-optional = true
+markers = "sys_platform!='win32'"
```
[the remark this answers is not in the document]
chkoss: I also created an issue for it, so this doesn't get lost https://github.com/RasaHQ/rasa/issues/6482

PR #6466, opened 2020-08-21
Train Response Selector on labels by default instead of text

dakshvar22:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/research/issues/111
- Adds `intent_response_key` attribute to `Message` class which will hold the complete `intent/response_key`(for example - `faq/languages`) label. This attribute is not dense featurizable.
- Added a flag - `use_text_as_label` in `ResponseSelector` to configure whether to train on text of responses or the `intent/response_key` label. Default being training on `intent/response_key` label.
- Sub-intents inside a retrieval intent(for example - `faq/languages` inside `faq`) can now have multiple response templates. These are handled by always using the first response template containing the `text` attribute for training and selecting a random template for bot utterance.
- Deprecate `full_retrieval_intent` from the output of the ResponseSelector and instead use `intent_response_key` to store the full intent name as it appears in the training data. This is more consistent with the naming schema used in the code.
- Deprecate `name` as the key to store the predicted response text and instead use `response_templates` to output all available response templates for the predicted `intent_response_key`.
- Added a key `all_retrieval_intents` to response selector output which will hold a list of all retrieval intents(`faq`,`chitchat`, etc.) that are present in the training data.
- So now the output looks like - 
```
"response_selector": {
    "all_retrieval_intents": ["faq"],
    "default": {
      "response": {
        "id": 1388783286124361986, "confidence": 1.0, "intent_response_key": "faq/is_legit",
        "response_templates": [
          {
            "text": "absolutely",
            "image": "https://i.imgur.com/nGF1K8f.jpg"
          },
          {
            "text": "I think so."
          }
        ],
      },
      "ranking": [
        {
          "id": 1388783286124361986,
          "confidence": 1.0,
          "intent_response_key": "faq/is_legit"
        },
      ]
```
- All response selector related evaluation artifacts - `report.json, successes.json, errors.json, confusion_matrix.png` now use the complete `intent_response_key` as the label instead of the actual response text. Makes them look more concise and clean and at the same time no information is lost.
- `response_selection_successes.json` looks like - 
```
{
      "text": "Is this really legit?",
      "intent_response_key_target": "faq/is_legit",
      "intent_response_key_prediction": {
              "name": "faq/is_legit",
              "confidence": 0.88
      }
}
```
- `response_selection_errors.json` looks like - 
```
{
      "text": "Is this really legit?",
      "intent_response_key_target": "faq/is_legit",
      "intent_response_key_prediction": {
              "name": "faq/ask_cost",
              "confidence": 0.6
      }
}
```


rasa/nlu/selectors/response_selector.py
```diff
@@ -204,6 +205,9 @@ def required_components(cls) -> List[Type[Component]]:
         MASKED_LM: False,
         # Name of the intent for which this response selector is to be trained
         RETRIEVAL_INTENT: None,
+        # Boolean flag to check if actual text of the response
+        # should be used as ground truth label for training the model.
+        TRAIN_ON_TEXT: True,
```
dakshvar22, line 210: Reminder for myself: Change this back to False, before the merge. It is `True` right now for one run of model regression tests.

rasa/nlu/test.py
```diff
@@ -1333,15 +1332,17 @@ def get_eval_data(
                 response_prediction_key, {}
             ).get(OPEN_UTTERANCE_PREDICTION_KEY, {})
 
+            # print(response_prediction)
+
```
tabergma, line 1336:
```suggestion
```

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -100,15 +107,30 @@ def process(self, message: Message, **kwargs: Any) -> None:
         tokens = self.tokenize(message, TEXT)
         message.set(TOKENS_NAMES[TEXT], tokens)
 
-    def _split_intent(self, message: Message):
-        text = message.get(INTENT)
+    def _tokenize_on_split_symbol(self, text: TEXT) -> List[TEXT]:
```
tabergma, line 110:
```suggestion
    def _tokenize_on_split_symbol(self, text: Text) -> List[Text]:
```

rasa/nlu/training_data/training_data.py
```diff
@@ -243,15 +239,22 @@ def _fill_response_phrases(self) -> None:
...
             # look for corresponding bot utterance
             story_lookup_intent = example.get_combined_intent_response_key()
             assistant_utterances = self.responses.get(story_lookup_intent, [])
             if assistant_utterances:
-                # selecting only first assistant utterance for now
-                example.set(RESPONSE, assistant_utterances[0].get(TEXT))
+
+                # Use the first response text as training label if needed downstream
+                for assistant_utterance in assistant_utterances:
+                    if assistant_utterance.get(TEXT, None):
```
tabergma, line 252:
```suggestion
                    if assistant_utterance.get(TEXT) is not None:
```
The default return value is already `None`.

rasa/nlu/training_data/training_data.py
```diff
@@ -421,7 +424,7 @@ def validate(self) -> None:
 
         # emit warnings for response intents without a response template
         for example in self.training_examples:
-            if example.get(RESPONSE_KEY_ATTRIBUTE):
+            if example.get(RESPONSE_KEY) and example.get(RESPONSE, None) is None:
```
tabergma, line 427:
```suggestion
            if example.get(RESPONSE_KEY) and example.get(RESPONSE) is None:
```

rasa/nlu/selectors/response_selector.py
```diff
@@ -313,20 +322,22 @@ def preprocess_train_data(self, training_data: TrainingData) -> RasaModelData:
...
         )
 
         self._check_input_dimension_consistency(model_data)
 
         return model_data
 
-    def _full_response(self, label: Dict[Text, Any]) -> Optional[Dict[Text, Any]]:
+    def _full_response(
+        self, label: Dict[Text, Any]
+    ) -> Tuple[Any, List[Dict[Text, Any]]]:
```
tabergma, line 340: not sure, but I think we can replace `Tuple[Any, List[Dict[Text, Any]]]` with `Tuple[Text, List[Dict[Text, Any]]]`

rasa/nlu/selectors/response_selector.py
```diff
@@ -366,9 +387,13 @@ def process(self, message: Message, **kwargs: Any) -> None:
         )
 
         prediction_dict = {
-            "response": self._full_response(label) or {TEXT: label.get("name")},
+            "response": {
```
tabergma, line 390: As we changed this, do we need to update the components page in our docs?

rasa/nlu/selectors/response_selector.py
```diff
@@ -366,9 +387,13 @@ def process(self, message: Message, **kwargs: Any) -> None:
...
+                "id": label["id"],
```
tabergma, line 391: just for my understanding, what exactly is the id of label?

rasa/nlu/selectors/response_selector.py
```diff
@@ -480,7 +505,11 @@ def _prepare_layers(self) -> None:
         self.label_name = TEXT if self.config[SHARE_HIDDEN_LAYERS] else LABEL
 
         self._prepare_sequence_layers(self.text_name)
-        self._prepare_sequence_layers(self.label_name)
+
+        # When training on labels, don't create transformer layers for the label
+        self._prepare_sequence_layers(
```
tabergma, line 510: just to clarify: the incoming user message is still going through the transformer. the responses are just going through the transformer if we are using the text, but not if we are using the label, in that case it is just a ffnn and an embedding layer, correct?

rasa/nlu/test.py
```diff
@@ -451,15 +455,10 @@ def evaluate_response_selections(
...
-            response_to_intent_target[label]
-            if label in response_to_intent_target
-            else f"'{label[:20]}...' (response not present in test data)"
-            for label in labels
-        ]
+
         plot_utils.plot_confusion_matrix(
             confusion_matrix,
-            classes=_labels,
+            classes=labels,
```
tabergma, line 461: So `labels` is now always the retrieval intent?

rasa/nlu/test.py
dakshvar22: Ahh, thanks for catching that.

rasa/nlu/tokenizers/tokenizer.py
dakshvar22: Raaiight, now I know why my type checks are failing 🙌

rasa/nlu/selectors/response_selector.py
dakshvar22: I thought so too, but type check fails I think bc `label` as type `Dict[Text, Any]` which is valid because a value in that dict can be `None` as well.

rasa/nlu/selectors/response_selector.py
dakshvar22: Yes, I'll update the docs in a separate PR.

rasa/nlu/selectors/response_selector.py
dakshvar22: It's the hash of whatever is stored in the `name` of label.

rasa/nlu/selectors/response_selector.py
dakshvar22: Yes, that's correct. Just the transformer layer is skipped if training on labels.

rasa/nlu/test.py
dakshvar22: Yes, it makes the files look much cleaner IMO and conveys all the information also anyways.

rasa/nlu/selectors/response_selector.py
```diff
@@ -298,12 +303,16 @@ def preprocess_train_data(self, training_data: TrainingData) -> RasaModelData:
                 "all retrieval intents."
             )
 
+        label_attribute = RESPONSE if self.train_on_text else INTENT_RESPONSE_KEY
```
tabergma, line 306: Minor: Can we somehow unify the names `RESPONSE` and `INTENT_RESPONSE_KEY`? If you think it does not make sense it is also fine.

rasa/nlu/selectors/response_selector.py
tabergma: Can we then change the type of `label` to `Dict[Text, Optional[Text]]`?

erohmensing:
> Move full_retrieval_intent key in the output of ResponseSelector inside the response key. 

I'm a bit concerned about this, as the `parse_data` that comes out of events could be considered a public API and moving this could break users' front ends (e.g. [rasa x](https://github.com/RasaHQ/rasa-x/pull/3488/files#diff-651c9f3a1a70660a008b93746c021e3fR442)), actions (e.g. [sara](https://github.com/RasaHQ/rasa-demo/blob/master/actions/actions.py#L228-L230) ), or analytics functions that consume events, etc. 

Ideally we'd have a way to deprecate this if we're moving it, but I'm not sure how one would go about that 🤔

tabergma (approved):
Looks good from my side 👍 . I don't have time to test it today, but I think Ella is doing it. Apart from that we should update the documentation and add also an example for the training data, but it is fine if that is addressed in a separate PR. 

Thanks for updating the response selector 💯

changelog/6466.improvement.rst
```diff
@@ -0,0 +1,6 @@
+[Response selector](retrieval-actions.mdx) is now trained on retrieval intent labels by default instead of actual response text.
+
+If you want to use the actual response text for training, then use ``train_on_text=True`` in the configuration of ``ResponseSelector``.
```
erohmensing, line 3: Can we maintain the description of the features as described here in the changelog, but also mention in the migration guide _why_ this change was made (`For most models, this should improve training time of the component with similar confidence results.`, if that's accurate) as well as how to revert to the default? It'd be good to have this info whenever we change defaults that we know could actually be bad for peoples' models, e.g. tom's example of chitchat/faq_1, chitchat/faq_2. example below:

dakshvar22:
@erohmensing Hmm, good point. I mean it would be good to maintain compatibility but we don't have to maintain API compatibility for a new major version right, specially for an earlier experimental feature? We definitely need to break it for the `name` key of the output. Earlier, we used to put the text of the response directly in the `name` key but now since we want to support multimedia elements and multiple response templates, that definitely has to change.
But happy to hear how we can gracefully deprecate it maybe. @tmbo any thoughts?

changelog/6466.improvement.rst
erohmensing: [Response Selectors](retrieval-actions.mdx) are now trained on retrieval intent labels by default instead of actual response text. For most models, this should improve training time of the `ResponseSelector`, with similar confidence results.

If you want to revert to the pre-2.0 default behavior, add the `train_on_text: true` parameter to your `ResponseSelector` component.

changelog/6466.improvement.rst
```diff
@@ -0,0 +1,6 @@
+[Response selector](retrieval-actions.mdx) is now trained on retrieval intent labels by default instead of actual response text.
```
erohmensing, line 1: This file should be `changelog/6466.improvement.md` and you no longer need the double backticking for code formatting!

changelog/6466.improvement.rst
```diff
@@ -0,0 +1,6 @@
...
+
+If you want to use the actual response text for training, then use ``train_on_text=True`` in the configuration of ``ResponseSelector``.
+You can now also have multiple response templates for a single sub-intent of a retrieval intent. The first response template
+containing the text attribute is picked for training(if ``train_on_text=True``) and a random template is picked for bot's utterance just as how other ``utter_`` templates are picked.
+All response selector related evaluation artifacts - ``report.json, successes.json, errors.json, confusion_matrix.png`` now use the sub-intent of the retrieval intent instead of the actual response text.
```
erohmensing, line 6:
```suggestion

All response selector related evaluation artifacts - ``report.json, successes.json, errors.json, confusion_matrix.png`` now use the sub-intent of the retrieval intent instead of the actual response text.
```

changelog/6466.improvement.rst
```diff
@@ -0,0 +1,6 @@
+[Response selector](retrieval-actions.mdx) is now trained on retrieval intent labels by default instead of actual response text.
+
+If you want to use the actual response text for training, then use ``train_on_text=True`` in the configuration of ``ResponseSelector``.
+You can now also have multiple response templates for a single sub-intent of a retrieval intent. The first response template
```
erohmensing, line 4:
```suggestion

You can now also have multiple response templates for a single sub-intent of a retrieval intent. The first response template
```

rasa/core/actions/action.py
```diff
@@ -248,10 +249,17 @@ async def run(
...
         logger.debug(f"Picking response from selector of type {query_key}")
         selected = response_selector_properties[query_key]
-        message = selected[OPEN_UTTERANCE_PREDICTION_KEY]
-        message["template_name"] = selected[FULL_RETRIEVAL_INTENT]
+        possible_messages = selected[OPEN_UTTERANCE_PREDICTION_KEY]["name"]
 
-        return [create_bot_utterance(message)]
+        # Pick a random message from list of candidate messages
+        picked_message_idx = random.randint(0, len(possible_messages) - 1)
+        picked_message = copy.deepcopy(possible_messages[picked_message_idx])
```
erohmensing, line 256: I know before you were looking into using the NLG. I think here we don't support channel-specific utterances? That is something I think users will expect support for

dakshvar22: Hmm that's tricky. I would definitely like to make use of the NLG for it which means we need to have the response templates shared between the domain and NLU training data. I'll prefer to keep it out of scope of this PR till we have handled that. I'll mention it in the changelog though.

rasa/nlu/selectors/response_selector.py
```diff
@@ -338,21 +349,31 @@ def _full_response(self, label: Dict[Text, Any]) -> Optional[Dict[Text, Any]]:
             the text but also buttons, images, ...
         """
         for key, responses in self.responses.items():
-            for response in responses:
-                if hash(response.get(TEXT, "")) == label.get("id"):
-                    return response
-        return None
+            if self.train_on_text:
```
tmbo, line 352: this function isn't only used training, this property should get a better name

rasa/core/actions/action.py
erohmensing: Sounds good, creating an issue for it would be good too :)

rasa/nlu/selectors/response_selector.py
```diff
@@ -338,21 +349,31 @@ def _full_response(self, label: Dict[Text, Any]) -> Optional[Dict[Text, Any]]:
...
+                for response in responses:
+                    if hash(response.get(TEXT, "")) == label.get("id"):
+                        return key, responses
```
tmbo, line 355: why does this need to return a tuple, can't the caller also use `self.responses[key]`?

rasa/nlu/selectors/response_selector.py
```diff
@@ -262,9 +267,9 @@ def _create_retrieval_intent_mapping(
 
         retrieval_intent_mapping = {}
         for example in training_data.intent_examples:
-            retrieval_intent_mapping[
-                example.get(RESPONSE)
-            ] = f"{example.get(INTENT)}/{example.get(RESPONSE_KEY_ATTRIBUTE)}"
+            retrieval_intent_mapping[example.get(RESPONSE)] = example.get(
```
tmbo, line 270: please add a function description of what the dict is that get's returned here

rasa/nlu/selectors/response_selector.py
```diff
@@ -262,9 +267,9 @@ def _create_retrieval_intent_mapping(
 
         retrieval_intent_mapping = {}
         for example in training_data.intent_examples:
```
tmbo, line 269: should this rather loop over `training_data.response_examples`? otherwise there is always going to be 
`None: None` entry in this dictionary

rasa/nlu/training_data/training_data.py
```diff
@@ -243,15 +239,22 @@ def _fill_response_phrases(self) -> None:
...
             if assistant_utterances:
-                # selecting only first assistant utterance for now
-                example.set(RESPONSE, assistant_utterances[0].get(TEXT))
+
+                # Use the first response text as training label if needed downstream
+                for assistant_utterance in assistant_utterances:
+                    if assistant_utterance.get(TEXT) is not None:
+                        example.set(RESPONSE, assistant_utterance.get(TEXT))
+
+                # If none of the response templates have a text attribute, then use the key as the training label
```
tmbo, line 255: this looks to long

rasa/nlu/training_data/training_data.py
```diff
@@ -243,15 +239,22 @@ def _fill_response_phrases(self) -> None:
...
             story_lookup_intent = example.get_combined_intent_response_key()
             assistant_utterances = self.responses.get(story_lookup_intent, [])
             if assistant_utterances:
-                # selecting only first assistant utterance for now
-                example.set(RESPONSE, assistant_utterances[0].get(TEXT))
+
+                # Use the first response text as training label if needed downstream
+                for assistant_utterance in assistant_utterances:
+                    if assistant_utterance.get(TEXT) is not None:
+                        example.set(RESPONSE, assistant_utterance.get(TEXT))
```
tmbo, line 253:
```suggestion
                        example.set(RESPONSE, assistant_utterance[TEXT])
```

rasa/nlu/training_data/training_data.py
```diff
@@ -243,15 +239,22 @@ def _fill_response_phrases(self) -> None:
...
             # look for corresponding bot utterance
             story_lookup_intent = example.get_combined_intent_response_key()
             assistant_utterances = self.responses.get(story_lookup_intent, [])
             if assistant_utterances:
-                # selecting only first assistant utterance for now
-                example.set(RESPONSE, assistant_utterances[0].get(TEXT))
+
+                # Use the first response text as training label if needed downstream
+                for assistant_utterance in assistant_utterances:
+                    if assistant_utterance.get(TEXT) is not None:
```
tmbo, line 252:
```suggestion
                    if assistant_utterance.get(TEXT):
```
I don't think we should train on `""`

tmbo:
I am currently reviewing this as well

rasa/nlu/selectors/response_selector.py
```diff
@@ -338,21 +349,31 @@ def _full_response(self, label: Dict[Text, Any]) -> Optional[Dict[Text, Any]]:
...
         out = self._predict(message)
         label, label_ranking = self._predict_label(out)
-        retrieval_intent_name = self.retrieval_intent_mapping.get(label.get("name"))
+
+        label_retrieval_intent, label_responses = self._full_response(label)
 
         for ranking in label_ranking:
-            ranking["full_retrieval_intent"] = self.retrieval_intent_mapping.get(
-                ranking.get("name")
+            ranking["full_retrieval_intent"] = (
```
tmbo, line 373: is a `full_retrieval_intent ` the same as `example.get_combined_intent_response_key`? if so please unify the naming

rasa/nlu/selectors/response_selector.py
```diff
@@ -338,21 +349,31 @@ def _full_response(self, label: Dict[Text, Any]) -> Optional[Dict[Text, Any]]:
...
+                self.retrieval_intent_mapping.get(ranking.get("name"))
```
tmbo, line 374: why do we need to make this differentiation?

rasa/nlu/selectors/response_selector.py
```diff
@@ -366,9 +387,13 @@ def process(self, message: Message, **kwargs: Any) -> None:
         )
 
         prediction_dict = {
-            "response": self._full_response(label) or {TEXT: label.get("name")},
+            "response": {
+                "id": label["id"],
+                "name": label_responses,
```
tmbo, line 392: why should this be called "name" ?

rasa/nlu/selectors/response_selector.py
dakshvar22: But `label_attribute` is actually intended to unify them for the later part of the code, so they are just used here in this one line.

tests/nlu/selectors/test_selectors.py
```diff
@@ -66,3 +69,60 @@ def test_train_selector(pipeline, component_builder, tmpdir):
         assert rank.get("name") is not None
         assert rank.get("confidence") is not None
         assert rank.get("full_retrieval_intent") is not None
+
+
+def test_training_label():
```
tmbo, line 74: ideally, this name should tell the reader what this is testing

tests/nlu/selectors/test_selectors.py
```diff
@@ -66,3 +69,60 @@ def test_train_selector(pipeline, component_builder, tmpdir):
...
+
+    rs = ResponseSelector(component_config={TRAIN_ON_TEXT: False})
+    rs.preprocess_train_data(td)
+
+    assert rs.responses == td.responses
+    assert sorted(list(rs.index_label_id_mapping.values())) == sorted(
+        list(td.responses.keys())
+    )
+
+    rs = ResponseSelector(component_config={TRAIN_ON_TEXT: True})
```
tmbo, line 89: this looks very much like a separate test

tests/nlu/selectors/test_selectors.py
```diff
@@ -66,3 +69,60 @@ def test_train_selector(pipeline, component_builder, tmpdir):
         assert rank.get("name") is not None
         assert rank.get("confidence") is not None
         assert rank.get("full_retrieval_intent") is not None
+
+
+def test_training_label():
+
+    # use data that include some responses
+    td = load_data("data/examples/rasa/demo-rasa.md")
```
tmbo, line 77: please use more readable names (`td`, `rs`, `r`). and please don't reassign them

tests/nlu/selectors/test_selectors.py
```diff
@@ -66,3 +69,60 @@ def test_train_selector(pipeline, component_builder, tmpdir):
...
+    assert rs.responses == td.responses
+    assert sorted(list(rs.index_label_id_mapping.values())) == sorted(
+        list(td.responses.keys())
+    )
+
+    rs = ResponseSelector(component_config={TRAIN_ON_TEXT: True})
+    rs.preprocess_train_data(td)
+
+    assert rs.responses == td.responses
+    assert sorted(list(rs.index_label_id_mapping.values())) == sorted(
```
tmbo, line 93: i don't know what this tests

tests/nlu/selectors/test_selectors.py
```diff
@@ -66,3 +69,60 @@ def test_train_selector(pipeline, component_builder, tmpdir):
...
+        [r[0].get("text") for r in td.responses.values()]
+    )
+
+
+@pytest.mark.parametrize(
+    "pred_label, train_on_text, full_response_label, full_response_text",
```
tmbo, line 99: what is a `full_response_label` was that not previously a `full_response_key` or `intent_response_key` or `combined_intent_response_key`. 

the naming is quite confusing:
- what is the name of `faq/ask_name`?
- what is the name of `faq`?
- what is the name of `ask_name`?

tmbo, line 99: `faq/ask_name` also doesn't sound like a `full_response_text`

tmbo (changes requested):
thanks a lot for fixing the bug in the response selector and for implementing multi template support for responses. 

I think the code would benefit from a better naming scheme: even after working on the responses, I find the naming very confusing in the code base. e.g. it is not clear to me what the difference between `INTENT_RESPONSE_KEY` and `FULL_RETRIEVAL_INTENT` is.

rasa/nlu/selectors/response_selector.py
```diff
@@ -249,6 +253,7 @@ def model_class() -> Type[RasaModel]:
 
     def _load_selector_params(self, config: Dict[Text, Any]) -> None:
         self.retrieval_intent = config[RETRIEVAL_INTENT]
+        self.train_on_text = config[TRAIN_ON_TEXT]
```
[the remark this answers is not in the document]
dakshvar22: The call to `_load_selector_params()` is in `_check_config_parameters` which is called from the constructor of the parent class. So the property is defined inside init. Would you recommend changing it still?

changelog/6466.improvement.rst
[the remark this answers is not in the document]
dakshvar22: Already in the migration guide - https://github.com/RasaHQ/rasa/pull/6466/files#diff-a6ba427077dfb18d7f8aea2d50f40a34

rasa/nlu/selectors/response_selector.py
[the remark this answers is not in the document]
tmbo: a property should always be initialised in `__init__` (a lot of tools depend on this).

it is fine to set the value here, but `def __init__` should have a `self.train_on_text = None` in that case (or some other default value like `False`)

dakshvar22:
@tmbo I agree that the naming has got very confusing. Some of them I couldn't change so that some of the existing keys that codes outside of Rasa(example - Rasa X, Sara) could rely on doesn't change. Does it make sense to make the naming more consistent once and for all since this goes into a major release? Related to things like the confusing keys - "name", "full_retrieval_intent", INTENT_RESPONSE_KEY, etc.

tests/nlu/selectors/test_selectors.py
dakshvar22: So, `faq` is the `retrieval intent` or just `intent`(depending on the context), `ask_name` is the `response_key` and `faq/ask_name` is the `intent_response_key`.
I'll change the names in the parameters here to better reflect this naming.

tmbo:
yes, the naming should be more consistent, and as far as I can see this is also not related to the API, since you changed that to a dict anyways and moved the properties. I'd rather add that deprecated information in a separate call using the old keys just in there than have confusing variable and function names all over the place. 

I don't want this to sound overly picky, but working on this and trying to implement the multimedia responses just showed how hard it was for me to understand the response selector code.

dakshvar22:
@tmbo Thanks for the feedback. I am working on cleaning/simplifying(including naming conventions) as much code as I can for the response selector.

rasa/nlu/selectors/response_selector.py
dakshvar22: Changed it to `use_text_as_label`.

rasa/nlu/selectors/response_selector.py
dakshvar22: Refactored the whole function to be cleaner.

rasa/nlu/selectors/response_selector.py
dakshvar22: Removed the usage of this function.

rasa/nlu/selectors/response_selector.py
dakshvar22: Changed it to `response_templates`

dakshvar22:
@tmbo The PR is ready for another review. I have made the following changes since the last review - 
1. I have tried to stick to the following naming scheme - If the intent name appears as `faq/ask_name` in the training data, then the message object should have the following attributes - 
`intent` = `faq`
`response_key` = `ask_name`
`intent_response_key` = `faq/ask_name`
`response` = All the response templates under this `faq/ask_name`
1. Deprecate `full_retrieval_intent` completely from the output of the `ResponseSelector` and instead use `intent_response_key` to store the full intent name as it appears in the training data. This is more consistent with the naming scheme used in the code.
2. Deprecate `name` as the key to store the predicted response text and instead use `response_templates` to output all available response templates for the predicted `intent_response_key`. The key `name` makes more sense for the output of intent classifiers and I had just used that earlier. My bad.
3. Change the schema of `ResponseSelectionEvaluationResult` to only have [`intent_response_key_target`, `intent_response_key_prediction`, `message`, `confidence`] as its attributes. Since we are now dealing with multiple response texts for each `intent_response_key`, it did not make sense to have the `response_target`(which used to hold the text of the response) in there as the match is only based on keys now and not on the actual text.
4. Deprecate `get_combined_intent_response_key` method from `Message` class since `intent_response_key` essentially holds that already. There is a new method `get_full_intent` which returns either `intent_response_key` attribute(if it exists) or `intent` attribute.
5. Change the new parameter name in `ResponseSelector` from `train_on_text` to `use_text_as_label`.
6. A couple of model level changes to simplify which exact implementation of DIET is used in different cases depending on `use_text_as_label=True/False`.


A couple of questions that I still have which could simplify the code further maybe - 
1. For 1&2 I am not sure how to implement a graceful deprecation for them because any client code could still be querying these keys. What's the best way to do that?
2. [This part of the code](https://github.com/RasaHQ/rasa/pull/6466/files#diff-75686dc39a842033808ce0c6137c102bR363-R372) seems overly complex because of the `or` construct that's needed as `_resolve_intent_response_key` can theoretically return `None`. Practically, if that happens there is already a problem in the training data which the user is warned about [here](https://github.com/RasaHQ/rasa/pull/6466/files#diff-0e775499654bc8adf9290d502dac3c0dR427-R433). Do you recommend getting around this in some other way?

docs/docs/migration-guide.mdx
```diff
@@ -341,6 +341,12 @@ Please see [Forms](./forms.mdx) if you have further customizations in your
 
 <a aria-hidden="true" tabIndex="-1" className="anchor enhancedAnchor" id="migration-to-rasa-1-8"></a>
 
+### Response Selectors
+
+Response Selectors are now trained on retrieval intent labels by default instead of the actual response text. For most models, this should improve training time and accuracy of the `ResponseSelector`.
+
+If you want to revert to the pre-2.0 default behavior, add the `use_text_as_label=true` parameter to your `ResponseSelector` component.
+
```
erohmensing, line 349: The output schema/api change should be in here too!

dakshvar22: I just added it :)

dakshvar22: Ohh, you meant in the migration guide. Adding it now.

rasa/nlu/selectors/response_selector.py
```diff
@@ -244,30 +256,20 @@ def label_key(self) -> Text:
         return LABEL_IDS
 
     @staticmethod
-    def model_class() -> Type[RasaModel]:
-        return DIET2DIET
+    def model_class(use_text_as_label: bool) -> Type[RasaModel]:
+        if use_text_as_label:
+            return DIET2DIET
+        else:
+            return DIET2BOW
```
dakshvar22, line 263: This and the almost the whole of `DIET2BOW` class was needed because metrics are tightly coupled with the model architecture right now. See https://github.com/RasaHQ/research/issues/112

rasa/nlu/selectors/response_selector.py
```diff
@@ -204,6 +210,9 @@ def required_components(cls) -> List[Type[Component]]:
         MASKED_LM: False,
         # Name of the intent for which this response selector is to be trained
         RETRIEVAL_INTENT: None,
+        # Boolean flag to check if actual text of the response
+        # should be used as ground truth label for training the model.
+        USE_TEXT_AS_LABEL: True,
```
dakshvar22, line 215: Reminder for myself: Change this to False before merging. This is `True` for model running model regression tests once.

docs/docs/migration-guide.mdx
```diff
@@ -341,6 +341,75 @@ Please see [Forms](./forms.mdx) if you have further customizations in your
...
+* If you want to revert to the pre-2.0 default behavior, add the `use_text_as_label=true`
+ parameter to your `ResponseSelector` component.
+
+* All response selector related evaluation artifacts - `report.json, successes.json,
+errors.json, confusion_matrix.png` now use the sub-intent of the retrieval intent as
+ the target and predicted labels instead of the actual response text.
+This means that the output in these files should look like -
+
+`response_selection_successes.json`
+```json
```
tmbo, line 359:
```suggestion
```json title="response_selection_successes.json"
```

docs/docs/migration-guide.mdx
```diff
@@ -341,6 +341,75 @@ Please see [Forms](./forms.mdx) if you have further customizations in your
...
+      "intent_response_key_target": "faq/is_legit",
+      "intent_response_key_prediction": {
+              "name": "faq/is_legit",
+              "confidence": 0.88
+      }
+}
+```
+
+`response_selection_errors.json`
+```json
```
tmbo, line 371:
```suggestion
```json title="response_selection_errors.json"
```

examples/responseselectorbot/credentials.yml
```diff
@@ -0,0 +1,32 @@
+twilio:
```
tmbo, line 1: I don't think these should be in here

examples/responseselectorbot/data/responses.yml
```diff
@@ -0,0 +1,12 @@
...
+responses:
+  chitchat/ask_name:
+  - image: "https://i.imgur.com/zTvA58i.jpeg"
+    text: hello, my name is retrieval bot.
+  - text: Oh yeah, I am called the retrieval bot.
+
+  chitchat/ask_weather:
+  - text: Oh, it does look sunny right now in Berlin.
+    image: "https://i.imgur.com/vwv7aHN.png"
+  - text: I am not sure of the whole week but I can see the sun is out today.
```
tmbo, line 12:
```suggestion
  - text: I am not sure of the whole week but I can see the sun is out today.

```

tests/core/test_actions.py
```diff
@@ -363,10 +370,12 @@ async def test_action_utter_retrieved_response(
     default_tracker.latest_message = UserMessage(
         "Who are you?",
         parse_data={
-            "response_selector": {
+            RESPONSE_SELECTOR_PROPERTY_NAME: {
```
tmbo, line 373: optional: I am not sure we should extensively use constants in the tests:
- it makes them less readable
- it tightly couples the test to the implementation: the test simulates an API user, if we use the same constants here as we do in the code, it is less obvious that changing that constant breaks that API (as the test will still pass)

examples/responseselectorbot/credentials.yml
dakshvar22: @tmbo Do you mean the actual credentials shouldn't be in here and we should put some dummy credentials?
I actually assumed these are dummy credentials that don't work because I took them from the moodbot example. We should change it there too if these are actually working credentials.

tests/core/test_actions.py
dakshvar22: Makes sense! I'll change them back to the actual strings.

examples/responseselectorbot/credentials.yml
dakshvar22: I see what you mean. We are not using channels in the example, so the file `credentials.yml` isn't needed even.

PR #6470, opened 2020-08-21
Switch to mypy part0

m-vdb:
## Summary of the first run

Found 812 errors in 136 files (checked 227 source files). Detailed text logs can be found [here](https://gist.github.com/m-vdb/94216371b17bc1a3918c79255d39684b).

### Errors by type

- [`[arg-type]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-argument-types-arg-type): 261 errors
- [`[assignment]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-types-in-assignment-statement-assignment): 146 errors ⚠️ 
- [`[var-annotated]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#require-annotation-if-variable-type-is-unclear-var-annotated): 59 errors
- [`[return-value]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-return-value-is-compatible-return-value): 53 errors
- [`[union-attr]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-attribute-exists-in-each-union-item-union-attr): 52 errors  ⚠️ 
- [`[override]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-validity-of-overrides-override): 47 errors
- [`[operator]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-uses-of-various-operators-operator): 41 errors  ⚠️ 
- [`[attr-defined]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-attribute-exists-attr-defined): 39 errors  ⚠️ 
- [`[index]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-indexing-operations-index): 31 errors  ⚠️ 
- [`[misc]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#miscellaneous-checks-misc): 21 errors
- [`[return]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-function-returns-a-value-return): 19 errors
- [`[call-arg]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-arguments-in-calls-call-arg): 17 errors  ⚠️ 
- [`[type-var]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-type-variable-values-type-var): 5 errors
- [`[list-item]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-list-items-list-item): 5 errors  ⚠️ 
- [`[has-type]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-type-of-target-is-known-has-type): 5 errors
- [`[valid-type]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-validity-of-types-valid-type): 4 errors
- [`[dict-item]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-dict-items-dict-item): 4 errors  ⚠️ 
- [`[no-redef]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-each-name-is-defined-once-no-redef): 2 errors
- [`[func-returns-value]`](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-called-function-returns-a-value-func-returns-value): 1 errors

### Errors by module:

- `rasa/core/`: 511 errors
- `rasa/nlu/`: 240 errors
- `rasa/utils/`: 50 errors
- `rasa/cli/`: 43 errors
- `rasa/train.py`: 18 errors
- `rasa/importers/`: 14 errors
- `rasa/test.py`: 13 errors
- `rasa/server.py`: 7 errors
- `rasa/model.py`: 7 errors
- `rasa/run.py`: 5 errors
- `rasa/validator.py`: 2 errors
- `rasa/jupyter.py`: 2 errors
- `rasa/constants.py`: 1 errors

### Files with more that 10 errors

- `rasa/core/training/interactive.py`: 39 errors
- `rasa/nlu/classifiers/diet_classifier.py`: 34 errors
- `rasa/core/training/structures.py`: 31 errors
- `rasa/core/training/generator.py`: 28 errors
- `rasa/core/featurizers.py`: 26 errors
- `rasa/core/tracker_store.py`: 24 errors
- `rasa/cli/train.py`: 23 errors
- `rasa/core/actions/forms.py`: 20 errors
- `rasa/core/actions/action.py`: 19 errors
- `rasa/train.py`: 18 errors
- `rasa/nlu/selectors/response_selector.py`: 18 errors
- `rasa/core/domain.py`: 18 errors
- `rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py`: 17 errors
- `rasa/core/policies/ensemble.py`: 16 errors
- `rasa/nlu/training_data/loading.py`: 15 errors
- `rasa/core/agent.py`: 15 errors
- `rasa/core/slots.py`: 14 errors
- `rasa/test.py`: 13 errors
- `rasa/core/events/__init__.py`: 13 errors
- `rasa/utils/tensorflow/models.py`: 12 errors
- `rasa/core/trackers.py`: 12 errors
- `rasa/core/test.py`: 12 errors
- `rasa/nlu/test.py`: 11 errors
- `rasa/nlu/components.py`: 11 errors
- `rasa/core/policies/mapping_policy.py`: 11 errors
- `rasa/core/policies/ted_policy.py`: 10 errors
- `rasa/core/channels/console.py`: 10 errors

## Strategy

1. 👉 Find a way to keep `mypy` and `pytype` side by side during the transition.
2. Current files with errors are ignored by mypy in `setup.cfg` (see 0d35a59). This makes sure we still type check on master when we merge this branch.
3. We'll work our way through the list of errors and split them in different tasks. 
I've added a  ⚠️ in front of the type errors that could yield to immediate errors and I think we should fix them first. Suggested approach:
  i. Group the error types with small numbers of errors together and tackle them into a single issue (<30 errors)
  ii. Treat error types with high number of errors in separate PRs (>30 errors)
  iii. Treat `[arg-type]` and `[assignment]` differently: we'll most probably do them in different passes. We might want to split by module in this case.

💡 Note: it doesn't seem to be possible to disable a list of error types in the config, it only works on a line by line basis. It would have been more convenient, but that's OK. Running `make types | grep -e "\[error-code\]"` does what you want and is really fast.


m-vdb:
@tmbo I've done an analysis of the type errors that we have in Rasa and suggested an approach. WDYT? I'd also like to share this to backend engineers to get feedback, they might have ideas (although some of them are off this week)

tmbo:
the approach sounds good :+1: thanks a lot for looking into this, replacing pytype will be a great way to speed up things on the CI but also in local development 🚀

tmbo:
I guess this will allow us to enable an error once we fixed the existing issues to avoid PRs introducing new errors right? (just double checking)

README.md
```diff
@@ -251,7 +251,7 @@ make formatter
 
 #### Type Checking
 
-If you want to check types on the codebase, install `pytype` using `poetry install`.
+If you want to check types on the codebase, install `mypy` using `poetry install`.
```
tmbo, line 254: should we keep the pytype check until we made enough progress on mypy? I am bit worried about the period of time where we disabled a lot of mypy checks and are also not running pytype

tmbo:
did you already discuss the switch with the backend team to make sure everyone is onboard?

README.md
m-vdb: yes this is a concern I have too 👍 I will see if there are issues if we keep both of them at the same time (for instance, there might be problems with `# pytype: disable` comments)

m-vdb:
@tmbo no I haven't, I wanted to get your green light for the plan. Will communicate today

m-vdb:
> I guess this will allow us to enable an error once we fixed the existing issues to avoid PRs introducing new errors right? (just double checking)

It's not perfect, as `mypy` config doesn't let you disable/enable all the error codes. But it lets you ignore files, which is why I included a list of files inside `setup.cfg`. And if you check the first PR I made to correct some of the type errors, I've updated the `setup.cfg` as I went: https://github.com/RasaHQ/rasa/pull/6511

wochinge:
> as mypy config doesn't let you disable/enable all the error codes.

this means handling grouped errors by type doesn't work? 

I like the iterative approach of doing it on a file per file base a lot. Imo the order of which we tackle them isn't critical as we have managed so far without having any of these resolved (assuming that we not stretch the transition for too long)

m-vdb:
@wochinge it doesn't work at the moment BUT happy news: I just saw that it's available on `mypy`'s master, so I've asked them if we could have a release on pypi [in this issue](https://github.com/python/mypy/issues/8975) 🚀  That would be awesome if we could use it to disable error codes one by one. That way we can do it iteratively both on the error codes and the files (there are files with a lot of errors, so it will help in these cases). cc @tmbo

README.md
m-vdb: 17f125d ✅

m-vdb:
Quick status update: [next `mypy` release is scheduled for the end of the week](https://github.com/python/mypy/issues/9290#issuecomment-703863780). After that, we'll be to enable / disable error codes in a more granular way! 🎉

m-vdb:
aaaaaaand disabling error codes one by one: 85c785d

pyproject.toml
```diff
@@ -152,6 +151,8 @@ toml = "^0.10.0"
 pep440-version-utils = "^0.3.0"
 pydoc-markdown = "^3.5.0"
 pytest-timeout = "^1.4.2"
+pytype = "^2020.6.1"
```
wochinge, line 154: should we add a comment here why we are keeping both for now?

rasa/utils/tensorflow/models.py
```diff
@@ -666,7 +666,7 @@ def __init__(
         )
 
         # set up tf layers
-        self._tf_layers: Dict[Text : tf.keras.layers.Layer] = {}
+        self._tf_layers: Dict[Text, tf.keras.layers.Layer] = {}
```
wochinge, line 669: 🤯

setup.cfg
```diff
@@ -19,3 +19,10 @@ log_cli_level = WARNING
 [flake8]
 max-line-length = 88
 ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, E231
+
+[mypy]
+ignore_missing_imports = True
```
wochinge, line 24: Why is this needed / is it worth adding a comment for it?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1115,7 +1115,9 @@ def __init__(
         self._create_metrics()
         self._update_metrics_to_log()
 
-        self.all_labels_embed = None  # needed for efficient prediction
+        # needed for efficient prediction
+        self.all_labels_embed: Optional[Tuple[tf.Tensor, tf.Tensor]] = None
```
wochinge, line 1119: judging based on [this line](https://github.com/RasaHQ/rasa/blob/077d5dfa6140771113cd06dffd068d2c3d36e446/rasa/nlu/classifiers/diet_classifier.py#L1814) I'd say it should be 
```suggestion
        self.all_labels_embed: Optional[tf.Tensor] = None
```

rasa/core/policies/ted_policy.py
```diff
@@ -541,7 +541,8 @@ def __init__(
         self.action_acc = tf.keras.metrics.Mean(name="acc")
         self.metrics_to_log += ["loss", "acc"]
 
-        self.all_labels_embed = None  # needed for efficient prediction
+        # needed for efficient prediction
+        self.all_labels_embed: Optional[Tuple[tf.Tensor, tf.Tensor]] = None
```
wochinge, line 545: judging based on [this line](https://github.com/RasaHQ/rasa/blob/077d5dfa6140771113cd06dffd068d2c3d36e446/rasa/nlu/classifiers/diet_classifier.py#L1814) I'd say it should be 
```suggestion
        self.all_labels_embed: Optional[tf.Tensor] = None
```

CONTRIBUTING.md
```diff
@@ -85,15 +85,15 @@ If your PR is greater than 500 lines, please consider splitting it into multiple
...
+To ensure a standardized code style we recommend using formatter black. To ensure our type annotations are correct we also suggest using the type checker `mypy`.
 
 #### Formatting and Type Checking
 
 If you want to automatically format your code on every commit, you can use pre-commit. Just install it via `pip install pre-commit` and execute `pre-commit install` in the root folder. This will add a hook to the repository, which reformats files on every commit.
 
 If you want to set it up manually, install black via `pip install -r requirements-dev.txt.` To reformat files execute `make formatter`.
 
-If you want to check types on the codebase, install pytype using `pip install -r requirements-dev.txt`. To check the types execute `make types`.
+If you want to check types on the codebase, install `mypy` using `pip install -r requirements-dev.txt`. To check the types execute `make types`.
```
wochinge, line 96: The `requirements-dev.txt` file doesn't exist anymore
```suggestion
If you want to check types on the codebase, install `mypy` using `poetry install`. To check the types execute `make types`.
```

wochinge (commented):
Excited to have a working type checker! Thanks for driving this!

rasa/utils/tensorflow/models.py
tmbo: it is very surprising that pytype accepts this 😂

.github/workflows/continous-integration.yml
```diff
@@ -102,6 +102,9 @@ jobs:
     - name: Check Types 📚
       run: make types
 
+    - name: Check Types (old) 📚
```
tmbo, line 105: do we need this if we are able to disable by error code?

setup.cfg
m-vdb: It's because there are modules without type hints or library stubs. I can add a comment

rasa/nlu/classifiers/diet_classifier.py
m-vdb: yes that's an issue I fixed in the other PR, don't know why I didn't fix it in this PR 🤷

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "poetry.masonry.api"
 [tool.black]
 line-length = 88
 target-version = [ "py36", "py37", "py38",]
-exclude = "((.eggs | .git | .pytype | .pytest_cache | build | dist))"
+exclude = "((.eggs | .git | .pytest_cache | build | dist))"
```
wochinge, line 8: do we also need to exclude the `mypy` cache here?

rasa/core/channels/rocketchat.py
```diff
@@ -108,13 +108,14 @@ def from_credentials(cls, credentials: Optional[Dict[Text, Any]]) -> InputChanne
...
-        # pytype: disable=attribute-error
+        # rasa-16-to-rasa-17
         return cls(
             credentials.get("user"),
             credentials.get("password"),
             credentials.get("server_url"),
         )
-        # pytype: enable=attribute-error
+
+    # rasa-16-to-rasa-17
```
wochinge, line 118: Does it work like a `begin ignore` / `end ignore` or can we just drop the 2nd `#rasa-16-to-rasa-17`?

wochinge (commented):
What is `rasa-16-to-rasa-17` for?

rasa/core/channels/rocketchat.py
m-vdb: I have no idea what this is 🤷

pyproject.toml
m-vdb: there is no need, since mypy cache contains json files instead python files (which was the case for pytype)

rasa/core/channels/rocketchat.py
wochinge: Ok, then let's remove the new version and old version of the comments?

m-vdb: soooo, apparently I introduced it myself in https://github.com/RasaHQ/rasa/pull/6470/commits/7fa00f42d5621b4ae10a8d179c0892b0f031929c... I've reverted and fixed it

rasa/core/actions/loops.py
```diff
@@ -43,13 +43,12 @@ async def is_activated(
         tracker: "DialogueStateTracker",
         domain: "Domain",
     ) -> bool:
-        # pytype: disable=attribute-error
+
```
wochinge, line 46:
```suggestion
```

rasa/core/channels/botframework.py
```diff
@@ -152,7 +152,7 @@ async def send_elements(
     async def send_custom_json(
         self, recipient_id: Text, json_message: Dict[Text, Any], **kwargs: Any
     ) -> None:
-        # pytype: disable=attribute-error
+
```
wochinge, line 155:
```suggestion
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1561,12 +1563,11 @@ def _calculate_entity_loss(
         # should call first to build weights
         pred_ids, _ = self._tf_layers[f"crf.{tag_name}"](logits, sequence_lengths)
         # pytype cannot infer that 'self._tf_layers["crf"]' has the method '.loss'
```
wochinge, line 1565:
```suggestion
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1817,15 +1818,14 @@ def _batch_predict_intents(
 
         # pytype cannot infer that 'self._tf_layers[f"loss.{LABEL}"]' has methods
         # like '.sim' or '.confidence_from_sim'
```
wochinge, line 1820:
```suggestion
```

rasa/shared/core/training_data/structures.py
```diff
@@ -168,20 +168,18 @@ def as_story_string(self, flat: bool = False, e2e: bool = False) -> Text:
     def is_action_listen(event: Event) -> bool:
         # this is not an `isinstance` because
         # we don't want to allow subclasses here
-        # pytype: disable=attribute-error
+
```
wochinge, line 171:
```suggestion
```

rasa/shared/core/training_data/structures.py
```diff
@@ -168,20 +168,18 @@ def as_story_string(self, flat: bool = False, e2e: bool = False) -> Text:
...
         return type(event) == ActionExecuted and event.action_name == ACTION_LISTEN_NAME
-        # pytype: enable=attribute-error
 
     @staticmethod
     def is_action_session_start(event: Event) -> bool:
         # this is not an `isinstance` because
         # we don't want to allow subclasses here
-        # pytype: disable=attribute-error
+
```
wochinge, line 178:
```suggestion
```

rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -171,14 +171,13 @@ def _parse_training_examples(
...
                 (
-                    # pytype: disable=attribute-error
                     example.get(KEY_INTENT_TEXT, "").strip(STRIP_SYMBOLS),
                     example.get(KEY_METADATA),
                 )
                 for example in examples
                 if example
             ]
-        # pytype: enable=attribute-error
+
```
wochinge, line 180:
```suggestion
```

rasa/shared/nlu/training_data/message.py
```diff
@@ -126,9 +126,9 @@ def build(
         if intent_metadata is not None:
             data[METADATA] = {METADATA_INTENT: intent_metadata}
         if example_metadata is not None:
-            # pytype: disable=unsupported-operands
+
```
wochinge, line 129:
```suggestion
```

wochinge (approved):
👍 

I found 4 more `# pytype: disable=` comments in `ted_policy.py` and `diet_classifier.py` which should also be removed please.

PR #6478, opened 2020-08-24
2.0 training data format improvement

### LINKED ISSUE #6361: Change training data version warning to be a debug log
In the new training data, you can put a `version` key in the files. If you don't do this, you will see warnings like:
`UserWarning: Training data file /Users/christina/rasa/rasa/cli/initial_project/domain.yml doesn't have a 'version' key. Rasa Open Source will read the file as a version '2.0' file.` (originating from [here](https://github.com/RasaHQ/rasa/blob/219baa461c066a199fb4fb10a123012ebc8182b3/rasa/validator.py#L294))

I'd suggest to not raise those warnings, but rather put them into a debug log.

My reasoning is this (see also the comments in this PR: https://github.com/RasaHQ/rasa/pull/6329#pullrequestreview-462397138):
Adding the version is not expected but rather an optional feature that people can use to get warned about incompatibilities with their Rasa Open Source version. (Or am I mistaken here?) The warnings seem to indicate that something is a problem, which is not the case.
Also, if we want to avoid that users see multiple warnings for each training data file when running `rasa init`, we would have to add a `version` key to the init project, too and manually maintain that to stay up to date.

degiz:
Closes #6361 

**Proposed changes**:
- Don't raise a warning in case `version` key is not specified in the training data

### DIFF
--- a/tests/nlu/training_data/formats/test_rasa_yaml.py
+++ b/tests/nlu/training_data/formats/test_rasa_yaml.py
@@ -3,13 +3,14 @@ from typing import Text
 
 import pytest
 
-from rasa.core.domain import InvalidDomain
 import rasa.utils.io as io_utils
 from rasa.constants import LATEST_TRAINING_DATA_FORMAT_VERSION
+from rasa.core.domain import InvalidDomain
 from rasa.nlu.constants import INTENT
 from rasa.nlu.training_data.formats.rasa_yaml import RasaYAMLReader, RasaYAMLWriter
 
 MULTILINE_INTENT_EXAMPLES = f"""
+version: "{LATEST_TRAINING_DATA_FORMAT_VERSION}"
 nlu:
 - intent: intent_name
   examples: |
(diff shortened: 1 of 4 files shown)


tests/nlu/training_data/formats/test_rasa_yaml.py
```diff
@@ -10,7 +10,6 @@
 from rasa.nlu.training_data.formats.rasa_yaml import RasaYAMLReader, RasaYAMLWriter
 
 MULTILINE_INTENT_EXAMPLES = f"""
-version: "{LATEST_TRAINING_DATA_FORMAT_VERSION}"
```
chkoss, line 13: I'd leave the version in here or add one more test example with a version, because the reader needs to be able to properly read the version, too.

tests/nlu/training_data/formats/test_rasa_yaml.py
```diff
@@ -10,7 +10,6 @@
 from rasa.nlu.training_data.formats.rasa_yaml import RasaYAMLReader, RasaYAMLWriter
 
 MULTILINE_INTENT_EXAMPLES = f"""
```
chkoss, line 12: This is just a normal string instead of f-string now (same for all the other examples below).

chkoss (changes requested):
Thanks! 🎉

I have a few points that don't belong to a specific line but to the tests in 'test_rasa_yaml' in general.

- Currently, `EXAMPLE_NO_VERSION_SPECIFIED` is not used, neither is the imported `LATEST_TRAINING_DATA_FORMAT_VERSION`, so they can be removed.
- There is a failing test (`test_rasa_yaml:test_yaml_examples_are_written`), because the dumped training data includes the version while the input doesn't necessarily.
- We don't have any test case to check that training data with version specified is read and dumped properly.

So I think we should add at least one example including a version, and the dumping test could be changed e.g. like this:
```python
def test_yaml_examples_are_written(example: Text):
    parser = RasaYAMLReader()
    writer = RasaYAMLWriter()

    training_data = parser.reads(example)

    if ... :  # check if there is a version in the example - not sure how to do this best
        expected_output = example
    else:
        expected_output = f"""
- version: "{LATEST_TRAINING_DATA_FORMAT_VERSION}"
{example}"
"""

    assert expected_output.strip() == writer.dumps(training_data).strip()
```

An alternative would be to just leave all the `test_rasa_yaml` examples and tests as they were (including the versions) and adjust only the `test_no_version_specified_raises_warning` to be a `test_no_version_specified` that checks if reading the data without version does the same as reading data with the latest version.

Wdyt?

degiz:
@chkoss good point, so what I did:
- Left few examples with the `LATEST_TRAINING_DATA_FORMAT_VERSION`, so that we know it's parsed
- `test_yaml_examples_are_written` now implicitly checks that the `version` is also dumped
- `test_minimal_valid_example` verifies that we don't raise warning in case the `version` is not specified

tests/nlu/training_data/formats/test_rasa_yaml.py
```diff
@@ -18,32 +18,22 @@
     - how much carbon will a one way flight from [new york]{{"entity": "city", "role": "from"}} to california produce?
```
chkoss, line 18:
```suggestion
    - how much carbon will a one way flight from [new york]{"entity": "city", "role": "from"} to california produce?
```

chkoss (approved):
Nice, looks great! 😄 

Maybe you could still add a comment to `test_yaml_examples_are_written` that only examples including a version number can be used for this test, to avoid future confusion.

tests/nlu/training_data/formats/test_rasa_yaml.py
degiz: Why? It's an `f-string`, I dont' think that would work.

chkoss: Ah sorry, my mistake!

degiz: no worries 😄

PR #6485, opened 2020-08-25
only suggest config that is required when doing 'rasa train nlu/core'

### LINKED ISSUE #6456: autoconfig chooses policies when training nlu only
When training nlu:
```
new ❯ rasa train nlu
The configuration for policies was chosen automatically. It was written into the config file at 'config.yml'.
```
I wouldn't expect it to choose policies for me because I am not using them in my model. My expectations for the opposite would be the same: if only training core, I wouldn't expect it to give me an NLU pipeline.

chkoss:
Fix #6456 

**Proposed changes**:
- When running `rasa train nlu`, only `pipeline` is suggested and not `policies`.
- When running `rasa train core`, only `policies` is suggested and not `pipeline`.
- As a side effect, for any custom importer there is now also the option to take an argument `nlu_or_core`. This argument will have the value `"core"` when running `rasa train core`, `"nlu"` when running `rasa train nlu` and `"both"` when running `rasa train`. So you can now choose to treat these cases differently in your custom importer.

**TODO**:
- [x] add tests for both `rasa train nlu` and `rasa train core` to make sure this is done correctly
- [x] write something in the docs about the new feature for custom importers


alwx:
This branch was refactored, some of the things were changed completely (like the usage of enums instead of strings), and tests were also added.

rasa/importers/importer.py
```diff
@@ -161,9 +167,12 @@ def _importer_from_dict(
                 logging.warning(f"Importer '{module_path}' not found.")
                 return None
 
+        importer_config["training_type"] = training_type
```
m-vdb, line 170: this line updates the `importer_config` dictionary, could it lead to unpredictable consequences, and would something like `importer_config = dict(training_type=training_type, **importer_config)` be more suitable? (or in 2 lines, copy() + update)

m-vdb (approved):
I did a few tests locally and it seems to work fine 💯 

I've left a minor comment and another one that I think we should address

PR #6489, opened 2020-08-26
E2E Featurization

tabergma:
**Proposed changes**:
- update `RasaModelData` format to better support e2e features
- update `TEDPolicy` and `DIETClassifier` to handle new `RasaModelData` format
- update core featurizers to produce `Features`


rasa/core/policies/policy.py
```diff
@@ -134,24 +136,31 @@ def featurize_for_training(
...
+            - the label ids (e.g. action ids) for every dialuge turn in all training
+              trackers
         """
 
-        training_data = self.featurizer.featurize_trackers(
+        X, label_ids = self.featurizer.featurize_trackers(
             training_trackers, domain, interpreter
         )
 
+        # TODO: does this still fit?
```
tabergma, line 153: @Ghostvv Does it make sense to just take the first x features from the list of list of list of features? Or do we need to change the logic here?

examples/formbot/config.yml
```diff
@@ -17,3 +17,4 @@ policies:
   - name: TEDPolicy
   - name: AugmentedMemoizationPolicy
   - name: RulePolicy
+  - name: TEDPolicy
```
Ghostvv, line 20: there is TED already
```suggestion

```

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -0,0 +1,246 @@
...
+        if not domain.action_names == []:
+            self._default_feature_states[ACTION_NAME] = convert_to_dict(
+                domain.action_names
+            )
+        if not domain.entities == []:
+            self._default_feature_states[ENTITIES] = convert_to_dict(domain.entities)
+        if not domain.slot_states == []:
+            self._default_feature_states[SLOTS] = convert_to_dict(domain.slot_states)
+        if not domain.form_names == []:
+            self._default_feature_states[FORM] = convert_to_dict(domain.form_names)
```
Ghostvv, line 51:
```suggestion
            self._default_feature_states[ACTIVE_LOOP] = convert_to_dict(domain.form_names)
```

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -0,0 +1,246 @@
...
+                    self._extract_features(
+                        sub_state, state_type, interpreter, sparse=True
+                    )
+                )
+            if state_type == USER:
+                if sub_state.get(ENTITIES):
+                    featurized_state.update(
+                        self._create_features(sub_state, ENTITIES, sparse=True)
+                    )
+            if state_type in {SLOTS, FORM}:
```
Ghostvv, line 183:
```suggestion
            if state_type in {SLOTS, ACTIVE_LOOP}:
```

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -0,0 +1,342 @@
...
+            )
+
+        self.state_featurizer.prepare_from_domain(domain)
+
+        trackers_as_states, trackers_as_actions = self.training_states_and_actions(
+            trackers, domain
+        )
+
+        # noinspection PyPep8Naming
+        X = self._featurize_states(trackers_as_states, interpreter)
```
Ghostvv, line 110: let's rename `X` since it is not actual numeric data anymore

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -0,0 +1,342 @@
...
+        self, trackers: List[DialogueStateTracker], domain: Domain
+    ) -> List[List[State]]:
+        """Transforms list of trackers to lists of states for prediction."""
+
+        raise NotImplementedError(
+            "Featurizer must have the capacity to create feature vector"
+        )
+
+    # noinspection PyPep8Naming
+    def create_X(
```
Ghostvv, line 125: same here, let's rename `X` here as well

rasa/core/policies/policy.py
```diff
@@ -134,24 +136,31 @@ def featurize_for_training(
...
-            the :class:`rasa.core.training.data.DialogueTrainingData`
+            - a dictionary of attribute (INTENT, TEXT, ACTION_NAME, ACTION_TEXT,
+              ENTITIES, SLOTS, FORM) to a list of features for all dialogue turns in
+              all training trackers
+            - the label ids (e.g. action ids) for every dialuge turn in all training
+              trackers
         """
 
-        training_data = self.featurizer.featurize_trackers(
+        X, label_ids = self.featurizer.featurize_trackers(
```
Ghostvv, line 149: let's rename `X`, since it is not pure numeric anymore

rasa/core/policies/policy.py
Ghostvv: what do you mean? we take first x examples

rasa/core/policies/ted_policy.py
```diff
@@ -1,35 +1,34 @@
...
-    E2ESingleStateFeaturizer,
 )
-from rasa.nlu.constants import ACTION_NAME, INTENT, ACTION_TEXT
-from rasa.core.interpreter import NaturalLanguageInterpreter, RegexInterpreter
+from rasa.core.featurizers.single_state_featurizer import SingleStateFeaturizer
+from rasa.nlu.constants import ACTION_NAME, INTENT, ACTION_TEXT, TEXT, ENTITIES
+from rasa.core.interpreter import NaturalLanguageInterpreter
 from rasa.core.policies.policy import Policy
-from rasa.core.constants import DEFAULT_POLICY_PRIORITY, DIALOGUE
+from rasa.core.constants import DEFAULT_POLICY_PRIORITY, DIALOGUE, FORM, SLOTS
```
Ghostvv, line 26:
```suggestion
from rasa.core.constants import DEFAULT_POLICY_PRIORITY, DIALOGUE, ACTIVE_LOOP, SLOTS
```

rasa/core/policies/ted_policy.py
```diff
@@ -65,14 +65,26 @@
...
+    from rasa.utils.features import Features
+
+
 logger = logging.getLogger(__name__)
 
-DIALOGUE_FEATURES = f"{DIALOGUE}_features"
-LABEL_FEATURES = f"{LABEL}_features"
-LABEL_IDS = f"{LABEL}_ids"
+MASK = "mask"
+LABEL_KEY = LABEL
```
Ghostvv, line 82: can we use `LABEL` directly?

rasa/core/policies/ted_policy.py
```diff
@@ -228,276 +248,311 @@ def _load_params(self, **kwargs: Dict[Text, Any]) -> None:
     # data helpers
     # noinspection PyPep8Naming
     @staticmethod
-    def _label_ids_for_Y(data_Y: np.ndarray) -> np.ndarray:
-        """Prepare Y data for training: extract label_ids.
-        label_ids are indices of labels, while `data_Y` contains one-hot encodings.
+    def _surface_attributes(
+        features: List[List[Dict[Text, List["Features"]]]]
```
Ghostvv, line 252: there is many different types of `features` depending where `Dict` is, let's try to create names that reflect that

rasa/core/policies/ted_policy.py
```diff
@@ -196,6 +213,7 @@ def __init__(
         priority: int = DEFAULT_POLICY_PRIORITY,
         max_history: Optional[int] = None,
         model: Optional[RasaModel] = None,
+        zero_features: Optional[Dict[Text, List["Features"]]] = None,
```
Ghostvv, line 216: is this rather `zero_example`? It contains `zero_features` for several attributes?

rasa/core/policies/ted_policy.py
```diff
@@ -509,16 +564,13 @@ def train(
         """Train the policy on given training trackers."""
 
         # dealing with training data
-        training_data = self.featurize_for_training(
+        X, label_ids = self.featurize_for_training(
```
Ghostvv, line 567: let's rename `X`

rasa/core/policies/ted_policy.py
```diff
@@ -604,7 +656,11 @@ def persist(self, path: Text) -> None:
             model_path / f"{SAVE_MODEL_FILE_NAME}.data_example.pkl", self.data_example
         )
         io_utils.pickle_dump(
-            model_path / f"{SAVE_MODEL_FILE_NAME}.label_data.pkl", self._label_data
+            model_path / f"{SAVE_MODEL_FILE_NAME}.zero_features.pkl", self.zero_features
```
Ghostvv, line 659: is `self.zero_features` is basically equivalent to `self.data_example` but with zeros?

rasa/core/policies/ted_policy.py
```diff
@@ -654,264 +716,181 @@ def load(cls, path: Text) -> "TEDPolicy":
...
-                else:
-                    _f = f
-                dense_features.append(self._tf_layers[f"sparse_to_dense.{name}"](_f))
+    def _prepare_utterance_level_layers(self, name: Text) -> None:
+        for feature_type in POSSIBLE_FEATURE_TYPES:
+            if (
+                name not in self.data_signature
+                or feature_type not in self.data_signature[name]
+            ):
+                continue
```
Ghostvv, line 822: why is continue here?

rasa/core/policies/ted_policy.py
```diff
@@ -654,264 +716,181 @@ def load(cls, path: Text) -> "TEDPolicy":
...
-            if key.startswith(LABEL_FEATURES):
-                all_labels.append(
-                    self._combine_sparse_dense_features(self.tf_label_data[key], key)
+            mask = None
+            if MASK in self.tf_label_data[key]:
+                mask = self.tf_label_data[key][MASK][0]
+
+            for sub_key in self.tf_label_data[key].keys():
+                if sub_key not in [SEQUENCE, SENTENCE]:
+                    continue
```
Ghostvv, line 865: could you please add comment why continue here

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -469,38 +470,30 @@ def _extract_features(
             sparse_sequence_features = None
             dense_sequence_features = None
 
-        return (
-            sparse_sequence_features,
-            sparse_sentence_features,
-            dense_sequence_features,
-            dense_sentence_features,
-        )
+        out = {}
```
Ghostvv, line 473: does it makes sense to make `out` dict of dict as well

rasa/utils/tensorflow/model_data.py
```diff
@@ -45,40 +61,61 @@ def __init__(
...
+        self.data = data or defaultdict(lambda: defaultdict(list))
         self.label_key = label_key
+        self.label_sub_key = label_sub_key
         # should be updated when features are added
         self.num_examples = self.number_of_examples()
 
-    def get_only(self, key: Text) -> Optional[np.ndarray]:
-        if key in self.data:
-            return self.data[key][0]
+    def get_only(self, key: Text, sub_key: Text) -> Optional[np.ndarray]:
```
Ghostvv, line 70: does it get array only if it's singular list? let's add the check then

rasa/nlu/classifiers/diet_classifier.py
tabergma: yes, that is on my list, but as it is a bit more complicated I postponed it for now

rasa/core/policies/ted_policy.py
tabergma: do you have any suggestions?

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -0,0 +1,246 @@
...
+                for i, value in enumerate(slot_as_feature)
+            }
+        else:
+            raise ValueError(
+                f"Given attribute '{attribute}' is not supported. "
+                f"It must be one of '{self._default_feature_states.keys()}'."
+            )
+
+        if attribute not in self._default_feature_states:
+            return
```
Ghostvv, line 102: return None doesn't correspond to return type

rasa/utils/tensorflow/model_data.py
tabergma: this is given from the data structure of `self.data`, which is `Dict[Text, Dict[Text, List[np.ndarray]]]`. So the first example of the list is always a `np.ndarray`

rasa/core/policies/ted_policy.py
Ghostvv: in single state featurizer, we use `features` for `"Features"`, all_features for `List["Features"]`, `featurized_state` for `Dict[Text, List["Features"]]`, I suggest using the same names for things before `_surface_attribute`, after it is surfaced maybe we can call it smth like `raw_data`, the rest seems to be fine

rasa/core/policies/policy.py
tabergma: was it also a list of lists before? if not we might need to change the logic. just wanted to double check.

rasa/core/policies/ted_policy.py
tabergma: i prefer `LABEL_KEY` as it fits more to the data structure. Also aligns with `LABEL_SUB_KEY`. But if you have a strong argument for `LABEL` we can also change it.

rasa/core/policies/policy.py
Ghostvv: before it was numpy ndarray, but the outer dimension is always batch dimension or examples dimension

rasa/core/policies/ted_policy.py
Ghostvv: no strong argument, was just a bit confused by constant duplication

rasa/core/policies/ted_policy.py
tabergma: No, it is not zero_example as it is just a dict of, for example, `"intent" -> [<features>]`. It does not contain the sub attributes like "sequence" in a second dict. However, the features itself have the type.

rasa/utils/tensorflow/model_data.py
Ghostvv: but why do we need the first example?

rasa/core/policies/ted_policy.py
tabergma: No. It might be that the first `data_example` does not have all attributes present. That would result in issue. I think it is best to store the `zero_features` and to load them again.

rasa/utils/tensorflow/model_data.py
Ghostvv: I meant if it meant to used if `Len(List[np.ndarray]) == 1` I'd add a check here, so that it is not used to get a first example of a long list

rasa/core/policies/policy.py
tabergma: ok, then I guess we should be fine. thanks.

rasa/utils/tensorflow/model_data.py
tabergma: it seems to be not used, I guess we can remove the method

rasa/core/policies/ted_policy.py
Ghostvv: ok, but let's update name `_features` here as in previous comment

rasa/core/policies/ted_policy.py
Ghostvv: ok, but let's update the name

rasa/core/policies/ted_policy.py
tabergma: In single state featurizer we have:
```
list_of_features                -> List["Features"]
state_features                 -> Dict[Text, List["Features"]
dialogue_state_features -> List[Dict[Text, List["Features"]]]
tracker_state_features   -> List[List[Dict[Text, List["Features"]]]]
```

after _` surface_attribute´:
```
state_to_tracker_features -> Dict[Text, List[List[List["Features"]]]]
tracker_features                -> List[List[List["Features"]]]
dialogue_features              -> List[List["Features"]]
turn_features                     -> List["Features"]
```
Does that work?

Ghostvv: usually we'd use `tracker_` and `dialogue_` as nearly synonyms. Does one more list correspond to examples?

tabergma: I guess we can also change `dialogue_features` to `turn_features` then.

Ghostvv: I think it is correct now

rasa/core/policies/ted_policy.py
```diff
@@ -654,264 +720,184 @@ def load(cls, path: Text) -> "TEDPolicy":
...
+
+            self._prepare_sparse_dense_dropout_layers(
+                f"{name}_{feature_type}", self.config[DROP_RATE]
+            )
+
+            if name not in STATE_LEVEL_FEATURES:
+                self._prepare_sparse_dense_layers(
+                    self.data_signature[name][feature_type],
+                    f"{name}_{feature_type}",
+                    10,
```
Ghostvv, line 837: what is this `10`?

rasa/core/policies/ted_policy.py
```diff
@@ -654,264 +720,184 @@ def load(cls, path: Text) -> "TEDPolicy":
...
-                )
+        for name in self.data_signature.keys():
+            self._prepare_utterance_level_layers(name)
 
-        self._tf_layers[f"ffnn.{DIALOGUE}"] = layers.Ffnn(
+        for name in FEATURES_TO_ENCODE:
+            self._prepare_encoding_layers(name)
+
+        self._prepare_ffnn_layer(
+            DIALOGUE,
```
Ghostvv, line 805: what is this ffnn layer?

rasa/core/policies/ted_policy.py
```diff
@@ -654,264 +720,184 @@ def load(cls, path: Text) -> "TEDPolicy":
...
             else:
-                dense_features.append(f)
+                self._prepare_sparse_dense_layers(
+                    self.data_signature[name][feature_type],
+                    f"{name}_{feature_type}",
+                    self.data_signature[name][feature_type][0][1],
+                )
+
+    def _prepare_encoding_layers(self, name: Text) -> None:
+        feature_type = SENTENCE
```
Ghostvv, line 847: what are these encoding layers? they're only dense sentence layers for `INTENT, TEXT, ACTION_NAME, ACTION_TEXT`

rasa/core/policies/ted_policy.py
```diff
@@ -923,62 +909,101 @@ def _embed_label(self, label_in: Union[tf.Tensor, np.ndarray]) -> tf.Tensor:
...
+            batch_encoded.get(ACTION_TEXT) is not None
+            and batch_encoded.get(ACTION_NAME) is not None
+        ):
+            batch_action = batch_encoded.pop(ACTION_TEXT) + batch_encoded.pop(
+                ACTION_NAME
+            )
+        elif batch_encoded.get(ACTION_TEXT) is not None:
+            batch_action = batch_encoded.pop(ACTION_TEXT)
+        else:
+            batch_action = batch_encoded.pop(ACTION_NAME)
```
Ghostvv, line 949: let's put these `if`s in a helper method, since they repeat below

Ghostvv: actually is it simpler to use `0` instead of `None` then simple summation should work

rasa/core/policies/ted_policy.py
```diff
@@ -923,62 +909,101 @@ def _embed_label(self, label_in: Union[tf.Tensor, np.ndarray]) -> tf.Tensor:
...
-                batch_features.append(
-                    text_features if text_features is not None else name_features
-                )
+        mask = batch[attribute][MASK][0]
+        attribute_features = self._combine_sparse_dense_features(
+            batch[attribute][SENTENCE], f"{attribute}_{SENTENCE}", mask=mask
+        )
+
+        if attribute in FEATURES_TO_ENCODE:
+            attribute_features = self._tf_layers[f"ffnn.{attribute}_{SENTENCE}"](
```
Ghostvv, line 924: let's add a `TODO` that here should be a transformer for text

rasa/core/policies/ted_policy.py
```diff
@@ -923,62 +909,101 @@ def _embed_label(self, label_in: Union[tf.Tensor, np.ndarray]) -> tf.Tensor:
...
+        if not batch[attribute]:
+            return None
 
-            if text_features is not None and name_features is not None:
-                batch_features.append(text_features + name_features)
-            else:
-                batch_features.append(
-                    text_features if text_features is not None else name_features
-                )
+        mask = batch[attribute][MASK][0]
```
Ghostvv, line 918:
```suggestion
        attribute_mask = batch[attribute][MASK][0]
```

rasa/core/policies/ted_policy.py
tabergma: @evgeniiaraz

rasa/core/policies/ted_policy.py
Ghostvv: also if this ffnn contains bias, we need to multiply it by `attribute_mask` as well, otherwise it'll create nonzero values

rasa/core/policies/ted_policy.py
tabergma: it's the FFNN layer before the transformer for all the dialogue features, e.g. action_text, intent, etc.

rasa/core/policies/ted_policy.py
tabergma: `_prepare_utterance_level_layers` does not create any ffnn layers, it just creates layers for combine sparse and dense features

rasa/core/policies/ted_policy.py
Ghostvv: aaahhh, the one used on line 924? could you please add explanatory comment

rasa/core/policies/ted_policy.py
Ghostvv: ok..., but not only for user utterance also for bot

Ghostvv: we need comments 😭

rasa/core/policies/ted_policy.py
tabergma: done

rasa/core/policies/ted_policy.py
tabergma: renamed the method and added a comment

rasa/core/policies/ted_policy.py
tabergma: not 100% sure what you mean

Ghostvv: if `batch_encoded.pop(ACTION_NAME)` or `batch_encoded.pop(ACTION_TEXT)` is `0` instead of `None` then we don't actually need to have these ifs, we can simply write `batch_action = batch_encoded.pop(ACTION_TEXT) + batch_encoded.pop(ACTION_NAME)`, no?

PR #6497, opened 2020-08-27
fixing DIET tests

evgeniiaraz:
**Proposed changes**:
- NLU test fixes

### DIFF
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -535,6 +535,8 @@ class TED(TransformerRasaModel):
 
         self.all_labels_embed = None  # needed for efficient prediction
 
+        self._prepare_layers()
+
     def _check_data(self) -> None:
         if not any(key in [INTENT, TEXT] for key in self.data_signature.keys()):
             raise ValueError(
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -781,7 +781,6 @@ class DIETClassifier(IntentClassifier, EntityExtractor):
 
         # create session data from message and convert it into a batch of 1
         model_data = self._create_model_data([message], training=False)
-        model_data.data = self._prepare_example_for_prediction(model_data)
 
         return self.model.predict(model_data)
 
@@ -937,17 +936,6 @@ class DIETClassifier(IntentClassifier, EntityExtractor):
         return {"file": file_name}
 
     @classmethod
-    def _prepare_example_for_prediction(
-        self, example: RasaModelData
-    ) -> Dict[Text, Dict[Text, Union[scipy.sparse.spmatrix, np.ndarray]]]:
-        data = {}
-        for feature_name, features in example.items():
-            if TEXT in feature_name:
-                new_features = {key: value for key, value in features.items() if value}
-                data[feature_name] = new_features
-        return data
-
-    @classmethod
     def load(
         cls,
         meta: Dict[Text, Any],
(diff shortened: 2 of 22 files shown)


rasa/core/policies/ted_policy.py
```diff
@@ -535,6 +535,8 @@ def __init__(
 
         self.all_labels_embed = None  # needed for efficient prediction
 
+        self._prepare_layers()
```
Ghostvv, line 538: `prepare_layers` is called inside `TransformerRasaModel`

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -781,6 +781,7 @@ def _predict(self, message: Message) -> Optional[Dict[Text, tf.Tensor]]:
 
         # create session data from message and convert it into a batch of 1
         model_data = self._create_model_data([message], training=False)
+        model_data.data = self._prepare_example_for_prediction(model_data)
```
Ghostvv, line 784: where is this comes from?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1090,6 +1097,7 @@ def __init__(
         self._update_metrics_to_log()
 
         self.all_labels_embed = None  # needed for efficient prediction
+        self._prepare_layers()
```
Ghostvv, line 1100: this is called in super class

evgeniiaraz: I changed it because it caused the `you should always start with super().__init__()` error. And there is other no way around it.

rasa/core/policies/ted_policy.py
evgeniiaraz: same as below :)

rasa/nlu/classifiers/diet_classifier.py
evgeniiaraz: this is required to make prediction work; otherwise we add a dictionary where the subdictionary value is empty (e.g., `"text": {"sequence" : some_sparse_matrix, "sentence": None, "length": []}`). This breaks the correspondence between shapes and surface form in `generator`.

rasa/nlu/classifiers/diet_classifier.py
Ghostvv: did you remove it from `TransformerRasaModel`?

rasa/nlu/classifiers/diet_classifier.py
Ghostvv: why did it work before?

rasa/nlu/classifiers/diet_classifier.py
evgeniiaraz: Yes, and added  into `TED` and `DIET`

rasa/nlu/classifiers/diet_classifier.py
evgeniiaraz: `rasa test nlu` didn't

Ghostvv: why do we add an empty value?

evgeniiaraz: we don't explicitly add it; but I'll double check, it might be in DIET's `_create_model_data`

evgeniiaraz: fixed it; it was a bug in `model_data.add_lengths`

Ghostvv (approved):

PR #6499, opened 2020-08-27
Ted update

Ghostvv:
**Proposed changes**:
- ...

### DIFF
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -555,7 +555,10 @@ class TED(TransformerRasaModel):
 
     def _prepare_layers(self) -> None:
         for name in self.data_signature.keys():
-            self._prepare_sparse_dense_layer_for(name)
+            self._prepare_sparse_dense_layer_for(name, self.data_signature)
+
+        for name in self.label_signature.keys():
+            self._prepare_sparse_dense_layer_for(name, self.label_signature)
 
         for name in FEATURES_TO_ENCODE:
             self._prepare_encoding_layers(name)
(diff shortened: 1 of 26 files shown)


rasa/utils/tensorflow/model_data_utils.py
```diff
@@ -86,79 +92,77 @@ def create_zero_features(features: List[List[List["Features"]]],) -> List["Featu
...
-                dense_features[key] = [value[0] for value in values]
-        else:
-            for key, values in _sparse_features.items():
-                sparse_features[key] = [scipy.sparse.vstack(value) for value in values]
-            for key, values in _dense_features.items():
-                dense_features[key] = [np.vstack(value) for value in values]
-
-        # TODO not sure about expand_dims
+        # vstack serves as removing dimension
+        # TODO check vstack for sequence features
```
evgeniiaraz, line 160: for sequence features `vstack ` won't work just like that, but we'll have to change a little bit how we pad a batches in RasaModelData to manage 4d tensors.

rasa/core/policies/ted_policy.py
```diff
@@ -677,25 +698,25 @@ def _emebed_dialogue(
             dialogue_transformed = tf.expand_dims(
                 self._last_token(dialogue_transformed, sequence_lengths), 1
             )
-            mask = self._last_token(mask, sequence_lengths)
+            mask = tf.expand_dims(self._last_token(mask, sequence_lengths), 1)
```
evgeniiaraz, line 701: why do we need to `expand_dims` now?

rasa/utils/tensorflow/model_data_utils.py
Ghostvv: sure, I left a `TODO` so that we don't forget

rasa/core/policies/ted_policy.py
Ghostvv: because we do that for `dialogue_transformed` above. Then both max history and full dialogue featurizers have sequence dimension

Ghostvv: all of these millions of `expand_dims` and `squeeze` need to be cleaned though, let's add it to further TODOs list

Issue #6498, opened 2020-08-27
Add configurable prefixes to Redis Tracker and Lock stores


bjbredis:
**Description of Problem**:
Currently Rasa does not allow using the same Redis instance for the TrackerStore and the LockStore because they use the same keys which overwrite each other or Rasa requires separate logical Redis DBs which most Redis services do not support multiple logical DBs (only `select 0` supported).
Related issues: 
- [#5813 Adding support for Redis Cluster mode to store both lock-store and tracker-store](https://github.com/RasaHQ/rasa/issues/5813)
- [#6227 RedisLockStore support for cluster mode ](https://github.com/RasaHQ/rasa/issues/6227)

**Overview of the Solution**:
Allow a new parameter in `endpoints.yml` to specify an alphanumeric prefix which is prepended to all keys for the given (Lock|Tracker)Store. The prefix should be optional for compatibility therefore the configurable prefix will be prepended as either `prefix:lock:<senderid` or `prefix:tracker:<senderid>`. The lock and tracker keys will not collide _and_ a given Redis instance can be used by more than one conversation tracker. 

**Definition of Done**:
- [y] ~Tests are added~ Existing tests provide sufficient coverage with minor changes.
- [y] Feature described the docs
- [y] Feature mentioned in the changlog

sara-tagger:
Thanks for submitting this feature request 🚀 @koaning will get back to you about it soon! ✨

alwx:
Closing because the PR is merged.

PR #6500, opened 2020-08-27
Added configurable prefixes for Redis Tracker and Lock stores

bjbredis:
**Proposed changes**:
Adding configurable prefixes to Redis Tracker and Lock stores so that a single Redis instance (and logical DB) can support multiple conversation trackers and lock.
All conversation keys will be prefixed with `tracker:...` and all locks prefixed with `lock:...`. Additionally, you can specify alphanumeric-only `prefix: value` in `endpoints.yml` such that keys in redis will take the form `value:tracker:...` and `value:lock:...` respectively.
#6498 

- [y] added some tests for the functionality
- [y] updated the documentation
- [y] updated the changelog 
- [y] reformat files using `black`


sara-tagger:

data/test_endpoints/example_endpoints.yml
```diff
@@ -14,6 +14,7 @@ tracker_store:
   port: 6379
   db: 0
   password: password
+  prefix: xyz
```
ricwo, line 17: this should be a useful example, how about
```suggestion
  prefix: conversation
```

docs/docs/lock-stores.mdx
```diff
@@ -76,6 +77,9 @@ address the same node when sending messages for a given conversation ID.
 
   * `db` (default: `1`): The number of your redis database
 
+  * `prefix` (default: `None`): The prefix to prepend to lock_store keys. Must
```
ricwo, line 80:
```suggestion
  * `prefix` (default: `None`): The prefix to prepend to lock store keys. Must
```

docs/docs/tracker-stores.mdx
```diff
@@ -191,13 +191,14 @@ USER 1001`}</code></pre>
...
-         port: <port of your redis instance, usually 6379>
-         db: <number of your database within redis, e.g. 0>
-         password: <password used for authentication>
-         use_ssl: <whether or not the communication is encrypted, default `false`>
+      tracker_store:
+        type: redis
+        url: <url of the redis instance, e.g. localhost>
+        port: <port of your redis instance, usually 6379>
+        db: <number of your database within redis, e.g. 0>
+        prefix: <alphanumeric value to prepend to tracker_store keys>
```
ricwo, line 199:
```suggestion
        prefix: <alphanumeric value to prepend to tracker store keys>
```

docs/docs/tracker-stores.mdx
```diff
@@ -219,12 +220,13 @@ USER 1001`}</code></pre>
...
-          url: <url of the redis instance, e.g. localhost>
-          port: <port of your redis instance, usually 6379>
-          db: <number of your database within redis, e.g. 0>
-          password: <password used for authentication>
-          use_ssl: <whether or not the communication is encrypted, default `false`>
+        type: redis
+        url: <url of the redis instance, e.g. localhost>
+        port: <port of your redis instance, usually 6379>
+        db: <number of your database within redis, e.g. 0>
+        prefix: <alphanumeric value to prepend to tracker_store keys>
```
ricwo, line 227:
```suggestion
        prefix: <alphanumeric value to prepend to tracker store keys>
```

docs/docs/tracker-stores.mdx
```diff
@@ -237,6 +239,9 @@ USER 1001`}</code></pre>
 
   * `db` (default: `0`): The number of your redis database
 
+  * `prefix` (default: `None`): The prefix to prepend to lock_store keys. Must
```
ricwo, line 242:
```suggestion
  * `prefix` (default: `None`): The prefix to prepend to lock store keys. Must
```

rasa/core/lock_store.py
```diff
@@ -210,6 +211,8 @@ def __init__(
             password: The password which should be used for authentication with the
                 Redis database.
             use_ssl: `True` if SSL should be used for the connection to Redis.
+            prefix: prefix to prepend to all keys used by the lockstore. Must be
```
ricwo, line 214:
```suggestion
            prefix: Prefix to prepend to all keys used by the lock tore. Must be
```

rasa/core/lock_store.py
```diff
@@ -223,19 +226,32 @@ def __init__(
             ssl=use_ssl,
             socket_timeout=socket_timeout,
         )
+
+        if (prefix is None) or (prefix == ""):
+            self.prefix = "lock:"
+        elif (prefix is not None) and (
+            (isinstance(prefix, str)) and (prefix.isalnum())
+        ):
+            self.prefix = prefix + ":lock:"
```
ricwo, line 235: why have `prefix:lock:` instead of just `prefix:` if one is provided? the same goes for the tracker store. I would expect `<prefix>:` to _replace_ the default `lock:` one

rasa/core/lock_store.py
```diff
@@ -223,19 +226,32 @@ def __init__(
             ssl=use_ssl,
             socket_timeout=socket_timeout,
         )
+
+        if (prefix is None) or (prefix == ""):
```
ricwo, line 230:
```suggestion
        if not prefix:
```

rasa/core/lock_store.py
```diff
@@ -223,19 +226,32 @@ def __init__(
...
+            self.prefix = "lock:"
+        elif (prefix is not None) and (
+            (isinstance(prefix, str)) and (prefix.isalnum())
+        ):
```
ricwo, line 234:
```suggestion
        elif isinstance(prefix, str) and prefix.isalnum():
```

rasa/core/tracker_store.py
```diff
@@ -271,6 +272,19 @@ def __init__(
             host=host, port=port, db=db, password=password, ssl=use_ssl
         )
         self.record_exp = record_exp
+
+        if (prefix is None) or (prefix == ""):
```
ricwo, line 276:
```suggestion
        if not prefix:
```

rasa/core/tracker_store.py
```diff
@@ -271,6 +272,19 @@ def __init__(
...
+            self.prefix = "tracker:"
+        elif (prefix is not None) and (
+            (isinstance(prefix, str)) and (prefix.isalnum())
+        ):
```
ricwo, line 280:
```suggestion
        elif isinstance(prefix, str) and prefix.isalnum():
```

rasa/core/lock_store.py
```diff
@@ -223,19 +226,32 @@ def __init__(
             ssl=use_ssl,
             socket_timeout=socket_timeout,
         )
+
+        if (prefix is None) or (prefix == ""):
+            self.prefix = "lock:"
```
ricwo, line 231: `lock` should probably be a (module-level) constant

rasa/core/tracker_store.py
```diff
@@ -271,6 +272,19 @@ def __init__(
             host=host, port=port, db=db, password=password, ssl=use_ssl
         )
         self.record_exp = record_exp
+
+        if (prefix is None) or (prefix == ""):
+            self.prefix = "tracker:"
```
ricwo, line 277: same here, `tracker` should be a module-level constant

rasa/core/lock_store.py
ricwo: let's add a debug message

rasa/core/lock_store.py
ricwo: same here, a debug message would be helpful

rasa/core/lock_store.py
ricwo, line 230: this logic should probably be its own method (`_set_prefix(self, prefix)` or something similar) with its own unit tests

ricwo (changes requested):
Thanks a lot for creating this PR! The change looks good overall. I'm not sure adding `lock:` (`tracker:`) is necessary if a prefix is provided

ricwo:
@bjbredis please let me know once this is ready for another review

sbkv:
@bjbredis Bump

rasa/core/lock_store.py
bjbredis: If we only use `<prefix>:` and it happens to be the same prefix as used in both the lock store and tracker stores then they will collide. If the keys are`<prefix>:lock:` and `<prefix>:tracker:` then they can work in the same redis db without conflict.

rasa/core/lock_store.py
bjbredis: Hi @ricwo are you thinking something like this?: 

`
async def test_redis_lock_store_with_prefixes(monkeypatch: MonkeyPatch):
    import redis.exceptions

    lock_store = FakeRedisLockStore()
    assert lock_store._set_redis_prefix("!asdf234 34#") == False
    
    monkeypatch.setattr(
        lock_store,
        lock_store.get_or_create_lock.__name__,
        Mock(side_effect=redis.exceptions.TimeoutError),
    )

    with pytest.raises(LockError):
        async with lock_store.lock("some sender"):
            pass

    lock_store = FakeRedisLockStore()
    assert lock_store._set_redis_prefix("prefix") == True
    
    monkeypatch.setattr(
        lock_store,
        lock_store.get_or_create_lock.__name__,
        Mock(side_effect=redis.exceptions.TimeoutError),
    )  

    with pytest.raises(LockError):
        async with lock_store.lock("some new sender"):
            pass
`

rasa/core/lock_store.py
ricwo: okay, ideally we should compare the prefixes provided for the lock store and the tracker store then. but I think this solution works as well. we should definitely omit a debug message indicating what the prefix is actually set to

rasa/core/lock_store.py
ricwo: this would be clearer if a method `_get_redis_prefix()` would return the actual prefix, so that in the constructor of the redis tracker/lock stores we can have
```python
self.prefix = self._get_redix_prefix()
```

ricwo: in addition, I think the attribute should be called something else to make it clear what the prefix pertains to. something like
```python
self.key_prefix = ...
```

bjbredis:
@ricwo  I tried to up-version my master to the rasa master in order to resubmit. in the process there are conflicts in my feature branch from my master. Can you have a look and let me know if the conflicts make sense and I should attempt to resolve, or alternatively the conflicts don't make sense and something else is needed. Thanks.

ricwo:
hi @bjbredis, yes, the two conflicts in `tracker_store.py` need to be resolved. for the first one, you can keep your version and simply add the `**kwargs` parameter to the `super()` init call. for the second one, please also use the master version, but change
```python
stored = self.red.get(sender_id)
```
to
```python
stored = self.red.get(self.prefix + sender_id)
```

ricwo (approved):

sbkv:
@bjbredis @ricwo Fantastic to see the changes approved!   Can't wait for them to be merged in.

PR #6504, opened 2020-08-28
`rasa data convert` should be able to convert forms to `YAML` format

### LINKED ISSUE #6501: Convert forms with `rasa convert`
Part of #6475

**Description of Problem**:
`rasa convert` should also convert forms, but warn users that the `FormPolicy` is deprecated.

degiz:
Closes #6501 

**Proposed changes**:
- `form` is converted to `active_loop`
- users are warned that `FormPolicy` has to be activated

### DIFF
--- a/rasa/core/training/story_writer/yaml_story_writer.py
+++ b/rasa/core/training/story_writer/yaml_story_writer.py
@@ -1,18 +1,16 @@
 from collections import OrderedDict
 from pathlib import Path
-
-from ruamel import yaml
 from typing import Any, Dict, List, Text, Union, Optional
 
+from ruamel import yaml
 from ruamel.yaml.comments import CommentedMap
 from ruamel.yaml.scalarstring import (
     DoubleQuotedScalarString,
     LiteralScalarString,
 )
 
-from rasa.utils.common import raise_warning
-
-from rasa.constants import LATEST_TRAINING_DATA_FORMAT_VERSION, DOCS_URL_STORIES
+import rasa.utils.io as io_utils
+from rasa.constants import LATEST_TRAINING_DATA_FORMAT_VERSION
 from rasa.core.events import UserUttered, ActionExecuted, SlotSet, ActiveLoop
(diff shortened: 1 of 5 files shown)


rasa/utils/io.py
```diff
@@ -293,6 +293,12 @@ def write_yaml(
     # no wrap lines
     dumper.width = YAML_LINE_MAX_WIDTH
 
+    # use `null` to represent `None`
```
federicotdn, line 296: `None` isn't mapped to `null` already? :thinking:

rasa/core/training/story_writer/yaml_story_writer.py
```diff
@@ -91,17 +88,6 @@ def process_story_step(self, story_step: StoryStep) -> Optional[OrderedDict]:
...
-        if self.story_contains_forms(story_step):
-            raise_warning(
-                f'Training data file contains a story "{story_step.block_name}" '
-                f"that has form(s) in it. This story cannot be converted automatically "
-                f"because of the new Rules system in Rasa Open Source "
-                f"version {LATEST_TRAINING_DATA_FORMAT_VERSION}. "
-                f"Please convert this story manually, it will be skipped now.",
-                docs=DOCS_URL_STORIES,
-            )
-            return None
```
federicotdn, line 103: With this removed, is the return type of this function still `Optional`?

federicotdn (approved):
Tested it locally and it's working great :+1: :+1:

rasa/utils/io.py
degiz: Somehow no, it wasn't! I've checked the sources, and there's a condition indeed when `null` is replaced by just empty space, but I couldn't quickly figure out what leads to it 😞

rasa/core/training/story_writer/yaml_story_writer.py
degiz: No, it's not, good catch!

PR #6511, opened 2020-08-31
Switch to mypy part1

m-vdb:
**Proposed changes**:
Fix the following typing issues:
- `dict-item`
- `list-item`
- `call-arg`

**Leftovers to merge this PR**
- [x] [this discussion in `rasa.core.actions.action`](https://github.com/RasaHQ/rasa/pull/6511#discussion_r503945014) to resolve _(needs input from reviewer)_
- [x] [this discussion in `rasa.core.actions.two_stage_fallback`](https://github.com/RasaHQ/rasa/pull/6511#discussion_r503948950) to resolve _(needs input from reviewer)_
- [x] [this discussion in `rasa.core.test`](https://github.com/RasaHQ/rasa/pull/6511#discussion_r503955238) to resolve _(ready to implement)_
- [x] [this discussion in `rasa.shared.nlu.training_data.lookup_tables_parser`](https://github.com/RasaHQ/rasa/pull/6511#discussion_r503967424) to resolve _(ready to implement)_
- [x] [this discussion on `tests`](https://github.com/RasaHQ/rasa/pull/6511#discussion_r503969854) to resolve _(needs input from reviewer)_
- [x] API spec to update (see `rasa.server` module)
- [x] merge `main` in the branch and fix merge conflicts
- [ ] ~Update `rasa.model.Fingerprint` typing (possibly in another task)~


m-vdb:
@wochinge after merging master into this branch, I have to fix more issues. I have a question about [`ActionExecuted`](https://github.com/RasaHQ/rasa/blob/056f6bd83fba1458559e67a941aaea00189b9dc9/rasa/shared/core/events.py#L1031-L1055). Is it safe to assume that it should only be initialised with `action_name` and/or `action_text`, but if both are `None`, then we should raise an error?

rasa/nlu/model.py
```diff
@@ -77,17 +77,16 @@ def load(model_dir: Text):
...
-            return Metadata(data, model_dir)
+            return Metadata(data)
         except Exception as e:
             abspath = os.path.abspath(os.path.join(model_dir, "metadata.json"))
             raise InvalidModelError(
                 f"Failed to load model metadata from '{abspath}'. {e}"
             )
 
-    def __init__(self, metadata: Dict[Text, Any], model_dir: Optional[Text]):
+    def __init__(self, metadata: Dict[Text, Any]):
```
wochinge, line 87:
```suggestion
    def __init__(self, metadata: Dict[Text, Any]) -> None:
```

rasa/nlu/model.py
```diff
@@ -389,10 +389,11 @@ def parse(
             output["text"] = ""
             return output
 
+        timestamp = str(int(time.timestamp())) if time else None
         data = self.default_output_attributes()
         data[TEXT] = text
 
-        message = Message(data=data, time=time)
+        message = Message(data=data, time=timestamp)
```
wochinge, line 396: seems we should rather pass an `int` here instead of `Text`. The only usages of `time` I found, cast to an `int` 🙈

rasa/nlu/model.py
```diff
@@ -203,10 +202,8 @@ def train(self, data: TrainingData, **kwargs: Any) -> "Interpreter":
         for i, component in enumerate(self.pipeline):
             logger.info(f"Starting to train component {component.name}")
             component.prepare_partial_processing(self.pipeline[:i], context)
-            updates = component.train(working_data, self.config, **context)
+            component.train(working_data, self.config, **context)
```
wochinge, line 205: omg 🤦

tests/cli/test_rasa_interactive.py
```diff
@@ -61,7 +62,7 @@ def test_pass_arguments_to_rasa_train(
 
     # Mock actual training
     mock = Mock()
-    monkeypatch.setattr(rasa, "train", mock.method)
+    monkeypatch.setattr(train, "rasa_train", mock.method)
```
wochinge, line 65: This change is wrong. It mocks the actual `train` function from the `rasa.train` module. 
It was not the best idea back then to have functions with the same name as modules on the same level 😬

rasa/server.py
```diff
@@ -235,7 +245,7 @@ async def get_tracker(
     _validate_tracker(tracker, conversation_id)
 
     # `_validate_tracker` ensures we can't return `None` so `Optional` is not needed
-    return tracker  # pytype: disable=bad-return-type
+    return cast(DialogueStateTracker, tracker)
```
wochinge, line 248: Rick is currently removing this in https://github.com/RasaHQ/rasa/pull/6979

wochinge (commented):
All done besides from a small snippet in `rasa.server`.

Nice work! Cool to see that mypy also helps us detecting and removing technical debt! 🥇

rasa/server.py
```diff
@@ -131,7 +139,7 @@ def decorated(*args, **kwargs):
 def requires_auth(app: Sanic, token: Optional[Text] = None) -> Callable[[Any], Any]:
     """Wraps a request handler with token authentication."""
 
-    def decorator(f: Callable[[Any, Any], Any]) -> Callable[[Any, Any], Any]:
+    def decorator(f: "SanicView") -> "SanicView":
```
wochinge, line 142: why do we need to define a `SanicView` here? (that concludes my review)

m-vdb:
@wochinge thanks for your review! This PR was not really ready for review, but [the "part0" is](https://github.com/RasaHQ/rasa/pull/6470). Would you mind taking a look?

wochinge:
🙈  will do

rasa/cli/train.py
```diff
@@ -7,6 +7,7 @@
 import rasa.cli.arguments.train as train_arguments
 
 import rasa.cli.utils
+import rasa.train
```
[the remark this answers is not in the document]
m-vdb: dammit, it's really misleading.

rasa/core/channels/rasa_chat.py
```diff
@@ -71,6 +72,9 @@ async def _fetch_public_key(self) -> None:
                     )
 
     def _decode_jwt(self, bearer_token: Text) -> Dict:
+        if self.jwt_key is None:
+            raise TypeError("JWT public key is none.")
```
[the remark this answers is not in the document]
m-vdb: usually python standard library functions raise a TypeError for this kind of errors (`None` is of type `NoneType`)

rasa/nlu/model.py
m-vdb: It's not really needed. It's only needed if your init method doesn't have any arguments and you still want it to be type checked, as far as I recall. And also, `-> NoReturn` would be more accurate anyway

rasa/nlu/model.py
m-vdb: I don't have enough knowledge about this to make the decision myself, but don't you think it would be a breaking change that could impact users?

rasa/server.py
m-vdb: well, in that case, we'll simply handle the merge in either of the PRs, depending on which one gets merged first :)

tests/cli/test_rasa_interactive.py
m-vdb: I'm confused, what should I do here?

rasa/core/actions/two_stage_fallback.py
```diff
@@ -194,7 +194,7 @@ def _second_affirmation_failed(tracker: DialogueStateTracker) -> bool:
 
 
 def _message_clarification(tracker: DialogueStateTracker) -> List[Event]:
-    clarification = copy.deepcopy(tracker.latest_message)
+    clarification = copy.deepcopy(cast(Event, tracker.latest_message))
```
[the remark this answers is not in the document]
m-vdb: I've added the type hint on the tracker, but there still an issue to fix to remove the casting: in case `latest_message` is empty, what should be done? Raise? Return `[ActionExecuted(ACTION_LISTEN_NAME)]`?

rasa/core/channels/rasa_chat.py
```diff
@@ -85,12 +89,13 @@ async def _decode_bearer_token(self, bearer_token: Text) -> Optional[Dict]:
         # noinspection PyBroadException
         try:
             return self._decode_jwt(bearer_token)
-        except jwt.exceptions.InvalidSignatureError:
+        except jwt.InvalidSignatureError:
             logger.error("JWT public key invalid, fetching new one.")
             await self._fetch_public_key()
             return self._decode_jwt(bearer_token)
         except Exception:
             logger.exception("Failed to decode bearer token.")
```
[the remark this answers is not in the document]
m-vdb: I think that in this case we should even report it to Sentry, because it's an unexpected exception.

rasa/core/test.py
```diff
@@ -357,9 +357,10 @@ def _collect_user_uttered_predictions(
     intent_gold = event.intent.get("name")
     predicted_intent = predicted.get(INTENT, {}).get("name")
 
-    user_uttered_eval_store.add_to_store(
-        intent_predictions=[predicted_intent], intent_targets=[intent_gold]
-    )
+    if intent_gold:
```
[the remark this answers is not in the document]
m-vdb: I think it will make it fail graciously, see `has_prediction_target_mismatch()`. Could you detail a bit more?

pyproject.toml
```diff
@@ -153,6 +153,7 @@ pydoc-markdown = "^3.5.0"
 pytest-timeout = "^1.4.2"
 pytype = "^2020.6.1"
 mypy = "^0.790"
+typing-extensions = "^3.7.4"
```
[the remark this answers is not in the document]
m-vdb: Not at the moment. I wouldn't add all the type stubs unless we need them. If, when enabling more type checks, we realise that we need it, then we should add it

rasa/core/actions/action.py
```diff
@@ -716,17 +716,20 @@ def has_user_affirmed(tracker: "DialogueStateTracker") -> bool:
...
     # User affirms the rephrased intent
     rephrased_intent = tracker.last_executed_action_has(
         name=ACTION_DEFAULT_ASK_REPHRASE_NAME, skip=1
     )
     if rephrased_intent:
         revert_events += _revert_rephrasing_events()
 
+    last_user_event = tracker.get_last_event_for(UserUttered)
+    if not last_user_event:
+        return revert_events
```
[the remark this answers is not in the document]
m-vdb: I'd go with an exception as well; what type of error should it be? do you think it should land in Sentry (I think it should)

rasa/model.py
```diff
@@ -36,7 +36,7 @@
 
 
 # Type alias for the fingerprint
-Fingerprint = Dict[Text, Union[Text, List[Text], int, float]]
+Fingerprint = Dict[Text, Union[Optional[Text], List[Text], int, float]]
```
m-vdb, line 39: I think it should be a TypedDict or a dataclass, not sure how hard it would be to make the change

rasa/shared/core/training_data/structures.py
```diff
@@ -149,7 +149,9 @@ def _or_string(story_step_element: List[Event], e2e: bool) -> Text:
                 )
 
         result = " OR ".join(
-            [element.as_story_string(e2e) for element in story_step_element]
+            # FIXME: this breaks below because not
+            #        all `as_story_string()` take a `e2e` argument.
```
m-vdb, line 153: I think we should address this

rasa/shared/nlu/training_data/lookup_tables_parser.py
```diff
@@ -17,4 +19,6 @@ def add_item_to_lookup_tables(
         existing_lookup_tables.append({"name": title, "elements": [item]})
     else:
         elements = matches[0]["elements"]
+        if not isinstance(elements, list):
```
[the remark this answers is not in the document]
m-vdb: 🤔 it sounds like a wrong type annotation then. Will see...

rasa/server.py
m-vdb: since we're manipulating arguments inside the decorator and call `f` with `request`, we need the type annotation to tell mypy that `f` is a function that accepts `request` as first argument. Does that make sense?

m-vdb:
@wochinge I finally took the time to come back to this PR. I think that I answered all your comments (with sometimes more questions). When you have the time, I'd appreciate a second round of feedback (no rush at all please)

wochinge:
Will have a look early next week then 👍

rasa/core/channels/rasa_chat.py
```diff
@@ -71,6 +72,9 @@ async def _fetch_public_key(self) -> None:
                     )
 
     def _decode_jwt(self, bearer_token: Text) -> Dict:
+        if self.jwt_key is None:
+            raise TypeError("JWT public key is `None`.")
```
wochinge, line 76: ?
```suggestion
            raise TypeError("JWT public key is `None`. This is likely caused by an error when retrieving the public key from Rasa X.")
```

rasa/core/channels/rasa_chat.py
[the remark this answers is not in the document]
wochinge: Sounds good. Then let's include `raise e`?

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -59,12 +59,15 @@ def convert_to_dict(feature_states: List[Text]) -> Dict[Text, int]:
     def _state_features_for_attribute(
         self, sub_state: SubState, attribute: Text
     ) -> Dict[Text, int]:
+        # FIXME: the code below is not type-safe, but fixing it
+        #        would require more refactoring, for instance using
```
wochinge, line 63: what is the not type safe part about it?

rasa/model.py
wochinge: Should be fairly simple 👍

rasa/nlu/model.py
wochinge: https://www.python.org/dev/peps/pep-0484/ says

> Note that the return type of __init__ ought to be annotated with -> None

I think it would be more consistent to annotate it everywhere with `None`.

rasa/nlu/model.py
wochinge: I think it's safe to change. `UserUttered` uses a separate `float` timestamp anyway which means users shouldn't see this timestamp ever.

rasa/shared/core/events.py
```diff
@@ -1210,7 +1210,9 @@ def as_sub_state(self) -> Dict[Text, Text]:
         if self.action_name:
             return {ACTION_NAME: self.action_name}
         else:
-            return {ACTION_TEXT: self.action_text}
+            # FIXME: we should define the type better here, and require either
+            #        `action_name` or `action_text`
+            return {ACTION_TEXT: cast(Text, self.action_text)}
```
wochinge, line 1215: Why is this a problem? We return a `Dict[Text, Optional[Text]]` either way

rasa/shared/core/trackers.py
```diff
@@ -18,6 +18,7 @@
     Union,
     FrozenSet,
     Tuple,
+    TypedDict,
```
wochinge, line 21:
```suggestion
```

rasa/shared/core/trackers.py
```diff
@@ -184,10 +199,10 @@ def __init__(
         self.followup_action = ACTION_LISTEN_NAME
         self.latest_action = None
         # Stores the most recent message sent by the user
-        self.latest_message = None
+        self.latest_message: Optional[Event] = None
```
wochinge, line 202: even more precise
```suggestion
        self.latest_message: Optional[UserUttered] = None
```

rasa/shared/core/training_data/structures.py
```diff
@@ -149,7 +149,12 @@ def _or_string(story_step_element: List[Event], e2e: bool) -> Text:
                 )
 
         result = " OR ".join(
-            [element.as_story_string(e2e) for element in story_step_element]
+            # FIXME: this breaks below because not
+            # all `as_story_string()` take a `e2e` argument.
```
wochinge, line 153: Can we change `_or_string(story_step_element: List[Event]` to `_or_string(story_step_element: List[UserUttered]`?

rasa/shared/nlu/training_data/lookup_tables_parser.py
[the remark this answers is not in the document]
wochinge: did you already revisit this one?

rasa/server.py
```diff
@@ -1207,15 +1250,15 @@ def _training_payload_from_json(request: Request) -> Dict[Text, Union[Text, bool
 def _validate_json_training_payload(rjs: Dict):
     if "config" not in rjs:
         raise ErrorResponse(
-            400,
+            HTTPStatus.BAD_REQUEST,
```
wochinge, line 1253: 🚀

rasa/server.py
```diff
@@ -947,18 +977,25 @@ async def evaluate_intents(request: Request) -> HTTPResponse:
         if not eval_agent.model_directory or not os.path.exists(
             eval_agent.model_directory
         ):
-            raise ErrorResponse(409, "Conflict", "Loaded model file not found.")
+            raise ErrorResponse(
+                HTTPStatus.CONFLICT, "Conflict", "Loaded model file not found."
```
wochinge, line 981: can you please update the api spec acoordingly?

rasa/server.py
wochinge: Yes, it does. And something like `Callable[[Request, ??], BaseHTTPResponse]` doesn't work as we only know the first argument, but not the others, right?

rasa/core/test.py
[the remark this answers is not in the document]
wochinge: If we add only one of them (either `intent_gold` or `predicted_intent`) then the lists `intent_predictions` / `intent_targets` would end up having different lengths. `has_prediction_target_mismatch` would for sure work. However [this](https://github.com/RasaHQ/rasa/blob/1b9e44c1215769b9e7552f7382d347ed04f7f13c/rasa/core/test.py#L276) access by index and would then fail. Probably enough to make the linked line fail safe then.

wochinge (commented):
A few more questions, but looks great! I comment for now but should be a fast final review.
Thanks for working on this 🙌

rasa/nlu/model.py
m-vdb: Directly quoting the boss -> https://github.com/python/mypy/issues/3358#issuecomment-465783430 (doesn't happen every day ahah)

rasa/shared/nlu/training_data/lookup_tables_parser.py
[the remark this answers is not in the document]
m-vdb: no I haven't implemented it yet, wanted to wait for an answer on other points first 👍🏻 will do

rasa/server.py
m-vdb: 🤔 will try `Callable[[Request, ...], BaseHTTPResponse]`

m-vdb: I was able to make it work with 297c1dcca4a (it's admittedly simpler)

rasa/core/actions/action.py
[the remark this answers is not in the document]
m-vdb: ping @wochinge do you have an opinion on this one?

rasa/core/channels/rasa_chat.py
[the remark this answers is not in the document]
m-vdb: if we raise, then `logging.exception` becomes useless; will remove

rasa/shared/core/events.py
m-vdb: I think it's a problem because an `ActionExecuted` instance should either have `action_text` or `action_name` defined (not both at the same time nor none), as I understand it (let me know if this is wrong).

I added the cast because the type definition of this function and others (including `DialogueStateTracker.set_latest_action()`) expect the type to be `Dict[Text, Text]`, not `Dict[Text, Optional[Text]]`.

rasa/shared/core/training_data/structures.py
m-vdb: I guess I didn't see that we explicitly check above that every element in the list if of type `UserUttered`. It's a `mypy` issue, because it doesn't infer the type correctly after the type guard. I've adapted using ab748f6307a

rasa/core/featurizers/single_state_featurizer.py
m-vdb: [`State`](https://github.com/RasaHQ/rasa/blob/b901d55907f9584dc4f37925e4f883c0fc7da9dc/rasa/shared/core/domain.py#L72e) is defined as a dictionary of `SubState`, which is defined as `Dict[Text, Union[Text, Tuple[Union[float, Text]]]]`, which is not actually true. The way the code uses the state: for a specific key of the `State` dict, the code expects a specific `SubState` structure. For instance:
- for `INTENT`, the `SubState` is `Dict[Text, Text]`
- for `ACTION_NAME`, the `SubState` is `TypedDict(intent=Text, text=Text, entities=List[Text])`
- for `SLOT`, the `SubState` is `Dict[Text, Union[Text, Tuple[float]]]`

`_state_features_for_attribute()` is kind of uncorrelating this dependency and assumes that because the attribute is of a certain value, the sub state will have a certain structure. Which can become risky. Does that make sense?

rasa/shared/core/events.py
wochinge: Yes, it's always either.  What do you think of doing something like:
```
        if self.action_name:
            return {ACTION_NAME: self.action_name}
        elif self.action_text:
            return {ACTION_TEXT: self.action_text}
       else:
           raise ValueError("{self.class.__name} does neither have {ACTION_NAME} nor {ACTION_TEXT} specified. Please check if your training data is valid."
```

This exception should never come up. However, this makes it fail loud and clear.

rasa/core/featurizers/single_state_featurizer.py
wochinge: Thanks, makes sense! Should we also fix the type annotation for `SubState` then? At least the `ACTION_NAME` case doesn't seem to be covered, is it?

m-vdb: honestly I've tried and I couldn't make it work without refactoring the code in `rasa/core/featurizers/single_state_featurizer.py` and I was afraid of the consequences 😬 I don't know how much effort would be required to to it properly

rasa/shared/core/events.py
m-vdb: yes I wanted to do that as well, but it won't fix the typing issue I'm afraid, as it's unlikely that mypy will pick it up. I'll try though

rasa/core/featurizers/single_state_featurizer.py
wochinge: ok, let's skip it then

wochinge:
@m-vdb What's the state of this? Do you require another review?

wochinge:
@m-vdb Can we wrap this up please? Happy to help btw. As this is sort of a preventive measure for potential bugs, I'd prefer to have this in asap.

m-vdb:
will give you a status update this week, probably tomorrow 👍🏻

m-vdb:
@wochinge I've updated the status in the PR description. Let me know how you want to proceed. If we want to merge this this week or the next, I'll need some help to implement the changes.

rasa/core/actions/action.py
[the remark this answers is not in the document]
wochinge: Let's go with the exception then. How about `ActionExecutionRejection` ? The message of `ActionExecutionRejection` might need some adaption tthough.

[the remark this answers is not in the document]
wochinge: Agree it should end up in Sentry. Tempted to just to an `assert last_user_event` as it's not supposed to ever happen. Alternatively something like a `ValueError`?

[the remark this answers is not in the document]
wochinge: Or even better: `raise TypeError` (https://docs.python.org/3/library/exceptions.html#TypeError) ?

rasa/core/actions/two_stage_fallback.py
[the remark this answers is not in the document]
wochinge: How about raising a `TypeError` in case it's `null`? There should always be a `latest_message` != None

tests/cli/test_rasa_interactive.py
wochinge: Revert the change to `"train"` . Seems you already did that 💯

wochinge:
@m-vdb Came back to the comments. If you have time to wrap it up it would be great. Otherwise I suggest to add it to the squad's board or assign it to me. Besides that I'd make it a priority for the `enable-squad` to enable the remaining checks asap (probably best to create issues for each category)

rasa/core/test.py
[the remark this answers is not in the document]
m-vdb: So apparently it's failing in `sklearn` with this traceback:
```
  File "/home/runner/work/rasa/rasa/.venv/bin/rasa", line 5, in <module>
    main()
  File "/home/runner/work/rasa/rasa/rasa/__main__.py", line 116, in main
  File "/home/runner/work/rasa/rasa/rasa/cli/test.py", line 222, in test
    run_core_test(args)
  File "/home/runner/work/rasa/rasa/rasa/cli/test.py", line 108, in run_core_test
    additional_arguments=vars(args),
  File "/home/runner/work/rasa/rasa/rasa/test.py", line 158, in test_core
    core_test(stories, _agent, out_directory=output, **kwargs)
  File "/home/runner/work/rasa/rasa/rasa/utils/common.py", line 307, in run_in_loop
  File "/home/runner/work/rasa/rasa/rasa/core/test.py", line 798, in test
  File "/home/runner/work/rasa/rasa/rasa/test.py", line 331, in get_evaluation_metrics
    targets, predictions, labels=labels, output_dict=output_dict
  File "/home/runner/work/rasa/rasa/.venv/lib/python3.7/site-packages/sklearn/utils/validation.py", line 63, in inner_f
    return f(*args, **kwargs)
  File "/home/runner/work/rasa/rasa/.venv/lib/python3.7/site-packages/sklearn/metrics/_classification.py", line 1966, in classification_report
    y_type, y_true, y_pred = _check_targets(y_true, y_pred)
  File "/home/runner/work/rasa/rasa/.venv/lib/python3.7/site-packages/sklearn/metrics/_classification.py", line 83, in _check_targets
    check_consistent_length(y_true, y_pred)
  File "/home/runner/work/rasa/rasa/.venv/lib/python3.7/site-packages/sklearn/utils/validation.py", line 263, in check_consistent_length
    " samples: %r" % [int(l) for l in lengths])
ValueError: Found input variables with inconsistent numbers of samples: [51, 50]
```

Reverted my changes and updated the types of the elements in the store in this commit: 39859b4da0c

rasa/shared/nlu/training_data/lookup_tables_parser.py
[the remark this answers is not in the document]
m-vdb: ok I think I finally get it. The typing is correct because the Markdown reader adds the lookup table file [here](https://github.com/RasaHQ/rasa/blob/8692ea37d03fa4c4244919e0ab310ed1691990a6/rasa/shared/nlu/training_data/formats/markdown.py#L106) (which is maybe questionable, but out of the scope here IMO). But if during parsing we're trying to add an item to this unloaded lookup table file, then it means that the parsing is broken. Implemented this behaviour here: 3db8c33d327

rasa/cli/train.py
```diff
@@ -16,7 +18,7 @@
     DEFAULT_DOMAIN_PATH,
     DEFAULT_DATA_PATH,
 )
-import rasa.utils.common
+from rasa.train import train as train_all, train_core, train_nlu
```
wochinge, line 21: this will probably be fixed by Tom [here](https://github.com/RasaHQ/rasa/pull/8141) but can we please use local imports here to avoid slow imports?

rasa/shared/core/trackers.py
```diff
@@ -64,9 +64,23 @@
...
+    from typing_extension import TypedDict
+
     from rasa.shared.core.training_data.structures import Story
     from rasa.shared.core.training_data.story_writer.story_writer import StoryWriter
 
+    # precise type definition for `DialogueStateTracker.active_loop`
+    TrackerActiveLoop = TypedDict(
+        "TrackerActiveLoop",
+        {
+            LOOP_NAME: Text,
```
wochinge, line 76:
```suggestion
            LOOP_NAME: Optional[Text],
```

rasa/shared/core/training_data/structures.py
```diff
@@ -160,35 +175,35 @@ def as_story_string(self, flat: bool = False, e2e: bool = False) -> Text:
...
             result = f"\n## {self.block_name}\n"
-            for s in self.start_checkpoints:
-                if s.name != STORY_START:
-                    result += self._checkpoint_string(s)
+            for checkpoint in self.start_checkpoints:
+                if checkpoint.name != STORY_START:
+                    result += self._checkpoint_string(checkpoint)
 
-        for s in self.events:
+        for event in self.events:
```
wochinge, line 182: 🙌🏻

rasa/shared/nlu/training_data/lookup_tables_parser.py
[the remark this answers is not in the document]
wochinge: Why can this be a `Union[Text, List]` in the first place? What was wrong with the old type annotation?

rasa/core/channels/rasa_chat.py
```diff
@@ -71,6 +73,12 @@ async def _fetch_public_key(self) -> None:
                     )
 
     def _decode_jwt(self, bearer_token: Text) -> Dict:
+        if self.jwt_key is None:
+            raise TypeError(
```
wochinge, line 77: Is this really a `TypeError` in this case? It rather points to a wrong configuration or some problems with Rasa X to me.

rasa/shared/nlu/training_data/lookup_tables_parser.py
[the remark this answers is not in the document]
m-vdb: I would answer this question the same way as in my previous comment, that's the way the code works now, the type annotation didn't reflect how the code works now 🤷🏻 if this answer is not good enough, can you detail a bit more what you're looking for? 😬

rasa/core/channels/rasa_chat.py
m-vdb: The way I understand the code: if this attribute is `None`, it's because [these lines](https://github.com/RasaHQ/rasa/blob/f89b953866d0c85a1228fca6f7852329df5cf522/rasa/core/channels/rasa_chat.py#L90-L91) failed to do their job. And actually we're all ready logging an error [here](https://github.com/RasaHQ/rasa/blob/f89b953866d0c85a1228fca6f7852329df5cf522/rasa/core/channels/rasa_chat.py#L67-L73). So maybe you're right, and this is more a problem with how the user uses rasa/rasa X, and in that case we could raise a RasaException (which wouldn't end up in Sentry)

rasa/shared/nlu/training_data/lookup_tables_parser.py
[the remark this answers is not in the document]
wochinge: ah, I see,the markdown reader is doing that so the old type annotation was wrong 👍🏻

rasa/core/channels/rasa_chat.py
wochinge: True, but e.g. what if the `key` is a blank string? I'd use some `RasaException` type 👍🏻

m-vdb: 👍🏻

wochinge (approved):
Nice to get this in! 🚀 Great work! Do we have follow up issues for the enable squad for this?

m-vdb:
@wochinge no we don't as I don't know how the squad wants to split the work on this yet. Maybe let's have a single placeholder issue to discuss this, and create other issues when we're ready to tackle it?

wochinge:
Do you have some up do date statistics how often each error type is occuring? How about putting this in the squad channel? In my opinion we should have ready to go issue for refinement and then sprint planning.

m-vdb:
let me compile this and share it

PR #6538, opened 2020-09-01
fix(model.py): Correct docstring

PLNech:
**Proposed changes**:
- Correct `Interpreter.load()`'s docstring for `skip_validation` parameter

### Rationale

`skip_validation`, defaulted to `False`, is passed in [rasa/nlu/model.py:301](https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/model.py#L302) to `Interpreter.create()`, [which uses it as such](https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/model.py#L323):
```py
# Before instantiating the component classes,
# lets check if all required packages are available
if not skip_validation:
    components.validate_requirements(model_metadata.component_classes)
```
Thus the current documentation, `if True, does not check [...]`, is misleading: the checking does happen **unless set to `True`** , which is the exact opposite of what the docstring states.


sara-tagger:

rgstephens (approved):
Thanks for the correction

PR #6540, opened 2020-09-01
update the command to get the latest message

JustinaPetr:
**Proposed changes**:
- Quite common question I get from the community:`tracker.latest_message` returns a dictionary so when users try to get the latest message by using the command `tracker.latest_message.text` as mentioned in the docs, it throws an error. Updated an example showing how to access the latest message by returning the value based on the key.


erohmensing (approved):
Thanks for catching this!

PR #6542, opened 2020-09-01
more test fixes for core featurization

wochinge:
**Proposed changes**:
- ...


tests/core/test_processor.py
```diff
@@ -642,9 +642,16 @@ async def test_handle_message_with_session_start(
             ],
         ),
         SlotSet(entity, slot_2[entity]),
+        ActionExecuted("utter_greet"),
```
wochinge, line 645: @Ghostvv This seems weird. Why did it work before 🤯  It works now, but can you please double check?

Ghostvv: your fix seems correct to me. Could it be that session start was broken somehow?

Ghostvv: what policy `default_processor` has?

evgeniiaraz: Augmented Memo

wochinge: with max_history=3 🤔

wochinge: could it be that the `max_history` was too short before?

wochinge: Nope, a longer `max_history` doesn't break it on master

wochinge:
@Ghostvv Should I merge this including the TODO or should we investigate the fixed and previously failing test?

tests/core/test_processor.py
Ghostvv: then on master action_listen is correct because aug memo doesn't recall anything?

Ghostvv:
yes, let's merge with TODO

PR #6546, opened 2020-09-02
Introduce `shared` package

wochinge:
**Proposed changes**:
- introduce the `rasa.shared` package. This package will contain items which are shared with downstream dependencies such as Rasa X **and** are specific to the Rasa Open Source domain. This means this package is supposed to include key representations such as `Event` or `TrainingData`, but ideally no helper functions (e.g. constants for the database connection). Utilities can be included in `rasa.shared` in case the `rasa.shared` package depends on these itself. Utilities shouldn't be included just because Rasa X might need them.

- fix https://github.com/RasaHQ/rasa-x/issues/3465
- fix https://github.com/RasaHQ/rasa-x/issues/3470
- add a unit test which ensures that the `rasa.shared` package is independent. `independent` means that `rasa.shared` only imports things from external libraries or `rasa.shared`, but not from other rasa packages such as `rasa.core`. This ensures that the `shared` package can easily moved to a separate library and the code is actually de-coupled.


m-vdb:
@wochinge one thought I had and maybe you already thought about this:
- how hard would it be to have a `rasa_shared` package (next to `rasa`, same git repo) instead of `rasa.shared`? would that be supported by `poetry` / `pip`?
- would that help us more with respect to decoupling?

wochinge:
> would that be supported by poetry / pip?

Yes, [it's supported](https://python-poetry.org/docs/versions/#path-dependencies). Let me think about the implications :-)

m-vdb:
for reference, discussed in meeting: we'll keep it like that (`rasa.shared`) and setup `CODEOWNERS` so that we're a bit more defensive

wochinge:
+ we merge each iteration directly to `master`
+ the `shared` subpackage should be completely independent of the rest of Rasa Open Source (I'll set up a check for that)

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -11,6 +11,7 @@
 
 from typing import Any, Dict, List, Optional, Text, Tuple, Union, Type, NamedTuple
 
+import rasa.shared.utils.io
```
alwx, line 14: but not the same here! :)

rasa/shared/core/slots.py
```diff
@@ -1,10 +1,10 @@
 import logging
-from typing import Any, Dict, List, NoReturn, Optional, Text, Type
 
-from rasa.core import utils
 from typing import Any, Dict, List, NoReturn, Optional, Text, Type
-from rasa.core.constants import DEFAULT_CATEGORICAL_SLOT_VALUE
-from rasa.utils.common import class_from_module_path, raise_warning
+
+import rasa.shared.utils.common
+from rasa.shared.core.constants import DEFAULT_CATEGORICAL_SLOT_VALUE
```
alwx, line 6: again, I find it better to import the constants module itself, not separate constants

rasa/shared/core/slots.py
```diff
@@ -1,10 +1,10 @@
...
+from rasa.shared.utils import io as io_utils, common as common_utils
```
alwx, line 7: well, I think we should choose one name for imported stuff, and use either `import rasa.shared.utils.io as io_utils` or just `import rasa.shared.utils.io`, but not both ways mixed together. 
the reason is that different synonyms for imported modules make it more difficult to search for specific functions and their usage using utils like `grep`

alwx: in addition, you import `rasa.shared.utils.common` two times here

line 5:
`import rasa.shared.utils.common`

line 7:
`from rasa.shared.utils import io as io_utils, common as common_utils`

rasa/train.py
```diff
@@ -12,13 +12,8 @@
...
 from rasa.utils.common import TempDirectoryPath
 
-from rasa.cli.utils import (
-    print_success,
-    print_warning,
-    print_error,
-    bcolors,
-    print_color,
-)
+from rasa.cli.utils import print_success, print_warning, print_error, print_color
```
alwx, line 15: again, functions are imported by their names while the module itself should be imported

alwx (changes requested):

rasa/importers/multi_project.py
```diff
@@ -10,7 +10,8 @@
 from rasa.importers import utils
 from rasa.nlu.training_data import TrainingData
 from rasa.core.training.structures import StoryGraph
-from rasa.utils.common import raise_warning, mark_as_experimental_feature
+from rasa.utils.common import mark_as_experimental_feature
+from rasa.shared.utils.io import raise_warning
```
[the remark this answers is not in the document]
wochinge: Will do - I skipped it to keep the diff smaller, but I agree that it's inconsistent 👍

rasa/shared/utils/common.py
```diff
@@ -0,0 +1,41 @@
+from typing import Text, Optional, Any, List
+
+
+def class_from_module_path(
+    module_path: Text, lookup_path: Optional[Text] = None
+) -> Any:
+    """Given the module name and path of a class, tries to retrieve the class.
+
+    The loaded class can be used to instantiate new objects. """
+    import importlib
```
m-vdb, line 10: can we move it to the top-level?

tests/shared/test_shared.py
```diff
@@ -0,0 +1,30 @@
...
+                line
+                for line in lines
+                if line.startswith("import ") or line.startswith("from ")
+            ]
+            rasa_imports = [line for line in imports if "rasa" in line]
+
+            outside_rasa_imports = [
+                import_line
+                for import_line in rasa_imports
+                if "rasa.shared" not in import_line
```
m-vdb, line 26: It wouldn't work if you did something like `from rasa import shared`, but that's admittedly a detail/edge case that we can accept.

tests/shared/utils/test_common.py
```diff
@@ -0,0 +1,14 @@
+import random
+
+import rasa.shared.utils.common
+
+
+def test_all_subclasses():
+    num = random.randint(1, 10)
```
m-vdb, line 7: I'm not sure if the use of `random` here is useful, at all

rasa/train.py
wochinge: I'll leave these as they are as this is going beyond the scope of this PR. Will change all imports from `rasa.shared` though which aren't imported by module.

tests/shared/utils/test_common.py
```diff
@@ -0,0 +1,14 @@
...
+
+    class TestClass:
+        pass
+
+    classes = [type(f"TestClass{i}", (TestClass,), {}) for i in range(num)]
```
m-vdb, line 12: it doesn't really test for sub-subclasses and doesn't check that if a class is not a subclass, then it's not included

(i know you haven't introduced the code, feel free to add a FIXME if lack of time)

wochinge:
@alwx Thanks for the review!  Implemented the changes

tests/shared/test_shared.py
```diff
@@ -0,0 +1,30 @@
...
+            ]
+
+            # The shared package is required to be independent of the rest of Rasa
+            assert not outside_rasa_imports
```
m-vdb, line 30: a more meaningful error message could be:
```suggestion
            assert not outside_rasa_imports, (
                f"File {file} imports code from outside "
                f"of `rasa.shared`: {','.join(outside_rasa_imports)}"
            )
```

And if there are several files that could raise errors, it might be nice to aggregate all the errors first and raise at the end with a summary of all files.

m-vdb (commented):
Looks good! I added a few comments.

One other thing that comes to mind and challenges a bit the use of `rasa.shared.utils` from inside Rasa: if we move things to `rasa.shared` because they constitute an interface for `rasa`, it might not make sense for `rasa` to import utils from there (mirroring what we decided for rasa x). I know it may sound like a lot of duplication (for instance 3 `raise_warning` functions).

So I would suggest: either decide that `raise_warning` is part of this interface, and do not do duplication, or decide that it's not part of the interface, and never import it from `rasa.shared`. But not do it half-half. This decision doesn't need to be done here, and can be done in the end when we have a clearer picture of what's in utils. WDYT?

wochinge:
> But not do it half-half.

By "half-half" you mean that Rasa X currently duplicates it and Rasa Open Source doesn't? I think we really have to see how this whole `utils` package develops when we are moving more things to `rasa.shared`. Currently I lean towards not having it duplicated (this would mean we can undo some of the duplication within Rasa X). So I agree to `this decision doesn't need to be done here, and can be done in the end when we have a clearer picture of what's in utils`  👍

tests/shared/test_shared.py
wochinge: it unlikely that somebody does `from rasa import shared` but added it 👍

tests/shared/utils/test_common.py
wochinge: > it doesn't really test for sub-subclasses 

It does actually create subclasses (https://docs.python.org/3/library/functions.html#type)

>  doesn't check that if a class is not a subclass, then it's not included

What would you like to test here? The fact that other, random classes, aren't found by this function?

alwx (approved):
🚀🚀🚀

tests/shared/utils/test_common.py
m-vdb: I meant sub-sub classes so that uses the recursion

wochinge: will address that in a separate PR 👍 (wanted to get this merged)

m-vdb:
@wochinge @alwx I think we forgot about the [`CODEOWNERS` file](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) :octocat:

wochinge:
will add it here https://github.com/RasaHQ/rasa/pull/6576

PR #6547, opened 2020-09-02
New tests

evgeniiaraz:
**Proposed changes**:
- Added more tests for `SingleStateFeaturizer `
- Created tests for `model_data_utils`

### DIFF
--- a/tests/core/test_featurizer.py
+++ b/tests/core/test_featurizer.py
@@ -39,7 +39,7 @@ def test_fail_to_load_non_existent_featurizer():
 
 def test_binary_featurizer_correctly_encodes_state():
     """
-    Check that all the attributes are correctly featurized when they should and not featurized when shouldn't;  
+    Check that all the attributes are correctly featurized when they should and not featurized when shouldn't;
     """
     f = BinarySingleStateFeaturizer()
     f._default_feature_states[INTENT] = {"a": 0, "b": 1}
--- a/tests/core/test_featurizer.py
+++ b/tests/core/test_featurizer.py
@@ -37,7 +37,7 @@ def test_fail_to_load_non_existent_featurizer():
     assert TrackerFeaturizer.load("non_existent_class") is None
 
 
-def test_single_state_featurizer_correctly_encodes_state():
+def test_single_state_featurizer_correctly_encodes_state_without_interpreter():
     """
     Check that all the attributes are correctly featurized when they should and not featurized when shouldn't;
     These tests are for encoding state without a trained interpreter.
--- a/tests/core/test_featurizer.py
+++ b/tests/core/test_featurizer.py
@@ -40,6 +40,7 @@ def test_fail_to_load_non_existent_featurizer():
 def test_single_state_featurizer_correctly_encodes_state():
     """
     Check that all the attributes are correctly featurized when they should and not featurized when shouldn't;
+    These tests are for encoding state without a trained interpreter.
     """
     f = SingleStateFeaturizer()
     f._default_feature_states[INTENT] = {"a": 0, "b": 1}
@@ -49,14 +50,14 @@ def test_single_state_featurizer_correctly_encodes_state():
 
     encoded = f.encode_state(
         {
-            "user": {"intent": "a"},
-            "prev_action": {"action_name": "d"},
+            "user": {"intent": "a", "text": "blah blah blah"},
+            "prev_action": {"action_name": "d", "action_text": "boom"},
             "active_loop": {"name": "i"},
             "slots": {"g": (1.0,)},
         },
         interpreter=None,
     )
-    # user input is ignored as prev action is not action_listen;
+    # user input is ignored as prev action is not action_listen
     assert list(encoded.keys()) == [ACTION_NAME, ACTIVE_LOOP, SLOTS]
     assert (
         encoded[ACTION_NAME][0].features != scipy.sparse.coo_matrix([[0, 1, 0]])
@@ -68,19 +69,36 @@ def test_single_state_featurizer_correctly_encodes_state():
 
     encoded = f.encode_state(
         {
-            "user": {"intent": "a"},
-            "prev_action": {"action_name": "action_listen"},
+            "user": {"intent": "a", "text": "blah blah blah"},
+            "prev_action": {"action_name": "action_listen", "action_text": "boom"},
             "active_loop": {"name": "k"},
             "slots": {"e": (1.0,)},
         },
         interpreter=None,
     )
-
+    # we featurize all the features except for *_text ones because NLU wasn't trained
     assert list(encoded.keys()) == [INTENT, ACTION_NAME, ACTIVE_LOOP, SLOTS]
     assert (encoded[INTENT][0].features != scipy.sparse.coo_matrix([[1, 0]])).nnz == 0
     assert (
         encoded[ACTION_NAME][0].features != scipy.sparse.coo_matrix([[0, 0, 1]])
     ).nnz == 0
+    assert (
+        encoded[ACTIVE_LOOP][0].features != scipy.sparse.coo_matrix([[0, 0, 0, 1]])
+    ).nnz == 0
+    assert (encoded[SLOTS][0].features != scipy.sparse.coo_matrix([[1, 0, 0]])).nnz == 0
+
+    # check that no intent / action_name features are added when the interpreter isn't there and
+    # intent / action_name not in input
+    encoded = f.encode_state(
+        {
+            "user": {"text": "blah blah blah"},
+            "prev_action": {"action_text": "boom"},
+            "active_loop": {"name": "k"},
+            "slots": {"e": (1.0,)},
+        },
+        interpreter=None,
+    )
+    assert list(encoded.keys()) == [ACTIVE_LOOP, SLOTS]
     assert (
         encoded[ACTIVE_LOOP][0].features != scipy.sparse.coo_matrix([[0, 0, 0, 1]])
     ).nnz == 0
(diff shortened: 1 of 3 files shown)


tests/core/test_featurizer.py
```diff
@@ -137,16 +167,36 @@ def test_single_state_featurizer_correctly_encodes_state(
...
         },
         interpreter=kwargs["interpreter"],
     )
-    assert all([attribute in encoded for attribute in [TEXT, ENTITIES, ACTION_NAME]])
+    assert all(
+        [
+            attribute in encoded
+            for attribute in [TEXT, ENTITIES, ACTION_NAME, SLOTS, ACTIVE_LOOP]
+        ]
+    )
```
Ghostvv, line 188: what does it check?

tests/core/test_featurizer.py
```diff
@@ -137,16 +167,36 @@ def test_single_state_featurizer_correctly_encodes_state(
...
+
+    encoded = f.encode_state(
+        {
+            "user": {"text": "a ball", "entities": ["c"]},
+            "prev_action": {"action_text": "throw a ball"},
+        },
+        interpreter=kwargs["interpreter"],
+    )
+
+    assert all([attribute in encoded for attribute in [ACTION_TEXT]])
```
Ghostvv, line 201: why can't we check `ACTION_TEXT in encoded`?

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+import numpy as np
+import copy
+
+from rasa.utils.tensorflow import model_data_utils
+from rasa.utils.features import Features
+from rasa.nlu.constants import INTENT, ACTION_NAME
+from rasa.utils.tensorflow.constants import SENTENCE
+
+
+def test_create_zero_features():
```
Ghostvv, line 12: @tabergma could you please check tests for model_data_utils?

Ghostvv (commented):
I think test can be simplified by providing state with all inputs (text and intent, action_text and action_name) for 2 cases for previous action some action and action listen.

tests/core/test_featurizer.py
evgeniiaraz: that all the features we expect to be encoded are encoded (to avoid the problem we had before with Forms and Slots)

Ghostvv: isn't the same as `assert list(encoded.keys()) == [TEXT, ENTITIES, ACTION_NAME, SLOTS, ACTIVE_LOOP]`

evgeniiaraz:
> I think test can be simplified by providing state with all inputs (text and intent, action_text and action_name) for 2 cases for previous action some action and action listen.

@Ghostvv 
which test did you mean?

Ghostvv:
SingleStateFeaturizer - there is no need to call encode multiple times, where there's only text, only action_text, etc

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    features = [[None, None, [sparse_feature_sentence_features]]]
+    zero_features = model_data_utils.create_zero_features(features)
+    assert len(zero_features) == 1
+    assert zero_features[0].is_sparse()
+    assert (zero_features[0].features != scipy.sparse.coo_matrix((1, 100))).nnz == 0
+
+
+def test_surface_attributes():
+    intent_features = np.zeros((1, 100))
+    intent_features[0, 50] = 1
```
tabergma, line 46: You are not using the intent features later on in the test. Ideally you check if the features are still the same.

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+                attribute=INTENT,
+                feature_type=SENTENCE,
+                origin=[],
+            )
+        ]
+    }
+
+    action_name_features = np.zeros((1, 100))
+    action_name_features[0, 57] = 1
+    action_name_features = scipy.sparse.coo_matrix(action_name_features)
```
tabergma, line 60: same here

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        ACTION_NAME: [
+            Features(
+                features=action_name_features,
+                attribute=ACTION_NAME,
+                feature_type=SENTENCE,
+                origin=[],
+            )
+        ]
+    }
+    state_features = copy.deepcopy(intent_features)
```
tabergma, line 71: why do we need to use `deepcopy´, does it not work without it?

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+from rasa.utils.tensorflow import model_data_utils
+from rasa.utils.features import Features
+from rasa.nlu.constants import INTENT, ACTION_NAME
+from rasa.utils.tensorflow.constants import SENTENCE
+
+
+def test_create_zero_features():
+    # DENSE FEATURES
+    dense_feature_sentence = np.zeros((1, 100))
+    dense_feature_sentence[0, 50] = 1
```
tabergma, line 15:
```suggestion
shape = (1, 100)
```

tests/utils/tensorflow/test_model_data_utils.py
evgeniiaraz: because if we update `intent_features` in place, then we won't be able to use them without `action_name` features

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    dense_feature_sentence_features = Features(
+        features=dense_feature_sentence,
```
tabergma, line 17:
```suggestion
        features=np.random.rand(shape)
```

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        attribute=INTENT,
+        feature_type=SENTENCE,
+        origin=[],
+    )
+    features = [[None, None, [dense_feature_sentence_features]]]
+
+    zero_features = model_data_utils.create_zero_features(features)
+    assert len(zero_features) == 1
+    assert zero_features[0].is_dense()
+    assert (zero_features[0].features == np.zeros((1, 100))).all()
```
tabergma, line 27:
```suggestion
    assert (zero_features[0].features == np.zeros((shape))).all()
```

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        features=sparse_feature_sentence,
+        attribute=INTENT,
+        feature_type=SENTENCE,
+        origin=[],
+    )
+    features = [[None, None, [sparse_feature_sentence_features]]]
+    zero_features = model_data_utils.create_zero_features(features)
+    assert len(zero_features) == 1
+    assert zero_features[0].is_sparse()
+    assert (zero_features[0].features != scipy.sparse.coo_matrix((1, 100))).nnz == 0
```
tabergma, line 41:
```suggestion
    assert (zero_features[0].features != scipy.sparse.coo_matrix(shape)).nnz == 0
```

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+    zero_features = model_data_utils.create_zero_features(features)
+    assert len(zero_features) == 1
+    assert zero_features[0].is_dense()
+    assert (zero_features[0].features == np.zeros((1, 100))).all()
+
+    # SPARSE FEATURES
+    sparse_feature_sentence = scipy.sparse.coo_matrix(dense_feature_sentence)
+    sparse_feature_sentence_features = Features(
+        features=sparse_feature_sentence,
```
tabergma, line 32:
```suggestion
        features=scipy.sparse.coo_matrix(np.random.rand(shape))
```

tests/utils/tensorflow/test_model_data_utils.py
```diff
@@ -0,0 +1,138 @@
...
+
+    (
+        attribute_masks,
+        dense_features,
+        sparse_features,
+    ) = model_data_utils.map_tracker_features(tracker_features, zero_features_list)
+    expected_mask = np.array([[1, 0, 1], [0, 0, 1], [1, 1, 1]])
+
+    assert np.all(np.squeeze(np.array(attribute_masks), 2) == expected_mask)
+    assert np.array(dense_features["sentence"]).shape[-1] == zero_features.shape[-1]
```
tabergma, line 138: what about sparse_features?

evgeniiaraz:
@Ghostvv isn't it the same as what we are doing now?
I'm doing 1 encode on state with `action_listen` and the other one on one with a different action
(both in `test_single_state_featurizer_correctly_encodes_state` and `test_single_state_featurizer_correctly_encodes_state`)
I'm not passing all the features because I also want to check that TEXT and INTENT will be featurized as they should; if intent was there, text wouldn't be featurized (and same for `action_text `and `action_name`)

tests/utils/tensorflow/test_model_data_utils.py
tabergma: ok, but then we don't need to deepcopy in the next lines

evgeniiaraz: yes, removed it! It was just me being overly cautious

tests/utils/tensorflow/test_model_data_utils.py
evgeniiaraz: added a check for them below.

Ghostvv:
> if intent was there, text wouldn't be featurized (and same for action_text and action_name)

that's not the case anymore. They should be both featurized. That's why we don't need to create different states for them

evgeniiaraz:
> that's not the case anymore. They should be both featurized. That's why we don't need to create different states for them
but won't we need to create 2 anyway -- one with action_listen and one without?

tests/core/test_featurizer.py
```diff
@@ -36,29 +37,76 @@ def test_fail_to_load_non_existent_featurizer():
...
+        {
+            "user": {"intent": "a", "text": "blah blah blah"},
+            "prev_action": {"action_name": "action_listen", "action_text": "boom"},
+            "active_loop": {"name": "k"},
+            "slots": {"e": (1.0,)},
+        },
         interpreter=None,
     )
-    assert list(encoded.keys()) == [INTENT, ACTION_NAME]
+    # we featurize all the features except for *_text ones because NLU wasn't trained
```
Ghostvv, line 79: let's put that also in the name of the test `..._without_interpreter`

Ghostvv (approved):

tests/core/test_featurizer.py
evgeniiaraz: done!

PR #6551, opened 2020-09-02
fix fingerprinting of domains with `triggers: `

wochinge:
**Proposed changes**:
- fix a bug when fingerprinting models which have a domain with `trigger: ` intents. The problem was that the fingerprinting removed the `responses` from the domain and then created domain which was checked for its validity. As the responses were missing, the related actions were also missing. This caused an error in the domain sanity check as the triggered actions aren't existing
- remove `test_create_fingerprint_from_invalid_paths`: This test just tested the behavior of `RasaFileImporter` but not the actual fingerprinting functionality
- added more tests, mainly to make sure that changed response texts only affect the NLG fingerprint
- added more typing

### DIFF
--- a/rasa/core/domain.py
+++ b/rasa/core/domain.py
@@ -450,11 +450,11 @@ class Domain:
         self._check_domain_sanity()
 
     def __hash__(self) -> int:
-
         self_as_dict = self.as_dict()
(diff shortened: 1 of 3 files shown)


federicotdn (approved):
Looks good :+1: :+1:

PR #6564, opened 2020-09-03
Revert "remove end-to-end reading yaml"

wochinge:
**Proposed changes**:
- re-add end-to-end importer so that NLU training receives the necessary action names
- don't re-add the actual reading of end-to-end stories as this is not supposed to be supported yet


Ghostvv (approved):
I guess this is the same code?

wochinge:
@evgeniiaraz Can you let me know when `new_core_featurization` passes so that I can then run the tests here, and merge? At the moment everything is breaking.

wochinge:
Or should I just merge? However, then it's on you to ping me if things break 🙈

evgeniiaraz:
@wochinge let's give it 5 mins, one of the commits (which I expect to fix things) shouls be done in 5. if it doesn't pass then, I'd just merge it. 
I'll give you a heads up once it's done

evgeniiaraz:
@wochinge So, I think now all the test code tests should work, while the type checks fail (I have an idea of where it's coming from). Though -- I don't know if we should wait for all the tests to finish because github seems to have a bit of a hard time with tests.

Issue #6222, opened 2020-07-16
Problem with integrating BERT (tensorflow.python.framework.errors_impl.InvalidArgumentError: indices[0,512] = 512 is not in [0, 512))


MajaRolevski:
**Rasa version**: 1.10.7 (also tried with 1.8.3)

**Python version**: 3.7

**Operating system** (windows, osx, ...): Linux and Windows

**Issue**:
I have a problem with integrating BERT into the configuration of my model. Namely, at the very beginning of training it gives the error message shown below. There is also a config.yml file I used. 

It is interesting that I am not getting this error when using a very small train dataset, so the model is able to finish with training successfully.

**Error (including full traceback)**:
```
2020-07-16 13:05:16 INFO     rasa.nlu.model  - Starting to train component HFTransformersNLP
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/bin/rasa", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/rasa/__main__.py", line 92, in main
    cmdline_arguments.func(cmdline_arguments)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/rasa/cli/train.py", line 140, in train_nlu
    persist_nlu_training_data=args.persist_nlu_data,
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/rasa/train.py", line 414, in train_nlu
    persist_nlu_training_data,
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/rasa/train.py", line 453, in _train_nlu_async
    persist_nlu_training_data=persist_nlu_training_data,
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/rasa/train.py", line 482, in _train_nlu_with_validated_data
    persist_nlu_training_data=persist_nlu_training_data,
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/rasa/nlu/train.py", line 90, in train
    interpreter = trainer.train(training_data, **kwargs)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/rasa/nlu/model.py", line 191, in train
    updates = component.train(working_data, self.config, **context)
[...]

PR #6567, opened 2020-09-03
Better handling of sequences longer than maximum sequence length for BERT style models

dakshvar22:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/6222


dakshvar22:
/modeltest

```yml

include:
 - dataset: ["all"]
   config: ["BERT + DIET(bow) + ResponseSelector(bow)", "BERT + DIET(seq) + ResponseSelector(t2t)"]

```

rasa/nlu/utils/hugging_face/hf_transformers.py
```diff
@@ -449,7 +602,10 @@ def _get_model_features_for_batch(
         return sentence_embeddings, sequence_final_embeddings
 
     def _get_docs_for_batch(
```
tabergma, line 604: Just for my understanding: Did we also update the call of `_get_docs_for_batch` in `train` and `predict`?

tests/nlu/utils/test_hf_transformers.py
```diff
@@ -0,0 +1,76 @@
...
+
+from rasa.nlu.utils.hugging_face.hf_transformers import HFTransformersNLP
+from rasa.nlu.training_data import Message
+
+
+@pytest.mark.parametrize(
+    "input_sequence_length, model_name, should_overflow",
+    [(20, "bert", False), (1000, "bert", True), (1000, "xlnet", False)],
+)
+def test_sequence_length_overflow_train(
```
tabergma, line 12: Can you please add types.

tabergma: also for the other tests

tabergma (commented):
Looks great 🚀  Added some comments.

tests/nlu/utils/test_hf_transformers.py
dakshvar22: I never saw types added to the tests and I think that's because the API should not be tested against specific types. The test should break if the input the API is wrong. Happy to add the types if that's not the case.

tabergma: part of our coding guidelines: https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e

dakshvar22: Thanks 👍  i'll add them

rasa/nlu/utils/hugging_face/hf_transformers.py
tabergma: But then how does the method know it is in inference mode or not? Sorry I am a bit confused.

dakshvar22: You are absolutely right! I had missed passing that argument in the call to `_get_docs_for_batch` in `process` method. The call in `train` stays stays the same since its False by default. I passed them in the tests correctly so, didn't catch the problem.
Thanks for catching this.

tabergma: Can/Should we add a test for this?

tabergma (approved):
Thanks for addressing my comments 💯 

If possible, it would be great to add a test that the inference mode is set to False during training and to True during prediction. But other than that, good to go 👍

dakshvar22:
@tabergma Thanks for the review, I was adding a test for it. Learnt how to test logging statements 😄

tests/nlu/utils/test_hf_transformers.py
```diff
@@ -81,3 +82,45 @@ def test_input_padding(
...
+
+    transformers_nlp = HFTransformersNLP(transformers_config)
+
+    text = " ".join(["hi"] * sequence_length)
+    message = Message(text)
+
+    caplog.set_level(logging.DEBUG)
+    transformers_nlp.process(message)
+    if should_overflow:
+        print(caplog.text)
```
tabergma, line 105: I guess this can be removed.

dakshvar22: Why would i miss that 🤯 
Thanks!

PR #6575, opened 2020-09-04
No warning during prediction if Response Selector was not trained

dakshvar22:
**Proposed changes**:
- `label_intent_response_key` is `None` when response selector wasn't trained. The check for response templates should be skipped then.


wochinge (approved):
Thanks for the quick fix!

PR #6580, opened 2020-09-04
Change default augmentation parameter

### LINKED ISSUE #6550: set `--augmentation 0` by default in `rasa train` and `rasa train core`

tabergma:
**Proposed changes**:
Change the default parameter for `augmentation` for `rasa train` commands from `50` to `0`.

closes #6550 

### DIFF
--- a/changelog/6550.improvement.md
+++ b/changelog/6550.improvement.md
@@ -1,0 +1,1 @@
+Change the default parameter for `augmentation` for `rasa train` commands from `50` to `0`.
--- a/rasa/cli/arguments/train.py
+++ b/rasa/cli/arguments/train.py
@@ -106,7 +106,7 @@ def add_augmentation_param(
     parser.add_argument(
         "--augmentation",
         type=int,
-        default=50,
+        default=0,
         help="How much data augmentation to use during training.",
     )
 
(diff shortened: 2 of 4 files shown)


Ghostvv:
before we do that, we need to run experiments on all available core datasets, to make sure it doesn't reduce performance. cc @akelad

tabergma:
Ok, will use the model regression tests for that.

Ghostvv:
but our model regression tests are nlu only, no?

tabergma:
Yeah, makes sense. Too bad we can not use the model regression tests for this. Will run some experiments.

Why do we actually want to have this change?

Ghostvv:
Because, it increases training time drastically, but we're not sure that with the new system (Rules+TED) it'd add to performance

tabergma:
Do we already have Sara data in the new training data format using rules? And what about carbon bot?

Ghostvv:
there are `rule` or `rules` branches, for them, but I don't know how fully updated they are

tabergma:
Ok, does it make sense then to wait until those datasets are fully available so that we can test properly? We could close the PR until then.

Ghostvv:
cc @evgeniiaraz

tabergma:
Ok, will close this PR for now. Let's reopen once the datasets to test this are ready.

PR #6584, opened 2020-09-07
nlu training to shared

wochinge:
**Proposed changes**:
- move `rasa.nlu.training_data` to `rasa.shared`
- move required utils / constants to `rasa.shared`


wochinge:
cc @alwx

rasa/shared/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class RasaException(Exception):
+    """Base exception class for all errors raised by Rasa."""
```
alwx, line 2: (definitely optional): I wonder if we really need a dedicated file for exceptions since there is only one very very generic one.

alwx:
We will definitely need to continue working on the shared package after the RC release :) I have a bunch of comments for your PR (like the usage of constants, imports, stuff like that) but I also feel that it all doesn’t really make sense — the job was to copy stuff and change references, not change the name of constants and refactor `rasa` :)

However, this package is going to be publicly available and theoretically used by third parties so I think we need to discuss how it all looks like after we finish working on decoupling and create a couple of follow-up stories after that.

rasa/shared/exceptions.py
wochinge: I thought more would come soon 🤔  I agree with you that we should have a look on the final package and then revisit what we actually still need or don't.

wochinge:
I completely agree. It's already enough changed lines. Refactoring of current behavior / constants should be done in separate PRs. 

Same for revisiting everything once we're done 👍  Let's see how the picture looks like when are done and what actually makes sense and what was too much brute force 👍

wochinge:
Seems the merge with `e2e` is gonna be fun - though it should be mostly small things. I hope the `e2e` branch goes in this night so I can start with this first thing in the morning.

wochinge:
@alwx Did you review this? So would it be ready to go in? In case we'd be ready before `e2e` is? 👀

alwx (approved):
Again, we should think more carefully about how we import stuff but it's a good first step anyway, and I think we should merge this one.

wochinge:
yeeees!

PR #6585, opened 2020-09-07
Update __init__.py

pranavdurai10:
**Proposed changes**:
-Un-used variable v in loop


sara-tagger:

pranavdurai10:
@sara-tagger @erohmensing Hey, any heads-up on my pull request?

rasa/core/events/__init__.py
```diff
@@ -83,7 +83,7 @@ def md_format_message(
 
 def first_key(d: Dict[Text, Any], default_key: Any) -> Any:
     if len(d) > 1:
-        for k, v in d.items():
+        for k in d.items():
```
erohmensing, line 86: In my opinion it's more confusing to only use the first value from a method that returns two values than to have an unused variable. what about this instead? 
```suggestion
        for k in d.keys():
```

erohmensing (changes requested):
Please be patient for a review when submitting a PR! 

See my comment for a suggestion :)

erohmensing (approved):
thanks for your first contribution!

PR #6589, opened 2020-09-07
Add option 'use_lemma' to CountVectorsFeaturizer

### LINKED ISSUE #6536: Lemmatization and CountVectorFeaturizers
**Rasa version**:

1.8 onwards 

**Python version**:

3.6 or 3.7

All 

**Issue**:

Currently if you use spaCy components in your pipeline then the behaviour of the CountVectorFeaturizer changes.  Currently we use the `lemma` that spaCy provides as an alternative for the text when we use the "word" analyser. There's potentially merit to this idea but currently:

1. The user does not know this is happening because this phenomenon is undocumented. At the very least we need to update the docs to reflect this.
2. The user cannot configure this, the user must use the `lemma` tokens at all times the user wants to use spaCy. 

I think the best way forward is to implement a configuration in the `CountVectorFeaturizer` that makes it possible to not use the `lemma`. Once this is implemented we can update the documentation.

tabergma:
**Proposed changes**:

closes #6536 


rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -77,6 +77,9 @@ def required_components(cls) -> List[Type[Component]]:
         # will be converted to lowercase if lowercase is True
         "OOV_token": None,  # string or None
         "OOV_words": [],  # string or list of strings
+        # indicates whether the featurizer should use the lemma of a word for
+        # counting (if available) or not
+        "use_lemma": True,
```
degiz, line 82: Optional: I see that it's not really common in this class, but since `use_lemma` is used more than once, I'd extract it (and other config options) to a constant.

tests/nlu/featurizers/test_count_vectors_featurizer.py
```diff
@@ -397,3 +400,46 @@ def test_count_vectors_featurizer_train():
...
+
+    seq_vecs, sen_vecs = test_message.get_sparse_features(TEXT, [])
+
+    assert isinstance(seq_vecs, scipy.sparse.coo_matrix)
+    assert isinstance(sen_vecs, scipy.sparse.coo_matrix)
+
+    actual_seq_vecs = seq_vecs.toarray()
+    actual_sen_vecs = sen_vecs.toarray()
+
+    print(actual_sen_vecs, sentence_features)
```
degiz, line 441: what's the goal of printing that in a test? To have the logs in case something fails?

docs/docs/components/featurizers.mdx
```diff
@@ -330,6 +330,10 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
   If you want to use character n-grams, set `analyzer` to `char` or `char_wb`.
   The lower and upper boundaries of the n-grams can be configured via the parameters `min_ngram` and `max_ngram`.
   By default both of them are set to `1`.
+  By default the featurizer takes the lemma of a word instead of the word directly if it is available. The lemma
+  of a word is for example set by the `SpacyTokenizer`. You can disable this behavior
```
degiz, line 334: Sorry if I missed something, but should we mention that we provide implementation only for `SpacyTokenizer`?

degiz (approved):

tests/nlu/featurizers/test_count_vectors_featurizer.py
tabergma: oh, was just for me when debugging, will remove this

PR #6493, opened 2020-08-26
Added version numbers to Rasa examples' training data

b-quachtran:
**Proposed changes**:
- Added version numbers to training data in Rasa examples

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6593, opened 2020-09-08
fix breaking master tests + endlessly running tests

wochinge:
**Proposed changes**:
- fix failing test `test_train_status_is_not_blocked_by_training` caused by [this PR](https://github.com/RasaHQ/rasa/pull/6493)
- add missing timeout: This avoids endless loops in case the test breaks
- remove duplicate `version: 2.0` entry when concatenating the files for the HTTP request. This led to a YAML error as the same key was specified multiple times. 

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -5,7 +5,6 @@ from unittest.mock import Mock, ANY
 
 import requests
 import time
-import tempfile
 import uuid
 
 from typing import List, Text, Type, Generator, NoReturn, Dict
@@ -18,6 +17,8 @@ import pytest
 from freezegun import freeze_time
 from mock import MagicMock
 from multiprocessing import Process, Manager
+
+from rasa.validator import KEY_TRAINING_DATA_FORMAT_VERSION
 
 import rasa
 import rasa.constants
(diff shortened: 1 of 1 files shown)


tests/test_server.py
```diff
@@ -200,8 +202,21 @@ def send_request() -> None:
             Path("data") / "nlu.yml",
         ]:
             full_path = project_path / file
-            payload += full_path.read_text()
+            content = full_path.read_text()
```
tmbo, line 205: should this rather read the content as a dict, combine the dicts, and then dump that dict to yaml as a payload?

wochinge: I think both is fine. Reading and concatenating as dict is cleaner, but then we'd have to dump to file and read it again from file. Happy to adapt if you think it's crucial.

wochinge: I'm also happy to do this in a separate PR so we can unblock `master` asap.

m-vdb: it seems a bit more future-proof to do it using Tom's suggestion. Let's merge this fix and do a follow-up PR?

m-vdb (approved):

tests/test_server.py
wochinge: will do

wochinge: https://github.com/RasaHQ/rasa/pull/6595

PR #6601, opened 2020-09-08
Fix boolean slot featurization

JEM-Mosig:
**Proposed changes**:
- Fixes boolean featurization bug, as reported [here](https://forum.rasa.com/t/questions-regarding-data-validation-script-and-with-slots-entites-and-followups-from-custom-actions/33405)


rasa/shared/core/slots.py
```diff
@@ -160,6 +160,21 @@ def feature_dimensionality(self) -> int:
         return len(self.as_feature())
 
 
+def bool_from_any(x: Any) -> bool:
+    """ Converts bool/float/int/str to bool or raises TypeError """
+    
+    if isinstance(x, bool):
+        return x
+    elif isinstance(x, (float, int)):
+        return x == 1.
```
tmbo, line 169:
```suggestion
        return x == 1.0
```

rasa/shared/core/slots.py
```diff
@@ -160,6 +160,21 @@ def feature_dimensionality(self) -> int:
...
+    elif isinstance(x, str) and x.isnumeric():
+        return float(x) == 1.
```
tmbo, line 171:
```suggestion
        return float(x) == 1.0
```

rasa/shared/core/slots.py
```diff
@@ -160,6 +160,21 @@ def feature_dimensionality(self) -> int:
...
+    elif isinstance(x, str):
+        return x.strip().lower() == "true"
```
tmbo, line 173: this will happily accept slot values like `"Berlin"` and set the boolean slot to `0.0`.

I think we should explicitly test for both cases, `"true"` and `"false"` strings and raise `ValueError` otherwise

tmbo (commented):
nice catch! let's add some tests to prevent this in the future

tests/shared/core/test_slots.py
```diff
@@ -96,11 +96,15 @@ def invalid_value(self, request):
         params=[
             (None, [0, 0]),
             (True, [1, 1]),
-            ("9", [1, 1]),
-            (12, [1, 1]),
+            ("9", [1, 0]),
+            (12, [1, 0]),
```
JEM-Mosig, line 100: @tmbo I don't know why this was supposed to be 1.

tmbo (approved):
looks good :+1: can you please add tests for the two exceptions?

JEM-Mosig:
Ups... I clicked on "ready to review" and assumed that would tag @tmbo , but it added half a dozen people! Sorry :grimacing:

wochinge (commented):
The change is fine for `rasa.shared` (it still needs a proper review)

changelog/6601.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Fixed a bug in the featurization of the boolean slot type. Previously, to set a slot value to "true", 
+you had to set it to "1", which is in conflict with the documentation.
```
erohmensing, line 2: You didn't have to set it to `"1"`, just to `true` (no quotes) AFAIK
```
- slot{"my_bool_slot": true}
```

JEM-Mosig: Ok, but that's still confusing users.

JEM-Mosig: I guess I make this an `improvement` then, instead of a bug fix? Also, I am not sure about the `float(x) == 1.0`. Perhaps it'd be better to have `float(x) != 0.0` as before, though that's a bit odd I think.

erohmensing: Oh I totally agree that this is a good improvement! (or bugfix if you expected `"true"` to work). Just wanted to point out that saying you had to set it to 1 would probably confuse a lot of people whose bots were working as expected before.

ortsaCniveK: @erohmensing just FYI that while just `true` might work for training time, I think there might be an issue with the JSON validation schema since I got [this](https://forum.rasa.com/t/questions-regarding-data-validation-script-and-with-slots-entites-and-followups-from-custom-actions/33405/8?u=nivek) error message when changing stories to use just `true`.

JEM-Mosig: I can confirm this error.

rasa/utils/schemas/stories.yml
```diff
@@ -36,7 +36,7 @@ mapping:
                 - type: "map"
                   mapping:
                     regex;(.*):
-                      type: "text"
+                      type: "any"
```
JEM-Mosig, line 39: @erohmensing This fixes the yaml problem and `true` without quotes works again. But I have no idea if this is the proper way of doing this.

degiz:
@JEM-Mosig 
I've noticed this [change](https://github.com/RasaHQ/rasa/blob/d978bc1ab310654daf915ab5e8482a49ebdca52c/rasa/utils/schemas/stories.yml#L39). Is there a reason for the type to be that generic? If there's a problem parsing boolean, then we can use the bool type instead, otherwise the schema validation is not that effective :(

JEM-Mosig:
> @JEM-Mosig
> I've noticed this [change](https://github.com/RasaHQ/rasa/blob/d978bc1ab310654daf915ab5e8482a49ebdca52c/rasa/utils/schemas/stories.yml#L39). Is there a reason for the type to be that generic? If there's a problem parsing boolean, then we can use the bool type instead, otherwise the schema validation is not that effective :(
I wasn't sure about this change. Strings and booleans should definitely be allowed. Also lists, since we have a list-type slot. And since you can define your own slot type, really anything should be allowed I guess...

degiz:
I'd rather allow `bool`, `text`, and list separately, and extend it on-the-go, as required. I have some high hopes that the detailed schema will help first adapters of `2.0` find error faster 😬

Issue #6588, opened 2020-09-07
`rasa data split nlu` no longer works


akelad:
Part of #6475 

**Rasa version**: master

Running `rasa data split nlu` creates an empty `train_test_split` directory as of the 2.0 changes to the data format. It's supposed to generate `test_data` and `training_data` files (they used to be .json, but we should probably change that to be .yaml as well)

I tested this on the `formbot` and `responseselectorbot` examples, both with the same result.

Potentially related to this issue: https://github.com/RasaHQ/rasa/issues/6389

Whoever fixes it, please make sure to update the `CLI` documentation as well!

m-vdb:
@akelad do you know why the default format was `json` in the first place?

akelad:
it's archaic functionality from when json was the only format we had :D

Issue #6363, opened 2020-08-10
Re-add test for `test_data_split_nlu` after YAML writer is ready


wochinge:
The changes of [this PR](https://github.com/RasaHQ/rasa/pull/6329/files) break the test for `test_data_split_nlu`. As this caused by the currently missing YAML writer implementation, and still works for old MD projects, the test was commented out. The test will work as soon as the writer for YAML data is implemented and merged (e.g. https://github.com/RasaHQ/rasa/issues/6252).

[It's these lines](https://github.com/RasaHQ/rasa/blob/ede603fd2b63f101174d4312ed77f710aaaeec3a/tests/cli/test_rasa_data.py#L23).

Blocked by:
- [ ] yaml writers for training data merged

PR #6591, opened 2020-09-07
New Response Selector training data format

dakshvar22:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/6480

The new training data format for using retrieval intents looks like this - 

nlu.yaml
```
- intent: chitchat/ask_name
  examples: |
    - What do people call you?
    - Do you have a name for yourself?

- intent: chitchat/ask_weather
  examples: |
    - Oh, do you mind checking the weather for me please?
    - I like sunny days in Berlin.
```

responses.yaml - 
```
responses:
  utter_chitchat/ask_name:
  - image: "https://i.imgur.com/zTvA58i.jpeg"
    text: hello, my name is retrieval bot.
  - text: Oh yeah, I am called the retrieval bot.

  utter_chitchat/ask_weather:
  - text: Oh, it does look sunny right now in Berlin.
    image: "https://i.imgur.com/vwv7aHN.png"
  - text: I am not sure of the whole week but I can see the sun is out today.
```

domain.yaml
```
intents:
    - chitchat
```

rules.yml
```
- rule: Response with a chitchat utterance whenever user indulges in some chitchat
  steps:
  - intent: chitchat
  - action: utter_chitchat
```

Two major changes - 

1. The bot builder doesn't need to know any difference between an utterance action and a retrieval action. Both of these actions now have the same `utter_` prefix
2. Response templates for response selector also now start with `utter_` prefix and hence are more consistent with other utterance templates.

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] updated the documentation
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

Issue #6566, opened 2020-09-03
Enable storing NLU, Strories, Rules in the same file


degiz:
Part of #6475

- Make it possible to store `nlu`,`stories`, `rules` keys within one `YAML` file.
- Domain stays out of the picture at the moment, because of `responses` key it shares with `RasaYAMLReader`, this should be implemented separately later.

PR #6604, opened 2020-09-08
Fix data split cli

m-vdb:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/6588
- Fixes https://github.com/RasaHQ/rasa/issues/6363


rasa/nlu/training_data/util.py
```diff
@@ -50,7 +50,7 @@ def get_file_format(resource_name: Text) -> Text:
     if fformat in [loading.MARKDOWN, loading.RASA_YAML] and all(
         f == fformat for f in file_formats
     ):
-        return fformat
+        return {loading.MARKDOWN: "md", loading.RASA_YAML: "yml"}[fformat]
```
m-vdb, line 53: @degiz I'd rather modify the value of the `RASA_YAML` constant to `"yml"` (instead of `"rasa_yaml"`). Is there any reason we're making this distinction? As far as I know we didn't make the distinction for the markdown format. That's what causing the issues with `rasa data split nlu`

degiz: Probably it's there to distinguish the `yml` as a file extension from the `yml` as a training data format, see `YAML_FILE_EXTENSIONS`.

Seems to me that this `get_file_format()` method, which is used only here btw, is broken by design, because from the usage it should return a file extension (defined in `rasa/data.py`) but in fact it return format name from `rasa/nlu/training_data/loading.py`.

m-vdb: yes exactly 👍 I'll keep the constant value and cleanup the design of `get_file_format()`

docs/docs/command-line-interface.mdx
```diff
@@ -255,8 +255,8 @@ If you have NLG data for retrieval actions, this will be saved to seperate files
 ```bash
 ls train_test_split
 
-      nlg_test_data.yml     test_data.json
-      nlg_training_data.yml training_data.json
+      nlg_test_data.yml     test_data.yml
+      nlg_training_data.yml training_data.yml
```
m-vdb, line 259: @akelad after reading the docs page on `rasa data split nlu`, this seemed like the only change required. Did you have something else in mind in your description of https://github.com/RasaHQ/rasa/issues/6588 ?

akelad: oh, since the response selector training data format changed, i wasn't actually sure that the nlg stuff is still relevant. but @dakshvar22 [PR](https://github.com/RasaHQ/rasa/pull/6591) isn't merged yet, so it may be up to him!

dakshvar22: It will still be relevant for the time being because response templates for responses still need to be defined outside the regular nlu, stories, rules yml files. See https://github.com/RasaHQ/rasa/issues/6566 cc @degiz

akelad: wait so `responses:` separately and `responses:` under domain function differently? that seems error prone

dakshvar22: I'll wait for sasha to answer that since he has been working on it.

rasa/cli/data.py
```diff
@@ -5,12 +5,12 @@
...
 
-import rasa.shared.data
+import rasa.cli.utils as cli_utils
+import rasa.shared.data as shared_data
 from rasa.cli.arguments import data as arguments
-import rasa.cli.utils
-from rasa.cli.utils import print_error_and_exit, print_info, print_warning
 from rasa.constants import DEFAULT_DATA_PATH
-from rasa.shared.data import is_valid_filetype
+from rasa.shared.nlu.training_data.loading import load_data
```
wochinge, line 12: please import the modules and not the functions directly

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -313,8 +313,8 @@ def persist_nlu(self, filename: Text = DEFAULT_TRAINING_DATA_OUTPUT_PATH) -> Non
         elif rasa.shared.data.is_likely_yaml_file(filename):
             rasa.shared.utils.io.write_text_file(self.nlg_as_yaml(), filename)
         else:
-            ValueError(
-                "Unsupported file format detected. Supported file formats are 'json' "
+            raise ValueError(
```
wochinge, line 316: 🙈

rasa/shared/nlu/training_data/util.py
```diff
@@ -50,6 +52,15 @@ def check_duplicate_synonym(
 
 
 def get_file_format(resource_name: Text) -> Text:
+    """
+    Get the file format of a resource. It supports both a folder and
+    a file, and tries to guess the format as follows:
```
wochinge, line 57: please use the google docstring format which also describe parameters / return values

rasa/shared/nlu/training_data/util.py
```diff
@@ -60,15 +71,17 @@ def get_file_format(resource_name: Text) -> Text:
     file_formats = list(map(lambda f: loading.guess_format(f), files))
 
     if not file_formats:
-        return "json"
+        return DEFAULT_FILE_FORMAT
+
+    knwon_file_formats = defaultdict(
```
wochinge, line 76: optional-ish: I actually had to look this up :-D I think it's simpler if it's just a regular dictionary and you use `dict.get(fformat, DEFAULT_FILE_FORMAT)` . Actually using a set would be even better, wouldn't it?

wochinge (commented):
thanks for taking this over :pray:

degiz (changes requested):
Thanks a lot for taking this over! 🚀 

I've tried this:
```bash
rasa data split nlu -u ./examples/responseselectorbot/data --out ~/Desktop/split
```
And it created 4 files with the same content. Then I've tried the same with a `moodbot`, and it gave me 4 empty files :(

Also one file is named `nlg_training_data..yml`, with two dots in the middle.

So I don't think that existing test is complete until it verifies the file's content.

rasa/shared/nlu/training_data/util.py
```diff
@@ -19,6 +20,7 @@
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_FILE_FORMAT = "yml"
```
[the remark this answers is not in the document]
m-vdb: well it's a `set()`, so I don't know if we want to define the default file format there or not :/

rasa/shared/nlu/training_data/util.py
m-vdb: yeah, its not needed. It was useful before but now it's useless

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -311,10 +311,10 @@ def persist_nlu(self, filename: Text = DEFAULT_TRAINING_DATA_OUTPUT_PATH) -> Non
         elif rasa.shared.data.is_likely_markdown_file(filename):
             rasa.shared.utils.io.write_text_file(self.nlu_as_markdown(), filename)
         elif rasa.shared.data.is_likely_yaml_file(filename):
-            rasa.shared.utils.io.write_text_file(self.nlg_as_yaml(), filename)
+            rasa.shared.utils.io.write_text_file(self.nlu_as_yaml(), filename)
```
m-vdb, line 314: this is the bad guy

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -340,12 +340,12 @@ def get_nlg_persist_filename(nlu_filename: Text) -> Text:
...
             # there is a special case: NLU is in json format, in this case we use
             # md as we do not have a NLG json format
-            extension = "md"
+            extension = ".md"
         # Add nlg_ as prefix and change extension to .md
         filename = (
             Path(nlu_filename)
             .with_name("nlg_" + Path(nlu_filename).name)
-            .with_suffix("." + extension)
+            .with_suffix(extension)
```
m-vdb, line 348: this is another bad guy

degiz (changes requested):
Thanks for fixing that! 
Existing tests didn't catch that, so I still think we need to extend them to be able to verify the content of the generated files.

degiz:
@m-vdb I have! I just think that it won't catch a situation when all four generated files have the same content.

m-vdb:
🤔 hum OK. I wasn't able to reproduce this one. I am not sure I'll be able to finish it before leaving. Can you take it over?

degiz (approved):
Fixed failing tests and accidentally fixed a bug with `--persist-nlu-data` flag for `rasa train`

Issue #6557, opened 2020-09-03
Restrict length of rules to one user turn 


erohmensing:
**Description of Problem**:

The RulePolicy is meant to replace the Form, Mapping, and Fallback policies. It's not meant to hardcode a state machine. We want to make sure that users understand this and provide ample training data to their machine learning policy to make predictions incase of unseen conversation paths. 

**Overview of the Solution**:

We restrict the length of rules to one _user_ turn (i.e. there can still be multiple bot events, including e.g. a form being filled and its subsequent submission.

wochinge:
How about a warning which can explicitly be disabled in the `RulePolicy` config? (should be `enabled: True` by default)

Ghostvv:
in my opinion warning is not enough

tmbo:
we shouldn't add undocumented options, either we think there is a use case for an option and describe that or there shouldn't be an option

Ghostvv:
we'll add an option to docs, just not advertise it

PR #6607, opened 2020-09-08
restrict rule snippets to single turn user inputs

Ghostvv:
**Proposed changes**:
- restrict rule snippets to single turn user inputs
- fixes https://github.com/RasaHQ/rasa/issues/6557

### DIFF
--- a/docs/docs/rules.mdx
+++ b/docs/docs/rules.mdx
@@ -54,6 +54,10 @@ rules:
 This example rule applies at the start of conversation as well as when the user decides
 to a send a message with an intent `greet` in the middle of an ongoing conversation.
 
+**Rules are not meant to hardcode a state machine.**
+Rules are restricted to have only one user turn, but there can still be multiple
+bot events, including e.g. a form being filled and its subsequent submission.
+
 ### Rules for the Conversation Start
 
 To write a rule which only applies at the beginning of a conversation, add a
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -3,7 +3,8 @@ from typing import List, Dict, Text, Optional, Any, Set, TYPE_CHECKING
 
 import json
 
-from rasa.core.events import FormValidation
+import rasa.shared.utils.io
+from rasa.core.events import FormValidation, UserUttered
 from rasa.core.featurizers.tracker_featurizers import TrackerFeaturizer
 from rasa.core.domain import Domain, InvalidDomain, State
 from rasa.core.interpreter import NaturalLanguageInterpreter
@@ -52,10 +53,27 @@ DO_NOT_VALIDATE_LOOP = "do_not_validate_loop"
 DO_NOT_PREDICT_LOOP_ACTION = "do_not_predict_loop_action"
 
 
+class InvalidRules(Exception):
+    """Exception that can be raised when domain is not valid."""
+
+    def __init__(self, message) -> None:
+        self.message = message
+
+    def __str__(self):
+        # return message in error colours
+        return rasa.shared.utils.io.wrap_with_color(
+            self.message, color=rasa.shared.utils.io.bcolors.FAIL
+        )
+
+
 class RulePolicy(MemoizationPolicy):
     """Policy which handles all the rules"""
 
+    # rules use explicit json strings
     ENABLE_FEATURE_STRING_COMPRESSION = False
+
+    # number of user inputs that is allowed in case rules are restricted
+    ALLOWED_NUMBER_OF_USER_INPUTS = 1
 
     @staticmethod
     def supported_data() -> SupportedData:
@@ -74,6 +92,7 @@ class RulePolicy(MemoizationPolicy):
         core_fallback_threshold: float = 0.3,
         core_fallback_action_name: Text = ACTION_DEFAULT_FALLBACK_NAME,
         enable_fallback_prediction: bool = True,
+        restrict_rules: bool = True,
     ) -> None:
         """Create a `RulePolicy` object.
 
@@ -95,6 +114,7 @@ class RulePolicy(MemoizationPolicy):
         self._core_fallback_threshold = core_fallback_threshold
         self._fallback_action_name = core_fallback_action_name
         self._enable_fallback_prediction = enable_fallback_prediction
+        self._restrict_rules = restrict_rules
 
         # max history is set to `None` in order to capture any lengths of rule stories
         super().__init__(
@@ -223,6 +243,23 @@ class RulePolicy(MemoizationPolicy):
                 lookup[feature_key] = DO_NOT_PREDICT_LOOP_ACTION
         return lookup
 
+    def _check_rule_trackers(self, rule_trackers):
+        bad_rules = []
+        for tracker in rule_trackers:
+            number_of_user_uttered = sum(
+                isinstance(event, UserUttered) for event in tracker.events
+            )
+            if number_of_user_uttered > self.ALLOWED_NUMBER_OF_USER_INPUTS:
+                bad_rules.append(tracker.sender_id)
+
+        if bad_rules:
+            raise InvalidRules(
+                f"Found rules {bad_rules} that contain more than "
+                f"{self.ALLOWED_NUMBER_OF_USER_INPUTS} user inputs. "
+                f"Rules are not meant to hardcode a state machine. "
+                f"Please use stories for these cases."
+            )
+
     def train(
         self,
         training_trackers: List[TrackerWithCachedStates],
@@ -243,6 +280,9 @@ class RulePolicy(MemoizationPolicy):
             rule_trackers_as_states,
             rule_trackers_as_actions,
         ) = self.featurizer.training_states_and_actions(rule_trackers, domain)
+
+        if self._restrict_rules:
+            self._check_rule_trackers(rule_trackers)
 
         rules_lookup = self._create_lookup_from_states(
             rule_trackers_as_states, rule_trackers_as_actions
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -107,8 +107,6 @@ class TEDPolicy(Policy):
           (https://arxiv.org/abs/1709.03856) idea.
     """
 
-    SUPPORTS_ONLINE_TRAINING = True
-
     # please make sure to update the docs when changing a default parameter
     defaults = {
         # ## Architecture of the used neural network
--- a/rasa/core/policies/memoization.py
+++ b/rasa/core/policies/memoization.py
@@ -53,8 +53,6 @@ class MemoizationPolicy(Policy):
 
     ENABLE_FEATURE_STRING_COMPRESSION = True
 
-    SUPPORTS_ONLINE_TRAINING = True
-
     USE_NLU_CONFIDENCE_AS_SCORE = False
 
     @staticmethod
--- a/rasa/core/policies/policy.py
+++ b/rasa/core/policies/policy.py
@@ -62,7 +62,6 @@ class SupportedData(Enum):
 
 
 class Policy:
-    SUPPORTS_ONLINE_TRAINING = False
 
     @staticmethod
     def supported_data() -> SupportedData:


Ghostvv:
@erohmensing I didn't find the description of `RulePolicy` config in the docs where I should include the parameter

wochinge:
Will have a look at this asap. I'm a bit swamped at the moment, sorry

Ghostvv:
@wochinge types failed after merging master

examples/rules/data/rules.yml
```diff
@@ -98,21 +98,19 @@ rules:
...
   - intent: deny
   - action: utter_ask_stop
-  - intent: deny
-  - action: utter_continue
 
-- rule: Detailed FAQ did not help - stop
-  condition:
-  - slot_was_set:
-    - detailed_faq: true
+- rule: Stop - stop
```
wochinge, line 102:
```suggestion
- rule: User decides to stop
```

examples/rules/data/rules.yml
```diff
@@ -98,21 +98,19 @@ rules:
...
   steps:
-  - action: utter_detailed_faq
-  - action: utter_ask_did_help
-  - intent: deny
   - action: utter_ask_stop
   - intent: affirm
   - action: utter_stop
 
+- rule: Stop - continue
```
wochinge, line 108:
```suggestion
- rule: User decides to continue
```

rasa/core/policies/rule_policy.py
```diff
@@ -53,11 +54,28 @@
 DO_NOT_PREDICT_LOOP_ACTION = "do_not_predict_loop_action"
 
 
+class InvalidRule(Exception):
+    """Exception that can be raised when domain is not valid."""
+
+    def __init__(self, message) -> None:
+        self.message = message
+
+    def __str__(self):
```
wochinge, line 63:
```suggestion
    def __str__(self) -> Text:
```

rasa/core/policies/rule_policy.py
```diff
@@ -53,11 +54,28 @@
 DO_NOT_PREDICT_LOOP_ACTION = "do_not_predict_loop_action"
 
 
+class InvalidRule(Exception):
+    """Exception that can be raised when domain is not valid."""
+
+    def __init__(self, message) -> None:
```
wochinge, line 60:
```suggestion
    def __init__(self, message: Text) -> None:
```

rasa/core/policies/rule_policy.py
```diff
@@ -75,6 +93,7 @@ def __init__(
         core_fallback_threshold: float = 0.3,
         core_fallback_action_name: Text = ACTION_DEFAULT_FALLBACK_NAME,
         enable_fallback_prediction: bool = True,
+        restrict_rules: bool = True,
```
wochinge, line 96: As the parameter name is user interfacing, I think this name would make it clearer what the parameter is doing.
```suggestion
        restrict_rule_length: bool = True,
```

rasa/core/policies/rule_policy.py
```diff
@@ -224,6 +244,23 @@ def _create_loop_unhappy_lookup_from_states(
                 lookup[feature_key] = DO_NOT_PREDICT_LOOP_ACTION
         return lookup
 
+    def _check_rule_trackers(self, rule_trackers):
```
wochinge, line 247:
```suggestion
    def _check_max_number_of_rule_turns(self, rule_trackers: List[TrackerWithCachedStates]) -> None:
```

rasa/core/policies/rule_policy.py
```diff
@@ -224,6 +244,23 @@ def _create_loop_unhappy_lookup_from_states(
...
+        bad_rules = []
```
wochinge, line 248:
```suggestion
        rules_exceeding_max_turns = []
```

rasa/core/policies/rule_policy.py
```diff
@@ -224,6 +244,23 @@ def _create_loop_unhappy_lookup_from_states(
...
+        for tracker in rule_trackers:
+            number_of_user_uttered = sum(
+                isinstance(event, UserUttered) for event in tracker.events
+            )
+            if number_of_user_uttered > self.ALLOWED_NUMBER_OF_USER_INPUTS:
+                bad_rules.append(tracker.sender_id)
+
+        if bad_rules:
+            raise InvalidRule(
+                f"Found rules {bad_rules} that contain more than "
```
wochinge, line 258: we should do something like `", ".join(bad_rules)`

rasa/core/policies/rule_policy.py
```diff
@@ -224,6 +244,23 @@ def _create_loop_unhappy_lookup_from_states(
...
+                f"{self.ALLOWED_NUMBER_OF_USER_INPUTS} user inputs. "
+                f"Rules are not meant to hardcode a state machine. "
+                f"Please use stories for these cases."
```
wochinge, line 261: let's include the link to the rule documentation. there is a constant in `rasa.constants` (DOCS_URL_RULES)

tests/core/policies/test_rule_policy.py
```diff
@@ -93,8 +93,37 @@ def _form_activation_rule(
...
+actions:
+- {UTTER_GREET_ACTION}
+    """
+    )
+    policy = RulePolicy(restrict_rules=True)
+    greet_events = [
+        UserUttered(intent={"name": GREET_INTENT_NAME}),
+        ActionExecuted(UTTER_GREET_ACTION),
+        ActionExecuted(ACTION_LISTEN_NAME),
+    ]
```
wochinge, line 110: let's re-use the events from `GREET_RULE`

tests/core/policies/test_rule_policy.py
```diff
@@ -93,8 +93,37 @@ def _form_activation_rule(
...
+        + greet_events * (policy.ALLOWED_NUMBER_OF_USER_INPUTS + 1),
+    )
+    forbidden_rule.is_rule_tracker = True
+    with pytest.raises(InvalidRule):
+        policy.train([forbidden_rule], domain, RegexInterpreter())
+
+
 def test_rule_policy_has_max_history_none():
-    policy = RulePolicy()
+    policy = RulePolicy(restrict_rules=False)
```
wochinge, line 126: why is this explicitly set to `False`?

tests/core/policies/test_rule_policy.py
```diff
@@ -108,7 +137,7 @@ def test_faq_rule():
     """
     )
 
-    policy = RulePolicy()
+    policy = RulePolicy(restrict_rules=False)
```
wochinge, line 140: why is this explicitly set to `False`?

tests/core/policies/test_rule_policy.py
```diff
@@ -613,7 +642,7 @@ async def test_form_unhappy_path_no_validation_from_rule():
         is_rule_tracker=True,
     )
 
-    policy = RulePolicy()
+    policy = RulePolicy(restrict_rules=False)
```
wochinge, line 645: this the only test which actually requires `restrict_rules=False`

docs/docs/rules.mdx
```diff
@@ -54,6 +54,10 @@ rules:
 This example rule applies at the start of conversation as well as when the user decides
 to a send a message with an intent `greet` in the middle of an ongoing conversation.
 
+**Rules are not meant to hardcode a state machine.**
+Rules are restricted to have only one user turn, but there can still be multiple
+bot events, including e.g. a form being filled and its subsequent submission.
+
```
wochinge, line 60:
```suggestion

:::caution
Rules are not meant to hardcode a state machine.
Rules are restricted to have only one user turn, but there can still be multiple
bot events, including e.g. a form being filled and its subsequent submission.
:::
```

docs/docs/rules.mdx
```diff
@@ -54,6 +54,10 @@ rules:
 This example rule applies at the start of conversation as well as when the user decides
 to a send a message with an intent `greet` in the middle of an ongoing conversation.
 
+**Rules are not meant to hardcode a state machine.**
```
wochinge, line 57: @b-quachtran Can you please double check if that makes sense?

wochinge (commented):
Thanks for taking over the implementation!

wochinge:
re type checks: https://rasa-hq.slack.com/archives/C0125DKLBEW/p1599727956128700

rasa/core/policies/rule_policy.py
Ghostvv: I'm not sure, since it doesn't restrict the length of the rule. It can contain arbitrary number of actions

tests/core/policies/test_rule_policy.py
Ghostvv: but it contain rule snippet action, which will need to be excluded, not sure whether it'll make it easier?

tests/core/policies/test_rule_policy.py
Ghostvv: I just set it to `False` in all our tests, so that if we change restriction, it'll not break other tests that don't test for it

tests/core/policies/test_rule_policy.py
Ghostvv: yes, but I just set it to False in all our tests, so that if we change restriction, it'll not break other tests that don't test for it

Ghostvv: to separate the test logics

rasa/core/policies/rule_policy.py
Ghostvv: Also, we might update the restriction if we get feedback that current restriction is bad

rasa/core/policies/rule_policy.py
Ghostvv: since max number of user turns in rule is only one possible restriction, I renamed it to `_check_rule_restriction`

Ghostvv:
@wochinge since they are related, I added checking rules to this PR (also to avoid multiple merge conflicts resolving created by. decoupling)

rasa/shared/core/trackers.py
```diff
@@ -367,7 +367,10 @@ def init_copy(self) -> "DialogueStateTracker":
         """Creates a new state tracker with the same initial values."""
 
         return DialogueStateTracker(
-            DEFAULT_SENDER_ID, self.slots.values(), self._max_event_history
+            DEFAULT_SENDER_ID,
```
wochinge, line 370: :rocket:

rasa/core/policies/ted_policy.py
```diff
@@ -320,6 +318,13 @@ def train(
     ) -> None:
         """Train the policy on given training trackers."""
 
+        if not training_trackers:
+            logger.error(
```
wochinge, line 322: it's rather a warning than an error

rasa/core/test.py
```diff
@@ -389,7 +389,7 @@ def _collect_user_uttered_predictions(
     return user_uttered_eval_store
 
 
-def _emulate_form_rejection(partial_tracker: DialogueStateTracker) -> None:
+def emulate_form_rejection(partial_tracker: DialogueStateTracker) -> None:
```
wochinge, line 392: please add a docstring if you make it public

rasa/core/policies/rule_policy.py
wochinge: > Also, we might update the restriction if we get feedback that current restriction is bad

I think we shouldn't write code for things which **maybe** will happen. Often enough this doesn't happen and then you're stuck with the complexity for no reason.

> I'm not sure, since it doesn't restrict the length of the rule. It can contain arbitrary number of actions

Good point 🤔 
@b-quachtran Do you have suggestions / objections regarding keep `restrict_rules`?

rasa/core/policies/rule_policy.py
```diff
@@ -71,6 +94,8 @@ def __init__(
         core_fallback_threshold: float = 0.3,
         core_fallback_action_name: Text = ACTION_DEFAULT_FALLBACK_NAME,
         enable_fallback_prediction: bool = True,
+        restrict_rules: bool = True,
+        check_rules_with_stories: bool = True,
```
wochinge, line 98: I'd prefer this one as it explains what is checked and doesn't expose to much of the "how"
```suggestion
        check_for_contradictions: bool = True,
```

rasa/core/policies/rule_policy.py
```diff
@@ -211,15 +241,132 @@ def _create_loop_unhappy_lookup_from_states(
...
+                # some action other than active_loop is predicted in unhappy path,
                 # therefore active_loop shouldn't be predicted by the rule
                 not is_prev_action_listen_in_state(states[-1])
-                and action not in {ACTION_LISTEN_NAME, active_loop}
+                and action != active_loop
             ):
                 lookup[feature_key] = DO_NOT_PREDICT_LOOP_ACTION
         return lookup
 
+    def _check_rule_restriction(self, rule_trackers):
```
wochinge, line 252: types please

rasa/core/policies/rule_policy.py
```diff
@@ -49,11 +55,28 @@
 DO_NOT_PREDICT_LOOP_ACTION = "do_not_predict_loop_action"
 
 
+class InvalidRule(Exception):
+    """Exception that can be raised when domain is not valid."""
```
wochinge, line 59: I think that one needs to be updated ;-)

rasa/core/policies/rule_policy.py
```diff
@@ -211,15 +241,132 @@ def _create_loop_unhappy_lookup_from_states(
...
+            rasa.core.test.emulate_form_rejection(tracker)
+            predicted_action_name = self._predict_next_action(
+                tracker, domain, interpreter
+            )
+            if not predicted_action_name or predicted_action_name == gold_action_name:
+                return
+
+        tracker_type = "rule" if tracker.is_rule_tracker else "story"
+        return (
+            f"Action '{gold_action_name}' in {tracker_type} "
```
wochinge, line 308: Can we add which tracker is contradicting which other tracker?

rasa/core/policies/rule_policy.py
```diff
@@ -211,15 +241,132 @@ def _create_loop_unhappy_lookup_from_states(
...
+            desc="Processed trackers",
+            disable=rasa.shared.utils.io.is_logging_disabled(),
+        )
+        for tracker in pbar:
+            running_tracker = tracker.init_copy()
+            running_tracker.sender_id = tracker.sender_id
+            # the first action is always unpredictable
+            next_action_is_unpredictable = True
+            for event in tracker.applied_events():
+                # do not run prediction on unpredictable actions
```
wochinge, line 334: great comments :rocket:

rasa/core/policies/rule_policy.py
```diff
@@ -211,15 +241,132 @@ def _create_loop_unhappy_lookup_from_states(
...
+                if error_message:
+                    error_messages.append(error_message)
+
+                running_tracker.update(event)
+
+        self._silent_prediction = False  # turn off silent logging
+        if error_messages:
+            error_messages.append(
+                "Please update your stories and rules so that "
+                "they don't contradict each other."
```
wochinge, line 364:
```suggestion
                "they don't contradict each other. "
```

rasa/core/policies/rule_policy.py
```diff
@@ -211,15 +241,132 @@ def _create_loop_unhappy_lookup_from_states(
...
+            )
+            raise InvalidRule("\n".join(error_messages))
```
wochinge, line 366: If there is a newline for every error message, then we should probably format it with `- ` to make it a list?

tests/core/policies/test_rule_policy.py
wochinge: I would not do that. If we change restrictions and tests break then this might provide valuable insight. Also I think it's weird to test a policy mainly in some sort of "experimental mode" .

rasa/core/policies/rule_policy.py
```diff
@@ -263,11 +413,18 @@ def train(
...
-        #  to check that stories don't contradict rules
+        # make this configurable because checking might take a lot of time
+        if self._check_rules_with_stories:
+            # using trackers here might not be the most efficient way, however
+            # it allows us to directly test `predict_action_probabilities` method
+            self._find_contradicting_rules(training_trackers, domain, interpreter)
 
         logger.debug(f"Memorized '{len(self.lookup[RULES])}' unique rules.")
 
+    def _output_debug_logs(self, message: Text) -> None:
```
wochinge, line 424: I think it's cleaner and more robust to temporarily modify the log level of the module logger using `logger.setLevel(logging.WARNING)`

wochinge (commented):
Slick 👍 

Side note: I think it would have been better to split up this PR into two PRs to make the changeset more concise.

tests/core/policies/test_rule_policy.py
Ghostvv: ok, will update tests, and put restriction false in only one of them with comment

rasa/core/policies/rule_policy.py
Ghostvv: > I think we shouldn't write code for things which maybe will happen.
ok

rasa/core/policies/ted_policy.py
Ghostvv: it is an error below if model data is empty

rasa/core/policies/rule_policy.py
Ghostvv: why do we need a space?

rasa/core/policies/rule_policy.py
Ghostvv: 👍 I forgot about it

rasa/core/policies/rule_policy.py
Ghostvv: this is a problem, we don't know other tracker, we know that rules have learned smth different. If we want to know which rule predicted it, we need to significantly update tracker featurization

Ghostvv: or maybe not

Ghostvv: nope, it's not that easy, also rules could contain intersecting identical pieces

docs/docs/rules.mdx
```diff
@@ -54,6 +54,14 @@ rules:
 This example rule applies at the start of conversation as well as when the user decides
 to a send a message with an intent `greet` in the middle of an ongoing conversation.
 
+After training, `RulePolicy` will perform a check to make sure that the rules don't
```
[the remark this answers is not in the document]
Ghostvv: added example

rasa/core/policies/rule_policy.py
Ghostvv: actually the comment needs to be moved

rasa/core/policies/ted_policy.py
Ghostvv: I just reused the old code

erohmensing:
@Ghostvv any info about configuring a specific policy should go under that policy on the policies page

rasa/core/policies/rule_policy.py
Ghostvv: anyhow, up for renaming

Ghostvv:
@erohmensing updated

wochinge:
@erohmensing @b-quachtran Could you please give the docs and parameters name a quick review? 🙌

docs/docs/policies.mdx
```diff
@@ -385,6 +385,16 @@ for details.
 The `Rule Policy` is a policy that handles conversation parts that follow
 a fixed behavior. Please see [Rules](./rules.mdx) for further information.
 
+You can remove rule restriction and check for contradictions by setting the following
```
wochinge, line 388: do we actually want to advertise this? (cc @akelad )

If we add this, we need to first explain what the "restrictions" are.

docs/docs/rules.mdx
```diff
@@ -54,6 +54,28 @@ rules:
...
+rules:
+- rule: Chitchat
+  steps:
+  - intent: chitchat
+  - action: utter_chitchat
+
+- rule: Greet instead of chitchat
+  steps:
+  - intent: chitchat
+  - action: utter_greet
```
wochinge, line 71: let's be super explicit
```suggestion
  - action: utter_greet. # `utter_greet` contradicts `utter_chitchat` from the rule above
```

rasa/core/policies/rule_policy.py
wochinge: Does it make sense to turn it around? `restrict` sounds like we are limiting something. Users might think they "unleash" the RulePolicy by setting this to `False`. Should we rather call the default something like  `safe_mode` ?

rasa/core/policies/rule_policy.py
```diff
@@ -212,15 +239,140 @@ def _create_loop_unhappy_lookup_from_states(
...
+
+    def _find_contradicting_rules(
+        self,
+        trackers: List[TrackerWithCachedStates],
+        domain: Domain,
+        interpreter: NaturalLanguageInterpreter,
+    ) -> None:
+        logger.debug("Started checking rules and stories for contradictions.")
+        # during training we run `predict_action_probabilities` to check for
+        # contradicting rules, silent prediction debug to avoid too many logs
```
wochinge, line 322:
```suggestion
        # contradicting rules. We silent prediction debug to avoid too many logs during these checks.
```

rasa/core/policies/rule_policy.py
```diff
@@ -290,6 +448,8 @@ def _does_rule_match_state(rule_state: State, conversation_state: State) -> bool
                     # it should be None or non existent in the state
                     value == SHOULD_NOT_BE_SET
                     and conversation_sub_state.get(key)
+                    # this is needed to test on rule snippets
```
wochinge, line 451: Why is this only needed during testing?

wochinge:
@Ghostvv I think we shouldn't model this as exception. The stack trace makes it look very ugly. I think we should rather use `print_error_and_exit`. 

Screenshot 2020-09-14 at 14 21 03


Another thing: When rules contradict the `RulePolicy` will always predict `action_default_fallback` which makes the error message confusing. Also, it will print two errors although the rules are contradicting with each other. Means it would be enough to print just one error (e.g. `Rule x contradicts with rule y` instead of `Rule x contradicts with rule y. Rule Y contradicts with Rule x`)

docs/docs/policies.mdx
akelad: we should document it, yes. But +1 to explaining what the restrictions are

docs/docs/rules.mdx
```diff
@@ -54,6 +54,28 @@ rules:
 This example rule applies at the start of conversation as well as when the user decides
 to a send a message with an intent `greet` in the middle of an ongoing conversation.
 
+After training, `RulePolicy` will perform a check to make sure that the rules don't
+contradict the stories and each other.
+For example the following two rules are contradicting:
```
akelad, line 59: should this be documented here and not under rule policy?

docs/docs/rules.mdx
```diff
@@ -54,6 +54,28 @@ rules:
...
+- rule: Greet instead of chitchat
+  steps:
+  - intent: chitchat
+  - action: utter_greet
+```
+
+:::caution
+Rules are not meant to hardcode a state machine.
+Rules are restricted to have only one user turn, but there can still be multiple
+bot events, including e.g. a form being filled and its subsequent submission.
```
akelad, line 77: i think this whole note might be gibberish to a user (esp. the mention of state machine).

akelad: I'll let @b-quachtran review anything further

docs/docs/policies.mdx
Ghostvv: the restriction is explained in `rules.mdx`

docs/docs/rules.mdx
Ghostvv: the `.` shouldn't be there, no?

rasa/core/policies/rule_policy.py
Ghostvv: because during prediction `SHOULD_NOT_BE_SET` doesn't exist

Ghostvv:
> Also, it will print two errors although the rules are contradicting with each other. Means it would be enough to print just one error

we cannot do it. We don't know which rule contradicts what

docs/docs/rules.mdx
Ghostvv: what do you mean by `here`?

docs/docs/rules.mdx
Ghostvv: I took it from the issue https://github.com/RasaHQ/rasa/issues/6557, and I thought it is rather good. Do you have a better suggestion?

Ghostvv:
> When rules contradict the RulePolicy will always predict action_default_fallback which makes the error message confusing.

that's what `RulePolicy` predicts by default. It could be `action_listen` or custom fallback depending on the settings. Do you have a suggestion for better phrasing?

rasa/core/policies/rule_policy.py
Ghostvv: Do you mean to create a constant with a string `"safe_mode"`?

Ghostvv: or creating `safe_mode: True` option?

docs/docs/rules.mdx
akelad: where it's documented currently - i don't understand why it's here and not under the RulePolicy page

docs/docs/rules.mdx
akelad: i'll let ben help you out with this, but "state machine" is something that users probably don't care about, or know about for that matter

wochinge:
> that's what RulePolicy predicts by default.

mhm, but there are rules and even they are contradicting none of them is predicting `action_default_fallback`. If the contradicting rules result in the `RulePolicy` predicting `action_default_fallback` then it doesn't much sense to add that information to the error message.

docs/docs/policies.mdx
wochinge: then let's link there

docs/docs/rules.mdx
wochinge: oh yes, sorry

rasa/core/policies/rule_policy.py
wochinge: then let's phrase the comment as something like `during training `SHOULD_NOT_BE_SET` is provided and hence we also have to check for the value of the slot state`

rasa/core/policies/rule_policy.py
wochinge: > or creating safe_mode: True option?

This one

Ghostvv:
but we test not what rules (as if rule snippets predict) we test what the whole rule policy predicts including all the complicated loop prediction system and fallback and default predictions. That's why I call `predict_action_probabilities` but if you think it'd be better to say:

`utter_chitchat` was not predicted by RulePolicy

I can change to that

docs/docs/rules.mdx
Ghostvv: oh, I thought it is common notion

docs/docs/rules.mdx
Ghostvv: From the look of it. This whole file describes the RulePolicy, so I added it here, I can move it to policies.mdx

docs/docs/policies.mdx
```diff
@@ -325,6 +325,15 @@ a fixed behavior. It makes predictions based on any `rules` you have in your
 training data. See the [Rules documentation](./rules.mdx) for further information
 on how to define rules.
 
+You can remove rule restriction and check for contradictions by setting the following
```
b-quachtran, line 328: Can this be explained in two separate sub-sections? That way there can be a full definition of what each param does (e.g. what is the rule restriction?) and why you would use it.

wochinge:
> That's why I call predict_action_probabilities but if you think it'd be better to say:

I think that's indeed better because users will intuitively look for `action_default_fallback` in their rules / stories and then not find it which is likely to be confusing.

docs/docs/rules.mdx
b-quachtran: Relevant to the comment I left in policies.mdx, this should be moved to a subsection under `RulePolicy` that includes a definition + example of contradicting stories.

docs/docs/rules.mdx
```diff
@@ -56,6 +56,28 @@ rules:
...
+  - action: utter_chitchat
+
+- rule: Greet instead of chitchat
+  steps:
+  - intent: chitchat
+  - action: utter_greet  # `utter_greet` contradicts `utter_chitchat` from the rule above
+```
+
+:::caution
+Rules are not meant to hardcode a state machine.
```
b-quachtran, line 77:
```suggestion
```

docs/docs/rules.mdx
```diff
@@ -56,6 +56,28 @@ rules:
...
+Rules are restricted to have only one user turn, but there can still be multiple
```
b-quachtran, line 78:
```suggestion
Rules are restricted to one user turn, but there can be multiple
```

b-quachtran (changes requested):
Looks good, just a few changes to the `RulePolicy` section's structure

rasa/core/policies/rule_policy.py
akelad: i honestly have no strong opinions on this. but i don't see any issues with `restrict_rules` tbh. `restrict_rule_length` kind of does make sense, even if it only restricts the number of user messages and not the number of actions. 
I would say up to you

wochinge: Ok, then let's leave it as it is 👍

wochinge:
the exception is actually fine. We can't do a `exit()` because that would break server mode .

rasa/core/policies/rule_policy.py
```diff
@@ -212,15 +242,137 @@ def _create_loop_unhappy_lookup_from_states(
...
+                if error_message:
+                    error_messages.append(error_message)
+
+                running_tracker.update(event)
+
+        logger.setLevel(logger_level)  # reset logger level
+        if error_messages:
+            error_messages = "\n".join(error_messages)
+            raise InvalidRule(
+                f"\nContradicting rules 🚨\n\n{error_messages}\n"
```
akelad, line 373:
```suggestion
                f"\nContradicting rules found🚨\n\n{error_messages}\n"
```

rasa/core/policies/rule_policy.py
```diff
@@ -212,15 +242,137 @@ def _create_loop_unhappy_lookup_from_states(
...
+                f"Please update your stories and rules so that they don't contradict "
+                f"each other."
```
akelad, line 375: qq: the error message says `- the prediction of the action 'loop_q_form' in rule 'Greet' is contradicting with another rule or story.`

Why don't we say which story it's contradicting with?

rasa/core/policies/rule_policy.py
```diff
@@ -212,15 +242,137 @@ def _create_loop_unhappy_lookup_from_states(
...
+            )
+
+        logger.debug("Found no contradictions between rules and stories")
```
akelad, line 378: should it say "Found no contradicting rules"? since rules can also contradict with other rules

rasa/core/policies/rule_policy.py
```diff
@@ -212,15 +242,137 @@ def _create_loop_unhappy_lookup_from_states(
...
+            )
+            if number_of_user_uttered > self.ALLOWED_NUMBER_OF_USER_INPUTS:
+                rules_exceeding_max_user_turns.append(tracker.sender_id)
+
+        if rules_exceeding_max_user_turns:
+            raise InvalidRule(
+                f"Found rules '{', '.join(rules_exceeding_max_user_turns)}' "
+                f"that contain more than {self.ALLOWED_NUMBER_OF_USER_INPUTS} "
+                f"user inputs. Rules are not meant to hardcode a state machine. "
+                f"Please use stories for these cases."
```
akelad, line 270:
```
rasa.core.policies.rule_policy.InvalidRule: Found rules 'Greet' that contain more than 1 user inputs. Rules are not meant to hardcode a state machine. Please use stories for these cases.
You can find more information about the usage of rules at https://rasa.com/docs/rasa/core/rules/. 
```
the error message is on the same line as `rasa.core.policies.rule_policy.InvalidRule` - can we make it so that it comes on a new line afterwards?

akelad: also would remove the sentence "Rules are not meant to hardcode a state machine"

rasa/core/policies/rule_policy.py
```diff
@@ -212,15 +242,137 @@ def _create_loop_unhappy_lookup_from_states(
...
+                isinstance(event, UserUttered) for event in tracker.events
+            )
+            if number_of_user_uttered > self.ALLOWED_NUMBER_OF_USER_INPUTS:
+                rules_exceeding_max_user_turns.append(tracker.sender_id)
+
+        if rules_exceeding_max_user_turns:
+            raise InvalidRule(
+                f"Found rules '{', '.join(rules_exceeding_max_user_turns)}' "
+                f"that contain more than {self.ALLOWED_NUMBER_OF_USER_INPUTS} "
+                f"user inputs. Rules are not meant to hardcode a state machine. "
```
akelad, line 269:
```suggestion
                f"user message. Rules are not meant to hardcode a state machine. "
```
?

rasa/core/policies/rule_policy.py
akelad: also not sure where the url is coming from, but the correct url will actually be: https://rasa.com/docs/rasa/rules

rasa/core/policies/rule_policy.py
```diff
@@ -212,15 +242,137 @@ def _create_loop_unhappy_lookup_from_states(
...
+                )
+                if error_message:
+                    error_messages.append(error_message)
+
+                running_tracker.update(event)
+
+        logger.setLevel(logger_level)  # reset logger level
+        if error_messages:
+            error_messages = "\n".join(error_messages)
+            raise InvalidRule(
```
akelad, line 372:
```
Contradicting rules 🚨

- the prediction of the action 'utter_greet' in story 'test' is contradicting with another rule or story.
Please update your stories and rules so that they don't contradict each other.
You can find more information about the usage of rules at https://rasa.com/docs/rasa/core/rules/. 
```
If the story is the contradicting one, then this error message makes less sense no? It says "contradicting rules" but only raises the story and not the rule

akelad (commented):
partial review on error messages

rasa/core/policies/rule_policy.py
wochinge: We don't know which story it's contradicting with.

docs/docs/policies.mdx
```diff
@@ -325,6 +325,48 @@ a fixed behavior. It makes predictions based on any `rules` you have in your
 training data. See the [Rules documentation](./rules.mdx) for further information
 on how to define rules.
 
+The `RulePolicy` has the following configuration options:
```
akelad, line 328:
```suggestion
The RulePolicy has the following configuration options:
```

docs/docs/policies.mdx
```diff
@@ -325,6 +325,48 @@ a fixed behavior. It makes predictions based on any `rules` you have in your
...
+* `core_fallback_threshold` (default: `0.3`): Please see the
+   [fallback documentation](fallback-handoff.mdx#handling-low-core-confidence) for
+   further information.
+* `core_fallback_action_name` (default: `action_default_fallback`): Please see the
+   [fallback documentation](fallback-handoff.mdx#handling-low-core-confidence) for
+   further information.
+* `enable_fallback_prediction` (default: `true`): Please see the
+   [fallback documentation](fallback-handoff.mdx#handling-low-core-confidence) for
+   further information.
+* `check_for_contradictions` (default: `true`): After training, the `RulePolicy` will
```
akelad, line 349:
```suggestion
* `check_for_contradictions` (default: `true`): After training, the RulePolicy will
```

rasa/core/policies/rule_policy.py
wochinge: it can raise for both rules and stories. Will do `Contradicting rules and stories`

docs/docs/policies.mdx
```diff
@@ -325,6 +325,48 @@ a fixed behavior. It makes predictions based on any `rules` you have in your
...
+   perform a check to make sure that the rules don't contradict the stories and each
```
akelad, line 350:
```suggestion
   perform a check to make sure that there are no rules that contradict with each other, or with any stories.
```

docs/docs/policies.mdx
```diff
@@ -325,6 +325,48 @@ a fixed behavior. It makes predictions based on any `rules` you have in your
...
+   other. The following snippet is an example for two contradicting rules:
```
akelad, line 351:
```suggestion
   The following snippet is an example for two contradicting rules:
```

docs/docs/policies.mdx
```diff
@@ -325,6 +325,48 @@ a fixed behavior. It makes predictions based on any `rules` you have in your
...
+      - action: utter_chitchat
+
+    - rule: Greet instead of chitchat
+      steps:
+      - intent: chitchat
+      - action: utter_greet  # `utter_greet` contradicts `utter_chitchat` from the rule above
+    ```
+ * `restrict_rules` (default: `true`): Rules are restricted to one user turn, but
+    there can be multiple bot events, including e.g. a form being filled and its
+    subsequent submission. It is **not recommended** to change this parameter to
```
akelad, line 367:
```suggestion
    Change this parameter to `false` at your own risk.
```

docs/docs/policies.mdx
```diff
@@ -325,6 +325,48 @@ a fixed behavior. It makes predictions based on any `rules` you have in your
...
+    `false`. Overusing rules for purposes out of the [recommended use cases](rules.mdx)
```
akelad, line 368:
```suggestion
    Overusing rules for purposes outside of the [recommended use cases](rules.mdx)
```

akelad (approved):
few comments on the documentation, but should be good to go otherwise

rasa/core/policies/rule_policy.py
akelad: maybe or stories?

rasa/core/policies/rule_policy.py
akelad: lol ok that's weird

wochinge: We find them by going through the training tracker and predicting each predictable state. If the `RulePolicy` predicts something else, it's contradicting.

rasa/core/policies/rule_policy.py
wochinge: it used `or`

PR #6611, opened 2020-09-09
Move actions.py -> actions/actions.py in rasa init

### LINKED ISSUE #5368: Inconsistency in initial project setup for actions running via rasa or via sdk
Currently, when running an action server with `rasa run actions`, the default looks in `actions.py` in the base of your project directory. `actions.py` is created there as part of the initial project. 

However, if you build in docker, the default place to find actions is in `actions/actions.py`. https://github.com/RasaHQ/rasa-sdk/blob/master/Dockerfile#L40

This makes it confusing -- when teaching the user to create custom actions in Docker, we have to tell them to create the new folder with an actions file, when `actions.py` already exists: https://rasa.com/docs/rasa/user-guide/how-to-deploy/#creating-a-custom-action

This seems to also have caused confusion here: https://github.com/RasaHQ/rasa-sdk/pull/140#issuecomment-586936047

We should make the default behavior consistent. I'm partial to having an `actions` folder as it's probably quite normal for a larger actions file to reference other files. However that will probably involve changing a decent amount of references in e.g. tutorials.

ricwo:
**Proposed changes**:
- fixes #5368 


docs/docs/docker/building-in-docker.mdx
```diff
@@ -166,17 +166,7 @@ To create more sophisticated assistants, you will want to use [Custom Actions](.
...
-
-```bash
-mkdir actions
-mv actions.py actions/actions.py
-# Rasa SDK expects a python module.
-# Therefore, make sure that you have this file in the directory.
-touch actions/__init__.py
-```
-
-Then build a custom action using the Rasa SDK by editing `actions/actions.py`, for example:
```
erohmensing, line 179: Ugh, beautiful 🙌

docs/docs/how-to-deploy.mdx
```diff
@@ -94,16 +94,8 @@ that are part of your action code but not included in the base `rasa/rasa-sdk` i
...
-
-   ```bash
-   mkdir actions
-   mv actions.py actions/actions.py
-   touch actions/__init__.py  # the init file indicates actions.py is a python module
-   ```
-
-   The `rasa/rasa-sdk` image will automatically look for the actions in `actions/actions.py`.
+1. Make sure your actions are defined in `actions/actions.py`. The `rasa/rasa-sdk`
+  image will automatically look for the actions in this file.
```
erohmensing, line 98: Just a thought, can you provide multiple actions files? If so it probably makes sense to update rasa sdk to pass the directory instead of that specific filename

erohmensing (approved):
Long overdue (by my fault!) -- thanks for moving forward with this!

docs/docs/how-to-deploy.mdx
ricwo: but we're using `DEFAULT_ACTIONS_PATH` which is just `actions`, so Rasa SDK will by default load the whole `actions` package

erohmensing: sorry, should have clarified, i meant in the dockerfile: https://github.com/RasaHQ/rasa-sdk/blob/master/Dockerfile#L40

ricwo: aah got it, good point! we should change that

Issue #6503, opened 2020-08-28
Model regression tests fail because of a dependency issue


dakshvar22:
**Rasa version**: [PR](https://github.com/RasaHQ/rasa/pull/6466)

**Operating system** (windows, osx, ...): Ubuntu

**Issue**: Model regression tests for some configurations start failing because of a [dependency issue](https://github.com/RasaHQ/rasa/runs/1040276701?check_suite_focus=true). This is weird because it's not consistent across configurations of the regression tests which means it has nothing to do with the code and the configurations and is a bit random probably.

m-vdb:
I've introduced the new `pydoc-markdown` dependency, and I didn't take a look at the setup for model regression tests. My bad. Let me take a quick look to see if I can fix it

m-vdb:
it actually seems to me that it's not linked to `pydoc-markdown` (sure, there is an internal error from pip, but that's not stopping the script). The real error is:
```bash
[TypeError]
expected string or bytes-like object
```

Which happens at various stages of the installation process on various builds. There are [this issue](https://github.com/python-poetry/poetry/issues/2027) and [this one](https://github.com/python-poetry/poetry/issues/1953) that talk about a possible problem with virtualenv.

dakshvar22:
Hmm, so would the recommendation be to try re-running the jobs in that case?

m-vdb:
maybe, but it doesn't seem like a suitable solution. Will retry the one you mentioned in the desc of this PR

m-vdb:
but apparently the "retry" didn't work @dakshvar22 . How can we retry those runs?

dakshvar22:
I just re-ran them by adding the appropriate labels on the PR. :)

m-vdb:
it looks like dependencies got installed [this time](https://github.com/RasaHQ/rasa/pull/6466/checks?check_run_id=1041994650) 🤷 😕

dakshvar22:
They do get installed for the [first configuration](https://github.com/RasaHQ/rasa/runs/1040276670?check_suite_focus=true) always and then the next one fails. 😕

m-vdb:
how different are the configurations in terms of OS?

dakshvar22:
They all run on the same OS configuration.

dakshvar22:
And it failed for the second  configuration :) - https://github.com/RasaHQ/rasa/pull/6466/checks?check_run_id=1041994673

m-vdb:
[...]

PR #6614, opened 2020-09-09
Use a more precise restore key for GH cache action

tczekajlo:
Use a more precise restore key for GH cache action - https://github.com/RasaHQ/rasa/issues/6503


tczekajlo:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]

```

.github/workflows/ci-model-regression.yml
```diff
@@ -388,7 +388,7 @@ jobs:
         with:
           path: ~/.cache/pypoetry/virtualenvs
           key: ${{ runner.os }}-poetry-3.8-${{ hashFiles('**/poetry.lock') }}
-          restore-keys: ${{ runner.os }}-poetry-3.8
+          restore-keys: ${{ runner.os }}-poetry-3.8-${{ hashFiles('**/poetry.lock') }}
```
m-vdb, line 391: [from the action's README](https://github.com/actions/cache#inputs):
> `restore-keys` - An ordered list of keys to use for restoring the cache if no cache hit occurred for key

So it sounds like we may want to remove the option altogether?

tczekajlo: No, we want it. The issue that we're turning to solve here is that cache is shared trough workflows, jobs, and so on. Currently, we use ${{ runner.os }}-poetry-3.8 which means that any cache that matches a restore key is used. It might happen that the cache will be overridden by different PR that why we want to be sure that we use cache related to our dependencies.

Does it make sense?

tczekajlo: Good point!

m-vdb (approved):
:octocat:

Issue #6602, opened 2020-09-08
Make metadata available in NLU components


degiz:
Part of #6475 

`2.0` Training data format introduced a new `metadata` field for the NLU training examples:

```yaml
nlu:
- intent: estimate_emissions
  # Arbitrary metadata
  metadata:
    author: Some example metadata!
    key: value
  # Multiline examples, each line is a separate training example.
  examples: |
     - how much CO2 will that use?
```

The content of metadata is not fixed and is up to the users to decide on. However, we should parse it and make it available from the custom NLU components user might have. One example - sentiments analyzer described [here](https://blog.rasa.com/enhancing-rasa-nlu-with-custom-components/). It uses a separate text file with sentiments, but now users should be able to store sentiments next to the training examples using the `metadata` field.

PR #6617, opened 2020-09-09
Support metadata parsing for intents and intent examples

m-vdb:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/6602
- API: `Message.data["metadata"]["intent"]` and `Message.data["metadata"]["example"]`


rasa/shared/nlu/training_data/message.py
```diff
@@ -116,6 +121,10 @@ def build(
                 data[INTENT_RESPONSE_KEY] = intent
         if entities:
             data[ENTITIES] = entities
+        if intent_metadata is not None:
+            data[METADATA] = {METADATA_INTENT: intent_metadata}
+        if metadata is not None:
+            data.setdefault(METADATA, {})[METADATA_EXAMPLE] = metadata
```
m-vdb, line 127: this will likely need to be documented. Where is the right place for this?

wochinge:
@alwx Do you have time for this one?

rasa/shared/nlu/training_data/message.py
```diff
@@ -93,7 +96,9 @@ def build(
         cls,
         text: Text,
         intent: Optional[Text] = None,
-        entities: List[Dict[Text, Any]] = None,
+        entities: Optional[List[Dict[Text, Any]]] = None,
+        intent_metadata: Optional[Any] = None,
+        metadata: Optional[Any] = None,
```
degiz, line 101: Can we update the docstring for this method?

rasa/shared/nlu/training_data/message.py
```diff
@@ -105,7 +110,7 @@ def build(
         Returns:
             Message
         """
-        data = {TEXT: text}
+        data: Dict[Text, Any] = {TEXT: text}
```
degiz, line 113: We don't usually define the types for local variables do we?

rasa/shared/nlu/training_data/message.py
degiz, line 101: also, maybe we can rename `metadata` parameter to `example_metadata` or something similar.

rasa/shared/nlu/training_data/message.py
degiz: I think we need to mention NLU `metadata` in this [doc](https://rasa.com/docs/rasa/next/training-data-format).
@akelad should we mention somewhere that `Message` class now has this metadata parsed?

degiz (approved):
Awesome and very clean 🚀

rasa/shared/nlu/training_data/message.py
akelad: agreed to mentioning it in the training data doc, perhaps we also add a quick explanation of how it can be used for custom components [here](https://rasa.com/docs/rasa/next/components/custom-nlu-components)? Or just link to the custom nlu component page from the training data format in that section.
About the `Message` class, I don't think we have that documented anywhere - but @m-vdb it's worth doing a quick search of the docs files locally to double check

rasa/shared/nlu/training_data/message.py
m-vdb: it's needed because the typechecker cannot infer the type correctly: this line creates a dict of type `Dict[Text, Text]` but we are adding other values that have types other than `Text`

rasa/shared/nlu/training_data/message.py
m-vdb: @akelad WDYT d1fa79e ? Any suggestions?

akelad: looks good to me

PR #6619, opened 2020-09-09
Implement telemetry reporting

### LINKED ISSUE #6613: Add Telemetry reporting
- [x]  add new CLI command
    - [x]  `rasa telemetry disable` to disable telemetry
    - [x]  `rasa telemetry enable` to enable telemetry
    - [x]  `rasa telemetry` to print information about the telemetry webpage

- [x]  implement telemetry reporting
    - [x]  implementation warnings in CLI
    - [x] docs message on install
    - [x] environment variable to enable a debug mode which prints the logged information to the commandline

- [x]  create a documentation page describing
    - [x]  why do we collect the information
    - [x]  how can the user opt out
    - [x]  what do we collect (not only high level, but in detail with examples)

tmbo:
**Proposed changes**:
- added telemetry reporting
- fixes #6613 


tmbo:
@akelad can you please take a look at the docs?

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,122 @@
+---
+id: telemetry
+sidebar_label: Rasa Telemetry
+title: Rasa Telemetry
+---
+
+Rasa uses telemetry to report anonymous usage information. This information is
+essential help improve Rasa Open Source for all users.
```
karen-white, line 8: missing 'to': essential to help improve

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,122 @@
...
+For the team working on Rasa Open Source it is important to understand how
+the product is used, it allows us to properly prioritize our research efforts
```
karen-white, line 10: Just a grammatical thing: comma splice. Should be 2 sentences:

For the team working on Rasa Open Source it is important to understand how
the product is used. It allows us to properly prioritize our research efforts

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,122 @@
...
+and feature development.
+
+You will be notified about the telemetry reporting when running Rasa Open Source
+for the first time.
+
+## How to opt-out
+
+You can always opt-out from telemetry reporting using
```
karen-white, line 18: You can always opt-out => You can opt out of telemetry reporting at any time using...

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,122 @@
...
+  This will enable us direct our research efforts towards projects
+  that have the most impact on text and dialogue handling.
+- We will be able to know data set sizes and general structure (e.g. the number
+  of intents). This allows us to better test our software on different types
+  of data sets and optimize the frameworks performance.
+- We will be able to get more detail on the types of errors you are running
+  into while building an assistant (e.g. initialization, training, etc.).
+  This will let us improve the quality of our framework and better focus our
+  time on solving more common, frustrating issues.
+
```
karen-white, line 49: I'd suggest moving the sensitive data section up higher, after the ## Why do we use telemetry reporting? section. 

Your first thought after you read the examples of data usage is whether Rasa can read your training data and user messages. Maybe even renaming the section ## What do we *not* report?

karen-white (approved):

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+async def track(
+    event_name: Text,
+    properties: Optional[Dict[Text, Any]] = None,
+    context: Optional[Dict[Text, Any]] = None,
+) -> None:
+    """Tracks a telemetry event.
+
+    It is OK to use this function from outside telemetry.py, but note that it
+    is recommended to create a new track_xyz() function for complex telemetry
+    events, or events that are generated from many parts of the Rasa X code.
```
federicotdn, line 389:
```suggestion
    events, or events that are generated from many parts of the Rasa OSS code.
```

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+        "project": _project_hash(),
+        "python": sys.version.split(" ")[0],
+        "rasa_open_source": rasa.__version__,
+        "gpu": len(tf.config.list_physical_devices("GPU")),
+        "cpu": multiprocessing.cpu_count(),
+    }
+
+
+@ensure_telemetry_enabled
+async def track(
```
federicotdn, line 380: The way `track` is implemented now, callers need to use `await track(...)`. This implies that, although other coroutines can run while the HTTP request is being sent, the caller of `track` will still need to wait for the request to end until they can continue with their code. For example, this happens on `await track_model_training()`.

My idea was to implement `track()` without the `async` keyword, but still have it use the event loop to call `_send_event`, but in an optional fire-and-forget fashion:
```python
def track(..., async_context: bool = True):
    coro = _send_event(...)
    
    if async_context:
        asyncio.ensure_future(coro)  # Coroutine will be added to the loop, no need to await its result
    else:
        asyncio.get_event_loop().run_until_complete(coro)  # Block until it's finished
```

So we would use `async_context=True` in contexts where there is already an event loop running, and we want to fire-and-forget the request. We would use `False` in contexts where there is no event loop running, like in `track_telemetry_disabled`.

The disadvantage of using fire-and-forget is that in some cases, the event loop could be closed before the coroutine is run. I can see this happening for example when we finish training the model and want to send an event right at the end.

All of this only applies if we want to be very conscious of potential delays by telemetry requests though. If we don't care about those then all of this isn't necessary and we can keep a simple `await`able interface.

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+    properties[TELEMETRY_ID] = telemetry_id
+
+    try:
+        await _send_event(
+            telemetry_id, event_name, properties, with_default_context_fields(context)
+        )
+    except Exception as e:  # skipcq:PYL-W0703
+        logger.debug(
+            f"An error occurred when trying to report the telemetry event: {e}"
+        )
```
federicotdn, line 415: This should be already handled by `ensure_telemetry_enabled`.

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+            "model_type": model_type,
+            # TODO: figure out how to track these
+            # "pipeline": config.get("pipeline"),
+            # "policies": config.get("policies"),
+            "num_intent_examples": len(nlu_data.intent_examples),
+            "num_entity_examples": len(nlu_data.entity_examples),
+            "num_actions": len(domain.action_names),
+            # Old nomenclature from when 'responses' were still called
+            # 'templates' in the domain
+            "num_templates": len(domain.templates),
```
federicotdn, line 485: Can't we use the new name here? Given that telemetry is new for OSS.

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+
+@ensure_telemetry_enabled
+def track_telemetry_disabled() -> None:
+    """Track when a user disables telemetry."""
+
+    loop = asyncio.new_event_loop()
+    try:
+        loop.run_until_complete(track(TELEMETRY_DISABLED))
+    finally:
+        loop.close()
```
federicotdn, line 506: See my comment in `track()`.

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+    "HUDSON_URL",
+    "JENKINS_URL",
+    "TEAMCITY_VERSION",
+    "TRAVIS",
+]
+
+# If updating or creating a new event, remember to update
+# https://rasa.com/docs/rasa/telemetry
+TRAINING_STARTED_EVENT = "Training Started"
+TRAINING_COMPLETED_EVENT = "Training Completed"
```
federicotdn, line 62: I ran `rasa init`, but never got a `Training Completed` event :thinking:

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+    "CONTINUOUS_INTEGRATION",
+    "GITHUB_ACTIONS",
+    "HUDSON_URL",
+    "JENKINS_URL",
+    "TEAMCITY_VERSION",
+    "TRAVIS",
+]
+
+# If updating or creating a new event, remember to update
+# https://rasa.com/docs/rasa/telemetry
```
federicotdn, line 60: WDYT of doing an approach using a JSON file (`events.json`) like we do for Rasa X?

federicotdn (changes requested):
Tested this locally, and it is working nicely. I've added a lot of small/medium comments and a possible discussion point in the implementation of `telemetry.track()`.

rasa/telemetry.py
tmbo: :D not implemented yet ;) will add it in a next PR, wanted to keep this as small as possible. it shouldn't be in here but there is also no reason to remove it now, will add that event in a second.

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+    # calling this even if the environment variable is set makes sure the
+    # configuration is created and there is a telemetry ID
+    is_enabled_in_configuration = _is_telemetry_enabled_in_configuration()
+
+    telemetry_environ = os.environ.get(TELEMETRY_ENVIRONMENT_VARIABLE)
+
+    if telemetry_environ is None:
+        return is_enabled_in_configuration
+    else:
+        return telemetry_environ.lower() == "true"
```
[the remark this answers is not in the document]
tmbo: This doesn't result in the same outcome: we always want the env to overwrite the configuration value. in the changed code an environment variable set to `False` will not overwrite the value stored in `is_enabled_in_configuration `

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+from subprocess import CalledProcessError, STDOUT, check_output  # skipcq:BAN-B404
+import sys
+import textwrap
+from typing import Any, Dict, Optional, Text
+import uuid
+
+import aiohttp
+from terminaltables import SingleTable
+
+import rasa
```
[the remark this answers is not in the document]
tmbo: mhm what is the benefit of an alias here?

rasa/telemetry.py
tmbo: I'd say yml is easier to write, but yes seems like a good idea.

rasa/telemetry.py
```diff
@@ -0,0 +1,506 @@
...
+        print("Telemetry Event:")
+        print(json.dumps(payload, indent=2))
+        return
+
+    write_key = telemetry_write_key()
+    if not write_key:
+        # If TELEMETRY_WRITE_KEY is empty or `None`, telemetry has not been
+        # enabled for this build (e.g. because it is running from source)
+        logger.debug("Skipping request to external service: telemetry key not set.")
+        return
```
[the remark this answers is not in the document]
tmbo: done :+1:

rasa/telemetry.py
tmbo: I wanted to keep the two events the same to have consistent naming in our telemetry db. I mean otherwise we are going to have different names in the db schema which isn't great either.

rasa/telemetry.py
tmbo: yes that is a good point, I changed it to run in the background in a slightly different way, using `enusre_future` in the event specific telemetry functions

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+**Anonymous** telemetry data allow us to prioritize our research efforts
+and feature development based on usage. We want to collect aggregated
+information on usage and reliability so that we can ensure a high-quality product.
+
+So how will we use the reported telemetry data? Here are some examples
+what we use the data for:
+
+- We will be able to know which languages, pipelines and policies are used.
+  This will enable us direct our research efforts towards projects
+  that have the most impact on text and dialogue handling.
```
akelad, line 41: i'm not sure this sentence makes sense - do we mean something like "This will enable us to direct our research efforts towards text and dialogue handling projects that will have the biggest impact for our users"?

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+  of intents). This allows us to better test our software on different types
+  of data sets and optimize the frameworks performance.
+- We will be able to get more detail on the types of errors you are running
+  into while building an assistant (e.g. initialization, training, etc.).
+  This will let us improve the quality of our framework and better focus our
+  time on solving more common, frustrating issues.
+
+## What about sensitive data?
+
+Your sensitive data never leaves your machine. We
```
akelad, line 52:
```suggestion
Your sensitive data never leaves your machine. We:
```

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+- don't report any personal identifiable information
+- don't report your training data
+- don't report any messages your assistant receives or sends
```
akelad, line 55: can we maybe bold the `don't` in all of these?

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+:::note Inspect what is reported
+You can view all the telemetry information that is reported
+by setting the environment variable `RASA_TELEMETRY_DEBUG=true`, e.g.
```
akelad, line 59:
```suggestion
by defining the environment variable `RASA_TELEMETRY_DEBUG=true`, for example when running the train command:
```

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+```bash
+RASA_TELEMETRY_DEBUG=true rasa train
+```
+When you set `RASA_TELEMETRY_DEBUG` no information will be send to any server,
```
akelad, line 63:
```suggestion
When you set `RASA_TELEMETRY_DEBUG` no information will be sent to any server,
```

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+Rasa reports aggregated usage details, command invocations, performance
+measurements and errors.
+We use the telemetry data to better understand usage patterns. The reported data
+will directly allow us to better decide how to design future features
+and prioritize current work.
+
+Specifically, we collect the following information for all telemetry events:
+
+- Type of the reported event (e.g. `Training Started`)
```
akelad, line 77: this isn't code, so it shouldn't be formatted with ` - maybe italics instead?

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+- Rasa machine ID: This is generated with UUID and stored in the global Rasa
```
akelad, line 78:
```suggestion
- Rasa machine ID: This is generated with a UUID and stored in the global Rasa
```

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+  config at `~/.config/rasa/global.yml` and sent as `metrics_id`
+- One-way hash of the current working directory or a hash of the git remote (`project`)
```
akelad, line 80: what does (`project`) mean?

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+    "project": "a0a7178e6e5f9e6484c5cfa3ea4497ffc0c96d0ad3f3ad8e9399a1edd88e3cf4",
+    "python": "3.7.5",
+    "rasa_open_source": "2.0.0",
+    "gpu": 0,
+    "cpu": 16
+  }
+}
+```
+
+We **cannot identify individual users** from the dataset. It is anonymized and
```
akelad, line 122: can you add "anonymized" to the typo CI exclusion list?

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
+---
+id: telemetry
+sidebar_label: Rasa Telemetry
+title: Rasa Telemetry
+---
+
+Rasa uses telemetry to report anonymous usage information. This information is
```
akelad, line 7: can you put this paragraph or at least part of it under `abstract:` below the title above? If you try doing `make livedocs` you'll see how that looks

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+For the team working on Rasa Open Source it is important to understand how
+the product is used. It allows us to properly prioritize our research efforts
+and feature development.
+
+You will be notified about the telemetry reporting when running Rasa Open Source
+for the first time.
+
+## How to opt-out
+
+You can opt out of telemetry reporting at any time using
```
akelad, line 18:
```suggestion
You can opt out of telemetry reporting at any time by running the command:
```

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+```bash
+rasa telemetry disable
+```
+
+or by setting `RASA_TELEMETRY_ENABLED=false` as an environment variable.
+
+If you want to enable reporting again, you can use
```
akelad, line 25:
```suggestion
If you want to enable reporting again, you can run:
```

docs/docs/telemetry.mdx
```diff
@@ -0,0 +1,123 @@
...
+- don't report any messages your assistant receives or sends
+
+:::note Inspect what is reported
+You can view all the telemetry information that is reported
+by setting the environment variable `RASA_TELEMETRY_DEBUG=true`, e.g.
+```bash
+RASA_TELEMETRY_DEBUG=true rasa train
+```
+When you set `RASA_TELEMETRY_DEBUG` no information will be send to any server,
+instead it will be logged to the commandline as a json dump for you to inspect.
```
akelad, line 64:
```suggestion
instead it will be logged to the commandline as a json dump for you to inspect.
:::
```
you need to finish a note with `:::` now

akelad (commented):
few changes to be made on the docs - good to double check everything renders correctly with `make livedocs` (note: lists don't render correctly atm, Lu is working on fixing that though)

docs/docs/telemetry.mdx
tmbo: done :+1:

tmbo:
thanks a lot @akelad for the comments :rocket:

tmbo:
@federicotdn ready for another review 🙂 

I've changed the reporting function to make sure it runs in the background :+1:

rasa/telemetry.py
```diff
@@ -0,0 +1,554 @@
...
+
+    headers = segment_request_header(write_key)
+
+    async with aiohttp.ClientSession() as session:
+        async with session.post(
```
wochinge, line 323: does this also catch these when there is no network connection?

tmbo: in case of no connection, no reporting will be done. the exception will be logged to debug

rasa/cli/utils.py
```diff
@@ -203,3 +205,38 @@ def payload_from_button_question(button_question: "Question") -> Text:
 def signal_handler(sig, frame) -> NoReturn:
     print("Goodbye 👋")
     sys.exit(0)
+
+
+def run_in_loop(f: Awaitable[T], loop: Optional[asyncio.AbstractEventLoop] = None) -> T:
```
wochinge, line 210: why do we need a Generic here? We know that it's gonna be a `Callable`, don't we?

wochinge, line 210: optional: this function is used from a couple of functions outside the `cli` package. This makes me wonder if `rasa.cli.utils` is really the best place for this to live.

rasa/train.py
```diff
@@ -118,8 +118,9 @@ async def handle_domain_if_not_exists(
         file_importer, output=output_path, fixed_model_name=fixed_model_name
     )
     print_warning(
-        "Core training was skipped because no valid domain file was found. Only an nlu-model was created."
-        "Please specify a valid domain using '--domain' argument or check if the provided domain file exists."
+        "Core training was skipped because no valid domain file was found. "
+        "Only an nlu-model was created. Please specify a valid domain using "
```
wochinge, line 122:
```suggestion
        "Only an NLU model was created. Please specify a valid domain using "
```

rasa/train.py
```diff
@@ -118,8 +118,9 @@ async def handle_domain_if_not_exists(
...
+        "'--domain' argument or check if the provided domain file exists."
```
wochinge, line 123:
```suggestion
        "the '--domain' argument or check if the provided domain file exists."
```

rasa/train.py
```diff
@@ -396,14 +399,15 @@ async def _train_core_with_validated_data(
...
         )
-        await rasa.core.train(
-            domain_file=domain,
-            training_resource=file_importer,
-            output_path=os.path.join(_train_path, DEFAULT_CORE_SUBDIRECTORY_NAME),
-            policy_config=config,
-            additional_arguments=additional_arguments,
-            interpreter=interpreter,
-        )
+        async with telemetry.track_model_training(file_importer, model_type="core"):
```
wochinge, line 402: when training a model, this will track 3 times: 
1. the entire training
2. the core training
3. the NLU training

rasa/utils/common.py
```diff
@@ -37,11 +36,15 @@ def __exit__(
             shutil.rmtree(self)
 
 
-def read_global_config() -> Dict[Text, Any]:
-    """Read global Rasa configuration."""
+def read_global_config(path: Text) -> Dict[Text, Any]:
+    """Read global Rasa configuration.
+
+    Args:
+        path: Path to the configuration
```
wochinge, line 43: please also describe the return value

rasa/utils/common.py
```diff
@@ -200,10 +208,14 @@ def not_found():
         else:
             raise ValueError(f"Configuration '{name}' key not found.")
 
-    if not os.path.exists(GLOBAL_USER_CONFIG_PATH):
+    # need to use `rasa.constants.GLOBAL_USER_CONFIG_PATH` to allow patching
+    # in tests
+    config_path = rasa.constants.GLOBAL_USER_CONFIG_PATH
+
+    if not os.path.exists(config_path):
```
wochinge, line 215: Can't we skip the the `os.path.exist` check and check whether we got an empty config or not?

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+TRAINING_STARTED_EVENT = "Training Started"
+TRAINING_COMPLETED_EVENT = "Training Completed"
+TELEMETRY_DISABLED = "Telemetry Disabled"
+
+
+def print_telemetry_reporting_info() -> None:
+    """Print telemetry information to std out."""
+    message = textwrap.dedent(
+        f"""
+      Rasa reports anonymous usage telemetry to help improve Rasa Open Source
```
wochinge, line 76: ?
```suggestion
      Rasa Open Source reports anonymous usage telemetry to help improve Rasa Open Source
```

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+# If updating or creating a new event, remember to update
+# https://rasa.com/docs/rasa/telemetry
+TRAINING_STARTED_EVENT = "Training Started"
+TRAINING_COMPLETED_EVENT = "Training Completed"
+TELEMETRY_DISABLED = "Telemetry Disabled"
+
+
+def print_telemetry_reporting_info() -> None:
+    """Print telemetry information to std out."""
+    message = textwrap.dedent(
```
wochinge, line 74: do we really need `textwrap`? You should be able to remove the indent to the left of the message without `black` complaining

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+
+def initialize_telemetry() -> bool:
+    """Read telemetry configuration from the user's Rasa config file in $HOME.
+
+    Creates a default configuration if no configuration exists.
+
+    Returns:
+        `True`, if telemetry is enabled, `False` otherwise.
+    """
+
```
wochinge, line 136:
```suggestion
```

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+        return stored_config[CONFIG_TELEMETRY_ENABLED]
+    except ValueError as e:
+        logger.debug(f"Could not read telemetry settings from configuration file: {e}")
+
+        # seems like there is no config, we'll create on and enable telemetry
+        _write_default_telemetry_configuration()
+        return TELEMETRY_ENABLED_BY_DEFAULT
+
+
+def initialize_telemetry() -> bool:
```
wochinge, line 128:
```suggestion
def _initialize_telemetry() -> bool:
```

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+
+    if telemetry_environ is None:
+        return is_enabled_in_configuration
+    else:
+        return telemetry_environ.lower() == "true"
+
+
+def ensure_telemetry_enabled(f: Callable[..., Any]) -> Callable[..., Any]:
+    """Function decorator for telemetry functions that only runs the decorated
+    function if telemetry is enabled."""
```
wochinge, line 151: let's use the google docstring format

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+
+    is_telemetry_enabled = initialize_telemetry()
+
+    # allows us to use the decorator for async and non async functions
+    if asyncio.iscoroutinefunction(f):
```
wochinge, line 156: we could also use `asyncio.coroutine` to wrap any non async functions which in turn would mean that we could deduplicate the block below. Or alternatively moving the `asyncio.iscoroutinefunction` within the decorator

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+
+    # noinspection PyBroadException
+    try:
+        with open(write_key_path) as f:
+            return f.read().strip()
+    except Exception:  # skipcq:PYL-W0703
+        return None
+
+
+def encode_base64(original: Text, encoding: Text = "utf-8") -> Text:
```
wochinge, line 211:
```suggestion
def _encode_base64(original: Text, encoding: Text = "utf-8") -> Text:
```

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+
+    async with aiohttp.ClientSession() as session:
+        async with session.post(
+            SEGMENT_ENDPOINT, headers=headers, json=payload,
+        ) as resp:
+            # handle different failure cases
+            if resp.status != 200:
+                response_text = await resp.text()
+                logger.debug(
+                    f"Segment telemetry request returned a {resp.status} response."
```
wochinge, line 330:
```suggestion
                    f"Segment telemetry request returned a {resp.status} response. "
```

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+                    f"Body: {response_text}"
+                )
+            else:
+                data = await resp.json()
+                if not data.get("success"):
+                    logger.debug(
+                        f"Segment telemetry request returned a failure."
```
wochinge, line 337:
```suggestion
                        f"Segment telemetry request returned a failure. "
```

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+
+def _project_hash() -> Text:
+    """Create a hash for the project in the current working directory.
+
+    Returns:
+        project hash
+    """
+    try:
+        remote = check_output(  # skipcq:BAN-B607,BAN-B603
+            ["git", "remote", "get-url", "origin"], stderr=STDOUT
```
wochinge, line 350: Why do we want to capture the error output? This should fail silently, shouldn't it?

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+    Args:
+        training_data: Training data used for the training.
+        model_type: Specifies the type of training, should be either "rasa", "core"
+            or "nlu".
+    """
+
+    config = await training_data.get_config()
+    stories = await training_data.get_stories()
+    nlu_data = await training_data.get_nlu_data()
+    domain = await training_data.get_domain()
```
wochinge, line 503: it doesn't matter but should be `asyncio.gather` in case it does matter at one point

rasa/telemetry.py
```diff
@@ -0,0 +1,553 @@
...
+            else:
+                data = await resp.json()
+                if not data.get("success"):
+                    logger.debug(
+                        f"Segment telemetry request returned a failure."
+                        f"Response: {data}"
+                    )
+
+
+def _project_hash() -> Text:
```
wochinge, line 342: makes sense to use `lru_cache` for all of the properties within `_default_context_fields`, doesn't it?

tests/test_telemetry.py
```diff
@@ -0,0 +1,196 @@
...
+TELEMETRY_TEST_USER = uuid.uuid4().hex
+TELEMETRY_TEST_KEY = uuid.uuid4().hex
+TELEMETRY_EVENTS_JSON = "docs/docs/telemetry/events.json"
+
+
+@pytest.fixture(autouse=True)
+def patch_global_config_path(monkeypatch: MonkeyPatch, tmp_path: Path):
+    """Ensure we use a unique config path for each test to avoid tests influencing
+    each other."""
+    default_location = rasa.constants.GLOBAL_USER_CONFIG_PATH
```
wochinge, line 25: why are we not monkeypatching this?

tests/test_telemetry.py
```diff
@@ -0,0 +1,196 @@
...
+    # is read from a temp file instead of the default location
+    assert "/.config/rasa" not in rasa.constants.GLOBAL_USER_CONFIG_PATH
+
+
+def test_segment_request_header():
+    assert telemetry.segment_request_header(TELEMETRY_TEST_KEY) == {
+        "Content-Type": "application/json",
+        "Authorization": "Basic {}".format(
+            telemetry.encode_base64(TELEMETRY_TEST_KEY + ":")
+        ),
```
wochinge, line 82: I think doing it this way doesn't make a lot of sense. E.g. when we for for some reason break `encode_base64` then this test will not fail. It would be better to hardcode the expected request header

changelog/6613.improvement.md
```diff
@@ -0,0 +1,2 @@
+Added telemetry reporting. Reporting will be opt-out. More information can
```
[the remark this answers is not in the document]
tmbo: I think the telemetry page is very verbose and since this doesn't require any change in user behaviour (e.g. like a change to our ML models). In addition to that users installing it will be notified again with a message in their CLI

rasa/cli/telemetry.py
```diff
@@ -0,0 +1,73 @@
...
+    print_success(
+        "\nYou can find more information about telemetry reporting at "
+        "" + DOCS_URL_TELEMETRY
+    )
+
+
+def disable_telemetry(_: argparse.Namespace) -> None:
+    """Disable telemetry tracking."""
+    rasa.cli.utils.run_in_loop(telemetry.track_telemetry_disabled())
+    telemetry.toggle_telemetry_reporting(is_enabled=False)
```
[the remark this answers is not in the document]
tmbo: I don't think `argparse` should use any methods outside the CLI package, e.g. I don't think it is cleaner to add a function to the telemetry package that receives `argparse.Namespace` as an argument.

rasa/cli/utils.py
tmbo: typevar is needed to properly set the return type. (annotation is wrong though, the callable part is missing :+1:, added that)

tmbo: moved it to common

rasa/telemetry.py
tmbo: in my opinion it makes it harder to read if a block of text is not properly indented like the rest of the code

changelog/6613.improvement.md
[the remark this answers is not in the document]
wochinge: > The changelog is going to be pretty packaged, so I'd keep this one short

But should it really be _this_ changelog which is kept short then? Communities are usually very wary about these kind of changes and I'd like to avoid the impression that we're trying to slip anything in. Imo being fully transparent (and potentially too verbose) reduces risk.

rasa/telemetry.py
tmbo: used outside of this module so shouldn't be private

wochinge: Sorry 🙈  for some reason pycharm didn't show this usage 🤔

rasa/telemetry.py
tmbo: `asyncio.coroutine` is deprecated

rasa/telemetry.py
tmbo: I do want to test it though which means it should not be protected.

rasa/telemetry.py
tmbo: I don't see the reason for that, `_default_context_fields` will only be called once so there is no need to use a cache for the other functions

rasa/telemetry.py
wochinge: then let's use an `if ... else` in the decorator?

rasa/telemetry.py
wochinge: I wouldn't change production code for the sake of testing code. 
I'd either just call it anyway from the tests or test it using a public function which calls that.

rasa/telemetry.py
tmbo: in examples usually, it gets pipped there if you don't want errors to end up in your error stream. I changed it to pipe it to dev null.

rasa/telemetry.py
tmbo: nah, doesn't matter - this is easier to read 🙂

rasa/train.py
tmbo: yes, this is intentional

rasa/utils/common.py
tmbo: it is not the same, the file might exist but be empty. the logic was like that before, I don't think we should change it

rasa/train.py
wochinge: What's the intention behind it?

tests/test_telemetry.py
tmbo: didn't work, but I didn't bother to investigate why

rasa/telemetry.py
wochinge: But what if a community member implements an importer that for example retrieves training data from a server? I don't think readability is such a big thing if you unpack it immediately after calling `gather(...)`

```
config, stories, nlu_data, domain = asyncio.gather(...)
```

tests/test_telemetry.py
wochinge: mhm, ok. then let's at least add the missing type annotation and drop the unused `monkeypatch` param

tests/test_telemetry.py
```diff
@@ -0,0 +1,196 @@
...
+    assert call_args[1] == "foobar"
+    assert call_args[2]["foo"] == "bar"
+    assert call_args[2]["metrics_id"] == telemetry.get_telemetry_id()
+    assert call_args[3]["baz"] == "foo"
+
+
+def test_toggle_telemetry_reporting(monkeypatch: MonkeyPatch):
+    # tests that toggling works if there is no config
+    telemetry.toggle_telemetry_reporting(True)
+    assert telemetry.initialize_telemetry() is True
```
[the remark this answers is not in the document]
tmbo: should we really? this is a private function so right now it is an implementation detail. so we would need to make it a public function which I am not sure makes sense. the API is tested through `test_events_schema` which ensures that the events properly make it through

tests/test_telemetry.py
tmbo: did that already :+1:

changelog/6613.improvement.md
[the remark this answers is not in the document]
tmbo: I don't think this is an issue, we have two documentation pages about this, including all the tracked events as well as a piece of information once you run this the first time. I really don't think users will feel that we are slipping this in. But I am also not sure debating this makes sense so I'll add two more sentences.

rasa/train.py
tmbo: we get runtimes for the separate parts and for the whole training (and e.g. model sizes if we want to add that)

wochinge: Makes sense 👍

rasa/telemetry.py
tmbo: it isn't specifically tested anyways now (which makes sense as it shouldn't be a public function) so made it protected.

tests/test_telemetry.py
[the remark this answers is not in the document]
tmbo: I'll add a test that calls `track(...)` and makes sure the segment API is called, but I don't think `_send_event` should be tested directly.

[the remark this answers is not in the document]
wochinge: I think we should discuss testing of protected / private in Slack (not as part of this PR). I had another look on this and as we have the `try ... catch` block around this in `ensure_telemetry_enabled` it probably doesn't matter too much as it won't do any harm for the user.

rasa/nlu/run.py
```diff
@@ -4,10 +4,12 @@
 from typing import Optional, Text
 
 from rasa.shared.utils.cli import print_success
+import rasa.cli.utils
```
wochinge, line 7: unused now as you moved the `run_in_loop`
```suggestion
```

tests/test_telemetry.py
```diff
@@ -0,0 +1,194 @@
...
+from rasa.shared.importers.importer import TrainingDataImporter
+from tests.conftest import DEFAULT_CONFIG_PATH
+
+TELEMETRY_TEST_USER = "083642a3e448423ca652134f00e7fc76"  # just some random static id
+TELEMETRY_TEST_KEY = "5640e893c1324090bff26f655456caf3"  # just some random static id
+TELEMETRY_EVENTS_JSON = "docs/docs/telemetry/events.json"
+
+
+@pytest.fixture(autouse=True)
+def patch_global_config_path(tmp_path: Path) -> None:
```
wochinge, line 22:
```suggestion
def patch_global_config_path(tmp_path: Path) -> Generator[None, None, None]:
```

rasa/telemetry.py
wochinge: @tmbo

wochinge (approved):
Looks great 👍

rasa/telemetry.py
tmbo: I am not quite sure what you mean 🤔 `ensure_telemetry_enabled ` is the decorator. do you mean inside of `decorated`? I don't think that would work since that function needs to be either `async def` or `def`

wochinge: Never mind :-) Played a bit around with it, and it's surprisingly complicated / impossible to deduplicate it further for this particular use case.

Issue #6616, opened 2020-09-09
Do not interpolate button payload


tabergma:
**Rasa version**: Rasa 2.0.0a4

**Issue**: If you define buttons with some payload like the following in your domain file, Rasa tries to interpolate the button payload when you are interacting with the assistant. It interprets the part between `{}` as placeholder and tries to fill it. However, in this case it refers to some entities and there is not matching template.

 utter_ask_nearby_pois:

*   text: Are you looking for an apartment near specific surroundings?
     buttons:
*   title: "no"
     payload: '/inform {"nearby_pois": "none"} 

    '

*   title: "school"
     payload: '/inform {"nearby_pois": "school"} 

    '

*   title: "train station"
*   title: "office"
     payload: '/inform {"nearby_pois": "office"} 

    '

The bot works as expected it just prints an ERROR to the command line. So, the intent and entities are recognized correctly and the bot behaves as expected.

**Idea:** @erohmensing suggested the following:

> honestly i think it might be enough, just to not print the warning if the segment starts with `/`. it will still perform fine as you’ve shown, and if they actually do have a weird slot somehow named `"nearby_pois": "office"`, it will still fill it.

**Error (including full traceback)**:

 2020-09-09 10:39:07 ERROR rasa.core.nlg.interpolator - Failed to fill utterance template '/inform

 {"nearby_pois": "office"} 

'. Tried to replace '"nearby_pois": "office"' but could not find a value for it. There is no slot with this name nor did you pass the value explicitly when calling the template. Return template without filling the template.
 Traceback (most recent call last):
 File "/Users/tabergma/Repositories/rasa/rasa/core/nlg/interpolator.py", line 28, in interpolate_text
 text = text.format(values)
 KeyError: '"nearby_pois": "office"'

**Command or request that led to error**:

 rasa shell

wochinge:
@erohmensing Didn't you fix this? Can I move this to CSE board

PR #6632, opened 2020-09-10
ignore interpolation warning if value starts with intent prefix

erohmensing:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6616

### DIFF
--- a/rasa/core/nlg/interpolator.py
+++ b/rasa/core/nlg/interpolator.py
@@ -2,6 +2,8 @@ import re
 import logging
 from typing import Text, Dict, Union, Any
 
+from rasa.core.constants import INTENT_MESSAGE_PREFIX
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


rasa/core/nlg/interpolator.py
```diff
@@ -36,14 +38,18 @@ def interpolate_text(template: Text, values: Dict[Text, Text]) -> Text:
...
     except KeyError as e:
-        logger.exception(
-            f"Failed to fill utterance template '{template}'. "
-            f"Tried to replace '{e.args[0]}' but could not find "
-            f"a value for it. There is no slot with this "
-            f"name nor did you pass the value explicitly "
-            f"when calling the template. Return template "
-            f"without filling the template. "
-        )
+        # we assume that if the value of the {key: value} pair starts with `/`,
```
wochinge, line 41: why do we interpolate it then in the first place?

wochinge (commented):
🚀 A test would be great please 🙏  And a changelog!

rasa/core/nlg/interpolator.py
erohmensing: Because we interpolate anything that has the curly braces in it! It's just an unfortunate thing about the regex format versus the slot interpolation format.  

We could try to make a nicer regex for this, but as far as I could see in the rest of the code, we don't have a regex for the /intent{"entity": "value"} format, we just pass things that that with `/` to the regexinterpreter. Though I guess that interpreter does have a regex for processing 🤔 

Basically one thing we could have is a button payload that is e.g. `/intent{"entity": {slot_name}}` and we want it to correctly fill to `/intent{"entity": "value"}` without yelling that it can't find a slot called `"entity": {slot_name}`. Any suggestions for improving that?

wochinge: Thanks for the explanation! Can't we restrict the interpolation to anything which **does not** start with `/`?

wochinge:
@erohmensing How is it going with this PR 😂 ?

ortsaCniveK:
Any update on this PR?  Let me know if you need some help pitching in to get this pushed up, would like to clean up my logs, since I use button payloads with slash commands a lot  😅

rasa/core/nlg/interpolator.py
raoulvm: That's what I thought first, too. But @erohmensing has a point - there could be use cases to set an entity/slot to the value of a different slot, maybe while confirming data from a backend inquiry or similar. So hiding the exception seem to be the way to go for me.

raoulvm: According to doc you should use double curly braces (standard f-string escaping for curly braces) if you want rasa NOT TO replace (interpolate) the alleged key words. You can still use curly braces inside, like

`/intent{{"entity": "{my_slot}"}}`

even though it gets harder to read. I would recommend putting that in the docs and close the PR.

wochinge: @erohmensing What do you think?

PR #6650, opened 2020-09-11
chore: update log level to exception in processor's call to action-server

moaazsidat:
Updating the log level to exception to log the exception info when processor is calling the action server. 
Reason: most production systems use a log level that's not `debug`, and this seems to be a fairly important exception to be able to dig through. 


rasa/core/processor.py
```diff
@@ -676,7 +676,7 @@ async def _run_action(
                 "Please check the logs of your action server for "
                 "more information."
             )
-            logger.debug(e, exc_info=True)
+            logger.error(str(e), exc_info=True)
```
edouardlp, line 679: Why not just add `exc_info=True` to the existing error log above?

moaazsidat: No specific reason, was mostly adhering to the pattern [elsewhere in the Rasa codebase](https://github.com/RasaHQ/rasa/blob/master/rasa/core/channels/slack.py#L343-L345)

sara-tagger:

rasa/core/processor.py
```diff
@@ -670,13 +670,12 @@ async def _run_action(
             tracker.update(events[0])
             return self.should_predict_another_action(action.name())
         except Exception as e:
-            logger.error(
-                f"Encountered an exception while running action '{action.name()}'. "
+            logger.exception(
+                f"Encountered an exception while running action '{action.name()}'. {e}"
```
hadrien, line 674:
```suggestion
            logger.exception(
                f"Encountered an exception while running action '{action.name()}'."
```

Because it is using `logger.exception`, it does not need the exception instance in the log message.

moaazsidat:
@tttthomasssss 👋  let me know if there's something else I can add here, relatively simple fix that's been helping us debug issues that may not directly be evident from action-server logs, e.g. issues that may arise from some configurations/timeouts set on a service mesh level as part of Kubernetes-based deployment.

tmbo (approved):
looks good and increasing the log level is very reasonable :+1:

PR #6656, opened 2020-09-13
Automated error reporting

tmbo:
**Proposed changes**:
- Automated error reporting


tmbo:
do you have time to take a look @wochinge ?

wochinge:
Yes 👍

.github/workflows/continous-integration.yml
```diff
@@ -18,6 +18,11 @@ on:
 #                       repo on hub.docker.com. used to pull and upload containers
 # - RASA_OSS_TELEMETRY_WRITE_KEY: key to write to segment. used to report telemetry
 #                                 they key will be added to the distributions
+# - RASA_OSS_EXCEPTION_WRITE_KEY: key to write to sentry. used to report exceptions
+#                                 they key will be added to the distributions
+#                                 key can be found at https://sentry.io/settings/rasahq/projects/rasa-open-source/install/python/
```
wochinge, line 23:
```suggestion
# - RASA_OSS_TELEMETRY_WRITE_KEY: key to write to segment. Used to report telemetry.
#                                 The key will be added to the distributions
# - RASA_OSS_EXCEPTION_WRITE_KEY: key to write to sentry. Used to report exceptions.
#                                 The key will be added to the distributions.
#                                 Key can be found at https://sentry.io/settings/rasahq/projects/rasa-open-source/install/python/
```

.github/workflows/continous-integration.yml
```diff
@@ -273,6 +284,15 @@ jobs:
...
         password: ${{ secrets.PYPI_TOKEN }}
 
+    - name: Notify Sentry about the release
+      env:
+        GITHUB_TAG: ${{ github.ref }}
+        SENTRY_ORG: rasahq
+        SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+      run: |
+        VERSION=${GITHUB_TAG/refs\/tags\//}
+        ./scripts/inform_sentry_about_release.sh
```
wochinge, line 294: optional: What's the benefits of using the script? I'd prefer to keep the CI workflows self-contained unless it de-duplicates things.

rasa/telemetry.py
```diff
@@ -195,15 +196,39 @@ def telemetry_write_key() -> Optional[Text]:
...
+    if os.environ.get(EXCEPTION_WRITE_KEY_ENVIRONMENT_VARIABLE):
+        # a write key set using the environment variable will always
+        # overwrite any key provided as part of the package (`keys` file)
+        return os.environ.get(EXCEPTION_WRITE_KEY_ENVIRONMENT_VARIABLE)
+
+    write_key_path = pkg_resources.resource_filename(name, "keys")
+
+    # noinspection PyBroadException
+    try:
+        with open(write_key_path) as f:
```
wochinge, line 230: let's deduplicate this part in a helper function

rasa/telemetry.py
```diff
@@ -484,6 +509,56 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+
+    # key for local testing can be found at
+    # https://sentry.io/settings/rasahq/projects/rasa-open-source/install/python/
+    # for local testing, set the key using `RASA_EXCEPTION_WRITE_KEY=key rasa <command>`
+    key = sentry_write_key()
+
+    if not key:
+        return
+
+    def strip_sensitive_data(event, hint):
```
wochinge, line 533: please add types

rasa/telemetry.py
```diff
@@ -484,6 +509,56 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+        f"https://{key}.ingest.sentry.io/2801673",
+        before_send=strip_sensitive_data,
+        integrations=[
+            ExcepthookIntegration(),
+            DedupeIntegration(),
+            AtexitIntegration(lambda _, __: None),
+        ],
+        send_default_pii=False,  # activate PII filter
+        server_name=get_telemetry_id() or "UNKNOWN",
+        in_app_include=["rasa"],  # only submit errors in this package
```
wochinge, line 554: I'm not sure about this one. E.g. if there is an error with the lock store then it would be great to capture the errors directly from the redis library.

rasa/telemetry.py
```diff
@@ -484,6 +509,56 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+        return event
+
+    # this is a very defensive configuration, avoiding as many integrations as
+    # possible. it also submits very little data (exception with error message
+    # and line numbers).
+    sentry_sdk.init(
+        f"https://{key}.ingest.sentry.io/2801673",
+        before_send=strip_sensitive_data,
+        integrations=[
+            ExcepthookIntegration(),
```
wochinge, line 548: optional: how about using https://docs.sentry.io/platforms/python/guides/logging/ + changing the sanic error logging to `logger.exception` so that we can also catch server errors?

rasa/telemetry.py
```diff
@@ -484,6 +509,56 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+    # for local testing, set the key using `RASA_EXCEPTION_WRITE_KEY=key rasa <command>`
+    key = sentry_write_key()
+
+    if not key:
+        return
+
+    def strip_sensitive_data(event, hint):
+        # removes any paths from stack traces (avoids e.g. sending
+        # a users home directory name if package is installed there)
+        for value in event.get("exception", {}).get("values", []):
```
wochinge, line 536: let's also add a test for this. The functionality is not huge, but I think the test would be interesting as it would clearly show how an event typically looks like and what its structure is like.

wochinge (approved):
Nice 🚀 That will be a nice help to find and fix bugs early 💯

.github/workflows/continous-integration.yml
tmbo: you're right, will move the commands here

rasa/telemetry.py
tmbo: I'd rather start with a narrow definition of this and then if we think this works for us extend this definition

tmbo: in general, I agree, we can be more open here but for the moment I think it is fine to start just with exceptions originating in our codebase.

rasa/telemetry.py
tmbo: I took a closer look at that and even though I think it is something we should do in the future it is to much effort to do it now: we can't just hook into the exception handling, as sanic uses exceptions to bubble up any errors (even expected ones) which would create to much noise. so this means this needs a bit more work to filter out any exceptions that are expected to reduce the logging noise. let's do that after the RC

wochinge: Can't we hook into https://github.com/RasaHQ/rasa/blob/f1055d18b46071312259057a23c80ce5506e158f/rasa/server.py#L90 and only do it for `5xx` status codes? Or do you think that's still too much noise?

tmbo: it might work, but I think it requires a bit more testing that I'd like to avoid just now given the current timeline. I think it's something that we can easily add after this is merged

PR #6661, opened 2020-09-14
refactor Rasa X local mode start

wochinge:
**Proposed changes**:
- `accept_terms_and_metrics` currently doesn't only check the terms and metrics but also initialize the telemetry in Rasa X. That's weird as we suddenly have to initialization points for Rasa X. This PR drops `accept_terms_and_metrics` from the Rasa Open Source code. Instead we will add this within `local.main(...)`. This way we have on designated entry point to Rasa X so that we can initialize all state properly. Also reduces the code coupling.
- added a `RasaXTermsError` so that `local.main(...)` can cleanly indicate that the terms weren't accepted.

[Discussion which started this](https://github.com/RasaHQ/rasa-x/pull/3555/files)

- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions) (I don't think it's needed as it's an internal change and Rasa X / Rasa Open Source compatibility breaks anyway with Rasa Open Source 2.0)


wochinge:
@tmbo Short doublecheck on this: I'm not removing the terms / metrics check, just moving it. There is no weird legal reason that we have to call the terms check from Rasa Open Source code, is there?

ricwo (approved):
💯

tmbo:
> @tmbo Short doublecheck on this: I'm not removing the terms / metrics check, just moving it. There is no weird legal reason that we have to call the terms check from Rasa Open Source code, is there?

I don't see any issues 👍

wochinge:
@alwx @ricwo @degiz This will cause failings tests when bumping Rasa X to the Rasa Open Source RC. I'm on vacation from Friday onwards, so I won't be around when we bump the version from the latest alpha to the RC. As I don't know who exactly is gonna bump the version, I'm tagging all of you so that you're aware of this and don't have to start debugging:

Changes which are required within `rasax.community.local`:
- call `check_license_and_metrics` within `local.main`
- raise a `RasaXTermsException` if the user didn't accept the terms
- in case [this PR](https://github.com/RasaHQ/rasa-x/pull/3167/) is already merged: move `global_state.initialize_global_state(NUMBER_OF_SANIC_WORKERS, is_local_mode=True)` from `check_license_and_metrics` to `local.main`

Thanks 🙌

PR #6663, opened 2020-09-14
New docs: theme updates II

lunelson:


docs/docusaurus.config.js
```diff
@@ -109,14 +101,15 @@ module.exports = {
         },
       ],
     },
-    footer: {
-      copyright: `Copyright © ${new Date().getFullYear()} Rasa Technologies GmbH`,
-    },
+    // footer: {
+    //   copyright: `Copyright © ${new Date().getFullYear()} Rasa Technologies GmbH`,
+    // },
```
erohmensing, line 106: Should we just remove this altogether if its being sourced by the theme?

docs/docusaurus.config.js
```diff
@@ -71,6 +53,11 @@ module.exports = {
   organizationName: 'RasaHQ',
   projectName: 'rasa',
   themeConfig: {
+    algolia: {
+      apiKey: '09ef5d111ebd9002df72575d516693af',
+      indexName: 'BH4D9OD16A',
+      // searchParameters: {}, // Optional (if provided by Algolia)
+    },
```
erohmensing, line 60: Are we merging the search while its still erroring out? When do we think that will be configured?

docs/docusaurus.config.js
```diff
@@ -5,63 +5,45 @@ const {
...
-  label: 'Legacy 1.x',
-  href: 'https://legacy-docs-v1.rasa.com',
-  target: '_self',
+const versionLabels = {
+  current:
+    isDev || isPreview
+      ? `Next (${isPreview ? 'deploy preview' : 'dev'})`
+      : existingVersions.length < 1
+      ? 'Current'
+      : 'Next',
```
erohmensing, line 35: Is this how we're switching it from current to next when there are multiple versions? If so, i'd prefer it it were `Master` (as i believe that's what it is going to be, right?)

docs/docusaurus.config.js
lunelson: I'm going to try to disable the form for now, and if that doesn't work I guess we'll leave it out. I had to insert some data here to get it to show up, because of the way Docusaurus validates its config internally

docs/docusaurus.config.js
lunelson: Yeah I might revert to this for the launch but kill it later

docs/docusaurus.config.js
lunelson: Yes, this is already fixed. I picked up Maxime's label "Master/Unreleased"

akelad:
looks good to me generally speaking, it builds fine etc :D 

few questions:

* is the lack of image here in the sidebar intentional?
image

* not sure the algolia search works properly, when I try to search it just takes me to a search of their default things instead of our docs:
image

* how come we moved the API reference here? there's an "APIs" section already 
image

lunelson:
@akelad to your comments:

1. the missing image is _intentional_ insofar as it's a placeholder, for a `product-logo` element which will be in the next patch.
2. the search UI you see is using DocSearch's own index as a demo, it's only enabled in development. In production the search button is disabled—see [staging](https://staging--rasahq-docsv2-rasa.netlify.app/docs/rasa/next/)
3. this is an interim solution: the wireframed idea was that API Spec pages would have their own links below the tree-menu, with an icon that gave some kind of indication that they were sort of a special case (they open in a page with no sidebar). Their specified in the `customFields` part of the config and I'm appending them to the tree menu as a final section, until I can get a new link element done for them that will sit on its own there

Issue #6659, opened 2020-09-14
NLU is re-trained when stories change


wochinge:
Whenever you change a story, the NLU model will also be re-trained. This is likely caused by [these lines](https://github.com/RasaHQ/rasa/blob/dcb9b5659ab7a2a87417a7b9af603a37ed30888f/rasa/shared/importers/importer.py#L488).

In addition we should de-duplicate the messages which are added to the NLU training data (sorting would also be great to avoid having diffs in the training data just because a new story was added)

PR #6605, opened 2020-09-08
Re add e2e reading

wochinge:
**Proposed changes**:
- last part of the revert of https://github.com/RasaHQ/rasa/pull/6518
- add support for reading end-to-end stories for **training** Rasa Open Source

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6674, opened 2020-09-15
sort and deduplicate story events before adding them to NLU training data

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6659
- A change of the story order or the order within the steps should not affect the NLU training data. If this constraint is not respected, then the NLU model will be retrained whenever the story or event order changed. To guarantee the stableness of the NLU data, the events are deduplicated and ordered before generating the `Message` objects from them. Note: New actions / utterances still need to cause a new training of the NLU model as this is required for the end-to-end support
- get rid of a couple of `asyncio.coroutine` usages as this function is deprecated as of Python 3.8

### DIFF
--- a/rasa/shared/importers/importer.py
+++ b/rasa/shared/importers/importer.py
@@ -1,6 +1,6 @@
 import asyncio
 from functools import reduce
-from typing import Text, Optional, List, Dict, Set, Any
+from typing import Text, Optional, List, Dict, Set, Any, Tuple
 import logging
 
 import rasa.shared.constants
@@ -8,7 +8,7 @@ import rasa.shared.utils.common
 import rasa.shared.core.constants
 import rasa.shared.utils.io
 from rasa.shared.core.domain import Domain
-from rasa.shared.core.events import ActionExecuted, UserUttered, Event
+from rasa.shared.core.events import ActionExecuted, UserUttered
 from rasa.shared.nlu.interpreter import NaturalLanguageInterpreter, RegexInterpreter
 from rasa.shared.core.training_data.structures import StoryGraph
 from rasa.shared.nlu.training_data.message import Message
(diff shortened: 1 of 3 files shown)


rasa/shared/importers/importer.py
```diff
@@ -488,12 +488,19 @@ async def get_nlu_data(self, language: Optional[Text] = "en") -> TrainingData:
...
-                    additional_messages_from_stories.append(message)
+        utterances, actions = _unique_events_from_stories(stories)
+
+        # Sort events to guarantee deterministic behavior and to avoid that the NLU
+        # model has to be retrained due to changes in the event order within
+        # the stories.
+        sorted_utterances = sorted(
+            list(utterances), key=lambda user: user.intent_name or user.text
+        )
+        sorted_actions = sorted(list(actions), key=lambda action: action.action_name)
```
evgeniiaraz, line 499: why are we not using `action_text` here?

rasa/shared/importers/importer.py
```diff
@@ -488,12 +488,19 @@ async def get_nlu_data(self, language: Optional[Text] = "en") -> TrainingData:
...
-                message = _message_from_conversation_event(event)
-                if message:
-                    additional_messages_from_stories.append(message)
+        utterances, actions = _unique_events_from_stories(stories)
+
+        # Sort events to guarantee deterministic behavior and to avoid that the NLU
+        # model has to be retrained due to changes in the event order within
+        # the stories.
+        sorted_utterances = sorted(
+            list(utterances), key=lambda user: user.intent_name or user.text
```
ricwo, line 497:
```suggestion
            utterances, key=lambda user: user.intent_name or user.text
```

rasa/shared/importers/importer.py
ricwo, line 499:
```suggestion
        sorted_actions = sorted(actions, key=lambda action: action.action_name)
```

tests/shared/importers/test_importer.py
```diff
@@ -228,40 +238,69 @@ async def test_import_nlu_training_data_from_e2e_stories(project: Text):
...
+    # Patch to return our test stories
+    default_importer.importer.get_stories = mocked_stories
+
+    training_data = await default_importer.get_nlu_data()
+
+    # Pretend the order of  the stories changed. This should have no
+    # effect on the NLU training data
+    stories = list(reversed(stories))
+
+    # Make sure importer doesnt cache stories
```
ricwo, line 274:
```suggestion
    # Make sure importer doesn't cache stories
```

evgeniiaraz:
Just asking for my understanding -- why `rasa.nlu.training_data.training_data.sanitize_examples` isn't enough? I though it did the ordering by turning training examples into ordered dict. Or are we by-passing the initialization somehow?

ricwo (approved):
Looks great 👍

rasa/shared/importers/importer.py
wochinge: I was under the impression that always `action_name==action_text` . Added it now 👍

wochinge: Makes sense that `sorted` returns a list 🙈

rasa/shared/importers/importer.py
```diff
@@ -496,7 +496,9 @@ async def _additional_training_data_from_stories(self) -> TrainingData:
         sorted_utterances = sorted(
             utterances, key=lambda user: user.intent_name or user.text
         )
-        sorted_actions = sorted(actions, key=lambda action: action.action_name)
+        sorted_actions = sorted(
+            actions, key=lambda action: action.action_name or action.action_text
```
ricwo, line 500: but an action name can't be `None` or `''` can it (unlike the intent name)?

wochinge: it will actually once we add end-to-end reading 👍 (see https://github.com/RasaHQ/rasa/pull/6605)

PR #6683, opened 2020-09-16
implemented additional telemetry events

tmbo:
**Proposed changes**:
- added additional telemetry events


rasa/cli/data.py
```diff
@@ -187,13 +192,11 @@ def validate_files(args: argparse.Namespace, stories_only: bool = False) -> None
             and _validate_story_structure(validator, args)
         )
 
+    rasa.utils.common.run_in_loop(telemetry.track_validate_files(all_good))
     if not all_good:
         rasa.shared.utils.cli.print_error_and_exit(
             "Project validation completed with errors."
         )
-        rasa.shared.utils.cli.print_error_and_exit(
```
wochinge, line 194: 🚀

rasa/cli/data.py
```diff
@@ -230,6 +233,8 @@ def _convert_nlu_data(args: argparse.Namespace) -> None:
         NLUMarkdownToYamlConverter,
     )
 
+    rasa.utils.common.run_in_loop(telemetry.track_data_convert(args.format, "nlu"))
```
wochinge: We track the "convert" events **before** we start the conversion, while the data others in here are track **after** the operations are done. Would be great to have this consistent.

rasa/cli/export.py
```diff
@@ -3,9 +3,11 @@
 import typing
 from typing import List, Text, Optional
 
+from rasa import telemetry
 from rasa.cli import SubParsersAction
 import rasa.core.utils
 import rasa.shared.utils.cli
+import rasa.utils.common
```
wochinge, line 10: can be removed then (if you haven't done it by the time I submitted the PR 😆 )

rasa/cli/shell.py
```diff
@@ -110,8 +113,12 @@ def shell(args: argparse.Namespace):
     if not core_model:
         import rasa.nlu.run
 
+        telemetry.track_shell_started("nlu")
+
         rasa.nlu.run.run_cmdline(nlu_model)
     else:
         import rasa.cli.run
 
+        telemetry.track_shell_started("rasa")
```
wochinge, line 122: optional: technically this could be only core model at this point as well

rasa/core/run.py
```diff
@@ -207,10 +207,13 @@ async def clear_model_files(_app: Sanic, _loop: Text) -> None:
         if app.agent.model_directory:
             shutil.rmtree(_app.agent.model_directory)
 
+    async def run_telemetry(_app: Sanic, _loop: AbstractEventLoop) -> None:
+        telemetry.track_server_start(input_channels, endpoints, app.agent)
+
     app.register_listener(clear_model_files, "after_server_stop")
+    app.register_listener(run_telemetry, "after_server_start")
```
wochinge, line 214: This will get fired `number_of_sanic_workers` times 😬 (btw: the number of sanic workers would be interesting to track as well 🕵️ )

rasa/core/test.py
```diff
@@ -302,22 +304,21 @@ async def _generate_trackers(
     agent: "Agent",
     max_stories: Optional[int] = None,
     use_e2e: bool = False,
-) -> List[Any]:
+) -> "TrainingDataGenerator":
```
wochinge, line 307: we should rename the function `_generate_trackers` then as it no longer generates the trackers.

rasa/nlu/test.py
```diff
@@ -1515,6 +1517,8 @@ def run_evaluation(
             disable_plotting,
         )
 
+    telemetry.track_nlu_model_test(test_data)
```
wochinge, line 1520: do we also want to track `perform_nlu_cross_validation`?

rasa/telemetry.py
```diff
@@ -10,28 +9,42 @@
...
-from rasa.shared.importers.importer import TrainingDataImporter
-import rasa.shared.utils.io
+from rasa import model
 from rasa.constants import (
     CONFIG_FILE_TELEMETRY_KEY,
     CONFIG_TELEMETRY_DATE,
     CONFIG_TELEMETRY_ENABLED,
     CONFIG_TELEMETRY_ID,
 )
+from rasa.model import project_fingerprint
```
wochinge, line 32: please import it by module

rasa/telemetry.py
```diff
@@ -370,40 +395,29 @@ async def _send_event(
...
     """
-    try:
-        remote = check_output(  # skipcq:BAN-B607,BAN-B603
-            ["git", "remote", "get-url", "origin"], stderr=DEVNULL
-        )
-        return hashlib.sha256(remote).hexdigest()
-    except (CalledProcessError, OSError):
-        working_dir = Path(os.getcwd()).absolute()
-        return hashlib.sha256(str(working_dir).encode("utf-8")).hexdigest()
+    working_dir = Path(path).absolute()
```
wochinge, line 419: in practice it's the working directory, but just looking at the function it could be any dir

rasa/telemetry.py
```diff
@@ -628,50 +643,238 @@ async def track_model_training(
...
+        number_of_exported_events: Number of events that got exported
+        tracker_store: Store used to retrieve the events from
+        event_broker: Broker the events are getting published towards
+    """
+    track(
+        TELEMETRY_TRACKER_EXPORTED_EVENT,
+        {
+            "number_of_exported_events": number_of_exported_events,
+            "tracker_store": type(tracker_store).__name__ if tracker_store else None,
+            "event_broker": type(event_broker).__name__ if event_broker else None,
```
wochinge, line 743: both can't be `None`

rasa/telemetry.py
```diff
@@ -370,40 +395,29 @@ async def _send_event(
...
-                if not data.get("success"):
-                    logger.debug(
-                        f"Segment telemetry request returned a failure. "
-                        f"Response: {data}"
-                    )
-
-
-def _project_hash() -> Text:
-    """Create a hash for the project in the current working directory.
+    resp = requests.post(SEGMENT_ENDPOINT, headers=headers, json=payload)
```
wochinge, line 398: We need to explicitly set a timeout here (by default there is no timeout)

rasa/telemetry.py
```diff
@@ -370,40 +395,29 @@ async def _send_event(
...
+    if resp.status_code != 200:
+        logger.debug(
+            f"Segment telemetry request returned a {resp.status_code} response. "
+            f"Body: {resp.text}"
+        )
+    else:
+        data = resp.json()
+        if not data.get("success"):
+            logger.debug(
+                f"Segment telemetry request returned a failure. " f"Response: {data}"
```
wochinge, line 409: `black` is playing pranks
```suggestion
                f"Segment telemetry request returned a failure. Response: {data}"
```

wochinge (commented):
Looks great! 💯  I think the timeout for `requests` is the only more serious thing 👍

docs/docs/telemetry/events.json
```diff
@@ -133,9 +133,260 @@
...
+        "project"
+      ]
+    },
+    "Project Created": {
+      "description": "Triggered when a project is created using rasa init.",
+      "type": "object",
+      "properties": {
+        "init_directory": {
+          "type": "string",
+          "description": "Hash of the directory the project is created in"
```
wochinge, line 305:
```suggestion
          "description": "Hash of the directory path the project is created in"
```

rasa/telemetry.py
```diff
@@ -1,5 +1,4 @@
 import asyncio
```
wochinge, line 1: `track` is called by all the other `track_` functions. As `track` also uses the `@ensure_telemetry_enabled` decorator, `ensure_telemetry_enabled` will be executed twice for every tracking call. How about making `track_` protected and removing the decorator from it?

docs/docs/telemetry/events.json
```diff
@@ -133,9 +133,260 @@
...
+      "type": "object",
+      "section": "Data Handling",
+      "properties": {
+        "fraction": {
+          "type": "number",
+          "description": "how much data goes into train and how much goes into test"
+        },
+        "type": {
+          "type": "string",
+          "description": "Type of data, either 'nlu', 'core' or 'rasa'."
```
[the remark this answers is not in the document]
tmbo: yes that is right, but it is copied from other descriptions of the same attribute, I think it makes sense to keep the same descriptions if we reuse property names

docs/docs/telemetry/events.json
```diff
@@ -133,9 +133,260 @@
...
+      "type": "object",
+      "section": "Model Testing",
+      "properties": {
+        "project": {
+          "type": ["string", "null"],
+          "description": "Fingerprint of the project the tested model got trained in."
+        },
+        "num_story_steps": {
+          "type": "integer",
+          "description": "Number of story steps used for testing"
```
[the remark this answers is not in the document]
tmbo: well these are really story steps so I'd keep that for now (there is a difference even though it is a small one)

rasa/core/run.py
tmbo: mhm that is interesting 🤔 I'll move this to a different location then

wochinge: we use `run_operation_in_single_sanic_worker` in Rasa X if that is of any help

rasa/nlu/test.py
tmbo: we can add that later, don't think it is an important one right now

rasa/telemetry.py
tmbo: yes will do

tmbo:
@wochinge ready for another review

rasa/cli/scaffold.py
```diff
@@ -3,8 +3,10 @@
 import sys
 from typing import List, Text
 
+from rasa import telemetry
 from rasa.cli import SubParsersAction
 import rasa.train
+import rasa.utils.common
```
wochinge, line 9: same for the others
```suggestion
```

rasa/telemetry.py
```diff
@@ -10,31 +9,46 @@
...
-from subprocess import CalledProcessError, DEVNULL, check_output  # skipcq:BAN-B404
 import sys
 import textwrap
-from typing import Any, Callable, Dict, Optional, Text
+import typing
+from typing import Any, Callable, Dict, List, Optional, Text
 import uuid
 
 import aiohttp
+import async_generator
```
wochinge, line 19: there are a couple of unused imports in here

rasa/telemetry.py
```diff
@@ -628,50 +645,249 @@ async def track_model_training(
...
+    input_channels: List["InputChannel"],
+    endpoints: Optional["AvailableEndpoints"],
+    model_directory: Optional[Text],
+    number_of_workers: int,
+    is_api_enabled: bool,
+) -> None:
+    """Track when a user starts a rasa server.
+
+    Args:
+        number_of_workers: number of used Sanic workers
```
wochinge, line 777: The order of the function args and the docstring description should be the same imo

rasa/telemetry.py
```diff
@@ -628,50 +645,249 @@ async def track_model_training(
...
+        TELEMETRY_DATA_CONVERTED_EVENT,
+        {"output_format": output_format, "type": data_type},
+    )
+
+
+@ensure_telemetry_enabled
+def track_tracker_export(
+    number_of_exported_events: int,
+    tracker_store: "TrackerStore",
+    event_broker: Optional["EventBroker"],
```
wochinge, line 731: optional as out of the scope of the PR: the type annotation for `_prepare_event_broker` is wrong as `EventBroker` can be `Optional` at this point. However, after `_prepare_event_broker` there is no way the broker can be `None` as this would have led to a nullpointer exception.

wochinge (approved):
Awesome 🚀 

Have you seen [this comment](https://github.com/RasaHQ/rasa/pull/6683#discussion_r489244116) @tmbo ?

rasa/cli/data.py
tmbo: yes will change that :+1:

rasa/telemetry.py
tmbo: fixed it :+1:

Issue #6393, opened 2020-08-12
Model regression tests: Show last performance on master


dakshvar22:
**Description of Problem**:
Currently the model regression tests displays the performance of a PR on the defined set of configurations and datasets. To validate if the PR improves/degrades the performance, the user needs to know the last recorded performance of the scheduled run on the master branch. While internally, this is viewable on Metabase, it's time taking to check for each configuration what numbers have changed from the last run. It would be useful if the numbers from the last scheduled run are shown in the comment of the tests.

**Overview of the Solution**:
1. Fetch the results of the latest scheduled run from segment
2. Show the difference between numbers captured in the PR run v/s the numbers extracted from the last scheduled run.

**Examples** (if relevant):
The resultant results table could look like - 

Dataset: `Carbon Bot`

| Configuration | Intent Classification Micro F1 | Entity Recognition Micro F1 | Response Selection Micro F1 |
|---------------|-----------------|-----------------|-------------------|
| `BERT + DIET(bow) + ResponseSelector(bow)`<br> test: `1m1s`, train: `2m29s`, total: `3m29s` |0.7709(0.2)|0.6260(0.0)|0.5133(-0.1)|

where the numbers in the brackets are the difference between numbers captured in the PR run v/s the numbers extracted from the last scheduled run.

cc @tczekajlo

PR #6699, opened 2020-09-17
Model regression tests - show last performance on master

tczekajlo:
Solves #6393


tczekajlo:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]

```

tczekajlo:
@dakshvar22 Can you 👀  if you have a minute?

dakshvar22:
@tczekajlo The format looks good 👍 I'll probably add a small dummy change today to one of our components to shoot down the accuracy and test if the difference comes up correctly.

dakshvar22:
@tczekajlo Doesn't look like that worked 🤔

rasa/nlu/utils/hugging_face/transformers_pre_post_processors.py
```diff
@@ -127,8 +127,8 @@ def bert_embeddings_post_processor(
     Returns:
         sentence level embedding and post-processed sequence level embedding.
     """
-    sentence_embedding = sequence_embeddings[0]
-    post_processed_embedding = sequence_embeddings[1:-1]
+    sentence_embedding = np.zeros_like(sequence_embeddings[0])
+    post_processed_embedding = np.ones_like(sequence_embeddings[1:-1])
```
dakshvar22, line 131: We should revert these back to original before merging the PR

tczekajlo:
> @tczekajlo Doesn't look like that worked 🤔

Hmm, let me fix it 😄

tczekajlo:
@dakshvar22 Now, it looks better, can you 👀  if you have a minute?

dakshvar22:
The format and output looks good to me! :rocket: 
Haven't reviewed the actual implementation as I am not familiar with the code.

tczekajlo:
> The format and output looks good to me! 🚀
> Haven't reviewed the actual implementation as I am not familiar with the code.

Cool, I wanted to be sure that format and output are ok. I'll catch someone else to take a look at the implementation part 😄

.github/workflows/ci-model-regression.yml
```diff
@@ -460,6 +460,22 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Download an artifact with the results for the schedule version
+        run: |
+          # Schedule workflows ID - 1498790
```
ricwo, line 465: is there a way we can fetch the ID for the _latest_ or _current_ version of `ci-model-regression-on-schedule.yml`? this is not ideal as it needs to be manually updated if the other file changes.

if it has to happen via ID, we need to add comments here that the ID has to be updated if `ci-model-regression-on-schedule.yml` changes, and also in `ci-model-regression-on-schedule.yml` to update the ID in this file

ricwo: also we need to think about the case where another PR changes `ci-model-regression-on-schedule.yml`, so simply getting the latest ID wouldn't be right. can we fetch the latest workflow ID of that workflow for the current branch?

ricwo (commented):
Implementation looks really good 👍  it would be great if we could do something about that static workflow ID

.github/workflows/ci-model-regression.yml
tczekajlo: The IDs for workflows are pretty static, they don't change even there are several branches that have different versions of the same workflow, additionally, schedule workflows can exist only in the master branch.

I've changed a bit this part and now, we fetch ID via GH API for the scheduled workflow.

ricwo (approved):
Awesome, thanks for making those changes 🚀

PR #6717, opened 2020-09-21
add telemetry prefix to docs commands

erohmensing:
**Proposed changes**:
- hopefully fix the output here: https://rasa.com/docs/rasa/command-line-interface

### DIFF
--- a/docs/plugins/program_output.js
+++ b/docs/plugins/program_output.js
@@ -24,7 +24,7 @@ const defaultOptions = {
     docsDir: './docs',
     sourceDir: './docs/sources',
     include: ['**.mdx', '**.md'],
-    commandPrefix: 'RASA_TELEMETRY_ENABLED=false',
+    commandPrefix: '',
 };
 
 /**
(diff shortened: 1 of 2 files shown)


erohmensing:
unfortunately i couldn't replicate the issue locally, but this seems like it will work 😄

docs/plugins/program_output.js
```diff
@@ -24,7 +24,7 @@ const defaultOptions = {
     docsDir: './docs',
     sourceDir: './docs/sources',
     include: ['**.mdx', '**.md'],
-    commandPrefix: '',
+    commandPrefix: 'RASA_TELEMETRY_ENABLED=false',
```
m-vdb, line 27: could you add it in the scripts/ folder instead?

https://github.com/RasaHQ/rasa/blob/8d4886a29623be4d703378bfc1a5e3f462dd364d/docs/scripts/compile_program_outputs.js#L7

m-vdb (changes requested):

docs/plugins/program_output.js
erohmensing: oh ho, gotcha, yes. makes more sense! Thanks 👍

m-vdb (approved):

PR #6722, opened 2020-09-21
Update responses docs 

ricwo:
**Proposed changes**:
- update `responses.mdx`
- cc @samsucik 


docs/docs/glossary.mdx
```diff
@@ -111,6 +111,10 @@ description: Glossary for all Rasa-related terms
 
   Incoming messages are processed by a sequence of components called a pipeline. A component can perform tasks ranging from entity extraction to intent classification to pre-processing.
 
+## [Response / Template / Utterance](./responses.mdx)
+
+  A message template that is used to respond to a user. This can include text, buttons, images, and other content.
```
erohmensing, line 116:
```suggestion
  A message that an assistant sends to a user. This can include text, buttons, images, and other content.
```

docs/docs/responses.mdx
```diff
@@ -2,73 +2,65 @@
 id: responses
 sidebar_label: Responses
 title: Responses
-abstract: Responses are messages that your assistant can send to the user. You can also use responses as actions,
-  making them the simplest type of action.
+abstract: Responses are messages that your assistant sends to the user. A response
+ is typically a text, but it can include other content like images and buttons.
```
erohmensing, line 6:
```suggestion
abstract: Responses are messages that your assistant sends to the user. A response
 is usually only text, but can also include content like images and buttons.
```

docs/docs/responses.mdx
```diff
@@ -2,73 +2,65 @@
...
-`utter_` prefix followed by the retrieval intent name (`chitchat`)
-and the associated response key (`ask_name`).
-
-### Using Variables
+:::note
+Notice the special format of each retrieval action's name. The name starts with `utter_`, 
+followed by the retrieval intent's name (here `chitchat`) and finally a suffix specifying 
+the different response keys (here `ask_name` and `ask_weather`). See 
+[Retrieval Actions](./retrieval-actions.mdx) and examples of 
+[retrieval intents](./training-data-format.mdx#training-examples) to learn more.
```
erohmensing, line 43:
```suggestion
[Retrieval Intents](./training-data-format.mdx#training-examples) to learn more.
```

docs/docs/responses.mdx
```diff
@@ -2,73 +2,65 @@
...
+:::
 
-You can use variables in your responses to insert information
-collected during the conversation. You can either do that in your custom action
-code or by using the automatic slot filling mechanism.
+### Using Variables in Responses
 
-#### Defining Response Variables
+You can use variables to insert information into responses.
+Within a response, a variable is enclosed in curly brackets, see the variable `name` below:
```
erohmensing, line 49:
```suggestion
Within a response, a variable is enclosed in curly brackets. For example, see the variable `name` below:
```

docs/docs/responses.mdx
```diff
@@ -2,73 +2,65 @@
...
+```yaml-rasa {3} title="domain.yml"
 responses:
   utter_greet:
   - text: "Hey, {name}. How are you?"
 ```
 
-Rasa will automatically fill that variable with a value found in a slot called
-`name`. If the slot name is not found or there is no value in the slot, it will be filled
-with `None`.
+Now, when the `utter_greet` response is used, Rasa automatically fills in the variable with the
```
erohmensing, line 57:
```suggestion
When the `utter_greet` response is used, Rasa automatically fills in the variable with the 
```

docs/docs/responses.mdx
```diff
@@ -97,98 +89,63 @@ to achieve the same result:
...
+```yaml-rasa {3-4} title="domain.yml"
 responses:
   utter_greet:
   - text: "Hey, {name}. How are you?"
   - text: "Hey, {name}. How is your day going?"
 ```
-
-When the model predicts `utter_greet`, Rasa will randomly pick one of them.
+In this example, when `utter_greet` gets predicted as the next action, Rasa will randomly pick one 
+of the two response variations and use it.
```
erohmensing, line 101:
```suggestion
of the two response variations to use.
```

docs/docs/responses.mdx
```diff
@@ -97,98 +89,63 @@ to achieve the same result:
...
-      payload: '/inform{"game": "chess"}'
-    - title: "Checkers"
-      payload: '/inform{"game": "checkers"}'
 ```
 
-When using channel-specific responses, the value of the `channel` key should match
-the name used to provide the credentials for that channel
-in the `credentials.yml` file.
+:::note
+Make sure the value of the `channel` key matches the channel name used in your `credentials.yml` file.
```
erohmensing, line 121: I am actually not sure if this is the case 🤔  I think we want them to use the `name` returned by the input channel. In the case of existing connectors, those match, but in the case of a custom one, your credentials would have path.to.class.module instead of the name of the channel.

docs/docs/responses.mdx
```diff
@@ -198,31 +155,32 @@ responses:
...
+
 Each button in the list of `buttons` should have two keys:
 * `title`: The text displayed on the buttons that the user sees.
-* `payload`: The message sent from the user to the bot when a user clicks a given button.
+* `payload`: The message sent from the user to the assistant when the button is clicked.
 
 :::note bypass nlu with buttons
-It's common to use buttons as a shortcut to bypass NLU.
-Rasa Open Source will send messages starting with a `/` to the
+It's common to use buttons as a shortcut to bypass the ML-based NLU interpreter.
```
erohmensing, line 164: I wouldn't shorten ML here

docs/docs/responses.mdx
```diff
@@ -198,31 +155,32 @@ responses:
...
+Messages starting with `/` are sent straight to the
 `RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format.
-In the above example, if a user clicks either button to respond with their mood, the user input
+In the example above, if the user clicks a button, the user input
 will be directly classified as either the `/mood_great` or `/mood_sad` intent.
```
erohmensing, line 168:
```suggestion
will be directly classified as either the `mood_great` or `mood_sad` intent.
```

docs/docs/responses.mdx
```diff
@@ -231,16 +189,13 @@ You can add images to a response by providing a URL to the image under the `imag
 ### Custom Output Payloads
```
erohmensing, line 189: I think I would still include something about this getting sent to the output as JSON

docs/docs/responses.mdx
```diff
@@ -257,58 +212,52 @@ responses:
...
+  steps:
+  - intent: greet
+  - action: utter_greet
 ```
 
-When `utter_greet` is run as an action, it will send the message from
+When `utter_greet` action runs, it will send the message from
 the response back to the user.
 
 :::note
```
erohmensing, line 241: Let's give this note a name

docs/docs/responses.mdx
```diff
@@ -257,58 +212,52 @@ responses:
...
-If you want to change the text, or any other part of the bots response,
+If you want to change the text, or any other part of the response,
 you need to retrain the assistant before these changes will be picked up.
-
 :::
 
 ### Calling Responses from Custom Actions
```
erohmensing, line 246: Can we clarify this is using SDK and also show how to do it in another action server?

erohmensing (approved):
This is looking great and is so much easier to understand  🚀

docs/docs/responses.mdx
samsucik: Well, there is no page/section/subsection called Retrieval Intents (maybe there should be!), this is why I left it uncapitalised. @erohmensing should I still go ahead and make this capitalised for the sake of consistency?

docs/docs/responses.mdx
samsucik: @erohmensing I know nothing about this, it's just that the original wording didn't make sense to me. Would you be able to suggest something that's technically correct and worded lightly?

docs/docs/responses.mdx
erohmensing: Good point, keep it as is with now. I'll figure out what the standard on this should be and add it to our styleguide :)

docs/docs/responses.mdx
erohmensing: Make sure the value of the `channel` key matches the value returned by the `name()` method of your input channel. If you are using a built-in channel, this value will also match the channel name used in your `credentials.yml` file. 

Does that make sense? ^

erohmensing: for example here https://github.com/RasaHQ/rasa/blob/master/rasa/core/channels/facebook.py#L26

samsucik: Makes sense to me, thanks :-)

PR #6746, opened 2020-09-22
Updated Rasa example bots for 2.0

b-quachtran:
**Proposed changes**:
- Updated Rasa example bots for 2.0


examples/formbot/actions/actions.py
```diff
@@ -1,140 +1,117 @@
...
-                # convert "out..." to True
-                return {"outdoor_seating": True}
-            elif "in" in value:
-                # convert "in..." to False
-                return {"outdoor_seating": False}
-            else:
-                dispatcher.utter_message(template="utter_wrong_outdoor_seating")
-                # validation failed, set slot to None
-                return {"outdoor_seating": None}
+        return "action_submit_restaurant_form"
```
ArjaanBuijk, line 16: I am not sure that we should still use the term `form` in the name of the class and the comments. It is just a regular action, and we need a rule to execute it.

By using the name `form` in the action code, it is possible to think that this is some kind of special Action.

Another approach would be to just call this: `ActionRestaurant`

And then have a `ValidateRestaurantForm`. It does make sense to use the therm `Form` in the Validate action, because it is called automatically, so it is a special kind of action.

examples/formbot/actions/actions.py
```diff
@@ -1,140 +1,117 @@
...
-            # affirm/deny was picked up as True/False by the from_intent mapping
-            return {"outdoor_seating": value}
-
-    def submit(
-        self,
-        dispatcher: CollectingDispatcher,
-        tracker: Tracker,
-        domain: Dict[Text, Any],
-    ) -> List[Dict]:
+    def run(self, dispatcher, tracker, domain) -> List[EventType]:
```
ArjaanBuijk, line 18: This should be made async.

examples/formbot/actions/actions.py
```diff
@@ -1,140 +1,117 @@
...
 
         dispatcher.utter_message(template="utter_submit")
         return []
+
+
+class ValidateRestaurantForm(Action):
+    def name(self) -> Text:
+        return "validate_restaurant_form"
+
+    def run(self, dispatcher, tracker, domain) -> List[EventType]:
```
ArjaanBuijk, line 29: Should be async

examples/formbot/domain.yml
```diff
@@ -77,35 +78,39 @@ responses:
...
-          entity: feedback
-        - type: from_text
+    cuisine:
+      - type: from_entity
+        entity: cuisine
+        not_intent: chitchat
+    num_people:
+      - type: from_entity
+        entity: number
+        intent: [inform, request_restaurant]
```
ArjaanBuijk, line 94: Is this the proper format?
I thought lists in yaml are done as:

```
intent:
 - inform
 - request_restaurant

**Review by ArjaanBuijk — CHANGES_REQUESTED** [review]

File: examples/formbot/domain.yml (hunk 4 above)
Comment (b-quachtran) [reply examples/formbot/domain.yml:94]: It's technically valid, but yeah this doesn't follow best conventions and needs to be updated.

File: examples/formbot/actions/actions.py [hunk 5]
```diff
@@ -1,140 +1,118 @@
...
+        dispatcher.utter_message(template="utter_slots_values")
         return []
+
+
+class ValidateRestaurantForm(Action):
+    def name(self) -> Text:
+        return "validate_restaurant_form"
+
+    async def run(self, dispatcher, tracker, domain) -> List[EventType]:
+        extracted_slots = tracker.form_slots_to_validate()
```
ArjaanBuijk, line 31: The method `form_slots_to_validate` was renamed to `slots_to_validate` in the latest rasa-sdk version.

ArjaanBuijk (changes requested):

b-quachtran:
@Ghostvv Would this be merged into 2.0.X or master?

Ghostvv:
@b-quachtran I don't know, what is the plan here regarding branches

examples/formbot/actions/actions.py
```diff
@@ -1,140 +1,118 @@
...
-        self,
-        dispatcher: CollectingDispatcher,
-        tracker: Tracker,
-        domain: Dict[Text, Any],
-    ) -> List[Dict]:
+    async def run(self, dispatcher, tracker, domain) -> List[EventType]:
         """Define what the form has to do after all required slots are filled."""
 
         dispatcher.utter_message(template="utter_submit")
+        dispatcher.utter_message(template="utter_slots_values")
```
Ghostvv, line 22: why do we need an action to utter 2 templates? can't they be predicted?

examples/formbot/data/rules.yml
```diff
@@ -15,22 +15,21 @@ rules:
...
-
   - rule: submit form
     condition:
     - active_loop: restaurant_form
     steps:
       - action: restaurant_form
       - active_loop: null
-      - action: utter_submit
-      - action: utter_slots_values
+      - action: action_restaurant
```
Ghostvv, line 35: why don't we want to predict 2 utter templates?

examples/formbot/actions/actions.py
b-quachtran: I can't remember why I made this change 😅, but I agree. Just updated

examples/knowledgebasebot/data/stories.yml
```diff
@@ -0,0 +1,8 @@
+version: "2.0"
+stories:
+  - story: greet and then query knowledge base
+    steps:
+    - intent: greet
+    - action: utter_greet
+    - intent: query_knowledge_base
+    - action: action_query_knowledge_base
```
tmbo, line 8:
```suggestion
    - action: action_query_knowledge_base

```

examples/reminderbot/data/stories.yml
```diff
@@ -0,0 +1,8 @@
+version: "2.0"
+stories:
+  - story: Happy path
+    steps:
+    - intent: greet
+    - action: utter_what_can_do
+    - intent: ask_remind_call
+    - action: action_set_reminder
```
tmbo, line 8:
```suggestion
    - action: action_set_reminder

```

examples/formbot/tests/test_stories.yml
```diff
@@ -11,6 +11,7 @@ stories:
   - action: restaurant_form
```
tmbo, line 11: should this also get a `version: "2.0"`?

examples/formbot/actions/actions.py
```diff
@@ -1,140 +1,103 @@
...
+from rasa_sdk.events import (
+    SlotSet,
+    EventType,
+)
 
+class ValidateRestaurantForm(Action):
+    def name(self) -> Text:
+        return "validate_restaurant_form"
+
+    async def run(self, dispatcher, tracker, domain) -> List[EventType]:
```
tmbo, line 15:
```suggestion
    async def run(self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: Domain) -> List[EventType]:
```

I think `Domain` needs to be imported

tmbo:
@b-quachtran should we get this ready to go and merged so that we can release it in a patch?

examples/formbot/tests/test_stories.yml
b-quachtran: Yep, it should. Just added!

b-quachtran:
> @b-quachtran should we get this ready to go and merged so that we can release it in a patch?

Yeah, just pushed some updates and I think it should be ready to go

examples/formbot/domain.yml
```diff
@@ -83,35 +84,38 @@ responses:
   utter_restart:
     - text: "restarted"
 
+actions:
```
tmbo, line 87: this is duplicated (already present in L120)

tmbo (approved):
looks great, thanks a lot for the update of the example bots 🚀

PR #6755, opened 2020-09-23
[bugfix][WIP] Fix the bug of Chinese tokenization in 1.10.12

### LINKED ISSUE #6754: Bug of Chinese tokenization in 1.10.12
**Rasa version**: 1.10.12

**Python version**: 3.7.4

**Operating system** (windows, osx, ...): ubuntu 20.04

**Command or request that led to error**: rasa train

**Issue**:
When Chinese training example contains an OOV character, it will be converted to `[UNK]` as token string, when rasa tries to align tokens with entities it will count the token's length, it treat `[UNK]` as 5 length string which is incorrect and it will create tokens that not match the true training example.

I will submit a PR to fix it.

howl-anderson:
**Proposed changes**:
- when count string length, treat the length of special token `[UNK]` as 1, to fix the OOV issue

fix #6754

### DIFF
--- a/rasa/nlu/tokenizers/convert_tokenizer.py
+++ b/rasa/nlu/tokenizers/convert_tokenizer.py
@@ -71,11 +71,8 @@ class ConveRTTokenizer(WhitespaceTokenizer):
             # clean tokens (remove special chars and empty tokens)
             split_token_strings = self._clean_tokens(split_token_strings)
 
-            # ConverRT models are removed from official github repo
-            # (https://github.com/RasaHQ/rasa/issues/6806),
-            # so here we pass a fake UNK token for pass the test for now
             tokens_out += train_utils.align_tokens(
-                split_token_strings, token_end, token_start, ""
+                split_token_strings, token_end, token_start
             )
 
         return tokens_out
--- a/rasa/nlu/tokenizers/convert_tokenizer.py
+++ b/rasa/nlu/tokenizers/convert_tokenizer.py
@@ -71,8 +71,11 @@ class ConveRTTokenizer(WhitespaceTokenizer):
             # clean tokens (remove special chars and empty tokens)
             split_token_strings = self._clean_tokens(split_token_strings)
 
+            # ConverRT models are removed from official github repo
+            # (https://github.com/RasaHQ/rasa/issues/6806),
+            # so here we pass a fake UNK token for pass the test for now
             tokens_out += train_utils.align_tokens(
-                split_token_strings, token_end, token_start
+                split_token_strings, token_end, token_start, ""
             )
 
         return tokens_out
--- a/rasa/utils/train_utils.py
+++ b/rasa/utils/train_utils.py
@@ -98,6 +98,13 @@ def align_tokens(
     current_token_offset = token_start
 
     for index, string in enumerate(tokens_in):
+        # There is no absolute guarantee that the length of OOV token always be 1.
+        # But some documents (e.g.
+        # https://mccormickml.com/2019/05/14/BERT-word-embeddings-tutorial/#22-tokenization)
+        # show that it very likely to be 1 in most case.
+        # It seems that OOV tokens in most language (except for Chinese) are emoji characters.
+        # Chinese language has lots of characters, some rare characters may become OOV.
+        # This is not a perfect solution, but in practice it can solve most issues related to OOV
         string_len = len(string) if unk_token is None or string != unk_token else 1
 
         if index == 0:
--- a/rasa/utils/train_utils.py
+++ b/rasa/utils/train_utils.py
@@ -80,7 +80,10 @@ def update_similarity_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
 
 
 def align_tokens(
-        tokens_in: List[Text], token_end: int, token_start: int, unk_token: str
+    tokens_in: List[Text],
+    token_end: int,
+    token_start: int,
+    unk_token: Optional[Text] = None,
 ) -> List[Token]:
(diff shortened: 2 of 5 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @rctatman will take a look at it as soon as possible ✨

rasa/utils/train_utils.py
```diff
@@ -95,22 +95,24 @@ def align_tokens(
     current_token_offset = token_start
 
     for index, string in enumerate(tokens_in):
+        string_len = len(string) if string != "[UNK]" else 1
```
tabergma, line 98: Where does the `"[UNK]"` token come from? Would it be always `"[UNK]"` or can it also be something else?

tabergma: Also why should the length always be 1? Can it also happen that the orginial word has a length longer than 1? Would that not break?

tabergma (changes requested):
Thanks for tackling this issue! 

I am not quite sure I fully understand the problem. Added some questions in the comments.

Also we would need to have a changelog entry and some tests. Thanks.

rasa/utils/train_utils.py
howl-anderson: Hi Tanja, "[UNK]" is the default OOV token of BERT. It's my fault to think that all tokenizers use the same OOV token, after more investigation, I found that different models may have different OOV token settings. Luckily, all those tokenizers used in Rasa inherit from PreTrainedTokenizer, which we can get OOV token by access the attribute `unk_token` of the instance. I will change the interface of align_tokens, make sure it always working on the correct OOV token.

The length of the OOV token is not guaranteed always be 1, some documents (https://mccormickml.com/2019/05/14/BERT-word-embeddings-tutorial/#22-tokenization; https://github.com/huggingface/tokenizers) gives the information that it seems that only emoji character can cause OOV in  English and most many other languages. Tokenizer for Chinese may treat the character as the token but Chinese has lots of characters, so OOV still occurs sometimes, and this patch will be very helpful to such a situation.

rasa/utils/train_utils.py
```diff
@@ -80,7 +80,7 @@ def update_similarity_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
 
 
 def align_tokens(
-    tokens_in: List[Text], token_end: int, token_start: int
+        tokens_in: List[Text], token_end: int, token_start: int, unk_token: str
```
tabergma, line 83:
```suggestion
        tokens_in: List[Text], token_end: int, token_start: int, unk_token: Optional[Text] = None
```

rasa/utils/train_utils.py
```diff
@@ -95,22 +95,24 @@ def align_tokens(
     current_token_offset = token_start
 
     for index, string in enumerate(tokens_in):
+        string_len = len(string) if string != unk_token else 1
```
tabergma, line 98:
```suggestion
        string_len = len(string) if unk_token is None or string != unk_token else 1
```

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -71,8 +71,11 @@ def tokenize(self, message: Message, attribute: Text) -> List[Token]:
             # clean tokens (remove special chars and empty tokens)
             split_token_strings = self._clean_tokens(split_token_strings)
 
+            # ConverRT models are removed from official github repo
+            # (https://github.com/RasaHQ/rasa/issues/6806),
+            # so here we pass a fake UNK token for pass the test for now
             tokens_out += train_utils.align_tokens(
-                split_token_strings, token_end, token_start
+                split_token_strings, token_end, token_start, ""
```
tabergma, line 78: We don't need this change if we set a default value for the `unk_token`, see other comments.

rasa/utils/train_utils.py
tabergma: Also can you please add a comment why `1` to the code? Thanks.

rasa/utils/train_utils.py
howl-anderson: suggestion accepted!

rasa/nlu/tokenizers/convert_tokenizer.py
howl-anderson: suggestion accepted!

rasa/utils/train_utils.py
howl-anderson: Good suggestion! I already have done that in a new commit!

tabergma:
Looks good so far! Can you please add a changelog entry? Thanks.

tabergma (approved):
Looks great! :+1: Thanks for tackling all my comments.

Can you please merge the latest version of `1.10.x` into this branch as the convert test fixes are merged there? Thanks.

rasa/utils/train_utils.py
```diff
@@ -95,22 +98,31 @@ def align_tokens(
     current_token_offset = token_start
 
     for index, string in enumerate(tokens_in):
+        # There is no absolute guarantee that the length of OOV token always be 1.
```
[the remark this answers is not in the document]
howl-anderson: suggestion accepted!

rasa/utils/train_utils.py
```diff
@@ -95,22 +98,31 @@ def align_tokens(
...
+        # But some documents (e.g.
+        # https://mccormickml.com/2019/05/14/BERT-word-embeddings-tutorial/#22-tokenization)
+        # show that it very likely to be 1 in most case.
+        # It seems that OOV tokens in most language (except for Chinese) are emoji characters.
```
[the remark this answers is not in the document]
howl-anderson: suggestion accepted!

PR #6757, opened 2020-09-23
fill other slot if its entity mapping is unique

Ghostvv:
**Proposed changes**:
- fill other slot if its entity mapping is unique


rasa/core/actions/forms.py
```diff
@@ -101,6 +104,31 @@ def get_mappings_for_slot(
 
         return requested_slot_mappings
 
+    def _create_unique_entity_mappings(self, domain: Domain) -> Set[Text]:
+        unique_entity_slot_mappings = set()
```
tabergma, line 108: Why should it happen that we have a duplicated slot mapping? Can you give an example?

rasa/core/actions/forms.py
```diff
@@ -118,44 +146,50 @@ def intent_is_desired(
...
-                requested_slot_mapping.get("group"),
+                slot_mapping.get("role"),
+                slot_mapping.get("group"),
             )
             slot_fulfils_entity_mapping = matching_values is not None
 
-        return slot_equals_entity or slot_fulfils_entity_mapping
+        return (
+            slot_equals_entity
+            or entity_mapping_is_unique
```
tabergma, line 190: Why do we add this condition? Can you please explain and add a comment? Thanks.

tabergma (changes requested):
Code wise it looks good :100: Would be great if you could add a couple more comments, otherwise it is hard to understand the logic in a couple of months. Also, please add a changelog entry.

rasa/core/actions/forms.py
Ghostvv: if we have the following slot_mappings:
```    departure_city:
      - type: from_entity
        entity: city
        role: from
      - type: from_entity
        entity: city
    arrival_city:
      - type: from_entity
        entity: city
        role: to
      - type: from_entity
        entity: city
    arrival_date:
      - type: from_entity
        entity: date
```
then 
```
      - type: from_entity
        entity: city
        role: from
      - type: from_entity
        entity: city
        role: to
      - type: from_entity
        entity: date
```
are unique, because they don't fill different slots. However
```
      - type: from_entity
        entity: city
```
is not unique it fills different slots depending on which one is requested

tabergma: Ah ok, that makes sense, was missing that we don't care about the actual slot to fill in this check.

Should we add a comment like:
``` 
A slot mapping of type "from_entity" is unique if there is no other slot mapping with the same conditions. For example:
Given the slot mappings 
  destination_city:
    - type: from_entity
      entity: city
departure_city:
    - type: from_entity
      entity: city
the slot mapping
    - type: from_entity
      entity: city
is not unique as it is used for two different slots.
```

Ghostvv: added explanation in a docstring

rasa/core/actions/forms.py
Ghostvv: added comment

Ghostvv:
sorry, forgot to add changelog files. I added comments and docs, could you please check if they make sense

changelog/6757.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Fixed the bug when entity was extracted even
+if it had a role a group but roles or groups were not expected.
```
tabergma, line 2:
```suggestion
if it had a role or group but roles or groups were not expected.
```

rasa/core/actions/forms.py
```diff
@@ -101,6 +104,60 @@ def get_mappings_for_slot(
 
         return requested_slot_mappings
 
+    def _create_unique_entity_mappings(self, domain: Domain) -> Set[Text]:
+        """Finds mappings of type `from_entity` that are uniquely set some slot.
```
tabergma, line 108:
```suggestion
        """Finds mappings of type `from_entity` that uniquely set a slot.
```

tabergma (approved):
Thanks for adding comments :+1:

docs/docs/forms.mdx
```diff
@@ -127,6 +127,40 @@ forms:
       not_intent: excluded_intent
 ```
 
+If the `from_entity` mapping is unique for some slot then
+it can also be used extract this slot from a given entity
+even if this slot wasn't requested by the form.
```
ArjaanBuijk, line 132: What about simplifying this to:

'In `from_entity` mapping, only when an extracted entity uniquely maps onto a requested slot, the slot will be filled. The extracted entity will be ignored if the mapping is not unique."

Ghostvv: I'm not sure, since in your phrasing, it is confusing to me as if it'll not be used for requested slot

Ghostvv: rephrased using your suggestion

Issue #6752, opened 2020-09-23
Bug of Chinese tokenization in 2.0.0rc1


howl-anderson:
**Rasa version**: 2.0.0rc1

**Python version**: 3.8.2

**Operating system** (windows, osx, ...): ubuntu 20.04

**Command or request that led to error**: rasa train

**Issue**:
Every Chinese train example is split as one and only one token by mistake, all Chinese character without whitespace inside will treated as one token with each character as subtoken. when the example comes with entities, it will cause `UserWarning: Misaligned entity annotation`. A hidden and terrible result is that since the tokenization is incorrected, the training component (such like a DIET component) is not correct too although there is no error shows up.

I will submit a PR to fix it.

sara-tagger:
Thanks for raising this issue, @tabergma will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

xkungfu:
@howl-anderson 
你好，借楼请教一个问题，我在试用rc1时也遇到了问题，但是因为我是新手，我不能确定是源码的问题，还是我自己使用不当造成的。网上查到的资料都是关于旧版的rasa的，2.0版还没有见到中文方面的资料，rasa执行时又没有报错，所以对于我来说我无法确定问题。正好您在使用中文的配置，可否帮助我看一下我的配置是不是有问题？谢谢！

Rasa version: 2.0.0rc1
Python version: 3.6.9
OS: ubuntu 18.04

config.yml
```

language: "zh"
pipeline:
- name: "JiebaTokenizer"
  dictionary_path: "jiebatemp"
  analyzer: "char_wb"
  min_ngram: 1
  max_ngram: 4

policies:
  - name: MemoizationPolicy
    max_history: 5
```

the dict.txt format like below:

 
> RSS订阅 3 n
> 江南Style 3 n
> 1号店 3 n
> 中国 129470 ns
> 中国下载 3 b


rasa train successfully. but parse intent by rasa shell nlu, the result is null.

howl-anderson:
Hi @xkungfu, this PR thread is setup for discuss PR related topic, you are asking for help about the usage of Rasa 2.0 for Chinese. This PR thread is not the place to discuss your usage problems. please send me an email (my email address: u1mail2me@gmail.com) so I can help you with all the details.

HatuneMiku:
[...]

PR #6792, opened 2020-09-25
Fix the bug of Chinese tokenization for rasa2.0.0rc1

HatuneMiku:
**Proposed changes**:
- Fix the bug of Chinese tokenization for rasa2.0.0rc1.   use rasa 1.9.x token logic which works fine.

### DIFF
--- a/rasa/nlu/utils/hugging_face/hf_transformers.py
+++ b/rasa/nlu/utils/hugging_face/hf_transformers.py
@@ -271,9 +271,9 @@ class HFTransformersNLP(Component):
 
             token_ids_out += split_token_ids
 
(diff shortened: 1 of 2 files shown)


HatuneMiku:
fix #6752

sara-tagger:

tabergma:
@HatuneMiku I don't think we want to bring back the old logic as this is the first step in the direction of getting rid of the library specific tokenizers which is planned for the future. Also reading the discussion in https://github.com/RasaHQ/rasa/issues/6752 it seems like there is another solution to the problem. Do you agree?

HatuneMiku:
@tabergma Ok, let me check. I will close this one.

Issue #6721, opened 2020-09-21
Setting slots before conversation starts leads to empty bot responses


ricwo:
**Rasa version**: 2.0.0rc1

**Python version**: 3.7.8

**Operating system** (windows, osx, ...): osx

**Issue**: Settings lots via the API for a new conversation ID leads to an empty bot response after the first regular user message. The issue was first reported here: https://forum.rasa.com/t/api-unexpected-behavior-bug/33400

chaneyjd:
I think the tracker is under the mistaken assumption that because it is not empty (but with only SlotSet events) it has interacted with the user (incorrect assumption).

I am assuming also that there might be other events (API) that should be ignored here (or at least not assumed to be user interaction events).

**I am able to "correct" the issue by the following:**
In the applied_events method in trackers.py (core):

**FIND:
---
            else:
                applied_events.append(event)
---
**CHANGE TO:
---
                if not isinstance(event, (SlotSet)):
                    applied_events.append(event)
                else:
                    logger.info("Skipped putting SetSlot Event on applied events queue")
---

chaneyjd:
and the diff:

--- a/rasa/shared/core/trackers.py
+++ b/rasa/shared/core/trackers.py
@@ -425,7 +425,10 @@ class DialogueStateTracker:
                     event.action_name, applied_events
                 )
             else:
-                applied_events.append(event)
+                if not isinstance(event, (SlotSet)):
+                    applied_events.append(event)
+                else:
+                    logger.info("Skipped putting SetSlot Event on applied events queue")

ricwo:
thanks @chaneyjd for looking into this. do you want to open a PR and we can discuss your solution there?

chaneyjd:
sure, let me get a PR today and I'll get with you there

chaneyjd:
issue #6721 fix #6785 PR created ...

coderunite:
This issue still exists in 2.0.3.

```
2020-11-06 06:15:24 INFO     root  - Rasa server is up and running.
2020-11-06 06:16:11 DEBUG    rasa.core.lock_store  - Issuing ticket for conversation 'testbug1'.
2020-11-06 06:16:11 DEBUG    rasa.core.lock_store  - Acquiring lock for conversation 'testbug1'.
2020-11-06 06:16:11 DEBUG    rasa.core.lock_store  - Acquired lock for conversation 'testbug1'.
[...]

PR #6798, opened 2020-09-25
added logic to skip SlotSet events addition to applied_events when ap…

chaneyjd:
…plied_events is empty

**Proposed changes**:
- added logic to skip putting SlotSet events on applied_events when applied_events is empty
this was causing the tracker to miss starting session for user_utterance when only SlotSet was on the applied_events List (issue #6721 )
Tested against scripts in issue #6721 ... shown to fix / validate

### DIFF
--- a/rasa/shared/core/trackers.py
+++ b/rasa/shared/core/trackers.py
@@ -425,7 +425,10 @@ class DialogueStateTracker:
                     event.action_name, applied_events
                 )
             else:
-                applied_events.append(event)
+                if (len(applied_events) == 0) and isinstance(event, SlotSet):
+                    logger.info("Skipping addition of SlotSet event on applied_events when applied_events is empty")
+                else:
+                    applied_events.append(event)
 
         return applied_events
 


sara-tagger:

rasa/shared/core/trackers.py
```diff
@@ -425,7 +425,10 @@ def applied_events(self) -> List[Event]:
                     event.action_name, applied_events
                 )
             else:
-                applied_events.append(event)
+                if (len(applied_events) == 0) and isinstance(event, SlotSet):
```
ricwo, line 428: I think this should be addressed at the level where the SlotSet events are being added to the tracker, not when they're retrieved

chaneyjd: Can you explain that suggestion, as I would not recommend changing how the slots are put on the tracker ? and technically I think it is the way the events are pulled from the tracker and used in the TED policy in this unique scenario , can you confirm that ?

ricwo:
the issue is with how the tracker is created when settings slots via the API. Closing this in favour of this PR: https://github.com/RasaHQ/rasa/pull/7020

PR #6805, opened 2020-09-27
Filter response templates for retrieval intents in domain during sync

dakshvar22:
**Proposed changes**:
- Previously, all response templates from the domain were added to the NLU training data. This is not efficient because if any of the template is changed the whole NLU model will be re-trained since the fingerprint of the NLU data would have changed. Instead, only the response templates defined for retrieval intents should be brought back to the NLU training data. 


dakshvar22:
@degiz What do you think about this? I think this is important to avoid repeated training of NLU model. Do you see any problems with this change? I'll convert this to a proper PR if none.

tests/shared/importers/test_importer.py
```diff
@@ -360,5 +360,5 @@ async def test_nlu_data_domain_sync_with_retrieval_intents(project: Text):
 
     assert domain.retrieval_intents == ["chitchat"]
     assert domain.intent_properties["chitchat"].get("is_retrieval_intent")
-    assert domain.templates == nlu_data.responses
+    assert domain.retrieval_intent_templates == nlu_data.responses
```
degiz, line 363: This test is using `data/test_domains/default_retrieval_intents.yml` that doesn't have any responses for the retrieval intents 😬 So the test is comparing empty lists. Can we probably change it?

I'd also keep one more check:
```python
assert domain.templates != nlu_data.responses
```

degiz (changes requested):
@dakshvar22 Thanks for addressing that! 🚀 
I can't think of any problems the PR can cause.

Left a comment about test.

tests/shared/importers/test_importer.py
dakshvar22: It's actually not empty lists because, the NLU data file used(`data/test_responses/default.md`) has responses for retrieval intents defined. So, when a 2-way sync happens inside `RetrievalModelsDataImporter`, both `domain.retrieval_intent_templates` and `nlu_data.responses` should have the same response templates.
I agree `assert domain.templates != nlu_data.responses` is a good check to have there.

dakshvar22: Although, `assert domain.templates != nlu_data.responses` is not valid if your assistant has only retrieval intent templates and no other response templates. So, the real check is that `domain.retrieval_intent_templates == nlu_data.responses` and `domain.retrieval_intent_templates` is a proper subset of `domain.templates`. What do you think?

degiz: > NLU data file used(data/test_responses/default.md)

Right, I think I missed the point that it's a both way sync happening 😄 
My idea is - since we're modifying the domain file's logic, to test the domain, not the syncing.
So I'd introduce a domain file with both non-retrival-intents responses AND retrival-intents responses, and then check that `domain.templates != domain.retrieval_intent_templates` and both are non-empty.
Does that make sense?

degiz: ⏫ And that should ideally be a pure domain test, without any sync with NLU data.

dakshvar22: That makes sense. Added an additional test for this.

degiz (approved):
Thanks! 👍

changelog/6805.bugfix.md
```diff
@@ -0,0 +1,4 @@
+Transfer only response templates for retrieval intents from domain to NLU Training Data.
+
+This avoids retraining the NLU model if one of the non retrieval intent response templates
```
dakshvar22, line 3:
```suggestion
This avoids retraining the NLU model if one of the non retrieval intent response templates are edited.
```

changelog/6805.bugfix.md
```diff
@@ -0,0 +1,4 @@
+Transfer only response templates for retrieval intents from domain to NLU Training Data.
+
+This avoids retraining the NLU model if one of the non retrieval intent response templates 
+are edited.
```
dakshvar22, line 4:
```suggestion

```

Issue #6809, opened 2020-09-28
Add `unfeaturized` flag to slots


wochinge:
# Motivation

Slots themselves cause confusion in another way -- users may define the type based on what they hold (e.g. “text”) but not include them in their stories because they shouldn’t affect the flow -- it’s common that users don’t understand that setting the type to text will actually featurize the slot, leading to unpredictable behavior.

# Suggested Solution

1. Add a flag `unfeaturized: true / false` to the configuration of each slot in the domain. Return the slots features in `as_feature` depending on the value of the `unfeaturized` flag
   
   ```yaml
   slots:
     os:
       type: categorical
       values:
       - Windows
       - MacOS
       - Linux
       unfeaturized: true
    # this would previously been
    # os2:
      # type: unfeaturized
    ````
    Open questions:
    - [ ] What should be the default value for `unfeaturized`? I'd vote for `false` because we will otherwise breaking existing bots. 
2. Deprecate `UnfeaturizedSlot`
    Open questions:
    - [ ] Should we introduce a slot type `Any` instead which can hold arbitrary values? I'd vote for `yes`. We can introduce a slot type `Any` which explicitly throws an error when used with `unfeaturized: false`. If users want to use an `Any` slot with `unfeaturized: false`, they need to implement their own slot type
3. Check whether / how this affects Rasa X / Rasa SDK

wochinge:
@akelad @tmbo Please double check

We currently don't touch the slot values for `Unfeaturized`. Means if you feed a dictionary (e.g. as part of a custom action), then it will stay a dictionary

wochinge:
Ok, let's featurize all slots by default (except the old `UnfeaturizedSlot` as we will break all existing bots otherwise.

tmbo:
> Ok, let's featurize all slots by default (except the old UnfeaturizedSlot as we will break all existing bots otherwise.
Yes this sounds reasonable.

> Should we introduce a slot type Any instead which can hold arbitrary values? I'd vote for yes.
I'd agree. The unfeaturized slot type should stay and throw a deprecation warning.

Regarding the naming of the flag:
- I think it is easier to understand if a flags name is not negated, so `featurized` instead of `unfeaturized`
- Maybe we can come up with a better naming, e.g. `influences_conversation` or `influences_prediction` or something that is slightly less ml heavy speech?

wochinge:
[...]

PR #6817, opened 2020-09-28
unfeaturized slots

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6809
- add a flag `influence_conversation` to each slot which determines whether the slot should featurized. Each slot type is featurized by default to avoid that existing bots break. Only `UnfeaturizedSlots` are not featurized by default.
- deprecate `UnfeaturizedSlot` as users should use the flag instead
- replace usages of `UnfeaturizedSlot` with equal usages of unfeaturized, but not deprecated slots
- introduce a new slot type `AnySlot` which can store anything but is required to be unfeaturized. Users have to implement custom slots if they want to featurize an `AnySlot`
- raise exceptions if somebody tries to make `UnfeaturizedSlot` or `AnySlot` featurized
- removed `DataSlot` which was unused and doesn't make sense as well
- adapt examples to use `influence_conversation: False` instead of using the deprecated `unfeaturized` slot
- adapt `slot` documentation and documentation which refers to `unfeaturized` slots
- add migration guide for `unfeaturized` slots

**Todos**
- [x] agree on a flag name
- [x] docs (migration guide, slots docs)


docs/docs/domain.mdx
```diff
@@ -82,40 +82,73 @@ which can be used to store information the user provided (e.g their home city)
...
+    type: text
+    # this slot will not influence the predictions
+    # of the dialogue policies
+    influence_conversation: false
+
+```
 
 ### Slot Types
 
 Make this a nice table instead.
```
wochinge, line 141: I think a table would be too wide and hence annoying. Also, this way we can link to individual slot types which is nice.

wochinge:
@akelad Who from CSE would be best to help me with the documentation? I did a draft for the slots documentation, next step is the migration guide then.

akelad:
@wochinge just to clarify, you're gonna write the migration guide? General process is person building the feature writes the docs, CSE reviews

wochinge:
@akelad Yep, all done but I still need review + support as especially the changes for the slots docs were a bit more

akelad:
sorry was just unclear on what you meant with your comment, sounds good!

docs/docs/domain.mdx
```diff
@@ -260,44 +288,35 @@ Make this a nice table instead.
...
-
-
 * **Description**
 
-  The feature of this slot is set to `1` if a value with a list is set,
-  where the list is not empty. If no value is set, or the empty list is the
-  set value, the feature will be `0`. The **length of the list stored in
+  If `influence_conversation` is set to `true`, the assistant's behavior will change
+  depending on whether the list is empty or not. The **length of the list stored in
   the slot does not influence the dialogue**.
```
melindaloubser1, line 296: adding this note because it's  not mentioned anywhere
```suggestion

If a `list` slot is [auto filled by an entity of the same name](domain.mdx#slot-auto-fill), each entity value will be appended to the value of the slot. 

```

docs/docs/domain.mdx
```diff
@@ -222,35 +258,27 @@ Make this a nice table instead.
...
-  happens for values above `max_value`. Hence, if `max_value` is set to
-  `1`, there is no difference between the slot values `2` and `3.5` in
-  terms of featurization (e.g. both values will influence the dialogue in
-  the same way and the model can not learn to differentiate between them).
-
+  If `influence_conversation` is set to `true`, the assistant's behavior will change
+  depending on the value of the slot. All values below `min_value` will be treated as
+  `min_value`, the same happens for values above `max_value`.
+  Hence, if `max_value` is set to `1`, there is no difference between the slot values
+  `2` and `3.5`.
```
melindaloubser1, line 271: How would this slot type be used in a story? Based on the description it sounds like it only matters whether the value is above max or below min, but is that true?

docs/docs/domain.mdx
```diff
@@ -260,44 +288,35 @@ Make this a nice table instead.
...
 
-
-
 * **Description**
 
-  There will not be any featurization of this slot, hence its value does
-  not influence the dialogue flow and is ignored when predicting the next
-  action the bot should run.
-
+  Slots of type `any` are always ignored during conversations. If you want to store
```
melindaloubser1, line 317:
```suggestion
  Slots of type `any` are always ignored during conversations. The property `influence_conversation` cannot be set to `true` for this slot type. If you want to store
```

docs/docs/forms.mdx
```diff
@@ -421,8 +421,11 @@ class ValidateSlots(Action):
 ### The requested_slot slot
 
 The slot `requested_slot` is automatically added to the domain as an
```
melindaloubser1, line 423:
```suggestion
The slot `requested_slot` is automatically added to the domain as a
```

docs/docs/forms.mdx
```diff
@@ -421,8 +421,11 @@ class ValidateSlots(Action):
...
-unfeaturized slot. If you want to make it featurized, you need to add it
-to your domain file as a categorical slot. You might want to do this if you
+slot of type [text](domain.mdx#text-slot). The value of the `requested_slot` will be
+ignored during conversations. If you want to change this behavior, you need to add
+the `requested_slot`  to your domain file as a categorical slot with
+`influence_conversation` set to `true`.
```
melindaloubser1, line 427: Noticing this here but in general - are slot types e.g. `text` code or not? They're inconsistently formatted

docs/docs/migration-guide.mdx
```diff
@@ -469,6 +469,44 @@ response_selector_output.get("default")
 
   Alternatively, you can train a Core model without NLU using `rasa train core`.
```
melindaloubser1, line 470: this is a separate thing - but we don't talk about `core` anymore, so this docs section should be renamed. I'll make a note for it on the docs to dos

docs/docs/domain.mdx
melindaloubser1: Or it could go under "auto fill slots"

docs/docs/unexpected-input.mdx
```diff
@@ -64,11 +65,12 @@ slots:
...
       - preferences
       - feedback
+    influence_conversation: true
 ```
 
 This means that the dialogue model will pay attention to the value of the slot when making a prediction
-(read more about other [featurized slots](./domain.mdx#slots), whereas
-unfeaturized slots are only used for storing information.
+(read more about other [featurized slots](./domain.mdx#slots), whereas slots with
+the property `influence_conversation` set to `false` are only used for storing information.
```
melindaloubser1, line 73:
```suggestion
(read more about how [slots influence the assistant's behaviour](./domain.mdx#slots-and-conversation-behavior)).
```

melindaloubser1 (changes requested):
Thanks for writing the docs 🚀 Just a few suggestions for clarity.

docs/docs/domain.mdx
```diff
@@ -82,40 +82,73 @@ which can be used to store information the user provided (e.g their home city)
 as well as information gathered about the outside world (e.g. the result of a
 database query).
 
-Most of the time, you want slots to influence how the dialogue progresses.
-There are different slot types for different behaviors.
+Slots are defined in the `slots` section of your domain with their name and
+[type](domain.mdx#slot-types). The slot type will determine the data type of the slot
+and how the slot
+[influences the assistant's behavior](domain.mdx#slots-and-conversation-behavior).
+The following example defines a slot with name `slot_name` and the type `text`.
```
ricwo, line 89:
```suggestion
The following example defines a slot with name `slot_name` and type `text`.
```

docs/docs/domain.mdx
```diff
@@ -82,40 +82,73 @@ which can be used to store information the user provided (e.g their home city)
...
+
+```yaml
+slots:
+  # the name of the slot
```
ricwo, line 93: I think it's clear enough from the sentence before the example
```suggestion
```

docs/docs/domain.mdx
```diff
@@ -82,40 +82,73 @@ which can be used to store information the user provided (e.g their home city)
...
+  slot_name:
+    # the type of the slot
```
ricwo, line 95:
```suggestion
```

docs/docs/domain.mdx
```diff
@@ -82,40 +82,73 @@ which can be used to store information the user provided (e.g their home city)
...
+
+### Slots and Conversation Behavior
 
+Most of the time, you want slots to influence how the dialogue progresses.
 For example, if your user has provided their home city, you might
-have a `text` slot called `home_city`. If the user asks for the
+have a slot called `home_city`. If the user asks for the
 weather, and you *don't* know their home city, you will have to ask
-them for it. A `text` slot only tells Rasa Core whether the slot
+them for it. A [text slot](domain.mdx#text-slot) only tells Rasa Core whether the slot
```
ricwo, line 105: not sure about this one:
```suggestion
them for it. A [text slot](domain.mdx#text-slot) only tells Rasa whether the slot
```

docs/docs/domain.mdx
```diff
@@ -82,40 +82,73 @@ which can be used to store information the user provided (e.g their home city)
...
+```yaml
+slots:
+  # this slot will influence the conversation depending on
+  # whether the slot is set or not
+  home_city:
+    type: text
+    influence_conversation: true
+```
+
+By default all slot types will influence the conversation except slots of the type
```
ricwo, line 118:
```suggestion
By default all slot types will influence the conversation except slots of type
```

docs/docs/domain.mdx
```diff
@@ -82,40 +82,73 @@ which can be used to store information the user provided (e.g their home city)
...
+that the assistant will ignore the value of the slot during the conversation.
+
+```yaml
+slots:
+  age:
+    type: text
+    # this slot will not influence the predictions
+    # of the dialogue policies
+    influence_conversation: false
+
```
ricwo, line 136:
```suggestion
```

docs/docs/domain.mdx
```diff
@@ -125,26 +158,36 @@ Make this a nice table instead.
...
   ```
 
-
-
 * **Description**
 
-  Results in the feature of the slot being set to `1` if any value is set.
-  Otherwise the feature will be set to `0` (no value is set).
+  If `influence_conversation` is set to `true`, the assistant's behavior will change
+  depending on whether the slot is set or not. Different texts do not influence the
```
ricwo, line 164:
```suggestion
  depending on whether the slot is set or not. Different slot values do not influence the
```

docs/docs/domain.mdx
```diff
@@ -154,26 +197,21 @@ Make this a nice table instead.
...
-
 * **Description**
 
-  Sets two boolean features. The first feature is `1` if the slot is set
-  and `0` otherwise. The second feature is `1` if the slot value is
-  `"true"`, `"1"`, or a variation of these, and `0` otherwise.
-
+  If `influence_conversation` is set to `true`, the assistant's behavior will change
+  depending on whether the slot is set to `true` or `false`. An empty `bool` slot
+  is the same as if the slot was set to `False`.
```
ricwo, line 204: > is the same as if the slot was set to `False`.

I find this surprising. does this have to be the case? this means a bool slot's default value is always `false`

ricwo, line 204:
```suggestion
  depending on whether the slot is set to `true` or `false`. An empty `bool` slot
  is the same as if the slot was set to `false`.
```

docs/docs/domain.mdx
```diff
@@ -187,30 +225,28 @@ Make this a nice table instead.
...
+* **Type**
 
   `float`
 
 * **Use For**
 
-  Continuous values
-
-
+  Storing continuous numbers.
```
ricwo, line 249:
```suggestion
  Storing real numbers.
```
or just
```suggestion
  Storing numbers.
```

docs/docs/domain.mdx
```diff
@@ -260,44 +288,35 @@ Make this a nice table instead.
...
-
-  `unfeaturized`
+  `any`
 
 * **Use For**
 
-  Data you want to store which shouldn't influence the dialogue flow
-
-
+  Storing arbitrary types of values.
```
ricwo, line 305:
```suggestion
  Storing arbitrary values (they can be objects of any type, such as dictionaries).
```
maybe it's just me, but "type" has a special meaning in slots, and I worry that that is confusing. `type` in this case refers to any legal yaml object, not the slot types previously defined

docs/docs/writing-stories.mdx
```diff
@@ -84,15 +84,16 @@ stories:
...
-dialogue flow based on the information stored in it.
+Slots act as your bot’s memory. When you define a slot, you can define whether a
+[slot](domain.mdx#slots) should influence the conversation or not.
+Slots with the property `influence_conversation` set to `false` can only store
+information. Slots with the property `influence_conversation` set to `true` can affect
+the dialogue flow based on the information stored in it.
 
-Because they affect the dialogue flow, you need to include any featurized slots set by
-[custom actions](./actions.mdx) in your stories.
+Slots which influence the conversation need to be added to your stories. This also
```
ricwo, line 95: should we mention rules here as well? e.g.
```suggestion
Slots which influence the conversation need to be added to your stories or rules. This also
```

rasa/shared/core/slots.py
```diff
@@ -270,22 +335,31 @@ def as_feature(self) -> List[float]:
...
 
-    def feature_dimensionality(self) -> int:
+    def _feature_dimensionality(self) -> int:
         return len(self.values)
 
 
-class DataSlot(Slot):
+class AnySlot(Slot):
+    """Slot which can be used to store any value. Users need to create a subclass of
+       `Slot` in case the information is supposed to get featurized."""
```
ricwo, line 344:
```suggestion
    `Slot` in case the information is supposed to get featurized."""
```

rasa/shared/core/slots.py
```diff
@@ -270,22 +335,31 @@ def as_feature(self) -> List[float]:
...
+        value_reset_delay: Optional[int] = None,
         auto_fill: bool = True,
-    ):
-        super().__init__(name, initial_value, value_reset_delay, auto_fill)
+        influence_conversation: bool = False,
+    ) -> None:
+        if influence_conversation:
+            raise ValueError(
+                f"An {AnySlot.__name__} cannot be featurized. "
+                f"Please use a different slot type instead. See the "
```
ricwo, line 359: should we hint on being able to subclass this slot if they want to? otherwise that hint is only in the class docstring of `AnySlot`, which is very hidden for someone seeing this warning

tests/shared/core/test_slots.py
```diff
@@ -18,21 +22,21 @@ class SlotTestCollection:
...
 
-    def value_feature_pair(self, request):
+    def value_feature_pair(self, request: SubRequest) -> Tuple[Any, List[float]]:
         """Values where featurization is defined and should be tested."""
         raise NotImplementedError
 
-    def invalid_value(self, request):
+    def invalid_value(self, request: SubRequest) -> Any:
         """Values, that should be handled gracefully but where the
         featurization is not defined."""
```
ricwo, line 35:
```suggestion
```

tests/shared/core/test_slots.py
```diff
@@ -220,5 +253,29 @@ def invalid_value(self, request):
             ("unseen value", [0, 0, 0, 0, 0, 1]),
         ]
     )
-    def value_feature_pair(self, request):
+    def value_feature_pair(self, request: SubRequest) -> Tuple[Any, List[float]]:
         return request.param
+
+
+class TestAnySlot(SlotTestCollection):
```
ricwo, line 260: 💯

rasa/shared/core/domain.py
```diff
@@ -278,9 +278,8 @@ def collect_slots(slot_dict: Dict[Text, Any]) -> List[Slot]:
         # make a copy to not alter the input dictionary
         slot_dict = copy.deepcopy(slot_dict)
         for slot_name in sorted(slot_dict):
-            slot_class = Slot.resolve_by_type(slot_dict[slot_name].get("type"))
-            if "type" in slot_dict[slot_name]:
-                del slot_dict[slot_name]["type"]
+            slot_class = Slot.resolve_by_type(slot_dict[slot_name].pop("type", None))
+
```
ricwo, line 282: why get rid of the key `type` deletion here?

ricwo (approved):
Should we add a special warning to `resolve_by_type` so that users are alerted of the removal of the `unfeaturized` property? Right now all they'll see is that that slot doesn't exist, but this can be confusing when you haven't read the changelog and/or migration guide.

Other than that, great work 💯

wochinge:
Thanks for your blazing fast reviews ❤️

docs/docs/domain.mdx
wochinge: I chose `texts` as different slot values _do_ influence the prediction (if the value is `None`).

rasa/shared/core/domain.py
wochinge: `pop` removes it in one go. I split it up in two lines as this is too nested

rasa/shared/core/slots.py
wochinge: I think they shouldn't subclass this one but rather `Slot` directly. There is no benefit in subclassing this one. Will add it to the error message 👍

docs/docs/domain.mdx
wochinge: Is a list an object? 🤔

docs/docs/domain.mdx
wochinge: The story will be different depending on the concrete value of the float. So the slot value is capped to the supplied min / max values and then everything within this range is different for the bot. Would writing
```
depending on the value of the number
```
make it clearer?

docs/docs/domain.mdx
wochinge: changed the whole sentence 🙈

docs/docs/domain.mdx
wochinge: @melindaloubser1 This means if there are previous values stored, they won't be overridden?

docs/docs/forms.mdx
wochinge: I think formatting them as code is reasonable. Sometimes it just makes more sense to directly link to them. What other formatting have you seen?

docs/docs/domain.mdx
wochinge: you can change the `initial_value`  🤔 I think changing this would also be out of the scope of this PR

docs/docs/domain.mdx
melindaloubser1:
```suggestion
  If `influence_conversation` is set to `true`, the assistant's behavior will change
  depending on the value of the slot. If the value is between `min_value` and `max_value`, the specific value of the number is used. All values below `min_value` will be treated as
  `min_value`, and all values above `max_value` will be treated as `max_value`.
  Hence, if `max_value` is set to `1`, there is no difference between the slot values
  `2` and `3.5`.
```

better explicit than confused?

docs/docs/forms.mdx
melindaloubser1: I've just seen some "it is a text slot" vs " it is a `text` slot"
They can be formatted as code & linked to:

```suggestion
slot of type [`text`](domain.mdx#text-slot). The value of the `requested_slot` will be
ignored during conversations. If you want to change this behavior, you need to add
the `requested_slot`  to your domain file as a categorical slot with
`influence_conversation` set to `true`.
```

docs/docs/domain.mdx
ricwo: 👍

docs/docs/domain.mdx
ricwo: or just
> they can be of any type, such as dictionaries or lists

docs/docs/forms.mdx
wochinge: ah nice! Didn't know that

docs/docs/domain.mdx
melindaloubser1: I believe so? That's what I thought, but one way or another we should document the behaviour

melindaloubser1: Ok nope, that's what [this forum post]() implied, but I checked and it overwrites.
So this type is only really useful for custom actions right?

wochinge: your link doesn't link to the forum post 😆

wochinge:
@ricwo Not sure I'm understanding this one

> Should we add a special warning to resolve_by_type so that users are alerted of the removal of the unfeaturized property?

There is a deprecation warning if you use an `UnfeaturizedSlot`.

ricwo:
@wochinge oh of course - the class still exists. please ignore my comment 👍

wochinge:
it made me reconsider the content of the deprecation warning which was helpful 👍

docs/docs/domain.mdx
wochinge: Judging from the code it should be replaced and not extended

melindaloubser1: Oh ugh now I can't find it, but yes it seems like that's always been the case

melindaloubser1 (approved):

PR #6821, opened 2020-09-28
AssistantBuilder component for Rasa Playground; UI improvements for Playground

alwx:
**Proposed changes**:
- Adds `AssistantBuilder` component to Rasa (according to https://www.figma.com/file/HwHDoLUE4DTjzy3IAgCk6b/Prototyping?node-id=0%3A1)
- Updates the existing Prototyper chat component
- Fixes https://github.com/RasaHQ/playground/issues/70
- Fixes https://github.com/RasaHQ/playground/issues/57

This change requires this PR to be merged first: https://github.com/RasaHQ/tabula/pull/31

Screenshot 2020-09-28 at 18 08 55


docs/docs/prototype-an-assistant.mdx
```diff
@@ -3,9 +3,12 @@ id: prototype-an-assistant
 sidebar_label: 'Rasa Playground'
 title: 'Rasa Playground'
 abstract: Learn the basics of building an assistant with Rasa Open Source with this interactive guide. You will be able to customize the assistant, talk to it and download the project so you can continue to build.
+hide_table_of_contents: true
+full_width: true
```
m-vdb, line 7: it doesn't appear to work:

https://drive.google.com/file/d/1LouMzeZDJAuIje1Xa1KWoIc9aZ_QhGKA/view?usp=sharing

m-vdb (commented):
Looks good, nice work! I've tried it locally and it looks nice. One thing though: I couldn't make it work locally.

BTW, does this need design review by Abhi?

docs/docs/prototype-an-assistant.mdx
alwx: That's because this change needs to be merged first: https://github.com/RasaHQ/tabula/pull/31

docs/themes/theme-custom/theme/AssistantBuilder/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+  )
+}
+
+export default {
+  Container,
+  Section,
+  Text,
+  Code,
+  NextStepButton
+}
```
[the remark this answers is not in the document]
alwx: Done!

alwx:
@m-vdb there is a separate issue for chat component and 503 errors there: https://github.com/RasaHQ/playground/issues/70

docs/docs/prototype-an-assistant.mdx
m-vdb: 👍

m-vdb (approved):
Then it's good from my POV 🚀 (maybe rick will have more input)

ricwo (approved):
🚀

alwx:
@m-vdb @ricwo I asked you review again because it now also contains changes to bundler & chat component. https://github.com/RasaHQ/rasa-x/pull/3215

It's mostly frontend (that's why I am also requesting review from @Saladdin) but some other parts were affected as well. 
Might also be a good idea for @abhilasharoy to review how it looks and is it good enough for the next iteration :)

ricwo (approved):
Tested it and it looks great - works as expected 💯

abhilasharoy:
Did a design review with @alwx this morning and here are some points we noted down:

- Consider making the text block in the docs less wide again (Lu)
- Selected tabs should have bold text
- Let’s remove “live editor” tag on the code block
- The bottom right corner of the component is missing a rounded corner
- The text and code block halves of the component should both ideally have the same font sizes and line heights
- Increase padding around the text on both halves of the component. Also increase padding between the component and the paragraph above it
- Grey line separator on top could only appear for the left side of the component
- The buttons seem to have no padding on the right and left of the text value
- Ensure that line breaks do not break words
- Consider horizontal scroll for code block, and consider not showing till the user activates it by scrolling
- Please double check that the latest version of the content is used
- The train button is a different size than the other buttons
- Rounded edges in the second component should have same radius as the component above
- “Updated” word break in the second component
- Add the purple label “ Try it out!” on the left side bar
- Responsive versions of the components

lunelson:
@alwx @abhilasharoy I've pushed a solution for this in the latest 0.8.0 release of the docs theme, which supports an `.mdx-box-max` class on components to allow their width to escape the text column, PRs for the version bump are here

https://github.com/RasaHQ/rasa/pull/6903
https://github.com/RasaHQ/rasa-x/pull/3658

alwx:
Done everything except these two things:
1) Add the purple label “Try it out!” on the left side bar
2) Responsive versions of the components

@abhilasharoy can we skip both for now and handle them separately? The first is mostly about `tabula` changes, and it might take a while, and the second seems to make little sense because our theme is not really responsive + chat component is also not responsive. Making everything responsive is a huge task that will require a lot of changes, and I am not sure if that's what we should be focusing on now, especially considering that it still seems hard to do prototyping on mobile devices. What do you think?

abhilasharoy:
Yea @alwx we can definitely consider the responsive versions for later. 

The "try it out" tag might have helped us guide new users to the playground tab, but if it isn't possible for today, let's create an issue for it and handle it soon?

cc @TyDunn

TyDunn:
@abhilasharoy I am not sure we should make it responsive. You won't be able to download the project on mobile device or tablet? That said, maybe we tell users on those devices to try it on a laptop / desktop, rather than it being a bad experience that does not work

abhilasharoy:
@TyDunn You may not be able to download and continue building from a mobile device, but the fact that the entire playground page becomes difficult to use, could still be addressed.

That way users could read and navigate the steps, train and talk to the assistant etc. from their phones if they wanted.

TyDunn:
> but the fact that the entire playground page becomes difficult to use, could still be addressed.

I agree with this. However, I am not sure about the solution. What if we made it a static page with the steps instead and encourage people to use a laptop / desktop to try the interactive version?

abhilasharoy:
@TyDunn [We discussed](https://rasa-hq.slack.com/archives/C016FLYV4QN/p1601464558070400?thread_ts=1601461156.068500&cid=C016FLYV4QN) two possible solutions:
1. The UI component goes away and the steps are simply rendered as a scrollable list on smaller devices. This is the easier option.

2. We have a [responsive version](https://rasa-hq.slack.com/archives/C016FLYV4QN/p1601464558070400?thread_ts=1601461156.068500&cid=C016FLYV4QN) of the component, this might not be neccessary.

alwx:
Both options are not the easiest to implement, actually.

But again — we don't really support mobile devices yet (open this page on your mobile, for example: https://rasa.com/docs), so I don't think we need to invest time now to make the component responsive or even show list with editors instead of tabbed component.
In addition, it's not only about the tabbed component — it's also about the chat component that is also not responsive, and making it responsive would require a lot of work on `rasa-ui` side.

abhilasharoy:
Okay, makes sense!

Do you think we should still create an issue for the future @alwx ? Or should we skip it altogether for now?

alwx:
I would say it's better to create an issue. I will do that.

docs/docs/prototype-an-assistant.mdx
```diff
@@ -150,18 +196,17 @@ rules:
...
 ## Train and talk to your assistant
 
 Once you have reviewed the steps above, you’re ready to train your assistant. The training process generates a new machine learning model based on the training data you’ve provided.
 
 To train the assistant on the NLU data, stories, forms, rules and responses above, click the `Train` button:
 
-<TrainButton />
-
-<div id="rasa-chat-block"></div>
+<div id="rasa-chat-block" class="mdx-box-max"></div>
```
lunelson, line 209: HTML in MDX still should be in JSX format, i.e. using `className` instead of `class`

alwx: done!

lunelson (commented):
LGTM

alwx:
@lunelson we should also release a new version of `tabula` first. It's just about changing the version in `project.json` and creating a tag, right? Can I do that or you want to?

lunelson:
@alwx the release workflow is done from the root with `yarn release` (plus you need auth to publish to our npm org), but the changes you're using here are already in the `v0.8.0` release from yesterday, which is already update here in master I believe; you just might not have it in this branch?

alwx:
The latest chat block changes are not released yet (https://github.com/RasaHQ/tabula/commit/3cb6530a6c28e0c176e4ccbef0c345f4a1d16d92)

lunelson:
@alwx Ah yes, OK I've published `v0.8.1` with your changes

lunelson (approved):

Issue #6765, opened 2020-09-23
improve accessibility in docs


erohmensing:
- [x] make sure that every image has a descriptive caption and descriptive alt text 
- [x] check for non-descriptive link texts, like `here` and `this page` and replace with the name of what is being linked to
- [ ] check for pages where lists/bullets are being used where headers would be more appropriate

PR #6822, opened 2020-09-28
Docs accessibility

erohmensing:
**Proposed changes**:
- close https://github.com/RasaHQ/rasa/issues/6765


docs/docs/architecture.mdx
```diff
@@ -13,27 +13,22 @@ responds to a message:
 
 
 
-<img alt="image" src={useBaseUrl("/img/rasa-message-processing.png")} />
+<img caption="Rasa Open Source architecture" src={useBaseUrl("/img/rasa-message-processing.png")}
```
lunelson, line 16: This will work, but for raster sources it's better for performance to use the `<Image />` component in the way we discussed in Slack, and also, it's arguably better to place the image next to the doc and reference it relatively. This way it will be versioned along with the docs, and we avoid:

1. a potentially orphaned image in the `static` folder
2. an image which is up-to-date with current docs but no longer the right image for the older versions

docs/docs/connectors/slack.mdx
```diff
@@ -12,7 +12,7 @@ Once you have them you can add these to your `credentials.yml`.
 
 **How to get the Slack credentials:** You need to set up a Slack app.
 
-1. To create the app go to [https://api.slack.com/apps](https://api.slack.com/apps) and click
+1. To create the app go to [Your Apps](https://api.slack.com/apps) and click
```
lunelson, line 15: From an a11y POV this link is a little bare, since the text "Your Apps" which would be read out-of-context by someone navigating with assistive tech, is not self-explanatory. You can add a title to a link like this
```md
To create the app go to [Your Apps](https://api.slack.com/apps "The Your Apps section of your Slack interface")
```

docs/docs/connectors/slack.mdx
```diff
@@ -12,7 +12,7 @@ Once you have them you can add these to your `credentials.yml`.
...
    on **Create New App**.
```
lunelson, line 16: I think we said we were going to mark up App functions with `<mark>` tags, like this `==Create New App==`

docs/docs/fallback-handoff.mdx
```diff
@@ -275,7 +275,7 @@ responses:
 
 Rasa provides default implementations for asking which intent the user
 meant and for asking the user to rephrase. To customize the behavior of these actions,
-see the [documentation default actions](default-actions.mdx).
+see the documentation on[default actions](default-actions.mdx).
```
lunelson, line 278: You might need a space here

docs/docs/migrate-from/google-dialogflow-to-rasa.mdx
```diff
@@ -15,13 +15,16 @@ Navigate to your agent's settings by clicking the gear icon.
...
+alt="In DialogFlow, the gear icon that indicates settings sits next to the dropdown menu for
+choosing between projects."/>
 
 Click on the 'Export and Import' tab and click on the 'Export as ZIP' button.
 
 
 
-<img alt="Dialogflow Export 2" src={useBaseUrl("/img/dialogflow_export_2.png")} width="675" />
+<img caption="Selecting Export and Import" src={useBaseUrl("/img/dialogflow_export_2.png")} width="675"
+alt="The Settings page will let you select from multiple headers, of which you want to select Export and Import."/>
```
lunelson, line 27: Same comments here, re optimizing images with `Image` and versioning them locally

docs/docs/tuning-your-model.mdx
```diff
@@ -130,7 +130,11 @@ The image below shows the call order during the training of this pipeline:
 
 
 
-<img alt="Component Lifecycle" src={useBaseUrl("/img/component_lifecycle.png")} />
+<img caption="Component Lifecycle" src={useBaseUrl("/img/component_lifecycle.png")
+alt="The component lifecycle during training. Components are processed in the order
+they're listed in the configuration file.
+All components are created and initialized in order before they are trained in
+order and then persisted in order."} />
```
lunelson, line 137: I think you might want to reverse `caption` and `alt` here—and use an `Image` component for this since `caption` won't do anything to an HTML `img` tag. Also same comment as above re versioning

docs/docs/writing-stories.mdx
```diff
@@ -399,7 +399,7 @@ In some cases, such as single-turn interjections, you can handle context-switchi
 using [rules](./rules.mdx) instead of through [stories](./stories.mdx).
 
 Consider this conversation scenario:
-<Chat caption="Single-turn interjection">
+<Chat caption="A single-turn interjection">
```
lunelson, line 402: You might get flaky results from MDX if you don't leave empty lines between markdown and JSX instructions

lunelson (changes requested):

docs/docs/connectors/slack.mdx
erohmensing: Ooh i did not know this! that's super helpful

docs/docs/architecture.mdx
erohmensing: what do you mean by `place the imge next to the doc`? as in sth like `docs/docs/architecture-img.png`?

lunelson: Yes, that's what I mean 😄 . So then, taking that example, your code, assuming the MDX file is in the same directory, would look something like

```jsx
import archImage from './architecture-img.png';

<Image img={archImage} alt='this text will be invisble, in the alt attribute of the img' caption='this text will be visible, in the figcaption following the img' />
```

lunelson (approved):

docs/docs/connectors/slack.mdx
erohmensing: Noted, I'm going to come back to this though

Issue #6762, opened 2020-09-23
Docs for GET /conversations/<conversation_id>/story endpoint shows markdown example instead of yaml


m-vdb:
Relates to https://github.com/RasaHQ/rasa/issues/6378

We might want to change this part of the API spec (and maybe other parts):

https://github.com/RasaHQ/rasa/blob/a754bd9eb1a14af384189a342d16afa3aeaff51f/docs/static/spec/rasa.yml#L869-L881

cc @wochinge @degiz

PR #6823, opened 2020-09-28
Update markdown example to yml

yennycheung:
**Proposed changes**:
- See https://github.com/RasaHQ/rasa/issues/6762
- Modify markdown example of GET /conversations/<conversation_id>/story to yml

**Note for reviewer**:
- I looked into https://github.com/RasaHQ/rasa/blob/master/rasa/cli/initial_project/tests/test_stories.yml and guessed the format for how the story should look like. 😅  I could be completely off.

**To do**:
- [x] There are other examples with markdown, after I get the first feedback I'll update those too.
- [ ] ~Looks like there are some css issues as well, the example has a white background, I might fix it while I'm at it. 
image~


docs/static/spec/rasa.yml
```diff
@@ -880,16 +880,19 @@ components:
     200Story:
       description: Success
       content:
-        text/markdown:
+        yml:
```
m-vdb, line 883: I think this refers to the mime type, so I would suggest to change it to
```suggestion
        text/yaml:
```

docs/static/spec/rasa.yml
```diff
@@ -880,16 +880,19 @@ components:
...
-             - utter_on_it
-             - utter_ask_location
+            - story: 00055028
+              steps:
+              - user: |
+                  hello
+                intent: greet
+              - action: utter_ask_howcanhelp
+              - user: |
+                  I'm looking for a [moderately priced](price:moderate) [Indian](cuisine) restaurant for [two](people) people
```
m-vdb, line 892: This format for entities is deprecated (or even gone I think), see [the doc here](https://rasa.com/docs/rasa/training-data-format#entities). I think you could go with:
```suggestion
                  I'm looking for a [moderately priced]{"entity": "price", "value": "moderate"} [Indian]{"entity": "cuisine"} restaurant for [two]({"entity": "people"}) people
```

docs/static/spec/rasa.yml
```diff
@@ -880,16 +880,19 @@ components:
...
           example: >-
-            ## story_00055028
-
-            * greet: hello
-             - utter_ask_howcanhelp
-            * inform: I'm looking for a [moderately priced](price:moderate)
-            [Indian](cuisine) restaurant for [two](people) people
-             - utter_on_it
-             - utter_ask_location
+            - story: 00055028
```
m-vdb, line 885: I think the original story title was:
```suggestion
            - story: story_00055028
```

m-vdb (changes requested):
Nice, thanks for taking these changes over 💯 

I left a couple of comments. Additionally, could you update other `text/markdown` content in this `rasa.yml` file? I think that the spec for `/model/test/intents` and `/model/test/stories` can be updated, as they now accept yml request bodies (and this is the default now).

docs/static/spec/rasa.yml
yennycheung: Will do, I wanted to get your opinion if I'm getting the yml format right first. I will update the others now. :D

docs/static/spec/rasa.yml
yennycheung: I see, thanks for the pointer! 😃

m-vdb:
> Looks like there are some css issues as well, the example has a white background, I might fix it while I'm at it.

@yennycheung about that, it might be a bit more complicated than that. Since things are still a bit WIP on the API spec, I think it's better to put that on @lunelson 's radar so that he knows about it (he is in charge about our docs theme, among other things)

yennycheung:
> > Looks like there are some css issues as well, the example has a white background, I might fix it while I'm at it.
> 
> @yennycheung about that, it might be a bit more complicated than that. Since things are still a bit WIP on the API spec, I think it's better to put that on @lunelson 's radar so that he knows about it (he is in charge about our docs theme, among other things)
Yeah I agree with you, I was just taking a look at the theme. I'll focus on the markdown changes then. :)

docs/static/spec/rasa.yml
```diff
@@ -1359,117 +1362,90 @@ components:
...
-        * chitchat/ask_name
+        - intent: chitchat/ask_name
+          examples: |
             - my name is Sara, Rasa's documentation bot!
 
-        ## ask weather
-        * chitchat/ask_weather
+        - intent: chitchat/ask_weather
+          examples: |
             - it's always sunny where I live
```
m-vdb, line 1412: These are responses, not intents
```suggestion
        chitchat/ask_name:
        - text: my name is Sara, Rasa's documentation bot!

        chitchat/ask_weather:
        - text: it's always sunny where I live
```

m-vdb (changes requested):
almost there 💪 only one example to change

docs/static/spec/rasa.yml
yennycheung: I see, let me read up more on this. Thanks for the review!

m-vdb (approved):
🐩

lunelson:
@yennycheung @m-vdb this CSS conflict is on my radar, it will be fixed this week.

Issue #4341, opened 2019-08-28
Accept pathlib model path


Batalex:
**Description of Problem**:
I am fond of pathlib and of this particular snippet to load the newest model I have:
```python
import pathlib

from rasa.core.agent import Agent

models_folder = pathlib.Path("models")
latest_model = max(models_folder.glob("*.tar.gz"), key=lambda p: p.stat().st_ctime)

agent = Agent.load(str(latest_model))
```

The only friction here is `str(latest_model)`. Adding support for pathlib paths would be quite convenient.

**Overview of the Solution**:

 Since `os.path` functions are compatible with `pathlib`, the only change we would need to do is right there https://github.com/RasaHQ/rasa/blob/master/rasa/core/agent.py#L355:
```python
model_path = get_model(str(model_path))
```
I could submit a PR if the feedback is positive.
The interrogation left would be about the type hints: `model_path` is a `Text` right now, would a `Union[Text, Path]` type be acceptable?

sara-tagger:
Thanks for submitting this feature request 🚀@JustinaPetr will get back to you about it soon!✨

silvasara:
Hi, @Henrike100 and I are interested in taking this issue and will start working on it

JustinaPetr:
That's great @silvasara. Please let us know if you need any help with this.

Batalex:
Thanks @silvasara and @Henrike100, it works like a charm 👍 
I am closing the issue.

PR #6826, opened 2020-09-29
Cast model_path to string

Henrike100:
**Proposed changes**:
- Change `Agent.load` method to support `pathlib` paths, resolving issue #4341


sara-tagger:

rasa/core/agent.py
```diff
@@ -408,7 +408,7 @@ def load(
             if not os.path.exists(model_path):
                 raise ModelNotFound(f"No file or directory at '{model_path}'.")
             if os.path.isfile(model_path):
-                model_path = get_model(model_path)
+                model_path = get_model(str(model_path))
```
ricwo, line 411: The type annotation for the `model_path` argument needs to be updated

ricwo (changes requested):
Thanks a lot for submitting this PR 👍

rasa/core/agent.py
silvasara: We have checked the signatures of the related functions (```load()``` and ```get_model()```) their type annotations seem to be correct, so we didn't understand the suggested modification. Could you please clarify?

ricwo: sure - casting the model path to `str` would only be necessary if `model_path` is of type `pathlib.Path`. the type annotations in the `load` function only have `model_path: Text`

Henrike100: We added type annotation to `pathlib.Path` and added type checking. When you can, check if the changes are correct, please.

rasa/core/agent.py
```diff
@@ -408,7 +409,10 @@ def load(
             if not os.path.exists(model_path):
```
ricwo, line 409:
```suggestion
            model_path = Path(model_path)
            if not model_path.exists():
```

rasa/core/agent.py
```diff
@@ -408,7 +409,10 @@ def load(
...
                 raise ModelNotFound(f"No file or directory at '{model_path}'.")
             if os.path.isfile(model_path):
```
ricwo, line 411:
```suggestion
            if model_path.is_file():
```

rasa/core/agent.py
```diff
@@ -408,7 +409,10 @@ def load(
...
-                model_path = get_model(model_path)
+                if isinstance(model_path, pathlib.Path):
+                    model_path = get_model(str(model_path))
+                else:
+                    model_path = get_model(model_path)
```
ricwo, line 415:
```suggestion
                model_path = get_model(model_path)
```

ricwo: `get_model()` should receive similar support for `Path` objects

rasa/core/agent.py
```diff
@@ -3,6 +3,7 @@
 import os
 import shutil
 import tempfile
+import pathlib
```
ricwo, line 6:
```suggestion
from pathlib import Path
```

rasa/core/agent.py
```diff
@@ -391,7 +392,7 @@ def update_model(
     @classmethod
     def load(
         cls,
-        model_path: Text,
+        model_path: Union[Text, pathlib.Path],
```
ricwo, line 395:
```suggestion
        model_path: Union[Text, Path],
```

ricwo (changes requested):
Thanks for making those changes 👍 

It would be great if you could add `Path` support to `get_model()` as well, so we don't do unnecessary conversions between the two functions.

Where did you come across having to use `Agent.load()` with a `Path` object? Did you use it somewhere in your own code, or is this an error you saw in the rasa codebase?

Henrike100:
Thanks for the review, we will make the changes soon.

> Where did you come across having to use Agent.load() with a Path object?

We don't. The owner of the issue said:

> Adding support for pathlib paths would be quite convenient.

silvasara:
> 
> It would be great if you could add `Path` support to `get_model()` as well, so we don't do unnecessary conversions between the two functions.
> 

For these suggestions we would have to change the `get_model` so that it accepts both Text and Path, but we already do this treatment in the `load`, so we don't understand the need for the change, could you elaborate better?

ricwo:
@Henrike100 can you say more about why this would be convenient? are you calling `Agent.load` from your own code?

@silvasara Updating `get_model()` doesn't necessarily have to happen in this PR. The suggestions I made are still possible if you include a `str(model_path)` conversion before passing the parameter into `get_model`

Henrike100:
> @Henrike100 can you say more about why this would be convenient? are you calling `Agent.load` from your own code?

I'm just solving issue #4341. It was the owner of the issue who said:

> Adding support for pathlib paths would be quite convenient.

I don't know why it would be convenient. It was the owner of the issue who said it would be. I haven't used `Agent.load` anywhere.

> The suggestions I made are still possible if you include a `str(model_path)` conversion before passing the parameter into `get_model`

Line 413 already does this:

`model_path = get_model(str(model_path))`

I can do this in a different way if you want. I can also make the other suggestions.

ricwo:
> I'm just solving issue #4341. It was the owner of the issue who said:

It would be great if you could add a changelog entry referencing this issue

> Line 413 already does this:

yes but the prior `if` is not necessary - you can just pass `str(model_path)` into `get_model()`, regardless of whether its type is Text or Path

Henrike100:
I made the changes you suggested, except for those involving `get_model`.

> It would be great if you could add a changelog entry referencing this issue

I already wrote the changelog. The file name is: `4341.improvement.md`.
Are you talking about this file?

ricwo (approved):
looks great, and thanks for your contribution 🎉

ricwo:
@Henrike100 would you mind merging in master once again?

Henrike100:
@ricwo Merged.

PR #6827, opened 2020-09-29
Fix bug when filling other slots from initial intend or when requested slot is of a different type.

### LINKED ISSUE #6843: Form slot filling does not fill slots other than the requested one if they are of a different type than the requested slots.
**Rasa version**:2.0.0rc2

**Rasa SDK version** (if used & relevant): not used

**Rasa X version** (if used & relevant): not used

**Python version**: Python 3.6.9

**Operating system** (windows, osx, ...): Ubuntu 18.04 (dockerized)

**Issue**:
When a form tries to fill other slots via `extract_other_slots(...)` it is using the entity type of the requested slot, not the slot in question. In particular, when reacting to the triggering intent, the requested slot is None and thus the entity type is None as well.

```
2020-09-30 01:36:16 DEBUG    rasa.core.actions.forms  - No pre-filled required slots to validate.
2020-09-30 01:36:16 DEBUG    rasa.core.actions.forms  - Validating user input 'UserUttered(text: i like cats, intent: {'id': 5319795724814210961, 'name': 'mention_good_animal', 'confidence': 0.9830133318901062}, entities: [{'entity': 'animal', 'start': 7, 'end': 11, 'confidence_entity': 0.7312353849411011, 'role': 'liked', 'confidence_role': 0.5919401049613953, 'value': 'cats', 'extractor': 'DIETClassifier'}])'.
2020-09-30 01:36:16 DEBUG    rasa.core.actions.forms  - Validating extracted slots: {}
2020-09-30 01:36:16 DEBUG    rasa.core.actions.forms  - Request next slot 'animal_liked'
2020-09-30 01:36:16 ERROR    rasa.core.actions.action  - Couldn't create message for response 'utter_ask_animal_liked'.
```

```
Expecting a form to fill a slot from an entity of the triggering event.
```

```yml
slots:
  animal_liked:
   type: text
forms:
  test_form:
    animal_liked:
    - type: from_entity
      role: liked
      entity: animal
```

This relates to #4644.
A suggested fix is implemented in #6827 .

hendrik-skubch:
**Proposed changes**:
- When explicitly filling slots other than the requested one, in `extract_other_slots` the entity type of the requested slot does not matter. Instead the entity of the slot to be filled should be checked. This PR fixes that.

Consider the following example:
Domain.yaml
```yml
slots:
  animal_liked:
   type: text
forms:
  test_form:
    animal_liked:
    - type: from_entity
      role: liked
      entity: animal
```
We expect that if the user message `i like [dogs]{"entity":"animal", "role":"liked"}` opens form `test_form`, `animal_liked` is filled.
The observed behavior leaves the slot blank, since request_slot is none.

- documentation update is not needed

Fixes #6843


sara-tagger:
Thanks for opening a draft pull request 🚀If you have any questions, you can direct them to @amn41 ✨

rasa/core/actions/forms.py
```diff
@@ -259,28 +259,22 @@ def extract_other_slots(
         """
         slot_to_fill = tracker.get_slot(REQUESTED_SLOT)
 
-        entity_type_of_slot_to_fill = self._get_entity_type_of_slot_to_fill(
```
Ghostvv, line 262: entity type of slot to fill matters. It checks for that when filling other slots with different roles and groups inside `self.entity_is_desired`

Ghostvv (changes requested):
It seems to me that the problem that you're trying to solve is that entity of a different name doesn't fill a slot. This behavior is expected. However if your `from_entity` mapping is unique, this PR should make it possible to fill other slot: https://github.com/RasaHQ/rasa/pull/6757

rasa/core/actions/forms.py
hendrik-skubch: So in `extract_other_slots` you only want to fill other slots if their entity type is the same as the requested slot?
If so, why bother calling it when `REQUESTED_SLOT` is none?
In addition, it looks a little bug prone in case the two entity types are different, though I am probably overlooking something.

Ghostvv: > fill other slots if their entity type is the same as the requested slot

or when their `from_entity` mapping is unique

> If so, why bother calling it when REQUESTED_SLOT is none?

because even if entity name is the same as the slot, we still need to extract the value for validation

Ghostvv: > it looks a little bug prone in case the two entity types are different

what do you mean?

hendrik-skubch: WellI I can see the case is caught in `entity_is_desired`, so it's not a problem, just a little late.
But in case of filling from the triggering intent, wouldn't you want to fill other slots if a unique mapping exists? At the moment `entity_type_of_slot_to_fill` would be `None` and thus `entity_is_desired` return `False`.

Ghostvv: > wouldn't you want to fill other slots if a unique mapping exists?
that's we do in the master

hendrik-skubch: > > wouldn't you want to fill other slots if a unique mapping exists?
> 
> that's we do in the master

Cool, I have to try that next week, it's basically midnight here.

PR #6829, opened 2020-09-29
Allow overriding of default intents

### LINKED ISSUE #6606: Merging domains is broken if out_of_scope intent is defined by user / make default actions configurable
**Rasa version**: master

**Issue**: The `out_of_scope` intent is added to the [list of default intents](https://github.com/RasaHQ/rasa/blame/master/rasa/core/constants.py#L46) and hence its omitted when the domain is fetched as a dict because of this [line](https://github.com/RasaHQ/rasa/blob/f62839e5cb0e8ecb8cb7d85ac0f8fb078035c8c8/rasa/core/domain.py#L857)
While merging multiple domains, the domains are [fetched as dicts](https://github.com/RasaHQ/rasa/blob/master/rasa/core/domain.py#L237) which would not contain `out_of_scope` intent even if the user has defined it in their domain.

@wochinge mentioned that `out_of_scope` intent was added to default intents because we need it for the `TwoStageFallbackAction` in order to know whether the user denied the suggestion. Currently there is no way to configure default actions and so the only way is to make certain things default. We could make the deny intent configurable, something like - 
```
forms:
- two_stage_fallback:
    deny_intent: "out_of_scope"
```

👉 [Slack thread](https://rasa-hq.slack.com/archives/C36SS4N8M/p1599581976115300)

ricwo:
**Proposed changes**:
- users can override default intents in their domain again
- fixes #6606 


rasa/core/actions/two_stage_fallback.py
```diff
@@ -85,7 +85,7 @@ async def is_done(
     ) -> bool:
         _user_clarified = _last_intent_name(tracker) not in [
             DEFAULT_NLU_FALLBACK_INTENT_NAME,
-            USER_INTENT_OUT_OF_SCOPE,
+            USER_INTENT_REJECT_SUGGESTION,
```
wochinge, line 88: @ricwo The problem I see with this approach is that it potentially breaks existing assistants which are currently using `out_of_scope`. In that sense it would be better to keep the old intent. I think remembering which intents were added as "default intents" and which were added by the users might make the most sense then?

ricwo: 👍  let's switch to the other approach - allowing overriding default intents

rasa/shared/core/domain.py
```diff
@@ -448,6 +462,24 @@ def __init__(
         self.store_entities_as_slots = store_entities_as_slots
         self._check_domain_sanity()
 
+    @staticmethod
```
wochinge, line 465: out of scope:I find it slightly weird that the other static methods are _above_ the constructor of `Domain`. Is that just me or is it also weird for you?

rasa/shared/core/domain.py
```diff
@@ -448,6 +462,24 @@ def __init__(
...
+    ) -> List[Text]:
+        """Collects the default intents overridden by the user.
+
+        Args:
+            intents: User-provided intents.
+
+        Returns:
+            User-defined intents that are default intents.
+        """
+        intent_names: Set[Text] = {
```
wochinge, line 477: optional: What I don't like about it is that it replicates some behavior of `collect_intent_properties`:
- checking if we have a `dict` or `str`
- looping over the intents again (that's ok imo)

Another thing which isn't ideal imo:
- we have a lot of `staticmethod` methods which all manipulate the state of the domain. That's slightly paradox imo. 

```python
def __init__(....):
  ...
  # don't add the default intents here
  self.intent_properties = self.collect_intent_properties(intents, entities)
  # add them here and remember which ones were overriden
  self.intent_properties, self.overriden_default_intents = self.add_missing_default_intents(self.intent_properties)
```

```

File: tests/shared/core/test_domain.py [hunk 4]
```diff
@@ -368,6 +368,23 @@ def test_merge_yaml_domains():
...
+    # this domain contains an overridden default intent
+    test_yaml_2 = f"""intents:
+- greet
+- {default_intent}"""
+
+    domain_1 = Domain.from_yaml(test_yaml_1)
+    domain_2 = Domain.from_yaml(test_yaml_2)
+    domain = domain_1.merge(domain_2)
+
+    assert default_intent in domain.intents
```
wochinge, line 384: can we also test that the `default_intent` is dumped as part of `as_dict` ?

wochinge (approved):
💯

tests/shared/core/test_domain.py
ricwo: 👍

rasa/shared/core/domain.py
ricwo: yeah I agree 😄

rasa/shared/core/domain.py
ricwo: > checking if we have a dict or str

I thought about this but didn't see a reasonable way to abstract the `dict` or `str` checking to be re-used in those methods. it would have to be at the expense of introducing a new method `is_intent_of_dict_type` or something similar which is not necessary. it only happens twice so I think that's OK

> we have a lot of staticmethod methods which all manipulate the state of the domain. That's slightly paradox imo.

I would say that's appropriate in these cases as they're used in the constructor

> add them here and remember which ones were overriden

That makes the `self.overriden_default_intents` more complicated than it needs to be (it can just be a list of intent names, whereas `self.intent_properties` is an object). we'd have to return an awkward tuple

Issue #6670, opened 2020-09-15
Bring e2e in *for real* left-overs


evgeniiaraz:
**What's left to do?** _Reading in e2e data_ 

*   <span class="error">[x]</span> Bring back YAML reading for end-to-end data

Research

*   <span class="error">[x]</span> Store in trackers whether intent / text were used to predict next action
*   <span class="error">[x]</span> In prediction, get the tracker history up to the current state depending on what action was predicted from (get the `intent/text` in corresponding steps)
*   <span class="error">[x]</span> figure out the condition when TED should use text or intent
*   <span class="error">[x]</span> TEDPolicy vs RulesPolicy: figure out what to do when TED uses text input, but rules predict an action based on intent
*   <span class="error">[x]</span> TEDPolicy vs RulesPolicy: what to do with nlu_fallback
*   <span class="error">[x]</span> create and test on edge cases
*   <span class="error">[x]</span> Make some mixed test stories to check it works (they can be Sara based)
*   <span class="error">[x]</span> 4D Tensors
*   <span class="error">[x]</span> Add utterance transformer to TED;
*   <span class="error">[x]</span> Ability to exclude the output of an NLU featurizer from being used for training and inference
*   <span class="error">[x]</span> increase test coverage of state and tracker featurizers
*   <span class="error">[x]</span> bring entities into TED [https://github.com/RasaHQ/research/issues/121](https://github.com/RasaHQ/research/issues/121)
*   <span class="error">[x]</span> remove fake features, and use 4d-3d converter instead
*   <span class="error">[x]</span> check if training entities in DIET is still working [https://github.com/RasaHQ/rasa/issues/7289](https://github.com/RasaHQ/rasa/issues/7289)
*   <span class="error">[x]</span> nuke from orbit entity labels from text as user input
*   <span class="error">[x]</span> pass entities extracted by ted to tracker [https://github.com/RasaHQ/rasa/pull/7443](https://github.com/RasaHQ/rasa/pull/7443)
*   <span class="error">[x]</span> create e2ebot example that makes sense
*   <span class="error">[x]</span> docs!!!!

Engineering

*   <span class="error">[x]</span> Bot utterances #6410
*   <span class="error">[x]</span> policies return events [https://github.com/RasaHQ/rasa/issues/7039](https://github.com/RasaHQ/rasa/issues/7039)
[...]

PR #6833, opened 2020-09-29
RasaModelData can handle 4D Tensors

tabergma:
**Proposed changes**:
- RasaModelData can handle 4D Tensors

related to https://github.com/RasaHQ/rasa/issues/6670


tabergma:
We have a lot of test failures. But I guess most of them or all of them come from the `e2e` itself. Should we fix them first on `e2e` branch itself?

Ghostvv:
> Should we fix them first on e2e branch itself?
I didn't look into tests there 🙈

rasa/utils/tensorflow/model_data.py
```diff
@@ -458,21 +514,37 @@ def _get_shapes_types(self) -> Tuple:
...
-            if isinstance(features[0], scipy.sparse.spmatrix):
-                # scipy matrix is converted into indices, data, shape
-                shapes.append((None, features[0].ndim + 1))
-                shapes.append((None,))
-                shapes.append((features[0].ndim + 1))
-            elif features[0].ndim == 0:
-                shapes.append((None,))
-            elif features[0].ndim == 1:
-                shapes.append((None, features[0].shape[-1]))
+        def append_shape(_features: Union[Features_4D, Features_3D]) -> None:
```
Ghostvv, line 517: it looks like in the rest of the code, `features` correspond to `Union[List[Features_4D], List[Features_3D]]` shall we create type for it?
and rename `_features` here to smth more explicit?

rasa/utils/tensorflow/model_data.py
```diff
@@ -31,14 +31,22 @@
...
 #   "numpy array containing sparse features for every training example"
 # ]}
-Data = Dict[Text, Dict[Text, List[np.ndarray]]]
+# In case of e2e data, we also deal with 4d tensors, e.g.
+# "action_text" -> { "sequence": [
+#   [["numpy array containing dense features for every training example"]],
+#   [["numpy array containing sparse features for every training example"]]
+# ]}
+Features_4D = List[np.ndarray]
+Features_3D = np.ndarray
```
Ghostvv, line 40: hm, so list in 4D correspond to batch dim?, however in 3D it's first dim of ndarray.
I'm afraid it's prone to confusion after some time. Shall we create a special structure for Features, that would be basically ndarray, but contain all the info: whether it is sparse, whether it is 4D, which dimensions are what etc?

rasa/utils/tensorflow/model_data.py
```diff
@@ -323,14 +361,18 @@ def split(
...
+                if RasaModelData.is_in_4d_format(f)
+                else f[counts > 1]
                 for attribute_data in self.data.values()
                 for features in attribute_data.values()
                 for f in features
             ]
             # collect data points that are unique for their label
             solo_values = [
-                f[counts == 1]
+                list(np.array(f)[counts == 1])
```
Ghostvv, line 373: that is a very strange thing 😂, I'd say it'd be better to use some special structure, to avoid things like that

rasa/utils/tensorflow/model_data.py
Ghostvv: and it seems to be if we keep the outer `ndarray`, it'll simplify things in the code. Can we subclass `ndarray`?

tabergma: why would you subclass `ndarray`?

I was first thinking of keeping both as numpy array with different dimensionalities, but then you would need to pad the data event before adding it to `RasaModelData` when creating the 4D features, which would have complicated the model data creation in TED. So, what do you suggest as a structure?

Ghostvv: but the ndarray in 3D features also contain non padded data, basically serving as a list

Ghostvv: > why would you subclass ndarray? what do you suggest as a structure?

smth that maintain the slicing of ndarray (needed for operating on batch dimension), but also containing the info about sparse, 4D, etc. to simplify our checks

Ghostvv: basically we're already using `ndarray` as ragged array. Because we're padding it only at last moment before creating tf.tensors

Ghostvv: I think we can even use ragged tensors, avoiding need for manual padding https://www.tensorflow.org/guide/ragged_tensor

tabergma: So, you are saying 4D should become np.array(np.array(np.array)) and 3D np.array(np.array)? And then we wrap it into some kind of structure so that we know if it is sparse, 4D, etc. So basically this object would already contain the information around FeatureSignature. Can try that.

I guess ragged tensors does not work for sparse, does it?

tabergma: ok, will try something, will ping you once it is ready for review

Ghostvv: I meant ragged tensors for dense

Ghostvv: >  3D np.array(np.array)

I believe we already have it

tabergma: yes, we do, just wanted to clarify

evgeniiaraz:
@tabergma I think I'll slack off the review here because I'd like to take my time to understand everything but I feel like by the time I do, I might not have proper access to Rasa github 😅 (not sure what the policies here are)

Ghostvv:
@evgeniiaraz this PR is public PR 😜

rasa/utils/tensorflow/model_data.py
```diff
@@ -24,21 +24,85 @@
...
+    ):
+        feature_array = np.asarray(input_array).view(cls)
+
+        if number_of_dimensions <= 2:
+            feature_array.feature_dimension = input_array.shape[-1]
+            feature_array.is_sparse = isinstance(input_array[0], scipy.sparse.spmatrix)
+        elif number_of_dimensions == 3:
+            feature_array.feature_dimension = input_array[0].shape[-1]
+            feature_array.is_sparse = isinstance(input_array[0], scipy.sparse.spmatrix)
+        else:
```
Ghostvv, line 41: should we explicitly put `elif number_of_dimensions == 4:` and `else: raise`

rasa/utils/tensorflow/model_data.py
```diff
@@ -96,7 +162,7 @@ def items(self) -> ItemsView:
         """
         return self.data.items()
 
-    def values(self) -> ValuesView[Dict[Text, List[np.ndarray]]]:
+    def values(self,) -> ValuesView[Dict[Text, List[FeatureArray]]]:
```
Ghostvv, line 165:
```suggestion
    def values(self) -> ValuesView[Dict[Text, List[FeatureArray]]]:
```

rasa/utils/tensorflow/model_data.py
```diff
@@ -209,9 +275,9 @@ def feature_dimension(self, key: Text, sub_key: Text) -> int:
             return 0
 
         number_of_features = 0
-        for data in self.data[key][sub_key]:
-            if data.size > 0:
-                number_of_features += data[0].shape[-1]
+        for features in self.data[key][sub_key]:
+            if len(features) > 0:
+                number_of_features += features.feature_dimension
```
Ghostvv, line 280: let's rename `number_of_features`

rasa/utils/tensorflow/model_data.py
```diff
@@ -24,21 +24,85 @@
...
+        if obj is None:
+            return
+
+        self.feature_dimension = getattr(obj, "feature_dimension", None)
+        self.number_of_dimensions = getattr(obj, "number_of_dimensions", None)
+        self.is_sparse = getattr(obj, "is_sparse", None)
+
+        default_attributes = {
+            "feature_dimension": self.feature_dimension,
+            "number_of_dimensions": self.number_of_dimensions,
```
Ghostvv, line 61: I think `feature_dimension` is a confusing name (since features have multiple dimensions). Not sure what would be better name. In keras this number is called `units`.

rasa/utils/tensorflow/model_data.py
```diff
@@ -458,21 +531,35 @@ def _get_shapes_types(self) -> Tuple:
...
-                # scipy matrix is converted into indices, data, shape
-                shapes.append((None, features[0].ndim + 1))
-                shapes.append((None,))
-                shapes.append((features[0].ndim + 1))
-            elif features[0].ndim == 0:
-                shapes.append((None,))
-            elif features[0].ndim == 1:
-                shapes.append((None, features[0].shape[-1]))
+        def append_shape(_features: FeatureArray) -> None:
+            if _features.number_of_dimensions == 4:
```
Ghostvv, line 535: is there a reason why `number_of_dimensions == 4` is first? Can it be put as one of the elifs below?

rasa/utils/tensorflow/model_data.py
```diff
@@ -458,21 +531,35 @@ def _get_shapes_types(self) -> Tuple:
...
-            if isinstance(features[0], scipy.sparse.spmatrix):
-                # scipy matrix is converted into indices, data, shape
-                shapes.append((None, features[0].ndim + 1))
-                shapes.append((None,))
-                shapes.append((features[0].ndim + 1))
-            elif features[0].ndim == 0:
-                shapes.append((None,))
-            elif features[0].ndim == 1:
-                shapes.append((None, features[0].shape[-1]))
+        def append_shape(_features: FeatureArray) -> None:
```
Ghostvv, line 534: can this method become a part of `FeatureArray` class?

rasa/utils/tensorflow/model_data.py
```diff
@@ -843,7 +939,39 @@ def _pad_dense_data(array_of_dense: np.ndarray) -> np.ndarray:
         return data_padded.astype(np.float32)
 
     @staticmethod
-    def _scipy_matrix_to_values(array_of_sparse: np.ndarray) -> List[np.ndarray]:
+    def _pad_4d_dense_data(array_of_array_of_dense: FeatureArray) -> np.ndarray:
```
Ghostvv, line 942: ragged tensors are not suitable for us?

rasa/utils/tensorflow/models.py
```diff
@@ -285,7 +285,7 @@ def train_on_batch(
         self.optimizer.apply_gradients(zip(gradients, self.trainable_variables))
 
     def build_for_predict(
-        self, predict_data: RasaModelData, eager: bool = False
+        self, predict_data: RasaModelData, eager: bool = True
```
Ghostvv, line 288:
```suggestion
        self, predict_data: RasaModelData, eager: bool = False
```

rasa/utils/tensorflow/models.py
```diff
@@ -541,6 +541,22 @@ def batch_to_model_data_format(
...
+                        batch_data[key][sub_key].append(
+                            tf.SparseTensor(
+                                batch[idx],
+                                batch[idx + 1],
+                                [
+                                    batch[idx + 2][0],
+                                    batch[idx + 2][1],
+                                    batch[idx + 2][2],
+                                    feature_dimension,
+                                ],
```
Ghostvv, line 556: isn't it easier: `shape = [batch[idx + 2][i] for i in range(number_of_dimensions - 1)] + [feature_dimension]`? then we don't need the `if` about `number_of_dimensions`

Ghostvv (commented):
looks great! I think it is way more convenient now, thanks for trying it

rasa/utils/tensorflow/model_data.py
tabergma: You are too fast with reviewing :smile:  It is not yet fully working, wanted to look into that once everything else works. Please be patient :slightly_smiling_face:

Ghostvv: oh sorry 😅 was excited to look how subclassing ndarray idea worked out

tabergma:
Implementation of `FeatureArray` is done. Did not had time to take a look at ragged tensors. Will do that on Monday.

rasa/utils/tensorflow/model_data.py
tabergma: @Ghostvv `RaggedTensors` look good, however they are currently not supported by `tf.data.Dataset.from_generator`... It is already fixed (https://github.com/tensorflow/tensorflow/pull/41981), but seem to be not yet released. We would also need to update some of our layers I guess. I'll create an issue for it to update it in the future.

tabergma: https://github.com/RasaHQ/rasa/issues/6909

rasa/utils/tensorflow/model_data.py
```diff
@@ -152,7 +157,7 @@ def _validate_number_of_dimensions(
                 f"array: {input_array}."
             )
 
-    def get_shape_type_info(self) -> Tuple[List[Tuple], List[Tuple]]:
+    def get_shape_type_info(self) -> Tuple[List, List]:
```
Ghostvv, line 160: do we know what these `List`s contain?

rasa/utils/tensorflow/model_data_utils.py
```diff
@@ -186,8 +186,8 @@ def _features_for_attribute(
         tracker_features, zero_state_features[attribute]
     )
 
-    sparse_features = defaultdict(list)
-    dense_features = defaultdict(list)
+    sparse_features = defaultdict()
+    dense_features = defaultdict()
```
Ghostvv, line 190: do we need a `defaultdict`? can't it simply be a `dict`?

rasa/utils/tensorflow/model_data.py
tabergma: It depends on the array. Sometimes it is a list of tuples of ints, sometimes it is just an int. I could change it to
```
Tuple[List[Union[int, Tuple[None], Tuple[None, int], Tuple[None, None, int], Tuple[None, None, None, int]]], List[int]]
```

Ghostvv: oh no

Ghostvv (approved):
Looks great! one small thing maybe: sparse 4D and 3D methods, seems to have quite a bit of shared code. Does it make sense to unite them somehow?

tabergma:
They actually do not really share that much code. The logic is the same but for 4D we always have an additional index or for loop. So if we would to extract some of the logic you would have a lot of if else statements I guess. So might be not worth it. 

The tests that are failing should be related to the changes in `e2e`. I guess we just merge and solve them later?

Ghostvv:
yes, let's merge then

Issue #6716, opened 2020-09-21
Converter tool not working for E2E stories


degiz:
Part of #6638 

### The Issue:

The converter script converts this markdown e2e story:

```markdown
* faq: what is [Rasa X](product)?
    - slot{"product": "x"}
    - respond_faq
    - action_set_faq_slot
```

into this yaml:

```yaml
- story: faqs
  steps:
  - checkpoint: STORY_START
  - intent: 'faq: what is [Rasa X](product)?'
  - slot_was_set:
    - product: x
  - action: respond_faq
  - action: action_set_faq_slot
```

... which doesn't match the E2E format in the docs ([https://rasa.com/docs/rasa/next/testing-your-assistant#end-to-end-testing](https://rasa.com/docs/rasa/next/testing-your-assistant#end-to-end-testing))

The `intent` section should look different:
```yaml
...
- user: |
      what is [Rasa X](product)?
  intent: faq
...
```

### How to reproduce:

 - create a text file `test_faq.md` with markdown snippet from above
 - run `rasa data convert core -f yaml --data {path_to_the_test_faq.md} --out {any_folder}` 
 - this will create a file called `test_faq_converted.yaml` in the output directory

Files to take a look to: 
- `rasa/cli/data.py`
- `rasa/core/training/converters/story_markdown_to_yaml_converter.py`
- `rasa/shared/core/training_data/story_writer/yaml_story_writer.py`

Test to take a look to:
- `tests/cli/test_rasa_data.py`
- `tests/shared/core/training_data/story_writer/test_yaml_story_writer.py`

PR #6839, opened 2020-09-29
fix converter tool to convert test stories to yaml

kalkbrennerei:
closing #6716 

**Proposed changes**:
- modified `markdown_to_yaml_converter` to check for test file extension and use the `use_e2e` label of the `markdown_story_reader`
- passed intent to the `story_step_builder` in `markdown_story_reader`


rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -241,6 +241,33 @@ def parse_e2e_message(line: Text) -> Message:
 
         return example
 
+    def _parse_message_e2e(self, message: Message, line_num: int) -> UserUttered:
```
degiz, line 244: This method looks very similar to the `_parse_message()`. Can we combine them together and use the `if self.use_e2e:` to reflect the necessary differencies?

rasa/core/training/converters/story_markdown_to_yaml_converter.py
```diff
@@ -42,6 +42,11 @@ async def convert_and_write(cls, source_path: Path, output_path: Path) -> None:
         )
 
         reader = MarkdownStoryReader(unfold_or_utterances=False)
+
+        # check if source file is test story file
+        if reader.is_markdown_test_stories_file(source_path):
```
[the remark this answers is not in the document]
degiz: Also, I'd first check `is_markdown_test_stories_file()` and then create an instance of `MarkdownStoryReader` and pass the `use_e2e` to the constructor.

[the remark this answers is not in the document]
degiz: And last point - I think for the test_strories we need to prepend the target filename with `test_` prefix.
And implement one test for that in `tests/cli/test_rasa_data.py`

tests/core/training/converters/test_story_markdown_to_yaml_converter.py
```diff
@@ -65,3 +74,47 @@ async def test_stories_are_converted(tmpdir: Path):
...
+
+    with open(f"{converted_data_folder}/test_stories_converted.yml", "r") as f:
+        content = f.read()
+        assert content == (
+            f'version: "{LATEST_TRAINING_DATA_FORMAT_VERSION}"\n'
+            "stories:\n"
+            "- story: ask product\n"
+            "  steps:\n"
+            "  - intent: faq\n"
+            "    user: |\n"
```
degiz, line 112: I checked and it's correct to expect the `|-` here, since `|` means there's a newline after the value.

degiz (changes requested):
Great job! 👍

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
kalkbrennerei: i thought about that, but then they are taking different inputs, one is taking only a string and one the whole message.

degiz (approved):
Great job @kalkbrennerei !
I've pushed a commit with small refactorings to spped up before the code freeze, but overall - everything's OK! 👍

PR #6845, opened 2020-09-30
add check for incomplete rules

Ghostvv:
**Proposed changes**:
- check action fingerprints in rules to make sure actions always set slots and active loops if they set it in other rules


docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
 * `enable_fallback_prediction` (default: `true`): Please see the
    [fallback documentation](fallback-handoff.mdx#handling-low-action-confidence) for
    further information.
-* `check_for_contradictions` (default: `true`): After training, the RulePolicy will
-   perform a check to make sure that there are no rules that contradict with each other, or with any stories.
+* `check_for_contradictions` (default: `true`):
+   Before training, the RulePolicy will perform a check to make sure that
+   actions always set slots and active loops that they set in other rules.
```
b-quachtran, line 351:
```suggestion
slots and active loops set by actions are defined consistently for all rules. 
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+   The following snippet is an example for an incomplete rule:
```
b-quachtran, line 352:
```suggestion
   The following snippet is an example of an incomplete rule:
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+
+   ```yaml-rasa
+   rules:
+   - rule: complete rule
+     steps:
+     - intent: do_something
```
b-quachtran, line 358:
```suggestion
     - intent: search_venues
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+     - action: action_that_sets_some_slot
```
b-quachtran, line 359:
```suggestion
     - action: action_search_venues
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+     - slot_was_set:
+       - some_slot: "some value"
```
b-quachtran, line 361:
```suggestion
       - venues: [{"name": "Big Arena", "reviews": 4.5}]
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+
+   - rule: incomplete rule
+     steps:
+     - intent: do_something
```
b-quachtran, line 365:
```suggestion
     - intent: search_venues
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+     - action: action_that_sets_some_slot
```
b-quachtran, line 366:
```suggestion
     - action: action_search_venues
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+   ```
+
+   In the second `incomplete rule`, `action_that_sets_some_slot` is expected to set
```
b-quachtran, line 369:
```suggestion
   In the second `incomplete rule`, `action_search_venues` should set
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+   `some_slot` but does not do it.
```
b-quachtran, line 370:
```suggestion
   the `venues` slot because it is set in `complete rule`, but this event is missing.
```

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+   rules:
+   - rule: fixes incomplete rule
+     steps:
+     - intent: do_something
+     - action: action_that_sets_some_slot
+     - slot_was_set:
+       - some_slot: null
+   ```
+
+   - if you want to deal with setting slot in a different rule or story,
```
b-quachtran, line 385: Even though it would solve the contradiction, it seems like bad practice for users to put slot sets only in certain rules and use the `wait_for_user_input` key. I'm leaning towards removing this recommendation.

docs/docs/policies.mdx
```diff
@@ -346,8 +346,56 @@ policies:
...
+     steps:
+     - intent: do_something
+     - action: action_that_sets_some_slot
+   ```
+
+   In the second `incomplete rule`, `action_that_sets_some_slot` is expected to set
+   `some_slot` but does not do it.
+   There is several possible ways to fix this rule, for example:
+
+   - if `some_slot` should not be set, you should explicitly set it to `null`
```
b-quachtran, line 373: There should be some clarification that adding `- some_slot: null` doesn't mean the slot gets set to `null` by the policy

b-quachtran (changes requested):
Awesome feature enhancement! Just a few changes to the rule examples and some questions around the recommendations for fixing contradictions.

docs/docs/policies.mdx
Ghostvv: I'm confused, how is that different from `- some_slot: some_value`? Users know that it is not set by the policy?

docs/docs/policies.mdx
Ghostvv: to me it's how rules should be written (especially taken into account that we restrict them by one user turn), but I can remove that

docs/docs/policies.mdx
Ghostvv: this is a weird format. Can a slot be a dictionary?

docs/docs/policies.mdx
Ghostvv: also, this is not a recommendation, it is a showcase of possible solutions

changelog/6845.improvement.md
```diff
@@ -0,0 +1,2 @@
+Check action fingerprints in rules to make sure actions always set slots
```
wochinge, line 1: What is an "action fingerprint"? I think we don't need a changelog for this as the rule policy in itself wasn't released yet.

rasa/core/policies/ensemble.py
```diff
@@ -737,6 +707,64 @@ def _check_policy_for_forms_available(
         )
 
 
+def _training_events_from_trackers(
```
wochinge, line 710: the name doesn't seem to correlate with what this is doing. Can we please find a better name? `_side_effects_per_action` or something?

rasa/core/policies/ensemble.py
```diff
@@ -737,6 +707,64 @@ def _check_policy_for_forms_available(
...
+    training_trackers: List[DialogueStateTracker],
+) -> Dict[Text, Set[Event]]:
+    """Creates a dictionary of action names and events that follow these actions.
```
wochinge, line 713: nice docstrings 🚀

rasa/core/policies/policy.py
```diff
@@ -53,8 +53,9 @@ class SupportedData(Enum):
 
     @staticmethod
     def trackers_for_policy(
-        policy: Union["Policy", Type["Policy"]], trackers: List[DialogueStateTracker]
-    ) -> List[DialogueStateTracker]:
+        policy: Union["Policy", Type["Policy"]],
+        trackers: Union[List[DialogueStateTracker], List[TrackerWithCachedStates]],
```
wochinge, line 57: `DialogueStateTracker` is the parent class of `TrackerWithCachedStates` and hence covers `TrackerWithCachedStates`. No need to change the type annotation here.

rasa/core/policies/ensemble.py
```diff
@@ -737,6 +707,64 @@ def _check_policy_for_forms_available(
...
+
+    action_fingerprints = {}
+    for k, vs in training_events.items():
+        slots = list({v.key for v in vs if isinstance(v, SlotSet)})
+        active_loops = list({v.name for v in vs if isinstance(v, ActiveLoop)})
+        action_fingerprints[k] = {
+            ACTION_FINGERPRINT_SLOTS: slots,
+            ACTION_FINGERPRINT_ACTIVE_LOOPS: active_loops,
+        }
+    return action_fingerprints
```
wochinge, line 765: Let's use a typed class instead of a dict.

wochinge: then we also don't need more constants

rasa/core/policies/rule_policy.py
```diff
@@ -395,6 +533,8 @@ def train(
         rule_trackers = [t for t in training_trackers if t.is_rule_tracker]
         if self._restrict_rules:
             self._check_rule_restriction(rule_trackers)
+        if self._check_for_contradictions:
+            self._check_rule_action_fingerprints(rule_trackers, domain)
```
wochinge, line 537: Why not calling it like this
```suggestion
            self._check_for_incomplete_rules(rule_trackers, domain)
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+                        tracker.sender_id,
+                    )
+                )
+
+        if error_messages:
+            error_messages = "\n".join(error_messages)
+            raise InvalidRule(
+                f"\nIncomplete rules found🚨\n\n{error_messages}\n"
+                f"Please note that if some slots or active loops should not be set "
+                f"during prediction you need explicitly set them to 'null' in the "
```
wochinge, line 405:
```suggestion
                f"during prediction you need to explicitly set them to 'null' in the "
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
                 f"Please use stories for these cases."
             )
 
+    @staticmethod
+    def _check_slots_fingerprint(
+        fingerprint: Dict[Text, List[Text]], state: State, featurized_slots: Set[Text]
+    ) -> Set[Text]:
+        # leave only featurized slots
+        expected_slots = set(fingerprint.get(ACTION_FINGERPRINT_SLOTS)).intersection(
```
wochinge, line 285: I think this should be done as part of the fingerprinting

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+    def _error_messages_from_fingerprints(
+        action_name: Text,
+        fingerprint_slots: Set[Text],
+        fingerprint_active_loops: Set[Text],
+        rule_name: Text,
+    ) -> List[Text]:
+        error_messages = []
+        if action_name and fingerprint_slots:
+            error_messages.append(
+                f"- the action '{action_name}' in rule '{rule_name}' does not set "
```
wochinge, line 319:
```suggestion
                f"- the action '{action_name}' in rule '{rule_name}' does not set all"
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+                f"slots, that it sets in other rules: "
+                f"'{', '.join(fingerprint_slots)}'. Please update the rule with "
+                f"an appropriate slot or if it is the last action add "
+                f"`wait_for_user_input: false` after this action."
```
wochinge, line 323:
```suggestion
                f"'wait_for_user_input: false' after this action."
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+            fingerprint_active_loops = set(
+                "null" if active_loop == SHOULD_NOT_BE_SET else active_loop
+                for active_loop in fingerprint_active_loops
+            )
+            # add action_name to active loop so that users
+            # know what to put in their rules
+            fingerprint_active_loops.add(action_name)
+
+            error_messages.append(
+                f"- the action '{action_name}' in rule '{rule_name}' doesn't set an "
```
wochinge, line 337:
```suggestion
                f"- the form '{action_name}' in rule '{rule_name}' doesn't set the "
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+                f"active_loop, that it sets in other rules: "
```
wochinge, line 338:
```suggestion
                f"'active_loop', that it sets in other rules: "
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+            # know what to put in their rules
+            fingerprint_active_loops = set(
+                "null" if active_loop == SHOULD_NOT_BE_SET else active_loop
+                for active_loop in fingerprint_active_loops
+            )
+            # add action_name to active loop so that users
+            # know what to put in their rules
+            fingerprint_active_loops.add(action_name)
+
+            error_messages.append(
```
wochinge, line 336: @b-quachtran Could you please also double check the error messages for phrasing. Thanks 🙌

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+                f"- the action '{action_name}' in rule '{rule_name}' doesn't set an "
+                f"active_loop, that it sets in other rules: "
+                f"'{', '.join(fingerprint_active_loops)}'. Please update the rule with "
+                f"an appropriate active loop or if it is the last action add "
```
wochinge, line 340:
```suggestion
                f"the appropriate 'active loop' property or if it is the last action add "
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+                f"`wait_for_user_input: false` after this action."
```
wochinge, line 341:
```suggestion
                f"'wait_for_user_input: false' after this action."
```

wochinge (commented):
I'll check the tests after my meeting

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +277,137 @@ def _check_rule_restriction(
...
+        return error_messages
+
+    def _check_rule_action_fingerprints(
+        self, rule_trackers: List[TrackerWithCachedStates], domain: Domain,
+    ) -> None:
+        import rasa.core.policies.ensemble
+
+        logger.debug("Started checking if some rules are incomplete.")
+        # we need to use only fingerprints from rules
+        rule_fingerprints = rasa.core.policies.ensemble.create_action_fingerprints(
```
wochinge, line 352: What was the reasoning behind putting these functions into to `ensemble`? To me this seems to rather fit into the `tracker` module

wochinge (commented):
tests look 👍

changelog/6845.improvement.md
Ghostvv: `action fingerprint` is smth already exists in rasa oss:
https://github.com/RasaHQ/rasa/blob/91a5b5f7ebcc50590beb0de5547b856cdd3992c2/rasa/core/policies/ensemble.py#L213

Ghostvv: I'll remove changelog

rasa/core/policies/policy.py
Ghostvv: it doesn't. `train` methods use properties that exist only in `TrackerWithCachedStates`

rasa/core/policies/ensemble.py
Ghostvv: this is an old dict, I'm not sure where it is used and how it is persisted.

rasa/core/policies/rule_policy.py
Ghostvv: because I came up with `incomplete` term afterwards 😅

rasa/core/policies/rule_policy.py
Ghostvv: fingerprinting is an old method, I don't know what effects of changing it will be

rasa/core/policies/rule_policy.py
Ghostvv: have no idea. this are the old functions that were in ensemble class

Ghostvv: I just found the new usage for them

Ghostvv:
@wochinge `action_fingerprints` is an old think. I can redesign it if you think it makes sense

rasa/core/policies/rule_policy.py
Ghostvv: moved to `trackers.py`

rasa/core/policies/rule_policy.py
Ghostvv: also I'd argue that fingerprints are independent of featurization, it's just in this concrete case we want to remove unfeaturized slots

rasa/core/policies/policy.py
wochinge: I know. The parent class is still enough as type annotation when it can be both. That serves as warning that we need an explicit `is_instance` check with the train function before using `TrackerWithCachedStates` methods

rasa/core/policies/ensemble.py
wochinge: I think yes as we can skip all these constants then.

rasa/core/policies/policy.py
Ghostvv: ok, removed it

rasa/core/policies/rule_policy.py
wochinge: Why are the fingerprints independent? They are used to check if stuff was already seen? As unfeaturized things don't matter I think it's fine to filter that out right away

rasa/core/policies/ensemble.py
Ghostvv: I reused `SLOTS` and `ACTIVE_LOOP` constants because they actually have to be coupled, so I removed these constants anyhow

rasa/core/policies/policy.py
Ghostvv: actually, it's return type that is the problem

rasa/core/policies/ensemble.py
Ghostvv: I suggest let's not over engineer it and keep a dict, it looks accurate right now

Ghostvv: also, it could happen that action doesn't set any slots or loops, so these keys are not guaranteed to be in there

docs/docs/policies.mdx
b-quachtran: Yeah, it's a list of dicts. I'm pulling it from the concertbot example.

docs/docs/policies.mdx
b-quachtran: Hmm, that's a good point. This solution should be good :thumbsup:

docs/docs/policies.mdx
b-quachtran: It solves the problem, but I can see users interpreting this as "use `wait_for_user_input` to solve conflict errors" when they should be re-examining their rule & story definitions.

Ghostvv: but this is not in the part of the docs where we recommend staff, it is in `policies.mdx` which describe capabilities of the policy. I think users should know that it is possible.
also, I don't see a problem in  "use `wait_for_user_input` to solve conflict errors", if it indeed solves them

rasa/core/policies/ensemble.py
```diff
@@ -27,22 +27,17 @@
...
+from rasa.shared.core.events import ActionExecutionRejected
 from rasa.core.exceptions import UnsupportedDialogueModelError
 from rasa.core.featurizers.tracker_featurizers import MaxHistoryTrackerFeaturizer
 from rasa.shared.nlu.interpreter import NaturalLanguageInterpreter, RegexInterpreter
 from rasa.core.policies.policy import Policy, SupportedData
 from rasa.core.policies.fallback import FallbackPolicy
 from rasa.core.policies.memoization import MemoizationPolicy, AugmentedMemoizationPolicy
 from rasa.core.policies.rule_policy import RulePolicy
-from rasa.shared.core.trackers import DialogueStateTracker
+from rasa.shared.core.trackers import DialogueStateTracker, create_action_fingerprints
```
wochinge, line 38: please refer to functions by their modules

rasa/core/policies/ensemble.py
```diff
@@ -193,7 +171,7 @@ def _emit_rule_policy_warning(
 
     def train(
         self,
-        training_trackers: List[DialogueStateTracker],
+        training_trackers: List[TrackerWithCachedStates],
```
wochinge, line 174: this causes a type clash in line 187 of `ensemble.py` as `trackers_for_policy` returns things of type `DialogueStateTracker`

rasa/core/policies/rule_policy.py
```diff
@@ -16,6 +16,7 @@
     DialogueStateTracker,
     get_active_loop_name,
     is_prev_action_listen_in_state,
+    create_action_fingerprints,
```
wochinge, line 19: please refer to the functions by their module

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +274,124 @@ def _check_rule_restriction(
                 f"Please use stories for these cases."
             )
 
+    @staticmethod
+    def _check_slots_fingerprint(
+        fingerprint: Dict[Text, List[Text]], state: State
+    ) -> Set[Text]:
+        # leave only featurized slots
```
wochinge, line 281: this comment seems outdated
```suggestion
```

rasa/core/policies/rule_policy.py
```diff
@@ -270,6 +274,124 @@ def _check_rule_restriction(
...
+    def _error_messages_from_fingerprints(
+        action_name: Text,
+        fingerprint_slots: Set[Text],
+        fingerprint_active_loops: Set[Text],
+        rule_name: Text,
+    ) -> List[Text]:
+        error_messages = []
+        if action_name and fingerprint_slots:
+            error_messages.append(
+                f"- the action '{action_name}' in rule '{rule_name}' does not set all"
```
wochinge, line 314:
```suggestion
                f"- the action '{action_name}' in rule '{rule_name}' does not set all "
```

rasa/core/policies/rule_policy.py
```diff
@@ -395,6 +517,8 @@ def train(
         rule_trackers = [t for t in training_trackers if t.is_rule_tracker]
         if self._restrict_rules:
             self._check_rule_restriction(rule_trackers)
+        if self._check_for_contradictions:
```
wochinge, line 520:
```suggestion
            self._check_rule_restriction(rule_trackers)

        if self._check_for_contradictions:
```

rasa/core/policies/ensemble.py
Ghostvv: I tried, it becomes impossibly long 🙈
this is not ambiguous function, no?

rasa/core/policies/ensemble.py
Ghostvv: exactly! that's why I returned `Union` in `trackers_for_policy`, but you told me to remove it

rasa/core/policies/ensemble.py
wochinge: it's our Python Code convention to refer to functions using their module so there is no way around it.

rasa/core/policies/policy.py
wochinge: ok, you're right. Please re-add. What a stupid problem.

rasa/shared/core/trackers.py
```diff
@@ -781,3 +783,68 @@ def is_prev_action_listen_in_state(state: State) -> bool:
...
+            that this action sets
+    """
+    events_after_actions = _find_events_after_actions(trackers)
+    if not events_after_actions:
+        return {}
+
+    # take into account only featurized slots
+    featurized_slots = {slot.name for slot in domain.slots if slot.has_features()}
+    action_fingerprints = defaultdict(dict)
+    for k, vs in events_after_actions.items():
```
wochinge, line 839:
```suggestion
    for action_name, events_after_action in events_after_actions.items():
```

rasa/core/policies/ensemble.py
wochinge: uargh. Let's readd it then 😬

Ghostvv: 🙈

rasa/core/policies/policy.py
Ghostvv: returned

Ghostvv:
@erohmensing sorry to drag you in, could you please check the formulation of errors and docs, since Ben is in different timezone?

docs/docs/policies.mdx
Ghostvv: @erohmensing what do you think?

rasa/core/policies/rule_policy.py
wochinge: cc @erohmensing as vova tagged you for the docs

rasa/shared/core/domain.py
```diff
@@ -1215,3 +1221,81 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
             The slot mapping or an empty dictionary in case no mapping was found.
         """
         return self.forms.get(form_name, {})
+
+    @staticmethod
+    def _find_events_after_actions(
```
wochinge, line 1226: Naah. I this has nothing to do with the `domain` in my opinion. Let's keep this in the `tracker` module. Just for the sake of keeping imports short we shouldn't move things to the wrong places.

Ghostvv: actually, it doesn't do anything with trackers either (that's why I moved it)

Ghostvv: I think it should go here: https://github.com/RasaHQ/rasa/blob/master/rasa/core/training/__init__.py

wochinge: I think having any code aside from imports in the `__init__.py` file is bad practice.

> actually, it doesn't do anything with trackers either (that's why I moved it)

it receives a list of trackers and goes through the conversation events of the trackers. I think it's very closely tied to trackers and their content.

Ghostvv: yes, but `trackers.py` deals with individual tracker. List of trackers is training data

wochinge: some resources on __init__.py and good practices https://docs.python-guide.org/writing/structure/#packages

Ghostvv: In my opinion it is much closer to domain than to trackers, because it creates a signature for actions which is basically part of domain

Ghostvv: there were other methods in `init.py`, totally up for moving all of them out

Ghostvv: funny enough, many of them are not used 😅

wochinge: > yes, but trackers.py deals with individual tracker. List of trackers is training data

But the resulting fingerprints are also used during prediction time (in `processor`). 

> In my opinion it is much closer to domain than to trackers, because it creates a signature for actions which is basically part of domain

But they only use the slot info from the domain. All other info is events and trackers. The only other place I can imagine is the module of `TrackerWithCachedStates` 

> there were other methods in init.py, totally up for moving all of them out

🎉  Let's do that in a different PR though. If the stuff isn't used -> feel free to delete it

Ghostvv: > But they only use the slot info from the domain. All other info is events and trackers. The only other place I can imagine is the module of TrackerWithCachedStates

the other place would be `policy.py` that's actually who deals with events list of trackers and domain

`trackers.py` don't deal with a list of trackers

Ghostvv: > But the resulting fingerprints are also used during prediction time
but not the method itself

Ghostvv: thinking about it more. I'd keep it in training, that's the one that deal with training data represented as list of trackers and that the main thing about this function

rasa/core/processor.py
```diff
@@ -533,7 +534,7 @@ def is_action_limit_reached(
         """Check whether the maximum number of predictions has been met.
 
         Args:
-            num_predictes_actions: Number of predicted actions.
+            num_predicted_actions: Number of predicted actions.
```
wochinge, line 537: I give you a ⭐ for this 😂

Ghostvv: it's pycharm

rasa/shared/core/trackers.py
```diff
@@ -1,6 +1,6 @@
 import copy
 import logging
-from collections import deque
+from collections import deque, defaultdict
```
wochinge, line 3:
```suggestion
from collections import deque
```

rasa/shared/core/trackers.py
```diff
@@ -16,6 +16,7 @@
     Union,
     FrozenSet,
     Tuple,
+    Set,
```
wochinge, line 19:
```suggestion
```

rasa/shared/core/trackers.py
```diff
@@ -40,6 +41,7 @@
     LOOP_VALIDATE,
     LOOP_REJECTED,
     TRIGGER_MESSAGE,
+    SLOTS,
```
wochinge, line 44:
```suggestion
```

wochinge (approved):
Approval for the code ⭐ ✅

wochinge:
@RasaHQ/cseteam Code freeze is ahead and there no feedback on the docs for a few days (code is ready). I suggest we merge this and if you have further improvement suggestion we can do a separate PR.

m-vdb:
@b-quachtran @erohmensing we merged this so that it makes it in the code freeze for 2.0. Can we continue the discussion on the docs part and merge any required changes before the actual release of 2.0? Thanks 🙏

b-quachtran:
@m-vdb Yep, we'll take care of the docs changes independent of the 2.0 release

PR #6855, opened 2020-09-30
strip paths from package paths

tmbo:
**Proposed changes**:
- we need to explicitly remove absolute path references to reduce the information sent


rasa/telemetry.py
```diff
@@ -558,7 +558,7 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
 
 def strip_sensitive_data_from_sentry_event(
     event: Dict[Text, Any], _unused_hint: Optional[Dict[Text, Any]] = None
-) -> Dict[Text, Any]:
+) -> Optional[Dict[Text, Any]]:
```
m-vdb, line 561: looks like the `Returns` in the docstring needs an update

rasa/telemetry.py
```diff
@@ -573,6 +573,18 @@ def strip_sensitive_data_from_sentry_event(
     for value in event.get("exception", {}).get("values", []):
         for frame in value.get("stacktrace", {}).get("frames", []):
             frame["abs_path"] = ""
+
+            if "site-packages" in frame["filename"]:
+                # drop site-packages and following slash / backslash
+                frame["filename"] = frame["filename"].split("site-packages")[-1][1:]
```
m-vdb, line 579: could split using `site-packages/` which would make it more obvious, but that's a detail

rasa/telemetry.py
```diff
@@ -573,6 +573,18 @@ def strip_sensitive_data_from_sentry_event(
...
+            if "dist-packages" in frame["filename"]:
+                # drop dist-packages and following slash / backslash
+                frame["filename"] = frame["filename"].split("dist-packages")[-1][1:]
```
m-vdb, line 582: Do you think it would still be valuable to keep `dist-packages/` and `site-packages/` in the filename? Could it influence Sentry's "App Only / Full" toggle behaviour?

m-vdb (approved):
Nice! I left a few comments, maybe you already thought about these

rasa/telemetry.py
tmbo: well, that won't work on windows due to different path separators... 😂

rasa/telemetry.py
tmbo: mhm that is a good point, yes it actually might :+1:

rasa/telemetry.py
tmbo: well, needs to work for windows as well and `\`... 😉

rasa/telemetry.py
tmbo: yes that is a good idea :+1:

rasa/telemetry.py
m-vdb: oh that's a good point!

PR #6856, opened 2020-09-30
Improve exception reporting for CLI commands

tmbo:
**Motivation**:
- We print unecessary stack traces for exceptions like failures to read a training data yaml file. This isn't helpful as we know that the issue is the syntax in the yaml file and a stack trace won't help.
- Many exceptions reported to sentry are things we expect to happen (errors in the training data format, a wrong action name, ...)

Before:
Bildschirmfoto 2020-10-01 um 11 15 22

After:
Bildschirmfoto 2020-10-01 um 11 16 52

**Proposed changes**:
- base class for all exceptions raised in Rasa Open Source that are "expected" (e.g. file format issues, missing action names, ... things we expect a user to get wrong. they will still bubble up, but are not going to create a stacktrace). `RasaOpenSourceException`
- all `RasaOpenSourceExceptions` will be caught if thrown as part of a CLI command and stack traces will not be printed by default, instead we use `print_error`
- changed `NameError` exception raised when an action can not be found to be a `ValueError` instead
- raise `InvalidYAMLFileException` instead of `YAMLError` (from ruamel) when failing to load a yaml file 
- made `InvalidDomain` a `RasaOpenSourceException`
- introduced `MissingDependencyException` as an exception raised if packages need to be installed
- added line numbers to output of YAML validation to the command line as paths on their own are hard to navigate:
```
InvalidYamlFileError: Failed to validate 'data/nlu.yml'. Please make sure the file is correct and all mandatory parameters are specified. Please take a look at the errors found during validation:
  in data/nlu.yml:2:
      Key 'ZZZZ' was not defined. Path: '/nlu/0'
```
- properly exit on ctrl+c / ctrl+d in `rasa shell nlu` command
- `RasaOpenSourceException`s are ignored by sentry

**TODO**
- `InvalidRule` is very user-unfriendly and vague, but out of this PRs scope


rasa/shared/utils/io.py
```diff
@@ -10,14 +10,18 @@
 from pathlib import Path
 from typing import Any, Text, Optional, Type, Union, List, Dict
 
+from ruamel.yaml.scanner import ScannerError
```
m-vdb, line 13: can this import be grouped with the other ones below?

rasa/shared/utils/io.py
```diff
@@ -241,7 +245,7 @@ def read_yaml(content: Text) -> Any:
 
     replace_environment_variables()
 
-    yaml_parser = yaml.YAML(typ="safe")
+    yaml_parser = yaml.YAML(typ=["safe", "rt"])
```
m-vdb, line 248: I couldn't find why it's useful in the doc. Is it worth a comment?

rasa/shared/utils/validation.py
```diff
@@ -22,23 +24,103 @@
 KEY_TRAINING_DATA_FORMAT_VERSION = "version"
 
 
-class InvalidYamlFileError(ValueError):
+class InvalidYamlFileError(ValueError, RasaOpenSourceException):
```
m-vdb, line 27: `InvalidYamlFileError` and `InvalidYamlFileException` have very similar name, I suggest we make one of them more specific

m-vdb: should we rename this one to `InvalidYamlSchemaError`?

rasa/shared/utils/validation.py
```diff
@@ -22,23 +24,103 @@
...
+            head, tail = path, ""
+
+        if head:
+            if isinstance(current, dict):
+                return self._line_number_for_path(current[head], tail) or this_line
+            elif isinstance(current, list) and head.isdigit():
+                return self._line_number_for_path(current[int(head)], tail) or this_line
+            else:
+                return this_line
+        else:
```
m-vdb, line 113: unnecessary `else` block

rasa/shared/utils/validation.py
```diff
@@ -67,28 +146,29 @@ def validate_yaml_schema(
...
         c.validate(raise_exception=True)
     except SchemaError:
         raise InvalidYamlFileError(
-            "Failed to validate yaml file. "
             "Please make sure the file is correct and all "
-            "mandatory parameters are specified; to do so, "
-            "take a look at the errors logged during "
-            "validation previous to this exception."
+            "mandatory parameters are specified. Please take a look at the errors "
+            "found during validation",
```
m-vdb, line 169: A sentence is starting with "Please" twice in a row, I'd suggest:
```suggestion
            "Please make sure the file is correct and all "
            "mandatory parameters are specified. Here are the errors "
            "found during validation",
```

rasa/telemetry.py
```diff
@@ -573,6 +574,11 @@ def strip_sensitive_data_from_sentry_event(
     for value in event.get("exception", {}).get("values", []):
         for frame in value.get("stacktrace", {}).get("frames", []):
             frame["abs_path"] = ""
+
+            if "rasa_sdk/executor.py" in frame["filename"]:
+                # this looks a lot like an exception in the SDK and hence custom code
+                # no need for us to deal with that
+                return None
```
m-vdb, line 581: looks more like a filter than stripping sensitive info, but arguable 👍

m-vdb (changes requested):
🚀 🚀 🚀  Amazing, thanks for doing these changes. I definitely think that they will improve the UX

rasa/core/policies/rule_policy.py
```diff
@@ -56,20 +57,17 @@
...
-        )
+        self.message = message
 
     def __str__(self) -> Text:
-        # return message in error colours
-        return rasa.shared.utils.io.wrap_with_color(
-            self.message, color=rasa.shared.utils.io.bcolors.FAIL
+        return self.message + (
+            f"\nYou can find more information about the usage of "
+            f"rules at {DOCS_URL_RULES}. "
```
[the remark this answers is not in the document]
tmbo: Yes I think this should become a behaviour of the `RasaOpenSourceException`. But I think it makes sense to do that separately afterwards :+1:

rasa/nlu/registry.py
```diff
@@ -110,31 +120,41 @@ def get_component_class(component_name: Text) -> Type["Component"]:
...
-            raise ModuleNotFoundError(exception_message)
+            raise ComponentNotFoundException(
+                f"Failed to load the component "
+                f"'{component_name}'. "
+                f"{exception_message} Either your "
+                f"pipeline configuration contain an error "
+                f"or the module you are trying to import "
+                f"is broken (e.g. the module is trying "
+                f"to import a package that is not "
+                f"installed). {traceback.format_exc()}"
```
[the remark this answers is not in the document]
tmbo: yes, there is a scenary where the stacktrace is very helpful: it might be the case that you configured the component correctly (using the right module path) but there is an importerror inside of your module, e.g. your class imports a package that does not exist. We can't figure this out on our own, so showing the stacktrace here will allow the user to see where in their custom code the error happend.

rasa/nlu/run.py
```diff
@@ -24,7 +24,12 @@ def run_cmdline(
...
-        message = input().strip()
+        try:
+            message = input().strip()
+        except (EOFError, KeyboardInterrupt):
+            print_info("Wrapping up command line chat...")
+            break
+
         if message.startswith(INTENT_MESSAGE_PREFIX):
             result = rasa.utils.common.run_in_loop(regex_interpreter.parse(message))
         else:
```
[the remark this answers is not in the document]
tmbo: that is a very good point. I've removed them (from this file but also from a couple other scripts). this has been deprecated for a number of versions now so save to remove :+1:

rasa/shared/core/domain.py
```diff
@@ -154,6 +156,7 @@ def from_yaml(cls, yaml: Text, original_filename: Text = "") -> "Domain":
                 yaml, rasa.shared.constants.DOMAIN_SCHEMA_FILE
             )
         except rasa.shared.utils.validation.InvalidYamlFileError as e:
+            e.filename = original_filename
```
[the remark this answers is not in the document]
tmbo: what do you mean? the `InvalidYamlFileError` exception handles this when converting itself to a string (it will not add a filename if it is not set or set to "")

rasa/shared/exceptions.py
```diff
@@ -1,10 +1,34 @@
+from typing import Text
+
+
 class RasaException(Exception):
     """Base exception class for all errors raised by Rasa."""
 
 
-class RasaCoreException(RasaException):
+class RasaOpenSourceException(RasaException):
```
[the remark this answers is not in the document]
tmbo: yes that is a good point, I'll reuse that one 👍

rasa/shared/utils/validation.py
tmbo: :+1: renamed them to `YamlSyntaxException` and `YamlValidationException`

rasa/telemetry.py
tmbo: well, but the intent is to remove sensitive info 😉

tmbo:
thanks a lot for the quick review super helpful to get this merged in time ⭐

rasa/shared/core/domain.py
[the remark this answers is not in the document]
m-vdb: ok gotcha, so let's keep it 👍

rasa/shared/utils/validation.py
m-vdb: great, that fixes it 👍

m-vdb (approved):
Great 💯 I think it's mergeable once the tests go green 🟢

Issue #6820, opened 2020-09-28
Automatically migrate MappingPolicy + FallbackPolicies


wochinge:
**This issue may be split in three parts**

**Motivation**
* [this feedback](https://www.notion.so/rasa/Feedback-on-2-0-4574eaacc0764967ac0b8b82fbc72904#e17d53c4dbb44ba1844b412b8ebaa782)
* there is actually no need to manually migrate `FallbackPolicy`, `TwoStageFallbackPolicy` or `MappingPolicy`. We can migrate all of them automatically. 

**Ideas**
* check out the migration guide and automate it as part of some CLI command (`rasa data convert`?)
* ? 

**Definition of Done**
- [ ] Feature implemented
- [ ] Migration guide updated

wochinge:
@degiz Do you think we integrate it into `rasa data convert`? There is no `rasa migrate`, right?

wochinge:
It might not be possible to migrate the `TwoStageFallbackPolicy` automatically as users might have customized the `deny_suggestion_intent_name` intent and we can't migrate custom actions of theirs which are using this intent.

degiz:
> There is no rasa migrate, right?

We were talking about possibly introducing `rasa migrate`, but wanted to use it to migrate the whole project from to `2.0`. So back then, when we only wanted to convert training data, we didn't do that.

Now, if you have an idea of how to do that, that makes sense! But Do you think we can do that before next week?

wochinge:
I mean it's just for these polices above. We can't migrate the whole thing (e.g. forms).
The alternative based on the feedback would be add a default rule for the two-stage-fallback but I don't think this is ideal as it's not transparent and users might wanna change the rule for the two stage fallback somehow. 

We could also do a `rasa data convert config` which would be a bit more selective than introducing a `rasa migrate`.

> But Do you think we can do that before next week?

It's blocking feedback and directly relevant to our okrs (smooth transition). I think it's definitely doable with a bit of force.

cc @m-vdb @tmbo Wdyt?

m-vdb:
> It might not be possible to migrate the TwoStageFallbackPolicy automatically as users might have customized the deny_suggestion_intent_name intent and we can't migrate custom actions of theirs which are using this intent.

If we can get a command that simplifies the migration for a good share of users that's great. What about something along those lines:
1. show a very visible warning about the fact that the rule is missing (can we detect it?)
2. have a command that converts the config to 2.0 format
[...]

PR #6857, opened 2020-09-30
function to automatically migrate the MappingPolicy

m-vdb:
**Proposed changes**:
- relates to #6820
- needs https://github.com/RasaHQ/rasa/pull/6854

### DIFF
--- a/rasa/cli/data.py
+++ b/rasa/cli/data.py
@@ -405,7 +405,13 @@ def _migrate_model_config(args: argparse.Namespace):
     # 1. Migrate FallbackPolicy
     # 2. Add telemetry
     # 3. Auto backup file?
-    rasa.core.config.migrate_mapping_policy_to_rules(model_configuration, domain, rules)
+    (
+        model_configuration,
+        domain,
+        rules,
+    ) = rasa.core.config.migrate_mapping_policy_to_rules(
+        model_configuration, domain, rules
+    )
 
     # dump config, domain, rules
     rasa.shared.utils.io.write_yaml(model_configuration, configuration_file)
--- a/rasa/cli/data.py
+++ b/rasa/cli/data.py
@@ -27,6 +27,7 @@ from rasa.utils.converter import TrainingDataConverter
 from rasa.validator import Validator
 from rasa.shared.core.domain import Domain, InvalidDomain
 import rasa.shared.utils.io
+import rasa.core.config
 from rasa.core.policies.form_policy import FormPolicy
 from rasa.core.policies.fallback import FallbackPolicy
 from rasa.core.policies.two_stage_fallback import TwoStageFallbackPolicy
@@ -398,12 +399,18 @@ def _migrate_model_config(args: argparse.Namespace):
     domain = _get_domain(domain_file)
 
     rule_output_file = _get_rules_path(Path(args.out))
+    rules = []
 
     # TODO:
-    # 1. Migrate
-    # 2. Dump config
-    # 3. Dump rules
-    # 4. Add telemetry
+    # 1. Migrate FallbackPolicy
+    # 2. Add telemetry
+    # 3. Auto backup file?
+    rasa.core.config.migrate_mapping_policy_to_rules(model_configuration, domain, rules)
+
+    # dump config, domain, rules
+    rasa.shared.utils.io.write_yaml(model_configuration, configuration_file)
+    domain.persist_clean(domain_file)
+    rasa.shared.utils.io.write_yaml({"rules": rules}, rule_output_file)
 
 
 def _get_configuration(path: Path) -> Dict:
@@ -446,7 +453,7 @@ def _assert_nlu_pipeline_given(config: Dict, policy_names: List[Text]) -> None:
         )
 
 
-def _assert_two_stage_fallack_policy_is_migratable(config: Dict,):
+def _assert_two_stage_fallack_policy_is_migratable(config: Dict):
     two_stage_fallback_config = next(
         (
             policy_config
--- a/rasa/core/config.py
+++ b/rasa/core/config.py
@@ -4,6 +4,8 @@ from typing import Any, Optional, Text, List, Dict, Union
 
 import rasa.shared.utils.io
 import rasa.utils.io
+from rasa.core.policies.mapping_policy import MappingPolicy
+from rasa.core.policies.rule_policy import RulePolicy
 
 if typing.TYPE_CHECKING:
     from rasa.core.policies.policy import Policy
@@ -44,9 +46,9 @@ def migrate_mapping_policy_to_rules(
     has_rule_policy = False
 
     for policy in policies:
-        if policy.get("name") == "MappingPolicy":
+        if policy.get("name") == MappingPolicy.__name__:
             has_mapping_policy = True
-        if policy.get("name") == "RulePolicy":
+        if policy.get("name") == RulePolicy.__name__:
             has_rule_policy = True
 
     if not has_mapping_policy:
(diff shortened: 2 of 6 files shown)


m-vdb:
cc @wochinge

wochinge:
@m-vdb Can you please rebase to https://github.com/RasaHQ/rasa/pull/6854/ ?

m-vdb:
yes sure if it's ready to be rebased to now

rasa/core/config.py
```diff
@@ -27,3 +28,45 @@ def load(config_file: Optional[Union[Text, Dict]]) -> List["Policy"]:
...
+    by updating the config, domain and generating rules.
+
+    This function modifies the config, the domain and the rules in place.
+    """
+    policies = config.get("policies", [])
+    has_mapping_policy = False
+    has_rule_policy = False
+
+    for policy in policies:
+        if policy.get("name") == "MappingPolicy":
```
wochinge, line 47:
```suggestion
        if policy.get("name") == MappingPolicy.__name__:
```

rasa/core/config.py
```diff
@@ -27,3 +28,45 @@ def load(config_file: Optional[Union[Text, Dict]]) -> List["Policy"]:
...
+            has_mapping_policy = True
+        if policy.get("name") == "RulePolicy":
```
wochinge, line 49:
```suggestion
        if policy.get("name") == RulePolicy.__name__:
```

rasa/core/config.py
```diff
@@ -27,3 +28,45 @@ def load(config_file: Optional[Union[Text, Dict]]) -> List["Policy"]:
...
+
+    has_one_triggered_action = False
+    for intent, properties in domain.intent_properties.items():
+        # remove triggers from intents, if any
+        triggered_action = properties.pop("triggers", None)
+        if triggered_action:
+            has_one_triggered_action = True
+            rules.append(
+                {
+                    "rule": f"Rule to map `{intent}` intent (automatic conversion)",
```
wochinge, line 63: How about including what it's mapped to?

rasa/core/config.py
```diff
@@ -27,3 +28,45 @@ def load(config_file: Optional[Union[Text, Dict]]) -> List["Policy"]:
...
+                    "steps": [{"intent": intent}, {"action": triggered_action},],
+                }
+            )
+
+    # finally update the policies
+    policies = [policy for policy in policies if policy.get("name") != "MappingPolicy"]
```
wochinge, line 69:
```suggestion
    policies = [policy for policy in policies if policy.get("name") != MappingPolicy.__name__]
```

rasa/core/config.py
```diff
@@ -27,3 +28,45 @@ def load(config_file: Optional[Union[Text, Dict]]) -> List["Policy"]:
...
+    if has_one_triggered_action and not has_rule_policy:
+        policies.append({"name": "RulePolicy"})
```
wochinge, line 71:
```suggestion
        policies.append({"name": RulePolicy.__name__})
```

tests/core/test_config.py
```diff
@@ -69,3 +72,58 @@ def check_form(p):
...
+def test_migrate_mapping_policy_to_rules(
+    config: Dict[Text, Any], domain_dict: Dict[Text, Any], expected: Dict[Text, Any],
+):
+    rules = []
+    domain = Domain.from_dict(domain_dict)
+    core_config.migrate_mapping_policy_to_rules(config, domain, rules)
+
+    assert config == expected["config"]
+    assert domain.cleaned_domain()["intents"] == expected["domain_intents"]
+    assert rules == expected["rules"]
```
wochinge, line 129: optional: also testing for the number of rules

tests/core/test_config.py
```diff
@@ -69,3 +72,58 @@ def check_form(p):
...
+                }
+            ],
+        },
+    ),
+]
+
+
+@pytest.mark.parametrize("config,domain_dict,expected", TEST_MIGRATED_MAPPING_POLICIES)
+def test_migrate_mapping_policy_to_rules(
+    config: Dict[Text, Any], domain_dict: Dict[Text, Any], expected: Dict[Text, Any],
```
wochinge, line 121: optional: feels a bit clearer
```suggestion
    config: Dict[Text, Any], domain_dict: Dict[Text, Any], expected_config: Dict[Text, Any],
```

rasa/cli/data.py
```diff
@@ -398,12 +399,18 @@ def _migrate_model_config(args: argparse.Namespace):
...
 
     # TODO:
-    # 1. Migrate
-    # 2. Dump config
-    # 3. Dump rules
-    # 4. Add telemetry
+    # 1. Migrate FallbackPolicy
+    # 2. Add telemetry
+    # 3. Auto backup file?
+    rasa.core.config.migrate_mapping_policy_to_rules(model_configuration, domain, rules)
```
wochinge, line 408: optional: Not a blocker, but I'd prefer it if this returns the changed the things instead of changing it inplace. It's somewhat stylistic as it comes from a functional mindset, but I think it would also be clearer because the reader doesn't have to checkout `migrate_mapping_policy_to_rules` to see what it might change.

wochinge (approved):
Thanks for your help 🙌 Slick testing and simple code 💯

rasa/cli/data.py
m-vdb: I considered both approaches and what made the cut for me was because I didn't know how to simply make an update to the domain without modifying it in place. If there is a simple why of copying and making changes to it, I'm happy to update the code. It led to that decision because I couldn't have some objects modified in place and other objects copied.

wochinge: I think that returning it would still make the inplace change more obvious. To avoid the inplace thing:  You can copy the domain by converting it to a `Dict`, `deepcopy` it, and `Domain.load` it again

rasa/core/config.py
wochinge, line 63: I'm currently at a similar place for the fallback rules: Using `DialogueStateTracker.from_events` feels more robust as the training data keys / format might change. Using pure yaml somewhat mixes the knowledge of the training data format into this module.

tests/core/test_config.py
m-vdb: I'll do `expected_results` because `expected_config["config"]` looks weird

tests/core/test_config.py
m-vdb: do you mean asserting that `len(rules) == expected["rules_count"]` or do you mean adding a new test config that should generate 2 rules or more?

rasa/core/config.py
wochinge: Or using `YAMLStoryReader.read_from_string` seems a good middle way as it validates the format

rasa/cli/data.py
m-vdb: sounds a bit painful - would we be ok adding the `__deepcopy__()` method to the Domain class?

```python
    def __deepcopy__(self, memo: Dict[int, Any]) -> "Domain":
        """
        This dunder method makes it possible to use deepcopy directly
        on a `Domain` instance.

        See https://docs.python.org/3/library/copy.html#copy.deepcopy
        for more implementation
        """
        domain_dict = self.as_dict()
        return self.__class__.from_dict(copy.deepcopy(domain_dict, memo))
```

Then it would make it possible to do `copy.deepcopy(domain)` easily, and make sure the domain is always deepcopied the right way

rasa/core/config.py
m-vdb: I'm not super familiar with the tracker. If you have an example to share that'd be simpler for me to understand 😄

rasa/cli/data.py
wochinge: yes please

tests/core/test_config.py
wochinge: > do you mean asserting that len(rules) == expected["rules_count"]

yes

rasa/core/config.py
wochinge: That's how I solved it now: https://github.com/RasaHQ/rasa/pull/6862/files#diff-ec2e0e4e9cc6048d7680bf9ccfe289beR115-R125 wdyt?

m-vdb: I don't think it's ideal but it's at least an improvement with the schema validation. It would be rad if we had data classes on which we could rely, and which the rules would be automatically loaded into

wochinge: We have data classes (`DialogueStateTracker`, `StoryStep`). Maybe their initializers are not ideal. I think defining the rule in yaml, parsing it and then adding is a good compromise

wochinge:
@m-vdb Feel free to merge this. I can consolidate things when in my fallback migration PR.

rasa/core/config.py
m-vdb: but there is no "Rule" and "RuleStep" dataclass, right? (and by data class, I mean https://docs.python.org/3/library/dataclasses.html)

m-vdb: I'll let you consolidate this in your PR then

m-vdb:
Merging. Tests pass now, lint is failing because of `rasa-migrate-config` branch

Issue #6861, opened 2020-10-01
Rules: One user message limit breaks handling of unhappy paths


N-Olbert:
**Rasa version**: 2.0rc3

Currently (changed in rc3 by #6607 because of #6557) a rule is not allowed to contain more than one user message/utterance action.

Anyways in practice there are cases where more user messages are useful, as also stated in the official [documentation](https://rasa.com/docs/rasa/forms#writing-stories--rules-for-unhappy-form-paths):
```
rules:
- rule: Example of an unhappy path
  condition:
  # Condition that form is active.
  - active_loop: restaurant_form
  steps:
  - intent: stop
  - action: utter_ask_continue
  - intent: stop
  - action: action_deactivate_loop
  - active_loop: null
```

Current error for this offical example: 

> Found rules 'Example of an unhappy path' that contain more than 1 user message. Rules are not meant to hardcode a state machine. Please use stories for these cases.

Suggestion: Increase the number of allowed user utterances to 2. This still prevents building (highly) complex state machines with rules, but allows common unhappy path handling with confirmation. 
Forcing the user to set *restrict_rules* to false does not seem to be a valid option, because according to the documentation it seems to be dangerous (I doubt so :D) and requires manual effort.

sara-tagger:
Thanks for raising this issue, @Ghostvv will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

Ghostvv:
thank you for spotting that, we'll update the docs.
You can convert unhappy path into 2 rules:
```
rules:
- rule: ask continue
  condition:
  # Condition that form is active.
  - active_loop: restaurant_form
  steps:
  - intent: stop
  - action: utter_ask_continue

- rule: stop form
  condition:
  # Condition that form is active.
  - active_loop: restaurant_form
  steps:
  - action: utter_ask_continue
  - intent: stop
  - action: action_deactivate_loop
  - active_loop: null
```

N-Olbert:
Hi @Ghostvv ,

first of all thanks for the quick reply.

Anyways, after some rethinking I think this replacement is not exactly the same.
Think about (hypothetical, I haven’t tested it in reality) stories like:


```
# story 1
  - active_loop: restaurant_form
  - intent: explain
  - action: action_explain_current_slot
  - action: utter_ask_continue
  - intent: confirm
  - action: restaurant_form
[...]

PR #6866, opened 2020-10-01
fix multi intent unhappy path in docs

Ghostvv:
**Proposed changes**:
- fix multi intent unhappy path in docs
- fixes https://github.com/RasaHQ/rasa/issues/6861

### DIFF
--- a/docs/docs/forms.mdx
+++ b/docs/docs/forms.mdx
@@ -254,13 +254,19 @@ the form and reset the requested slot. An example story of such conversation cou
 look as follows:
 
 ```yaml-rasa
-rules:
-- rule: Example of an unhappy path
+- rule: Example of an unhappy path: ask if the form should continue
   condition:
   # Condition that form is active.
   - active_loop: restaurant_form
   steps:
   - intent: stop
+  - action: utter_ask_continue
+
+- rule: Example of an unhappy path: stop the form
+  condition:
+  # Condition that form is active.
+  - active_loop: restaurant_form
+  steps:
   - action: utter_ask_continue
   - intent: stop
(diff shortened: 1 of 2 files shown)


erohmensing:
Akela wins, gonna make it a story for less confusing 😬

Ghostvv:
@erohmensing should I close the PR then?

erohmensing (approved):

erohmensing:
@Ghostvv looks fine right?

docs/docs/forms.mdx
```diff
@@ -251,18 +252,19 @@ rules:
...
 the form and reset the requested slot. An example story of such conversation could
 look as follows:
 
 ```yaml-rasa
-rules:
-- rule: Example of an unhappy path
-  condition:
-  # Condition that form is active.
-  - active_loop: restaurant_form
+- story: User interrupts the form and doesn't want to continue
```
Ghostvv, line 263: do we need key `stories:` above

docs/docs/forms.mdx
```diff
@@ -251,18 +252,19 @@ rules:
...
   steps:
+  - intent: request_restaurant
+  - action: restaurant_form
+  - active_loop: restaurant_form
```
Ghostvv, line 267: the problem with this story is that it'll not work if there was `chitchat` before `stop`

tmbo:
should this target 2.x?

docs/docs/forms.mdx
erohmensing: I think that's fine, here we're only trying to show what to do if the user wants to stop, not handle all cases

PR #6876, opened 2020-10-01
Remaining migration polishing

wochinge:
**Proposed changes**:
- add telemetry
- backup files before overriding them
- add validations if we actually need to migrate stuff
- add more prints


rasa/cli/data.py
```diff
@@ -410,27 +418,35 @@ def _migrate_model_config(args: argparse.Namespace) -> None:
...
-    # 3. Auto backup file?
     (
         model_configuration,
         domain,
         new_rules,
     ) = rasa.core.config.migrate_mapping_policy_to_rules(model_configuration, domain)
 
+    if new_rules:
+        _backup(domain_file)
+        domain.persist_clean(domain_file)
```
m-vdb, line 429: why not do this at the end with the config?

rasa/cli/data.py
```diff
@@ -410,27 +418,35 @@ def _migrate_model_config(args: argparse.Namespace) -> None:
...
     if new_rules:
+        _backup(configuration_file)
+        rasa.shared.utils.io.write_yaml(model_configuration, configuration_file)
         _dump_rules(rule_output_file, new_rules)
 
+    telemetry.track_data_convert("yaml", "config")
+
+    rasa.shared.utils.cli.print_success(
+        f"Finished migrating your policy configuration 🎉.\n"
+        f"The migration generated {len(new_rules)} which were added to "
```
m-vdb, line 446:
```suggestion
        f"The migration generated {len(new_rules)} rule(s) which were added to "
```

m-vdb: but ideally we would pluralise the "(s)" and the "was/were"

rasa/cli/data.py
```diff
@@ -410,27 +418,35 @@ def _migrate_model_config(args: argparse.Namespace) -> None:
...
+        f"'{rule_output_file}'"
```
m-vdb, line 447:
```suggestion
        f"'{rule_output_file}'."
```

m-vdb (approved):
Looks good!

rasa/cli/data.py
wochinge: this saved me one additional check 😀Added the check and moved this to the end of the function so all dumping things are at the end.

PR #6878, opened 2020-10-01
Update tensorflow-hub and tensorflow-probability

tabergma:
**Proposed changes**:
- Update `tensorflow-hub` to `0.9.0`
- Update `tensorflow-probability` to `0.11.1`


tabergma:
/modeltest
```
include:
 - dataset: ["all"]
   config: ["all"]
```

tabergma:
/modeltest

```yml
include:
 - dataset: ["all"]
   config: ["all"]

```

wochinge (approved):
Does it make sense to use `>=` and `<=` to also work with the old versions or was there a breaking change?

dakshvar22:
There wasn't a breaking change but we wanted tensorflow deps to be updated to the latest version before 2.0 release.
Do you suggest including the older version in a range in that case?

wochinge:
> Do you suggest including the older version in a range in that case?

Imo a bigger range of compatible versions is better for our users as they might have other things installed. However, this might be slightly different for tensorflow things so up to you.

dakshvar22:
I think for tensorflow related deps it's better to pin them since they tend to be fragile with each other. For example, `tensorflow-text==2.2` may not be compatible with `tensorflow==2.3` and then finding that combination is hard from the ranges specified.

PR #6877, opened 2020-10-01
Update business-logic.mdx to specify form as dictionary not a list

Robsteranium:
**Proposed changes**:

Configuring the `domain.yml` as originally specified leads to the following error on training:
> rasa/shared/utils/io.py:87: UserWarning: The `forms` section in the domain needs to contain a dictionary. Instead found an object of type '<class 'list'>'.
>  More info at https://rasa.com/docs/rasa/forms

This change resolves the problem.

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [X] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6594, opened 2020-09-08
Change forms in domain.yml to be dicts

ricwo:
**Proposed changes**:
- fixes #6535 

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [x] updated the documentation
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6881, opened 2020-10-02
update form schema to map

erohmensing:
**Proposed changes**:
- update schema to catch forms being provided as lists (see https://github.com/RasaHQ/rasa/pull/6877)


ricwo:
they can be lists or dicts, hence type `any` in the schema. see https://github.com/RasaHQ/rasa/pull/6594

erohmensing:
Ah hmmm interesting. That doesn't seem to be implied by the error message: 
```
rasa/shared/utils/io.py:87: UserWarning: The forms section in the domain needs to contain a dictionary. Instead found an object of type '<class 'list'>'.
```
but I can try providing a list to see what happens

erohmensing:
I'll assume that we only expect people to run into this when they're trying to use the dict type, so the error message makes sense in that case, closing this

Thanks!

PR #6883, opened 2020-10-02
Allow old school rule-like policies with RulePolicy

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6820#issuecomment-702246054
- make it possible to use the `RulePolicy` in conjunction with the old, deprecated "rule-style" policies. Motivation: As a user I want to migrate the assistant step by step. The procedure is roughly the following
   1. Migrate one policy
   2. Train assistant
   3. test
   4. Migrate next policy
  This was not possible without this PR as the training would have thrown an exception about the combined usage. This exception is now relaxed to a warning.
- give the `RulePolicy` a higher priority to make sure it can be used in conjunction with `MappingPolicy` and co
- change check for `FormPolicy` in `rasa data convert config` to a warning

### DIFF
--- a/docs/docs/policies.mdx
+++ b/docs/docs/policies.mdx
@@ -52,7 +52,7 @@ and Rule Policies might both predict with confidence 1), the priority of the
 policies is considered. Rasa policies have default priorities that are set to ensure the
 expected outcome in the case of a tie. They look like this, where higher numbers have higher priority:
 
-5. `RulePolicy`
+6. `RulePolicy`
 
 3. `MemoizationPolicy` and `AugmentedMemoizationPolicy`
 
--- a/rasa/cli/data.py
+++ b/rasa/cli/data.py
@@ -457,11 +457,13 @@ def _get_configuration(path: Path) -> Dict:
 
     policy_names = [p.get("name") for p in config.get("policies", [])]
 
-    _assert_no_form_policy_present(policy_names)
     _assert_config_needs_migration(policy_names)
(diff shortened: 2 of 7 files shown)


wochinge:
I'll merge this directly into master. This is just targeted to `rasa-migrate-config` to keep the diff small.

m-vdb (commented):
LGTM 👍  but I think it's worth a review from someone who knows more about the Rule Policy

docs/docs/policies.mdx
```diff
@@ -52,7 +52,7 @@ and Rule Policies might both predict with confidence 1), the priority of the
 policies is considered. Rasa policies have default priorities that are set to ensure the
 expected outcome in the case of a tie. They look like this, where higher numbers have higher priority:
 
-5. `RulePolicy`
+6. `RulePolicy`
```
Ghostvv, line 55: I think these numbers will be confusing for new 2.0 users who are not aware of old policies

Ghostvv (approved):

docs/docs/policies.mdx
wochinge: @erohmensing I think Vova has a point. Why did we remove the deprecated polices from this? I'm gonna merge my change anyway as this outside of the scope this PR imo.

wochinge:
I thought `rasa-migrate-config` would be long merged but it isn't. Merging everything together now to handle only 2 prs instead of 3

PR #6885, opened 2020-10-02
Test stories in YAML format can now have user uterances

### LINKED ISSUE #6638: Post-RC1 Training data format items
- [x] `rasa train nlu` and Domain #6635 
- [x] make `StoryReader::read_from_file` synchronous #6641
- [x] Converter tool not working for E2E stories #6716 
- [x] Test stories don't work if they aren't in the end-to-end format #6732 
- [x] StoryMarkdownToYamlConverter fails to convert entity annotation in e2e story #6867

### LINKED ISSUE #6732: Test stories don't work if they aren't in the end-to-end format
Part of #6638 

**Rasa version**: 2.0.0rc1

Test stories don't work if they aren't in the end-to-end format, even if they are still valid story files e.g.

```yaml
version: "2.0"
stories:
- story: prompt for getting started + confirm
  steps:
  - intent: get_started_step1
  - action: action_greet_user
  - intent: affirm
  - action: utter_getstarted
  - action: utter_first_bot_with_rasa
```

This story will fail because the intent `get_started_step1` gets interpreted as a message instead of an intent. The output file `failed_test_stories.yml` shows this:

```yaml
version: "2.0"
stories:
- story: prompt for getting started + confirm (/var/folders/wc/lzjxcb8j28b2cvrwl9jv9k_h0000gn/T/tmpebavhe42/c8b65dd1df414dc4ae9ec0824829f728_test_stories_converted.yml)
  steps:
  - action: action_listen
  - intent: get_started_step1  # predicted: enter_data: get_started_step[1](number)
    user: |-
      get_started_step1
  - action: action_greet_user
  - action: action_listen
  - intent: affirm
    user: |-
      affirm
  - action: utter_getstarted
  - action: utter_first_bot_with_rasa
  - action: action_listen
```

degiz:
Fixes #6732

**Proposed changes**:
- Fixed existing bug with YAML test stories
- Added one more NLU training example for the initial project


rasa/core/test.py
```diff
@@ -539,7 +539,15 @@ async def _predict_tracker_actions(
             num_predicted_actions += 1
 
         elif use_e2e and isinstance(event, UserUttered):
-            predicted = await processor.parse_message(UserMessage(event.text))
+            # This means that user utterance didn't have a user message, only intent,
+            # so we can skip the NLU part and take the parse data directly.
+            # Indirectly that means that the test story was in YAML format.
+            if not event.text:
```
wochinge, line 545: I think we should print a warning in this case. Doesn't make sense to have labeled data without text in the tests. Might be even worth to throw an exception at the user

degiz: Imagine a situation when there was a button with intents - there's no user utterance, only intent.
These stories are still valid and should be testable.

wochinge: still. Then they should have text `/greet` or so.

wochinge (approved):

rasa/core/test.py
wochinge: I think more feedback for the user would be preferable, but up to you.

wochinge:
great work with test btw 👍

rasa/core/test.py
degiz: Hm, not throwing an exception is the whole point here, I wanted to make this test story:
```yaml
stories:
- story: happy path 1
  steps:
  - intent: greet
  - action: utter_greet
  - intent: mood_great
  - action: utter_happy
```
.. a valid one. Otherwise they will look like this:
```yaml
stories:
- story: happy path 1
  steps:
  - intent: greet
     user: /greet
  - action: utter_greet
  - intent: mood_great
     user: /mood_great
  - action: utter_happy

```
.. which takes away advantages of using explicit `intent` key.

degiz: @akelad wdyt?

Issue #6819, opened 2020-09-28
Custom action class for validating slots with one custom action


wochinge:
**This is actually an SDK issue but I added it here so we can add it to the milestone.**

**Description of Problem**:
Docs specify that all slot validations for a given form is happening as a single function, which is confusing because validation is always unique for each slot. Users will still have to define separate validation functions, but they'll also have to write the logic inside run() for calling each validate method. 

Also see the long discussions [here](https://www.notion.so/rasa/Feedback-on-2-0-4574eaacc0764967ac0b8b82fbc72904#ac27f8bae79e414e8023cdeb70283581)
 
**Overview of the Solution**:
Some magic class ala 

```python
class AbstractFormValidatorAction(Action):

  def run(...):
    slots_to_validate = tracker.form_slots_to_validate()
    for slot_name, value in slots_to_validate.items():
       function_name = f"validate_{slot_name}"
       fn = getattr(self, function_name)
       fn(value, tracker, bla)
```

**Definition of Done**:
- [ ] Implementation
- [ ] Updated Rasa Open Source docs

wochinge:
Please find a better name for `AbstractFormValidatorAction`. That's just a placeholder 😂

PR #6887, opened 2020-10-02
Form validation doc update

alwx:
**Proposed changes**:
- doc changes for #6819 
- depends on https://github.com/RasaHQ/rasa-sdk/pull/288

### DIFF
--- a/changelog/6819.doc.md
+++ b/changelog/6819.doc.md
@@ -1,1 +1,2 @@
-Update "Validating Form Input" section to include details about the usage of `FormValidationAction` class.
+Update "Validating Form Input" section to include details about how `FormValidationAction` class makes 
+it easier to validate form slots in custom actions and how to use it.
--- a/changelog/6819.doc.md
+++ b/changelog/6819.doc.md
@@ -1,0 +1,1 @@
+Update "Validating Form Input" section to include details about the usage of `FormValidationAction` class.
--- a/changelog/6819.doc.md
+++ b/changelog/6819.doc.md
@@ -1,2 +1,2 @@
-Update "Validating Form Input" section to include details about how `FormValidationAction` class makes 
-it easier to validate form slots in custom actions and how to use it.
+Update ["Validating Form Input" section](./forms.mdx#validating-form-input) to include details about 
+how `FormValidationAction` class makes it easier to validate form slots in custom actions and how to use it.
--- a/docs/docs/forms.mdx
+++ b/docs/docs/forms.mdx
@@ -344,7 +344,7 @@ class ValidateRestaurantForm(FormValidationAction):
             return {"cuisine": None}
 ```
 
-You can also extend `Action` class and retrieve extracted slots with `tracker.slots_to_validate`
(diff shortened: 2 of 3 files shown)


docs/docs/forms.mdx
```diff
@@ -297,55 +297,37 @@ actions:
...
-            else:
-                # Return a `SlotSet` event with value `None` to indicate that this
-                # slot still needs to be filled.
-                validation_events.append(SlotSet(slot_name, None))
-
-        return validation_events
-
-    def is_valid(slot_value: Any) -> bool:
-        # Implementation of the validate function.
+        return [SlotSet("slot1", "value")]
```
wochinge, line 330: Can we please give an example of a real validation here so that users see what to do in case the slot is valid and what to in case the slot is invalid?

changelog/6819.doc.md
```diff
@@ -0,0 +1 @@
+Update "Validating Form Input" section to include details about the usage of `FormValidationAction` class.
```
wochinge, line 1: Let's include a sentence what this function can be used for ("makes it easier to validate form slots in custom actions")

wochinge (commented):
Thanks 💯 

We also need to update the `migration-guide` as there is also a section on `Migrating from the Form Policy` which also shows how to implement the validation.

docs/docs/forms.mdx
wochinge: just realized that this is different to your implementation (you return a dict there).

docs/docs/forms.mdx
```diff
@@ -297,55 +297,51 @@ actions:
 - validate_restaurant_form
 ```
 
-When the form is executed it will run your custom action. In your custom action
-you can either
+When the form is executed it will run your custom action.
 
-- validate already extracted slots. You can retrieve them from the tracker by running
-  `tracker.form_slots_to_validate`.
```
wochinge, line 304: can we at least mention this in a note or at the end to let them know that they can use this to fully customize the validation? We are already seeing people who are trying to deactivate / activate forms for their validation methods and I think they'd be disappointed if they don't see a way to this with a regular `Action`

alwx: I mentioned that there is `tracker.form_slots_to_validate` and it can be used to customize the validation process in the end of the section.

docs/docs/forms.mdx
```diff
@@ -297,57 +297,56 @@ actions:
...
+        if slot_value.lower() in self.cuisine_db():
+            # validation succeeded, set the value of the "cuisine" slot to value
+            return {"cuisine": slot_value}
+        else:
+            # validation failed, set this slot to None so that the
+            # user will be asked for the slot again
+            return {"cuisine": None}
 ```
 
+You can also extend `Action` class and retrieve extracted slots with `tracker.slots_to_validate`
```
wochinge, line 347:
```suggestion
You can also extend the `Action` class and retrieve extracted slots with `tracker.slots_to_validate`
```

wochinge (approved):
🥇

PR #6890, opened 2020-10-03
Fixed bug with conversion of MD test stories

### LINKED ISSUE #6867: StoryMarkdownToYamlConverter fails to convert entity annotation in e2e story
**Rasa version**: 2.0.0rc3

**Python version**: 3.7.9

**Operating system** (windows, osx, ...): OSX

**Issue**:
When converting this e2e markdown story into yaml:
```markdown
## story_id
* enter_data: [two]{"entity": "number", "value": 2} plus [two]{"entity": "number", "value": 2}
```
the entities get muffled and I end up with:
```yaml
version: "2.0"
stories:
- story: story_id
  steps:
  - intent: 'enter_data: [two]'
```

```
/Users/sams/rasa/rasa/shared/utils/io.py:87: UserWarning: Failed to parse arguments in line 'enter_data: [two]{"entity": "number", "value": 2} plus [two]{"entity": "number", "value": 2}'. Failed to decode parameters as a json object. Make sure the intent is followed by a proper json object. Error: Extra data: line 1 column 34 (char 33)
```

```python
loop = asyncio.new_event_loop()
asyncio.set_event_loop(loop)
result = loop.run_until_complete(StoryMarkdownToYamlConverter.convert_and_write(Path("tmp_stories.md"), Path(".")))
```

```yml

```

```yml

```

degiz:
Fixes #6867

**Proposed changes**:
- During conversion of test stories from MD to YAML, entities are kept in the user utterance 


rasa/core/training/converters/story_markdown_to_yaml_converter.py
```diff
@@ -41,12 +41,12 @@ async def convert_and_write(cls, source_path: Path, output_path: Path) -> None:
 
         # check if source file is test stories file
         if MarkdownStoryReader.is_test_stories_file(source_path):
-            reader = MarkdownStoryReader(unfold_or_utterances=False, use_e2e=True)
+            reader = MarkdownStoryReader(unfold_utterances=False, use_e2e=True)
```
wochinge, line 44: why do we need `unfold_utterances` if we have `use_e2e`. That's the same thing, isn't it?

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -231,6 +231,11 @@ def parse_e2e_message(line: Text) -> Message:
         intent = match.group(2)
         message = match.group(4)
         example = entities_parser.parse_training_example(message, intent)
+        if not unfold_utterances:
```
wochinge, line 234: Why is this a change in the `MarkdownReader`. I would expect that we have to change the `YAMLStoryWriter` to inline the entity annotations when `use_e2e=True`?

wochinge (commented):
A couple of questions first which I need to understand before I can finish the review 🙈

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
degiz: Because the "unfoling" of i.g. "OR" statements happens during reading. The writer takes a list of StorySteps that are already modified 😬

rasa/core/training/converters/story_markdown_to_yaml_converter.py
degiz: Because when you read a test story that will not be written later (i.g. the actual `rasa test` command with `use_e2e=True`), you actually want to parse the entities 😄

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
degiz: So the `unfold_utterances=False` is used only with `StoryMarkdownToYamlConverter`

rasa/core/training/converters/story_markdown_to_yaml_converter.py
wochinge: Sure. So why not always unfold them?

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
wochinge: Sorry, I'm not getting this 🙈 Can't the writer "fold" it back?

degiz: No, it can't

degiz: "OR" -> copy over the steps that were before with no back trace.

rasa/core/training/converters/story_markdown_to_yaml_converter.py
degiz: because we want to convert the original user stories and keep the "OR" statements they've been using.

wochinge: Ok, I see now why `e2e=True` is different to `unfold_utterances` 👍

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
wochinge: Ok, but why don't we parse the entities as we're used to in `parse_e2e_messages` and then fold them back in in `YAMLStoryWriter` whenever we write an end-to-end-test story?

degiz: Because we replace the original text of the training example and remove all `{` `}` from it, so we don't have this information anymore.

degiz: And this is exactly what this PR is targeting - preserving this information, bc for simple conversion we don't want to replace the original text.

wochinge: > Because we replace the original text of the training example and remove all { } from it, so we don't have this information anymore.

But although we don't have the entities in the text anymore, we still got is as `entities` properties within the `UserUttered` events, do we? So we can generate the text including the brackets in the writer.

wochinge: Just checked the `YAMLStoryWriter` code: so the `YAMLStoryWriter` has no clue whether it's currently in end-to-end mode or not? And we are trying to achieve different behavior in the output format by making the reader change?

degiz: Yes, but why would we do that? 😄

wochinge: It seems super awkward to me to change the reader if we want to have a different output format. Also, we need this when dumping actual (not converting) end-to-end stories, don't we?

wochinge: @degiz another solution I could imagine is to subclass `MarkdownStoryReader` with something like `MarkdownStoryReaderForConversion` which implements the special behavior which you need for converting the stories. This way the converter-specific behavior would be isolated and clearly identifiable

degiz: @wochinge  I like the idea with MarkdownStoryReaderForConversion, but that's quite an overkill for a reader that we deprecate, because that would require some refactorings.
Do you think it's clearer what's happening if I rename unfold_utterances to keep_original_user_utterances? Bc that's what it does essentially - keeps utterance the same as they were in a text file.

wochinge: how about naming it `is_used_for_conversion`? Then it's clear to everyone that this is a flag which should only be used when converting training data, but at no point else?

degiz: sounds good to me 👍

wochinge: you have to invert your boolean logic then 🙈

wochinge (approved):
Thanks 🙌

PR #5991, opened 2020-06-10
Fix link to rasa-demo actions file in docs

melindaloubser1:
**Proposed changes**:
- fix link to rasa-demo actions file

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [x] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #5887, opened 2020-05-24
Fix code line in actions.py for spreadsheet info

vreyespue:
The code line should be updated to the specific portion of the code where the info is stored into a spreadsheet. In addition, after applying this fix it should be ensured that the links contained in the doc site (https://rasa.com/docs/rasa/user-guide/building-assistants) are correctly updated (previously incorrectly pointing to https://github.com/RasaHQ/rasa-demo/blob/master/demo/actions.py#L69 and delivering 404 error).

**Proposed changes**:
- ...

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [x] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #6891, opened 2020-10-03
Omit line anchors in links to functions because it is error prone

vreyespue:
**Proposed changes**:

As discussed in [this stackoverflow post](https://stackoverflow.com/questions/18220650/github-link-to-function-in-source) we can have anchors to a certain line, but if the source changes that line will become irrelevant in the future. Thus, using line anchors is in general error prone. 

From here there are two possible solutions:
- we might link to a certain revision of the file (/RasaHQ/rasa/blob/{hash}/{file}#L{line}), or
- we simply link to the whole file instead of the specific line, and clarify what to look for.

I propose to using the second solution rather than the first, because in that case we will refer the reader to a previous version of the function, which might have changed fundamentally since that specific time point.

Hope this helps :)
Many thanks and best regards.

Relates to #5887, #5991
CC @melindaloubser1, @erohmensing, @TyDunn 


sara-tagger:

melindaloubser1:
Thanks for the PR @vreyespue, that seems like a good solution to me. Are these two the only cases of line-specific links you found?

vreyespue:
> Thanks for the PR @vreyespue, that seems like a good solution to me. Are these two the only cases of line-specific links you found?

Hi @melindaloubser1, yes, these are the only two cases that I found using a recursive grep:
```
RasaHQ/rasa $ grep -InHR '#L' *
docs/docs/components.mdx:1089:  [MITIE trainer code](https://github.com/mit-nlp/MITIE/blob/master/mitielib/src/text_categorizer_trainer.cpp#L222)).
rasa/core/nlg/callback.py:128:                "https://github.com/RasaHQ/rasa/blob/master/rasa/core/nlg/callback.py#L12"
tests/core/test_processor.py:322:    name = "wkjbgr,34(,*&%^^&*(OP#LKMN V#NF# # #R"
```
Note that the last line belongs to a random string for a test, and therefore it is not relevant.
Many thanks and best regards :)

melindaloubser1 (approved):

PR #6894, opened 2020-10-04
HTTP Proxy Support for Slack Connector in Rasa 1.10.x

BrainConverter:
**Proposed changes**:
- Environment variables for HTTP-Proxies (`HTTP_PROXY`, `HTTPS_PROXY`) will be trusted by the slackclient. Since Rasa is using the async mode of the slackclient, the `trust_env` parameter has to be passed to the [AIOHTTP Client](https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support).

### DIFF
--- a/rasa/core/channels/slack.py
+++ b/rasa/core/channels/slack.py
@@ -8,6 +8,8 @@ from rasa.utils.common import raise_warning
 from sanic import Blueprint, response
 from sanic.request import Request
 from sanic.response import HTTPResponse
+
+from aiohttp import ClientSession
 from slack import WebClient
 
 logger = logging.getLogger(__name__)
@@ -23,7 +25,7 @@ class SlackBot(OutputChannel):
     def __init__(self, token: Text, slack_channel: Optional[Text] = None) -> None:
 
         self.slack_channel = slack_channel
-        self.client = WebClient(token, run_async=True)
+        self.client = WebClient(token, run_async=True, session=ClientSession(trust_env=True))
         super().__init__()
 
     @staticmethod
--- a/rasa/core/channels/slack.py
+++ b/rasa/core/channels/slack.py
@@ -25,7 +25,9 @@ class SlackBot(OutputChannel):
     def __init__(self, token: Text, slack_channel: Optional[Text] = None) -> None:
 
         self.slack_channel = slack_channel
-        self.client = WebClient(token, run_async=True, session=ClientSession(trust_env=True))
+        self.client = WebClient(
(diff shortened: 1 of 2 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @koaning will take a look at it as soon as possible ✨

wochinge:
@BrainConverter Thanks for your PR! I will probably not get to it today, but will have a look on Monday 👍

rasa/core/channels/slack.py
```diff
@@ -23,7 +25,9 @@ def name(cls) -> Text:
     def __init__(self, token: Text, slack_channel: Optional[Text] = None) -> None:
 
         self.slack_channel = slack_channel
-        self.client = WebClient(token, run_async=True)
+        self.client = WebClient(
+            token, run_async=True, session=ClientSession(trust_env=True)
```
wochinge, line 29: @BrainConverter I'm a bit torn at the moment. By creating the `session` manually, we loose access to slacks default configuration (namely the configured timeout). I discovered that `WebClient` also accepts a `proxy` argument which you could use to pass the proxy URL. 
In addition I think it could be preferable for users to configure the proxy URL via the `credentials.yml` as all config for the channel is in one place then. Hence, I think it be slightly preferable to set the `proxy` arg via `credentials.yml`. What do you think?

BrainConverter: You are right this might not be the best implementation.
I'll think about another solution and put this PR back to draft.

wochinge: Please feel free to tag me and discuss ideas / opinions with me.

BrainConverter: Since the functionality is already implemented in Rasa 2.x is there a best practice how to sync the different versions of the Slack connector (Rasa 2.x and 1.x) ?

wochinge:
@BrainConverter In response to your comment [here](https://github.com/RasaHQ/rasa/pull/6894#discussion_r512189796)

Sorry for not realizing this sooner, but we only ship bug fixes or documentation improvements for old minor versions (such as `1.10`). If you can't or don't want to migrate to Rasa Open Source 2.0, then I'd suggested that you take the channel implementation from 2.0 and refer to it by its module path as described [here](https://rasa.com/docs/rasa/connectors/custom-connectors). So your `credentials.yaml` would roughly look like this:

```
path_to_rasa_open_source_2_module.SlackInput:
  # your config for the channel
```

BrainConverter:
Thanks for that advice!

PR #6898, opened 2020-10-05
allow empty string text key and extra button keys in domain validation

### LINKED ISSUE #6897: Domain schema validation fails with empty string for text key
**Rasa version**:
2.0.0rc3

**Issue**:

Domain validation fails for an empty string in the text key. For example if you add this: 
```
  utter_only_image:
    - text: ""
      image: "https://hi.com" 
```
This is the recommended way to send image-only messages as we require either the `text` or `response` key, but the check doesn't work correctly.

erohmensing:
**Proposed changes**:
- close #6897: allow `text: ""` for the "at least one of text or custom" validation
- allow extra keys in addition to `title` and `payload` for buttons


rasa/shared/utils/pykwalify_extensions.py
```diff
@@ -20,7 +20,7 @@ def require_response_keys(
             # this is handled by other validation rules
             continue
 
-        if not response.get("text") and not response.get("custom"):
+        if not (response.get("text") is not None) and not response.get("custom"):
```
tmbo, line 23: mhm this is odd, what's the `not is not` for, so should it be

```
if response.get("text") is None and not response.get("custom"):

Comment (erohmensing) [reply rasa/shared/utils/pykwalify_extensions.py:23]: Aha, good point! I was just processing the information in a weird way 🤔 
```suggestion
        if response.get("text") is None and not response.get("custom"):
```

erohmensing, line 23:
```suggestion
        if response.get("text") is None and not response.get("custom"):
```

tmbo (approved):

PR #6905, opened 2020-10-05
form deactivations during validation

wochinge:
**Proposed changes**:
- forms are now correctly disabled when a custom validation action sets the next requested slot to `None`
- form can now be disabled during validation by setting the currently `active_loop` to None

### DIFF
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -15,7 +15,7 @@ from rasa.shared.core.constants import (
     LOOP_INTERRUPTED,
 )
 from rasa.shared.constants import UTTER_PREFIX
-from rasa.shared.core.events import Event, SlotSet, ActionExecuted
+from rasa.shared.core.events import Event, SlotSet, ActionExecuted, ActiveLoop
 from rasa.core.nlg import NaturalLanguageGenerator
 from rasa.shared.core.trackers import DialogueStateTracker
 from rasa.utils.endpoints import EndpointConfig
@@ -469,8 +469,10 @@ class FormAction(LoopAction):
         """Request the next slot and utter template if needed, else return `None`."""
         request_slot_events = []
 
-        # If this is not `None` it means that the custom action specified a next slot
-        # to request
+        if await self.is_done(output_channel, nlg, tracker, domain, events_so_far):
+            # The custom action for slot validation decided to stop the form early
+            return [SlotSet(REQUESTED_SLOT, None)]
+
         slot_to_request = next(
             (
                 event.value
@@ -479,6 +481,7 @@ class FormAction(LoopAction):
             ),
             None,
         )
+
         temp_tracker = self._temporary_tracker(tracker, events_so_far, domain)
 
         if not slot_to_request:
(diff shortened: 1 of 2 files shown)


rasa/core/actions/forms.py
```diff
@@ -664,7 +667,25 @@ async def is_done(
...
+        ) == SlotSet(REQUESTED_SLOT, None) or next(
+            (
+                event
+                for event in reversed(events_so_far)
+                if isinstance(event, ActiveLoop)
+            ),
+            None,
+        ) == ActiveLoop(
+            None
+        )
```
Ghostvv, line 688: `in` operator doesn't work?

Ghostvv: `SlotSet(REQUESTED_SLOT, None) in events_so_far` seems to me much more readable

wochinge: I explicitly wanna only check against the last occurrences of these events. I wanna avoid that something earlier returns `ActiveLoop(None)` but something later is trying to override it with `ActiveLoop("dasdasd")` and it doesn't work. It should usually be the same, but I think this is slightly more robust

Ghostvv: ok, not sure if we should catch multiple settings differently, but for now could you please add this comment to the code, so that in the future we could return to it, if smth doesn't work as expected

wochinge: done, do I get a 👍 then?

Ghostvv (approved):
🚀

wochinge:
My code didn't change any imports @alwx So Imo it's not my fault that the startup speed decreased :-D

JulianGerhard21:
Hi @wochinge ,

with `rasa==2.0.3` I currently get:

```
  File "\sanic\app.py", line 976, in handle_request
    response = await response
  File "rasa_sdk\endpoint.py", line 102, in webhook
    result = await executor.run(action_call)
  File "rasa_sdk\executor.py", line 390, in run
    events = await action(dispatcher, tracker, domain)
  File "rasa_sdk\forms.py", line 752, in run
    return await self.validate(dispatcher, tracker, domain)
  File "rasa_sdk\forms.py", line 738, in validate
    slots.update(validation_output)
TypeError: cannot convert dictionary update sequence element #0 to a sequence
```

when using:

```
def validate_contact_persistence_permission(
        self,
        slot_value: Any,
        tracker: Tracker,
        domain: DomainDict,
    ) -> Dict[Text, Any]:
        """Validate the persistence permission."""

        if not slot_value:
            dispatcher.utter_message(template="utter_no_permission")
            return [ActiveLoop(None)]

        return {"contact_persistence_permission": slot_value}
```

However:

```
if not slot_value:
    dispatcher.utter_message(template='utter_no_permission')
    return {"requested_slot": None}
```

works as expected.

Maybe this isn't intended? 

Kind regards
Julian

wochinge:
@JulianGerhard21  The problem is that you're returning the wrong type when using `[ActiveLoop(None)]`. In this case it's a list, but we expect a `Dict`. I'm currently working on something to make this all a bit easier: https://github.com/RasaHQ/rasa-sdk/pull/329

JulianGerhard21:
@wochinge Sorry - I should have made that clearer - of course I have noticed that you want to update the slots dictionary after the return statement and a `List` of `Any` (Events) would be a problem. 

I just wanted to mention, that [this forum post](https://forum.rasa.com/t/rasa-2-0-how-to-deactivate-a-form-during-validation/35445/9?u=juliangerhard) gives the impression, that both approaches will work the same way - apparently they don't :-) 

Thanks for making that clear - I appreciate the thoughts on the linked sdk proposed changes.

wochinge:
> I just wanted to mention, that this forum post gives the impression, that both approaches will work the same way - apparently they don't :-)

If you override the `FormValidationAction` it indeed doesn't work 🙈 And I think that's ok. The new PR will hopefully makes this easier 🙂  You can use plain `Action` to implement this as well btw. Then you can also use the `ActiveLoop(None)` event.

JulianGerhard21:
@wochinge Absolutely OK. Thanks for the explanation!

PR #6910, opened 2020-10-05
add validation of docstrings (pydocstyle)

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa-x/issues/3568
- only run against diff to avoid too many errors. 
  * We use the `flake8-pydocstyle` plugin so we can use `flake8 diff` to run only against the diff. `pydocstyle` doesn't natively support on option to only check the diff.
  * `git diff` includes one line to the top and bottom in the diff which actually wasn't changed. We could get rid of these extra diffs, but I think it's even good to capture a tiny bit larger diff so that we arrive at "fixed" state quicker and can run `pydocstyle` on everything by default
- list of things which can be validated [here](http://www.pydocstyle.org/en/stable/error_codes.html)

**Known Errors**:
- missing arg documentation is  not caught if `Returns` is present: https://github.com/PyCQA/pydocstyle/issues/459
- there is no option to disallow one line docstrings. They are currently valid. We could add add test on our side to fix this?
- can't check for descriptive phrasing (only for imperative)

**FAQ**
* _Are there alternatives to `pydocstyle`_ - No.
* _Can we ever run `pydocstyle` on everything instead of the diff?_ - Yes. However, it's best to fix our issues iteratively before switching to the entire diff.


rasa/shared/core/domain.py
```diff
@@ -641,6 +641,14 @@ def raise_action_not_found_exception(self, action_name) -> NoReturn:
         )
 
     def random_template_for(self, utter_action: Text) -> Optional[Dict[Text, Any]]:
```
wochinge, line 643: that's just for provoking errors and to show case the CI errors

Makefile
```diff
@@ -59,9 +61,16 @@ formatter:
 	poetry run black rasa tests
 
 lint:
-	poetry run flake8 rasa tests
+    # Ignore docstring errors when running on the entire project
```
ricwo, line 64:
```suggestion
     # Ignore docstring errors when running on the entire project
```

Makefile
```diff
@@ -59,9 +61,16 @@ formatter:
...
+	poetry run flake8 rasa tests --extend-ignore D
 	poetry run black --check rasa tests
 
+branch ?= master
+lint-docs:
+    # Lint docstrings only against the the diff to avoid too many errors.
+    # Check only production code. Ignore other flake errors which are captured by `lint`
```
ricwo, line 71:
```suggestion
     # Lint docstrings only against the the diff to avoid too many errors.
     # Check only production code. Ignore other flake errors which are captured by `lint`
```

.github/workflows/continous-integration.yml
```diff
@@ -99,6 +99,15 @@ jobs:
...
       run: make lint
 
+    - name: Checkout target branch to be able to diff
+      if: github.event_name == 'pull_request'
+      run: |
+        git fetch --depth=1 origin ${{ github.base_ref }}
+
+    - name: Check Docstring conventions 📃
+      if: github.event_name == 'pull_request'
+      run: make lint-docs branch=origin/${{ github.base_ref }}
```
ricwo, line 109: is there any way these checks can happen before the full `install poetry` stage? that way we wouldn't have to wait ten minutes before docstring errors are caught.

The only libs required for this are `flake8` and `flake8-docstrings` right?

setup.cfg
```diff
@@ -18,4 +18,7 @@ log_cli_level = WARNING
 
 [flake8]
 max-line-length = 88
-ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, E231
+ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, E231,
+    D100,  # ignore missing module docstrings
```
ricwo, line 22:
```suggestion
    D100,  # ignore missing docstrings in public module
```

setup.cfg
```diff
@@ -18,4 +18,7 @@ log_cli_level = WARNING
...
+    D104,  # ignore missing package docstrings
```
ricwo, line 23:
```suggestion
    D104,  # ignore missing docstrings in public package
```

setup.cfg
```diff
@@ -18,4 +18,7 @@ log_cli_level = WARNING
 
 [flake8]
 max-line-length = 88
-ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, E231
+ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, E231,
```
ricwo, line 21: probably out of scope, but it would be very useful to document these codes as well with comments

rasa/shared/core/domain.py
```diff
@@ -877,6 +890,39 @@ def as_dict(self) -> Dict[Text, Any]:
             KEY_E2E_ACTIONS: self.action_texts,
         }
 
+    def new_function(self, arg1, arg2, arg3, arg4):
```
ricwo, line 893: should we add all these tests to a unit test file `test_docstrings` which is exempted from the default tests, but is covered in unit tests?

rasa/shared/core/domain.py
```diff
@@ -877,6 +890,39 @@ def as_dict(self) -> Dict[Text, Any]:
...
+        pass
+
+    def persist(self, filename: Union[Text, Path]) -> None:
+        """Write domain to a file."""
+
+        domain_data = self.as_dict()
+        rasa.shared.utils.io.write_yaml(
+            domain_data, filename, should_preserve_key_order=True
+        )
+
```
ricwo, line 925: what about a valid docstring that's not in Google style, but, say, in Numpy?

Makefile
```diff
@@ -14,6 +14,8 @@ help:
 	@echo "        Apply black formatting to code."
 	@echo "    lint"
 	@echo "        Lint code with flake8, and check if black formatter should be applied."
```
ricwo, line 16: what do you think about including this as part of `make lint`?

ricwo (commented):
This looks really awesome! thanks for exploring this 🚀

Makefile
```diff
@@ -59,9 +61,16 @@ formatter:
 	poetry run black rasa tests
 
 lint:
-	poetry run flake8 rasa tests
+    # Ignore docstring errors when running on the entire project
+	poetry run flake8 rasa tests --extend-ignore D
 	poetry run black --check rasa tests
 
+branch ?= master
+lint-docs:
```
m-vdb, line 69: I would rename this to `lint-docstrings:` because it's more explicit. "docs" can be misinterpreted with the actual documentation (related but not the same thing

Makefile
```diff
@@ -59,9 +61,16 @@ formatter:
 	poetry run black rasa tests
 
 lint:
-	poetry run flake8 rasa tests
+    # Ignore docstring errors when running on the entire project
+	poetry run flake8 rasa tests --extend-ignore D
 	poetry run black --check rasa tests
 
+branch ?= master
```
m-vdb, line 68: this checks for a mandatory `branch` parameter? neat, didn't know about this

rasa/shared/core/domain.py
m-vdb: I don't think it's a good idea, it would mean that we're testing flake8/pydocstyle themselves

rasa/shared/core/domain.py
m-vdb: I say we only allow one style, for consistency

setup.cfg
m-vdb: 🤔 but flake8 does this as well. Maybe simply a link to their doc?

m-vdb (changes requested):
@wochinge sorry for the late review, and thanks a lot for the initiative, that's 💯 !

I left a couple of comments. I was also wondering if you already went through the [list of error codes](http://www.pydocstyle.org/en/stable/error_codes.html) and ignored the ones that we do not care about (for instance I saw that you already ignored D100 and D104). My concern I guess, is that if we do not tweak the config enough in this PR:
1. people will spend time fixing warnings that we do not care about, hindering the whole dev experience
2. people will require some sort of support to figure out if they need to do 1.

Does that make sense? What's your take on this?

rasa/shared/core/domain.py
ricwo: yes but I mean a test (even if it's just for the draft PR stage) that checks this

.github/workflows/continous-integration.yml
m-vdb: 🤔 but why should this run before any other checks?

rasa/shared/core/domain.py
m-vdb: works for me on only if it's at the draft PR stage :) otherwise I don't see the point of testing flake8 lib

.github/workflows/continous-integration.yml
```diff
@@ -99,6 +99,15 @@ jobs:
     - name: Lint Code 🎎
       run: make lint
 
+    - name: Checkout target branch to be able to diff
+      if: github.event_name == 'pull_request'
+      run: |
+        git fetch --depth=1 origin ${{ github.base_ref }}
```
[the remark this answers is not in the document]
wochinge: by default it only checks out the last commit of the target branch. We could change it do `depth: 0` but that would do a clone of the entire repository and its history. The `fetch` is the quickest way hence.

Makefile
wochinge: `make lint` runs against the entire codebase and not just the diff. I could concatenate it with `&&` though 🤔

rasa/shared/core/domain.py
wochinge: I agree with Maxime 👍

setup.cfg
wochinge: added links

Makefile
wochinge: No, it provides a default it isn't provided. Added a comment.

Makefile
wochinge: done

rasa/shared/core/domain.py
wochinge: Wait - should I add one to make sure `pydocstyle` verifies it?

wochinge:
> . I was also wondering if you already went through the list of error codes and ignored the ones that we do not care about (for instance I saw that you already ignored D100 and D104). 

In addition to my exclusions the `google` convention will exclude some more.

I think all the remaining error codes seem legit. I think it's also fine if it's too pedantic. Enforcing consistency across our code base will only make it easier. And I don't think it will hinder the entire dev experience because we can 1) Quickly add an ignore if we really don't want a certain error code and 2) The error codes are pretty self-explanatory and and easy to fix.

rasa/shared/core/domain.py
m-vdb: could be a good idea just to double check, but it doesn't need to be merged in the codebase in my opinion

m-vdb:
ok sounds good! I kind of missed the `docstring-convention` directive 🤷

rasa/shared/core/domain.py
wochinge: unfortunately it doesn't fail on docstrings with the wrong convention 😬 Seems they will fix this in version 6 of pydocstyle (see https://github.com/PyCQA/pydocstyle/issues/482#issuecomment-679403107)

wochinge: pushed an example with a numpy docstring btw

Makefile
```diff
@@ -59,8 +61,16 @@ formatter:
 	poetry run black rasa tests
 
 lint:
-	poetry run flake8 rasa tests
+     # Ignore docstring errors when running on the entire project
+	poetry run flake8 rasa tests --extend-ignore D
 	poetry run black --check rasa tests
+	make lint-docstrings
+
+branch ?= master # Compare against `master` if no branch was provided
```
m-vdb, line 69: minor: should we make the variable name uppercase like JOBS above?

rasa/shared/core/domain.py
```diff
@@ -641,6 +641,14 @@ def raise_action_not_found_exception(self, action_name) -> NoReturn:
         )
 
     def random_template_for(self, utter_action: Text) -> Optional[Dict[Text, Any]]:
+        """Great docstring but no dots at the end
```
m-vdb, line 644: it says "First line should end with a period, question mark, or exclamation point", does it support multi-line though? Would:
```suggestion
        """Great docstring but no dots
        at the end
```
Fail? What if there is the dot only on the second line?

rasa/shared/core/domain.py
```diff
@@ -763,6 +775,7 @@ def _get_active_loop_sub_state(
             tracker: dialog state tracker containing the dialog so far
         Returns:
             a dictionary mapping "name" to active loop name if present
+
         """
```
m-vdb, line 780: any idea why the check doesn't fail at this line with "D202 No blank lines allowed after function docstring" ?

m-vdb (commented):
Works for me 👌 I've left a couple of other questions, sorry I miss this in the first round. I think it's mergeable once the test code is removed 💯 thanks for doing this @wochinge

ricwo (approved):
💯

rasa/shared/core/domain.py
wochinge: The first line should actually never be multiline as stated [here](https://google.github.io/styleguide/pyguide.html#382-modules)

rasa/shared/core/domain.py
wochinge: It now works. Some validations seem a bit fragile. In this case the empty line after the first line of the docstring was missing.

Makefile
wochinge: done

rasa/shared/core/domain.py
m-vdb: I didn't know about this recommendation. Not sure everyone is aware

m-vdb (approved):
Looks great 🚀  @wochinge mergeable once the testing code is removed

Question: do you plan on setting it up on Rasa X as well?

rasa/shared/core/domain.py
wochinge: I think it makes a lot of sense to keep the summary a summary + that's a habit which should be quick to adapt (in the end it forces you to write less 😂 )

wochinge:
> Question: do you plan on setting it up on Rasa X as well?

👍

m-vdb:
🎉 @wochinge have you thought about communicating these changes on Slack to eng/cse/research?

wochinge:
Jep, just did

PR #6912, opened 2020-10-05
dump domain as yaml including the training data version

wochinge:
**Proposed changes**:
- fix this warning:
  ```
  020-10-05 16:53:40 WARNING  rasa.shared.utils.validation  - Training data file 
  /var/folders/h_/jnhnb5q137x0t5cfbv6cc9hr0000gp/T/tmpdevsvzgf/core/domain.yml doesn't have a 'version' key. Rasa Open 
  Source will read the file as a version '2.0' file. See https://rasa.com/docs/rasa.
  ```
  The problem was that we persisted the domain without `version` key after training the model.
- include `version: "2.0"` in domain yaml when dumping it
- base `dump_obj_as_yaml_to_string` on `write_yaml` instead of re-inventing the wheel


rasa/shared/utils/io.py
```diff
@@ -366,22 +366,23 @@ def dump_obj_as_json_to_file(filename: Union[Text, Path], obj: Any) -> None:
...
-    yaml_writer.unicode_supplementary = True
-    yaml_writer.default_flow_style = False
-    yaml_writer.version = YAML_VERSION
+    Args:
+        obj: The object to dump. Has to be serializable.
+        should_preserve_key_order: Whether to force preserve key order in `data`.
 
-    yaml_writer.dump(obj, output)
+    Returns:
+        The object as YAML string.
```
degiz, line 379:
```suggestion
        The object converted to a YAML string.
```

rasa/shared/utils/io.py
```diff
@@ -366,22 +366,23 @@ def dump_obj_as_json_to_file(filename: Union[Text, Path], obj: Any) -> None:
...
+    obj: Dict, should_preserve_key_order: bool = False
+) -> Text:
+    """Writes data (python dict) to a yaml string.
 
-    yaml_writer = ruamel.yaml.YAML(pure=True, typ="safe")
-    yaml_writer.unicode_supplementary = True
-    yaml_writer.default_flow_style = False
-    yaml_writer.version = YAML_VERSION
+    Args:
+        obj: The object to dump. Has to be serializable.
```
degiz, line 375: It's already a dict according to the type hint, do you think we should mention `serializable`?

degiz (approved):
Thanks for fixing that! 🚀

rasa/shared/utils/io.py
wochinge: type hint was wrong 👍 Can also be `list` or `Text` or whatever

PR #6916, opened 2020-10-05
Added Language Support Page to 2.0 Docs

b-quachtran:
**Proposed changes**:
- Added Language Support Page to 2.0 Docs

### DIFF
--- a/docs/docs/language-support.mdx
+++ b/docs/docs/language-support.mdx
@@ -5,25 +5,21 @@ title: Language Support
 abstract: You can use Rasa to build assistants in any language you want.
 ---
 
-Rasa's ``supervised_embeddings`` pipeline can be used on training data in **any language**.
-This pipeline creates word embeddings from scratch with the data you provide.
+Your Rasa assistant can be used on training data in **any language**.
+If there are no word embeddings for your language, you can train your featurizers
+from scratch with the data you provide.
 
 In addition, we also support pre-trained word embeddings such as spaCy. For information on
 what pipeline is best for your use case, check out [choosing a pipeline](./tuning-your-model.mdx#how-to-choose-a-pipeline).
 
 ## Training a Model in Any Language
 
-Rasa's ``supervised_embeddings`` pipeline can be used to train models in any language, because
-it uses your own training data to create custom word embeddings. This means that the vector
-representation of any specific word will depend on its relationship with the other words in your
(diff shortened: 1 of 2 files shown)


erohmensing (changes requested):
I don't know if this is the right spot for this. I think this page is mostly about making sure that people who are starting know that Rasa can support their language. Here's what I'd propose: 

Move this to the getting started section (which we removed bc it had only one entry, but can add back in: 

Building Assisnants
  Getting Started
    Installation
    Language Support

Make this page itself less technical, along the lines of "you can build assistants in any lanuguage you want. Rasa provides components that can learn directly from your data, ... 
You can also incorporate existing language models...." And then we can link to the suggested pipelines for the different recommendations (not sure if those will continue to live in "tuning your nlu model"

I'm interested to see if @karen-white has ideas for this page, as it sort of seems like something that fits in the marketing FAQ as well, but we do need search in the docs to turn something up along these lines

erohmensing:
Considering this is a highly requested page, I'll add it as is, but would like us to reconsider the content and where it goes as part of the "finish documentation updates" for Q4

docs/docs/language-support.mdx
```diff
@@ -0,0 +1,71 @@
...
+```
+
+:::note
+Even more so when training word embeddings from scratch, more training data will lead to a
+better model! If you find your model is having trouble discerning your inputs, try training
+with more example sentences.
+:::
+
+
+## Pre-trained Word Vectors
```
erohmensing, line 43:
```suggestion
## Using Pre-trained Language Models
```

docs/docs/language-support.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+If you can find them in your language, pre-trained word vectors are a great way to get started with less data,
+as the word vectors are trained on large amounts of data such as Wikipedia.
```
erohmensing, line 46:
```suggestion
If you can find them in your language, language models with pre-trained word vectors are a great way to get started with less data,
as the word vectors are trained on large amounts of data such as Wikipedia.
```

erohmensing (approved):

karen-white:
@erohmensing we include a multi-language mention in our FAQ page (need to update the link though, it looks like): https://info.rasa.com/faqs#multi-lang

agree it would be good to think about how we can enrich this page and make it discoverable :)

PR #6925, opened 2020-10-06
Delete model directory if it wasn't loaded correctly, when fetching from server

federicotdn:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa-x/issues/3597.

### DIFF
--- a/rasa/core/agent.py
+++ b/rasa/core/agent.py
@@ -115,40 +115,26 @@ def _load_domain_and_policy_ensemble(
 
 def _load_and_set_updated_model(
     agent: "Agent", model_directory: Text, fingerprint: Text
-) -> bool:
+) -> None:
     """Load the persisted model into memory and set the model on the agent.
 
     Args:
         agent: Instance of `Agent` to update with the new model.
         model_directory: Rasa model directory.
         fingerprint: Fingerprint of the supplied model at `model_directory`.
-
-    Returns:
-        `True` if the model was loaded correctly.
     """
     logger.debug(f"Found new model with fingerprint {fingerprint}. Loading...")
 
     core_path, nlu_path = get_model_subdirectories(model_directory)
 
-    try:
-        interpreter = _load_interpreter(agent, nlu_path)
-        domain, policy_ensemble = _load_domain_and_policy_ensemble(core_path)
+    interpreter = _load_interpreter(agent, nlu_path)
+    domain, policy_ensemble = _load_domain_and_policy_ensemble(core_path)
 
-        agent.update_model(
-            domain, policy_ensemble, fingerprint, interpreter, model_directory
-        )
+    agent.update_model(
+        domain, policy_ensemble, fingerprint, interpreter, model_directory
+    )
 
-        logger.debug("Finished updating agent to new model.")
-
-        return True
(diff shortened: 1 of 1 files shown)


rasa/core/agent.py
```diff
@@ -152,25 +159,39 @@ async def _update_model_from_server(
...
-    )
-    if model_directory_and_fingerprint:
-        model_directory, new_model_fingerprint = model_directory_and_fingerprint
-        _load_and_set_updated_model(agent, model_directory, new_model_fingerprint)
-    else:
-        logger.debug(f"No new model found at URL {model_server.url}")
+    model_directory = tempfile.mkdtemp()
+    loaded = False
+
+    try:
```
wochinge, line 165: what do we need the `try: ` block for?

wochinge: If you expect an unzip error - shouldn't this be handled as part of `_pull_model_and_fingerprint` ?

wochinge (approved):
Thanks for the additional docstring work!

rasa/core/agent.py
```diff
@@ -115,13 +115,16 @@ def _load_domain_and_policy_ensemble(
 
 def _load_and_set_updated_model(
     agent: "Agent", model_directory: Text, fingerprint: Text
-) -> None:
+) -> bool:
```
wochinge, line 118: one more thought: Should we delete `core_path`, `nlu_path` in here when an exception occurs?

wochinge:
Should we also delete the old `model_directory` in `update_model` if `model_directory` is a temp directory?

rasa/core/agent.py
federicotdn: I think handling the creation/deletion should be outside `_pull_model_and_fingerprint`, because after the model is downloaded there is still room for errors to happen. So it is easier to move this management to `_update_model_from_server` and handle it there.

I used try/finally because we only want to delete the directory if an error happened. I would've preferred to use `with tempfile.TemporaryDirectory() as ...`, but that would delete it no matter what.

rasa/core/agent.py
federicotdn: See the other comment - since `model_directory` is being received as an argument, I think it is better to leave the management of the directory to the caller of this function.

wochinge: ah. I also overlooked that both directories are within `model_directory` 👍

rasa/core/agent.py
wochinge: Makes sense. But using this logic `_load_and_set_updated_model` should not have a `try ... except` block then, should it?

federicotdn: I think we could let the exception bubble up from `_load_and_set_updated_model`. However then we would need to catch it in `_update_model_from_server` because it at the moment it is being converted into a log message and then ignored.

wochinge: yep, that's what I mean. I think it would be a bit cleaner to handle all exception stuff in one place. Anyway, up to you as we're quite close to the release and it doesn't really matter.

federicotdn:
@wochinge yes that would be good. Let me check if there's a reliable way of knowing if a directory is temporary or not.

PR #6928, opened 2020-10-06
Expose YAML dict structure used by RasaYAMLWriter

federicotdn:
**Proposed changes**:
To be used from the Rasa X side to make training new models easier.


rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -378,6 +378,23 @@ def dump(
             target: Name of the target object to write the YAML to.
             training_data: TrainingData object.
         """
+        result = self.training_data_to_yaml(training_data)
+
+        if result:
+            rasa.shared.utils.io.write_yaml(result, target, True)
+
+    def training_data_to_yaml(
```
degiz, line 386: maybe `training_data_as_dict`? 
`training_data_to_yaml` sounds confusing, I'd expect it to return a YAML string, not a dict.

rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -378,6 +378,23 @@ def dump(
...
+        self, training_data: "TrainingData"
+    ) -> Optional[OrderedDict]:
+        """Converts NLU training data to a dict/list structure ready to be
+        serialized as YAML.
+
+        Args:
+            training_data: `TrainingDa` to convert.
```
degiz, line 393:
```suggestion
            training_data: `TrainingData` to convert.
```

rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -378,6 +378,23 @@ def dump(
...
         """
+        result = self.training_data_to_yaml(training_data)
+
+        if result:
+            rasa.shared.utils.io.write_yaml(result, target, True)
+
+    def training_data_to_yaml(
+        self, training_data: "TrainingData"
+    ) -> Optional[OrderedDict]:
+        """Converts NLU training data to a dict/list structure ready to be
```
degiz, line 389: Maybe `Represents` instead of `Converts`?

tmbo (approved):

rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -378,6 +378,23 @@ def dump(
...
+        serialized as YAML.
+
+        Args:
+            training_data: `TrainingDa` to convert.
+
+        Returns:
+            `OrderedDict` containing all training data as dictionaries and lists.
```
degiz, line 396:
```suggestion
            `OrderedDict` containing all training data.
```

rasa/shared/nlu/training_data/formats/rasa_yaml.py
degiz, line 386: Thinking out loud - is there actually any reason for this method not to be static?

federicotdn: This was to be consistent with `YAMLStoryWriter`, which has a method named `stories_to_yaml` that does something similar. But I can change this one since it is true that `dict` is more descriptive.

federicotdn: That's true, I've changed it now

PR #6929, opened 2020-10-06
rasa test core: check if model was provided

tabergma:
**Proposed changes**:
Check if a model was provided when executing `rasa test core`. If not, print a useful error message and stop.


rasa/cli/test.py
```diff
@@ -83,6 +84,12 @@ def run_core_test(args: argparse.Namespace) -> None:
...
 
+    if args.model is None:
+        rasa.shared.utils.cli.print_error(
+            "No model provided. Please make sure to specify the model to test with '--model'."
+        )
+        return
+
     if isinstance(args.model, str):
         model_path = rasa.cli.utils.get_validated_path(
             args.model, "model", DEFAULT_MODELS_PATH
```
wochinge, line 95: The default value doesn't make sense then, does it?

wochinge (approved):
Thanks! 
Let's also add a test for this 👍

rasa/cli/test.py
```diff
@@ -83,6 +84,12 @@ def run_core_test(args: argparse.Namespace) -> None:
     if isinstance(args.model, list) and len(args.model) == 1:
         args.model = args.model[0]
 
+    if args.model is None:
```
wochinge, line 87: do we need the same for `train_nlu`?

tabergma: `rasa test nlu` does not fail. In this case `args.model` can not be a list. We use the default model path in case the provided one is not given. So it is already handled.

rasa/cli/test.py
tabergma: It does. The errors happens if `args.model = [None]`.

tests/cli/test_rasa_test.py
```diff
@@ -21,6 +21,17 @@ def test_test_core_no_plot(run_in_simple_project: Callable[..., RunResult]):
     assert not os.path.exists(f"results/{CONFUSION_MATRIX_STORIES_FILE}")
 
 
+def test_test_core_with_no_model(run_in_simple_project: Callable[..., RunResult]):
+    assert not os.path.exists("models")
+
+    output = run_in_simple_project("test", "core")
```
wochinge, line 27: 🚀

rasa/cli/test.py
wochinge: but `[None]` already means that there is no model in the `DEFAULT_MODELS_PATH`, no?

tabergma: No. The default value of `--model` for `rasa test core` is set to the following: https://github.com/RasaHQ/rasa/blob/master/rasa/cli/arguments/test.py#L155
In case there is no model folder, it is set to `[None]`. As it is a list of length 1 we extract the first value with is `None`. As `None` is not of type `str` it goes into the `else` part, which failed.

PR #6934, opened 2020-10-06
enable conversation sessions by default

wochinge:
**Proposed changes**:
- enable conversation sessions even if no `session_config` is given in the domain. This is a breaking change in contrast to Rasa Open Source 1 where this would result in disabled conversation sessions. The default value for `session_expiration_time` is the same one which we have been using in our `rasa init` project.
- update migration guide accordingly
- remove duplicate constants (caused when introducing `rasa.shared`).


docs/docs/migration-guide.mdx
```diff
@@ -29,6 +29,17 @@ how you can migrate from one version to another.
   component in your `policies` configuration (`config.yml`) you can replace it
   with `TEDPolicy`. It accepts the same configuration parameters.
 
+* [Conversation sessions](domain.mdx#session-configuration) are now enabled by default
```
wochinge, line 32: The title of the targeted section is currently `Session Configuration`. I'd vote for renaming this to `Conversation Sessions`.

rasa/shared/core/domain.py
```diff
@@ -185,7 +184,6 @@ def from_dict(cls, data: Dict) -> "Domain":
     def _get_session_config(session_config: Dict) -> SessionConfig:
         session_expiration_time_min = session_config.get(SESSION_EXPIRATION_TIME_KEY)
 
-        # TODO: 2.0 reconsider how to apply sessions to old projects and legacy trackers
```
erohmensing, line 188: I guess we've just decided to leave them as is?

erohmensing (approved):
Looks good, this behavior makes more sense with regards to the ["default configuration"](https://rasa.com/docs/rasa/domain#session-configuration) we already reference in the docs 👌

rasa/shared/core/domain.py
wochinge: No, I changed the default from `0` to `60` which essentially enables them by default.

docs/docs/migration-guide.mdx
wochinge: @erohmensing ?

erohmensing: Mm, I think it makes the most sense to have the title of the section to reflect the high level keys in the domain

wochinge: Okay, then let's leave it as it is.

rasa/shared/core/domain.py
erohmensing: i guess i assumed that this note was about "should we read all of the conversations from 2018 as one conversation even if we introduce sessions now? with regards to applying them to legacy trackers

wochinge: Ah, okay. Yes, these will have sessions enabled, too, now

Issue #6896, opened 2020-10-05
Reduce readings of training data files during training


degiz:
**Rasa version**: 2.0.0rc3

As mentioned in the [forum](https://forum.rasa.com/t/rasa-2-reads-many-many-times-the-data/35325), Rasa Open Source reads training data file and Domain more than once during training, whilst the files don't change.
Even though that doesn't slow down things, it produces confusing logs and not optimal usage of resources anyway.
We should reduce the reading of the training data files.

wochinge:
fixed by https://github.com/RasaHQ/rasa/pull/6938

PR #6938, opened 2020-10-06
cache computation of each importer

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6896
- add a utility decorator `cached_method` which caches the results of the `args` and `kwargs` of the call
- add this decorator to all methods of `TrainingDataImporter` classes who do computation which shouldn't be repeated
- remove simple cache mechanism which was used in `E2EImporter` and which would have been duplicate

**Considered Alternatives**
* Just caching within `CombinedDataImporter`: This would have just cached the file reading but not the stuff of all the other importers which loop through stories etc.
* Just cache the outermost importer: In this case we wouldn't have cached the requests from sub importers (e.g. `E2EImporter` to `CombinedDataImporter`)
* Use a parent class `CachedImporter`: Less generic solution and felt also less intuitive


wochinge:
@tmbo Do you have an example for when the warnings happened? I just implemented a caching solution without applying it to the actual problem yet 🙈

tmbo:
a simple one is using an action in a story that is not listed in the domain. this will create a userwarning

wochinge:
defined an `utter_` action without response template 👍 Works as expected (no duplicate warnings).

rasa/shared/utils/common.py
```diff
@@ -66,6 +68,36 @@ def _lazyprop(self):
...
+    Works for `async` and `sync` methods. Don't apply this to functions.
+
+    Returns:
+        The return value which the method gave for the first call.
+    """
+    if asyncio.iscoroutinefunction(f):
+
+        @functools.wraps(f)
+        async def decorated(self: object, *args: Any, **kwargs: Any) -> Any:
+            cache_name = f"_cached_{self.__class__.__name__}_{f.__name__}"
```
tmbo, line 83: don't want to play devils advocate, but doesn't this reuse the same cache result for different argument values? (e.g. language parameter when loading nlu)

wochinge:
something is wrong - will fix and rerequest review

rasa/shared/utils/common.py
tmbo: maybe we should use a std lib implementation for this? `functools.lru_cache`

wochinge: `lru_cache` doesn't work with async. 

> but doesn't this reuse the same cache result for different argument values?

Yes, but I think it's fine. A new importer should be created if used with different parameters (we never user it with the same parameters in one go)

tmbo: mhm it is a source of very hidden bugs though - should we move the arguments to init then?

tmbo: yes very true - if that is possible I think that would be perfect

rasa/shared/utils/common.py
```diff
@@ -66,6 +68,61 @@ def _lazyprop(self):
...
+    """
+
+    class Cache:
+        """Helper class to abstract the caching details."""
+
+        def __init__(self, caching_object: Any, args: Any, kwargs: Any) -> None:
+            self.caching_object = caching_object
+            self.cache = getattr(caching_object, self._cache_name(), {})
+            # noinspection PyUnresolvedReferences
+            self.cache_key = functools._make_key(args, kwargs, typed=False)
```
wochinge, line 87: slightly nasty but should be stable as it's part of the standardlib

rasa/shared/utils/common.py
wochinge: So it ended up being better, but not perfect (see my other comment). At this point we're solving two problems:
* caching the calls
* having an async cache function

I still think the decorator and its currently relatively lightweight implementation are the best solution. The next best thing would be to move all arguments to `__init__` which is both implementation pain + pain for users (potentially even Rasa X, don't know how the training data reading is implemented there for 2.0)

rasa/shared/utils/common.py
```diff
@@ -66,6 +68,64 @@ def _lazyprop(self):
...
+def cached_method(f: Callable[..., Any]) -> Callable[..., Any]:
+    """Caches method calls based on the call's `args` and `kwargs`.
+
+    Works for `async` and `sync` methods. Don't apply this to functions.
+
+    Returns:
+        The return value which the method gives for the first call with the given
+        arguments.
+    """
+
```
tmbo, line 80: should we check if the first argument of `f` is `self` to ensure it is a method?

rasa/shared/utils/common.py
```diff
@@ -66,6 +68,64 @@ def _lazyprop(self):
     return _lazyprop
 
 
+def cached_method(f: Callable[..., Any]) -> Callable[..., Any]:
+    """Caches method calls based on the call's `args` and `kwargs`.
+
+    Works for `async` and `sync` methods. Don't apply this to functions.
+
```
tmbo, line 75: missing `Args:`

rasa/shared/utils/common.py
```diff
@@ -66,6 +68,64 @@ def _lazyprop(self):
...
+        def cached_result(self) -> Any:
+            return self.cache[self.cache_key]
+
+    if asyncio.iscoroutinefunction(f):
+
+        @functools.wraps(f)
+        async def decorated(self: object, *args: Any, **kwargs: Any) -> Any:
+            cache = Cache(self, args, kwargs)
+            if not cache.is_cached():
+                to_cache = await f(self, *args, **kwargs)
```
tmbo, line 111: should we rather cache the future instead of the awaited result? we might call f multiple times if we are still awaiting the return of f but another caller called this function before it completed

tests/shared/utils/test_common.py
```diff
@@ -55,3 +56,65 @@ def test_transform_collection_to_sentence(
...
+    class Test:
+        @rasa.shared.utils.common.cached_method
+        async def f(self):
+            return mock()
+
+    test_instance = Test()
+    assert await test_instance.f() == expected
+    assert await test_instance.f() == expected
+
+    mock.assert_called_once()
```
tmbo, line 90:
```suggestion
async def test_cached_method_with_async_method():
    expected = 5
    mock = Mock(return_value=expected)

    class Test:
        @rasa.shared.utils.common.cached_method
        async def f(self):
            return mock()

    test_instance = Test()
    assert await test_instance.f() == expected
    assert await test_instance.f() == expected

    mock.assert_called_once()


async def test_cached_method_with_async_method_and_sleep():
    expected = 5
    mock = Mock(return_value=expected)

    class Test:
        @rasa.shared.utils.common.cached_method
        async def f(self):
            await asyncio.sleep(1)
            return mock()

    test_instance = Test()
    assert await test_instance.f() == expected
    assert await test_instance.f() == expected

    mock.assert_called_once()
```

tmbo (approved):
looks good :+1: thanks a lot for the quick implementation.

there is one edge case for async method (the proposed test should fail). I think we can merge this without a fix but we should be aware there is a limitation and the cached function might be called multiple times

tmbo:
in a separate step we should unite this with usages of `rasa.shared.utils.common.lazy_property`

tests/shared/utils/test_common.py
wochinge: Do we really need two test cases for that? Imo it's enough to add the `sleep` to the existing test function and be done with it. (I'd use `sleep(0)`  though)

tmbo: also fine with a single test :+1:

rasa/shared/utils/common.py
wochinge: Done. I couldn't use `inspect.ismethod` as it all coroutines are functions 💩

rasa/shared/utils/common.py
wochinge: Done. I wrote a test for this edge case but decided not to add it as it depends on timing and I don't a test in our CI which depends on timing.

tests/shared/utils/test_common.py
wochinge: Now I got what you meant with the test 👍

wochinge:
> in a separate step we should unite this with usages of rasa.shared.utils.common.lazy_property

Not sure what would be the benefit of it. One is for properties - one is for methods. What would be the benefit of unifying them? A next step is imo to make the decorator work with functions, too.

wochinge:
Thanks for the great comments, Tom! Especially the one regarding the edge case was very nice 🚀

tmbo:
> Not sure what would be the benefit of it. One is for properties - one is for methods. What would be the benefit of unifying them? A next step is imo to make the decorator work with functions, too.

right, never mind - that decorator does work a little differently

wochinge:
@tmbo Force merge? 👀 Windows and Ubuntu tests all passed

PR #6962, opened 2020-10-07
properly fix intent_name by removing it

Ghostvv:
**Proposed changes**:
- Remove `INTENT_NAME` constant
- Consider ony NLU training examples during NLU testing, in classifiers and entity extractors
- NLU training examples should have an intent and a text


tmbo (dismissed):
💯

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -607,7 +606,7 @@ def _training_examples_without_empty_e2e_examples(self) -> List[Message]:
         return [
             example
             for example in self.training_examples
-            if not example.get(ACTION_NAME) and not example.get(INTENT_NAME)
+            if not example.get(ACTION_NAME) and not example.get(INTENT)
```
tmbo, line 609: this condition breaks things

Ghostvv: yes, just pushed the fix with correct condition

Ghostvv:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]
```

tabergma:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]
   config: ["all"]
```

tabergma:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]
   config: ["all"]
```

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -577,38 +577,23 @@ def is_empty(self) -> bool:
...
-        regex_features = copy.deepcopy(self.regex_features)
-        lookup_tables = copy.deepcopy(self.lookup_tables)
-        responses = copy.deepcopy(self.responses)
-        copied = TrainingData(
-            training_examples, entity_synonyms, regex_features, lookup_tables, responses
-        )
-        copied.training_examples = self._training_examples_without_empty_e2e_examples()
-
-        return copied
+    def can_train_nlu_model(self) -> bool:
```
Ghostvv, line 587: maybe
```suggestion
    def nlu_data_is_empty(self) -> bool:
```
?

rasa/train.py
```diff
@@ -174,7 +174,7 @@ async def _train_async_internal(
             additional_arguments=nlu_additional_arguments,
         )
 
-    if nlu_data.is_empty():
+    if nlu_data.can_train_nlu_model():
         print_warning("No NLU data present. Just a Rasa Core model will be trained.")
```
Ghostvv, line 178: will it train nlu featurizers?

tabergma: good point, no

We have a test that checks that if only a core model was trained, by default a RegexInterpreter will be used during inference time. If this would be `nlu_data.is_empty` and we have core data but no nlu data, it will train a KeywordClassifier by default, which restulted the test to fail.

tabergma: but that might be something that we can tackle in the `e2e` branch as we don't have proper `e2e` yet?

tabergma: Ah, just took a closer look at the test, it actually uses a config with a `KeywordIntentClassifier`, so that is why the classifier was trained. So, I guess we could also update the test `test_test` to use a empty pipeline? Or would we use the default pipeline then? Also, in what kind of situations should we not train NLU when the user executed `rasa train`?

Ghostvv: >in what kind of situations should we not train NLU when the user executed rasa train?

if there is no data at all or no nlu config

> it actually uses a config with a KeywordIntentClassifier,

why a classifier will be trained if there is no data?

we should train nlu (the components that are in nlu config) if there is data from core (but only those components that have data).

Ghostvv: I guess we can tackle it in e2e or minor

wochinge: > if there is data from core

This exclude data from default actions, right?

tabergma: > why a classifier will be trained if there is no data?

Good question. It is basically an empty `KeywordIntentClassifier`. 

Yeahc let's tackle this in e2e and make sure to cover all possible scenarios.

Ghostvv: we can include them, they exist if core data exists

Ghostvv: @tabergma could you please add it to the list in e2e for real issue?

tests/nlu/extractors/test_regex_entity_extractor.py
```diff
@@ -86,12 +86,17 @@ def test_process(
     training_data.lookup_tables = lookup
     training_data.training_examples = [
         Message(
-            data={TEXT: "Hi Max!", "entities": [{"entity": "person", "value": "Max"}]}
+            data={
+                TEXT: "Hi Max!",
+                INTENT: "intent",
```
tmbo, line 91: why do we need dict entries set to `"intent"`?

tabergma: The `RegexEntityExtractor` uses `training_data.nlu_examples`. We define an NLU example as an training example, that has an intent and a name set.

tmbo: should we rather use a value that indicates that this is an intent then, e.g. `greet`?

tabergma: you mean in the test? sure we could update the value from `intent` to `greet`

tmbo: yes, it'd be great if you can do that at the other places as well. I don't think it helps readability if a value is set to `intent` or `text`, for me an actual example is way easier to digest

tmbo, line 91: the requirement to have an intent on an example to do entity extraction doesn't sound right, esp. for components that don't rely on that data - are we sure this is the right way to implement this?

Ghostvv: it depends. I don't think new data format allows to define examples with entities without intent, no?

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -130,21 +130,21 @@ def sanitize_examples(examples: List[Message]) -> List[Message]:
...
     def entity_examples(self) -> List[Message]:
-        return [ex for ex in self.training_examples if ex.get(ENTITIES)]
-
-    @lazy_property
-    def nlu_examples(self) -> List[Message]:
-        return [ex for ex in self.training_examples if not ex.is_core_message()]
+        return [ex for ex in self.nlu_examples if ex.get(ENTITIES)]
 
     @lazy_property
     def intents(self) -> Set[Text]:
```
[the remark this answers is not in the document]
wochinge: ok, you should definitely do it as a different PR. E.g. all the `number_of_` methods also duplicate logic here.

tests/shared/importers/test_importer.py
```diff
@@ -190,14 +190,14 @@ async def test_import_nlu_training_data_from_e2e_stories(
             StoryStep(
                 events=[
                     SlotSet("some slot", "doesn't matter"),
-                    UserUttered("greet_from_stories", {"name": "greet_from_stories"}),
+                    UserUttered(intent={"name": "greet_from_stories"}),
```
wochinge, line 193: optional: would be great to have on example with entities (can also be done in the `e2e` PR)

tests/nlu/extractors/test_regex_entity_extractor.py
wochinge: I was first with Tom, but then after going through the code it makes sense. intents are labels and entities are labels during the training, too. Either we have a labeled training example or not. But a mix wouldn't make sense.

wochinge (approved):
Thanks for fixing this throughout the night and today ❤️ 💯 !  Most comments are optional. I think  the one about using `can_train_nlu_model` in the `train` module is the most critical one.

tests/nlu/test_train.py
```diff
@@ -116,6 +116,7 @@ def test_all_components_are_in_at_least_one_test_pipeline():
         ), "`all_components` template is missing component."
 
 
+@pytest.mark.timeout(600)
```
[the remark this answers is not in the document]
dakshvar22: @wochinge Yes, it would be downloaded the first time(by the first test that needed it) and the model will be cached in a directory for further tests to use it.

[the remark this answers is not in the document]
wochinge: Awesome 👍

rasa/shared/nlu/training_data/training_data.py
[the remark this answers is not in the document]
tabergma: Yes, I guess for we should update this to `can_train_nlu_model`. I think there is a test somewhere that checks that. But as those tests do not cover any e2e stuff, there is basically no difference. I guess we will add tests with the e2e branch for this.

rasa/shared/importers/importer.py
```diff
@@ -533,18 +533,23 @@ def _unique_events_from_stories(
...
-        action_name=event.action_name, action_text=event.action_text or ""
-    )
+    # sub state correctly encodes action_name vs action_text
+    return Message(data=event.as_sub_state())
 
 
 def _additional_training_data_from_default_actions() -> TrainingData:
     additional_messages_from_default_actions = [
-        Message.build_from_action(action_name=action_name)
+        Message(data={ACTION_NAME: action_name})
```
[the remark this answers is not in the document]
Ghostvv: > create an ActionExecuted event for each default action

this seems to be an overkill here, creating an object just to extract name back is strange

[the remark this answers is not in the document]
wochinge: it's would be just done once for our few default actions. It has the advantage that we only have to fix one place the next time something breaks / changes here.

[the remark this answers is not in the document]
Ghostvv: ok, I suggest let' revisit it as part of e2e, because that's where we can properly test it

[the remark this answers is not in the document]
wochinge: My argument is rather that the tests nearly passed and we don't wanna trigger the CI again 😂

tmbo:
seems like we are ready to go or are there more changes you are working on? @Ghostvv @tabergma @wochinge

tabergma:
No, good to go :rocket:

PR #6972, opened 2020-10-08
Fix description of previous event in output of `rasa data validate stories`

### LINKED ISSUE #6691: `rasa data validate` shows wrong conflicting action name
**Rasa version**:
Rasa Version     : 2.0.0a5 (commit 30d2af7847dd97b02fdad4a4abf7ddb64d8115aa)
Rasa SDK Version : 2.0.0a5
Rasa X Version   : 0.32.0

**Python version**:
Python Version   : 3.7.0 (default, Oct  9 2018, 10:31:47) 

Operating System : Linux-4.15.0-117-generic-x86_64-with-debian-buster-sid

**Issue**:
`rasa data validate` shows wrong conflicting action name. Specifically, if a conflict is present (regardless whether in a story or in a rule), it prints `Story structure conflict after action 'action'`, where the last `action` should be the name of the action instead of the word "action".

**Error**:
With stories
```
- story: happy path
  steps:
  - intent: greet
  - action: utter_greet
  - intent: mood_great
  - action: utter_happy

- story: sad path 1
  steps:
  - intent: greet
  - action: utter_greet
  - intent: mood_unhappy
  - action: utter_cheer_up
  - action: utter_did_that_help
  - intent: affirm
  - action: utter_happy

- story: sad path 2
  steps:
  - intent: greet
  - action: utter_goodbye
  - intent: mood_unhappy
  - action: utter_cheer_up
  - action: utter_did_that_help
  - intent: deny
  - action: utter_goodbye
```
the wrong output is
```
rasa.validator  - Story structure conflict after action 'action':
  utter_greet predicted in 'sad path 1' and 'happy path'
  utter_goodbye predicted in 'sad path 2'
```

```
rasa init
# now add a conflicting story, e.g. by replacing a utter_goodbye by a utter_greet
rasa data validate
```

JEM-Mosig:
**Proposed changes**:
- Fixes #6691 


rasa/core/training/story_conflict.py
```diff
@@ -304,17 +304,28 @@ def _get_previous_event(
...
-            return "action", turn_label.replace(PREV_PREFIX, "")
-        elif turn_label.startswith(INTENT + "_"):
+    for origin, substate in state.items():
+        if origin == PREVIOUS_ACTION:
+            if (
+                "action_name" in substate
+                and substate["action_name"] != ACTION_LISTEN_NAME
+            ):
+                # The `prev_...` was an action that was NOT `action_listen`
+                assert isinstance(substate["action_name"], str)
```
tabergma, line 314: I think it would be better to do a proper check and raise an exception with a good explanation if the action name is not a string. In what kind of situations might action name be not a string?

rasa/core/training/story_conflict.py
```diff
@@ -304,17 +304,28 @@ def _get_previous_event(
     # A typical state is, for example,
     # `{'prev_action_listen': 1.0, 'intent_greet': 1.0, 'slot_cuisine_0': 1.0}`.
     # We need to look out for `prev_` and `intent_` prefixes in the labels.
```
tabergma, line 306: I guess the example is not correct anymore. Please update.

rasa/core/training/story_conflict.py
```diff
@@ -304,17 +304,28 @@ def _get_previous_event(
...
+    for origin, substate in state.items():
+        if origin == PREVIOUS_ACTION:
+            if (
+                "action_name" in substate
+                and substate["action_name"] != ACTION_LISTEN_NAME
+            ):
+                # The `prev_...` was an action that was NOT `action_listen`
+                assert isinstance(substate["action_name"], str)
+                return "action", substate["action_name"]
+            elif "action_text" in substate:
```
tabergma, line 316: As you return in the previous line you could change the `elif` to an `if`.

tabergma (commented):
Looks good 👍 Left a couple of comments.

rasa/core/training/story_conflict.py
JEM-Mosig: I think action name cannot be anything but a string. It's just that `Substate` doesn't require that in terms of types.

rasa/core/training/story_conflict.py
JEM-Mosig: Don't you think `elif` makes it easier to read? Then you don't have to care about what happens inside the `if`-block. One of our code-quality checks used to always tell me that I should use `elif` to reduce "cognitive complexity" and I think that makes sense.

rasa/core/training/story_conflict.py
JEM-Mosig: So I'd think that if this assertion fails, somebody must have modified the Rasa source code.

JEM-Mosig: Is there a way to ensure the type of the value of one key (`action_name`) is actually `Text`?

rasa/core/training/story_conflict.py
tabergma: I think we added something to our coding guidelines that if you do return in one if branch you should not continue with else or elif. But not 100% sure. Do not have a strong opinion here. We can also leave it as it is.

rasa/core/training/story_conflict.py
tabergma: No, as far as I know you cannot make the type dependent on a specific value. 
I would nevertheless go with raising a specific exception instead of the assertion. That makes it also more clear to the users and the developers. You could maybe also add a comment why this exception is raised.

JEM-Mosig: ok, will do :)

rasa/core/training/story_conflict.py
JEM-Mosig: Done

rasa/core/training/story_conflict.py
```diff
@@ -302,19 +302,38 @@ def _get_previous_event(
...
-            previous_event_name = turn_label.replace(INTENT + "_", "")
-
+            if "intent" in substate:
+                previous_event_type = "intent"
+                previous_event_name = substate["intent"]
+            elif "text" in substate:
+                previous_event_type = "user utterance"
+                previous_event_name = substate["text"]
+
+    assert isinstance(previous_event_name, (str, type(None)))
```
tabergma, line 338: I guess we should raise the same exception here.

Looking at the code again, you could also the logic of 
```
                previous_event_type = "intent"
                previous_event_name = substate["intent"]
```
in the if branch `if origin == PREVIOUS_ACTION:`. Then you would just need to raise the exception once.

JEM-Mosig: Ah, forgot about that. :man_facepalming: 
Actually... now that states are dicts we don't even need the loop anymore I think...

tabergma (approved):
Thanks for addressing my comments 💯

PR #6974, opened 2020-10-09
[Port for 2.0.x ] Make sure that compare the same fields for schedule version

tczekajlo:
Port for 2.0.x

In some cases, a report for the schedule version might contain different fields than a report in a PR. To be sure that we compare the same fields we should check if fields that are compared exist in both reports.


tczekajlo:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]
   config: ["all"]

```

ricwo (approved):
looks great 👍

PR #6978, opened 2020-10-09
Entity Roles and Groups influence dialogue predictions

### LINKED ISSUE #5715: Entity Roles and Groups influencing dialogue predictions
**Description of Problem**:
If might be useful to allow roles and groups to influence our dialogue predictions. Currently, they are not considered during the dialogue prediction at all.
Example:
```
# some story
* I want to fly from [Berlin]{"entity": "airport", "role": "departure"}
   - action_handle_departure
   - ....
# another story
* I want to fly to [Berlin]{"entity": "airport", "role": "destination"}
  - action_handle_destination
```

**Overview of the Solution**:
We would need to update the core featurization to also consider entity roles/groups [here](https://github.com/RasaHQ/rasa/blob/master/rasa/core/domain.py#L630). 
Things to consider:
- How should the training data format for stories look like? Currently, you can define a story like this
  ```
  * <intent>{"<entity-type>": "<entity-value>"}
  ```
  You can also specify multiple entities. How would we add roles/group in there? One suggestion
  ```
  * <intent>{"entities": [{"entity": "<entity-type>", "role": "<role-name>", "value": "<entity-value>"}]}
  ```
- We most likely would need to add the roles and groups to their respective entities in the domain file. The entities keyword in the domain file would then become a dictionary instead of a list. 
  ```
  entities:
    city:
      roles:
        - departure
        - destination 
  ```
- What about support in Rasa X?
- This change would break existing models. Assume we would add featurize roles and groups. A user is using entity roles and groups, but did not use them inside their stories. If the user trains a new model, the model would not know anything about roles and groups as the training data do not contain them. However, during prediction we might end up in a situation where an role is detected and featurized. The model would not know what to do. To solve this, the user would need to update the training data.

tabergma:
**Proposed changes**:
closes #5715 


rasa/shared/core/domain.py
```diff
@@ -659,6 +736,46 @@ def slot_states(self) -> List[Text]:
...
+    def entity_states(self) -> List[Text]:
+        """Returns all available entity state strings."""
+        entity_states = copy.deepcopy(self.entities)
+        entity_states.extend(Domain.concatenate_entity_labels(self.roles))
+        entity_states.extend(Domain.concatenate_entity_labels(self.groups))
+        return entity_states
+
+    @staticmethod
+    def concatenate_entity_labels(
+        entities: Dict[Text, List[Text]], entity: Optional[Text] = None
```
Ghostvv, line 750: not sure what would be better, but `entities` here is confusing

rasa/shared/core/domain.py
```diff
@@ -682,8 +799,18 @@ def _get_featurized_entities(self, latest_message: UserUttered) -> Set[Text]:
...
+            | set(
+                f"{entity['entity']}{ENTITY_LABEL_SEPARATOR}{entity['role']}"
+                for entity in entities
+                if "entity" in entity.keys() and "role" in entity.keys()
+            )
+            | set(
+                f"{entity['entity']}{ENTITY_LABEL_SEPARATOR}{entity['group']}"
+                for entity in entities
+                if "entity" in entity.keys() and "group" in entity.keys()
+            )
```
Ghostvv, line 813: could you please add comments, what is happening here

Ghostvv (commented):
looks good to me, since it makes changes in `shared`, someone form engineering team should approve the PR.

Could you please add a test for `SingleStateFeaturizer` that it correctly featurizes roles and groups

tabergma:
@degiz Do you have time to review this PR soon? If not, I would ask someone else. Thanks.

federicotdn:
@tabergma sorry for the delay! I'm not being able to find time to review this - do you think you can assign someone else?

wochinge:
I am on vacation until Tuesday. I think @ricwo is available

rasa/shared/core/domain.py
```diff
@@ -682,10 +802,26 @@ def _get_featurized_entities(self, latest_message: UserUttered) -> Set[Text]:
         )
         intent_config = self.intent_config(intent_name)
         entities = latest_message.entities
-        entity_names = set(
-            entity["entity"] for entity in entities if "entity" in entity.keys()
+
+        # If Entity Roles and Groups is used, we also need to make sure the roles and groups get featurized.
+        # We concatenate the entity label with the role/group label using a special separator to make
+        # sure that the resulting label is unique (as you can have the same role/group label for different entities).
```
ricwo, line 808:
```suggestion
        # If Entity Roles and Groups is used, we also need to make sure the roles and
        # groups get featurized. We concatenate the entity label with the role/group
        # label using a special separator to make sure that the resulting label is
        # unique (as you can have the same role/group label for different entities).
```

rasa/shared/core/domain.py
```diff
@@ -682,10 +802,26 @@ def _get_featurized_entities(self, latest_message: UserUttered) -> Set[Text]:
...
+            )
+            | set(
+                f"{entity['entity']}{ENTITY_LABEL_SEPARATOR}{entity['group']}"
+                for entity in entities
+                if "entity" in entity.keys() and "group" in entity.keys()
+            )
         )
 
+        # the USED_ENTITIES_KEY of an intent also contains the entity labels and the concatenated entity labels with
+        # their corresponding roles and groups labels
```
ricwo, line 824:
```suggestion
        # the USED_ENTITIES_KEY of an intent also contains the entity labels and the
        # concatenated entity labels with their corresponding roles and groups labels
```

rasa/shared/core/domain.py
```diff
@@ -682,10 +802,26 @@ def _get_featurized_entities(self, latest_message: UserUttered) -> Set[Text]:
...
         wanted_entities = set(intent_config.get(USED_ENTITIES_KEY, entity_names))
 
         return entity_names.intersection(wanted_entities)
```
ricwo, line 827: Optional: consistent with using set operators above

```suggestion
        return entity_names & wanted_entities
```

ricwo, line 827: how do we guarantee that the order of items is the same between different featurizations? should this be sorted somewhere?

rasa/shared/core/domain.py
```diff
@@ -897,7 +1033,15 @@ def _transform_intents_for_file(self) -> List[Union[Text, Dict[Text, Any]]]:
             ):
                 # Default intents should be not dumped with the domain
                 continue
-            use_entities = set(intent_props[USED_ENTITIES_KEY])
+            # `use_entities` and `ignore_entities` in the domain file do not consider the role and group labels
+            # remove them from the list to make sure to not put them into the domain file
```
ricwo, line 1037:
```suggestion
            # `use_entities` and `ignore_entities` in the domain file do not consider
            # the role and group labels remove them from the list to make sure to not
            # put them into the domain file
```

rasa/shared/core/domain.py
```diff
@@ -897,7 +1033,15 @@ def _transform_intents_for_file(self) -> List[Union[Text, Dict[Text, Any]]]:
...
+            use_entities = set(
+                [
+                    entity
+                    for entity in intent_props[USED_ENTITIES_KEY]
+                    if ENTITY_LABEL_SEPARATOR not in entity
+                ]
+            )
```
ricwo, line 1044:
```suggestion
            use_entities = set(
                entity
                for entity in intent_props[USED_ENTITIES_KEY]
                if ENTITY_LABEL_SEPARATOR not in entity
            )
```

rasa/shared/core/domain.py
```diff
@@ -910,6 +1054,37 @@ def _transform_intents_for_file(self) -> List[Union[Text, Dict[Text, Any]]]:
 
         return intents_for_file
 
+    def _transform_entities_for_file(self) -> List[Union[Text, Dict[Text, Any]]]:
+        """Transform entity properties for displaying or writing into a domain file.
```
ricwo, line 1058:
```suggestion
        """Transform entity properties for displaying or writing to a domain file.
```

rasa/shared/core/events.py
```diff
@@ -414,6 +417,21 @@ def as_sub_state(self) -> Dict[Text, Union[None, Text, List[Optional[Text]]]]:
             a dictionary with intent name, text and entities
         """
         entities = [entity.get(ENTITY_ATTRIBUTE_TYPE) for entity in self.entities]
+        entities.extend(
+            [
+                f"{entity.get(ENTITY_ATTRIBUTE_TYPE)}{ENTITY_LABEL_SEPARATOR}{entity.get(ENTITY_ATTRIBUTE_ROLE)}"
+                for entity in self.entities
+                if ENTITY_ATTRIBUTE_ROLE in entity
+            ]
+        )
```
ricwo, line 426:
```suggestion
        entities.extend(
            (
                f"{entity.get(ENTITY_ATTRIBUTE_TYPE)}{ENTITY_LABEL_SEPARATOR}"
                f"{entity.get(ENTITY_ATTRIBUTE_ROLE)}"
            )
            for entity in self.entities
            if ENTITY_ATTRIBUTE_ROLE in entity
        )
```

rasa/shared/core/events.py
```diff
@@ -414,6 +417,21 @@ def as_sub_state(self) -> Dict[Text, Union[None, Text, List[Optional[Text]]]]:
...
+        entities.extend(
+            [
+                f"{entity.get(ENTITY_ATTRIBUTE_TYPE)}{ENTITY_LABEL_SEPARATOR}{entity.get(ENTITY_ATTRIBUTE_GROUP)}"
+                for entity in self.entities
+                if ENTITY_ATTRIBUTE_GROUP in entity
+            ]
+        )
```
ricwo, line 433:
```suggestion
        entities.extend(
            (
                f"{entity.get(ENTITY_ATTRIBUTE_TYPE)}{ENTITY_LABEL_SEPARATOR}"
                f"{entity.get(ENTITY_ATTRIBUTE_GROUP)}"
            )
            for entity in self.entities
            if ENTITY_ATTRIBUTE_GROUP in entity
        )
```

rasa/shared/core/events.py
```diff
@@ -457,14 +475,13 @@ def as_story_string(self, e2e: bool = False) -> Text:
...
+            parse_string = f"{self.intent.get(INTENT_NAME_KEY, '')}{ent_string}"
+
             if e2e:
                 message = md_format_message(
                     self.text, self.intent.get(INTENT_NAME_KEY), self.entities
                 )
-                return "{}: {}".format(self.intent.get(INTENT_NAME_KEY), message)
+                return f"{self.intent.get(INTENT_NAME_KEY)}: {message}"
             else:
                 return parse_string
```
ricwo, line 486:
```suggestion

            return parse_string
```

rasa/utils/schemas/domain.yml
```diff
@@ -18,7 +18,18 @@ mapping:
   entities:
     type: "seq"
```
ricwo, line 19:
```suggestion
    type: "seq"
    matching: "any"
```

rasa/shared/core/domain.py
```diff
@@ -352,15 +375,56 @@ def retrieval_intents(self) -> List[Text]:
...
+                entities.append(entity)
+            elif isinstance(entity, Dict):
+                for _entity, sub_labels in entity.items():
+                    entities.append(_entity)
+                    if ENTITY_ROLES_KEY in sub_labels:
+                        roles[_entity] = sub_labels[ENTITY_ROLES_KEY]
+                    if ENTITY_GROUPS_KEY in sub_labels:
+                        groups[_entity] = sub_labels[ENTITY_GROUPS_KEY]
+            else:
+                raise InvalidDomain(
```
ricwo, line 407: when would you expect the type to be something other than `dict` or `str`? shouldn't that be caught in the validation before?

rasa/shared/core/domain.py
```diff
@@ -352,15 +375,56 @@ def retrieval_intents(self) -> List[Text]:
...
+            A dictionary of entity names to groups.
+        """
+        entities: List[Text] = []
+        roles: Dict[Text, List[Text]] = {}
+        groups: Dict[Text, List[Text]] = {}
+
+        for entity in domain_entities:
+            if isinstance(entity, str):
+                entities.append(entity)
+            elif isinstance(entity, Dict):
```
ricwo, line 399:
```suggestion
            elif isinstance(entity, dict):
```

ricwo (approved):
🎉

docs/docs/nlu-training-data.mdx
```diff
@@ -246,8 +246,41 @@ For example, you should  include examples like `fly TO y FROM x`, not only `fly
...
 
-### BILOU Entity Tagging
+If you want to influence the dialogue predictions by roles or groups, you need to modify your stories to contain
+the desired role or group label. You also need to list the corresponding roles and groups of an entity in your
+[domain file](./domain.mdx#entities).
+
+Let's assume you want to output a different sentence depending on what the user's location is. E.g.
+if the user just arrived from London, you might want to ask how the trip to London was. But if the user is on the way
+to Madrid, you might want to wish the user a good stay. You can achieve this with the
+following two stories.
```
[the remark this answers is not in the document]
tabergma: Don't think it is needed as it is basically the same principal.

rasa/shared/core/domain.py
tabergma: Yes, it should be caught before.

rasa/shared/core/domain.py
tabergma: the order of the entities should not matter, they are converted into a one-hot-encoding later on with a fixed entity to index mapping

PR #6979, opened 2020-10-09
Return all sessions for GET /story

### LINKED ISSUE #6860: /story endpoint does not fetch full tracker
**Issue**: The `GET /story` endpoint currently only fetches the conversation after the last session start. The endpoint should
1. fetch the story for the entire tracker
2. not cause a new session start sequence if the session expiration time has elapsed (this endpoint should not modify the tracker)

ricwo:
**Proposed changes**:
- fixes #6860 


changelog/6860.improvement.md
```diff
@@ -0,0 +1,12 @@
+Fetching test stories from the HTTP API endpoint 
+`GET /conversations/<conversation_id>/story` no longer triggers an update 
+of the conversation session.
```
wochinge, line 3: optional: Link to docs for conversation sessions

changelog/6860.improvement.md
```diff
@@ -0,0 +1,12 @@
+Fetching test stories from the HTTP API endpoint 
+`GET /conversations/<conversation_id>/story` no longer triggers an update
```
wochinge, line 2: optional: link to the API spec docs for this endpoint

rasa/core/processor.py
```diff
@@ -42,11 +43,14 @@
...
 from rasa.core.nlg import NaturalLanguageGenerator
 from rasa.core.policies.ensemble import PolicyEnsemble
-from rasa.core.tracker_store import TrackerStore
+import rasa.core.tracker_store
 from rasa.shared.core.trackers import DialogueStateTracker, EventVerbosity
 from rasa.shared.nlu.constants import INTENT_NAME_KEY
 from rasa.utils.endpoints import EndpointConfig
 
+if TYPE_CHECKING:
+    from rasa.core.tracker_store import TrackerStore
```
wochinge, line 52: doesn't really make sense imo as we already import the entire module in line 46

rasa/core/processor.py
```diff
@@ -190,14 +194,12 @@ async def get_tracker_with_session_start(
         """
 
         tracker = self.get_tracker(sender_id)
-        if not tracker:
```
wochinge, line 193: 🚀

rasa/core/processor.py
```diff
@@ -213,10 +215,39 @@ def get_tracker(self, conversation_id: Text) -> Optional[DialogueStateTracker]:
             conversation.
         """
         conversation_id = conversation_id or DEFAULT_SENDER_ID
```
wochinge, line 217: do we actually need this line? Or should it be `Optional[Text]` for `conversation_id`?

rasa/core/tracker_store.py
```diff
@@ -1115,3 +1113,22 @@ def _load_from_module_name_in_endpoint_config(
...
+    """Context manager that ensures `tracker_store` fetches all tracker sessions.
+
+    Args:
+        tracker_store: Tracker store to modify.
+    """
+    previous_value = tracker_store.load_events_from_previous_conversation_sessions
+
+    tracker_store.load_events_from_previous_conversation_sessions = True
+
+    try:
```
wochinge, line 1131: I'm a bit worried about the implementation as generator. Somebody might see and use this function with an `await` expression inside which would have side effects on other requests which are using trackers.

* Why not making it atomic - means returning the `DialogueStateTracker` right away?
* An alternative would be to add a new function `retrieve_full_tracker` to `TrackerStore` and implement this for the `TrackerStore` class and the trackers which are supporting sessions. By default it could just point to `retrieve`

rasa/core/processor.py
```diff
@@ -213,10 +215,39 @@ def get_tracker(self, conversation_id: Text) -> Optional[DialogueStateTracker]:
...
+        for `conversation_id`.
+
+        Args:
+            conversation_id: The ID of the conversation for which the trackers should
+                be retrieved.
+
+        Returns:
+            Trackers for the conversation.
+        """
+        import rasa.shared.core.trackers
```
wochinge, line 238: it's already imported as top level import. A local import doesn't add any benefit then.

rasa/shared/core/trackers.py
```diff
@@ -813,3 +828,56 @@ def is_prev_action_listen_in_state(state: State) -> bool:
     """
     prev_action_name = state.get(PREVIOUS_ACTION, {}).get(ACTION_NAME)
     return prev_action_name == ACTION_LISTEN_NAME
+
+
+def get_trackers_for_conversation_sessions(
```
wochinge, line 833: Can we consolidate this one with [_split_conversation_at_restarts](https://github.com/RasaHQ/rasa/blob/cee30cbacf1655f4796d49d15a9ad56a6ec2dc5e/rasa/core/training/interactive.py#L747)?

rasa/shared/core/trackers.py
```diff
@@ -813,3 +828,56 @@ def is_prev_action_listen_in_state(state: State) -> bool:
...
+        tracker: Instance of `DialogueStateTracker` to split.
+
+    Returns:
+        The trackers split by conversation sessions.
+    """
+    trackers: List["DialogueStateTracker"] = []
+    tracker_slots = tracker.slots
+
+    applied_events = tracker.applied_events(
+        include_restarts=True, include_previous_conversation_sessions=True
```
wochinge, line 848: I think these flags dilute the meaning of `applied_events` a bit. What about splitting at the restarts / session starts and calling `applied_events` on each tracker?

rasa/shared/core/trackers.py
```diff
@@ -130,7 +131,7 @@ def from_events(
         cls,
         sender_id: Text,
         evts: List[Event],
-        slots: Optional[List[Slot]] = None,
+        slots: Optional[Iterable[Slot]] = None,
```
wochinge, line 134: 😍

rasa/server.py
```diff
@@ -227,27 +227,85 @@ def event_verbosity_parameter(
...
-    # `_validate_tracker` ensures we can't return `None` so `Optional` is not needed
-    return tracker  # pytype: disable=bad-return-type
+    Args:
+        processor: An instance of `MessageProcessor`.
+        conversation_id: Conversation ID to fetch the tracker for.
 
+    Returns:
+        The tracker for `conversation_id` with an updated conversation session.
+    """
+    return await processor.get_tracker_with_session_start(conversation_id)
```
wochinge, line 242: I don't think it's intuitive that a `get_` function has side effects not only on the objects but even on the persistence layer. However, somewhat out of the scope of this PR 😬

rasa/server.py
```diff
@@ -625,7 +684,7 @@ async def execute_action(request: Request, conversation_id: Text):
 
         try:
             async with app.agent.lock_store.lock(conversation_id):
-                tracker = await get_tracker(
+                tracker = await get_tracker_with_session_start(
```
wochinge, line 687: This collides with the [API spec](https://rasa.com/docs/rasa/pages/http-api#operation/predictConversationAction) which states

> The state of the tracker is not modified.

I think it makes sense to keep the validation of an empty tracker for this (and to not modify the sessions)

rasa/server.py
```diff
@@ -227,27 +227,85 @@ def event_verbosity_parameter(
...
+        fetch_all_sessions: Whether to fetch stories for all conversation sessions.
+            If `False`, only the last conversation session is retrieved.
+
+    Returns:
+        The stories for `conversation_id` in test format.
+    """
+    if fetch_all_sessions:
+        trackers: List[
+            DialogueStateTracker
+        ] = processor.get_trackers_for_all_conversation_sessions(conversation_id)
```
wochinge, line 267: shouldn't we raise in case the `conversation_id` is invalid?

rasa/server.py
```diff
@@ -227,27 +227,85 @@ def event_verbosity_parameter(
...
+
+    more_than_one_story = len(trackers) > 1
+
+    for i, tracker in enumerate(trackers, 1):
+        for story_step in tracker.as_story().story_steps:
+            story_step.block_name = conversation_id
+            if more_than_one_story:
+                story_step.block_name += f", story {i}"
+
+            story_steps.append(story_step)
```
wochinge, line 291: ?
```suggestion
    for i, tracker in enumerate(trackers, 1):
        if more_than_one_story:
            tracker.sender_id = f"{tracker.sender_id}, story {i}"
        story_steps += tracker.as_story().story_steps
```

wochinge (commented):
Juicy PR in the middle of some technical debt 🚀 
I made a lot of comments - most of them are discussion items.

rasa/core/processor.py
ricwo: I would keep it as it is in order to cover `conversation_id = ''`

rasa/shared/core/trackers.py
ricwo: I don't see the benefit in trying to combine the two - the one in `interactive.py` works on serialised events, this one doesn't

rasa/core/tracker_store.py
ricwo: I'm not sure I follow - why would someone use this in an `await` expression? It is a context manager - the decorator and the function description says so.

I could also add an explicit example of how to use it in the function docstring. what do you think?

rasa/shared/core/trackers.py
ricwo: great idea!

rasa/server.py
ricwo: yeah we should but this goes for all endpoints that depend on the `conversation_id`. I would suggest handling this separately, for example in a decorator that uses the processor to validate that `conversation_id` exists

ricwo: no hang on, the function was just badly named before: 

https://github.com/RasaHQ/rasa/pull/6979/files#diff-394c9a945911d29e59b1eae237cdf29d7bc1276940b6dfc2cd95dd5a51117143L230

it was still executing the session start

rasa/server.py
ricwo: no hang on, the function was just badly named before:

https://github.com/RasaHQ/rasa/pull/6979/files#diff-394c9a945911d29e59b1eae237cdf29d7bc1276940b6dfc2cd95dd5a51117143L230

it was still executing the session start

ricwo: I guess fixing that should happen in a separate PR

rasa/core/tracker_store.py
wochinge: I meant the following use case:

```python
        with rasa.core.tracker_store.tracker_store_with_full_conversation_retrieval(
            self.tracker_store
        ):
            tracker = self.tracker_store.get_or_create_tracker(
                conversation_id, append_action_listen=False
            )
            # imagine any `async` call here
            # the await will allow other requests to be processed while this is blocking
            # These requests will then use a tracker store with disabled sessions.
            await asyncio.sleep(123)
```

rasa/server.py
wochinge: Sounds good. Can you please create an issue for that?

rasa/shared/core/trackers.py
wochinge: I don't think we should duplicate the logic to split conversations. E.g. we could make the one in `interactive` work on serialized events, too. It's not ideal for performance, but it will be called just once, so it's fine.

rasa/server.py
wochinge: I think it would be great to tackle it right away as we're otherwise never gonna tackle it, but the PR already has a decent size. Can you please create an issue? I think it would be good for someone who is getting started at Rasa.

rasa/core/tracker_store.py
ricwo: I don't think so - other calls to `self.tracker_store` within that class will happen _outside_ the context manager, so should not be affected by this

rasa/shared/core/trackers.py
ricwo: but the split is a very different one. in interactive we split on the `restart` event - here we split on `ActionExecuted(action_session_start)`. All of this could be parametrised in the function but I don't see the benefit in doing that, especially since we'd also have to deserialise events in the interactive case

rasa/core/tracker_store.py
ricwo: I agree with the suggestion to make this a tracker store method 👍

rasa/server.py
ricwo: https://github.com/RasaHQ/rasa/issues/7022

rasa/server.py
ricwo: will handle this in a separate PR: https://github.com/RasaHQ/rasa/pull/7020

rasa/shared/core/trackers.py
ricwo: Okay I've just convinced myself 😄   I've moved the split

rasa/core/tracker_store.py
```diff
@@ -68,38 +66,51 @@ def __init__(
...
+        # TODO: Remove this in Rasa Open Source 3.0
+        self.retrieve_events_from_previous_conversation_sessions: Optional[bool] = None
+        self._set_deprecated_kwargs_and_emit_warning(kwargs)
+
+    def _set_deprecated_kwargs_and_emit_warning(self, kwargs: Dict[Text, Any]) -> None:
+        retrieve_events_from_previous_conversation_sessions = kwargs.get(
+            "retrieve_events_from_previous_conversation_sessions"
         )
 
+        if retrieve_events_from_previous_conversation_sessions:
```
wochinge, line 92: `False` shouldn't be passed in any longer by default right? So any usage of this flag should be deprecated.
```suggestion
        if retrieve_events_from_previous_conversation_sessions is not None:
```

rasa/core/tracker_store.py
```diff
@@ -164,6 +175,22 @@ def retrieve(self, sender_id: Text) -> Optional[DialogueStateTracker]:
         """Retrieve method that will be overridden by specific tracker"""
         raise NotImplementedError()
 
+    def retrieve_full_tracker(
```
wochinge, line 178: 🙌

rasa/core/tracker_store.py
```diff
@@ -272,14 +303,15 @@ def __init__(
         event_broker: Optional[EventBroker] = None,
         record_exp: Optional[float] = None,
         use_ssl: bool = False,
+        **kwargs: Dict[Text, Any],
     ):
```
wochinge, line 307:
```suggestion
    ) -> None:
```

rasa/core/tracker_store.py
```diff
@@ -558,11 +596,46 @@ def retrieve(self, sender_id: Text) -> Optional[DialogueStateTracker]:
...
+            return self.retrieve_full_tracker(sender_id)
+
+        events = self._retrieve(sender_id, fetch_events_from_all_sessions=False)
+
         return DialogueStateTracker.from_dict(sender_id, events, self.domain.slots)
 
+    def retrieve_full_tracker(
+        self, conversation_id: Text
+    ) -> Optional[DialogueStateTracker]:
+        """Retrieves tracker containing all conversation sessions.
```
wochinge, line 625: Optional: Imo it's not required to re-specify docstrings for methods of the abstract base class which are overriden.

rasa/shared/core/events.py
```diff
@@ -103,6 +102,69 @@ def first_key(d: Dict[Text, Any], default_key: Any) -> Any:
...
+        if not isinstance(evt, event_type_to_split_on):
+            return False
+
+        # the type is correct and there are no further conditions
+        if not additional_splitting_conditions:
+            return True
+
+        # there are further conditions - check those
+        return all(
+            getattr(evt, k) == v for k, v in additional_splitting_conditions.items()
```
wochinge, line 148: That would avoid an `AttributeError` in case the event for some reason doesn't have this attribute yet.
```suggestion
            getattr(evt, k, None) == v for k, v in additional_splitting_conditions.items()
```

wochinge (approved):
Nice! I think the new `TrackerStore` method makes it way cleaner 🥇

PR #6996, opened 2020-10-12
Docker: Suggest mounting /tmp as tmpfs

orgads:
Without this, rasa leaves behind tons of temp files when it unpacks the models.

### DIFF
--- a/docs/docs/docker/building-in-docker.mdx
+++ b/docs/docs/docker/building-in-docker.mdx
@@ -33,7 +33,7 @@ The only difference is that you'll be running Rasa inside a Docker container, us
 the image `rasa/rasa`. To initialize your project, run:
 
 <pre><code parentName="pre" className="language-bash">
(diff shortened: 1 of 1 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @amn41 will take a look at it as soon as possible ✨

tczekajlo (changes requested):
@orgads Thanks a lot for the changes 😄  I left one comment. Additionally, the model unpacking issue shouldn’t happen as of 1.10.15 or 2.0.0:  https://github.com/RasaHQ/rasa/pull/6925

@erohmensing I have doubts if we should suggest something that has platform limitations. Especially, if the issue unpacking the model is fixed. 

What do you think?

tczekajlo:
@erohmensing Taking intro account the platform limitation on which `tmpfs` can be used and the fact that the issue with unpacking the model is fixed I'm gonna drop the PR.  

@orgads Thank you a lot for your PR, and I'm looking forward to seeing your next PRs 😄  🙌

orgads:
Yes, the issue seems to be fixed. Thank you.

PR #6999, opened 2020-10-13
Redirections Redux

lunelson:
This PR changes the build path of the `docusaurus build` command to match the subpath configured as `baseUrl` in `docusarus.config.js`. Doing this avoids the additional rewrite of sub-directory paths back to the root, and enables content 301 redirections to be maintained in this repo, rather than in `rasa-website`.

This corresponds to [this PR on rasa-website](https://github.com/RasaHQ/rasa-website/pull/863), which should be merged first.

NOTE: the `/docs/static/_redirects` file has been replaced by a `/docs/netlify.toml` instead, because due to the changed build path it would now end up in a subfolder where it wouldn't work. Depending on how the build/deploy is scripted, `netlify.toml` may not work. In this case we could revert to `_redirects`, but would have to place the `_redirects` file manually at the root of the build, as part of the script, not in `/static/`.


m-vdb:
docs build on master branch is currently broken due to https://github.com/RasaHQ/rasa/pull/7093 (will be temporarily fixed once the [2.0.x branch is merged into master](https://github.com/RasaHQ/rasa/pull/7092) though)

m-vdb (approved):
👌

PR #7012, opened 2020-10-14
enrich error messages from contradicting rules

Ghostvv:
**Proposed changes**:
- hopefully make error messages from contradicting rules clear

### DIFF
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -58,13 +58,12 @@ DEFAULT_ACTION_MAPPINGS = {
 
 RULES = "rules"
 RULES_FOR_LOOP_UNHAPPY_PATH = "rules_for_loop_unhappy_path"
-RULES_SOURCES = "rules_sources"
 
 DO_NOT_VALIDATE_LOOP = "do_not_validate_loop"
 DO_NOT_PREDICT_LOOP_ACTION = "do_not_predict_loop_action"
 
-DEFAULT_RULES = "default_rules"
-LOOP_RULES = "loop_rules"
+DEFAULT_RULES = "predicting default action"
+LOOP_RULES = "handling active loops and forms"
 
 
 class InvalidRule(RasaException):
@@ -134,6 +133,7 @@ class RulePolicy(MemoizationPolicy):
         self._check_for_contradictions = check_for_contradictions
 
         self._prediction_source = None
+        self._rules_sources = None
 
         # max history is set to `None` in order to capture any lengths of rule stories
         super().__init__(
@@ -441,8 +441,14 @@ class RulePolicy(MemoizationPolicy):
         if collect_sources:
             # we need to remember which action should be predicted by the rule
             # in order to correctly output the names of the contradicting rules
-            self.lookup[RULES_SOURCES][self._prediction_source].append(
-                (tracker.sender_id, gold_action_name)
+            rule_name = tracker.sender_id
+            if self._prediction_source in {DEFAULT_RULES, LOOP_RULES}:
+                # the real gold action contradict the one in the rules in this case
+                gold_action_name = predicted_action_name
+                rule_name = self._prediction_source
+
+            self._rules_sources[self._prediction_source].append(
+                (rule_name, gold_action_name)
             )
             return
 
@@ -452,9 +458,7 @@ class RulePolicy(MemoizationPolicy):
         tracker_type = "rule" if tracker.is_rule_tracker else "story"
         contradicting_rules = {
             rule_name
-            for rule_name, action_name in self.lookup[RULES_SOURCES][
-                self._prediction_source
-            ]
+            for rule_name, action_name in self._rules_sources[self._prediction_source]
             if action_name != gold_action_name
         }
 
@@ -477,7 +481,7 @@ class RulePolicy(MemoizationPolicy):
         collect_sources: bool,
     ) -> List[Text]:
         if collect_sources:
-            self.lookup[RULES_SOURCES] = defaultdict(list)
+            self._rules_sources = defaultdict(list)
 
         error_messages = []
         pbar = tqdm(
@@ -795,6 +799,8 @@ class RulePolicy(MemoizationPolicy):
         else:
             logger.debug("There is no applicable rule.")
 
+        # if we didn't predict anything from the rules, then the feature key created
+        # from states can be used as an indicator that this state will lead to fallback
         return predicted_action_name, best_rule_key or self._create_feature_key(states)
 
     def predict_action_probabilities(
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -450,13 +450,13 @@ class RulePolicy(MemoizationPolicy):
             return
 
         tracker_type = "rule" if tracker.is_rule_tracker else "story"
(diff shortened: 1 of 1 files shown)


Ghostvv:
@akelad could you please review new error messages

rasa/core/policies/rule_policy.py
```diff
@@ -415,40 +423,62 @@ def _check_prediction(
...
+                self._prediction_source
+            ]
+            if action_name != gold_action_name
+        )
+
+        error_message = (
             f"- the prediction of the action '{gold_action_name}' in {tracker_type} "
-            f"'{tracker.sender_id}' is contradicting with another rule or story."
+            f"'{tracker.sender_id}' "
+            f"is contradicting with rules '{', '.join(contradicting_rules)}'"
```
akelad, line 464:
```suggestion
            f"is contradicting with rule(s) '{', '.join(contradicting_rules)}'"
```

akelad:
looks good to me (with the one tiny change)! just tested out locally as well. one thing that i noticed is that the TEDPolicy trained before this error showed up - shouldn't we check this before training any sort of policy?

Ghostvv:
@akelad we have a separate issue to train `RulePolicy` first, however I couldn't find it (cc @wochinge)

rasa/core/policies/rule_policy.py
```diff
@@ -415,40 +423,62 @@ def _check_prediction(
...
                 tracker, domain, interpreter
             )
-            if not predicted_action_name or predicted_action_name == gold_action_name:
-                return None
+
+        if collect_sources:
+            # we need to remember which action should be predicted by the rule
+            # in order to correctly output the names of the contradicting rules
+            self.lookup[RULES_SOURCES][self._prediction_source].append(
+                [tracker.sender_id, gold_action_name]
```
wochinge, line 445:
```suggestion
                (tracker.sender_id, gold_action_name)
```

rasa/core/policies/rule_policy.py
```diff
@@ -415,40 +423,62 @@ def _check_prediction(
...
 
         tracker_type = "rule" if tracker.is_rule_tracker else "story"
-        return (
+        contradicting_rules = set(
+            rule_name
+            for rule_name, action_name in self.lookup[RULES_SOURCES][
+                self._prediction_source
+            ]
+            if action_name != gold_action_name
+        )
```
wochinge, line 459:
```suggestion
        contradicting_rules = {
            rule_name
            for rule_name, action_name in self.lookup[RULES_SOURCES][
                self._prediction_source
            ]
            if action_name != gold_action_name
        }
```

rasa/core/policies/rule_policy.py
```diff
@@ -734,7 +795,7 @@ def _find_action_from_rules(
         else:
             logger.debug("There is no applicable rule.")
 
-        return predicted_action_name
+        return predicted_action_name, best_rule_key or self._create_feature_key(states)
```
wochinge, line 798: let's add a comment why `self._create_feature_key(states)` is used a fallback source

rasa/core/policies/rule_policy.py
```diff
@@ -415,40 +423,62 @@ def _check_prediction(
...
-        logger.debug("Started checking rules and stories for contradictions.")
-        # during training we run `predict_action_probabilities` to check for
-        # contradicting rules.
-        # We silent prediction debug to avoid too many logs during these checks.
-        logger_level = logger.level
-        logger.setLevel(logging.WARNING)
+        collect_sources: bool,
+    ) -> List[Text]:
+        if collect_sources:
+            self.lookup[RULES_SOURCES] = defaultdict(list)
```
wochinge, line 480: why do we store this in `self.lookup`? Let's just use a new class attribute of this instead.

rasa/core/policies/rule_policy.py
```diff
@@ -415,40 +423,62 @@ def _check_prediction(
...
+            and predicted_action_name == tracker.active_loop_name
+        ):
             rasa.core.test.emulate_loop_rejection(tracker)
             predicted_action_name = self._predict_next_action(
                 tracker, domain, interpreter
             )
-            if not predicted_action_name or predicted_action_name == gold_action_name:
-                return None
+
+        if collect_sources:
```
wochinge, line 441: Can't we move up lines 441 to 447 to the top of the method to skip the actual prediction part? Should be a faster.

wochinge: or making it an `if else` thin within `_run_prediction_on_trackers`?

wochinge (commented):
🚀

rasa/core/policies/rule_policy.py
Ghostvv: Then we will automatically persist it. And I think it can be a great origin for logging information for debugging rules during prediction. I thought of adding it in different PR, but can do it in this PR as well

wochinge: > Then we will automatically persist it. 

We can also persist it if it's a separate attribute. Storing everything in this dictionary seems very hacky.

Ghostvv: why would we create additional persistence of smth that is already persisted. However, it might be better to update original `self.lookup[RULES]` with origin of the rule in addition to the action name

rasa/core/policies/rule_policy.py
Ghostvv: the actual prediction part is necessary, because `predict_action_probabilities` updates `self._prediction_source` that is used in new lookup

Ghostvv: and we cannot simply call `predict_action_probabilities` because we need to emulate form rejection

rasa/core/policies/rule_policy.py
Ghostvv: I renamed it to private variable `self._rules_sources`

rasa/core/policies/rule_policy.py
```diff
@@ -415,40 +423,66 @@ def _check_prediction(
...
+
+        if not predicted_action_name or predicted_action_name == gold_action_name:
+            return
 
         tracker_type = "rule" if tracker.is_rule_tracker else "story"
-        return (
+        contradicting_rules = {
+            rule_name
+            for rule_name, action_name in self._rules_sources[self._prediction_source]
+            if action_name != gold_action_name
```
wochinge, line 462: This avoids error in case a rule contains several actions, doesn't it?

```suggestion
            if action_name != gold_action_name and tracker.sender_id != rule_name
```

wochinge (approved):
I left one more comment. Besides, ready to go 👍

rasa/core/policies/rule_policy.py
Ghostvv: Users might create 2 rules with the same name, but different actions, then we should still output the rule name

rasa/core/policies/rule_policy.py
```diff
@@ -415,40 +423,66 @@ def _check_prediction(
...
+        }
+
+        error_message = (
             f"- the prediction of the action '{gold_action_name}' in {tracker_type} "
```
wochinge, line 466: makes more sense this way, doesn't it?
```suggestion
            f"- the prediction of the action '{predicted_action_name}' in {tracker_type} "
```

PR #7018, opened 2020-10-15
Create correct KafkaProducer for PLAINTEXT and SASL_SSL security protocols.

alwx:
**Proposed changes**:
- update `_create_producer` function in `kafka.py` to create correct `KafkaProducer`s

### DIFF
--- a/changelog/7018.bugfix.md
+++ b/changelog/7018.bugfix.md
@@ -1,0 +1,1 @@
+Create correct `KafkaProducer` for `PLAINTEXT` and `SASL_SSL` security protocols.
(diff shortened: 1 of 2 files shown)


wochinge (commented):
Thanks for the fix 🙌 

- Is this related to an open issue?
- Can we please remove `group_id` from the `KafkaEventBroker` constructor as it doesn't make sense for an `KafkaProducer`
- Can we add `client_id` to every `KafkaProducer` we create?

alwx:
@wochinge no, it's not linked to any issue — we merged some updates to KafkaProducer about a week ago, and now we discovered that there is an issue with those updates.

wochinge (approved):
🥇 
Can you please target `2.0.x` instead of `master`?

PR #7020, opened 2020-10-15
Setting slots via API does not influence conversation

### LINKED ISSUE #6721: Setting slots before conversation starts leads to empty bot responses

ricwo:
**Proposed changes**:
- fixes #6721 
- the bug comes from the fact that previously no session start was run if slots were appended to a new tracker
- this manifests itself in different `prediction_states` depending on whether or not the session start sequence was run at the beginning of the tracker
- this leads to different policy behaviour - in the linked issue, it makes the difference between the memo policy kicking in or not
- this PR fixes this behaviour by making sure that when appending events to a new tracker, there is always a session start

### DIFF
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -178,7 +178,7 @@ class MessageProcessor:
         output_channel: Optional[OutputChannel] = None,
         metadata: Optional[Dict] = None,
     ) -> DialogueStateTracker:
-        """Get tracker for `sender_id` or create a new tracker for `sender_id`.
+        """Fetch tracker for `sender_id` and update its conversation session.
 
         If a new tracker is created, `action_session_start` is run.
 
@@ -188,7 +188,7 @@ class MessageProcessor:
             sender_id: Conversation ID for which to fetch the tracker.
 
         Returns:
-              Tracker for `sender_id` if available, `None` otherwise.
+              Tracker for `sender_id` if available.
         """
 
         tracker = self.get_tracker(sender_id)
@@ -211,9 +211,8 @@ class MessageProcessor:
             sender_id: Conversation ID for which to fetch the tracker.
 
         Returns:
-              Tracker for `sender_id` if available, `None` otherwise.
+              Tracker for `sender_id`.
         """
-
         tracker = self.get_tracker(sender_id)
 
         # run session start only if the tracker is empty
@@ -815,7 +814,6 @@ class MessageProcessor:
         Returns:
             `True` if the session in `tracker` has expired, `False` otherwise.
         """
-
         if not self.domain.session_config.are_sessions_enabled():
             # tracker has never expired if sessions are disabled
             return False
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -189,6 +189,36 @@ class MessageProcessor:
 
         return tracker
 
+    async def get_tracker_with_session_start(
+        self,
+        sender_id: Text,
+        output_channel: Optional[OutputChannel] = None,
+        metadata: Optional[Dict] = None,
+    ) -> DialogueStateTracker:
+        """Fetches tracker for `sender_id` and updates its conversation session.
+
+        If a new tracker is created, `action_session_start` is run.
+
+        Args:
+            metadata: Data sent from client associated with the incoming user message.
+            output_channel: Output channel associated with the incoming user message.
+            sender_id: Conversation ID for which to fetch the tracker.
+
+        Returns:
+              Tracker for `sender_id`.
+        """
+        # TODO: Remove in Rasa Open Source 3.0
+        rasa.shared.utils.io.raise_warning(
+            "`get_tracker_with_session_start()` is deprecated and will be removed "
+            "in Rasa Open Source version 3.0. Please use "
+            "`fetch_tracker_and_update_session()` instead.",
+            category=DeprecationWarning,
+        )
+
+        return await self.fetch_tracker_and_update_session(
+            sender_id, output_channel, metadata
+        )
+
     async def fetch_tracker_with_initial_session(
         self,
         sender_id: Text,
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -189,36 +189,6 @@ class MessageProcessor:
 
         return tracker
 
-    async def get_tracker_with_session_start(
-        self,
-        sender_id: Text,
-        output_channel: Optional[OutputChannel] = None,
-        metadata: Optional[Dict] = None,
-    ) -> DialogueStateTracker:
-        """Fetches tracker for `sender_id` and updates its conversation session.
-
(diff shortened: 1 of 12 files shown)


rasa/core/processor.py
```diff
@@ -188,19 +205,49 @@ async def get_tracker_with_session_start(
             sender_id: Conversation ID for which to fetch the tracker.
 
         Returns:
-              Tracker for `sender_id` if available, `None` otherwise.
+              Tracker for `sender_id`.
         """
+        # TODO: Remove in Rasa Open Source 3.0
+        rasa.shared.utils.io.raise_warning(
```
wochinge, line 211: I'd personally not consider the `processor` module a public API (in contrast to `Agent`). So imo fine to just drop it. What do you think?

rasa/core/processor.py
```diff
@@ -172,13 +165,37 @@ async def _update_tracker_session(
                 metadata=metadata,
             )
 
+    async def fetch_tracker_and_update_session(
```
wochinge, line 168: optional optional: All methods use `fetch_tracker` as prefix except `get_tracker` which is slightly inconsistent.

rasa/core/tracker_store.py
```diff
@@ -282,9 +281,9 @@ def retrieve(self, sender_id: Text) -> Optional[DialogueStateTracker]:
         if sender_id in self.store:
             logger.debug(f"Recreating tracker for id '{sender_id}'")
             return self.deserialise_tracker(sender_id, self.store[sender_id])
-        else:
-            logger.debug(f"Creating a new tracker for id '{sender_id}'.")
-            return None
+
+        logger.debug(f"Creating a new tracker for id '{sender_id}'.")
```
wochinge, line 285: This comment doesn't make sense for me at this location. We should either say that there was no tracker for the user or move the log to where the tracker is actually created.

rasa/server.py
```diff
@@ -564,7 +542,9 @@ async def append_events(request: Request, conversation_id: Text):
         try:
             async with app.agent.lock_store.lock(conversation_id):
                 processor = app.agent.create_processor()
-                tracker = processor.get_tracker(conversation_id)
+                tracker = await processor.fetch_tracker_with_initial_session(
```
wochinge, line 545: I think [this piece of code](https://github.com/RasaHQ/rasa-x/blob/4e2e5a4613eef15816968389be4eb9d616f47fca/src/community/rasax/community/api/blueprints/stack.py#L219) in the Rasa X code base depends on the fact that the tracker is actually empty.

wochinge: I think we should either check if the new events already a starting sequence or not adding the starting sequence. What do you think?

wochinge (commented):
Could you please explain the bug? Currently not sure if it's not a usage error.

rasa/core/processor.py
ricwo: I've named the methods with side effects that "get and do something else" as `fetch_X_and_Y`. those that simply "get" should be called `get_X` IMO, wdyt?

rasa/core/processor.py
ricwo: 👍

rasa/core/tracker_store.py
```diff
@@ -229,13 +230,13 @@ def serialise_tracker(tracker: DialogueStateTracker) -> Text:
     def _deserialise_dialogue_from_pickle(
         sender_id: Text, serialised_tracker: bytes
     ) -> Dialogue:
-
         rasa.shared.utils.io.raise_deprecation_warning(
```
ricwo, line 233: this could have been dropped in 2.0 😬

ricwo: I'll add a note for 3.0

rasa/core/processor.py
wochinge: Is that a convention? Makes sense - not sure if other developers would immediately see `fetch_` as a warning for side effects though.

rasa/core/tracker_store.py
```diff
@@ -154,16 +157,13 @@ def create_tracker(
...
         tracker = self.init_tracker(sender_id)
 
-        if tracker:
-            if append_action_listen:
-                tracker.update(ActionExecuted(ACTION_LISTEN_NAME))
+        if append_action_listen:
+            tracker.update(ActionExecuted(ACTION_LISTEN_NAME))
 
-            self.save(tracker)
+        self.save(tracker)
```
wochinge, line 166: out of the scope of this PR: isn't this redundant? We'd save the tracker anyway later on, wouldn't we?

rasa/server.py
```diff
@@ -564,19 +571,21 @@ async def append_events(request: Request, conversation_id: Text):
...
             async with app.agent.lock_store.lock(conversation_id):
                 processor = app.agent.create_processor()
-                tracker = processor.get_tracker(conversation_id)
-                output_channel = _get_output_channel(request, tracker)
-
                 events = _get_events_from_request_body(request)
 
-                for event in events:
-                    tracker.update(event, app.agent.domain)
+                tracker = await fetch_tracker_and_update_with_events(
```
wochinge, line 576: optional / nit picking: I'm not a huge fan of `and` in function names as a functions are supposed to do one thing and in this case the name seems a bit too focused on the `how` instead of the `what` . What about something like this? 
```suggestion
                tracker = await update_conversation_with_events(
```

rasa/server.py
```diff
@@ -684,9 +693,10 @@ async def execute_action(request: Request, conversation_id: Text):
 
         try:
             async with app.agent.lock_store.lock(conversation_id):
-                tracker = await get_tracker_with_session_start(
-                    app.agent.create_processor(), conversation_id
+                tracker = await app.agent.create_processor().fetch_tracker_and_update_session(
```
wochinge, line 696: out of scope: Your change really makes it obvious that we don't really care about the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) 😆 In my had head all tracker store interaction should happen in `Agent`. The processor should only work on `tracker` events and the Agent would store the results in the tracker store once the `processor` is done.

wochinge (commented):
Nice refactoring and bug fix 🚀

wochinge (approved):

rasa/server.py
ricwo: I agree that this is odd, but as you said this is a larger refactoring

rasa/core/tracker_store.py
ricwo: you're right, this fn is only called from `server.py` which always performs a `save()` afterwards

Issue #6926, opened 2020-10-06
Rasa 2.0 version bug in windows


matthewkos:
**Rasa version**:
2.0.0rc4
**Rasa SDK version** (if used & relevant):
2.0.0rc4
**Rasa X version** (if used & relevant):
NA
**Python version**:
3.7.9
**Operating system** (windows, osx, ...):
windows 10
**Issue**:
sys.version in windows 10 does not contain \n
please fix by 

**Error (including full traceback)**:
```
Traceback (most recent call last):
  File "e:\env\anaconda\envs\rasa\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "e:\env\anaconda\envs\rasa\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "E:\Env\Anaconda\envs\rasa\Scripts\rasa.exe\__main__.py", line 7, in <module>
  File "e:\env\anaconda\envs\rasa\lib\site-packages\rasa\__main__.py", line 118, in main
    print_version()
  File "e:\env\anaconda\envs\rasa\lib\site-packages\rasa\__main__.py", line 78, in print_version
    python_version, os_info = sys.version.split("\n")
ValueError: not enough values to unpack (expected 2, got 1)
```

**Command or request that led to error**:
```
rasa --version
```

**Content of configuration file (config.yml)** (if relevant):
```yml

```

**Content of domain file (domain.yml)** (if relevant):
```yml

```

sara-tagger:
Thanks for the issue, @tttthomasssss will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

manada-jwheat:
I can second this error

Rasa version:
2.0.0 release
Rasa SDK version (if used & relevant):
2.0.0 release
Rasa X version (if used & relevant):
NA
Python version:
3.8
Operating system (windows, osx, ...):
Windows 10 / Using an anaconda environment

koaning:
Ah my bad. I should've adressed it a while ago. I've added a PR for it!

wochinge:
fixed by https://github.com/RasaHQ/rasa/pull/7023

Issue #5637, opened 2020-04-15
Easier to Post Version Numbers on Github Issues


koaning:
**Description of Problem**:

When I write a bug request, I gotta do a whole lot of accounting. 

![image](https://user-images.githubusercontent.com/1019791/79348658-918aac00-7f35-11ea-8fa1-f6341e49d8a7.png)

We need this, but this is effort from the user. And techies, they don't like to do manual labor. 

**Overview of the Solution**:

I propose that we add a verbose flag to the version command that will output information that a user can just copy. 

```
> rasa --version --verbose
**Enrivonment Info**
Darwin 19.3.0 - posix
Python 3.6.8
Rasa 1.9.6
Rasa-SDK 1.9.0
```

This information saves the user time and may make it easier for folks to give good issues. We should be able to get all the rasa packages that are installed with versions as well as the operating system in one go. Heck, we might even be able to generate a nice table format :) 

**Definition of Done**:
<!-- What needs to be there to consider this feature as done?
- [ ] Tests are added
- [ ] Feature described the docs
- [ ] Feature mentioned in the changlog

koaning:
I wouldn't mind picking this up but I would like to get a green light before attempting this.

eibhleag:
This feels like a @RasaHQ/production-engineers issue so best to double check with them 🙂

wochinge:
I think it's a cool idea and you can of course do it 👍 Let's just get on more opinion on it 🙂 @erohmensing What do you think?

erohmensing:
Yeah seems like a great idea to me! So long as it's making sure to check for things in the right environments and such :)

koaning:
I need some feedback on design. 

# Question One: How to Call it

There's multiple places in the cli where I can add this behavior and I'm not sure where we'd prefer it. 

### Behavior One 

We could add a `--verbose` flag. 

```
> rasa --version --verbose
```

This means that I have to add a flag to the global `rasa` command that is only valid when `--version` is set. Setting `--verbose` usually has an effect on the logger, which isn't the case here so this feels like an anti-pattern. 

### Behavior Two 

We could also just add this to the original command `rasa --version`. 

```
> rasa --version
Python Environment:
3.6.8 (v3.6.8:3c6b436a57, Dec 24 2018, 02:04:31) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)]
Rasa Version: 1.9.7
Rasa SDK Version: 1.9.0
[...]

PR #7023, opened 2020-10-15
#6926 Created a fix for --version on Windows

koaning:
While fixing https://github.com/RasaHQ/rasa/issues/5637 I accidentally broke the `rasa --version` command on some windows machines. It turns out that `sys.version` returns a single line of information in some older versions. It unfortunately passed our CI but there were people who got hit by it https://github.com/RasaHQ/rasa/issues/6926. 

In this PR I've written a small patch that should fix it. The only caveat is that I'm not sure how to add a test for this since the Windows CI doesn't seem to be able to pick this up.


koaning:
Closing this PR in favor of https://github.com/RasaHQ/rasa/pull/7064.

Issue #6133, opened 2020-07-02
Format tracker state for logging


melindaloubser1:
**Description of Problem**:
It would be useful if  the tracker state was displayed in logs in a way that's easier to parse.
At the moment, it's displayed in a list of dictionaries which can be hard to  read when debugging

```
[{'intent_thankyou': 1.0, 'prev_action_listen': 1.0}, {'prev_utter_noworries': 1.0, 'intent_thankyou': 1.0}, {'intent_goodbye': 1.0, 'prev_action_listen': 1.0}, {'intent_goodbye': 1.0, 'prev_utter_goodbye': 1.0}, {'intent_pay_cc': 1.0, 'prev_action_listen': 1.0}]
```

**Overview of the Solution**:
It would be easier for debugging if it was displayed similar to how slots are displayed i.e. in a human-readable list

**Definition of Done**:

Raghibshams456:
I want to solve this problem but I am not able  to understand the problem properly..can you explain me please..

rajsukanya:
I would like to work on this issue.

melindaloubser1:
@rajsukanya do you have any questions about it? Feel free to open a PR

MarthaDs:
Hi, me and @aleronupe are interesting in this issue and started to wok on it. @melindaloubser1 do you have any example of how you think the logging should look like?

melindaloubser1:
Not sold on a specific format, but it should represent a human-readable version of the state. It'll be important to keep in mind what the state really means: Each item in the list is the last action + last intent (which was not necessarily the previous step, there could be multiple actions). If the action is `action_listen`, then that intent was the previous step.  The example shown in the issue description translates to this state: 
```
- action: action_listen 1.0
- intent: thankyou 1.0
- action: utter_noworries 1.0
- action: action_listen 1.0
- intent: goodbye 1.0
- action: utter_goodbye 1.0
- action: action_listen 1.0
- intent: pay_cc 1.0
```


However, this was in Rasa 1.x. Since the release candidate for 2.x is currently out,  new features/enhancements should target 2.x. The way state is stored and shown is different for 2.x, which means this issue is probably not worth addressing in 1.x.

aleronupe:
Thank you for your answer @melindaloubser1. We will proceed to work on this issue targeting the 2.x version currently on master.
We noticed that there are a few differences on tracker state between the two versions of rasa, hence the new format is as follows:

```
[...]

PR #7026, opened 2020-10-15
Add human-readable format to tracker states logging

MarthaDs:
Co-authored-by: Martha Dantas <marthadntas@gmail.com>

**Proposed changes**:
- Improve display of tracker states to a more human-readable list. Resolve issue #6133.

**Status**:


sara-tagger:

akelad:
this looks like a cool addition! @Ghostvv @wochinge could either of you take over reviewing this to make sure everything is correct?

rasa/core/policies/memoization.py
```diff
@@ -236,6 +237,17 @@ def predict_action_probabilities(
 
         return result
 
+    def format_tracker_states(self, states):
```
Ghostvv, line 240: could you please also add the rest of the information like active loops, slots and entities?

Ghostvv: I think it is better to move this method to `Policy` superclass, then you can call `self. format_tracker_states` in all other policies

rasa/core/policies/memoization.py
```diff
@@ -226,7 +226,8 @@ def predict_action_probabilities(
 
         tracker_as_states = self.featurizer.prediction_states([tracker], domain)
         states = tracker_as_states[0]
-        logger.debug(f"Current tracker state {states}")
+        current_states = ' '.join(self.format_tracker_states(states))
```
Ghostvv, line 229: I think it makes sense to directly return a string, so that you don't need to call `" ".join()`

Ghostvv (changes requested):
Great idea thank you for implementing it. Let's add the rest of the states as well.

rasa/core/policies/memoization.py
MarthaDs: Thanks for your review! We pushed updates, now ` format_tracker_states` is into  `policy.py`. But We not sure about where the rest of the information should be displayed in logging. Would you mind telling us where do you think is a good place for it?

rasa/core/policies/memoization.py
MarthaDs: Done it ?! We still using `.join()` but now it has been used in the return of the method.

rasa/core/policies/policy.py
```diff
@@ -292,6 +292,23 @@ def _default_predictions(domain: Domain) -> List[float]:
...
 
+    def format_tracker_states(self, states: List[Dict]) -> Text:
+        """Format tracker states to human readable format on debug log
+
+        Args: 
+            states: list of tracker states dicts
+        Returns:
+            the string of the states with user intents and actions
+        """
+        formatted_states = [""]
```
Ghostvv, line 303: why do we need empty string here?
```suggestion
        formatted_states = []
```

rasa/core/policies/memoization.py
Ghostvv: the state lists might contain keys like `USER, PREVIOUS_ACTION, ACTIVE_LOOP, SLOTS` (these are constant names for strings defined in `rasa.core.constants`).
`USER` can contain keys defined here: https://github.com/RasaHQ/rasa/blob/005f8b1e4742efca12ee56194bce4f1985e73555/rasa/shared/core/events.py#L390
`PREVIOUS_ACTION` can contain keys defined here: https://github.com/RasaHQ/rasa/blob/005f8b1e4742efca12ee56194bce4f1985e73555/rasa/shared/core/events.py#L1167


If you want to reformat `states` dict, then all of possible states should be outputted and not only

rasa/core/policies/memoization.py
Ghostvv: yes, thank you. That looks good, and seems to simplify the usage of this method

rasa/core/policies/policy.py
aleronupe: Because the string returned by the function **format_tracker_states**  should begin with a line break , as it returns a join on `\n`, this empty string will guarantee that each state, even the first, will be displayed in a single line, with the following format:

```python
DEBUG rasa.core.policies.memoization  - Current tracker state:
[state 1] previous action: action_listen | user intent: greet
[state 2] previous action: utter_greet | user intent: greet
```

Instead, with an empty array to initialize the variable, the line breaks would only exist between the tracker states, resulting on the following: 

```python
DEBUG rasa.core.policies.memoization  - Current tracker state: [state 1] previous action: action_listen | user intent: greet
[state 2] previous action: utter_greet | user intent: greet
```

Ghostvv: could you please add brief comment to the code

rasa/core/policies/memoization.py
MarthaDs: @Ghostvv Sorry we took so long!!! We not sure our solution is the most elegant one but was the one that works. We will be glad if could take a look.

aleronupe: We also developed an alternate solution, focusing on specific dict keys to build the final message. We could, also, implement this same approach on a generic form, without using the specific names of the keys but iterating through the dict. Here's how we pictured it could work:

```python
def format_tracker_states(self, states: List[Dict]) -> Text:
        """Format tracker states to human readable format on debug log

        Args: 
            states: list of tracker states dicts
        Returns:
            the string of the states with user intents and actions
        """
        # empty string to insert line break before first state
        formatted_states = [""]
        if states:
            for index, state in enumerate(states):
                state_messages = []
                if state:
                    if USER in state:
                        if TEXT in state[USER]:
                            state_messages.append(
                                f"user text: {str(state[USER][TEXT])}"
                            )
                        if INTENT in state[USER]:
                            state_messages.append(
                                f"user intent: {str(state[USER][INTENT])}"
                            )
                        if ENTITIES in state[USER]:
                            state_messages.append(
                                f"user entities: {str(state[USER][ENTITIES])}"
                            )
                    if PREVIOUS_ACTION in state:
                        if ACTION_NAME in state[PREVIOUS_ACTION]:
                            state_messages.append(
                                f"previous action name: {str(state[PREVIOUS_ACTION][ACTION_NAME])}"
                            )
                        if ACTION_TEXT in state[PREVIOUS_ACTION]:
                            state_messages.append(
                                f"previous action text: {str(state[PREVIOUS_ACTION][ACTION_TEXT])}"
                            )
                    if ACTIVE_LOOP in state:
                        state_messages.append(
                            f"active loop: {str(state[ACTIVE_LOOP])}"
                        )
                    if SLOTS in state:
                        state_messages.append(
                            f"slots: {str(state[SLOTS])}"
                        )
                    state_message_formatted = " | ".join(state_messages)
                    state_formatted = f"[state {str(index)}] {state_message_formatted}" 
                    formatted_states.append(state_formatted)

        return "\n".join(formatted_states)
```

Ghostvv: the explicit approach looks good to me, we also use similar explicit keys in single state featurizer. Not sure if key agnostic for over dict will make it simpler since you need to create key specific strings anyhow

Ghostvv (commented):
looks good to me

melindaloubser1:
This looks good! @MarthaDs could you update your branch with master and fix the trailing whitespace issue?

melindaloubser1 (approved):

MarthaDs:
@melindaloubser1  All done :)

Issue #108, opened 2017-01-18
Deprecating visualizer


amn41:
The visualizer was intended as a placeholder until we had a proper GUI for creating training data. Thanks to @azazdeaz we now have one, which you can check out [here](https://azazdeaz.github.io/rasa-nlu-trainer/). 

With version 0.6 we will remove the visualizer, in favour of using this GUI directly. There will be a separate tool for converting wit/LUIS/api exports to the rasa NLU format, so you'll use it like: 

```bash
python -m rasa_nlu.convert -d wit_download.json -o rasa_data.json
```

Where the `rasa_data.json` will be ready to upload to the web app. 

Questions / comments / concerns welcome!

pilot:
👍

amn41:
PR is #122

PR #7027, opened 2020-10-15
Remove library specific tokenizers #108

koernerfelicia:
Research [issue 108](https://github.com/RasaHQ/research/issues/108)
**Proposed changes**
- Remove LanguageModelTokenizer:

I set the sub-token information in HFTransformersNLP. This requires that tokenization occurs **before** the information is set.

We used to handle this by always tokenizing with the WhitespaceTokenizer inside HFTransformersNLP. Now, the Tokenizer must be placed **before** HFTransformersNLP in the pipeline, but any Tokenizer can be used. This may be confusing to the user.

One alternative would be to move the logic of setting the sub-token information into LanguageModelFeaturizer. This requires moving several functions from HFTransformersNLP as well and reloading the model in LanguageModelFeaturizer.
 
Alternatively, the featurization and tokenization from HFTransformersNLP can be moved into LanguageModelFeaturizer, doing away with HFTransformersNLP entirely. The drawback maybe less flexibility should we want to apply features from HFTransformersNLP differently in the future. I'm not sure if this is likely, maybe @dakshvar22 knows?

- Remove ConveRTTokenizer: the sub-token information is set in ConveRTFeaturizer.

Both Tokenizers will give a deprecation warning if used, and will revert to WhitespaceTokenizer.


dakshvar22:
@koernerfelicia I started looking into the PR, I think the way how you have implemented your intended logic, it looks good.
However, I have questions on the solution approach itself 😅  - 
1. You've mentioned this as well and I agree that having the `Tokenizer` before `HFTransformersNLP` is super counter-intuitive. If we do that and decide to do away with `LanguageModelTokenizer` then it makes a lot of sense to move the complete logic of `HFTransformersNLP` inside the `LanguageModelFeaturizer` and remove `HFTransformersNLP`

2. However, we lose a bit of flexibility if we follow the suggestion in (1). Having `HFTransformersNLP` also allowed users to build a custom component based on the sub-tokens that `HFTransformersNLP` set. If that logic is moved directly inside `LanguageModelFeaturizer`, we should still make that information available inside the `Message` object somehow. Also, for some languages like Chinese whitespace tokenization does not work and users rely on either `MitieTokenizer` or the output of `LanguageModelTokenizer`. Both of them are not guaranteed to output the same set of tokens for the same input sentence.

I am not sure what's the ideal solution here but I'll think a bit more on the original problem itself - 
> certain featurizer can just be used if certain tokenizers are used in the pipeline. Due to that it is not possible to mix different word embeddings as just one tokenizer per pipeline is allowed.

What if we get around that problem by relaxing the constraint of certain tokenizers should necessarily be used for certain featurizers?

@tabergma Thoughts on the above?

tabergma:
I think we definitely need to get rid of the dependency between `LanguageModelTokenizer` and `WhitespaceTokenizer`. For example, currently it is not possible to use the `LanguageModel` components with a Chinese BERT model as the sentence are whitespace tokenized which does not work for Chinese. We cannot just remove the constraint that certain featurizers rely on certain tokenizers. So, I think moving the functionality completely to the `LanguageModelFeaturizer` is the best approach. I think it is fine if this component sets the sub-token information as well. And we can also try to decouple the functionality as much as possible so that users building a new component with a similar logic can be easily done.

koernerfelicia:
@tabergma @dakshvar22 To summarize: I'm still not sure how to change the logic so that the `Tokenizer` can come after `HFTransformersNLP` unless we replace the `HFTransformersNLP` with a `Featurizer`. Even if we figure out some solution to pass the sub-token information to the `LanguageModelFeaturizer` without attaching it to a token, the featurizing inside `HFTransformersNLP` requires some sort of tokens (e.g. see line 529 `_get_model_features_for_batch`). I think the only reasonable solutions are:
1. (currently implemented in this PR) Require the `Tokenizer` to come **before** `HFTransformersNLP` and use the tokens produced by the selected `Tokenizer` to set sub-token information, and produce features in `HFTransformersNLP`. The features are assigned to tokens in `LanguageModelFeaturizer`.
2. (like Daksh said) Move the complete logic of `HFTransformersNLP` inside the `LanguageModelFeaturizer` and remove `HFTransformersNLP`.

Which solution should I select?

tabergma:
I would go with option 2.

(Just as a note: But we need to keep the old components and deprecate them. We cannot simply remove them.)

dakshvar22:
@tabergma @koernerfelicia I agree option 2 seems much better. Just wondering about this - 
> For example, currently it is not possible to use the LanguageModel components with a Chinese BERT model as the sentence are whitespace tokenized which does not work for Chinese. 

So, if we move all the logic inside `LanguageModelFeaturizer`, then the user can only use `MitieTokenizer` for tokenization and not the tokenizer from Bert model, right?

tabergma:
Yes. For Chinese our users would be able to use the `MitieTokenizer` to tokenize their models. The `LanguageModelFeaturizer` can then use the Chinese BERT model to featurize the incoming utterance. It will first "tokenizes"  the individual tokens to get the information how many sub-tokens one token has. This information is then used to calculate the avg feature vector per token.

dakshvar22:
Okay, sounds good. 👍

koernerfelicia:
@dakshvar22 @tabergma I've moved the logic from `HFTransformersNLP` into `LanguageModelFeaturizer`.

One tricky thing -- since we are only deprecating `HFTransformersNLP` I assume we want the behaviour of pipelines containing `HFTransformersNLP` to be preserved. This means if users have specified a model in `HFTransformersNLP`, this is the model that needs to be used to featurize the dense-featurizable messages (**not** the default model in `LanguageModelFeaturizer`). I have a check in the `LanguageModelFeaturizer`,  which outputs a warning to `DEBUG `(per message), this may be too noisy, however. I wasn't able to get caplog in tests to show me anything in `TRACE`.

What do you think?

docs/docs/components.mdx
```diff
@@ -137,85 +137,6 @@ word vectors in your pipeline.
   by passing in your language identifier as the `language` option.
 
 
-### HFTransformersNLP
```
tabergma, line 140: We should not just remove the docs. I would keep it the section and add a deprecation warning, similar to https://rasa.com/docs/rasa/policies#two-stage-fallback-policy.

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -26,6 +29,10 @@
 
 logger = logging.getLogger(__name__)
 
+TF_HUB_MODULE_URL = (
+    "https://github.com/PolyAI-LDN/polyai-models/releases/download/v1.0/model.tar.gz"
+)
```
tabergma, line 34: I think we should not point to that URL anymore as it does not exist any longer. I think @dakshvar22 was working on a solution.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -1,65 +1,757 @@
...
+            if message.get(attribute):
+                self._set_lm_features(
+                    self._get_docs_for_batch(
+                        [message], attribute=attribute, inference_mode=True
+                    )[0],
+                    message,
+                    attribute,
+                )
+
+    def _set_lm_features(self, doc, message: Message, attribute: Text = TEXT) -> None:
```
tabergma, line 740: type for `doc` is missing

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -1,65 +1,757 @@
...
+            # This should only be the case if the deprecated
+            # HFTransformersNLP component is used in the pipeline
+            # TODO: remove this when HFTransformersNLP is removed for good
+            logging.debug(
+                f"{LANGUAGE_MODEL_DOCS[attribute]} set: this "
+                f"indicates you're using the deprecated component "
+                f"HFTransformersNLP, please remove it from your "
+                f"pipeline."
+            )
+            doc = hf_transformers_doc
```
tabergma, line 754: Shouldn't this check already happen before line 714 and 732? Otherwise we are constructing the doc all over again, which is double work.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -1,65 +1,757 @@
...
+        )
+
+        # A doc consists of
+        # {'token_ids': ..., 'tokens': ..., 'sequence_features': ...,
+        # 'sentence_features': ...}
+        batch_docs = []
+        for index in range(len(batch_examples)):
+            doc = {
+                TOKEN_IDS: batch_token_ids[index],
+                TOKENS: batch_tokens[index],
```
tabergma, line 672: We don't need those anymore.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
tabergma: If we move it before those lines, we could also just print the warning just once per batch. We could check if the doc is set for the first message in the batch and if yes print the warning. If it is set for one message it should be set for all.

tabergma (commented):
Just took a brief look and added some first comments. 

I think we can move the check so that we just print the warning once per batch. See inline comments. 

Please, also add a changelog entry.

docs/docs/components.mdx
koernerfelicia: I moved it to [`Deprecated Components`](
https://github.com/RasaHQ/rasa/pull/7027/files/0eca21ad7e82249ae18f07e1a27ad1fa2336d11d#diff-a1167895182b95a8cde4e98af355375597a90d92e76698823e2a61ac4d5521f8R2614)

Do you think it needs more info?

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
koernerfelicia: @dakshvar22 let me know what you'd like me to do

dakshvar22: Yeah that's a separate PR which I'll submit today, let's leave it as it is here.

docs/docs/components.mdx
tabergma: So far we just added the deprecation warning to the section of the component itself. Not sure if we should introduce a new section for it as users might be confused.

koernerfelicia: I'm not sure I understand, sorrry.

I introduced the new "Deprecated Components" section and moved the component headings as well as the warning to this new section. I think this is also how it is done in [https://rasa.com/docs/rasa/policies#two-stage-fallback-policy](https://rasa.com/docs/rasa/policies#two-stage-fallback-policy). Would you rather I do something else?

tabergma: ah, never mind, did not saw that, let's keep it as it is

dakshvar22:
@koernerfelicia We should merge this after https://github.com/RasaHQ/rasa/pull/7089 is merged into `2.0.x` branch. We can then deprecate `ConveRTTokenizer` by shifting the mandatory `model_url` parameter to `ConveRTFeaturizer` as part of this PR(I'll take that on myself).

tabergma:
@koernerfelicia Let me know if this is ready for another review round.

koernerfelicia:
@tabergma I think it is ready for another review. You might notice I moved the [check](https://github.com/RasaHQ/rasa/pull/7027#discussion_r511867149) into `_get_docs_for_batch`, rather than into `process` and `train` that way it happens before we construct the doc again (like you suggested) but the code is only in there once. This means I had to change the corresponding test as well, as it previously relied on the docs built in  . Let me know if this is unclear

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -1,65 +1,776 @@
...
+            batch_token_ids,
+            batch_tokens,
+            batch_examples,
+            attribute,
+            inference_mode,
+        )
+
+        # A doc consists of
+        # {'token_ids': ..., 'tokens': ..., 'sequence_features': ...,
+        # 'sentence_features': ...}
```
tabergma, line 699:
```suggestion
        # {'sequence_features': ..., 'sentence_features': ...}
```

tabergma (commented):
Looks great 🚀 Good work 💯 

Just a few minor comments. I guess we need to wait until Daksh merged his PR and then we can wrap this up.

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -2,11 +2,14 @@
 
 from typing import Any, Dict, List, NoReturn, Optional, Text, Tuple, Type
 from tqdm import tqdm
+import os
 
 import rasa.shared.utils.io
-from rasa.nlu.tokenizers.convert_tokenizer import ConveRTTokenizer
+from rasa.core.utils import get_dict_hash
```
tabergma, line 8: Please, do not import functions directly but rather the module itself. See https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#70d0247fdd1f48d2aec341fd91fe1dac.

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -17,15 +20,33 @@
...
 
+# URL to the old remote location of the model which
+# users might use. The model is no longer hosted here.
+ORIGINAL_TF_HUB_MODULE_URL = (
+    "https://github.com/PolyAI-LDN/polyai-models/releases/download/v1.0/model.tar.gz"
+)
+
+# Warning: This URL is only intended for running pytests on ConveRT
+# related components. This URL should not be allowed to be used by the user.
+RESTRICTED_ACCESS_URL = "https://storage.googleapis.com/continuous-integration-model-storage/convert_tf2.tar.gz"
```
tabergma, line 48: Can't users simply download the model from here and use the model than locally? Should we somehow protect this URL a bit more?

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -1,66 +1,781 @@
-from typing import Any, Optional, Text, List, Type
+import numpy as np
+import logging
 
+from typing import Any, Optional, Text, List, Type, Dict, Tuple
+
+from rasa.core.utils import get_dict_hash
```
tabergma, line 6: Import the module instead of the function.

tabergma (approved):
Great work 🚀 Looks good from my side. Just left a few comments. Please address them before you merge. Also there is still one test failure. You might need to change the configuration (use a different model) as the current configuration is producing a timeout.

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
tabergma: cc @dakshvar22 
I guess this should be tackled in a separate PR if we want to change it.

dakshvar22: We discussed a few other approaches briefly but decided to go this route because others didn't really seem feasible. See this [slack thread](https://rasa-hq.slack.com/archives/C36SS4N8M/p1603736235407200).

koernerfelicia:
/modeltest
```yml
include:
 - dataset: ["all"]
   config: ["BERT + DIET(bow) + ResponseSelector(bow)","BERT + DIET(seq) + ResponseSelector(t2t)","Sparse + BERT + DIET(bow) + ResponseSelector(bow)","Sparse + BERT + DIET(seq) + ResponseSelector(t2t)"]

```

howl-anderson:
Hi, @koernerfelicia it seems that the issue page

> Research [issue 108](https://github.com/RasaHQ/research/issues/108)

got a 404 error. Is that a typo error or the expected result (e.g. private repo)?

koernerfelicia:
Hi @howl-anderson, you're right, it's a link to a private repo. Do you have any questions about the issue that I can answer?

howl-anderson:
Hi @koernerfelicia, I am just working on an issue related to this, so far everything is OK, if I have a question, I will let you know. Thank you!

koernerfelicia:
/modeltest
```yml
include:
 - dataset: ["Carbon Bot"]
   config: ["BERT + DIET(bow) + ResponseSelector(bow)"]

```

Issue #6977, opened 2020-10-09
Send potential `ActionExecutionRejected` after form validation


wochinge:
The `FormAction` currently returns an `ActionExecutionRejected` event in case there were no slot candidates extracted (see [here](https://github.com/RasaHQ/rasa/blob/6764365ed8f1104d938d77c489f5ea9fcfe00e6f/rasa/core/actions/forms.py#L452)).
The problem is that with Rasa Open Source 2.0 users can extract custom slots in their custom `validate` action. In case they only extract slots manually, it currently throws a `ActionExecutionRejected`. Instead it should run the custom validate function and decide then whether a `ActionExecutionRejected` event should be returned.

Notes by @Ghostvv 
> we need to think how custom vs default validation works and how to notify form through custom validation action

ArjaanBuijk:
I tried to use ActionExecutionRejected for implementing some logic, but got stuck because it is not possible to set slots at the same time. It would be a good enhancement I think to allow to set slots.

PR #7035, opened 2020-10-16
check if `ActionExecutionRejected` should be thrown after custom action for slot validations was called

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6977
- check for a required `ActionExecutionRejection` error **after** custom action for slot validations / extractions was called. This doesn't reject forms before users could intervene.
- don't deactivate the form in case the user manually returned a `ActionExecutionRejected` event
- raise `InvalidDomain` instead of `ValueError` for invalid slot mappings to avoid unnecessary Sentry errors
- a custom action can reject its execution by returning a `ActionExecutionRejected` event and still apply other events (e.g. slots)
- update related documentation


changelog/6977.improvement.md
```diff
@@ -0,0 +1,11 @@
+[Forms](forms.mdx) no longer reject their execution before a potential custom 
+action for validating / extracting slots was executed. 
+Forms continue to reject in two cases automatically:
+- A slot was requested to be filled, but no slot mapping applied to the latest user
+  message and there was no custom action for potentially extracting other slots.
+- A slot was requested to be filled, but the custom action for validating / extracting
+  slots didn't return any slot event.
```
Ghostvv, line 7: what happens if you want to simply utter "please rephrase" in your custom validation without setting any slot?

rasa/core/actions/forms.py
```diff
@@ -666,6 +685,9 @@ async def is_done(
         domain: "Domain",
         events_so_far: List[Event],
     ) -> bool:
+        if any(isinstance(event, ActionExecutionRejected) for event in events_so_far):
+            return False
+
```
Ghostvv, line 690: I don't thin it will cause policies to predict different action, though, because `ActionExecuted(form_name)` will be added to `tracker.events`, no?
Also, isn't returning `False` here will deactivate the form? That's not what we want to do

Ghostvv (changes requested):

changelog/6977.improvement.md
wochinge: In that case they'd do that in response to invalid slots, correct? As long as they return `SlotSet(<slot name>, None)` events there won't be an automatic rejection.

rasa/core/actions/forms.py
wochinge: return `False` here means `is_done = False` which will **not** deactivate the form.

> I don't thin it will cause policies to predict different action, though, because ActionExecuted(form_name) will be added to tracker.events, no?

Good point. This will work as long as it's added as last event in the tracker. I'll make the test more robust.

wochinge: Arg. It's actually a bit more complicated than that. @ArjaanBuijk What are you trying to do by rejection but still setting slots?

wochinge: fixed. rejections in conjunction with other events are now possible

changelog/6977.improvement.md
Ghostvv: aaaah, so you'd still need to set requested slot to `None`, I think it is a bit of a hassle from user perspective but could work if we document it well. @ArjaanBuijk what do you think?

wochinge: > so you'd still need to set requested slot to None

No. When the action server retrieves a request to run the validation action **and** this doesn't return any `SlotSet` events at all, then we automatically raise ` ActionExecutionRejected`. If the validation action marks all slots as invalid (by returning `SlotSet(<slot name>, None)` events) then we don't raise a rejection. They can request next slots / do utterances as they desire.

I think this in line with the Rasa Open Source 1.x behavior to be honest. If you use the `FormValidation` action then you'd always return something like `{"slot_name": <value>}`. So not returning any `SlotSet` event in the validation action should really be the exception.

Ghostvv: sorry, that's what I meant. I didn't finish the sentence that was in my head:
"you'd still need to set requested slot to None if you don't want to reject the form"

Ghostvv: @wochinge could you please write documentation for this as part of this PR

Ghostvv (commented):
please update the docs as part of this PR, otherwise looks good to me 🚀

wochinge:
@ArjaanBuijk Could you please check the documentation?

changelog/6977.improvement.md
wochinge: done 👍

wochinge:
@tczekajlo The security check for the secrets is currently failing. it's not required but I think we need to exclude `mdx` files from the check

wochinge:
@melindaloubser1 Requested review from you since @ArjaanBuijk doesn't seems to be busy

melindaloubser1 (approved):
I made some small phrasing changes directly since it won't allow me to request changes on deleted lines

wochinge:
Thank you so much for this blazing fast review and changes @melindaloubser1 🙌

wochinge:
@Ghostvv I'm gonna dismiss your `changes requested` because you said it's good to go apart from the docs and they were approved now.

PR #7040, opened 2020-10-16
make converter convert numbers as entity type

### LINKED ISSUE #7031: Converter won't convert entities with numbers as digits
**Rasa version**: 2.0.0

**Issue**:

The convertor fails on cases where entity value is a number not enclosed in `""`.

If you try to convert this story:
```
## story_id
* technical_question: how to download [7]{"entity": "number", "value": 7}
    - action_docs_search
```
..the exception is thrown. 

That shouldn't be a case and converter should be able to parse that kind of entities successfully.

kalkbrennerei:
Closes #7031

**Proposed changes**:
- added `number` as possible `value` type for in `_common_entity_properties`


degiz (changes requested):
OMG, great detective work finding the `data_schema.py` 🚀 🕵️ 

- One ask about the test - could we rather test it with a YAML story? MD is deprecated now 😬 
- Another ask - could you please add a changelog entry?

kalkbrennerei:
>     * One ask about the test - could we rather test it with a YAML story? MD is deprecated now 😬
> 

@degiz this test is about testing the converter tool from markdown to YAML - so it should be in MD right?

degiz:
@kalkbrennerei 

`tests/core/training/converters/test_story_markdown_to_yaml_converter.py` is indeed converting MD to the YAML.
But we could use other unit tests to verify the new functionality 😃

degiz (changes requested):
`tests/core/training/converters/test_story_markdown_to_yaml_converter.py` is indeed converting MD to the YAML.

But we could use other unit tests to verify the new functionality 😃

degiz (approved):
Thank for fixing that! 👍 
I think we can safely remove `test_test_stories_contains_number` since the  `test_entity_dict_is_valid` is sufficient!

PR #7079, opened 2020-10-21
Add option "featurizers" to TEDPolicy

tabergma:
**Proposed changes**:
`TEDPolicy` now has the option `featurizers`. It can be set to a list of featurizer names (name of `alias`). Only the given featurizers will be considers when the dialogues are featurized.

related to https://github.com/RasaHQ/rasa/issues/6670

### DIFF
--- a/rasa/utils/tensorflow/model_data_utils.py
+++ b/rasa/utils/tensorflow/model_data_utils.py
@@ -23,6 +23,9 @@ if typing.TYPE_CHECKING:
     from rasa.nlu.classifiers.diet_classifier import EntityTagSpec
 
 
+ENTITY_TAG_ORIGIN = "origin_entity_tag_ids"
+
+
 def convert_training_examples(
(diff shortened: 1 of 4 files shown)


rasa/utils/tensorflow/model_data_utils.py
```diff
@@ -177,6 +211,7 @@ def convert_to_data_format(
     ],
     zero_features: Optional[Dict[Text, List["Features"]]] = None,
     consider_dialogue_dimension: bool = True,
+    featurizers: Optional[List[Text]] = None,
```
Ghostvv, line 214: diet also uses `convert_to_data_format` but don't use this option. Does it make sense to unify the way ted and diet use featurizers?

tabergma: would do that as part of the bigger refactoring

Ghostvv (approved):

Issue #7045, opened 2020-10-19
Restrict ConveRTTokenizer to work only if model_url was specified.


dakshvar22:
Since ConveRT was taken offline by the original authors, the tokenizers and featurizers cannot download the model from the original URL of the model.
However, [some users](https://github.com/RasaHQ/rasa/issues/6806#issuecomment-700017350) may already have a copy of the model and they can host the model on their own if they wish to and use it inside the components. To do so, they can specify the URL of the model or path to a local directory as the value to the parameter `model_url` in the configuration of `ConveRTTokenizer`. `ConveRTTokenizer` should throw an error if the `model_url` is not specified in the configuration or it is the same as the URL of the original model which the authors hosted.
Currently, without this check in place the components error out with a model not found error which is not helpful for the user.

👉  [Slack Thread](https://rasa-hq.slack.com/archives/C7B08Q5FX/p1601385771081700)

dakshvar22:
Bumping up the priority as it affects a lot of users.

AdityaSingh06:
Hi Team,

Even I was facing the ConveRTTokenizer unavailibility issue before so i upgraded rasa to '1.10.14' (using the command "pip3 install rasa[convert] ==1.10.14") and made the following changes to the config.yml file --

pipeline:

name: "ConveRTTokenizer"
"model_url": https://github.com/connorbrinton/polyai-models/releases/download/v1.0/model.tar.gz
name: "ConveRTFeaturizer"
"model_url": https://github.com/connorbrinton/polyai-models/releases/download/v1.0/model.tar.gz
But ever since I made the above two mentioned changes, my model has become extremely erratic (unable to identify entities properly, misclassifying intensions etc.).
My data has not changed and only above mentioned changes are made from my side. Is there anything i am missing or need to do here?
Any suggestion or pointer would be of great help. Thanks community.

Issue #6806, opened 2020-09-28
PolyAI Models - model.tar.gz - No Longer Available?


RedTint:
**Rasa version**: 2.0.0rc2
**Python version**: 3.6.12

**Operating system** (windows, osx, ...): OSX

**Issue**: Failed to run `rasa init`. Upon checking the model source, it is no longer available. It was taken down `2 hours ago` upon posting this.

Edited to include the message from https://github.com/PolyAI-LDN/polyai-models:
> After much consideration, the PolyAI team has decided to take down the ConveRT models from the public domain.
> Over the course of last year, we have been very excited to see ConveRT gaining a huge amount of traction in various communities - that was something we didn't expect when we first released it. However, with the amount of business growing and the shift of our team's priorities, we no longer have the resources to responsibly maintain or provide support for these models.
> PolyAI is working to create end-to-end voice assistants. If you're interested in helping us, check out our careers page at polyai.com/careers. On the other hand, if you are interested in knowing how our solutions can help you transform your contact center, please get in touch at contact@polyai.com.

**Error (including full traceback)**:
```
Warning: Output is not to a terminal (fd=1).
Warning: Input is not to a terminal (fd=0).
2020-09-28 04:00:48 INFO     absl  - Using /tmp/tfhub_modules to cache modules.
2020-09-28 04:00:48 INFO     absl  - Downloading TF-Hub Module 'https://github.com/PolyAI-LDN/polyai-models/releases/download/v1.0/model.tar.gz'.
Welcome to Rasa! 🤖

To get started quickly, an initial project will be created.
If you need some help, check out the documentation at https://rasa.com/docs/rasa.

Created project directory at '/usr/src/app'.
Finished creating project structure.
Training an initial model...
The configuration for policies and pipeline was chosen automatically. It was written into the config file at './config.yml'.
Training NLU model...
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/rasa/utils/train_utils.py", line 142, in load_tf_hub_model
    return tfhub.load(model_url)
  File "/usr/local/lib/python3.6/site-packages/tensorflow_hub/module_v2.py", line 97, in load
    module_path = resolve(handle)
  File "/usr/local/lib/python3.6/site-packages/tensorflow_hub/module_v2.py", line 53, in resolve
    return registry.resolver(handle)
  File "/usr/local/lib/python3.6/site-packages/tensorflow_hub/registry.py", line 42, in __call__
[...]

PR #7089, opened 2020-10-22
Restrict acceptable values for model_url parameter of ConveRTTokenizer

dakshvar22:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/7045
- Closes https://github.com/RasaHQ/rasa/issues/6806


rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -6,15 +6,23 @@
...
+from rasa.utils import common, io
 import rasa.utils.train_utils as train_utils
+from rasa.exceptions import RasaException
 import tensorflow as tf
+import os
 
 
-TF_HUB_MODULE_URL = (
+# URL to the old remote location of the model which
+# users might use. The model is not longer hosted here.
```
dakshvar22, line 17:
```suggestion
# users might use. The model is no longer hosted here.
```

wochinge:
@dakshvar22 Seems the remaining docstring issues are valid.

dakshvar22:
Just pushed fixes for them.

changelog/7089.bugfix.md
```diff
@@ -0,0 +1,6 @@
+Fix `ConveRTTokenizer` failing because of wrong model URL by making `model_url` parameter of `ConveRTTokenizer` mandatory.
```
wochinge, line 1: 🤔 
```suggestion
Fix  `ConveRTTokenizer` failing because of wrong model URL by making the `model_url` parameter of `ConveRTTokenizer` mandatory.
```

wochinge, line 1: can we please also add a link to the docs of the `ConveRTTokenizer`?

changelog/7089.bugfix.md
```diff
@@ -0,0 +1,6 @@
...
+
+Since the ConveRT model was taken [offline](https://github.com/RasaHQ/rasa/issues/6806), we can no longer use
```
wochinge, line 3: optional: give example of how to configure the `model_url`

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
-    def __init__(self, component_config: Dict[Text, Any] = None) -> None:
-        """Construct a new tokenizer using the WhitespaceTokenizer framework."""
+    def _validate_model_files_exist(self, model_directory: Text) -> None:
+        """Check if essential model files exist inside the model_directory.
 
+        Args:
+            model_directory: Directory to investigate
+        """
+        files_to_check = [
+            os.path.join(model_directory, "saved_model.pb"),
```
wochinge, line 53: optional: let's use the `pathlib` to concatenate /deal with the paths.

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+            os.path.join(model_directory, "variables/variables.data-00001-of-00002"),
+            os.path.join(model_directory, "variables/variables.data-00000-of-00002"),
+        ]
+
+        for file_path in files_to_check:
+            if not os.path.exists(file_path):
+                raise RasaException(
+                    f"""File {file_path} does not exist.
+                    Re-check the files inside the directory {model_directory}.
+                    It should contain the following model files - {files_to_check}"""
```
wochinge, line 64: should we concatenate the list with `,` instead of printing the pure python list? There is also a utility function for it as far as I know

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+                )
+
+    def _get_validated_model_url(self, ignore_exceptions: bool) -> Text:
+        """Validates the specified model_url parameter.
```
wochinge, line 68:
```suggestion
        """Validates the specified `model_url` parameter.
```

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+
+        The model_url parameter cannot be left empty. It can either
```
wochinge, line 70:
```suggestion
        The `model_url` parameter cannot be left empty. It can either
```

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+        be set to a remote URL where the model is hosted or it can be
```
wochinge, line 71:
```suggestion
        be set to a remote URL where the model is hosted or it can be a
```

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+        path to a local directory.
+
+        Args:
+            ignore_exceptions: Whether to ignore all exception checks. This
+            is strictly only for pytests.
```
wochinge, line 76: I'd rather monkeypatch this method then instead of changing production code for tests

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+
+    def __init__(
+        self, component_config: Dict[Text, Any] = None, ignore_exceptions: bool = False
+    ) -> None:
+        """Construct a new tokenizer using the WhitespaceTokenizer framework.
+
+        Args:
+            component_config: User configuration for the component
+            ignore_exceptions: Whether to ignore all exception checks. This
+            is strictly only for pytests.
```
wochinge, line 151: This seems misleading. It only disables exception checks for the model url, but not for anything else.

rasa/utils/io.py
```diff
@@ -212,3 +213,26 @@ def json_pickle(file_name: Union[Text, Path], obj: Any) -> None:
...
+    """Check whether the url specified is a well formed one.
+
+    Regex taken from https://stackoverflow.com/a/7160778/3001665
+
+    Args:
+        url: Remote URL to validate
+
+    Returns: Boolean for validity
+    """
+    regex = re.compile(
```
wochinge, line 228: Can we use something like https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse and check if the protocol is given and `http(s)`? seems more robust in my opinion.

tests/nlu/featurizers/test_convert_featurizer.py
```diff
@@ -98,9 +101,9 @@ def test_convert_featurizer_train(component_builder):
         ("ńöñàśçií", "ńöñàśçií"),
     ],
 )
-@pytest.mark.skip
-def test_convert_featurizer_tokens_to_text(component_builder, sentence, expected_text):
-    tokenizer = component_builder.create_component_from_class(ConveRTTokenizer)
+def test_convert_featurizer_tokens_to_text(sentence, expected_text):
```
wochinge, line 104:
```suggestion
def test_convert_featurizer_tokens_to_text(sentence: Text, expected_text: Text):
```

wochinge: optional: types for the other test functions would be amazing, too

tests/nlu/tokenizers/test_convert_tokenizer.py
```diff
@@ -25,55 +31,133 @@
...
 
-    tk.train(TrainingData([message]))
+    tokenizer.train(TrainingData([message]))
 
     assert [
         t.get(NUMBER_OF_SUB_TOKENS) for t in message.get(TOKENS_NAMES[TEXT])
     ] == expected_number_of_sub_tokens
+
+
+@pytest.mark.skip_on_windows
```
wochinge, line 92: this test actually works on all platforms, right? Same for the next ones.

tests/nlu/tokenizers/test_convert_tokenizer.py
```diff
@@ -25,55 +31,133 @@
...
+    "model_url, exception_phrase",
+    [
+        (ORIGINAL_TF_HUB_MODULE_URL, "which does not contain the model any longer"),
+        (
+            RESTRICTED_ACCESS_URL,
+            "which is strictly reserved for pytests of Rasa Open Source only",
+        ),
+    ],
+)
+def test_raise_invalid_urls(model_url: Text, exception_phrase: Text):
```
wochinge, line 116: can we merge the `no_url` with this test function?

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,16 +39,120 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+        else:
+            raise RasaException(
+                f"""{model_url} is neither a valid remote URL nor a local directory.
+                You can either use a community hosted URL or if you have a
+                local copy of the model, pass the path to
+                the directory containing the model files."""
+            )
+        return model_url
+
+    def __init__(
```
wochinge, line 143: let's keep the constructor at the top and have these helpers below it. I think it's more intuitive to see the constructor right away when jumping to a class definition instead of semi-related helper methods. What do you think?

tests/nlu/tokenizers/test_convert_tokenizer.py
```diff
@@ -25,55 +31,133 @@
...
+        component_config = {"name": "ConveRTTokenizer", "model_url": temp_dir}
+
+        with pytest.raises(RasaException) as excinfo:
+            _ = ConveRTTokenizer(component_config)
+
+        assert "Re-check the files inside the directory" in str(excinfo.value)
+
+
+@pytest.mark.skip_on_windows
+def test_raise_wrong_model_file():
```
wochinge, line 139: that's a pytest fixture which gives you a temporary directory. No need for using `tempfile` (same for the others)
```suggestion
def test_raise_wrong_model_file(tmp_path: Path):
```

tests/utils/test_io.py
```diff
@@ -91,3 +92,18 @@ def test_write_utf_8_yaml_file(tmp_path: Path):
...
     assert rasa.shared.utils.io.read_file(file_path) == "data: amazing 🌈\n"
+
+
+@pytest.mark.parametrize(
+    "url, result",
+    [
+        ("a/b/c", False),
+        ("a", False),
+        ("https://google.com", True),
+        ("https://www.google.com", True),
```
wochinge, line 103: I think a test case for ips would be great, too.

wochinge (commented):
Great testing!

rasa/utils/io.py
dakshvar22: I tried using it and looking like the check would be something like if `scheme` and `netloc` are defined then it is a valid URL but it would fail in a simple case like `https://a/b/c` where `a` is extracted as `netloc` and `b/c` as path. Ideally it should have been `a/b/c` as path and then `netloc` just empty. Not sure if that is more robust that a regex.

tests/nlu/tokenizers/test_convert_tokenizer.py
dakshvar22: They do, but it is an unnecessary test for them right because we do not expect folks to run `ConveRTTokenizer` on for e.g. windows.

rasa/nlu/tokenizers/convert_tokenizer.py
dakshvar22: Good point. Changed it!

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -30,21 +39,117 @@ class ConveRTTokenizer(WhitespaceTokenizer):
...
+        for file_path in files_to_check:
+            if not os.path.exists(file_path):
+                raise RasaException(
+                    f"""File {file_path} does not exist.
+                    Re-check the files inside the directory {model_directory}.
+                    It should contain the following model 
+                    files - [{", ".join(files_to_check)}]"""
+                )
+
+    def _get_validated_model_url(self) -> Text:
```
wochinge, line 83: We could remove some indentation levels by doing something like this, what do you think?

```python
        if not model_url:
            raise RasaException(
                """Parameter "model_url" was not specified in the configuration of "ConveRTTokenizer".
                You can either use a community hosted URL of the model
                or if you have a local copy of the model, pass the
                path to the directory containing the model files."""
            )

        if model_url == ORIGINAL_TF_HUB_MODULE_URL:
            # Can't use the originally hosted URL
            raise RasaException(
                f"""Parameter "model_url" of "ConveRTTokenizer" was
                set to "{model_url}" which does not contain the model any longer.
                You can either use a community hosted URL or if you have a
                local copy of the model, pass the path to the directory
                containing the model files."""
            )

        if model_url == RESTRICTED_ACCESS_URL:
            # Can't use the URL that is reserved for tests only
            raise RasaException(
                f"""Parameter "model_url" of "ConveRTTokenizer" was
                set to "{model_url}" which is strictly reserved for pytests of Rasa Open Source only.
                Due to licensing issues you are not allowed to use the model from this URL.
                You can either use a community hosted URL or if you have a
                local copy of the model, pass the path to the directory
                containing the model files."""
                )

        if io.is_valid_remote_url(model_url):
            return model_url

        if os.path.isfile(model_url):
            # Definitely invalid since the specified path should be a directory
            raise RasaException(
                """Parameter "model_url" of "ConveRTTokenizer" was
                set to the path of a file which is invalid. You
                can either use a community hosted URL or if you have a
                local copy of the model, pass the path to the directory
                containing the model files."""
            )

        if os.path.isdir(model_url):
            # Looks like a local directory. Inspect the directory
            # to see if model files exist.
            self._validate_model_files_exist(model_url)
            # Convert the path to an absolute one since
            # TFHUB doesn't like relative paths
            return os.path.abspath(model_url)

        raise RasaException(
            f"""{model_url} is neither a valid remote URL nor a local directory.
            You can either use a community hosted URL or if you have a
            local copy of the model, pass the path to
            the directory containing the model files."""
        )
```

wochinge (approved):
Please check the remaining comments. Go to go otherwise 👍

rasa/utils/io.py
dakshvar22: Makes sense. Didn't know this function existed. I have replaced the original regex here with the new regex. Handles other cases like IP, ports, etc.

rasa/utils/io.py
```diff
@@ -10,6 +10,7 @@
 from io import BytesIO as IOReader
 from pathlib import Path
 from typing import Text, Any, Union, List, Type, Callable, TYPE_CHECKING
+import re
```
dakshvar22, line 13:
```suggestion
```

Issue #7011, opened 2020-10-14
'Edit this page' links to wrong url


erohmensing:
Due to the way the docs are deployed and versioned, the "edit this page" from `[https://rasa.com/docs/rasa/installation](https://rasa.com/docs/rasa/installation)` goes to `[https://github.com/rasahq/rasa/edit/master/docs/versioned_docs/version-2.0.x/installation.mdx](https://github.com/rasahq/rasa/edit/master/docs/versioned_docs/version-2.0.x/installation.mdx)`. We'd like it to go to `[https://github.com/RasaHQ/rasa/blob/master/docs/docs/installation.mdx](https://github.com/RasaHQ/rasa/blob/master/docs/docs/installation.mdx)`. This goes for all pages, in rasa-sdk as well!

cc @m-vdb

lunelson:
@erohmensing thanks for the issue. IMO it's a bit misleading for the link to say "edit this page" in that case, since a user who is reading a versioned page would expect to be led to a place where they can contribute an edit to _the versioned page_; moreover, not all versioned pages will exist in the master (unreleased) branch, so there is still potential for the path to 404.

Perhaps versioned pages should go to the documentation branch, and unreleased (current) pages should link to the master branch?

In any case think this issue should be transfered to the theme repo, so we can rewrite this logic for all sites, wdyt?

erohmensing:
@lunelson perhaps we could only make it so that the main docs, not the versioned branches, have the "Edit this page"?

lunelson:
@erohmensing We could also disable the feature altogether: do you see this as something people will do? If we only present this link on the `master/unreleased` version, I think most users will never see it anyway, since this is not the main view

erohmensing:
We've definitely seen people do it before! At that time it was at the top, which makes it more visible, but I could still see people doing it especially if we ask them to.

Should have clarified, I guess it makes most sense to have it only on the main (2.0.x) branch and have it edit the 2.0.x branch, as that is the branch any updates to documentation should go into anyway, in comparison to the master branch

lunelson:
I don't think this is something I can generalize in the theme: the only thing the theme can consistently assume is the path to the versioned document within the repo, because the naming pattern of versioned subfolders is determined by Docusaurus. We could customize the edit URL in terms of which branch actually contains all the versions, if that doesn't happen to be `master`; but beyond this, every repo will differ and there's no way to extrapolate from version-strings to branch-names, AFAICT 😬

[...]

PR #7097, opened 2020-10-23
Bump theme to 0.8.2, configure `rootEditUrl`

lunelson:
This is a patch theme update to support a `customFields.rootEditUrl` property, as a quick fix for [404 errors on "Edit This Page" links](https://github.com/RasaHQ/rasa/issues/7011). 

[Theme changes can be seen here](https://github.com/RasaHQ/tabula/commit/1feccb59e035563ffabf726163fcf975e1519857)

My suggestion for this property is to create a section in the README which explains how to submit PRs for docs fixes, including which branches/versions will be allowed, and put that anchor link in.

Closes https://github.com/RasaHQ/tabula/issues/49

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -25,7 +25,10 @@ const versionLabels = {
 
 module.exports = {
   customFields: {
-    rootEditUrl: 'https://github.com/rasahq/rasa/', // NOTE: this could be directed to a specfic README or a heading within a README
+    // FIXME: this is a simplistic solution to https://github.com/RasaHQ/rasa/issues/7011
+    // either (A): create a more sophisticated solution to link the precise branch and doc to be edited, according to branch settings
+    // or (B): create a README document (or a section in the main README) which explains how to contribute docs fixes, and link all edit links to this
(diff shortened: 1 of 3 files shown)


docs/docusaurus.config.js
```diff
@@ -25,6 +25,7 @@ const versionLabels = {
 
 module.exports = {
   customFields: {
+    rootEditUrl: 'https://github.com/rasahq/rasa/', // NOTE: this could be directed to a specfic README or a heading within a README
```
m-vdb, line 28: should we add . FIXME that references this issue: https://github.com/RasaHQ/rasa/issues/7011 ?

also I think the NOTE is not that useful 🤷

m-vdb (approved):
Looks good, thanks for doing this 💯 Do we also want to bump the theme on `rasa-sdk`?

docs/docusaurus.config.js
lunelson: OK, have updated this to a FIXME with issue ref and more explanation about future solution options

m-vdb (approved):
💯

PR #7098, opened 2020-10-23
make rasa test support retrieval intents for test stories

### LINKED ISSUE #7002: Rasa 2.0: responses in tests fail
**Rasa version**:
2.0

2.0

**Python version**:
3.8

Fedora 32

**Issue**:
rasa tests does not seem to work for responses.
Using a test story like this:
```
- story: chitchat_el/activities0el01
  steps:
  - user: |
      <some user text>
    intent: chitchat_el/activities0el01
  - action: utter_chitchat_el/activities0el01
```
raises an error (about chitchat_el/activities0el01 not existing in the domain).
If I use a test story like:
```
- story: chitchat_el/activities0el01
  steps:
  - user: |
      <some user text>
    intent: chitchat_el
  - action: utter_chitchat_el/activities0el01
```
all metrics are 0.

```

```

```
rasa test
```

```yml

```

```yml

```

kalkbrennerei:
closes #7002 

**Proposed changes**:
- make `rasa test` support `retrieval intents` in `test stories`, so either `base intents` or `full retrieval intents` can be specified
- adjusted warning about `retrieval intents` in `domain` in `yaml_story_reader`
- make `core/test.py::_collect_user_uttered_predictions` check for `intent_response_key` in `prediction` to avoid wrong mismatches
- make comments in `failed_test_stories.yml` consistent, always make the report show the full retrieval intent


data/test_evaluations/retrieval_intent_story.yml
```diff
@@ -0,0 +1,12 @@
...
+  steps:
+  - user: |
+      What is your name?
+    intent: chitchat/ask_name
+  - action: utter_chitchat
+  - user: |
+      What's the weather like today?
+    intent: chitchat
+  - action: utter_chitchat
+
```
tmbo, line 12:
```suggestion

```

data/test_evaluations/retrieval_intent_wrong_prediction.yml
```diff
@@ -0,0 +1,8 @@
+stories:
+- story: chitchat name
+  steps:
+  - user: |
+      What is your name?
+    intent: mood_unhappy
+  - action: utter_chitchat
+
```
tmbo, line 8:
```suggestion

```

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -450,9 +451,15 @@ def prediction_states(
         ]
         # TODO there is no prediction support for e2e input right now, therefore
         #  temporary remove TEXT features from USER state during prediction
+        #  also temporary remove full retrieval intent
         for states in trackers_as_states:
             for state in states:
                 if state.get(USER, {}).get(TEXT):
                     del state[USER][TEXT]
 
+                # remove full intent, only keep base intent
```
tmbo, line 460: why do we need to remove the full intent?

rasa/core/test.py
```diff
@@ -351,6 +359,37 @@ def _clean_entity_results(
...
+
+    base_intent = parsed.get(INTENT, {}).get(INTENT_NAME_KEY, {})
+
+    try:
+        # return only base intent if intent is not a retrieval intent
+        if (
+            base_intent
+            not in parsed[RESPONSE_SELECTOR][RESPONSE_SELECTOR_RETRIEVAL_INTENTS]
+        ):
+            return base_intent
```
tmbo, line 372: does this really need to be within the `try`?

tests/core/conftest.py
```diff
@@ -219,3 +225,18 @@ async def form_bot_agent(trained_async) -> Agent:
     )
 
     return Agent.load_local_model(zipped_model)
+
+
+@pytest.fixture
```
tmbo, line 230: this shouldn't be a function scoped fixture (nor should the other examples in this file...) as it will retrain the model for every test which is just a waste of time. should rather be session scoped

tests/core/conftest.py
```diff
@@ -219,3 +225,18 @@ async def form_bot_agent(trained_async) -> Agent:
...
+async def response_selector_agent(trained_async) -> Agent:
```
tmbo, line 231: missing type annotation
```suggestion
async def response_selector_agent(trained_async: Callable) -> Agent:
```

tests/core/test_evaluation.py
```diff
@@ -242,3 +244,38 @@ def test_event_has_proper_implementation(
...
+    generator = await _create_data_generator(
+        RETRIEVAL_INTENT_STORY_FILE, response_selector_agent, use_e2e=True
+    )
+    test_stories = generator.generate_story_trackers()
+
+    story_evaluation, num_stories = await _collect_story_predictions(
+        test_stories, response_selector_agent, use_e2e=True
+    )
+    # check that test story can either specify base intent or full retrieval intent
+    assert not story_evaluation.evaluation_store.has_prediction_target_mismatch()
```
tmbo, line 259: I think it would be better if these would be two test cases. one testing with base intent and the other with the full retrieval intent

tmbo (commented):
great work :+1: added some style suggestions. the only thing that's unclear to me is the change to the featurization

data/test_evaluations/retrieval_intent_story.yml
degiz: I think this file should be in `data/test_yaml_stories` folder instead.

data/test_evaluations/retrieval_intent_wrong_prediction.yml
degiz: And this file too!

rasa/core/featurizers/tracker_featurizers.py
degiz: @kalkbrennerei were we supposed to detect if the `intent` is a `retrieval intent` here, and take only the `base intent` from it? If yes, then:
- we should adjust the comment, since `temporary remove full retrieval intent` and `remove full intent` sound scary 🙀 
- `INTENT_MESSAGE_PREFIX` -> `RESPONSE_IDENTIFIER_DELIMITER` **or** check my suggestion below, we probably don't even need to check anything manually 👍

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -450,9 +451,15 @@ def prediction_states(
...
+                intent = state.get(USER, {}).get(INTENT)
+                if intent and INTENT_MESSAGE_PREFIX in intent:
+                    state[USER][INTENT] = intent.split(INTENT_MESSAGE_PREFIX)[0]
```
degiz, line 463:
```suggestion
                base_intent, _ = Message.separate_intent_response_key(intent)
                state[USER][INTENT] = base_intent
```

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -302,8 +302,14 @@ def _parse_user_utterance(self, step: Dict[Text, Any]) -> None:
             self.current_step_builder.add_user_messages([utterance])
 
     def _validate_that_utterance_is_in_domain(self, utterance: UserUttered) -> None:
+
         intent_name = utterance.intent.get(INTENT_NAME_KEY)
 
+        # check if this is a retrieval intent
+        # in this case check only for the base intent in domain
+        if INTENT_MESSAGE_PREFIX in intent_name:
```
degiz, line 310: The same comment with `INTENT_MESSAGE_PREFIX` -> `RESPONSE_IDENTIFIER_DELIMITER` and using `Message.separate_intent_response_key()` instead.

changelog/7002.bugfix.md
```diff
@@ -0,0 +1,2 @@
+Test stories can now contain base intents or full retrieval intents. The report `failed_test_stories.yml` also specifies the full retrieval intent.
```
degiz, line 1:
```suggestion
Test stories can now contain both: normal intents and retrieval intents. The ``failed_test_stories.yml``, generated by ``rasa test``, also specifies the full retrieval intent now.
```

changelog/7002.bugfix.md
```diff
@@ -0,0 +1,2 @@
...
+Previously rasa test would mismatch test stories that specified full retrieval intents.
```
degiz, line 2:
```suggestion
Previously ``rasa test`` would fail on test stories that specified retrieval intents.
```

rasa/core/test.py
```diff
@@ -351,6 +359,37 @@ def _clean_entity_results(
...
+
+        # extract full retrieval intent
+        # if the response selector parameter was not specified in config,
+        # the response selector contains a "default" key
+        if RESPONSE_SELECTOR_DEFAULT_INTENT in parsed.get(RESPONSE_SELECTOR, {}):
```
degiz, line 377: Are these two `if - else` cases are mutual exclusive? Should we check for the `else` condition first?

rasa/core/test.py
degiz: plus 1, doesn't seem that we should throw an exception here (because how should we recover from it?) Since we deal with `dict` here, we could use `.get()` method and specify the default value in case the `key` was not found.

rasa/core/test.py
```diff
@@ -359,9 +398,18 @@ def _collect_user_uttered_predictions(
 ) -> EvaluationStore:
     user_uttered_eval_store = EvaluationStore()
 
+    # intent from the test story, may either be base intent or full retrieval intent
```
degiz, line 401: Thanks a lot for adding comments here! 🚀

rasa/core/test.py
```diff
@@ -351,6 +359,37 @@ def _clean_entity_results(
     return cleaned_entities
 
 
+def _intent_response_key_from_parsed_data(parsed: Dict[Text, Any]) -> Text:
```
degiz, line 362: Could we add a docstring here? This method is doing some black magic, would be great to have a guide for the future 👍

tests/core/conftest.py
```diff
@@ -38,6 +38,12 @@
 
 END_TO_END_STORY_FILE = "data/test_evaluations/end_to_end_story.md"
 
+RETRIEVAL_INTENT_STORY_FILE = "data/test_evaluations/retrieval_intent_story.yml"
```
degiz, line 41: Seems like we're using the `RETRIEVAL_INTENT_STORY_FILE` and `RETRIEVAL_INTENT_WRONG_PREDICTION` only once, probably that doesn't make sense to extract them as constants, and just reference them in the actual tests that are using them.

degiz (changes requested):
Thanks a lot for digging this! 🚀

rasa/core/featurizers/tracker_featurizers.py
degiz: @dakshvar22 What do you think about this change to the `rasa/core/featurizers/tracker_featurizers.py`?
Do we also need to do that for both `FullDialogueTrackerFeaturizer` and `MaxHistoryTrackerFeaturizer`?

dakshvar22: Just so that I understand this correctly, a retrieval intent with its response key can only be present in the tracker during these E2E test stories because they are written in a way to have them there? I don't think they can be present during inferencing/training because the response key is never added to the intent present in the tracker.
From that perspective, yes making this change makes sense in both `FullDialogueTrackerFeaturizer` and `MaxHistoryTrackerFeaturizer`. @degiz Also, I agree your suggestion below is a better way to do it since it reuses the method for separating the intent from response key.
As an alternate approach, would it be possible to create the intents inside the trackers without the response key for prediction in core? They are still needed to validate if NLU predicted both intent and response key correctly. Then we don't need to have this logic here.

kalkbrennerei: maybe I can add a bit more context on that one:

- for the **why**: I added this because the prediction was confused by seeing the `full retrieval intent` and this would lead to wrong `action predictions`. This happened to me when using `rasa-demo` with a test story that contains `retrieval intents`. Shall I provide more info on that to reproduce?
- for the **where**: I assumed that this was related to the missing prediction support for `e2e input`, that's why i moved it there. Please feel free to suggest to move it to another position in the code!!
- also note that this is about the `full retrieval intent` of the `gold intent` and also only removed temporarily for the `prediction`

I'm happy to hear about your opinions on this @tmbo @degiz

degiz:
@tmbo Would you mind taking a look again? I've removed the `featurizer` changes in favor of storing the full retrieval intent along with other parsed data.

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -361,13 +374,30 @@ def _user_intent_from_step(self, step: Dict[Text, Any]) -> Text:
...
-        return user_intent
+
+        # StoryStep should never contain a full retrieval intent, only the base intent.
+        # However, users can specify full retrieval intents in their test stories file
+        # for the NLU testing purposes.
+        base_intent, response_key = Message.separate_intent_response_key(user_intent)
+        if response_key and not self.is_test_stories_file(self.source_name):
+            rasa.shared.utils.io.raise_warning(
+                f"Issue found in '{self.source_name}':\n"
+                f"User intent '{user_intent}' is a retrieval intent. "
```
tmbo, line 385: this error should be slightly more specific, e.g. we have the file in there but the  user would need to search the whole file to find the issue

rasa/shared/nlu/constants.py
```diff
@@ -1,10 +1,12 @@
 TEXT = "text"
 INTENT = "intent"
 RESPONSE = "response"
+RESPONSE_SELECTOR = "response_selector"
 INTENT_RESPONSE_KEY = "intent_response_key"
 ACTION_TEXT = "action_text"
 ACTION_NAME = "action_name"
 INTENT_NAME_KEY = "name"
+FULL_RETRIEVAL_INTENT_NAME_KEY = "full_retrieval_intent_name"
```
tmbo, line 9: why does this end with `_key`?

tests/shared/core/test_trackers.py
```diff
@@ -128,16 +133,16 @@ def test_tracker_store_storage_and_retrieval(store: TrackerStore):
     assert list(tracker.events) == [ActionExecuted(ACTION_LISTEN_NAME)]
 
     # lets log a test message
-    intent = {"name": "greet", "confidence": 1.0}
+    intent = {INTENT_NAME_KEY: "greet", "confidence": 1.0}
```
tmbo, line 136: I don't think we should use these constants in the tests, but I guess this is a broader discussion to have. 

Tests are to some extent used to make sure the API stays stable if we reuse constants we use in the code - we won't get failures if an API suddenly changes.

rasa/core/test.py
```diff
@@ -352,6 +362,45 @@ def _clean_entity_results(
...
+    """Return full retrieval intent, if it's present, or normal intent otherwise.
+
+    Args:
+        parsed: Predicted parsed data.
+
+    Returns:
+        The extracted intent.
+    """
+    base_intent = parsed.get(INTENT, {}).get(INTENT_NAME_KEY)
+
```
tmbo, line 375: `parsed_get(RESPONSE_SELECTOR, {})` is used a couple times, should get a variable

rasa/core/test.py
```diff
@@ -360,8 +409,19 @@ def _collect_user_uttered_predictions(
...
 
-    intent_gold = event.intent.get("name")
-    predicted_intent = predicted.get(INTENT, {}).get("name")
+    # intent from the test story, may either be base intent or full retrieval intent
+    base_intent = event.intent.get(INTENT_NAME_KEY)
+    full_retrieval_intent = event.intent.get(FULL_RETRIEVAL_INTENT_NAME_KEY)
+    intent_gold = full_retrieval_intent if full_retrieval_intent else base_intent
+
+    # predicted intent: note that this is only the base intent at this point
+    predicted_intent = predicted.get(INTENT, {}).get(INTENT_NAME_KEY)
```
tmbo, line 418: should this rather be `predicted_base_intent` then?

tmbo (approved):
looks good so far :+1: some minor comments

rasa/shared/nlu/constants.py
degiz: To empathize that's it a name of a key in a certain dictionary, and shouldn't be confused with any other constant.
Do you think it's too much details for a name of a constant?

tests/shared/core/test_trackers.py
degiz: Makes sense, switched back 👍

Issue #6852, opened 2020-09-30
Make entity extraction for entities separated by comma configurable 


akelad:
With PR https://github.com/RasaHQ/rasa/pull/6191, entities separated by comma now get extracted separately. This makes sense for something like: 

please add buffalo, ranch, mustard, barbeque sauces results in {'entity': 'sauces', 'start': 11, 'end': 43, 'value': 'buffalo, ranch, mustard, barbeque', 'extractor': 'DIETClassifier'}

But maybe shouldn't work this way for address entities like `5306 Walnut Ave., Building A, Sacramento, CA 95841`. See also issue https://github.com/RasaHQ/rasa/issues/6795.

We should make this behaviour configurable in the entity extractors, and maybe even on an entity level, given you might have both types of entities in one bot.

tttthomasssss:
@tabergma @akelad just to be clear - is the idea to make this configurable per entity type (on the basis of the entity types that are annotated in the NLU training data)? For example, supposing that `address` and `ingredients` are two entity types in the data, users simply need to specify sth like the following in the config (with default behaviour of True?):

```
language: en
pipeline:
  ...
  - name: DIETClassifier
    epochs: 100
    entity_recognition: True
    split_entities_by_comma: 
      address: False
      ingredients: True
  ...
```

tabergma:
Something like this sounds good I would say. I guess it would be even better if we support both variants:
```
language: en
pipeline:
  ...
  - name: DIETClassifier
    epochs: 100
    entity_recognition: True
    split_entities_by_comma: False/True
  ...
```
and 
```
language: en
pipeline:
  ...
  - name: DIETClassifier
    epochs: 100
    entity_recognition: True
    split_entities_by_comma: 
      address: False
      ingredients: True
  ...
```
We also should be able to handle the cases where not all entities are listed in the second configuration. And this should also be available for the `CRFEntityExtractor`.

PR #7103, opened 2020-10-26
Split entities by comma

tttthomasssss:
**Proposed changes**:

This PR addresses #6852.

Adds the option to configure whether extracted entities should be split by comma (`","`) or not. The default behaviour is `True` - i.e. split any list of extracted entities by comma. This makes sense for a list of ingredients in a recipie, for example `"avocado, tofu, cauliflower"`, however doesn't make sense for an address such as `"Schönhauser Allee 175, 10119 Berlin, Germany"`.

In the latter case, add a new option to your config, e.g. if you are using the `DIETClassifier` this becomes:

```yaml
...
- name: DIETClassifier
  split_entities_by_comma: False
...
```

in which case, none of the extracted entities will be split by comma. To switch it on/off for specific entity types you can use:

```yaml
...
- name: DIETClassifier
  split_entities_by_comma: 
    address: False
    ingredient: True
...
```

where both `address` and `ingredient` are two entity types. 

This feature is also available for `CRFEntityExtractor`.


docs/docs/components.mdx
```diff
@@ -1842,6 +1854,10 @@ The `SpacyEntityExtractor` extractor does not provide a `confidence` level and w
     # Name of dense featurizers to use.
     # If list is empty all available dense features are used.
     "featurizers": []
+    # Indicated whether a list of extracted entities should be split into individual tokens for a given entity type
```
tabergma, line 1857:
```suggestion
    # Indicated whether a list of extracted entities should be split into individual entities for a given entity type
```

docs/docs/components.mdx
```diff
@@ -2675,6 +2691,18 @@ However, additional parameters exist that can be adapted.
...
+|                                 |                  | globally, or set per entity type, such as:                    |
+|                                 |                  | ```                                                           |
+|                                 |                  | ...                                                           |
+|                                 |                  | - name: DIETClassifier                                        |
+|                                 |                  |   split_entities_by_comma:                                    |
+|                                 |                  |     address: True                                             |
+|                                 |                  |     ...                                                       |
+|                                 |                  | ...                                                           |
+|                                 |                  | ```                                                           |
++---------------------------------+------------------+---------------------------------------------------------------+
```
tabergma, line 2705: I guess this section is added to the `ResponseSelector`. Should be removed.

rasa/nlu/extractors/extractor.py
```diff
@@ -208,6 +210,18 @@ def convert_predictions_into_entities(
                 entities[-1][ENTITY_ATTRIBUTE_END] = token.end
                 if confidences is not None:
                     self._update_confidence_values(entities, confidences, idx)
+
+            elif self._check_is_single_entity(
+                text, token, last_token_end, split_entities_config, current_entity_tag
+            ):
+                # current token has the same entity tag as the token before and
+                # the two tokens are separated by at most 3 symbols, where each
+                # of the symbols has to bei either punctuation (e.g. "." or ",")
```
tabergma, line 219:
```suggestion
                # of the symbols has to be either punctuation (e.g. "." or ",")
```

rasa/nlu/extractors/extractor.py
```diff
@@ -208,6 +210,18 @@ def convert_predictions_into_entities(
                 entities[-1][ENTITY_ATTRIBUTE_END] = token.end
                 if confidences is not None:
                     self._update_confidence_values(entities, confidences, idx)
+
+            elif self._check_is_single_entity(
```
tabergma, line 214: The check before, e.g. `elif token.start - last_token_end <= 1:`. should be covered by this as well.

rasa/nlu/extractors/extractor.py
```diff
@@ -208,6 +210,18 @@ def convert_predictions_into_entities(
...
+                # current token has the same entity tag as the token before and
+                # the two tokens are separated by at most 3 symbols, where each
+                # of the symbols has to bei either punctuation (e.g. "." or ",")
+                # and a whitespace.
+                entities[-1][ENTITY_ATTRIBUTE_END] = token.end
+                if confidences is not None:
+                    self._update_confidence_values(entities, confidences, idx)
+
             else:
                 # the token has the same entity tag as the token before but the two
```
tabergma, line 226: this comment might need to be updated

rasa/nlu/extractors/extractor.py
```diff
@@ -254,6 +268,40 @@ def _update_confidence_values(
...
+    @staticmethod
+    def _check_is_single_entity(
+        text: Text,
+        token: Token,
+        last_token_end: int,
+        split_entities_config: Dict[Text, bool],
+        current_entity_tag: Text,
+    ):
+        # Tokens need to be no further than 3 positions apart
+        tokens_within_range = token.start - last_token_end <= 3
```
tabergma, line 280: why 3?

rasa/nlu/extractors/extractor.py
```diff
@@ -254,6 +268,40 @@ def _update_confidence_values(
...
+
+        # The interleaving tokens *must* be a full stop, a comma, or a whitespace
```
tabergma, line 282: what about other symbols like `;`?

rasa/nlu/extractors/extractor.py
```diff
@@ -119,6 +120,7 @@ def convert_predictions_into_entities(
         tokens: List[Token],
         tags: Dict[Text, List[Text]],
         confidences: Optional[Dict[Text, List[float]]] = None,
+        split_entities_config: Optional[Dict[Text, bool]] = None,
```
tabergma, line 123: I guess this should not be optional? Also the doc string for this parameter is missing.

rasa/nlu/extractors/extractor.py
```diff
@@ -254,6 +268,40 @@ def _update_confidence_values(
...
+            )
+        )
+
+        # The current entity type must match with the config (default value is True)
+        default_value = split_entities_config[SPLIT_ENTITIES_BY_COMMA]
+        split_current_entity_type = split_entities_config.get(
+            current_entity_tag, default_value
+        )
+
+        # matching_entity_types = True # TODO: Check the current entity tag matches the pre-configured one form current_entitiy_tag and the split_entities_config dict
```
tabergma, line 297: This seems to be relevant and should be added, shouldn't it?

tabergma (changes requested):
Great start 🚀  Left a few comments. Please also update the tests to cover different scenarios. Thanks.

rasa/nlu/extractors/extractor.py
tttthomasssss: `_check_is_single_entity(...)` contains this check: `tokens_within_range = token.start - last_token_end <= 3` - is that what you meant?

rasa/nlu/extractors/extractor.py
tttthomasssss: basically to catch things like `Schönhauser Allee 175, 10119 Berlin` (2 tokens apart) as well as abbreviated addresses like, `22 Powderhall Rd., EH7 4GB` which are 3 tokens apart.

I am happy to take other suggestions, but 3 seemed to be the tightest bound that will work in most cases.

rasa/nlu/extractors/extractor.py
tttthomasssss: Should we maybe move the allowed character set to a constant somewhere, so it can be extended more easily? I can certainly add a semi-colon.

rasa/nlu/extractors/extractor.py
tttthomasssss: oh dear, forgot to delete this...its superseded by what is above.

rasa/nlu/extractors/extractor.py
tabergma: yes, but then we should remove the if part in line 209

rasa/nlu/extractors/extractor.py
tabergma: make sense, would just add a comment, otherwise this is just a magic number

rasa/nlu/extractors/extractor.py
tttthomasssss: hmm, not sure I understand correctly - the former `if` statement catches things like compounding `New York`, and the new code addresses a slightly different case. Or do I misunderstand something here?

tabergma: That is correct, but your check should also cover that, right? And `_check_is_single_entity` should in my opinion cover all scenarios that indicates that the two tokens belong to one entity.  So I would just do something like
```
if new_tag_found:
    # new entity found, e.g. create new entity
    ...
elif self._check_is_single_entity(...):
    # same entity tags and same entity
    ...
else:
    # same entity tags but different entities
    ...
```

tttthomasssss: ah, right, got it :) - thank you!

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -326,12 +331,25 @@ def __init__(
...
+        split_entities_config = self.component_config.get(
+            SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE
+        )
+        if isinstance(split_entities_config, bool):
+            split_entities_config = {SPLIT_ENTITIES_BY_COMMA: split_entities_config}
+        else:
+            split_entities_config[SPLIT_ENTITIES_BY_COMMA] = self.defaults[
+                SPLIT_ENTITIES_BY_COMMA
+            ]
+        self.split_entities_config = split_entities_config
```
tabergma, line 343: I would move this to a new method in the `EntityExtractor`. As both, `DIETClassifier` and the `CRFEntityExtractor` are subclasses of the `EntityExtractor` they can call the new method in the init.

rasa/nlu/extractors/extractor.py
```diff
@@ -254,6 +262,50 @@ def _update_confidence_values(
                 confidences[ENTITY_ATTRIBUTE_GROUP][idx],
             )
 
+    @staticmethod
+    def _check_is_single_entity(
```
tabergma, line 266: What about a case like `We need to get more of tofu cauliflower avocado`? The user was too lazy to input the commas. No matter what we would always extract `tofu cauliflower avocado` as a single entity, correct? Do you see a way how we could avoid that or is that something we don't want to support.

tabergma (approved):
Looks good! Thanks for addressing my comments 💯 

Just one minor comment left and one question.

rasa/nlu/extractors/extractor.py
tttthomasssss: Thats a tricky case and I have a feeling that trying to handle this might introduce more complexity and problems than it solves. 

We might want to re-visit this issue though if it turns out to be an annoying & recurring thing for users.

tabergma: yeah, not sure if we can actually support all the different scenarios. Let's merge as it is and then we can see what users think about it. What do you think?

tttthomasssss: 💯 - that sounds like the right plan!

PR #7105, opened 2020-10-26
Hotfix: tracking consent in docs

lunelson:
This is a hotfix, to include our [Klaro](https://kiprotect.com/docs/klaro/) consent management script in the docs deployments.

### DIFF
--- a/docs/plugins/google-tagmanager/index.js
+++ b/docs/plugins/google-tagmanager/index.js
@@ -20,8 +20,6 @@ module.exports = function (context, options) {
     );
   }
 
-  const isProd = process.env.NODE_ENV === 'production';
-
   return {
     name: 'google-tagmanager',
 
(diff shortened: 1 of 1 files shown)


docs/plugins/google-tagmanager/index.js
```diff
@@ -20,19 +20,14 @@ module.exports = function (context, options) {
...
   return {
     name: 'google-tagmanager',
 
     getClientModules() {
-      return isProd ? [path.resolve(__dirname, './client')] : [];
+      return [path.resolve(__dirname, './client')];
     },
 
     injectHtmlTags() {
-      if (!isProd) {
```
m-vdb, line 33: why do we want to include Klaro in dev/demo versions?

m-vdb (commented):
thanks for the hotfix

docs/plugins/google-tagmanager/index.js
lunelson: This is just because I need a way to test it locally. Results from non-production domains are still filtered out in GA though

m-vdb: ok 👍

m-vdb (approved):
🍏

PR #7106, opened 2020-10-26
Replace "PERSON" to "PER" in the dimensions

brunohjs:
**Proposed changes**:
- Change "PERSON" to "PER" in the dimensions parameter of the SpacyEntityExtractor component. This is how the component is identifying this entity in Rasa.

### DIFF
--- a/docs/docs/components.mdx
+++ b/docs/docs/components.mdx
@@ -1711,7 +1711,7 @@ The `SpacyEntityExtractor` extractor does not provide a `confidence` level and w
   pipeline:
   - name: "SpacyEntityExtractor"
     # dimensions to extract
-    dimensions: ["PERSON", "LOC", "ORG", "PRODUCT"]
+    dimensions: ["PER", "LOC", "ORG", "PRODUCT"]
   ```
 
 


sara-tagger:

akelad:
Hey @brunohjs, are you sure about this? As far as I'm aware the spacy entity is `PERSON` and not `PER`?

brunohjs:
> Hey @brunohjs, are you sure about this? As far as I'm aware the spacy entity is `PERSON` and not `PER`?

Yes, I'm sure. I know Spacy considers "PERSON", but Rasa appears as "PER". Look at this print:
image

brunohjs:
Now it's clear. When using the `en_core_web_md` model, the entity is `PERSON`. When using the `pt_core_news_lg` model (my case), the entity is `PER`.

image


So, there is no need to change the documentation. Maybe add a note for this case.

akelad:
oh, that's strange on spacy's part! i think given that the example in the docs is just an example, there's no need to add a note. By default rasa will extract all spacy entities anyways. but let me know if you think there's a better place to make a note of this

brunohjs:
I found in the spaCy documentation the reason for this problem. I found out that the Portuguese models are based on Wikipedia corpus and, because of that, have different entity tags.
image

English models are based on OneNote 5 corpus.

image

Link: https://spacy.io/api/annotation#ner-wikipedia-scheme

Maybe create a forum thread to make a note of that?

PR #7107, opened 2020-10-27
regex support non-whitespace language

w7lau:
add non-withespace language list

**Proposed changes**:
- add the language pass to pattern_utils
- check the language if non-withespace so do not add the word boundary

non-withespace language like Chinese should not add the word boundary


sara-tagger:

rasa/nlu/utils/pattern_utils.py
```diff
@@ -49,7 +53,9 @@ def _generate_lookup_regex(lookup_table: Dict[Text, Union[Text, List[Text]]]) ->
 
     # sanitize the regex, escape special characters
     elements_sanitized = [re.escape(e) for e in elements_to_regex]
-
+    not_boundary_language = ["zh", "ja", "th"]
+    if str(config_language).lower in not_boundary_language:
```
erohmensing, line 57:
```suggestion
    if str(config_language).lower() in not_boundary_language:
```

rasa/nlu/utils/pattern_utils.py
```diff
@@ -49,7 +53,9 @@ def _generate_lookup_regex(lookup_table: Dict[Text, Union[Text, List[Text]]]) ->
 
     # sanitize the regex, escape special characters
     elements_sanitized = [re.escape(e) for e in elements_to_regex]
-
+    not_boundary_language = ["zh", "ja", "th"]
```
erohmensing, line 56: can we move these into a constant `NON_WHITESPACE_LANGUAGES` in `rasa/nlu/constants.py`?

rasa/nlu/utils/pattern_utils.py
```diff
@@ -21,13 +22,16 @@ def _convert_lookup_tables_to_regex(
...
+        regex_pattern = _generate_lookup_regex(table,config_language=config_language)
         lookup_regex = {"name": table["name"], "pattern": regex_pattern}
         patterns.append(lookup_regex)
     return patterns
 
 
-def _generate_lookup_regex(lookup_table: Dict[Text, Union[Text, List[Text]]]) -> Text:
+def _generate_lookup_regex(
+        lookup_table: Dict[Text, Union[Text, List[Text]]],
+        config_language: Text = None,
```
erohmensing, line 33: Each config must have a language, I can't think of a time where this would be `None`, can you? If we provide `None` for this, we'll get an exception when we try to `lower()` it.

erohmensing (changes requested):
Thanks for the PR! This improvement makes a lot of sense 👌 I have some suggestions around the implementation. 

Would you also do a few things as per the PR template:
- format the code with `black`, otherwise it will not pass CI
- add a changelog entry, so that other users using non-whitespaced tokenized languages know they can use this feature 
- add a test in the `test_patten_utils.py`

rasa/nlu/utils/pattern_utils.py
w7lau: Of course !

erohmensing:
Going to remove my request until you're ready for another review after addressing comments 🙂

rasa/nlu/utils/pattern_utils.py
w7lau: I am a newer in python, so i don't know what will happen and how it work, i will listen to your suggestions

w7lau:
> Going to remove my request until you're ready for another review after addressing comments 🙂



> Going to remove my request until you're ready for another review after addressing comments 🙂

It's ok, i change the .lower() and check the language is not none, if none let it go to default code.

w7lau:
> Thanks for the PR! This improvement makes a lot of sense 👌 I have some suggestions around the implementation.
> 
> Would you also do a few things as per the PR template:
> 
> * format the code with `black`, otherwise it will not pass CI
> * add a changelog entry, so that other users using non-whitespaced tokenized languages know they can use this feature
> * add a test in the `test_patten_utils.py`

Ok, i will do this.

w7lau:
@erohmensing What issue type should i use in changelog?

w7lau:
> @erohmensing What issue type should i use in changelog?

And how can i know the test_patten_utils.py is successful ?

PR #7131, opened 2020-10-28
Bring DIET into TED

tabergma:
**Proposed changes**:

- add utterance transformer to TED

part of #6670 


rasa/core/policies/ted_policy.py
```diff
@@ -755,39 +812,110 @@ def _emebed_dialogue(
...
+        else:
+            # resulting attribute features will have shape
+            # combined batch dimension and dialogue length x 1 x units
+            attribute_features = self._combine_sparse_dense_features(
+                tf_batch_data[attribute][SENTENCE],
+                f"{attribute}_{SENTENCE}",
+                mask=attribute_mask,
+            )
+
+        if attribute in SENTENCE_FEATURES_TO_ENCODE + LABEL_FEATURES_TO_ENCODE:
```
Ghostvv, line 865: I think we still need to do that for sequence features because we convert from transformer units (128) to encoding units (50), or remove it all together and transform everything into 128

rasa/core/policies/ted_policy.py
```diff
@@ -755,39 +812,110 @@ def _emebed_dialogue(
...
+
+            attribute_features = self._convert_to_original_shape(
+                attribute_features, dialogue_lengths
+            )
+
+            # create a attribute mask that has the shape
+            # batch x dialogue length
+            attribute_mask = tf.expand_dims(
+                tf.squeeze(self._compute_mask(tf.squeeze(dialogue_lengths))), axis=-1
+            )
```
Ghostvv, line 883: `attribute_mask` is a different mask, it is not about dialogue lengths, but it tracks whether the input is fake input (zero features) or real input. E.g. where there is intent, there is no text and vice versa

rasa/utils/tensorflow/model_data.py
```diff
@@ -1123,10 +1143,19 @@ def _scipy_matrix_to_values(array_of_sparse: FeatureArray) -> List[np.ndarray]:
...
 
     @staticmethod
-    def _4d_scipy_matrix_to_values(array_of_array_of_sparse: FeatureArray):
+    def _4d_scipy_matrix_to_values(
+        array_of_array_of_sparse: FeatureArray,
+    ) -> List[np.ndarray]:
         # in case of dialogue data we may have 4 dimensions
         # batch size x dialogue history length x sequence length x number of features
 
+        # as transformers cannot handle 4D tensors pad and reshape the data
```
Ghostvv, line 1152:
```suggestion
        # transformers cannot handle 4D tensors, therefore pad and reshape the data
```

rasa/core/policies/ted_policy.py
```diff
@@ -17,7 +17,15 @@
...
 from rasa.core.featurizers.single_state_featurizer import SingleStateFeaturizer
-from rasa.shared.nlu.constants import ACTION_TEXT, ACTION_NAME, INTENT, TEXT, ENTITIES
+from rasa.shared.nlu.constants import (
+    ACTION_TEXT,
+    ACTION_NAME,
+    INTENT,
+    TEXT,
+    ENTITIES,
+    VALID_FEATURE_TYPES,
+    FEATURE_TYPE_SENTENCE,
```
Ghostvv, line 27: what is the difference to `SENTENCE` from `rasa.utils.tensorflow.constants`?

tabergma: basically the naming, would make it explicit and use the actual constant for `feature_type` when we are dealing with the feature type. just in case we change it.

rasa/core/policies/ted_policy.py
tabergma: `SENTENCE_FEATURES_TO_ENCODE` includes the `SEQUENCE_FEATURES_TO_ENCODE`, but will update the check to make it more explicit

rasa/core/policies/ted_policy.py
tabergma: that is actually exactly what `dialogue_length` is storing in the 3D way... it is 32 x 27 x 1, with just 1 or 0 depending on whether the dialogue turn is present or not. should we maybe rename it?

Ghostvv: but it's not about the dialogue turn, attribute mask stores whether the certain attribute is present or it is the fake one

rasa/core/policies/ted_policy.py
Ghostvv: then, shall we rename constants in `rasa.utils.tensorflow.constants`?

Ghostvv: I'd stick to the shorter names, but I don't have strong opinion here

rasa/core/policies/ted_policy.py
tabergma: just realized that dialogue length is not attribute specific. so it just indicates in general if we have any actual (aka non zero) features for this dialogue or not. I think we actually need the mask per attribute... we might need to add more info to the model data in that case.

tabergma: found a much simpler way... we can just reuse the "4D" attribute mask

tabergma:
argh... it is not working in non-eager mode 😞 need to take another look tomorrow.

rasa/core/policies/rule_policy.py
```diff
@@ -853,7 +853,9 @@ def predict_action_probabilities(
                 None,
             )
 
-        rules_action_name, source = self._find_action_from_rules(tracker, domain)
+        rules_action_name, source = self._find_action_from_rules(
+            tracker, domain, use_text_for_last_user_input=False
+        )
```
tabergma, line 858: @Ghostvv I merged master into `e2e` and got some merge conflicts. I think I messed up something here. Also one test is failing that should be related to that. Can you please take a look?

tabergma: This was the merge commit: https://github.com/RasaHQ/rasa/commit/3ed6d0421d47a74244ebe0a9ab76aeb07c713703

Ghostvv: pushed to e2e and merged e2e here

tabergma:
@Ghostvv Ready to be merged from my side. Can you please do a final review? Thanks.

rasa/utils/tensorflow/model_data.py
```diff
@@ -1135,8 +1167,8 @@ def _4d_scipy_matrix_to_values(array_of_array_of_sparse: FeatureArray):
                 for array_of_sparse in array_of_array_of_sparse
             ]
 
-        max_dialogue_len = max(
-            [len(array_of_sparse) for array_of_sparse in array_of_array_of_sparse]
+        max_dialogue_len = sum(
```
Ghostvv, line 1170: it is not a `max_dialogue_len`, right? I think smth like `combined_dialogue_len` would make more sense here

rasa/utils/tensorflow/model_data_utils.py
```diff
@@ -404,7 +399,7 @@ def _extract_features(
         # create a mask for every state
         # to capture which turn has which input
         attribute_mask = np.expand_dims(
-            np.ones(len(list_of_list_of_features), np.float32), -1
+            np.expand_dims(np.ones(len(list_of_list_of_features), np.float32), -1), -1
```
Ghostvv, line 402: could you please add comment about expand dims, and what is resulting shape

rasa/utils/tensorflow/model_data_utils.py
```diff
@@ -415,9 +410,10 @@ def _extract_features(
                 list_of_features = zero_features
```
Ghostvv, line 410: on line before, do we need to update `attribute_mask[i] = 0` due to additional dim?

Ghostvv (approved):
So simple in the end, great job finding the right tf methods!

Issue #7039, opened 2020-10-16
Next steps End-to-end: Implementation Proposal + Implementation


wochinge:
[Short scope document](https://www.notion.so/rasa/End-to-End-processing-aab7d87a1b9b43c3aca27cee68537972)

Policies have to be able to return events to store information about the used featurization in the tracker.

wochinge:
I think it would be cool to use a context manager to save the tracker. Currently we save the tracker quite often in between which is 1) Complex from a code-perspective and 2) Slow.

I'd imagine something like 

```
with tracker_store.get_tracker(sender_id) as tracker:
   # ... do things
```

This would mean that a tracker would only be saved **after** at the end of the interaction (aka we're waiting for the next user message).

wochinge:
@Ghostvv I thought about it some more:
* I'd implement "policies returning events" against `master`. This keeps the diff for end-to-end smaller and you can just use it in your `e2e` implementation then
* We need a deprecation warning for policies not returning the new items
* I think `DefinePrevUserUtteredFeaturization` could use `apply_to` to update the `UserUttered` event instead of `applied_events`. This way the whole logic would be nicely encapsulated within the event and not spread out too much.
* I think a separate event is indeed better than manipulating the `UserUttered` event directly. This has two advantages for me
  - more in line with our philosophy of unmodifiable events and that the tracker events give a precise history of what happened
  - allows the policies to manipulate the tracker by only using events. Directly modifying the event would require some less abstracted operations which are less generalizable / reusable. 

My idea from the comment above would be a nice refactoring, but it's not required for these changes

Ghostvv:
> I'd implement "policies returning events" against master.

ok, the only problem, that you couldn't properly test it

> I think `DefinePrevUserUtteredFeaturization` could use apply_to to update the UserUttered event instead of applied_events. 

as we discussed I prefer this one as well, but it'll work only if trackers are stored at the right moment

> more in line with our philosophy of unmodifiable events and that the tracker events give a precise history of what happened

but if you do it with `DefinePrevUserUtteredFeaturization.apply_to` you will modify `UserUttered` event

wochinge:
> ok, the only problem, that you couldn't properly test it

I mean your approach already works, right? That would simply be the same changes just a bit more polished. And we can write tests for it and so on. Worst case is that you can cherry pick the changes and we quickly try it out in `e2e`?

[...]

PR #5673, opened 2020-04-20
Attention weight logging

JEM-Mosig:
**Proposed changes**:
- Closes https://github.com/RasaHQ/research/issues/73

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)
- [ ] List this as experimental feature [here](https://www.notion.so/rasa/Experimental-Features-bc65f942fc924fd89d74b6eda2023163)

PR #7136, opened 2020-10-29
Policy Predictions for End-to-End

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7039
- `Policy`s now return a `PolicyPrediction` object instead of the plain probabilities. This allows us to return multiple pieces after a prediction. This is not only useful for this PR but also for [this PR](https://github.com/RasaHQ/rasa/pull/5673#discussion_r514164086)
- `PolicyEnsemble` now returns a `PolicyPrediction` object instead of `Tuple[probabilities, policy_name]`. The reason is the same for the change in the `Policy` interface: This allows us to encapsulate more information and pass this up to the `MessageProcessor`
- type the whole `test_policies` module
- add method `predict_next_with_tracker` to avoid having to reimplement some predict logic within `server`
- introduce a constant for `followup_action`


rasa/core/policies/policy.py
```diff
@@ -293,6 +293,78 @@ def _default_predictions(domain: Domain) -> List[float]:
...
+class PolicyPrediction:
+    """Stores information about the prediction of a `Policy`."""
+
+    def __init__(
+        self,
+        probabilities: List[float],
+        policy_priority: int,
+        events: Optional[List[Event]] = None,
+        optional_events: Optional[List[Event]] = None,
+        is_end_to_end_prediction: bool = False,
```
wochinge, line 305: should we also include the policy name here?

tests/core/test_policies.py
```diff
@@ -1,9 +1,14 @@
 from pathlib import Path
-from typing import Type, List, Text
+from typing import Type, List, Text, Tuple, Optional, Any
 from unittest.mock import Mock, patch
 
 import numpy as np
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
```
wochinge, line 7: sorry, typing got a bit out of hand here 🙈

wochinge:
I'll have an engineer review this as soon as you're ok with the changes in general @Ghostvv

rasa/core/policies/ensemble.py
```diff
@@ -526,14 +539,22 @@ def _pick_best_policy(
...
                 best_policy_name = form_policy_name
 
-        return predictions[best_policy_name].probabilities, best_policy_name
+        best_prediction = predictions[best_policy_name]
+
+        # Apply policy events to tracker
+        policy_events += best_prediction.optional_events
+
+        for event in policy_events:
+            tracker.update(event, domain)
```
Ghostvv, line 548: this is still illegal thing to do?

rasa/core/policies/ensemble.py
```diff
@@ -526,14 +539,22 @@ def _pick_best_policy(
...
+
+        return best_prediction, best_policy_name
```
Ghostvv, line 550: why don't we put `best_policy_name` inside `best_prediction`?

rasa/core/policies/policy.py
Ghostvv: I think so, since we any how return a structure, why not add additional field, then return of several methods can be simplified

rasa/core/policies/policy.py
```diff
@@ -293,6 +293,78 @@ def _default_predictions(domain: Domain) -> List[float]:
...
+        return (
+            self.probabilities == other.probabilities
+            and self.policy_priority == other.policy_priority
+            and self.events == other.events
+            and self.optional_events == other.events
+            and self.is_end_to_end_prediction == other.is_end_to_end_prediction
+        )
+
+    @property
+    def max_index(self) -> int:
```
Ghostvv, line 350: I don't think `max_index` is intuitive name, in numpy it is called `argmax`, or we can call it here explicitly `max_confidence_index` or smth

rasa/core/policies/rule_policy.py
```diff
@@ -828,15 +839,27 @@ def predict_action_probabilities(
...
-        rules_action_name, source = self._find_action_from_rules(tracker, domain)
+        rules_action_name, source, entered_unhappy_path = self._find_action_from_rules(
+            tracker, domain
+        )
         # we want to remember the source even if rules didn't predict any action
         self._prediction_source = source
+
+        policy_events = []
+        if entered_unhappy_path:
+            policy_events.append(LoopInterrupted(True))
```
Ghostvv, line 855: what about simplifying it?
```suggestion
        policy_events = [LoopInterrupted(True)] if entered_unhappy_path else []
```

rasa/core/policies/rule_policy.py
```diff
@@ -783,14 +787,14 @@ def _find_action_from_rules(
...
-                    return active_loop_name, LOOP_RULES
+                    return active_loop_name, LOOP_RULES, entered_unhappy_path
 
                 # do not predict anything
                 predicted_action_name = None
 
             if DO_NOT_VALIDATE_LOOP in unhappy_path_conditions:
-                logger.debug("Added `FormValidation(False)` event.")
-                tracker.update(LoopInterrupted(True))
+                logger.debug("Entered unhappy path. Loop will be interrupted.")
```
Ghostvv, line 796: this is happening when we return from unhappy path to the loop
```suggestion
                logger.debug("Returning from unhappy path. Loop will be notified that it was interrupted.")
```

rasa/core/policies/rule_policy.py
```diff
@@ -739,12 +741,14 @@ def _find_action_from_loop_happy_path(
...
     def _find_action_from_rules(
         self, tracker: DialogueStateTracker, domain: Domain
-    ) -> Tuple[Optional[Text], Optional[Text]]:
+    ) -> Tuple[Optional[Text], Optional[Text], bool]:
         tracker_as_states = self.featurizer.prediction_states([tracker], domain)
         states = tracker_as_states[0]
 
         logger.debug(f"Current tracker state: {states}")
 
+        entered_unhappy_path = False
```
Ghostvv, line 750: please add comment why it is here
```suggestion
        returning_from_unhappy_path = False
```

rasa/core/policies/rule_policy.py
```diff
@@ -783,14 +787,14 @@ def _find_action_from_rules(
                         f"Predicted loop '{active_loop_name}' by overwriting "
                         f"'{ACTION_LISTEN_NAME}' predicted by general rule."
                     )
-                    return active_loop_name, LOOP_RULES
+                    return active_loop_name, LOOP_RULES, entered_unhappy_path
 
                 # do not predict anything
                 predicted_action_name = None
 
             if DO_NOT_VALIDATE_LOOP in unhappy_path_conditions:
```
Ghostvv, line 795: maybe rename `DO_NOT_VALIDATE_LOOP` also

rasa/core/policies/ensemble.py
wochinge: Why is it illegal? We add the events **after** all policies made their prediction so they are not affecting each other for the current action.

Ghostvv: but we update the object that we pass as an input

Ghostvv:
part of https://github.com/RasaHQ/rasa/issues/6670

rasa/core/policies/ensemble.py
wochinge: what's the problem with that?

Ghostvv: you pass one object to the method, but after it you effectively get the different one. I don't know. I thought this was the problem with updating tracker from inside the policies, because there is not so much difference whether it is done in ensemble or policies. cc @tmbo

wochinge: In my opinion it's different if you do it **within** the policies as this might affect the predictions of other policies for the current state. Consider having an ensemble of two policies. The first one is asked to predict the probabilities and adds a featurized `SlotSet` event to the tracker. The second policy would then predict based on a different state. In this PR we update the tracker **after** all predictions were done which will potentially affects the next action prediction but not the current one.

Ghostvv: with slot set yes, but back then when we had only `FormValidation` that we know doesn't influence any prediction, we had a discussion that tracker shouldn't be updated anywhere outside processor (or whatever class handles adding predicted events), because it is a bad practice from programming point view, because it is preferred to have immutable inputs or whatever is the reason

wochinge: I think it makes sense for the `ensemble` to apply the events because there is some logic to it (e.g. optional events are only applied for the winning policy) which in my opinion is tied to the model (I don't see the `processor` as part of the model). 

> because it is a bad practice from programming point view, because it is preferred to have immutable inputs or whatever is the reason

I don't see the risk in doing so. The `tracker` represents the conversation state so it's normal that that things get appended to it as we go. I can expose the the policy events to the processor but I'm not sure if it makes anything simpler or more transparent.

rasa/core/policies/ensemble.py
wochinge: we'd loose the information about the index of the policy 😬

wochinge: as far as I know policy names aren't unique, right?

rasa/core/policies/ensemble.py
Ghostvv: I don't know, I just remember that there was a discussion that we'd like to avoid it. Let's ask in the engineering team what others think.

wochinge: I had another look and it would reduce the amount of individual parameters which we pass around in `processor` by encapsulating everything within a `PolicyPrediction` object which would be desirable. 

@Ghostvv Would we throw away the policy events in case the predicted action was rejected? My intuition says yes.

wochinge: @federicotdn @degiz What do you think? 

Context: `Policy`s can now return `Event`s. Previously only actions and NLU predictions returned events which were applied to the tracker within the `processor` module. I am currently applying the events which are returned by the `Policy`s within the `PolicyEnsemble`. I am indifferent where we apply these events to the tracker. I'm currently slightly leaning towards `Processor` we could use this to reduce the number of parameters in our methods within `processor`. Namely we wouldn't pass around `action_name`, `confidence`, and `policy_name` separately, but could bundle them within one `prediction` argument.

Ghostvv: yes, if action was rejected, it never happened, therefore all the events which came with its predictions are irrelevant

rasa/core/policies/ensemble.py
Ghostvv: aaahh. yes, we need an index. Maybe add a comment about it

rasa/core/policies/ensemble.py
wochinge: But what if the action needs the policy events? 😆 
This is actually something which requires us to move the application of the events to the tracker to processor.

Ghostvv: aaah, yeah good catch

wochinge: It was definitely worth doing it. However, the change got bigger than expected. Will work in the remaining todos tomorrow.

wochinge:
I'll self-review this later, polish, add docstrings and then request the actual review.

rasa/core/policies/policy.py
```diff
@@ -353,6 +358,104 @@ def format_tracker_states(self, states: List[Dict]) -> Text:
...
 
 
+class PolicyPrediction:
+    """Stores information about the prediction of a `Policy`."""
+
+    def __init__(
+        self,
+        probabilities: List[float],
+        policy_name: Optional[Text],
+        policy_priority: int = 1,
```
wochinge, line 368: I think a default of `1` is way better than a default of `None` as it's more robust. Developers tend to forget that it can `None` and then everything explods.

wochinge:
@degiz Do you have time? It looks bigger than it is. Mostly typing and renames.

degiz:
@wochinge can review that tomorrow in the morning, would it be fine?

wochinge:
Sure, absolutely no rush!

rasa/core/policies/policy.py
```diff
@@ -353,6 +358,104 @@ def format_tracker_states(self, states: List[Dict]) -> Text:
...
+        self.events = events or []
+        self.optional_events = optional_events or []
+        self.is_end_to_end_prediction = is_end_to_end_prediction
+
+    @staticmethod
+    def for_action_name(
+        domain: Domain,
+        action_name: Text,
+        policy_name: Optional[Text] = None,
+        confidence: float = 1,
```
degiz, line 402:
```suggestion
        confidence: float = 1.0,
```

rasa/core/policies/policy.py
```diff
@@ -353,6 +358,104 @@ def format_tracker_states(self, states: List[Dict]) -> Text:
...
+        return self.probabilities.index(self.max_confidence)
+
+    @property
+    def max_confidence(self) -> float:
+        """Gets the highest predicted probability.
+
+        Returns:
+            The highest predicted probability.
+        """
+        return max(self.probabilities, default=0)
```
degiz, line 456:
```suggestion
        return max(self.probabilities, default=0.0)
```

rasa/core/policies/rule_policy.py
```diff
@@ -59,7 +59,7 @@
 RULES = "rules"
 RULES_FOR_LOOP_UNHAPPY_PATH = "rules_for_loop_unhappy_path"
 
-DO_NOT_VALIDATE_LOOP = "do_not_validate_loop"
+LOOP_WAS_INTERRUPTED = "do_not_validate_loop"
```
degiz, line 62: Does it make sense to keep the old version on the string value 🤔 ? Perhaps the backward compatibility?

rasa/core/policies/rule_policy.py
```diff
@@ -739,13 +741,18 @@ def _find_action_from_loop_happy_path(
 
     def _find_action_from_rules(
```
degiz, line 742: Could we add a docstring at least for the return types here?

rasa/core/processor.py
```diff
@@ -828,33 +875,36 @@ def _has_session_expired(self, tracker: DialogueStateTracker) -> bool:
...
-            if result:
-                return result
-            else:
-                logger.error(
-                    f"Trying to run unknown follow-up action '{followup_action}'!"
-                    "Instead of running that, we will ignore the action "
-                    "and predict the next action."
+            if followup_action in self.domain.action_names:
+                return PolicyPrediction.for_action_name(
+                    self.domain, followup_action, "followup_action"
```
degiz, line 887: Could we extract `followup_action` as a const?

rasa/core/processor.py
```diff
@@ -828,33 +875,36 @@ def _has_session_expired(self, tracker: DialogueStateTracker) -> bool:
...
             tracker, self.domain, self.interpreter
         )
+
+        if isinstance(prediction, PolicyPrediction):
+            return prediction
+
+        rasa.shared.utils.io.raise_deprecation_warning(
+            "Returning a tuple of probabilities and policy name for "
+            "`PolicyEnsemble.probabilities_using_best_policy` is deprecated and will "
+            "be removed in Rasa Open Source 3.0. Please return a `PolicyPrediction` "
```
degiz, line 906: The same comment about `3.0` VS `3.0.0`

rasa/shared/core/trackers.py
```diff
@@ -589,10 +590,31 @@ def update(self, event: Event, domain: Optional[Domain] = None) -> None:
...
+        new_events: List[Event],
+        domain: Optional[Domain],
+        override_timestamp: bool = True,
+    ) -> None:
+        """Adds multiple events to the tracker.
+
+        Args:
+            new_events: Events to apply.
+            domain: The current model's domain.
+            override_timestamp: If `True` reset all timestamps of the events. As the
```
degiz, line 604: Maybe `actualize` instead of `reset`? Reset sounds to me like "set to zero".

tests/core/test_processor.py
```diff
@@ -866,3 +880,155 @@ async def mocked_run(self, *args: Any, **kwargs: Any) -> List[Event]:
 
     assert ActionExecuted("utter_greet") not in logged_events
     assert all(event in logged_events for event in rejection_events)
+
+
+def test_predict_next_action_with_deprecated_ensemble(
```
degiz, line 885: 🚀

degiz (approved):
Great job adding all the docstrings, comments, types and tests 🚀 🚀

rasa/core/agent.py
```diff
@@ -548,14 +548,16 @@ async def execute_action(
...
-        policy: Text,
-        confidence: float,
+        policy: Optional[Text],
+        confidence: Optional[float],
     ) -> Optional[DialogueStateTracker]:
         """Handle a single message."""
-
         processor = self.create_processor()
+        prediction = PolicyPrediction.for_action_name(
+            self.domain, action, policy, confidence or 1
```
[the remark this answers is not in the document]
Ghostvv: memo can predict action_listen with confidence `0`, let's keep it `0`

rasa/core/policies/policy.py
Ghostvv: wait, no. I think, it shouldn't have default value. `Policy._prediciton` method should set it to `Policy.priority`

rasa/core/policies/rule_policy.py
```diff
@@ -739,13 +741,18 @@ def _find_action_from_loop_happy_path(
...
+    ) -> Tuple[Optional[Text], Optional[Text], bool]:
         tracker_as_states = self.featurizer.prediction_states([tracker], domain)
         states = tracker_as_states[0]
         current_states = self.format_tracker_states(states)
 
         logger.debug(f"Current tracker state:{current_states}")
 
+        # Tracks if we are returning after an unhappy loop path. If this becomes `True`
+        # the policy returns an event which disables the slot validation for the next
+        # loop execution.
```
Ghostvv, line 753: this comment mixes forms and loops. Maybe smth along the lines:
"For example, if `active_loop` is a form, then slot validation is disabled"

Ghostvv (approved):
🚀

rasa/core/policies/ensemble.py
```diff
@@ -449,18 +450,24 @@ def _check_if_rule_policy_used_with_rule_like_policies(
...
+        Args:
+            policy_name: The name of the policy.
+            max_confidence: The max confidence of the policy's prediction.
+
+        Returns: `True` if it's a `MemoizationPolicy`, `False` otherwise.
+        """
+        if not policy_name:
+            return True
+
         is_memo = policy_name.endswith("_" + MemoizationPolicy.__name__)
```
wochinge, line 471: @Ghostvv That should now also include the `RulePolicy`, shouldn't it?

rasa/core/policies/ensemble.py
```diff
@@ -449,18 +450,24 @@ def _check_if_rule_policy_used_with_rule_like_policies(
...
-
 class SimplePolicyEnsemble(PolicyEnsemble):
+    """Default implementation of a `Policy` ensemble."""
+
     @staticmethod
     def is_not_memo_policy(
-        policy_name: Text, max_confidence: Optional[float] = None
+        policy_name: Optional[Text], max_confidence: Optional[float] = None
     ) -> bool:
+        """Checks if the policy is the `MemoizationPolicy` or one of its subclasses.
```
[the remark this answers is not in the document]
wochinge: re-named the function and changed the docstring. Better?

rasa/core/policies/ensemble.py
```diff
@@ -449,18 +450,24 @@ def _check_if_rule_policy_used_with_rule_like_policies(
...
-    priority: int
-
-
 class SimplePolicyEnsemble(PolicyEnsemble):
+    """Default implementation of a `Policy` ensemble."""
+
     @staticmethod
     def is_not_memo_policy(
-        policy_name: Text, max_confidence: Optional[float] = None
+        policy_name: Optional[Text], max_confidence: Optional[float] = None
```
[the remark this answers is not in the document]
wochinge: No as this would otherwise evaluate to `True` for the call [here](https://github.com/RasaHQ/rasa/blob/82e98163d3701f490716760213b92619283a4a00/rasa/core/test.py#L575)

rasa/core/policies/ensemble.py
```diff
@@ -526,14 +539,24 @@ def _pick_best_policy(
...
-        return predictions[best_policy_name].probabilities, best_policy_name
+        best_prediction = predictions[best_policy_name]
+
+        policy_events += best_prediction.optional_events
+
+        return PolicyPrediction(
+            best_prediction.probabilities,
+            best_policy_name,
+            best_prediction.policy_priority,
+            policy_events,
```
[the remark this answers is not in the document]
wochinge: `optional_events` of policy predictions are only considered if a policy wins (see line 544). So we have to separate "must apply" events from "optional events" for the single predictions of the policies. Once we have settled on a winner we can merge the "must apply" events from all policies with the optional events from the winning policy. Does that make sense?

rasa/core/policies/ensemble.py
```diff
@@ -608,15 +631,27 @@ def _get_prediction(
...
+                "adapt your custom `Policy` implementation. Support for returning "
+                "a list of floats will be removed in Rasa Open Source 3.0."
+            )
+            prediction = PolicyPrediction(
+                prediction, policy.__class__.__name__, policy_priority=policy.priority
+            )
+
+        return prediction
 
     def _fallback_after_listen(
```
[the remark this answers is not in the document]
wochinge: It gets quite long and the docstring explains it quite well. I think the precise thing would be something like `_fallback_after_second_action_listen_in_a_row`  🙈 I'd opt to not do this if it's ok.

rasa/core/policies/rule_policy.py
wochinge: I don't think the actual string value matters much. If we change it, then it means that we need to raise the minimal compatible model version which I'd like to avoid (cc @Ghostvv )

rasa/core/policies/rule_policy.py
wochinge: 👍 
It's easy to mix currently as every loop currently is a form 🙈

rasa/shared/core/trackers.py
wochinge: Changed it to `refresh`.

rasa/core/policies/ensemble.py
Ghostvv: hm, good catch! Probably yes, since we allow all policies together now

rasa/core/policies/rule_policy.py
Ghostvv: not really. two stage fallback is not a form

rasa/core/policies/ensemble.py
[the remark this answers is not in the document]
Ghostvv: I'd keep short version

rasa/core/policies/rule_policy.py
Ghostvv: this will be a part of minor release anyhow? So users should retrain. I'd change the value as well, to avoid confusion

wochinge: @tabergma How do I change `MINIMUM_COMPATIBLE_VERSION`? Intuitively I'd change the version in `version.py` as well because otherwise it won't allow training on master. However, the comment in `version.py` tells me to not modify it ? 🙈

tmbo: you can run `make release` and do an alpha

wochinge: but then I also need an SDK alpha 🙈 At that point it seems simpler to create an issue for the release and add the add update of `MINIMUM_COMPATIBLE_VERSION` as `TODO` there

rasa/core/policies/ensemble.py
wochinge: added + tests for it 👍

rasa/core/policies/rule_policy.py
wochinge: created an issue for it [here](https://github.com/RasaHQ/rasa/issues/7224) and will do it during the releaes

PR #7137, opened 2020-10-29
Merge 2.0.x to master

dakshvar22:
**Proposed changes**:
- Merge `2.0.x` to master after `2.0.3` release.


dakshvar22:
@wochinge There was a merge conflict in `pyproject.toml`. Looks like we added a new version of `pytype` in `2.0.x` which wasn't present in master. Because of that there were merge conflicts in `poetry.lock` as well. I kept the `2.0.x` version in that case and re-ran `poetry update`.

pyproject.toml
```diff
@@ -144,6 +144,7 @@ mongomock = "^3.18.0"
 black = "^19.10b0"
 flake8 = "^3.8.3"
 flake8-docstrings = "^1.5.0"
+pytype = "^2020.6.1"
```
wochinge, line 147: please remove this. We are using `mytype` on `master`

rasa/nlu/tokenizers/convert_tokenizer.py
```diff
@@ -7,17 +7,27 @@
 from rasa.nlu.tokenizers.whitespace_tokenizer import WhitespaceTokenizer
 from rasa.shared.nlu.training_data.message import Message
 from rasa.utils import common
+from rasa.nlu import utils as nlu_utils
```
wochinge, line 10: minor: In our [code guidelines](https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#4fbe14b7c3a04fbfa20fa0e76995ecf7) we suggested to use the absolute import instead of aliases

wochinge (changes requested):

pyproject.toml
dakshvar22: Okay, exactly that was what I wasn't sure of. Should I then revert back changes made to `poetry.lock` as well because it seems like there isn't any dependency update to do if this goes away?

wochinge: sounds good 👍

rasa/nlu/tokenizers/convert_tokenizer.py
dakshvar22: Hmm, but for imports like those of util functions, shouldn't we use aliases to make the import clearer because we could be importing functions from multiple util modules like `rasa.nlu.utils`, `rasa.shared.utils`? If these functions are declared in the `__init__.py` of the respective modules(like in the case above) then imports will clash.

dakshvar22: Ahh I guess you mean importing it as `rasa.nlu.utils` and not `from rasa.nlu import utils` 👍

wochinge: There is a whole discussion about this somewhere in Slack and in Notion. In the end we decided to go for this. 

> If these functions are declared in the __init__.py of the respective modules(like in the case above) then imports will clash.

With which other things would that crash?

wochinge: > Ahh I guess you mean importing it as rasa.nlu.utils and not from rasa.nlu import utils 👍

Yes, exactly 👍

wochinge (approved):
💯

PR #7141, opened 2020-10-30
Fix failure of rasa x CLI command if git is not installed

### LINKED ISSUE #7140: ImportError: Failed to initialize: Bad git executable.
Sentry Issue: [RASA-OPEN-SOURCE-180](https://sentry.io/organizations/rasahq/issues/1973792341/?referrer=github_integration)

```
ImportError: Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh()

All git commands will error until this is rectified.

This initial warning can be silenced or aggravated in the future by setting the
$GIT_PYTHON_REFRESH environment variable. Use one of the following values:
    - quiet|q|silence|s|none|n|0: for no warning or exception
    - warn|w|warning|1: for a print...
  File "__init__.py", line 83, in <module>
    refresh()
  File "__init__.py", line 73, in refresh
    if not Git.refresh(path=path):
  File "site-packages/git/cmd.py", line 278, in refresh
    raise ImportError(err)

ImportError: Failed to initialize: Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh()

All git commands will error until this is rectified.

This initial warning can be silenced or aggravated in the future by setting the
$GIT_PYTHON_REFRESH environment variable. Use one of the following values:
    - quiet|q|silence|s|none|n|0: for no warning or exception
    - warn|w|...
(3 additional frame(s) were not displayed)
...
  File "rasa/__main__.py", line 115, in main
  File "rasa/cli/x.py", line 339, in rasa_x
    run_locally(args)
  File "rasa/cli/x.py", line 428, in run_locally
    from rasax.community import local  # pytype: disable=import-error
```

tmbo:
**Proposed changes**:
- fixes #7140 

- [ ] ~added some tests for the functionality~

### DIFF
--- a/changelog/7140.bugfix.md
+++ b/changelog/7140.bugfix.md
@@ -1,0 +1,1 @@
+Fixed error when starting Rasa X locally without a proper git setup.
--- a/rasa/cli/x.py
+++ b/rasa/cli/x.py
@@ -426,15 +426,20 @@ def _get_credentials_and_endpoints_paths(
 def _prevent_failure_if_git_is_not_available() -> None:
     """Rasa X uses the `git` package, which will fail to import if git is not available.
 
-
     Git isn't needed locally, which means we can silence this error to allow
     users to use local mode even if git is not available on their machine.
-    Fixes regression https://github.com/RasaHQ/rasa/issues/7140"""
+    Fixes regression https://github.com/RasaHQ/rasa/issues/7140
+    """
     if os.environ.get("GIT_PYTHON_REFRESH") is None:
         os.environ["GIT_PYTHON_REFRESH"] = "quiet"
 
 
-def run_locally(args: argparse.Namespace):
+def run_locally(args: argparse.Namespace) -> None:
+    """Run a Rasa X instance locally.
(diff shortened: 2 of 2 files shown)


degiz (approved):
Thanks a lot for fixing that! 👍 
Do you think we need to add a log message in case `git` is not available? I'm not sure myself, since the IVC integration shouldn't be used locally anyway 🤔

tmbo:
> Do you think we need to add a log message in case git is not available? I'm not sure myself, since the IVC integration shouldn't be used locally anyway 🤔
That is a good point, I do think though that should happen at the place where we want to use git (and I think there might already be logging there, if not exceptions will be thrown on the first call to a git command). At this place in the code, we really are just importing and not using git, so adding a logger message would IMO just confuse the user.

Issue #7068, opened 2020-10-20
Allow forms without default slot mappings / only custom slots mappings


wochinge:
Currently, you need to add at least one slot mapping to your form in the domain. Otherwise, it will choose the Rasa Open Source 1.x `FormAction` for execution (see [here](https://github.com/RasaHQ/rasa/blob/ae8eee310a9cb00573d5d44650d2654436475cc4/rasa/core/actions/action.py#L127))

We need to detect the deprecated usage of old forms differently and allow that users can define a form like the following:

```yaml
forms:
   my_form:
     # no slot mappings
     # all slots are extracted in the custom validation action
```

wochinge:
@Ghostvv @akelad 

**Current Situation**
In case there are no slot mappings defined for a form, we currently assume that they want to use the deprecated Rasa Open Source 1 `FormAction`. We have to replace this check with a different one in order to allow forms without any slot mappings in the domain.

**Solutions**
I gathered two solutions which make sense in my opinion:
1. Allow users to specify a key `use_deprecated_form_action` for each form. This has more implications for Rasa X and code wise, but would be very explicit.
2. Allow users to override the form by specifying a custom action with the name of the form. This would be more general and would require less changes. However, it might be a bit unintuitive for users.

In general I'd prefer option 2 as it seems more "Rasa-style" and blends with how we override other "default" actions. Any strong opinions?

Ghostvv:
the first version still technically breaks compatibility? I thought the idea is to allow users to train/run old 1.x bots without any changes to the files

Ghostvv:
I'm not sure, I understand the solution 2, do you mean that if there is an action with the same name, we override the default? I thought we already do it, at least we did for `action_listen` in some old version

wochinge:
> the first version still technically breaks compatibility? I thought the idea is to allow users to train/run old 1.x bots without any changes to the files

In both cases we can check if the `forms` section is a list (old format) or a dict (new format) and set the flag / add the custom action for it on the fly. 

> 2, do you mean that if there is an action with the same name, we override the default? I thought we already do it, at least we did for action_listen in some old version

Yep, that would be the default. Currently overriding the default form action however isn't possible due to our current check which we have in place.

Ghostvv:
then I'd say number 2 is natural solution

Ghostvv:
the only problem, users might start to create their custom forms, and they wouldn't work, because specific events wouldn't be returned

wochinge:
[...]

PR #7171, opened 2020-11-03
optional slot mappings

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7068
- forms do not have to have slot mappings any more. They can be empty instead which allows moving all slot mappings to the custom validate action.
- the usage of the Rasa Open Source 2 `FormAction` can be overridden by defining a custom action with the name of the form. This can e.g. be used to use the deprecated `FormAction` from the Rasa SDK.
- When a `Domain` defines forms as a list of strings (Rasa Open Source 1 format) then it's assumed that the user wants to use the deprecated `FormAction` as well. A warning is printed.


Ghostvv:
what is the difference between defining forms as list of strings vs list of dicts with no values?

wochinge:
> what is the difference between defining forms as list of strings vs list of dicts with no values?

The list format is the old format for the `forms:` section which was used in Rasa Open Source. 
Rasa Open Source expects the `forms:` section to be provided as `Dict`

Ghostvv:
how does it look from user point of view in domain file? Should we document that?

wochinge:
> how does it look from user point of view in domain file? Should we document that?

What exactly do you mean? I don't think we should document an outdated format. Handling that is just a failsafe for users who haven't migrated their domain format yet. I can add another section to the documentation which describes how they can use the old SDK `FormAction` with a bunch of warnings 👍

Ghostvv:
what I mean is. If I were to define a bunch of forms without slot mappings, I'd put them in domain file similar as actions. How would it look different if I need to define them as a list of dicts

wochinge:
> f I were to define a bunch of forms without slot mappings, I'd put them in domain file similar as actions. How would it look different if I need to define them as a list of dicts

I made an example of forms without slot mappings in the changelog. It's simply as you would usually define forms - just that you leave apart the slot mappings. The old format within Rasa Open Source 1 was:

```
forms:
- form1
- form2
```

Ghostvv:
now I'm confused, sorry 🙈 but then it is a list of strings. But I'd like to use 2.0 forms

wochinge:
That's a Rasa Open Source 2.0 forms without slot mapping
```
forms:
  my_form:
    # no slot mappings
```

@akelad updated the migration guide with instructions

Ghostvv:
so 
```
forms:
  form1:
  form2:
```
or do I have to add a comment?

wochinge:
> or do I have to add a comment?

No comment needed

Ghostvv:
that is very weird format to me, but ok

rasa/shared/core/domain.py
```diff
@@ -569,34 +574,55 @@ def _collect_overridden_default_intents(
...
             rasa.shared.utils.io.raise_warning(
-                f"The `forms` section in the domain needs to contain a dictionary. "
-                f"Instead found an object of type '{type(forms)}'.",
+                "The `forms` section in the domain used the old Rasa Open Source 1 "
+                "list format to define forms. Rasa Open Source will be configured to "
+                "use the deprecated `FormAction` within the Rasa SDK. If you want to "
+                "use the new Rasa Open Source 2 `FormAction` adapt your `forms` "
+                "section as described in the documentation. Support for the "
+                "deprecated `FormAction` in the Rasa SDK will be removed in Rasa Open "
+                "Source 3.0.",
```
federicotdn, line 610:
```suggestion
                "The `forms` section in the domain used the old Rasa Open Source 1.0 "
                "list format to define forms. Rasa Open Source will be configured to "
                "use the deprecated `FormAction` within the Rasa SDK. If you want to "
                "use the new Rasa Open Source 2.0 `FormAction` adapt your `forms` "
                "section as described in the documentation. Support for the "
                "deprecated `FormAction` in the Rasa SDK will be removed in Rasa Open "
                "Source 3.0.",
```

rasa/core/actions/action.py
```diff
@@ -158,30 +148,40 @@ def is_retrieval_action(action_name: Text, retrieval_intents: List[Text]) -> boo
...
-    elif name.startswith(UTTER_PREFIX):
+
+    if name.startswith(UTTER_PREFIX):
         return ActionUtterTemplate(name)
-    elif should_use_form_action:
+
+    is_form = name in domain.form_names
+    # Users can override the form by defining an action with the same name as the form
+    user_overrode_form_action = is_form and name in domain.user_actions
+    if is_form and not user_overrode_form_action:
```
federicotdn, line 179:
```python
    if is_form and not user_overrode_form_action:
```
is equivalent to (`X and not (X and Y)`):
```python
    if is_form and not (is_form and name in domain.user_actions):
```
which in turn is equivalent to (`X and not Y`):
```python
    if is_form and name not in domain.user_actions:
```
So then maybe we could just write:
```python
    if name in domain.form_names and name not in domain.user_actions:
```

federicotdn (approved):
:+1: :+1:

rasa/shared/core/domain.py
wochinge: why do we need the `.0`? It was not only part of Rasa Open Source 1.0, but of all Rasa Open Source 1.x versions, right? Same for `2.0`: It's part of `2.x`

rasa/core/actions/action.py
wochinge: I assigned names to the single checks to make their meaning more explicit.

Developers might wonder what `name not in domain.user_actions` means while `user_overrode_form_action` conveys way more information about this. What do you think?

wochinge:
@melindaloubser1 Could you please have a look at the docs?

docs/docs/default-actions.mdx
```diff
@@ -148,3 +148,17 @@ their message.
...
+[form logic](forms.mdx). You can override this default action with a custom action by
+adding a custom action with the form's name to the domain. This can e.g. be used to use
+the deprecated `FormAction` which is part of the Rasa SDK and used to implement forms
+for Rasa Open Source 1.
+
+:::caution
+Overriding the default action for forms should **only** be used during the process of
+migrating from Rasa Open Source 1 to 2. It is highly discouraged to use this to
+implement custom form versions. Instead we recommend to use the customizations which
+are described in the [Advanced Usage](forms.mdx#advanced-usage) section of forms.
```
melindaloubser1, line 163:
```suggestion
Overriding the default action for forms should **only** be used during the process of
migrating from Rasa Open Source 1 to 2. Do not to use this to
implement custom form versions. To customize forms, see the [Advanced Usage](forms.mdx#advanced-usage) section of forms.
```
May be overkill on the simplification but "it's recommended/discouraged" is  🙅‍♀️

docs/docs/migration-guide.mdx
```diff
@@ -378,6 +378,25 @@ class RestaurantFormValidator(FormValidationAction):
             return {"cuisine": None}
 ```
 
+It is also possible to migrate forms to Rasa Open Source iteratively. You can for
```
melindaloubser1, line 381:
```suggestion
You can also migrate forms from Rasa SDK to Rasa Open Source 2 iteratively. You can for
```

docs/docs/migration-guide.mdx
```diff
@@ -378,6 +378,25 @@ class RestaurantFormValidator(FormValidationAction):
...
+example migrate one form to the Rasa Open Source 2 implementation while continue using
+the deprecated Rasa SDK implementation for another form. If you want to continue to use
+the deprecated Rasa SDK `FormAction`s, then you can instruct Rasa Open Source to do so
+by adding a custom action with the name of your form to your domain. Note that it is
+recommended to complete the migration as soon as possible as the deprecated `FormAction`
```
melindaloubser1, line 386:
```suggestion
the deprecated Rasa SDK implementation for another form. To continue to use
the deprecated Rasa SDK `FormAction`s, add a custom action with the name of your form to your domain. Note that you should complete the migration as soon as possible as the deprecated `FormAction`
```

docs/docs/default-actions.mdx
```diff
@@ -148,3 +148,17 @@ their message.
 This action undoes the last user-bot interaction. It can be triggered by the user
 by sending a "/back" message to the assistant if the [RulePolicy](./policies.mdx#rule-policy) is configured.
                                                                                                                                                                                                                                                                         |
+## Form Action
+
+By default Rasa Open Source uses the inbuilt `FormAction` for processing any
+[form logic](forms.mdx). You can override this default action with a custom action by
+adding a custom action with the form's name to the domain. This can e.g. be used to use
+the deprecated `FormAction` which is part of the Rasa SDK and used to implement forms
+for Rasa Open Source 1.
```
melindaloubser1, line 157:
```suggestion
By default Rasa Open Source uses `FormAction` for processing any
[form logic](forms.mdx). You can override this default action with a custom action by
adding a custom action with the form's name to the domain. This can e.g. be used to use
the deprecated `FormAction` which is part of the Rasa SDK and used to implement forms
for Rasa Open Source 1.
```

melindaloubser1 (changes requested):
Just some phrasing changes

docs/docs/default-actions.mdx
akelad: @wochinge didn't I say that we should just remove the "caution" note here and just state in the text that overriding should ONLY be done during migration? So basically say this directly in the text, rather than saying "This can e.g. be used".

wochinge: Yes, but I haven't worked that in yet.

wochinge: @melindaloubser1 Merged your suggestion with Akela's

docs/docs/default-actions.mdx
wochinge: Deleted the caution section as Akela suggested. Merged some parts of your suggestion to the new version

melindaloubser1 (approved):

docs/docs/default-actions.mdx
akelad: haha ok sorry, i assumed since you requested Melinda for review you weren't going to ;)

wochinge:
@Ghostvv @Ghostvv Do you want to have another look or can I merge?

Ghostvv (approved):
I don't think the format:
```
forms:
  form1:
  form2:
```
is intuitive or clear, the need or no need of `-` in `yaml` is a mess. Otherwise looks good to me

SuunnyyGoo:
Hello Everybody,

First thanks for your job, Rasa is a great toolbox, it helps us a lot ! 

I am facing the current issue and i dont found it clear, i'm using dist 0.37.0 | 2.3.1 | 2.3.1 with gke rasa-x-helm install.

I have a form used to meet the user, at first I started with slot mapping from the domain, but looking for more accuracy, I created a validate_form class in my custom action server to improve slot extraction. One by one I finnally added all my slots mappings in my custom action but now the form is still called but not effective anymore.
I read this topic and the linked ones but i'm still confused about the use of the FormAction which is said as a way to manage depreciation, is that case one sort of depreciation ? Because i don't want te create a whole FormAction, i just want to find out my validate_form class

So I tried to implement a FormAction inherited from Rasa Core with only the name and the required slots (keeping the validate_form class nearby, this FormAction might call it ?..), but it is not registred when starting the action server and my form is achieved right after starting (without asking questions).

So I tried to implement a FormAction from Rasa_sdk (which is commented as deprecated) with overides for the name, the required slots and the submit methods. Actions my_form and validate_my_form are well reguistred ! But when my form is accessed it isn't able to use the nlu extraction and always run in fallback, stucking on 1st question or getting out the form. 

My domain contains :
```
actions:
- my_form
- validate_my_form
forms:
  my_form: null
  other_form:
    slot:(...)
```
The "null" is added by Rasa X.

How do I reach validate_my_form when there are no commun mapping ? Which FormAction must i use to explicit the form's slots and access the validate_form (inside action server, i guess the core's one, but i only achieve register sdk's one).

Do i have to make my class inherit from both FormAction and FormValidationAction ? 

Thank you a lot for reading, let's give a nudge to your karma by helping me ! :D

wochinge:
@SuunnyyGoo Please ask usage questions in [Rasa forum](https//forum.rasa.com)

travelmail26:
Hi, rasa newbie here. Is there an example of what this looks like somewhere? thanks!

melindaloubser1:
@travelmail26  Please ask usage questions in [Rasa forum](https://forum.rasa.com/). 
You can also look at [these docs](https://rasa.com/docs/rasa/forms#custom-slot-mappings)

PR #7179, opened 2020-11-04
Create security.txt

tmbo:
**Proposed changes**:
- added security.txt file
- conforms with https://tools.ietf.org/html/draft-foudil-securitytxt-10#section-3.5.7


m-vdb (approved):
🔒

Issue #6865, opened 2020-10-01
Migration from old response selector format


dakshvar22:
**Description of Problem**:
Since, we changed the [training data format for response selector](https://github.com/RasaHQ/rasa/pull/6591), it would be good to help users smoothly transition to the new format through a cli command.

**Overview of the Solution**:
It can be part of the `rasa data convert` command that now exists.
Migration changes that need to be covered - 

- [ ] Change all response templates that are defined for retrieval intents to have the `utter_` prefix. For e.g.
```
## ask_name
* chitchat/ask_name
  - My name is retrieval bot
```
should get converted to
```
responses:
   utter_chitchat/ask_name:
     - text: My name is retrieval bot
```

- [ ] Change all occurrences of actions with `respond_` prefix to use `utter_` prefix instead. These would be in domain and stories.

**Definition of Done**:
- [ ] Tests are added
- [ ] Feature described the docs
- [ ] Feature mentioned in the changlog

m-vdb:
Apparently we already have a `rasa data convert nlg` command that does the conversion from Markdown to Yaml, so I'll modify the command to add the behaviour.

m-vdb:
@dakshvar22 I am not sure how to implement the updates to the stories and domain files (your second point). My initial thoughts where:
1. use the `rasa data convert nlg` command
2. rename the response names before the yaml file is dumped, then dump the file
3. for every response renaming that was done, update the stories and domain files

1 and 2 are straightforward. For 3, I'm not sure how people would use it. Should they provide a `--domain` and `--stories` argument? Or should we detect where the domain and the stories files reside? For the former, it's straightforward. For the latter, I am not sure how to do it.

cc @degiz do you have any thoughts on this?

dakshvar22:
I think the default can be that the user provides path to the training data. In that path, we try to search for - 
1. NLU files(including responses.yml),
2. Story files
Additionally, we also search for domain in its default path. Based on what all we get, we should be able to modify.
The command can also let the user explicitly specify the paths for domain and stories which would override the default paths.

What do you think?

m-vdb:
The thing is that with this approach, you have to consider the conversion of the whole project: are stories in Markdown or already converted to Yaml? I don't know if we can detect this reasonably well, I'd have to check.

degiz:
> The thing is that with this approach, you have to consider the conversion of the whole project: are stories in Markdown or already converted to Yaml

[...]

PR #7183, opened 2020-11-04
Response converter

m-vdb:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/6865


rasa/cli/arguments/data.py
```diff
@@ -14,7 +14,6 @@ def set_convert_arguments(parser: argparse.ArgumentParser, data_type: Text):
     parser.add_argument(
         "-f",
         "--format",
-        required=True,
```
m-vdb, line 17: otherwise, if you do not specify `--format`, it gives this error:
```
usage: rasa data convert nlg [-h] [-v] [-vv] [--quiet] -f {json,md,yaml}
                             --data DATA --out OUT [-l LANGUAGE]
rasa data convert nlg: error: the following arguments are required: -f/--format
```

m-vdb: there might be a lot of other occurrences of this

rasa/cli/data.py
```diff
@@ -336,6 +343,119 @@ def _convert_nlg_data(args: argparse.Namespace) -> None:
...
+    Args:
+        args: the CLI arguments
+    """
+    domain_file = Path(args.domain)
+    domain = _get_domain(domain_file)
+
+    domain_dict = domain.cleaned_domain()
+    domain_dict["actions"] = [
+        _normalize_response_name(action) for action in domain_dict["actions"]
+    ]
```
m-vdb, line 373: quickest way I found to update actions of a domain, not sure if there is another better way?

m-vdb:
@degiz would you mind giving a first review on this PR? I haven't run nor written tests yet, I want to get some feedback on the general approach (no need to look at docstrings and function names for instance)

rasa/cli/arguments/data.py
```diff
@@ -26,7 +25,6 @@ def set_convert_arguments(parser: argparse.ArgumentParser, data_type: Text):
 
     add_out_param(
         parser,
-        required=True,
```
degiz, line 29: I think it still make sense *not* to convert the files in-place, and rather create converted copies using the `out` parameter.

rasa/cli/arguments/data.py
degiz: I think I'd still keep the `required`! 
We're adding a new sub command `rasa data convert responses`, and it can have it's custom params.

rasa/cli/data.py
```diff
@@ -121,28 +128,28 @@ def _add_data_convert_parsers(
         help="Migrate model configuration between Rasa Open Source versions.",
     )
     migrate_config_parser.set_defaults(func=_migrate_model_config)
-    _add_migrate_config_arguments(migrate_config_parser)
-
-
-def _add_migrate_config_arguments(parser: argparse.ArgumentParser) -> None:
```
degiz, line 127: Not sure here - why do we remove `_add_migrate_config_arguments`?
IMO we can have a new `_add_convert_responses_arguments` instead.

rasa/cli/data.py
```diff
@@ -121,28 +128,28 @@ def _add_data_convert_parsers(
...
-        "-o",
-        "--out",
-        type=str,
+    default_arguments.add_config_param(migrate_config_parser)
+    default_arguments.add_domain_param(migrate_config_parser)
+    default_arguments.add_out_param(
+        migrate_config_parser,
         default=os.path.join(DEFAULT_DATA_PATH, "rules.yml"),
-        help="Path to the file which should contain any rules which are created as "
+        help_text="Path to the file which should contain any rules which are created as "
```
degiz, line 136: 🚀

rasa/cli/data.py
degiz: Sounds ok to me 👍

degiz (commented):
Thanks a lot for working on this @m-vdb !!
One note about the general approach - can we use `TrainingDataConverter` interface? This way we can:
- implement all needed conversions in another module and keep `data.py` from getting too long
- reuse methods like `TrainingDataConverter::generate_path_for_converted_training_data_file`

rasa/cli/arguments/data.py
m-vdb: the reason why I did this change is not because of this, see [this comment](https://github.com/RasaHQ/rasa/pull/7183#discussion_r517883807)

rasa/cli/data.py
m-vdb: initially the method was here (made by @wochinge I think) because it did the legwork of adding custom arguments. I realised that it could leverage the `default_arguments` module, so I made the change in [this commit](https://github.com/RasaHQ/rasa/pull/7183/commits/9ef7bbea43468bda1dea79c7a65b0eb96b9d5c7f). The code is shorter and leverages existing interface. Since it's very short now, I thought that containing it in `_add_migrate_config_arguments()` became useless

rasa/cli/arguments/data.py
m-vdb: but why specify a default argument then? I'm confused

m-vdb: to be even clearer: the command's `--help` says:
> Output format the training data should be converted
                        into. Note: currently training data can be converted
                        to 'yaml' format only from 'md' format (**default: yaml**)

So I think we should either:
1. remove `required=True` and keep the default behaviour
2. remove "default: yaml" from the help message because it's misleading

I went with 1. because the UX seemed better to me

rasa/nlu/training_data/converters/nlg_markdown_to_yaml_converter.py
```diff
@@ -35,7 +38,29 @@ async def convert_and_write(cls, source_path: Path, output_path: Path) -> None:
...
+        writer.dump(output_nlg_path, converted_training_data)
 
         print_success(f"Converted NLG file: '{source_path}' >> '{output_nlg_path}'.")
+
+    @staticmethod
+    def _normalize_response_name(response_name: Text) -> Text:
+        """During the converstion, ensure responses start with `utter_`
+
+        Args:
+            response_name: the name of the response, e.g. "chitchat/ask_name"
```
degiz, line 58:
```suggestion
            response_name: The name of the response, e.g. "chitchat/ask_name".
```

rasa/nlu/training_data/converters/nlg_markdown_to_yaml_converter.py
```diff
@@ -35,7 +38,29 @@ async def convert_and_write(cls, source_path: Path, output_path: Path) -> None:
...
+
+        Returns:
+            A normalised response name starting with "utter_"
```
degiz, line 61:
```suggestion
            A normalised response name starting with "utter_".
```

tests/cli/test_rasa_data.py
```diff
@@ -340,6 +340,96 @@ def test_rasa_data_convert_nlg_to_yaml(
...
+
+    expected_domain = (
+        "session_config:\n"
+        "  session_expiration_time: 60\n"
+        "  carry_over_slots_to_new_session: true\n"
+        "actions:\n"
+        "- utter_chitchat\n"
+        "- utter_greet\n"
+        "- utter_cheer_up\n"
+        "version: '2.0'\n"
```
degiz, line 370: Why is the `version` key on the last line? 🤔

rasa/cli/data.py
```diff
@@ -336,6 +342,50 @@ def _convert_nlg_data(args: argparse.Namespace) -> None:
...
+        )
+        telemetry.track_data_convert(args.format, "responses")
+    else:
+        rasa.shared.utils.cli.print_error_and_exit(
+            "Could not recognize output format. Supported output formats: "
+            "'yaml'. Specify the desired output format with '--format'."
+        )
+
+
+def _migrate_responses_in_domain(args: argparse.Namespace):
```
degiz, line 363: Perfectly it would be awesome to convert the domain using `TrainingDataConverter` interface as well 😬 
In order to do that we need to pass `args.out` and `args.data` (`args.domain` for domain) directly to the `_convert_to_yaml()`, and the rest will be taken care of!

Do you still have capacity to do that?

degiz (commented):
Thanks a lot for jumping in and implementing the changes!! 🚀 💯 

One note - with this PR `rasa data convert nlg` will leave project in a broken state (domain and stories still not converted), could we extent the help text for this command and mention, that they should also run `rasa data convert responses`?

rasa/cli/data.py
```diff
@@ -336,6 +342,50 @@ def _convert_nlg_data(args: argparse.Namespace) -> None:
...
+    """Migrate retrieval intent responses to the new 2.0 format.
+
+    Before 2.0, retrieval intent responses needed to start
+    with `respond_`. Now, they need to start with `utter_`.
+
+    This function updates the domain file in place.
```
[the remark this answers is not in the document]
m-vdb: yeah it doesn't anymore, removing

rasa/cli/data.py
```diff
@@ -336,6 +342,50 @@ def _convert_nlg_data(args: argparse.Namespace) -> None:
...
+
```
m-vdb, line 370:
```suggestion
```

rasa/core/training/converters/story_responses_prefix_converter.py
```diff
@@ -0,0 +1,66 @@
...
+    def filter(cls, source_path: Path) -> bool:
+        """Only consider YAML story files.
+
+        Args:
+            source_path: Path to the training data file.
+
+        Returns:
+            `True` if the given file can is a YAML stories file, `False` otherwise
+        """
+        return YAMLStoryReader.is_stories_file(source_path)
```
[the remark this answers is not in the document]
m-vdb: I'd change the signature of `is_stories_file()` to accept both

tests/cli/test_rasa_data.py
m-vdb: because it's added last to the dictionary before it's returned... Happy to change it in another PR though, I fear it could break other tests. WDYT?

rasa/cli/data.py
```diff
@@ -121,28 +127,28 @@ def _add_data_convert_parsers(
...
+        "responses",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        parents=parents,
+        help=(
+            "Convert retrieval intent responses between Rasa Open Source versions. "
+            "Please also run `rasa data convert nlg` to convert responses to the right format."
+        ),
+    )
+    convert_responses_parser.set_defaults(func=_migrate_responses)
+    arguments.set_convert_arguments(convert_responses_parser, data_type="Rasa NLU")
```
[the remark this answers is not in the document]
m-vdb: This argument is used for `add_data_param()`, and in the case of this command, it's mainly stories (because the domain is handled by another argument). I'll change to `stories`

rasa/cli/data.py
m-vdb: done in e908f5a2868 ✅ One question: is it ok to keep the name `TrainingData` in `TrainingDataConverter` now that it can also take care of the domain? I think it's fine, just wanted to double check

tests/cli/test_rasa_data.py
m-vdb: done in ef7686a021a, let's see if some tests fail ⏳

docs/docs/migration-guide.mdx
```diff
@@ -403,52 +403,14 @@ Response Selectors are a stable feature as of version 2.0.
 
 The [conversion command](./migration-guide.mdx#training-data-files) will automatically
 convert your `responses.md` file, stories and nlu training data to the new yaml format.
-Additionally you will need to rename the `respond_` actions in your stories files to use the
-`utter_` prefix instead. For example:
+It will also take care of adding the `utter_` prefix to your responses.
+Additionally, `respond_` actions in your domain file and stories files will need to be
+renamed with the `utter_` prefix instead. You can run the following command to do so:
 
-```yaml-rasa
```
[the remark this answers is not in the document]
m-vdb: 🤔 that's a good point. I'll keep them and rephrase

tests/cli/test_rasa_data.py
m-vdb: should I add a changelog fragment for this?

rasa/cli/data.py
degiz: I think it's ok! The only alternative I can think about is `ProjectData`, but that doesn't look any better 😬

rasa/shared/core/domain.py
```diff
@@ -1165,14 +1165,25 @@ def persist_clean(self, filename: Union[Text, Path]) -> None:
...
+
+        Args:
+            clean_before_dump: When set to `True`, this method returns
+                               a version of the domain without internal
+                               information. Defaults to `False`.
+        Returns:
+            A string in YAML format representing the domain.
+        """
+        # setting the `version` key first so that it appears at the top of YAML files
+        # thanks to the `should_preserve_key_order` argument of `dump_obj_as_yaml_to_string`
```
degiz, line 1179:
```suggestion
        # thanks to the `should_preserve_key_order` argument 
        # of `dump_obj_as_yaml_to_string`
```

rasa/shared/core/domain.py
```diff
@@ -1165,14 +1165,25 @@ def persist_clean(self, filename: Union[Text, Path]) -> None:
...
+        domain_data: Dict[Text, Any] = {
```
degiz, line 1180: I'm quite surprised this even works 🤔 it's still a Dict, and the keys order is not preserved 😬 
If tests fails, we could change the `dict` to the `OrderedDict` right away!

degiz (approved):
Thanks a lot! 🚀🚀🚀

Let's keep an eye on that domain dict key order 😬

rasa/shared/core/domain.py
m-vdb: because they changed it in Python 3.7 \0/ => https://softwaremaniacs.org/blog/2020/02/05/dicts-ordered/

m-vdb:
thanks for the rapid reviews @degiz 🙏 🙏 🙏

PR #7195, opened 2020-11-05
Accept multiple domain files in Rasa X local mode

zammitjames:
Proposed changes:
- Add `--domain` argument for `rasa x command`

**Do not merge** before https://github.com/RasaHQ/rasa-x/pull/3967 is merged!


sara-tagger:

rgstephens:
@tabergma could you look at this?

rasa/cli/x.py
```diff
@@ -445,13 +456,19 @@ def run_locally(args: argparse.Namespace):
...
     # noinspection PyBroadException
     try:
         local.main(
-            args, project_path, args.data, token=rasa_x_token, config_path=config_path
+            args,
+            project_path,
+            args.data,
+            token=rasa_x_token,
+            config_path=config_path,
+            domain_path=domain_path,
```
alwx, line 471: Unfortunately, `domain_path` argument is not yet supported by Rasa X.

rasa/cli/arguments/x.py
```diff
@@ -12,6 +12,7 @@ def set_x_arguments(parser: argparse.ArgumentParser):
         parser, default=DEFAULT_DATA_PATH, data_type="stories and Rasa NLU "
     )
     default_arguments.add_config_param(parser)
+    default_arguments.add_domain_param(parser)
```
alwx, line 15: The description of domain parameter says "This can be a single YAML file, or a directory that contains several files with domain specifications in it". I am not sure if it's correct because it seems like the rest of the code expects `domain` to be a file, not a directory.

alwx (changes requested):
Thanks for your PR, @zammitjames! 

Unfortunately, this PR requires changes to Rasa X, and these changes cannot be done by community members. We will take a closer look and make sure to update Rasa X accordingly.

We will also keep you updated here.

rasa/cli/x.py
degiz: done in RasaHQ/rasa-x#3967

rasa/cli/arguments/x.py
degiz: Could you please point me to an example? we've changed the description of the "--domain" as of `2.0`, and I was supposed to update to docs as well 😬

alwx: @degiz here it is: https://github.com/RasaHQ/rasa/blob/master/rasa/cli/arguments/default_arguments.py#L60

degiz: @alwx right, this doc is something I'm aware of, I meant this:
```
the rest of the code expects domain to be a file, not a directory.
```
I think the rest of the code works correctly with the situation when the domain is a folder.

tmbo:
@degiz does this work properly with the split domain implementation in rasa x?

degiz:
@tmbo yes, I was working on both of them in parallel. Could you review it?

rasa/cli/arguments/x.py
alwx: Yes, you seem to be right. Ok, I think we can merge this one then.

alwx (approved):

PR #7204, opened 2020-11-08
Require slack signing secret

tmbo:
**Proposed changes**:
- Require slack signing secret
- renamed and moved `InvalidConfigurationError`


rasa/nlu/components.py
```diff
@@ -7,7 +7,8 @@
 from rasa.exceptions import MissingDependencyException
 from rasa.shared.exceptions import RasaException
 from rasa.shared.nlu.constants import TRAINABLE_EXTRACTORS
-from rasa.nlu.config import RasaNLUModelConfig, override_defaults, InvalidConfigError
+from rasa.nlu.config import RasaNLUModelConfig, override_defaults
```
wochinge, line 10: optional: Don't use the function directly but reference it using the module.

rasa/nlu/components.py
```diff
@@ -58,7 +59,7 @@ def validate_requirements(component_names: List[Optional[Text]]) -> None:
     failed_imports = {}
     for component_name in component_names:
         if component_name is None:
-            raise InvalidConfigError(
+            raise InvalidConfigException(
```
wochinge, line 62: can you please adapt the docstring for this function? The `Raises` mentions `InvalidConfigError`

rasa/shared/exceptions.py
```diff
@@ -69,3 +69,7 @@ class FileNotFoundException(RasaException, FileNotFoundError):
 
 class FileIOException(RasaException):
     """Raised if there is an error while doing file IO."""
+
+
+class InvalidConfigException(ValueError, RasaException):
```
wochinge, line 74: for my understanding: Why did we change the suffix from `Error` to `Exception`? Configuration errors are usually something we can't recover from - so `Error` seems more suited in my opinion (following [this](https://www.geeksforgeeks.org/errors-v-s-exceptions-in-java/) or [this](https://www.datacamp.com/community/tutorials/exception-handling-python) guide.

wochinge (approved):
🚀

rasa/shared/exceptions.py
tmbo: I agree with you, it should be an `Error`. I was thinking about this as well and thought that we have quite a number of `Exceptions` that should rather be `Errors` by that definition (e.g. the above yaml errors) that it might be more consistent now to just always use `Exception`. I am also not sure if this is always a useful distinction. E.g. I certainly agree that OOM is something different than a model not found exception, but things start to break when you look at the `ValueError` which in some cases should be treated as an Exception and in some places more like an Error. It assumes that the throwing function knows what the caller is able to handle and recover from, which I am not sure is the case (e.g. why shouldn't there be a function that is able to recover from an `InvalidConfigException` by "repairing" a common error).

Long story short rather than using a differentiation where I am not sure if it provides any value I aimed to keep the naming consistent between the exceptions we raise. But honestly, this is a tossup and happy to change it to `Error`.

wochinge: > It assumes that the throwing function knows what the caller is able to handle and recover from, which I am not sure is the case (e.g. why shouldn't there be a function that is able to recover from an InvalidConfigException by "repairing" a common error)

I agree 👍 

> Long story short rather than using a differentiation where I am not sure if it provides any value I aimed to keep the naming consistent between the exceptions we raise

Makes sense 👍 Just wanted to know the reason since we recently named a couple of exceptions / errors in Rasa X `XyError` .

tmbo:
@tczekajlo is there a way to tell the security check that these secrets are ok and not an issue?

virtualroot:
@tmbo You can configure [gitleaks](https://github.com/zricethezav/gitleaks/wiki/Configuration) to ignore the file if you add it into the allowlist section https://github.com/RasaHQ/rasa/blob/master/.gitleaks.toml#L107

tczekajlo:
> @tczekajlo is there a way to tell the security check that these secrets are ok and not an issue?

Yes, there is the `.gitleaks.toml` file with the configuration for the gitleaks. I've added extra configuration that should help with your issue and similar in the future.

tmbo:
perfect, thanks a lot :rocket

Issue #7122, opened 2020-10-28
add slot mappings validator 


Ghostvv:
some slot mappings require certain info to be set, like `from_entity` requires `entity`. We need to add validation of the slot mappings that users create in domain, to prevent hidden errors like: https://forum.rasa.com/t/got-mapping-but-faild-to-extract-slot-name-in-rasa-formaction-anyone-please-help/36310

PR #7210, opened 2020-11-09
add validation for slot mappings

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7122
- move `SlotMapping` to `Domain` as we require it in the `rasa.shared` package
- implement the validations as Python code as `pykwalify`
  - doesn't allow different types for dictionary keys (`forms` can be a `list` or `dict`) (see https://github.com/Grokzen/pykwalify/issues/112)
  - we need different validations based on the `type` field of the slot mapping

### DIFF
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -171,7 +171,7 @@ class Domain:
         intents = data.get(KEY_INTENTS, {})
         forms = data.get(KEY_FORMS, {})
 
-        validate_slot_mappings(forms)
+        _validate_slot_mappings(forms)
 
         return cls(
             intents,
@@ -1473,10 +1473,18 @@ class SlotMapping(Enum):
     FROM_TEXT = 3
 
     def __str__(self) -> Text:
+        """Returns a string representation of the object."""
         return self.name.lower()
 
     @staticmethod
     def validate(mapping: Any, form_name: Text, slot_name: Text) -> None:
+        """Validates a slot mapping.
+
+        Args:
+            mapping: The mapping which is validated.
+            form_name: The name of the form which uses this slot mapping.
+            slot_name: The name of the slot which is mapped by this mapping.
+        """
         if not isinstance(mapping, dict):
             raise InvalidDomain(
                 f"Please make sure that the slot mappings for slot '{slot_name}' in "
@@ -1508,7 +1516,7 @@ class SlotMapping(Enum):
                 )
 
 
-def validate_slot_mappings(forms: Union[Dict, List]) -> None:
+def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
     if isinstance(forms, list):
         if not all(isinstance(form_name, str) for form_name in forms):
             raise InvalidDomain("Not all form names are strings.")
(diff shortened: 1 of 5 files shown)


rasa/shared/core/domain.py
```diff
@@ -1446,3 +1450,85 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
...
+            str(SlotMapping.FROM_TRIGGER_INTENT): ["value"],
+            str(SlotMapping.FROM_TEXT): [],
+        }
+
+        mapping_type = mapping.get("type")
+        required_keys = validations.get(mapping_type)
+
+        if required_keys is None:
+            raise InvalidDomain(
+                f"Your form '{form_name}' uses an invalid slot mapping of type "
```
wochinge, line 1494: add docs link everywhere

rasa/shared/core/domain.py
```diff
@@ -1446,3 +1450,94 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
...
+    FROM_INTENT = 1
+    FROM_TRIGGER_INTENT = 2
+    FROM_TEXT = 3
+
+    def __str__(self) -> Text:
+        """Returns a string representation of the object."""
+        return self.name.lower()
+
+    @staticmethod
+    def validate(mapping: Any, form_name: Text, slot_name: Text) -> None:
```
joejuzl, line 1468: Should the type of `mapping` be `Dict`?

rasa/shared/core/domain.py
```diff
@@ -1446,3 +1450,94 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
...
+                    f"for slot '{slot_name}' in the form '{form_name}'. Please see "
+                    f"{rasa.shared.constants.DOCS_URL_FORMS} for more information."
+                )
+
+
+def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
+    if isinstance(forms, list):
+        if not all(isinstance(form_name, str) for form_name in forms):
+            raise InvalidDomain(
+                f"If use the deprecated list syntax for forms, "
```
joejuzl, line 1514:
```suggestion
                f"If you use the deprecated list syntax for forms, "
```

rasa/shared/core/domain.py
```diff
@@ -1446,3 +1450,94 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
...
+                f"{rasa.shared.constants.DOCS_URL_FORMS} for more information."
+            )
+
+        return
+
+    if not isinstance(forms, dict):
+        raise InvalidDomain("Forms have to be specified as dictionary.")
+
+    for form_name, slots in forms.items():
+        if not isinstance(slots, Dict) and slots is not None:
```
joejuzl, line 1525: Could first have something like: 
```
if not slots:
    continue
```
Then you wouldn't need `and slots is not None` in this check or the reassignment `slots = slots or {}` below.

joejuzl (changes requested):

rasa/shared/core/domain.py
wochinge: We don't know at this point. That's why I used `Any`.

joejuzl: But if we throw an error when it is not `Dict` then shouldn't we have the type here be `Dict` so potential errors can be caught using static analysis?

joejuzl (approved):

PR #7217, opened 2020-11-10
Pika multiprocessing fix

alwx:
**Proposed changes**:
- Prevent the message handling process in `PikaEventBroker` from being terminated.

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #7219, opened 2020-11-10
Pika multiprocessing fix for 1.10.x

alwx:
**Proposed changes**:
- based on what's done in this PR: https://github.com/RasaHQ/rasa/pull/7217

What's essential here is that now we prevent the message handling process from being terminated.

### DIFF
--- a/rasa/core/brokers/pika.py
+++ b/rasa/core/brokers/pika.py
@@ -307,6 +307,7 @@ class PikaMessageProcessor:
         self,
         parameters: "Parameters",
         queues: Union[List[Text], Tuple[Text], Text, None],
+        **kwargs: Any,
     ) -> None:
         """Initialise Pika connector.
 
@@ -315,7 +316,7 @@ class PikaMessageProcessor:
             queues: Pika queues to declare and publish to
         """
         self.parameters: "Parameters" = parameters
-        self.queues: List[Text] = self._get_queues_from_args(queues)
+        self.queues: List[Text] = self._get_queues_from_args(queues, kwargs)
 
         self._connection: Optional["SelectConnection"] = None
         self._channel: Optional["Channel"] = None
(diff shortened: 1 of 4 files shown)


rasa/core/brokers/pika.py
```diff
@@ -498,16 +527,25 @@ def process_messages(self) -> None:
...
         starting the IOLoop to block and allow the SelectConnection to operate.
+
         This function is blocking and indefinite thus it
         should be started in a separate process.
         """
-        self._connection = self._connect()
-
-        # noinspection PyUnresolvedReferences
-        self._connection.ioloop.start()
+        # TODO: Fix logging
```
wochinge, line 545: logging works for you right? Then let's dump this todo
```suggestion
```

rasa/core/brokers/pika.py
```diff
@@ -602,53 +641,24 @@ def _connect(self) -> None:
...
         )
-
-        self.process = self._start_pika_process()
+        self.process = self._start_pika_connection_process()
 
     def _get_mp_context(self) -> multiprocessing.context.BaseContext:
         return multiprocessing.get_context(self.MP_CONTEXT)
 
-    def _start_pika_process(self) -> Optional[multiprocessing.Process]:
+    def _start_pika_connection_process(self) -> Optional[multiprocessing.Process]:
```
wochinge, line 652: optional: What does the additional `connection` mean?

wochinge (commented):
code looks good. Thanks for looking into this 🚀  Will give it an additional manual run and then approve 👍

wochinge:
a bunch of tests fail btw

rasa/core/brokers/pika.py
```diff
@@ -602,53 +641,24 @@ def _connect(self) -> None:
...
-    def _publish(self, body: Text, headers: MessageHeaders = None) -> None:
-        if not self.pika_message_processor:
-            self._connect()
-
-        if (
-            self.process and self.process.is_alive()
-        ) or self.should_keep_unpublished_messages:
-            self.process_queue.put((body, headers))
-
-    def is_ready(
```
wochinge, line 636: we can return if the process is alive here. It's used by the `exporter` module afaik

wochinge (approved):
Logging works (why didn't it for 2.0? 🤔 ) and connection seems stable. Would be great if @melindaloubser1 could confirm though.

melindaloubser1:
Yes, just tested it on microk8s, all seems good

melindaloubser1 (approved):

benos:
@wochinge @alwx 

If I am reading the code correctly, this commit has silently removed the deprecated 'queue' argument:

https://github.com/RasaHQ/rasa/pull/7219/commits/253afa0f8f9cf04d78647d20b2ee10b1a3b5ce96

I imagine it was unintentional to introduce a breaking change on a patch level release, but you might want to update the changelog for 1.10.17 so that those of us still using the old argument know about it.

Also, the warning still refers to the 'queue' argument, adding to the confusion: 

UserWarning: No `queues` or `queue` argument provided. It is suggested to explicitly specify a queue as described in https://rasa.com/docs/rasa/api/event-brokers/#pika-event-broker. Using the default queue 'rasa_core_events' for now.

wochinge:
Thanks for pointing this out @benos 
@alwx Could you please take care of it? In my opinion it would be clearer to re-add support for the old argument, and cut another micro instead of having such a major change in the changelog / code, what do you think?

Issue #7211, opened 2020-11-09
Update docs for `FormValidationAction`


wochinge:
Update the docs for `FormValidationAction` as followup of https://github.com/RasaHQ/rasa-sdk/pull/329

PR #7233, opened 2020-11-10
update documentation for custom slot mappings

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7211
- recommend to use `FormValidationAction` everywhere.
- update how custom slots are extracted
- describe how to deal with dynamic slot mappings

### DIFF
--- a/docs/docs/forms.mdx
+++ b/docs/docs/forms.mdx
@@ -428,11 +428,11 @@ This is, for example, useful when you need further details based on how
 a previous slot was filled or you want to change the order in which slots are requested.
 
 If you are using the Rasa SDK, we recommend you to use the `FormValidationAction` and
-override `required_slots` to fit your dynamic behavior. Please remember to implement
+override `required_slots` to fit your dynamic behavior. You should implement
 a method `extract_<slot name>` for every slot which doesn't use a predefined mapping,
 as described in [Custom Slot Mappings](forms.mdx#custom-slot-mappings).
 The example below will ask the user if they want to sit in
-the shade or in the sun in case they told us that they want to sit outside.
+the shade or in the sun in case they said they want to sit outside.
 
 ```python
 from typing import Text, List, Optional
--- a/docs/docs/forms.mdx
+++ b/docs/docs/forms.mdx
@@ -367,12 +367,12 @@ If you're using the Rasa SDK we recommend you to extend the provided
 `FormValidationAction`. When using the `FormValidationAction`, two things are needed to
 extract customs slots:
 
-1. Define a method `extract_<slot_name>` for every custom slot which you want to
+1. Define a method `extract_<slot_name>` for every custom slot mapping which you want to
    extract.
-2. Optional: If a custom slot is not part of your mapped slots in the domain,
-   override `required_slots` to include the additional custom slot.
+2. Optional: If a slot is not part of your mapped slots in the domain,
+   override `required_slots` to make sure the additional slot is considered by the form.
 
-The following example shows the implementation of a form which extracts a custom slot
+The following example shows the implementation of a form which extracts a slot
 `outdoor_seating` in addition to the slots which were mapped in the domain.
 The method `extract_outdoor_seating` sets the slot `outdoor_seating` based on whether
 the keyword `outdoor` was present in the last user message.
@@ -426,7 +426,9 @@ This is, for example, useful when you need further details based on how
 a previous slot was filled or you want to change the order in which slots are requested.
 
 If you are using the Rasa SDK, we recommend you to use the `FormValidationAction` and
-override `required_slots` to fit your dynamic behavior.
+override `required_slots` to fit your dynamic behavior. Please remember to implement
+a method `extract_<slot name>` for every slot which isn't defined in the slot mappings
+of your domain as it is described [here](forms.mdx#custom-slot-mappings).
 The example below will ask the user if they want to sit in
 the shade or in the sun in case they told us that they want to sit outside.
 
--- a/docs/docs/forms.mdx
+++ b/docs/docs/forms.mdx
@@ -364,16 +364,19 @@ case, you can use the
 extraction code. Rasa Open Source will trigger this action when the form is run.
 
 If you're using the Rasa SDK we recommend you to extend the provided
-`FormValidationAction`. When using the `FormValidationAction`, two things are needed to
-extract customs slots:
+`FormValidationAction`. When using the `FormValidationAction`, three steps are required
+to extract customs slots:
 
-1. Define a method `extract_<slot_name>` for every custom slot mapping which you want to
-   extract.
-2. Optional: If a slot is not part of your mapped slots in the domain,
-   override `required_slots` to make sure the additional slot is considered by the form.
+1. Define a method `extract_<slot_name>` for every slot that should be mapped in a
+   custom way.
+2. Make sure that in the domain file you list for your form only those slots that don't
+   use custom mappings.
+3. Override `required_slots` to add all slots with custom mappings to the list of slots
+   the form should request.
 
 The following example shows the implementation of a form which extracts a slot
-`outdoor_seating` in addition to the slots which were mapped in the domain.
+`outdoor_seating` in a custom way, in addition to the slots whose mappings are defined
+in the domain.
 The method `extract_outdoor_seating` sets the slot `outdoor_seating` based on whether
 the keyword `outdoor` was present in the last user message.
 
--- a/docs/docs/forms.mdx
+++ b/docs/docs/forms.mdx
@@ -367,14 +367,15 @@ If you're using the Rasa SDK we recommend you to extend the provided
 `FormValidationAction`. When using the `FormValidationAction`, two things are needed to
 extract customs slots:
 
-1. Override `required_slots`. It should return a list of slot names which your form
-   requires to be filled.
-2. Define a method `extract_<slot_name>` for every custom slot which you want to
+1. Define a method `extract_<slot_name>` for every custom slot which you want to
    extract.
+2. Optional: If a custom slot is not part of your mapped slots in the domain,
+   override `required_slots` to include the additional custom slot.
 
-The following example shows the implementation of a custom slot mapping which sets
-the slot `outdoor_seating` based on whether the keyword `outdoor` was present in the
-last user message.
+The following example shows the implementation of a form which extracts a custom slot
+`outdoor_seating` in addition to the slots which were mapped in the domain.
+The method `extract_outdoor_seating` sets the slot `outdoor_seating` based on whether
+the keyword `outdoor` was present in the last user message.
 
 ```python
 from typing import Dict, Text, List, Optional, Any
@@ -390,11 +391,13 @@ class ValidateRestaurantForm(FormValidationAction):
 
     async def required_slots(
         self,
+        slots_mapped_in_domain: List[Text],
         dispatcher: "CollectingDispatcher",
         tracker: "Tracker",
         domain: "DomainDict",
     ) -> Optional[List[Text]]:
-        return ["outdoor_seating"]
+        required_slots = slots_mapped_in_domain + ["outdoor_seating"]
+        return required_slots
 
     async def extract_outdoor_seating(
         self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: Dict
@@ -409,68 +412,18 @@ By default the `FormValidationAction`
 will automatically set the [`requested_slot`](forms.mdx#the-requested_slot-slot) to the
 first slot specified in `required_slots` which is not filled.
 
-### Changing the Next Slot to Request
+### Dynamic Form Behavior
 
 By default Rasa Open Source will ask for the next
 [`requested_slot`](forms.mdx#the-requested_slot-slot) defined in the slot mappings
-in our domain, which is not filled. If you use
+in the domain, which is not filled yet. If you use
 [custom slot mappings](forms.mdx#custom-slot-mappings) and the `FormValidationAction`,
-it will ask for the first empty slot returned by your `required_slots` method. If all
-slots in `required_slots` are filled, it will check if all domain slot mappings are
-filled as well and then notify the form that it can be deactivated.
-
-You can change the way the next slot is requested by overriding `next_requested_slot`.
-Return `SlotSet(requested_slot", "my_slot")` to ask the user to fill slot `my_slot`
-next. Return `SlotSet("requested_slot": None)` to notify Rasa Open Source that all
-required slots are filled and that the form should be deactivated.
-
-```python
-from typing import Dict, Text, List, Optional, Any
-
-from rasa_sdk import Tracker
-from rasa_sdk.executor import CollectingDispatcher
-from rasa_sdk.forms import FormValidationAction
-from rasa_sdk.events import EventType, SlotSet
-
-class ValidateRestaurantForm(FormValidationAction):
-    def name(self) -> Text:
-        return "validate_restaurant_form"
-
-    async def required_slots(
-        self,
-        dispatcher: "CollectingDispatcher",
-        tracker: "Tracker",
-        domain: "DomainDict",
-    ) -> Optional[List[Text]]:
-        return ["outdoor_seating"]
-
-    async def extract_outdoor_seating(
-        self, dispatcher: CollectingDispatcher, tracker: Tracker, domain: Dict
-    ) -> Dict[Text, Any]:
-        text_of_last_user_message = tracker.latest_message.get("text")
-        sit_outside = "outdoor" in text_of_last_user_message
-
-        return {"outdoor_seating": sit_outside}
-
-    async def next_requested_slot(
-        self,
-        dispatcher: "CollectingDispatcher",
-        tracker: "Tracker",
-        domain: "DomainDict",
-    ) -> Optional[EventType]:
-        for slot_name in await self.required_slots(dispatcher, tracker, domain):
-            if not tracker.slots.get(slot_name):
-                return SlotSet("requested_slot", slot_name)
-
-        # all required slots are filled
-        return SlotSet("requested_slot", None)
-```
-
-### Dynamic Slot Mappings
+it will ask for the first empty slot returned by the `required_slots` method. If all
+slots in `required_slots` are filled the form will be be deactivated.
 
 If needed, you can update the required slots of your form dynamically.
-This is for example useful when you need further details based on how
-a previous slot was filled.
+This is, for example, useful when you need further details based on how
+a previous slot was filled or you want to change the order in which slots are requested.
 
 If you are using the Rasa SDK, we recommend you to use the `FormValidationAction` and
 override `required_slots` to fit your dynamic behavior.
@@ -488,23 +441,19 @@ class ValidateRestaurantForm(FormValidationAction):
 
     async def required_slots(
         self,
+        slots_mapped_in_domain: List[Text],
         dispatcher: "CollectingDispatcher",
         tracker: "Tracker",
         domain: "DomainDict",
     ) -> Optional[List[Text]]:
-        required_slots = ["outdoor_seating"]
+        additional_slots = ["outdoor_seating"]
         if tracker.slots.get("outdoor_seating") is True:
             # If the user wants to sit outside, ask
             # if they want to sit in the shade or in the sun.
-            required_slots.append("shade_or_sun")
+            additional_slots.append("shade_or_sun")
 
-        return required_slots
+        return additional_slots + slots_mapped_in_domain
 ```
-
-`required_slots` is called with the latest slot values during the execution of
-[`next_requested_slot`](forms.mdx#changing-the-next-slot-to-request) which means that
-the `FormValidation` action will take the latest state in account when deciding which
-slot to request next.
 
 ### The requested_slot slot
 
--- a/docs/docs/forms.mdx
+++ b/docs/docs/forms.mdx
@@ -361,74 +361,150 @@ to fully customize the validation process.
 If none of the predefined [Slot Mappings](./forms.mdx#slot-mappings) fit your use
 case, you can use the
 [Custom Action](./actions.mdx#custom-actions) `validate_{form_name}` to write your own
-extraction code. Rasa Open Source will trigger this function when the form is run.
+extraction code. Rasa Open Source will trigger this action when the form is run.
 
-Make sure your custom action returns `SlotSet` events for every extracted value.
+If you're using the Rasa SDK we recommend you to extend the provided
+`FormValidationAction`. When using the `FormValidationAction`, two things are needed to
+extract customs slots:
+
+1. Override `required_slots`. It should return a list of slot names which your form
+   requires to be filled.
+2. Define a method `extract_<slot_name>` for every custom slot which you want to
+   extract.
+
 The following example shows the implementation of a custom slot mapping which sets
-a slot based on the length of the last user message.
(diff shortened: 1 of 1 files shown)


wochinge:
@JulianGerhard21 Happy to hear your feedback on this as well!

docs/docs/forms.mdx
```diff
@@ -361,75 +361,151 @@ to fully customize the validation process.
...
+extraction code. Rasa Open Source will trigger this action when the form is run.
+
+If you're using the Rasa SDK we recommend you to extend the provided
+`FormValidationAction`. When using the `FormValidationAction`, two things are needed to
+extract customs slots:
+
+1. Override `required_slots`. It should return a list of slot names which your form
+   requires to be filled.
+2. Define a method `extract_<slot_name>` for every custom slot which you want to
+   extract.
```
kearnsw, line 374: I'm not sure that these two are dependent. Isn't it possible to dynamically generate the required_slot list without needing to define any `extract_<slot_name>` methods or has that changed in Rasa 2.0? It may help to make a clearer distinction between these two use cases, so users don't think they need to define these for every slot.

kearnsw (commented):
@wochinge, I provided some suggestions for clarity. HTH

Otherwise it LGTM, thanks for adding this!

samsucik:
Let's define 3 use cases these changes should facilitate (there may be more). For each, I give the steps to do in a custom validation action (subclassing `FormValidationAction`) that I thought would be necessary & sufficient to solve the use case.

## Use case 1: change mapping for an "existing" slot 
_"Custom Slot Mappings" in docs_

Conditions:
- The slot is listed for this form in the domain
- Want to change only how the slot is extracted

Solution:
- Implement `extract_{slot}` for this slot

## Use case 2: add slot to be requested
Conditions:
- Some list of slots listed for this form in the domain
- Want to change this list by only 1) adding a slot not listed for this form in the domain

Solution:
- Override `required_slots` where I insert my new slot into the existing list of slots coming from the domain (I may want to append, prepend, insert the new slot…)
- Implement `extract_{slot}` only for my new slot

## Use case 3: change set of requested slots
_"Dynamic Slot Mappings" in docs_

Conditions:
- Some list of slots listed for this form in the domain
- Want to change this list by both 1) adding some more slots, 2) not requiring some of the original ones

Solution:
- Override `required_slots` where I insert my new slot into the existing list of slots coming from the domain and remove from the list the slots I don’t want to request
- Implement `extract_{slot}` only for my new slot

+++++++++++++++++++++++++++++++++++++++++++++++++++++++

Now my comments -- mostly about where the solution I imagine doesn't align with the current implementation or with the proposed docs changes:
1. For UC1 I shouldn't need to override `required_slots` because, as @kearnsw pointed out, changing mappings is a different use case than changing which slots the form requires.
2. UC3 could be renamed from "Dynamic Slot Mappings" to "Dynamic Requested Slots" (proposed by @akelad) because it's not (necessarily) about mappings.
3. Also re the solution to UC3, @wochinge explained: 
    > "if you defined slot mappings in the domain then Rasa Open Source will consider them by default. In that case you'd need to override `missing_slots` or `next_requested_slot`. TBH, it's best to not define any slot mappings in the domain if you advanced stuff like this in your FormValidationAction"

    I think the name `required_slots` doesn't reflect the true purpose of this function. As a user, I'd expect that from there I can change which slots the form requires, but currently I can only _add_ some more slots that the form will require. Could we perhaps change how `missing_slots` and `required_slots` are implemented such that inside `required_slots` the default implementation will grab all slots for this form from the domain, and in `missing_slots` the default will merely check for slots that aren't filled yet?

samsucik:
Prompted by a question from @wochinge I realised that for UC1 we might not be able to figure out which `extract_{slot}` method to use if there some mapping defined for a slot in the domain but the user also implements custom mapping, i.e. `extract_{slot}` in the form validation action. The issue as I understand it is that we first extract slots in `FormAction` (following the mappings from the domain) and only then call the validation action. If `FormAction` could "see" that there's a custom `extract_{slot}` defined, it wouldn't try to extract the given slot at all, but it can't see this. @wochinge is my understanding right?

wochinge:
> For UC1 I shouldn't need to override required_slots because, as @kearnsw pointed out, changing mappings is a different use case than changing which slots the form requires.

How would we know which `extract_` methods we should call then? We can get this by inspecting the class, but I think being explicit is more helpful then having a lot of logic going on behind the curtains.

> UC3 could be renamed from "Dynamic Slot Mappings" to "Dynamic Requested Slots" (proposed by @akelad) because it's not (necessarily) about mappings.

Sure 👍 

> Could we perhaps change how missing_slots and required_slots are implemented such that inside required_slots the default implementation will grab all slots for this form from the domain, and in missing_slots the default will merely check for slots that aren't filled yet?

mhm, that would mean that the `FormValidationAction` would always set the `requested_slot` even if the action only validates slots. I'd like to avoid doing this and rather set the `requested_slot` in the Rasa Open Source code then 🤔 We could add a check that the `FormValidation` action only sets the `requested_slot` in case the output of `required_slots(...)` is different than the slots defined in the domain though 👍

wochinge:
> I realised that for UC1 we might not be able to figure out which extract_{slot} method to use if there some mapping defined for a slot in the domain but the user also implements custom mapping, i.e. extract_{slot} in the form validation action.

I think if an `extract_` method is defined **and** a slot mapping in the domain, we should print a warning. Currently the `extract_` method would override any extractions based on mappings in the domain for this slot.

> If FormAction could "see" that there's a custom extract_{slot} defined, it wouldn't try to extract the given slot at all, but it can't see this. @wochinge is my understanding right?

I don't think the `FormAction` has to see if there is `extract_` method defined in the SDK. We can just override the extracted slot in the custom action. 

My issue is regarding the request to override a slot by defining a `extract_<slot name>` method in the `FormValidationAction` **without** overriding `required_slots`. We can inspect the python class during runtime and simply call all `extract_<slot name>` methods we find. However, I consider this too much magic and would rather go for explicitness in this case.

samsucik:
> I think if an extract_ method is defined and a slot mapping in the domain, we should print a warning. Currently the extract_ method would override any extractions based on mappings in the domain for this slot.

Now thinking about it, I agree. So then the solution to UC1 would be to implement custom `extract_{slot}` AND ensure that no mapping is defined for the slot in the domain, right?

> We can inspect the python class during runtime and simply call all extract_<slot name> methods we find. However, I consider this too much magic and would rather go for explicitness in this case.

I don't have a strong opinion here. Maybe we can use magic if it simplifies the job of the user. If we go for less magic and more transparency, then I'd create a new method (or even just an instance variable of `FormValidationAction`) called `slots_with_custom_mapping` or similar, where the user would list the slots they want to extract using custom functions. And then `required_slots` would be used to manage all the slots required by the form in UC2&3.

wochinge:
>  So then the solution to UC1 would be to implement custom extract_{slot} AND ensure that no mapping is defined for the slot in the domain, right?

Yes! We can implement a warning in the SDK in case both are defined.

> Then I'd create a new method (or even just an instance variable of FormValidationAction) called slots_with_custom_mapping or similar, where the user would list the slots they want to extract using custom functions. And then required_slots would be used to manage all the slots required by the form in UC2&3.

What about this approach?
> We could add a check that the FormValidation action only sets the `requested_slot` in case the output of `required_slots(...)` is different than the slots defined in the domain though 👍

samsucik:
> We could add a check that the `FormValidation` action only sets the `requested_slot` in case the output of `required_slots(...)` is different than the slots defined in the domain though 👍

In the context of UC1, personally, I would still go with a separate place where to specify the slots that should be extracted using custom `extract_{slot}` functions. For the sake of clarity. Because `required_slots` is, I think, a misleading name for someone who just wants to do UC1.

In the wider context, though, I think this kind of check could lead users towards using `FormValidationAction` the intended way. Though we should be careful here: The way I saw people use `required_slots`, sometimes they don't return the whole list of slots that the form needs to request, only those that haven't been filled yet. Also, the ordering of slots should be taken into account too. So, strictly speaking, I'd say that we should catch the cases where the empty slots in the list returned by `required_slots` are the same (and in the same order) as the empty slots from the list of slots for the form in the domain.

wochinge:
> The way I saw people use required_slots, sometimes they don't return the whole list of slots that the form needs to request

That wouldn't be a problem 👍 

> So, strictly speaking, I'd say that we should catch the cases where the empty slots in the list returned by required_slots are the same (and in the same order) as the empty slots from the list of slots for the form in the domain.

Thinking about it, having a function `additional_slots` (or `slots_with_custom_mapping`) seems actually cleaner and less error prone. Then `required_slots` would be `domain slots` + `slots with custom mapping`.
And we only set a `requested_slot` in case `required_slots` is different (different order, different values) than the pure slots from the domain. Does that make sense? 

Then things would look like this:

- Use case 1: change mapping for an "existing" slot: They can actually use `validate_<my slot>` for this
- Use case 2: add slot to be requested: Override `additional_slots` and define `extract_<my function>` 
- Use case 3: change set of requested slots: Override `additional_slots` and maybe change `next_requested_slot`

What do you think?

samsucik:
> having a function `additional_slots` (or `slots_with_custom_mapping`) seems actually cleaner and less error prone.

I agree -- if we're still talking about the thing meant for UC1.

> Then `required_slots` would be `domain slots` + `slots with custom mapping`.

Hmmm, this isn't aligned with the way I understood UC1. I thought the user already had all the required slots listed for the form in the domain, but for the custom mapping ones they didn't define any mapping in the domain and instead implemented custom `extract_{slot}` methods. Then the role of `slots_with_custom_mapping` as I understand it is only to remember which ones of the domain slots should be extracted using the custom extractor methods. In other words, `slots_with_custom_mapping` would always be a subset of `domain_slots`. Otherwise we bring in the issues from UC2&3.

> Use case 1: They can actually use `validate_<my slot>` for this

I have doubts about this because `validate` and `extract` are conceptually different things. But I've seen people doing extraction in `validate_{slot}` and maybe sometimes the difference isn't so clear. Or maybe I didn't get you and you can explain :-)

> Use case 2: Override `additional_slots` and define `extract_<my function>`

I think this introduces an unnatural gap between use cases 2 & 3 as we discussed at some point earlier. Because while this works for UC2, you (I assume) can't easily change whether the additional slot(s) will be requested before, after, or in between some of the remaining domain slots. Then the moment you want to change this ordering, you are in UC3 and need a place where you fully customise the list of required slots.

> Use case 3: Override `additional_slots` and maybe change `next_requested_slot`

Here's the problem. By overriding `additional_slots` you can't remove some of the domain slots from the required slots list. You also can't change the order. All of these changes you'd have to do somewhere else and `next_requested_slot` doesn't sound like a natural place. For UC3, I think the right place to make your changes is `required_slots`, where you have access to both the (unfilled) domain slots and the (unfilled) custom mapping ones and you can re-shuffle them and/or remove some as you need.

docs/docs/forms.mdx
wochinge: > without needing to define any extract_<slot_name> methods 

@kearnsw Not sure I'm understanding this part. You'll always need to define an `extract_<slot name>` method for custom extractions. We could indeed dynamically generate `required_slots` but I'd rather go for explicitness than magic here. @samsucik and I iterated a bit more on the solution and updated the PR accordingly. Hope it makes more sense now.

wochinge:
@samsucik Updated it according to our talk. Could you please have another look?

docs/docs/forms.mdx
```diff
@@ -360,74 +360,99 @@ to fully customize the validation process.
...
-extraction code. Rasa Open Source will trigger this function when the form is run.
+[Custom Action](./actions.mdx#custom-actions) `validate_<form_name>` to write your own
+extraction code. Rasa Open Source will trigger this action when the form is run.
 
-Make sure your custom action returns `SlotSet` events for every extracted value.
-The following example shows the implementation of a custom slot mapping which sets
-a slot based on the length of the last user message.
+If you're using the Rasa SDK we recommend you to extend the provided
+`FormValidationAction`. When using the `FormValidationAction`, two things are needed to
+extract customs slots:
```
samsucik, line 368: I think "custom slot" can lead to confusion. (It has confused me in the past.) We might want to consistently talk about "custom mapping", even though it will make some sentences more complicated...

docs/docs/forms.mdx
```diff
@@ -361,75 +361,151 @@ to fully customize the validation process.
...
+        sit_outside = "outdoor" in text_of_last_user_message
 
-        return [SlotSet("user_message_length", len(text_of_last_user_message))]
+        return {"outdoor_seating": sit_outside}
 ```
 
-### Requesting Extra Slots
+By default the `FormValidationAction`
+will automatically set the [`requested_slot`](forms.mdx#the-requested_slot-slot) to the
+first slot specified in `required_slots` which is not filled.
```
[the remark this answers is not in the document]
samsucik: I think the above comment is still relevant. Could we mention what's the default behaviour, i.e. when the user doesn't override `required_slots`?

docs/docs/forms.mdx
```diff
@@ -360,74 +360,99 @@ to fully customize the validation process.
...
+        tracker: "Tracker",
+        domain: "DomainDict",
+    ) -> Optional[List[Text]]:
+        additional_slots = ["outdoor_seating"]
+        if tracker.slots.get("outdoor_seating") is True:
+            # If the user wants to sit outside, ask
+            # if they want to sit in the shade or in the sun.
+            additional_slots.append("shade_or_sun")
+
+        return additional_slots + slots_mapped_in_domain
```
samsucik, line 455: I'm a huge fan of this <3

samsucik: Though we should also add a note or a warning here, reminding the user to implement `extract_<slot_name>` when a new slot is added that isn't mapped in the domain.

samsucik: I mean in the docs, not a warning in the code (there, it would be an error :-D).

samsucik (changes requested):
Looks good to me! I didn't necessarily catch _all_ the cosmetic issues but I think this is good to go conceptually.

docs/docs/forms.mdx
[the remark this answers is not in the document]
wochinge: > that if the user doesn't return anything from required_slots (or doesn't override required_slots at all), then the form won't request any further slots.

If `required_slots` returns `[]` it will indeed deactivate the form.

>  Could we mention what's the default behaviour, i.e. when the user doesn't override required_slots?

You meant that Rasa Open Source will then request the next slot  based on the slot mappings defined in the domain?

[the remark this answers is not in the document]
wochinge: > My understanding was that if I just want to implement custom slot mapping, I implement the corresponding extract_{slot}, nothing more.

That **only** works if this slot was mapped in the domain!

> And if I just want to change the set of required slots, I override required_slots and possibly add extract_{slot} for slots that aren't listed for this form in my domain file.

Exactly.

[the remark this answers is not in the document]
samsucik: > If `required_slots` returns `[]` it will indeed deactivate the form.

Ah, that's fair now with the new way `required_slot` will be implemented in `FormValidationAction`.

> You meant that Rasa Open Source will then request the next slot based on the slot mappings defined in the domain?

Exactly.

> That only works if this slot was mapped in the domain!

Ah, right, sure. And we plan to throw a warning if the user writes a custom extractor for a slot that's already mapped in the domain, right? So, then the only recommended way to do custom mapping would be to **not** map that slot in the domain, write a custom extractor method, and add the slot to the list inside `required_slots`, right?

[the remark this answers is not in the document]
wochinge: > You meant that Rasa Open Source will then request the next slot based on the slot mappings defined in the domain?

> Exactly.

Will do 👍 

> And we plan to throw a warning if the user writes a custom extractor for a slot that's already mapped in the domain, right? 

Can do 👍 

> So, then the only recommended way to do custom mapping would be to not map that slot in the domain, write a custom extractor method, and add the slot to the list inside required_slots, right?

Sounds good 👍

docs/docs/forms.mdx
wochinge: Sometimes I now just went with `additional slot` as the mapping focuses upon the extraction (not yet pushed)

docs/docs/forms.mdx
```diff
@@ -360,74 +360,101 @@ to fully customize the validation process.
...
 
-Make sure your custom action returns `SlotSet` events for every extracted value.
-The following example shows the implementation of a custom slot mapping which sets
-a slot based on the length of the last user message.
+If you're using the Rasa SDK we recommend you to extend the provided
+`FormValidationAction`. When using the `FormValidationAction`, two things are needed to
+extract customs slots:
+
+1. Define a method `extract_<slot_name>` for every custom slot mapping which you want to
+   extract.
```
samsucik, line 371: What about this:
```suggestion
1. Define a method `extract_<slot_name>` for every slot that should be extracted using a custom mapping.
```
My motivation is to get rid of the `custom slot` term as much as possible to stress out that it is the mapping that is `custom`. Though the possible confusion between `extracting` and `mapping` remains... Then perhaps this could partly do the job?
```suggestion
1. Define a method `extract_<slot_name>` for every slot that should be mapped in a custom way.
```

docs/docs/forms.mdx
```diff
@@ -360,74 +360,101 @@ to fully customize the validation process.
...
+2. Optional: If a slot is not part of your mapped slots in the domain,
+   override `required_slots` to make sure the additional slot is considered by the form.
```
samsucik, line 373: Wait, currently we want users to **not** include in the domain (for the given form) any slots that should use custom mappings, right? So this 2nd point should be more like two different points:
```suggestion
2. Make sure that in the domain file you list for your form only those slots that don't use custom mappings.
3. Override `required_slots` to add all slots with custom mappings to the list of slots the form should request.
```

docs/docs/forms.mdx
```diff
@@ -360,74 +360,101 @@ to fully customize the validation process.
...
+
+The following example shows the implementation of a form which extracts a slot
+`outdoor_seating` in addition to the slots which were mapped in the domain.
```
samsucik, line 376: How about:
```suggestion
The following example shows the implementation of a form which extracts a slot
`outdoor_seating` in a custom way, in addition to the slots whose mappings are defined in the domain.
```
(trying to get rid of "mapped in the domain")

samsucik (changes requested):
Now that the rough shapes are more or less clear, I also doubled down on some smaller terminology issues :-)

wochinge:
I also made the PR with the changes to the SDK https://github.com/RasaHQ/rasa-sdk/pull/338

docs/docs/forms.mdx
```diff
@@ -360,74 +360,104 @@ to fully customize the validation process.
...
-the form validates user input. Set the slot `requested_slot` to the name of the slot
-which should be extracted next. If all desired slots are filled, set `requested_slot`
-to `None`.
+### Dynamic Form Behavior
 
-The following example shows the implementation of a custom action which requests
-the three slots `cuisine`, `num_people`, and `outdoor_seating`.
+By default Rasa Open Source will ask for the next
+[`requested_slot`](forms.mdx#the-requested_slot-slot) defined in the slot mappings
+in the domain, which is not filled yet. If you use
```
samsucik, line 422: Hm, I'm trying to think of a better way to refer to the slots listed for the form in the domain. How about:
```suggestion
By default Rasa Open Source will ask for the next empty slot from the slots
listed for your form in the domain file. If you use
```

docs/docs/forms.mdx
```diff
@@ -360,74 +360,104 @@ to fully customize the validation process.
...
 
-from rasa_sdk import Tracker
-from rasa_sdk.events import EventType
-from rasa_sdk.executor import CollectingDispatcher
-from rasa_sdk import Action
-from rasa_sdk.events import SlotSet
+If you are using the Rasa SDK, we recommend you to use the `FormValidationAction` and
+override `required_slots` to fit your dynamic behavior. Please remember to implement
+a method `extract_<slot name>` for every slot which isn't defined in the slot mappings
+of your domain as it is described [here](forms.mdx#custom-slot-mappings).
```
samsucik, line 434:
```suggestion
a method `extract_<slot name>` for every slot which doesn't use a predefined mapping,
as described in [Custom Slot Mappings](forms.mdx#custom-slot-mappings).
```

docs/docs/forms.mdx
```diff
@@ -360,74 +360,104 @@ to fully customize the validation process.
...
+extraction code. Rasa Open Source will trigger this action when the form is run.
+
+If you're using the Rasa SDK we recommend you to extend the provided
+`FormValidationAction`. When using the `FormValidationAction`, three steps are required
+to extract customs slots:
+
+1. Define a method `extract_<slot_name>` for every slot that should be mapped in a
+   custom way.
+2. Make sure that in the domain file you list for your form only those slots that don't
+   use custom mappings.
```
samsucik, line 373: I'm thinking about using "predefined mapping" more often. @wochinge what do you think?
```suggestion
2. Make sure that in the domain file you list for your form only those slots that use 
[predefined mappings](./forms.mdx#slot-mappings).
```

docs/docs/forms.mdx
```diff
@@ -360,74 +360,104 @@ to fully customize the validation process.
...
+3. Override `required_slots` to add all slots with custom mappings to the list of slots
+   the form should request.
+
+The following example shows the implementation of a form which extracts a slot
+`outdoor_seating` in a custom way, in addition to the slots whose mappings are defined
+in the domain.
```
samsucik, line 379:
```suggestion
`outdoor_seating` in a custom way, in addition to the slots which use 
[predefined mappings](./forms.mdx#slot-mappings).
```

samsucik (approved):
Looks good, I just left a few comments regarding terminology. In particular, I've tried to reinforce the consistent use of "predefined mappings" vs "custom mappings".

docs/docs/forms.mdx
wochinge: nice, I like it!

wochinge:
@melindaloubser1 @JulianGerhard21 @kearnsw Any further comments or can I merge?

docs/docs/forms.mdx
```diff
@@ -360,74 +360,103 @@ to fully customize the validation process.
...
+This is, for example, useful when you need further details based on how
+a previous slot was filled or you want to change the order in which slots are requested.
 
-from rasa_sdk import Tracker
-from rasa_sdk.events import EventType
-from rasa_sdk.executor import CollectingDispatcher
-from rasa_sdk import Action
-from rasa_sdk.events import SlotSet
+If you are using the Rasa SDK, we recommend you to use the `FormValidationAction` and
+override `required_slots` to fit your dynamic behavior. Please remember to implement
```
melindaloubser1, line 431:
```suggestion
override `required_slots` to fit your dynamic behavior. You should implement
```

docs/docs/forms.mdx
```diff
@@ -360,74 +360,103 @@ to fully customize the validation process.
...
+a method `extract_<slot name>` for every slot which doesn't use a predefined mapping,
+as described in [Custom Slot Mappings](forms.mdx#custom-slot-mappings).
+The example below will ask the user if they want to sit in
+the shade or in the sun in case they told us that they want to sit outside.
```
melindaloubser1, line 435:
```suggestion
the shade or in the sun in case they said they want to sit outside.
```

melindaloubser1 (approved):
It's been covered pretty thoroughly, just commented two small language things

samsucik (approved):
Awesome!

PR #7235, opened 2020-11-10
fix conversation starts with slots initial value

Ghostvv:
**Proposed changes**:
- fix rules with conversation starts with slots initial value

### DIFF
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -281,7 +281,7 @@ class RulePolicy(MemoizationPolicy):
             )
 
     @staticmethod
-    def _check_slots_fingerprint(
+    def _expected_but_missing_slots(
         fingerprint: Dict[Text, List[Text]], state: State
     ) -> Set[Text]:
         expected_slots = set(fingerprint.get(SLOTS, {}))
@@ -314,9 +314,9 @@ class RulePolicy(MemoizationPolicy):
         if action_name and missing_fingerprint_slots:
             error_messages.append(
                 f"- the action '{action_name}' in rule '{rule_name}' does not set some "
-                f"of the slots that it sets in other rules. Slots not set in rule '{rule_name}': "
-                f"'{', '.join(missing_fingerprint_slots)}'. Please update the rule with "
-                f"an appropriate slot or if it is the last action "
+                f"of the slots that it sets in other rules. Slots not set in rule "
+                f"'{rule_name}': '{', '.join(missing_fingerprint_slots)}'. Please "
+                f"update the rule with an appropriate slot or if it is the last action "
                 f"add 'wait_for_user_input: false' after this action."
             )
         if action_name and fingerprint_active_loops:
@@ -372,7 +372,7 @@ class RulePolicy(MemoizationPolicy):
                     # for a previous action if current action is rule snippet action
                     continue
 
-                missing_expected_slots = self._check_slots_fingerprint(
+                missing_expected_slots = self._expected_but_missing_slots(
                     fingerprint, state
                 )
                 expected_active_loops = self._check_active_loops_fingerprint(
@@ -660,22 +660,34 @@ class RulePolicy(MemoizationPolicy):
         # turn_index goes back in time
         reversed_rule_states = list(reversed(self._rule_key_to_state(rule_key)))
 
-        return bool(
-            # rule is shorter than current turn index
-            turn_index >= len(reversed_rule_states)
-            # current rule and state turns are conversation starters
-            or (
-                not reversed_rule_states[turn_index].get(PREVIOUS_ACTION)
-                and not conversation_state.get(PREVIOUS_ACTION)
-            )
-            # check that current rule turn features are present in current state turn
-            or (
-                reversed_rule_states[turn_index].get(PREVIOUS_ACTION)
-                and conversation_state.get(PREVIOUS_ACTION)
-                and self._does_rule_match_state(
-                    reversed_rule_states[turn_index], conversation_state
-                )
-            )
+        # the rule must be applicable because we got (without any applicability issues)
+        # further in the conversation history than the rule's length
+        if turn_index >= len(reversed_rule_states):
+            return True
+
+        # a state has previous action if and only if it is not a conversation start
+        # state
+        current_previous_action = conversation_state.get(PREVIOUS_ACTION)
+        rule_previous_action = reversed_rule_states[turn_index].get(PREVIOUS_ACTION)
+
+        # current rule state is a conversation starter (due to conversation_start: true)
+        # but current conversation state is not.
+        # or
+        # current conversation state is a starter (due to initial_value set for a slot)
+        # but current rule state is not a starter
+        if current_previous_action and not rule_previous_action:
+            return False
+
+        # current conversation state and rule state are conversation starters.
+        # any slots with initial_value set will necessarily be in both states and don't
+        # need to be checked.
+        if not current_previous_action:
+            # Ignore other states for conversation starts
+            return True
+
+        # check: current rule state features are present in current conversation state
+        return self._does_rule_match_state(
+            reversed_rule_states[turn_index], conversation_state
         )
 
     def _get_possible_keys(
@@ -744,6 +756,8 @@ class RulePolicy(MemoizationPolicy):
     ) -> Tuple[Optional[Text], Optional[Text]]:
         tracker_as_states = self.featurizer.prediction_states([tracker], domain)
         states = tracker_as_states[0]
+        for s in states:
+            print(f":::::{s}")
 
         logger.debug(f"Current tracker state: {states}")
 
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -125,7 +125,6 @@ class RulePolicy(MemoizationPolicy):
             enable_fallback_prediction: If `True` `core_fallback_action_name` is
                 predicted in case no rule matched.
         """
-
         self._core_fallback_threshold = core_fallback_threshold
         self._fallback_action_name = core_fallback_action_name
         self._enable_fallback_prediction = enable_fallback_prediction
@@ -194,7 +193,6 @@ class RulePolicy(MemoizationPolicy):
         Returns:
             modified states
         """
-
         # leave only last 2 dialogue turns to
         # - capture previous meaningful action before action_listen
         # - ignore previous intent
@@ -226,7 +224,6 @@ class RulePolicy(MemoizationPolicy):
         Returns:
             lookup dictionary
         """
-
         lookup = {}
         for states, actions in zip(trackers_as_states, trackers_as_actions):
             action = actions[0]
@@ -665,7 +662,6 @@ class RulePolicy(MemoizationPolicy):
         Returns:
             a boolean that says whether the rule is applicable to current state
         """
-
         # turn_index goes back in time
         reversed_rule_states = list(reversed(self._rule_key_to_state(rule_key)))
 
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -664,12 +664,15 @@ class RulePolicy(MemoizationPolicy):
         return bool(
             # rule is shorter than current turn index
             turn_index >= len(reversed_rule_states)
-            # current rule and state turns are empty
-            or (not reversed_rule_states[turn_index] and not conversation_state)
+            # current rule and state turns are conversation starters
+            or (
+                not reversed_rule_states[turn_index].get(PREVIOUS_ACTION)
+                and not conversation_state.get(PREVIOUS_ACTION)
+            )
             # check that current rule turn features are present in current state turn
             or (
-                reversed_rule_states[turn_index]
-                and conversation_state
+                reversed_rule_states[turn_index].get(PREVIOUS_ACTION)
+                and conversation_state.get(PREVIOUS_ACTION)
                 and self._does_rule_match_state(
                     reversed_rule_states[turn_index], conversation_state
                 )
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -655,7 +655,16 @@ class RulePolicy(MemoizationPolicy):
     def _is_rule_applicable(
         self, rule_key: Text, turn_index: int, conversation_state: State
     ) -> bool:
-        """Check if rule is satisfied with current state at turn."""
+        """Check if rule is satisfied with current state at turn.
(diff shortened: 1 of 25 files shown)


Ghostvv:
@samsucik as agreed during debug session, could you please add tests?

samsucik:
@Ghostvv I've added 2 tests:
1. Checking that two contradicting rules don't lead to contradiction if one of them is marked as `conversation_start: true`.
1. Checking that the stuff from 1) works when a slot has an initial value set in the domain, whether or not the slot is also explicitly set at the beginning of a story on which you want to make predictions.

Does this sound right, or did you have another test case in mind?

wochinge:
Please change the target of this PR to `master` or `2.1.x` once I release it this afternoon.

Ghostvv:
@wochinge why? we don't support `2.0` bug fixes anymore?

wochinge:
Yes, you can do that but then you need to take care of 1) merging these changes to `master` and cutting the release

tests/core/policies/test_rule_policy.py
```diff
@@ -92,6 +92,138 @@ def _form_activation_rule(
...
+        "greet rule at conversation start",
+        domain=domain,
+        slots=domain.slots,
+        evts=[
+            ActionExecuted(ACTION_LISTEN_NAME),
+            UserUttered(intent={"name": GREET_INTENT_NAME}),
+            ActionExecuted(UTTER_GREET_ACTION),
+        ],
+    )
+    greet_rule_at_conversation_start.is_rule_tracker = True
```
Ghostvv, line 121: I think `is_rule_tracker` can be a part of `from_events`

tests/core/policies/test_rule_policy.py
```diff
@@ -92,6 +92,138 @@ def _form_activation_rule(
...
+        evts=[
+            ActionExecuted(RULE_SNIPPET_ACTION_NAME),
+            ActionExecuted(ACTION_LISTEN_NAME),
+            UserUttered(intent={"name": GREET_INTENT_NAME}),
+            ActionExecuted(utter_anti_greet_action),
+        ],
+    )
+    anti_greet_rule.is_rule_tracker = True
+
+    policy.train(
```
Ghostvv, line 136: could you please add comment that by calling `policy.train` we check that no contradiction is raised

tests/core/policies/test_rule_policy.py
```diff
@@ -92,6 +92,138 @@ def _form_activation_rule(
...
+    action_probabilities_1 = policy.predict_action_probabilities(
+        DialogueStateTracker.from_events(
+            "test conversation", evts=conversation_events, slots=domain.slots
+        ),
+        domain,
+        RegexInterpreter(),
+    )
+    assert_predicted_action(action_probabilities_1, domain, UTTER_GREET_ACTION)
+
+    conversation_events_with_initial_slot_explicit = [
```
Ghostvv, line 210: hm, I don't see the difference between `conversation_events` and `conversation_events_with_initial_slot_explicit`

tests/core/policies/test_rule_policy.py
```diff
@@ -168,7 +300,10 @@ def test_incomplete_rules_due_to_slots():
         policy.train([complete_rule, incomplete_rule], domain, RegexInterpreter())
     assert all(
         name in execinfo.value.message
-        for name in {some_action, incomplete_rule.sender_id,}
+        for name in {
+            some_action,
+            incomplete_rule.sender_id,
+        }
```
Ghostvv, line 306: what's wrong with keeping it on one line?
```suggestion
        for name in {some_action, incomplete_rule.sender_id}
```

Ghostvv (commented):
@samsucik that are the tests that I had in my mind. Left a couple of comments.
@wochinge could you please give it a review, since I did the original code change, it wouldn't be good if I review the PR

tests/core/policies/test_rule_policy.py
samsucik: Good point. I was just copy-pasting existing tests and changing those, that's why I didn't even know about this option :-)

tests/core/policies/test_rule_policy.py
samsucik: Sure.

tests/core/policies/test_rule_policy.py
samsucik: My bad; basically, `conversation_events` shouldn't have the `SlotSet` event at the start.

tests/core/policies/test_rule_policy.py
samsucik: Ask the formatter tool I'd say :-D But it's true that I should've checked the changes made by `black` to the code. @wochinge any thoughts?

samsucik: For now, I've pushed without fixing this. Partly because I didn't want to mark the relevant lines as `# fmt:off/on` (because to me this smells as it would need to be done everywhere), and because I didn't want to push something that doesn't pass the `black` checks.

Ghostvv: what do you mean? my suggestion should pass the `black` check

Ghostvv: also, just do `black .` in your rasa repo, it'll do the job automatically for you

Ghostvv: except removing a comma (but this you can do manually if you spot it)

samsucik: @Ghostvv the changes were made by my pre-commit hook, not by me. I've just checked and it looks like my pre-commit hook uses maybe a different version (or config) of `black` because the pre-commit prefers the multi-line form of the set expression above while `make formatter` (i.e. `black rasa tests`) prefers the single-line version. I have no idea what to do...

Ghostvv: did you try updating `black`? I have no idea what is "pre-commit hook" 😅

samsucik: I've solved it by deleting the pre-commit hook altogether because it was actually using a newer version of `black` than `make formatter`. Those two versions will have to be synced though because in the repo readme [here](https://github.com/RasaHQ/rasa#formatting) we tell people how to use `black` in a pre-commit hook (so that reformatting is done automatically when you `git commit` and you don't have to run it manually).

rasa/core/policies/rule_policy.py
```diff
@@ -286,11 +286,11 @@ def _check_slots_fingerprint(
...
         expected_slots = set(fingerprint.get(SLOTS, {}))
         current_slots = set(state.get(SLOTS, {}).keys())
-        if expected_slots == current_slots:
-            # all expected slots are satisfied
+        if expected_slots.issubset(current_slots):
+            # all expected slots are found in the currently set slots
             return set()
 
-        return expected_slots
+        return expected_slots.difference(current_slots)
```
Ghostvv, line 293: can we simply return `expected_slots.difference(current_slots)`? in case of subset, it'll be exactly `set()`, no?

samsucik, line 293: @Ghostvv what do you think, shall I change this similarly in `_check_active_loops_fingerprint` too? Also, I've noticed that there we use `expected_active_loops` but this will always be at most one loop, or not?

Ghostvv: yes in the other case I think it can be only one loop

Ghostvv: oh, effectively in case of loops, I check if it a subset

Ghostvv: I think, you can leave it as is

Ghostvv: aaah

Ghostvv: one action can set 2 loops: `some_loop` and `null`, that's why it is expected loops

samsucik: You are right, should've done it right away.

samsucik: Ah, right, now it makes sense to me. Thanks! I'll leave everything as it is then.

wochinge:
Hey, not sure if I can get to it today. I'm at a conference today. Will try to get it done by latest friday.

Ghostvv:
@wochinge would you have time to review it today, or should we ask someone else?

wochinge:
I'll have a look today 👍

rasa/core/policies/rule_policy.py
```diff
@@ -309,16 +306,16 @@ def _check_active_loops_fingerprint(
...
     ) -> List[Text]:
         error_messages = []
-        if action_name and fingerprint_slots:
+        if action_name and missing_fingerprint_slots:
             error_messages.append(
-                f"- the action '{action_name}' in rule '{rule_name}' does not set all "
-                f"the slots, that it sets in other rules: "
-                f"'{', '.join(fingerprint_slots)}'. Please update the rule with "
+                f"- the action '{action_name}' in rule '{rule_name}' does not set some "
+                f"of the slots that it sets in other rules. Slots not set in rule '{rule_name}': "
```
wochinge, line 317: can you please make sure that this line doesn't exceed the 88 char limit?

rasa/core/policies/rule_policy.py
```diff
@@ -664,12 +663,15 @@ def _is_rule_applicable(
         return bool(
             # rule is shorter than current turn index
             turn_index >= len(reversed_rule_states)
-            # current rule and state turns are empty
-            or (not reversed_rule_states[turn_index] and not conversation_state)
+            # current rule and state turns are conversation starters
```
wochinge, line 666: what do you think of this rewrite?
It avoids repeating the index access with `turn_index` and the `get` call to the dictionary

```python
    def _is_rule_applicable(
        self, rule_key: Text, turn_index: int, conversation_state: State
    ) -> bool:
        """Check if rule is satisfied with current state at turn."""

        # turn_index goes back in time
        reversed_rule_states = list(reversed(self._rule_key_to_state(rule_key)))

        if turn_index >= len(reversed_rule_states):
            return True

        current_previous_action = conversation_state.get(PREVIOUS_ACTION)
        rule_previous_action = reversed_rule_states[turn_index].get(PREVIOUS_ACTION)

        if current_previous_action != rule_previous_action:
            return False

        if not current_previous_action:
            # Ignore other states for conversation starts
            return True

        return self._does_rule_match_state(
            reversed_rule_states[turn_index], conversation_state
        )
```

wochinge, line 666: I think the comment is slightly useful, but we could make it more useful by stating something like:
```suggestion
            # current rule and state turns are conversation starters.
            # Don't consider any other state than actions in this case.
```

rasa/core/policies/rule_policy.py
```diff
@@ -286,11 +286,8 @@ def _check_slots_fingerprint(
     ) -> Set[Text]:
```
wochinge, line 286: can we change the function name `_check_slots_fingerprint` to something like `_expected_but_missing_slots`?

wochinge (commented):
Nice, edge cases 💯 
* could you please add a changelog entry?

rasa/core/policies/rule_policy.py
samsucik: Yes!

rasa/core/policies/rule_policy.py
samsucik: I'm re-writing the comments + adding some more :-)

tests/core/policies/test_rule_policy.py
```diff
@@ -92,6 +92,191 @@ def _form_activation_rule(
...
+    action_probabilities_2 = policy.predict_action_probabilities(
+        DialogueStateTracker.from_events(
+            "test conversation with initial slot value explicitly set",
+            evts=conversation_events_with_initial_slot_explicit,
+            slots=domain.slots,
+        ),
+        domain,
+        RegexInterpreter(),
+    )
+    assert_predicted_action(action_probabilities_2, domain, UTTER_GREET_ACTION)
```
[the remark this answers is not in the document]
samsucik: Good point :-)

rasa/core/policies/rule_policy.py
samsucik: Thanks, I've adopted this rewritten form, but I also added some more comments :-)

rasa/core/policies/rule_policy.py
```diff
@@ -661,19 +660,34 @@ def _is_rule_applicable(
...
+        # a state has previous action if and only if it is not a conversation start
+        # state
+        current_previous_action = conversation_state.get(PREVIOUS_ACTION)
+        rule_previous_action = reversed_rule_states[turn_index].get(PREVIOUS_ACTION)
+
+        # current rule state is a conversation starter (due to conversation_start: true)
+        # but current conversation state is not.
+        # or
+        # current conversation state is a starter (due to initial_value set for a slot)
+        # but current rule state is not a starter
```
samsucik, line 677: @Ghostvv could you look at this case? I don't see why it would make the rule not applicable. But I also don't think this case can every happen with our current implementation... Either way, my brain hurts from trying to understand this and trying to write a test case that would point out the possible mistake here.

Ghostvv: not sure what you mean, but I updated your code, it should be equivalent to the bool now

samsucik: Thanks for the changes!

Ghostvv:
@wochinge I think it is ready for final review

Ghostvv:
@wochinge do you have an idea why docs PR is failing? we didn't change any docs in this PR. Maybe it is failing on 2.0.x branch?

rasa/core/policies/rule_policy.py
```diff
@@ -656,24 +654,46 @@ def _rule_key_to_state(rule_key: Text) -> List[State]:
...
-            or (not reversed_rule_states[turn_index] and not conversation_state)
-            # check that current rule turn features are present in current state turn
-            or (
-                reversed_rule_states[turn_index]
-                and conversation_state
-                and self._does_rule_match_state(
-                    reversed_rule_states[turn_index], conversation_state
-                )
-            )
+        # the rule must be applicable because we got (without any applicability issues)
```
wochinge, line 670: Nice comments!

wochinge (approved):

Ghostvv:
@samsucik after it is merged, could you please release a minor for 2.0, merge `2.0.x` into `2.1.x`, release a minor for 2.1, merge `2.1.x` into `master`? sorry for the longest chain 🙈

samsucik:
@Ghostvv sure. This will be my first time; sounds like a good opportunity to learn something new! Wish me luck though 😄

PR #7246, opened 2020-11-11
Fixed changing fingerprints even though training data did not change

tmbo:
**Issue**:
- we relied on hashes in some places to fingerprint parts of the configuration / training data. hashes of strings are not stable across python (>3.3) runs, which means that the fingerprint changes even though the configuration / data is still the same

**Proposed changes**:
- implement proper fingerprinting using md5 for strings instead of python native hashes

### DIFF
--- a/rasa/shared/utils/io.py
+++ b/rasa/shared/utils/io.py
@@ -231,7 +231,7 @@ def get_dictionary_fingerprint(
         encoding: encoding used for dumping objects as strings
 
     Returns:
-        the hash of the dictionary
+        The hash of the dictionary
     """
     stringified = json.dumps(
         {
--- a/rasa/shared/utils/io.py
+++ b/rasa/shared/utils/io.py
@@ -245,18 +245,20 @@ def get_dictionary_fingerprint(
     return get_text_hash(stringified, encoding)
 
 
-def get_list_fingerprint(l: List[Any], encoding: Text = DEFAULT_ENCODING) -> Text:
+def get_list_fingerprint(
+    elements: List[Any], encoding: Text = DEFAULT_ENCODING
+) -> Text:
     """Calculate a fingerprint for an unordered list.
 
     Args:
-        l: unordered list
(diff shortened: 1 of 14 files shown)


tests/utils/test_io.py
```diff
@@ -91,3 +93,34 @@ def test_write_utf_8_yaml_file(tmp_path: Path):
...
+        [],
+        ["a"],
+        [{}],
+        [None],
+    ],
+)
+def test_fingerprint_dicionaries(dictionary):
+    assert rasa.shared.utils.io.deep_container_fingerprint(
+        dictionary
+    ) == rasa.shared.utils.io.deep_container_fingerprint(copy.deepcopy(dictionary))
```
tmbo, line 115:
```suggestion
@pytest.mark.parametrize(
    "container",
    [
        {},
        {"hello": "world"},
        {1: 2},
        {"foo": ["bar"]},
        {"a": []},
        [],
        ["a"],
        [{}],
        [None],
    ],
)
def test_fingerprint_containers(container):
    assert rasa.shared.utils.io.deep_container_fingerprint(
        container
    ) == rasa.shared.utils.io.deep_container_fingerprint(copy.deepcopy(container))
```

rasa/shared/utils/io.py
```diff
@@ -194,6 +194,75 @@ def list_subdirectories(path: Text) -> List[Text]:
...
+
+    The dictionary can contain any keys and values which are either a dict,
+    a list or a elements which can be dumped as a string.
+
+    Args:
+        dictionary: dictionary to be hashed
+        encoding: encoding used for dumping objects as strings
+
+    Returns:
+        the hash of the dictionary
```
alwx, line 234: We usually start from a capital letter here.

alwx: But it's probably tested automatically anyway :)

alwx (approved):
Looks great! 🚀

PR #7247, opened 2020-11-11
Add list handling in the send_custom_json method on facebook.py

### LINKED ISSUE #3153: consider replacing os.path uses with pathlib
The standard python library `pathlib` makes interactions with paths much easier. We should consider replacing the old `os.path` with the pathlib `functions`.

### LINKED ISSUE #5657: Type error in facebook.py
Hi,

I tried deploying the bot with Facebook and one error appears when I start a form:

TypeError: pop() takes at most 1 argument (2 given) (in rasa/core/channels/facebook.py : line 229).

I did a workaround but I thought you guys might want to take a look at this.

Cheers.

silvasara:
**Proposed changes**:
- Add list handling in the `send_custom_json` method on `channels/facebook.py`.
- Resolves #5657 


sara-tagger:
Thanks for submitting a pull request 🚀 @wochinge will take a look at it as soon as possible ✨

wochinge (commented):
Hi @silvasara 

Thanks for your contribution! 
Can you please add tests for this?

silvasara:
@wochinge  I don't know if I can do the tests, but I'll try. Where would I put this test? The only test I found for this file was this:

https://github.com/RasaHQ/rasa/blob/0ccfa20ef72f7f6d4e4e1f3e0bf88d8568bfcb7e/tests/core/test_channels.py#L146-L165

wochinge:
@silvasara Yep, it would go into `test_channels`. You can also create a new file `test_facebook.py` in `tests.core.channels` and move all facebook tests there

jppgomes:
Hello, I checked the missing tests for the correction, can I implement and send the pr? @wochinge

wochinge:
@jppgomes Sure! It would be a pleasure!

.pre-commit-config.yaml
```diff
@@ -1,10 +1,10 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 19.10b0
+    rev: 20.8b1
```
wochinge, line 3: can you please change the items in this file back? These hooks should work as they are once you do `poetry install` in the cloned repository

tests/test_facebook.py
```diff
@@ -0,0 +1,61 @@
...
+    # END DOC INCLUDE
+    # the above marker marks the end of the code snipped included
+    # in the docs
+    routes_list = utils.list_routes(s)
+
+    assert routes_list["fb_webhook.health"].startswith("/webhooks/facebook")
+    assert routes_list["fb_webhook.webhook"].startswith("/webhooks/facebook/webhook")
+
+
+def test_facebook_send_custon_json_list():
```
wochinge, line 52: this doesn't test `send_custom_json` - we need to call the actual code in this test (same in the one below)

rasa/core/channels/facebook.py
```diff
@@ -279,8 +279,18 @@ async def send_custom_json(
         self, recipient_id: Text, json_message: Dict[Text, Any], **kwargs: Any
     ) -> None:
         """Sends custom json data to the output."""
-
-        recipient_id = json_message.pop("sender", {}).pop("id", None) or recipient_id
+        if isinstance(json_message, list):
+            if isinstance(json_message.pop(), list):
```
wochinge, line 283: `pop` gets **and** removes stuff from the list - do we really want that?

wavymazy:
Hi guys, I'm the author of the post for this issue. I had some time today so I though might as well update this PR.

**Context**:

I have retested the issue with Facebook to understand where it comes from.
The issue stems from the custom template/response if you have any. Basically, if we follow the template (https://rasa.com/docs/rasa/responses/#custom-output-payloads), it is said that we can send any arbitrary output we want. But what will end up happening if we have 2 "type" items (under "blocks"), is that the output will be a list (under "blocks"). The MessengerBot.send_custom_json() previously didn't handle this well at all.

**Propose solution:**

I think the best way to handle this is by only looking for the sender ID in the custom template only if that template is of type Dict. Otherwise we could use the ID that is given as a function paramter.

wochinge:
that would be awesome if you could wrap this up @wavymazy . 

> Otherwise we could use the ID that is given as a function paramter.

Would do you mean by this?

wavymazy:
Hi @wochinge,

Unfortunately it seems I can't make a direct change to this PR unless given access.

With regards to your question, the ID I'm refering to is the `recipient_id` that is given as a paramter. 

`    async def send_custom_json(
        self, recipient_id: Text, json_message: Dict[Text, Any], **kwargs: Any
    ) -> None:`

wochinge:
@wavymazy How about forking the fork / cherry-picking the commits and then you create a new PR?

wavymazy:
Ok @wochinge. I can try this.

wavymazy:
Hi, @wochinge, I created my PR, cherry picking from this one (link : https://github.com/RasaHQ/rasa/pull/8332).

wochinge:
Awesome! I'll close this one then.

PR #7263, opened 2020-11-12
replace `pika` with `aio-pika`

wochinge:
**Proposed changes**:
- replace `pika` with `aio-pika`
- `EventBroker.create` and `EventBroker.from_endpoint_config` are now `async`. The built-in brokers were adapted. A warning will be issued if brokers loaded from module string don't have `async` versions of these methods. This was required to run the `async` `connect` method of `PikaEventBroker`. I left a comment in the PR about this as it's not necessarily required.
- It's now deprecated to specify `RABBITMQ_SSL_CA_FILE` and `RABBITMQ_SSL_KEY_PASSWORD` via env variables as `aio-pika` doesn't support this in their SSL options. `RABBITMQ_SSL_CA_FILE` can be supplied via URL as described in the changelog. `RABBITMQ_SSL_KEY_PASSWORD` is dropped.
- publishing to RabbitMQ is no longer blocking. As we add the publishing to the event loop of Sanic, this should improve the performance of predictions. 
- update the examples in the docs.
- clean up `EventBroker` connections when server is shutting down. This is done via the `Sanic` callback `after_server_stop`. This is more reliable then using `__del__` and also support `async` functions

**Side Effects**
- As we use `connect_robustly` we instruct `aio-pika` to automatically try to re-connect in case RabbitMQ goes down.


rasa/core/brokers/broker.py
```diff
@@ -11,17 +12,22 @@ class EventBroker:
     """Base class for any event broker implementation."""
 
     @staticmethod
-    def create(
+    async def create(
```
wochinge, line 15: validate who calls this

rasa/core/brokers/broker.py
```diff
@@ -11,17 +12,22 @@ class EventBroker:
...
         obj: Union["EventBroker", EndpointConfig, None],
+        loop: Optional[AbstractEventLoop] = None,
     ) -> Optional["EventBroker"]:
```
wochinge, line 18: docstrings please!

rasa/core/brokers/kafka.py
```diff
@@ -75,7 +77,12 @@ def __init__(
         logging.getLogger("kafka").setLevel(loglevel)
 
     @classmethod
-    def from_endpoint_config(cls, broker_config) -> Optional["KafkaEventBroker"]:
+    async def from_endpoint_config(
```
wochinge, line 80: we could technically do it without that by using `loop.run_until_complete` 🤔 Currently it wouldn't make a difference but it might be nicer in the future if we wanna do something else in between

changelog/7263.removal.md
```diff
@@ -0,0 +1,13 @@
+The [Pika Event Broker](event-brokers.mdx#pika-event-broker) no longer supports
+the environment variables `RABBITMQ_SSL_CA_FILE` and `RABBITMQ_SSL_KEY_PASSWORD`.
+You can alternatively specify them in RabbitMQ connection url as described in the 
+[RabbitMQ documentation](https://www.rabbitmq.com/uri-query-parameters.html).
+
+```yaml-rasa title="endpoints.yml
+event_broker:
+ type: pika
+ url: "amqps://user:password@host?cacertfile=path_to_ca_cert&password=private_key_password"
```
wochinge, line 9: I'm not sure about the `password` part. It's not officially documented 😬

wochinge:
@alwx Does the poetry error mean I have to downgrade poetry and then create the lockfile again?

wochinge:
I'd like to get this in the 2.1 release on Tuesday btw.

rasa/core/brokers/pika.py
```diff
@@ -371,343 +123,213 @@ def _get_queues_from_args(
...
+        await queue.bind(exchange, "")
 
     def __del__(self) -> None:
-        if self.pika_message_processor:
-            self.pika_message_processor.close()
-
-        if self.process and self.process.is_alive():
-            self.process.terminate()
+        """Closes connection when object is destroyed."""
+        self._loop.run_until_complete(self._close())
```
wochinge, line 224: this might break as the loop is already running

changelog/7263.removal.md
alwx: maybe makes sense to remove it then

alwx: by the way, why does the password specified twice in this string?

rasa/core/brokers/pika.py
```diff
@@ -371,343 +123,213 @@ def _get_queues_from_args(
...
-        Returns:
-            `PikaEventBroker` if `broker_config` was supplied, else `None`.
-        """
-        if broker_config is None:
-            return None
-
-        return cls(broker_config.url, **broker_config.kwargs)
+    def is_ready(self) -> bool:
+        """Return `True` if a connection was established."""
+        return self._exchange is not None
```
alwx, line 241: What do you think of actually checking if the connection is ready here? Or at least updating the docstring to make it less confusing why we check for `self._exchange` to be `not None`.

rasa/core/brokers/pika.py
```diff
@@ -371,343 +123,213 @@ def _get_queues_from_args(
...
+    if client_certificate_path and client_key_path:
+        logger.debug(f"Configuring SSL context for RabbitMQ host '{rabbitmq_host}'.")
+        return {
+            "certfile": client_certificate_path,
+            "client_key_path": client_key_path,
+            "cert_reqs": ssl.CERT_REQUIRED,
+        }
 
-        logger.error(f"Failed to publish Pika event on host '{self.host}':\n{body}")
+    return None
```
alwx, line 335: Cool to see this file becoming two times smaller! 🚀

alwx (approved):

alwx:
Added some small comments, but the code looks fine in general!

wochinge:
@alwx Thanks for the quick review! What about https://github.com/RasaHQ/rasa/pull/7263#issuecomment-726662495 ?

changelog/7263.removal.md
wochinge: The second one is hopefully the one for the private key 🤔

rasa/core/brokers/pika.py
wochinge: adapted it

changelog/7263.removal.md
wochinge: removed the `password` stuff and made it clear that it's not longer supported.

wochinge:
@alwx Added a bit more code which now cleans up the pika connection when the server shuts down using the Sanic listener `after_server_stop`.

changelog/7263.removal.md
```diff
@@ -0,0 +1,29 @@
...
+ queues:
+ - my_queue
+
+```
+
+Support for `RABBITMQ_SSL_KEY_PASSWORD` was removed entirely.
```
wochinge, line 16: @erohmensing Short question. Do you know if this used by any customer? 😬

erohmensing: Oooh, my guess would be yes, but I can check. Is there still support for SSL for rabbit through a different setup or env var?

erohmensing: Or is it just the password that is not supported?

erohmensing: Sorry, have read the description now. We definitely need support for CA files as its common to use it with self-signed certs 😬

wochinge: it's just the password for the private key which is no longer supported.

rasa/core/brokers/broker.py
```diff
@@ -38,53 +55,62 @@ def is_ready(self) -> bool:
...
-        return event_broker_class.from_endpoint_config(broker_config)
+        if not asyncio.iscoroutinefunction(event_broker_class.from_endpoint_config):
+            rasa.shared.utils.io.raise_deprecation_warning(
+                f"The method "
+                f"'{EventBroker.__name__}.{EventBroker.from_endpoint_config.__name__}' "
+                f"was changed to be 'async'. Your event broker '{broker_config.type}' "
+                f"is currently synchronous. Support for synchronous implementations of "
+                f"'{EventBroker.from_endpoint_config.__name__}' will be removed in "
+                f"3.0.0."
+            )
```
alwx, line 111: 🚀

rasa/core/exporter.py
```diff
@@ -87,7 +87,17 @@ def publish_events(self) -> int:
...
                 raise PublishingError(current_timestamp)
 
-        self.event_broker.close()
+        if asyncio.iscoroutinefunction(self.event_broker.close):
+            rasa.shared.utils.io.raise_deprecation_warning(
+                f"The method '{EventBroker.__name__}.{EventBroker.close.__name__} was "
+                f"changed to be asynchronous. Please adapt your custom event broker "
+                f"accordingly. Support for synchronous implementations will be removed "
+                f"in Rasa Open Source 2.2.0."
+            )
```
alwx, line 96: Have we created issues (= reminders for ourselves) for that already?

alwx (approved):

rasa/core/exporter.py
wochinge: will do! Good catch!

Issue #7244, opened 2020-11-11
Invalid rule when training when setting a slot of type float


nbeuchat:
**Rasa version**: 2.0.3

**Python version**: 3.7.9

**Operating system**: Ubuntu 18.04

**Issue**:
I have two rules which start with the same action but have different behaviors depending on the value of the slot that has been set. The slot (`matches_found`) is of type `float` with a min of 0 and a max of 1. The rules fail to be validated with the following error (full error below): `the action 'action_search' in rule 'Search with results' does not set all the slots, that it sets in other rules: 'matches_found'.` 

If I use a different type of slot (a boolean slot for example), it does not throw an error. Even if I set `matches_found` to the exact same value in both rules, it still fails. 

**Rules**
```
version: "2.0"
rules:
- rule: Search with results
  steps:
  - action: action_search
  - slot_was_set:
    - matches_found: 1
  - action: action_display_apt_selector

- rule: Search without results
  steps:
  - action: action_search
  - slot_was_set:
    - matches_found: 0
  - action: action_no_search_results
```

**Error (including full traceback)**:
```
InvalidRule: 
Incomplete rules found🚨

- the action 'action_search' in rule 'Search with results' does not set all the slots, that it sets in other rules: 'matches_found'. Please update the rule with an appropriate slot or if it is the last action add 'wait_for_user_input: false' after this action.
- the action 'action_search' in rule 'Search without results' does not set all the slots, that it sets in other rules: 'matches_found'. Please update the rule with an appropriate slot or if it is the last action add 'wait_for_user_input: false' after this action.
Please note that if some slots or active loops should not be set during prediction you need to explicitly set them to 'null' in the rules.
You can find more information about the usage of rules at https://rasa.com/docs/rasa/rules. 
```

**Command or request that led to error**:
```
rasa train core
```

**Content of configuration file (config.yml)** (if relevant):
```yml
policies:
  - name: TEDPolicy
    max_history: 7
    epochs: 5
  - name: MemoizationPolicy
    max_history: 5
  - name: RulePolicy
```

**Content of domain file (domain.yml)** (if relevant):
```yml
slots:
  matches_found:
    max_value: 1
    min_value: 0
    type: float
```

davidhsv:
Same problem with type text

sara-tagger:
Thanks for the issue, @koernerfelicia will get back to you about it soon!

[...]

PR #7268, opened 2020-11-12
FloatSlot._as_feature should return unambiguous values

koernerfelicia:
Currently, if a float slot value is set to `0.0` and the min value is `0.0`, `FloatSlot._as_feature` will return `0.0`. However, this is also the default value returned when the value could not be converted. As a result, the slot is considered "unfilled" and isn't added:

Screenshot 2020-11-12 at 20 47 20

See user issue [here](https://github.com/RasaHQ/rasa/issues/7244)

**Proposed changes**:
- Make `FloatSlot` return two floats:
`[1.0, <value>]` if successfully converted
`[0.0, 0.0]` if not
This mimics the behaviour of `BooleanSlot`

- [X] added some tests for the functionality -> (updated existing tests)
- [X] updated the documentation -> (not really user-facing, the functionality should stay the same from the outside)


tmbo:
this is a model breaking change and needs to increase the minimum compatible model version

tmbo:
@dakshvar22 should we run the model evaluation tests on this?

tmbo (changes requested):
looks good, some minor remarks + breaking change as mentioned

dakshvar22:
@tmbo Model regression tests are only for NLU at the moment, so not needed here.

koernerfelicia:
>this is a model breaking change and needs to increase the minimum compatible model version
@tmbo: thank you for taking a look! What does the above mean in terms of action items? I'm guessing it has to be documented somewhere, just not sure where

tmbo:
your changelog item should include a warning, e.g. here is an example: https://rasa.com/docs/rasa/changelog#deprecations-and-removals

and you'll need to increase the version in https://github.com/RasaHQ/rasa/blob/master/rasa/constants.py#L27

koernerfelicia:
@tmbo sorry, could you take another look? I'm not sure why this is failing - I updated the rasa-sdk version by hand, maybe I shouldn't have done this?

koernerfelicia:
Hi @tmbo, I think this is finally ready. I'm sorry it got so messy and required so many revisions... I definitely know better now for the future.

tmbo:
don't worry, e.g. the poetry issues and needed updates aren't your fault - was actually good that we ran into that to fix it :+1:

tmbo (approved):
looks great :rocket:

Issue #6406, opened 2020-08-14
LexicalSyntacticFeaturizer computes unnecessary operations !


sahinbatmaz:
**Description of Problem**:

LexicalSyntacticFeaturizer computes unnecessary operations when creating features.

[https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py#L184](https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py#L184)

In the '_tokens_to_features' function, tokens are converted to appropriate features. However, when a window of features is given and there are common feature names among window elements, the same function is applied on the same token unnecessarily. The function is applied in the following line.

[https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py#L268](https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py#L268)

 value = self.function_dict<span class="error">[feature]</span>(token)

**Overview of the Solution**: In each sentence, keep a dictionary for tokens. If a function is applied on a token, keep the result in a dictionary. Probably using dictionary keys as a tuple of token text and given function would work. And the value will be the result.

Actually, there should not be a dictionary for each sentence. There could be one dictionary for all training data, since result will be same when the token text and the function are given.

This solution is somehow similar to Dynamic Programming in terms of keeping the result, and using it over and over when needed.

**Examples** : In my configuration file, I am a little bit greedy about this layer. My configuration looks like the following. The window size is 7 in the configuration.

*   name: LexicalSyntacticFeaturizer
     "features": [
     <span class="error">["low", "prefix5", "prefix2", "suffix5", "suffix3", "suffix2", "suffix1", "upper", "title", "digit",]</span>,
     <span class="error">["low", "prefix5", "prefix2", "suffix5", "suffix3", "suffix2", "suffix1", "upper", "title", "digit",]</span>,
     <span class="error">["low", "prefix5", "prefix2", "suffix5", "suffix3", "suffix2", "suffix1", "upper", "title", "digit",]</span>,
     <span class="error">["low", "prefix5", "prefix2", "suffix5", "suffix3", "suffix2", "suffix1", "upper", "title", "digit",]</span>,
[...]

PR #7280, opened 2020-11-13
Speed up LexicalSyntacticFeaturizer when using window features

alopez:
**Proposed changes**:
- Memoizes feature computations when window features are used are used in LexicalSyntacticFeaturizer. This results in some minor speedups since features are no longer computed repeatedly for each token., as noted by https://github.com/RasaHQ/rasa/issues/6406. It doesn't completely solve their problem, since more features still require more time to compute (e.g. since there are larger arrays in _features_to_onehot).

### DIFF
--- a/rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
+++ b/rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
@@ -189,14 +189,34 @@ class LexicalSyntacticFeaturizer(SparseFeaturizer):
         configured_features = self.component_config["features"]
         sentence_features = []
 
-        for token_idx in range(len(tokens)):
+        # A function can be applied to the same token multiple times if
+        # if configured_features includes functions of the window around a token.
+        # This can become unnecessarily slow if the window size is large.
+        # For efficiency, we memoize the function call on each token.
+        functions = set(
+            [
+                function
+                for token_functions in configured_features
+                for function in token_functions
+            ]
+        )
+        token_to_function_values = []
+        for token_idx, token in enumerate(tokens):
+            function_values = {}
+            for function in functions:
+                function_values[function] = self._get_feature_value(
+                    function, token, token_idx, len(tokens)
+                )
+            token_to_function_values.append(function_values)
+
+        for token_idx, token in enumerate(tokens):
             # get the window size (e.g. before, word, after) of the configured features
             # in case of an even number we will look at one more word before,
             # e.g. window size 4 will result in a window range of
             # [-2, -1, 0, 1] (0 = current word in sentence)
             window_size = len(configured_features)
             half_window_size = window_size // 2
-            window_range = range(-half_window_size, half_window_size + window_size % 2)
+            window_range = [i - half_window_size for i in range(window_size)]
 
             prefixes = [str(i) for i in window_range]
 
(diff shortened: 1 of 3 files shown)


tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
```diff
@@ -195,3 +195,93 @@ def test_text_featurizer_using_pos_with_action_text(
...
+                    "pos",
+                    "upper",
+                    "digit",
+                ],
+                ["prefix2", "suffix1", "suffix3", "pos", "upper"],
+            ],
+        )
+    ],
+)
+def test_text_featurizer_using_feature_to_idx_dict(
```
dakshvar22, line 222: Can we break this test into multiple tests since it's testing too many things at once IMO?
If we want to test the memoization implementation I suggest we write a unit test just for that. To do that, it would be better if we extract out the memoization code lines 192-212 in `_tokens_to_features` to a separate function so that we can test it separately.

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
```diff
@@ -195,3 +195,93 @@ def test_text_featurizer_using_pos_with_action_text(
...
+                        featurizer.function_dict[feature_function](tokens[position])
+                    )
+                    feature_idx = featurizer.feature_to_idx_dict[feature_name][
+                        feature_value_str
+                    ]
+                    assert token_features[feature_idx] == 1.0
+                    token_features[feature_idx] = 0.0
+
+        # Since all expected features were zeroed out, result should be all zeros
+        assert not np.any(token_features)
```
dakshvar22, line 287: I didn't fully understand what this test is doing but I think this should ideally be a separate test. Also, there is a lot of "logic" inside the test on what the code should be doing and it's a bit hard to follow. Are we testing a particular edge case here?

dakshvar22 (changes requested):
Implementation is really clean! A few suggestions on how we could restructure it a bit further.

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
alopez: @dakshvar22 There aren't any direct tests of the memoization: if memoization is implemented correctly, then the behavior of the feature is unchanged, but it's faster. The existing tests check behavior; I verified the speedup with cProfile. 

So technically the test is unrelated to the change. However, it exercises more parts of the code than the other tests and doesn't rely on any expected outputs, as in the other tests. By specifying those outputs, the other tests create undocumented (afaict) requirements on the order of the features, overconstraining the implementation. This test does not have any expected output. Rather, it checks whether the features are internally consistent, irrespective of how the features are ordered.

alopez: I guess then the question is whether this test should be separated out or removed altogether?

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
alopez: Just so I understand: do you mean that this should be separated out from the rest of the test?

In a nutshell, the this test is checking that the features returned for the test input are 1 if and only if they were features requested in `featurizer_config`. Rather than use a hardcoded expected features (see comment above), it does this by indexing the features by name via `feature_to_idx_dict`. This is arguably a whitebox test, but afaict `feature_to_idx_dict` is public so it seems ok to test it.

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -189,14 +189,34 @@ def _tokens_to_features(self, tokens: List[Token]) -> List[Dict[Text, Any]]:
...
+            ]
+        )
+        token_to_function_values = []
+        for token_idx, token in enumerate(tokens):
+            function_values = {}
+            for function in functions:
+                function_values[function] = self._get_feature_value(
+                    function, token, token_idx, len(tokens)
+                )
+            token_to_function_values.append(function_values)
```
dakshvar22, line 210: I think it would also help to move this memoization bit to an independent function. Would be easier to follow IMO

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
dakshvar22: As I said above, having this as a complete test is still useful, so let's keep it.

PR #7284, opened 2020-11-16
Poc bandit

m-vdb:
**Proposed changes**:
- Relates to https://github.com/RasaHQ/rasa-x/issues/3567
- Setup [Bandit](https://bandit.readthedocs.io/en/latest/) to detect python security issues
- If we're happy with it, we can make it as required in the Branch Protection settings
- (this replaces some of DeepSource checks)


Makefile
```diff
@@ -79,6 +79,10 @@ endif
 	# Diff of uncommitted changes for running locally
 	git diff HEAD -- rasa | poetry run flake8 --select D --diff
 
+lint-security:
+	poetry run pip install bandit\>=1.6.2,\<2.0
```
m-vdb, line 83: it's quicker to install it this way in CI (you do not have to run `make install`) but it's harder to maintain. Can we specify this requirement inside `pyproject.toml`?

rasa/core/tracker_store.py
```diff
@@ -887,7 +887,9 @@ def _create_database(engine: "Engine", db: Text):
 
         cursor = conn.connection.cursor()
         cursor.execute("COMMIT")
-        cursor.execute(f"SELECT 1 FROM pg_catalog.pg_database WHERE datname = '{db}'")
+        cursor.execute(
+            f"SELECT 1 FROM pg_catalog.pg_database WHERE datname = ?", (db,)  # nosec
+        )
```
m-vdb, line 892: the SQL injection is now fixed, but `bandit` still detects this as an issue, so I'm ignoring it with `# nosec`

Makefile
joejuzl: hmm could we use something like https://python-poetry.org/docs/pyproject/#extras for this?

rasa/core/tracker_store.py
joejuzl: what does it complain about?

.github/workflows/security-scans.yml
```diff
@@ -27,3 +42,35 @@ jobs:
...
       uses: zricethezav/gitleaks-action@v1.1.4
+
+  bandit:
+    name: Detect python security issues
+    runs-on: ubuntu-latest
+    needs: [changes]
+
+    steps:
+    - name: Checkout git repository 🕝
+      if: needs.changes.outputs.backend == 'true'
```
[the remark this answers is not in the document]
m-vdb: no not really, because the workflow are required to pass by the branch protection mechanism. We cannot skip the workflows altogether, so we have to live with this today 😄

Makefile
```diff
@@ -79,6 +79,10 @@ endif
 	# Diff of uncommitted changes for running locally
 	git diff HEAD -- rasa | poetry run flake8 --select D --diff
 
+lint-security:
```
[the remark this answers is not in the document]
m-vdb: not really, but I wanted to avoid naming it `run-bandit`, in case we change the tool itself one day 🤷🏻‍♂️

Makefile
m-vdb: no not really, because when you run `poetry install --extras stuff` it installs all the dependencies + the extra ones, which in our case doesn't help 😕

rasa/core/tracker_store.py
m-vdb: it throws [this error](https://bandit.readthedocs.io/en/latest/plugins/b608_hardcoded_sql_expressions.html). It did the same before I changed the line.

Makefile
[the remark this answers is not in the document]
m-vdb: My reasoning for running it separately is that we'd want to run it in another workflow with other security checks (cf. the GH workflow file). But if we want to run it with the rest of the CI checks, we could merge it with `make lint`.

I'd be cautious about merging them though because:
- if we end up merging checks, we do not achieve parallelism
- it'd make it hard for developers to see all the failing checks at once and they'd have to wait for other CI runs to see all errors (for instance: python lint + security lint both fail. You fix first the python lint issue, push your changes, **wait for another run**, and then realise the security lint fails too, etc...)

Makefile
joejuzl: aha I see, that's annoying

rasa/core/tracker_store.py
joejuzl: We could use https://sqlalchemy-utils.readthedocs.io/en/latest/database_helpers.html#database-exists

Makefile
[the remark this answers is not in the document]
joejuzl: Yup definitely agree with having them split in the CI. We could have an overriding `make` command for running all of them locally though. Like `make static-checks` or something.

rasa/core/tracker_store.py
m-vdb: 👍🏻 yeah I found out about this one as well. Not sure it makes sense to install the extra dependency to only replace this function call. Interested to see if there other use cases hidden in the code somewhere 🤔

Makefile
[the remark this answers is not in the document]
m-vdb: Ah yeah that's true, so it'd be only a developer tool 🤔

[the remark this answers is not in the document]
joejuzl: Yeah - I find it useful to have one command I can run to ensure that I won't break any of the non-test checks in the CI.

rasa/core/tracker_store.py
joejuzl: it would also mean we could replace the line `cursor.execute(f"CREATE DATABASE {db}")` as it has functionality for creating databases. Potentially also useful for this upcoming work: https://github.com/RasaHQ/rasa-x/issues/4027.

m-vdb: that's a good point for Rasa X, but maybe not for Rasa: the only place where we create a database in rasa is in the Tracker, and it already uses postgres. I'm a bit reluctant to add a 140kb dependency only for getting rid of this issue. WDYT?

Makefile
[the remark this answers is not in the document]
m-vdb: added ✅

rasa/core/tracker_store.py
joejuzl: I would usually choose having an extra dependency over having to do something manually using SQL. But in this case I don't feel super strongly, and this is the only place like you say - so your call.

joejuzl (approved):

rasa/core/tracker_store.py
m-vdb: waiting on @wochinge input on this too, as he might have another perspective :)

tmbo:
@m-vdb should we get this merged or is there something we need to test first?

m-vdb:
I wanted to get input from @wochinge as well. I know he didn't have the time last week, hopefully he's able to do a review this week 🤞🏻

Makefile
wochinge: why not having it in the `dev-dependencies`?

bandit.yml
```diff
@@ -0,0 +1 @@
+skips: ['B322', 'B104', 'B301', 'B303']
```
wochinge, line 1: can we add comments why we are skipping these?

rasa/core/tracker_store.py
wochinge: > the SQL injection is now fixed, but bandit still detects this as an issue, so I'm ignoring it with # nosec

Is this maybe due to the `1`? 

I am against the lib as the lib actually reintroduces the security issue: https://github.com/kvesteri/sqlalchemy-utils/blob/2e8ee0093f4a33a5c7479bc9aaf16d7863a74a16/sqlalchemy_utils/functions/database.py#L465

wochinge (approved):
Sorry for the late review 😬 

Thanks for taking this on 🚀

Makefile
m-vdb: that was also a possibility - do you think it would be useful to run it locally? We'd be sacrificing some performances in the CI but I think we can live with that if it's actually useful to have it locally.

rasa/core/tracker_store.py
m-vdb: good, good catch for the lib 🤦🏻 

regarding the 1: I don't think it is the reason, but I'll try to remove it to make sure

m-vdb: so it wasn't that, but what's weird is that now it works without the `# nosec` comment. Not sure what happened there, I must have made a mistake while testing it 🤷🏻 so all is good 👍🏻

wochinge: Nice 💯

bandit.yml
```diff
@@ -0,0 +1,5 @@
+# B322: checks for `input()` which is unsafe in Python 2, but safe in Python 3.
+# B104: checks for binding 0.0.0.0 interface, which should be fine for containers.
+# B301: checks for pickle usage, which is a necessary evil.
+# B303: checks for insecure hash functions like md5, which is good enough for current use cases.
```
m-vdb, line 4: @wochinge @joejuzl I'm having a doubt about this one. I think that the current use we make of sha1/md5 is fine today, but we cannot guarantee that the future uses are fine as well. It might make sense to not skip it, and locally disable the checks by adding `# nosec` comments. What do you think?

joejuzl: Yeah I agree - safer to skip it on a case-by-case basis where we know it is not being used for security purposes.

Makefile
wochinge: Ideally it's restored from cache anyway so I don't see the performance impact on the CI as critical.
We run all other linters locally, so I think this one should be set up similarly.

m-vdb: ok gotcha, will go with your reco then. I don't know why exactly I thought I could make an exception of it to the detriment of easy dep management...

Issue #7253, opened 2020-11-12
Rule with `slot_was_set` doesn't work as expected


degiz:
**Rasa version**: 2.0.6

**Issue**:

`slot_was_set` condition doesn't work with a Rule below:
```yaml
rules:
- rule: Only say `hello` if the user provided a name
  condition:
  - slot_was_set:
    - some_text_slot
```

**Expected behavior**: Rule is applied when the `some_text_slot` was set to a not-null value.

**Actual behavior**: Rule is applied when the slot is not filled.

davidhsv:
Ideas for testing this issue:

**Test1:**
```
rules:
  - rule: Only say `hello` if the user provided a name
    condition:
    - slot_was_set:
      - some_text_slot
    - steps:
      - intent: greet_user
  - rule: Say `Hi strange` if the user does not provided a name
    condition:
    - slot_was_set:
      - some_text_slot: null
    - steps:
      - intent: greet_user_anon
```
**Test2:** 
```
rules:
  - rule: Say `foo` if the user provided a name 'foo'
    condition:
    - slot_was_set:
      - name: 'foo'
    - steps:
      - intent: greet_user_foo
  - rule: Say `bar` if the user provided a name 'bar'
    condition:
    - slot_was_set:
      - name: 'bar'
    - steps:
      - intent: greet_user_bar
  - rule: Say `bar` if the user provided a name
    condition:
      - slot_was_set:
        - some_text_slot: null
      - steps:
        - intent: greet_user_generic
  - rule: Say `bar` if the user does not provided a name
    condition:
      - slot_was_set:
        - some_text_slot: null
      - steps:
        - intent: greet_user_anon
```

nbeuchat:
Maybe related to this issue that I was facing with rules validation for `float` slots? https://github.com/RasaHQ/rasa/issues/7244

nbeuchat:
By the way, I don't think your rules are expected to work for a `text` slot. A `text` slot value does not matter, it only matters if the text is filled or not. If the value should matter (see: https://rasa.com/docs/rasa/domain#slot-types). You should use a categorical slot for that: https://rasa.com/docs/rasa/domain#categorical-slot

davidhsv:
@nbeuchat I don't agree. I think you should be able to make a rule with a condition for a specific string. That why the rules exists, to make a very specific **programmatic** rule before any NLP policy.

[...]

PR #7288, opened 2020-11-16
Add default values for `text` and `bool` slots if referenced by name

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7253
- Background: For `text` and `list` slots it only matters whether they are set or not. Their actual values doesn't matter. In case they were referenced by name only, we can fill in a value, which will featurize them as `set`. This doesn't work for something like `float` or `categorical` slots as in that case the value matters, and we don't know the value. 
- Raise warning if a slot was referenced by name which needs to have a value specified.
- clarify usage of `boolean` slots as it's different if they are empty, `True`, or `False`


wochinge:
@akelad @erohmensing Does it make sense this way? I generally find referencing slots by names only a bit error prone.

melindaloubser1:
@wochinge `text` and `list` slots are featurized as set or not - `bool` is featurized differently for `true` and `false`. Empty is featurized as `false`. I think it would be better to not allow listing `bool` slots with name only, although technically it is pointless to list it if it isn't supposed to be `true`.

wochinge:
@melindaloubser1 Adapted it accordingly 👍

akelad:
the bool slot featurization for empty is different to `false` just fyi: https://github.com/RasaHQ/rasa/blob/master/rasa/shared/core/slots.py#L197

i think:

empty = [0,0]
False = [1,0]
True = [1,1]

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -436,6 +441,26 @@ def _parse_slot(self, step: Dict[Text, Any]) -> None:
...
+
+        default_value = slot_types_with_default_types.get(type(slot))
+        if default_value is None and slot.has_features():
+            rasa.shared.utils.io.raise_warning(
+                f"Slot '{slot_name}' was referenced by its name only. As slot "
+                f"'{slot_name}' is of type '{slot.type_name}' you need to specify a "
+                f"value for it. Slot '{slot_name}' will be treated as if it's value "
+                f"is empty.",
+                docs=DOCS_URL_SLOTS,
+            )
```
melindaloubser1, line 460: Should this be an error instead of a warning?

melindaloubser1:
@akelad In that case we need to update the [docs](https://rasa.com/docs/rasa/domain#boolean-slot) 😬

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
wochinge: All the other stuff is warnings, that's why I chose `raise_warning` to be consistent.

wochinge:
> @akelad In that case we need to update the docs 😬

Done!

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
melindaloubser1: 👍

melindaloubser1 (approved):

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -46,6 +49,8 @@
 
 
 CORE_SCHEMA_FILE = "utils/schemas/stories.yml"
+DEFAULT_VALUE_TEXT_SLOTS = "filled"
```
joejuzl, line 52: So, these two stories would still be equivalent?

```yaml
stories:
- story: Pasta 1
  steps:
  - intent: order_pasta
  - slot_was_set:
    - pasta_type
- story: Pasta 2
  steps:
  - intent: order_pasta
  - slot_was_set:
    - pasta_type: "filled"
```
😬 😬 😬

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -425,7 +430,9 @@ def _parse_slot(self, step: Dict[Text, Any]) -> None:
                 for key, value in slot.items():
                     self._add_event(SlotSet.type_name, {key: value})
             elif isinstance(slot, str):
-                self._add_event(SlotSet.type_name, {slot: None})
+                self._add_event(
+                    SlotSet.type_name, {slot: self._slot_default_value(slot)}
```
joejuzl, line 434: So this is added every time, even if the `slot` has a value?

wochinge: If the slot is referenced by name only, it can't have a value at the same time. If `isinstance(slot, dict)` then we take the given value

joejuzl: Gotcha 👍

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
wochinge: Yes 👍  It doesn't matter for `text` slots as the ML model only looks **if** the slot was filled. So the actual text doesn't matter at all.

joejuzl: Aha ok - it never looks at the value... Makes sense! (I missed that in the description 🤦 )

wochinge: Yeah. Not a huge fan of this "reference by name" only thing 🤷‍♂️

PR #7292, opened 2020-11-17
Iterate on Sentry filtering

m-vdb:
**Proposed changes**:
- Ignore pymongo connection errors ([sentry/1939788804](https://sentry.io/organizations/rasahq/issues/1939788804/?project=2801673))
- Raise `MarkdownException` when Markdown cannot be read ([sentry/1953580685](https://sentry.io/organizations/rasahq/issues/1953580685/?project=2801673))
- Ignore `CancelledError` ([sentry/1966753788](https://sentry.io/organizations/rasahq/issues/1966753788/?project=2801673))
- Do not load endpoint config if it is `None` ([sentry/1946541728](https://sentry.io/organizations/rasahq/issues/1946541728/?project=2801673))
- Raise `InvalidEntityFormatException` when entities cannot be parsed, and ignore it ([sentry/1945365040](https://sentry.io/organizations/rasahq/issues/1945365040/?project=2801673))
- Ignore `MemoryError` ([sentry/2009040566](https://sentry.io/organizations/rasahq/issues/2009040566/?project=2801673))
- Ignore `psycopg2.OperationalError` ([sentry/2009384160](https://sentry.io/organizations/rasahq/issues/2009384160/?project=2801673))
- Ignore `botocore.NoCrendentialsError` ([sentry/2005727807](https://sentry.io/organizations/rasahq/issues/2005727807/?project=2801673))
- Ensure classes using `class_from_module_path()` are in fact classes ([sentry/1952538184/071b08fc49494d3b96c16d0ed19b9913](https://sentry.io/organizations/rasahq/issues/1952538184/events/071b08fc49494d3b96c16d0ed19b9913/?project=2801673) ⚠️ there are various issues under this sentry issue, I'm addressing one type)
- Ignore Pika connection errors ([sentry/2071539640](https://sentry.io/organizations/rasahq/issues/2071539640/?project=2801673))


m-vdb:
There are some issues coming up today into Sentry ([1958583765](https://sentry.io/organizations/rasahq/issues/1958583765/?project=2801673), [1952538184](https://sentry.io/organizations/rasahq/issues/1952538184/?project=2801673), [1962256159](https://sentry.io/organizations/rasahq/issues/1962256159/?project=2801673), [1951854941](https://sentry.io/organizations/rasahq/issues/1951854941/?project=2801673), [1956288097](https://sentry.io/organizations/rasahq/issues/1956288097/?project=2801673), [1946342362](https://sentry.io/organizations/rasahq/issues/1946342362/?project=2801673)) that come from schema validation issues of config / domain files. While exploring solutions, I've hit a roadblock with `pykwalify` (the lib we use to validate yaml against a schema), especially on the slot types definition: with this lib, it's not possible to validate an object against an union of schema ("any of" use case). I don't think it's the main limitation that we have today, would be happy to hear some feedback on this point.

To me the impact is at different levels:
- At the user level: the code fails with an exception, not telling the user what the problem is. They might even think that rasa is broken
- At the rasa engineer level: hard to pin-point the initial issue when supporting a user on the forum, noise in Sentry.

rasa/telemetry.py
```diff
@@ -608,6 +608,28 @@ def initialize_error_reporting() -> None:
...
+        ignored_errors.append(PyMongoError)
+    except ImportError:
+        pass
+
+    try:
+        from psycopg2 import OperationalError
+
+        ignored_errors.append(OperationalError)
+    except ImportError:
+        pass
```
m-vdb, line 631: it might be better to define these closer to the classes that rely on pymongo / postgres 🤔

rasa/telemetry.py
```diff
@@ -631,7 +634,19 @@ def initialize_error_reporting() -> None:
...
-        ignore_errors=[KeyboardInterrupt, RasaException, NotImplementedError],
+        ignore_errors=[
+            # std lib errors
+            KeyboardInterrupt,
+            MemoryError,
+            NotImplementedError,
+            asyncio.CancelledError,
+            # rasa errors
+            RasaException,
+            # 3rd party errors
```
m-vdb, line 645: it might be better to define these closer to the classes that rely on pymongo / postgres / boto 🤔 that would make it easier to maintain (e.g. closer to channels, tracker store, etc...).

rasa/telemetry.py
```diff
@@ -631,7 +634,19 @@ def initialize_error_reporting() -> None:
...
+            botocore.exceptions.NoCredentialsError,
```
m-vdb, line 646: I think we should ignore more connection errors too (see [all the errors here](https://github.com/boto/botocore/blob/develop/botocore/exceptions.py))

rasa/shared/utils/common.py
```diff
@@ -1,35 +1,44 @@
...
             m = importlib.import_module(lookup_path)
-            return getattr(m, module_path)
-        else:
-            raise ImportError(f"Cannot retrieve class from path {module_path}.")
+            klass = getattr(m, module_path)
+
+    if klass is None:
+        raise ImportError(f"Cannot retrieve class from path {module_path}.")
+
+    if ensure_class and not inspect.isclass(klass):
```
m-vdb, line 37: I don't know enough of the usage of this function, but the name `class_from_module_path ()` suggests that it should always be returning a class, whereas the code at line 25 suggests otherwise. Would need feedback on this

if it's expected to return a class all the time, i'd make `ensure_class` the default (effectively removing it from the func signature), and raise a deprecation warning.

(the code I wrote for this one is pretty basic, mainly because I wanted to show a draft version earlier)

m-vdb:
@tmbo could I get your feedback on this one? Nothing urgent and it should be quick for you I think ^^ (tests + changelog fragments are obviously missing for the draft)

rasa/shared/utils/common.py
```diff
@@ -1,35 +1,44 @@
...
+        if klass is None and lookup_path:
             # last resort: try to import the class from the lookup path
             m = importlib.import_module(lookup_path)
-            return getattr(m, module_path)
-        else:
-            raise ImportError(f"Cannot retrieve class from path {module_path}.")
+            klass = getattr(m, module_path)
+
+    if klass is None:
+        raise ImportError(f"Cannot retrieve class from path {module_path}.")
```
tmbo, line 35: this will get logged to sentry though, right?

rasa/shared/utils/common.py
tmbo: I do think it needs to always return a class.

The condition in L19 is for the two different cases `mypackage.MyModel` and `MyModel` as they need to be loaded in different ways. In the case of the later (handled in the `else` part) the module is already in scope and hence needs to be retrieved from there (this is the `globals` and `locals` part) - this should have gotten a link to a stackoverflow post where this likely has its origin 😅

rasa/shared/utils/common.py
```diff
@@ -1,35 +1,44 @@
...
+
+    if ensure_class and not inspect.isclass(klass):
+        # FIXME: surely another error more adapted
+        raise TypeError(f"Object at {module_path} is not a valid class.")
```
tmbo, line 39: should this be a RasaException as well (and hence be ignored by sentry)?

rasa/telemetry.py
tmbo: Maybe there is a better approach: I am not sure raising these original exceptions directly is the best approach anyways. e.g. it might be better to define a `ConnectionException` error pointing to more information on where to fix an issue and use the original exception as a cause of this one (`raise ConnectionException(...) from NoCredentialsError`)

tmbo (commented):
thanks a lot for digging into these issues 🚀  left some thoughts

rasa/shared/utils/common.py
m-vdb: yes that was the plan, wanted to get feedback first 👍🏻 💯

rasa/shared/utils/common.py
m-vdb: yes 👍🏻

rasa/core/brokers/broker.py
```diff
@@ -3,13 +3,20 @@
...
+
 import rasa.shared.utils.common
 import rasa.shared.utils.io
+from rasa.shared.exceptions import ConnectionException
 from rasa.utils.endpoints import EndpointConfig
 
 logger = logging.getLogger(__name__)
 
 
+CONNECTION_ERRORS = (psycopg2.OperationalError, aio_pika.exceptions.AMQPConnectionError)
```
m-vdb, line 17: questions for this constant and the other one below:
- do we care about importing modules like `psycopg2`, `pymongo`, etc... at the top-level? Is it impacting the start-up time of the CLI and do we have evidence of that?
- would it make sense to have only one shared `CONNECTION_ERRORS` tuple somewhere in `rasa.shared.constants` or `rasa.constants`? I'm sitting between that and defining the constant closer to the the code that needs it.

rasa/core/tracker_store.py
```diff
@@ -63,9 +66,15 @@
 # default value for key prefix in RedisTrackerStore
 DEFAULT_REDIS_TRACKER_STORE_KEY_PREFIX = "tracker:"
 
+CONNECTION_ERRORS = (
+    NoCredentialsError,
```
m-vdb, line 70: it might make sense to ignore more errors from `botocore` (see [all the errors here](https://github.com/boto/botocore/blob/develop/botocore/exceptions.py))

tests/shared/utils/test_common.py
```diff
@@ -140,3 +141,44 @@ async def my_function():
...
 
         await my_function()
+
+
+@pytest.mark.parametrize(
+    "module_path, lookup_path, outcome",
+    [
+        ("rasa.shared.core.domain.Domain", None, "Domain"),
+        # lookup in locals()
+        ("Domain", None, "Domain"),
```
m-vdb, line 151: this test fails. I am not sure how the function is supposed to work with `globals()` (cf. [docs](https://docs.python.org/3/library/functions.html#globals)). I think we assumed it did something, and actually it doesn't work.

m-vdb: or do we expect that the `globals()` are updated after the function is called successfully, making the subsequent calls work? 🤷🏻

wochinge:
@m-vdb I won't get to this before my vacation. Could you please request somebody else for review?

m-vdb:
@wochinge it can wait. If @joejuzl you have time next week, feel free to take a look, but it's not urgent.

tests/shared/utils/test_common.py
```diff
@@ -140,3 +142,47 @@ async def my_function():
...
+        # import errors below
+        ("rasa.shared.core.domain.FunkyDomain", None, None),
+        ("FunkyDomain", None, None),
+        ("FunkyDomain", "rasa.shared.core.domain", None),
+    ],
+)
+def test_class_from_module_path(
+    module_path: Text, lookup_path: Optional[Text], outcome: Optional[Text]
+):
+    if outcome is None:
```
tmbo, line 164: should we just create two different tests instead of having this if?

tests/shared/utils/test_common.py
tmbo: when you are saying "this test fails" do you mean for the parameters in the line we are talking about here or do all of these fail?

m-vdb: I'm talking about these specific parameters. All the other tests pass

data/test_endpoints/event_brokers/connexion_error_pika_endpoint.yml
```diff
@@ -0,0 +1,9 @@
+event_broker:
```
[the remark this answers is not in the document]
wochinge: actually in all new `test_endpoints` files

rasa/core/brokers/broker.py
```diff
@@ -3,13 +3,25 @@
...
 import rasa.shared.utils.common
 import rasa.shared.utils.io
+from rasa.shared.exceptions import ConnectionException
 from rasa.utils.endpoints import EndpointConfig
 
 logger = logging.getLogger(__name__)
 
 
+CONNECTION_ERRORS = (
+    psycopg2.OperationalError,
```
wochinge, line 19: there is no event broker using`Postgres` afaik? What about the Kafka broker?

rasa/core/tracker_store.py
```diff
@@ -63,9 +67,16 @@
...
+    psycopg2.OperationalError,
```
wochinge, line 72: is there a generic SQLAlchemy error which covers this one as well (e.g. sqlachemy.exc.OperationalError). I'm brining this up as this wouldn't work for other types of tracker stores or other SQL dialects (e.g. oracle)

rasa/core/tracker_store.py
```diff
@@ -63,9 +67,16 @@
...
+    pymongo.errors.ConnectionFailure,
+    sqlalchemy.exc.OperationalError,
+)
+
 
 class TrackerStore:
-    """Class to hold all of the TrackerStore classes"""
+    """Class to hold all of the TrackerStore classes."""
```
wochinge, line 79: how about this:
```suggestion
    """Represents common behavior and interface for all `TrackerStore`s."""
```

rasa/shared/exceptions.py
```diff
@@ -77,3 +81,15 @@ class InvalidConfigException(ValueError, RasaException):
...
+
+class InvalidEntityFormatException(RasaException):
+    """Raised if the format of an entity is invalid."""
+
+
+class ConnectionException(RasaException):
+    """Raised when a connection to a 3rd party service fails.
+
+    It's used by our broker and tracker store classes, when
+    they can't connect to services like postgres, s3, mongo.
```
wochinge, line 94: S3 is nor a broker nor a tracker store.

rasa/shared/nlu/training_data/entities_parser.py
```diff
@@ -141,12 +142,10 @@ def get_validated_dict(json_str: Text) -> Dict[Text, Text]:
     try:
         data = json.loads(f"{{{json_str}}}")
     except JSONDecodeError as e:
-        rasa.shared.utils.io.raise_warning(
+        raise InvalidEntityFormatException(
```
wochinge, line 145: can you please include that in the docstring?

rasa/shared/nlu/training_data/entities_parser.py
```diff
@@ -141,12 +142,10 @@ def get_validated_dict(json_str: Text) -> Dict[Text, Text]:
...
             f"Incorrect training data format ('{{{json_str}}}'). Make sure your "
-            f"data is valid.",
-            docs=DOCS_URL_TRAINING_DATA_NLU,
-        )
-        raise e
+            f"data is valid. More info at {DOCS_URL_TRAINING_DATA_NLU}"
+        ) from e
 
     validation_utils.validate_training_data(data, schema.entity_dict_schema())
```
wochinge, line 150: what about the `ValidationError` here? Should in my opinion also raise a `InvalidEntityFormatException`

rasa/shared/utils/common.py
```diff
@@ -5,32 +5,59 @@
 import logging
 from typing import Text, Dict, Optional, Any, List, Callable, Collection
 
+import rasa.shared.utils.io as io_utils
```
wochinge, line 8: please use the complete module path (python code convention)

rasa/shared/utils/common.py
```diff
@@ -5,32 +5,59 @@
...
-    # load the module, will raise ImportError if module cannot be loaded
+    The loaded class can be used to instantiate new objects.
+
+    Args:
+        module_path: either an absolute path to a Python class,
+                     or the name of the class in the local / global scope.
+        lookup_path: a path where to load the class from, if it cannot
+                     be found in the local / global scope.
+        ensure_class: a boolean that will ensure that the loaded Python class
+                      is actually class if True (by default).
```
wochinge, line 28: In my opinion it's clear from the method signature that it's the default. Keeping information about the default in two places seems error prone to me.
```suggestion
                      is actually class if `True`.
```

rasa/shared/utils/common.py
```diff
@@ -5,32 +5,59 @@
...
-            raise ImportError(f"Cannot retrieve class from path {module_path}.")
+            klass = getattr(m, module_path, None)
+
+    if klass is None:
+        raise ImportError(f"Cannot retrieve class from path {module_path}.")
+
+    if ensure_class and not inspect.isclass(klass):
+        io_utils.raise_deprecation_warning(
+            f"`class_from_module_path()` is expected to return a class, "
+            f"but {module_path} is not one. "
```
wochinge, line 55: what do you think? 
```suggestion
            f"but {module_path} is of type '{type(klass)}'. "
```

rasa/telemetry.py
```diff
@@ -631,7 +631,15 @@ def initialize_error_reporting() -> None:
         ],
         send_default_pii=False,  # activate PII filter
         server_name=telemetry_id or "UNKNOWN",
-        ignore_errors=[KeyboardInterrupt, RasaException, NotImplementedError],
+        ignore_errors=[
+            # std lib errors
```
wochinge, line 635: how about adding explanations when which of these can happen?

tests/core/test_broker.py
```diff
@@ -301,3 +303,26 @@ def test_warning_if_unsupported_ssl_env_variables(monkeypatch: MonkeyPatch):
...
         pika._create_rabbitmq_ssl_options()
+
+
+async def test_pika_connection_error(monkeypatch: MonkeyPatch):
+    # patch PikaEventBroker to raise an AMQP connection error
+    async def connect(self) -> None:
+        raise aio_pika.exceptions.ProbableAuthenticationError("Oups")
+
+    monkeypatch.setattr(PikaEventBroker, "connect", connect)
+    cfg = read_endpoint_config(
```
wochinge, line 314: optional: I personally find it easier if all test data is directly in the test. Especially if it isn't re-used. This makes it easier to read the test as I don't have to jump to the file and also avoids us having thousands of files with often only minor differences.

tests/shared/nlu/training_data/test_entities_parser.py
```diff
@@ -137,3 +138,12 @@ def test_parse_training_example_with_entities():
     assert message.get("entities") == [
         {"start": 10, "end": 16, "value": "Berlin", "entity": "city"}
     ]
+
+
+def test_markdown_entity_regex_error_handling():
```
wochinge, line 143: there was no test for the error case? 🤯 Can we also add a test where it's a valid json but doesn't match the expected schema?

tests/shared/utils/test_common.py
wochinge: +1

tests/utils/test_endpoints.py
```diff
@@ -120,3 +122,21 @@ async def test_request_non_json_response():
...
+    "filename, endpoint_type, expected",
+    [
+        ("", "tracker_store", False),
+        (DEFAULT_ENDPOINTS_FILE, "stuff", False,),
+        (DEFAULT_ENDPOINTS_FILE, "empty", False,),
+        ("/unknown/path.yml", "tracker_store", False,),
+        (DEFAULT_ENDPOINTS_FILE, "tracker_store", True,),
+    ],
+)
+def test_read_endpoint_config(filename: Text, endpoint_type: Text, expected: bool):
```
wochinge, line 137: optional: I'ld vote for splitting this in two test cases

rasa/core/brokers/broker.py
m-vdb: - what about `SQLEventBroker` that can be instantiated with `postgres` dialect ([people are using it](https://sentry.io/organizations/rasahq/issues/2009384160/?project=2801673))?
- Didn't find enough kafka related errors to ignore them here. Which errors do you suggest we catch from [this list](https://github.com/dpkp/kafka-python/blob/master/kafka/errors.py)? Not sure if we should catch the base `KafkaError`, it could hide errors from us.

rasa/core/tracker_store.py
m-vdb: actually, I think we can remove this, as SQLAlchemy wraps exceptions and re-raise using their own class `sqlalchemy.exc.OperationalError`. Good catch

rasa/shared/utils/common.py
m-vdb: 🤔 I don't think it adds much. The path will be something like `path.to.Class`, I think there's enough information already, `is of type 'Class'` doesn't help.

rasa/shared/utils/common.py
m-vdb: gosh I don't like this convention :D

rasa/shared/nlu/training_data/entities_parser.py
m-vdb: I'm afraid it can break some behaviours, especially the `RasaYAMLReader`'s. I'd rather rename the exception to something more explicit if it's not already the case: the parsing broke because of an invalid format (not JSON). Not because the format didn't follow a specific schema (ValidationError). WDYT?

rasa/shared/exceptions.py
m-vdb: I think I meant AWS. But i'll write DynamoDB

tests/shared/nlu/training_data/test_entities_parser.py
m-vdb: sure

rasa/core/tracker_store.py
m-vdb: @wochinge I dunno if you saw this. Any opinion?

rasa/core/brokers/broker.py
m-vdb: @wochinge I dunno if you saw this. Any opinion?

tests/core/test_tracker_stores.py
```diff
@@ -896,3 +897,17 @@ def test_current_state_without_events(default_domain: Domain):
...
+
+    with pytest.raises(ConnectionException):
+        TrackerStore.create(store, default_domain)
+
+
+def test_dynamo_connection_error(default_domain: Domain):
+    store = EndpointConfig.from_dict({"type": "dynamo",})
+
+    with pytest.raises(ConnectionException):
+        TrackerStore.create(store, default_domain)
```
wochinge, line 913: optional: seems like a good use case for `pytest.mark.parametrize`

wochinge (commented):
Great iteration! Waiting for the approval due to my comment about Rasa X implications.

rasa/core/brokers/broker.py
wochinge: > what about SQLEventBroker that can be instantiated with postgres dialect

True 👍🏻 They definitely shouldn't use it though. If you go the extra step to use Rasa Open Source with an extra service, then they'd should use something right for the job.

wochinge: > Not sure if we should catch the base KafkaError, it could hide errors from us.

I agree. Seems too broad. How about

* `NoBrokersAvailable`
* `NodeNotReadyError`
* `KafkaProtocolError`
* `UnrecognizedBrokerVersion`
* `IncompatibleBrokerVersion`
* `AuthenticationMethodNotSupported`
* `AuthenticationFailedError`
* `TopicAuthorizationFailedError`
* `GroupAuthorizationFailedError`
* `ClusterAuthorizationFailedError`
* `UnsupportedVersionError`
*  `KafkaUnavailableError`
* `KafkaConnectionError`
* `ProtocolError`
* `KafkaConfigurationError`


Maybe `RETRY_BACKOFF_ERROR_TYPES` is also enough for now?

Haha, have you seen [this](https://github.com/dpkp/kafka-python/blob/9feeb79140ed10e3a7f2036491fc07573740c231/kafka/errors.py#L104) one? 😆

m-vdb: holy crap 😄

rasa/core/tracker_store.py
m-vdb: It could make sense to do that indeed. I still have a concern that if we're ignoring all `BotoCoreErrors` in Sentry, we'll ignore both errors generated by users (e.g. wrong connection URIs for instance, which is good) and errors generated by rasa (e.g. some parameters that we send as part of the TrackerStore do not pass validation) 🤔

rasa/core/brokers/broker.py
wochinge: Very good point. I think **in theory** the `broker` / `tracker_store` modules should only be imported when they are actually used. Which means that they shouldn't impact things like `rasa --help` but of course will impact things like `rasa run` (which is fine). From that perspective it's in my opinion fine to have the imports of the exceptions as top level imports. 

You could argue that **in practise** we have no clue when / where these modules are imported and that it could be easier to use local imports in this one module instead of having to watch out in all other modules. 

As far as I know these 3rd party imports are only used in a few modules so I think that the 2nd approach (using local imports) might indeed be the easier one. 

>  Is it impacting the start-up time of the CLI and do we have evidence of that?

we once had a test for the CLI startup time but removed it as far as I know 😬 

> would it make sense to have only one shared CONNECTION_ERRORS tuple somewhere in rasa.shared.constants or rasa.constants?

I personally would favor locality. Downstream dependencies (like Rasa X) should handle their connection issues themselves (which speaks against `shared`) and I think it makes sense to keep the 3rd party integrations as much as possible separate which would speak against the `rasa.constants` (the last point is a rather weak argument to be honest).

rasa/shared/nlu/training_data/formats/markdown.py
```diff
@@ -175,7 +176,7 @@ def _get_validated_dict(json_str: Text) -> Dict[Text, Text]:
     def _set_current_section(self, section: Text, title: Text) -> None:
         """Update parsing mode."""
         if section not in AVAILABLE_SECTIONS:
-            raise ValueError(
+            raise MarkdownException(
```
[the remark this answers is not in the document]
m-vdb: yes, good point re backwards compatibility

rasa/core/tracker_store.py
m-vdb: 🤔 but I think it's better that way. And it's not like errors are gonna be silent, users will still see them.

rasa/shared/nlu/training_data/entities_parser.py
m-vdb: yes better 👍🏻 I've made the changes in ffb283f1d45

rasa/core/brokers/broker.py
m-vdb: I will defer to your advice here. I am not a big fan of having local imports here and there without a clear understanding of why we're doing this (right it's more or less a guess). That'd be way better if we could make educated decisions using performance arguments and data points. Changes made in 230c773d4b0

rasa/core/brokers/broker.py
m-vdb: I think your list is better than the `RETRY_BACKOFF_ERROR_TYPES` list.

m-vdb: @wochinge actually the Kafka event broker doesn't try connecting when you instantiate it, so there is no way it raises an exception: https://github.com/RasaHQ/rasa/blob/3d4f2659a6516fdac2c9cf809071cbb9abacaeda/rasa/core/brokers/kafka.py#L15-L74

tests/shared/utils/test_common.py
m-vdb: cc @wochinge I suspect that there is a use case that never works in this function (or I am missing something): I wrote a test case with that should go down [this branch](https://github.com/RasaHQ/rasa/blob/230c773d4b0ad0c1dc158f8efa1d8252fc15eb42/rasa/shared/utils/common.py#L43) and return a loaded class, but it never works. (cf. my two first comments in this thread).

m-vdb:
@wochinge thanks for all your review comments 💯 I think I addressed all your comments and there is [this remaining open question](https://github.com/RasaHQ/rasa/pull/7292#discussion_r545051621) before this can be merged

rasa/core/brokers/broker.py
```diff
@@ -22,7 +26,18 @@ async def create(
         if isinstance(obj, EventBroker):
             return obj
 
-        return await _create_from_endpoint_config(obj, loop)
+        import aio_pika.exceptions
```
joejuzl, line 29: are these duplicate imports?

rasa/core/brokers/broker.py
```diff
@@ -22,7 +26,18 @@ async def create(
...
+        import sqlalchemy.exc
+
+        connection_errors = (
+            sqlalchemy.exc.OperationalError,
+            aio_pika.exceptions.AMQPConnectionError,
+        )
+
+        try:
+            return await _create_from_endpoint_config(obj, loop)
+        except connection_errors as error:
```
joejuzl, line 39: I would just inline this I think

rasa/core/tracker_store.py
```diff
@@ -505,16 +520,15 @@ def retrieve(self, sender_id: Text) -> Optional[DialogueStateTracker]:
         )
 
     def keys(self) -> Iterable[Text]:
-        """Returns sender_ids of the DynamoTrackerStore"""
+        """Returns sender_ids of the DynamoTrackerStore."""
```
joejuzl, line 523:
```suggestion
        """Returns sender_ids of the `DynamoTrackerStore`."""
```

rasa/shared/exceptions.py
```diff
@@ -17,6 +18,10 @@ class InvalidParameterException(RasaException, ValueError):
     """Raised when an invalid parameter is used."""
 
 
+class MarkdownException(RasaException, ValueError):
```
joejuzl, line 21: Why is this a `RasaException`?

rasa/shared/exceptions.py
```diff
@@ -77,3 +82,19 @@ class InvalidConfigException(ValueError, RasaException):
 
 class UnsupportedFeatureException(RasaCoreException):
     """Raised if a requested feature is not supported."""
+
+
+class SchemaValidationError(RasaException, jsonschema.ValidationError):
+    """Raised if schema validation via `jsonschema` failed."""
+
+
+class InvalidEntityFormatException(RasaException):
```
joejuzl, line 91: Should this also inherit from some kind of json error?

rasa/shared/utils/common.py
```diff
@@ -5,32 +5,60 @@
...
+import rasa.shared.utils.io
+from rasa.shared.constants import NEXT_MAJOR_VERSION_FOR_DEPRECATIONS
+
+
 logger = logging.getLogger(__name__)
 
 
 def class_from_module_path(
-    module_path: Text, lookup_path: Optional[Text] = None
+    module_path: Text, lookup_path: Optional[Text] = None, ensure_class: bool = True
```
joejuzl, line 16: As currently `ensure_class` only raises a warning, is there any situation we have it as `False`?

rasa/shared/exceptions.py
m-vdb: it's an exception generated by user input which is expected, checkout the PR desc:
> Raise MarkdownException when Markdown cannot be read ([sentry/1953580685](https://sentry.io/organizations/rasahq/issues/1953580685/?project=2801673))

rasa/core/brokers/broker.py
m-vdb: forgot to remove the top-level ones 🤦🏻

rasa/shared/exceptions.py
m-vdb: can you detail why do you suggest this?

rasa/shared/utils/common.py
m-vdb: no I don't think so. It's raising a deprecation warning, that will be transformed into an exception in the next major.

m-vdb: 🤔 I can't remember why exactly I introduced it (extra caution?). I think we can remove it actually: b6cceb5e90a

rasa/shared/exceptions.py
joejuzl: Ah ok - I interpreted "Markdown cannot be read" as meaning invalid markdown, rather than not fitting our spec.

rasa/shared/exceptions.py
joejuzl: Because of the comment: "InvalidEntityFormatException if provided entity is not valid json."
I would generally assume that if data (e.g. json or yaml) is an incorrect format i.e. cannot be parsed, then it would be some kind of decode, parse or value error. 
And if it can be parsed but does not match our schema, then it would be a Rasa exception.
Does this make sense? Or do we see `RasaException` as broader than that?

m-vdb: `RasaException` is a base class for expected exceptions raised when using Rasa Open Source (and most probably resulting from a wrong usage). These exceptions do not get sent to Sentry (there is nothing we can do about it from an engineering POV) but are reported to the users CLI. I'll update the docstring of the exception to reflect this, I think it's missing. Does that make sense?

`InvalidEntityFormatException` -> I think it's a good point too. I'll see if I can easily inherit the [`JSONDecodeError`](https://docs.python.org/3/library/json.html#json.JSONDecodeError) (constructor requires more arguments). It's also better in terms of backwards-compatibility (if you expect the code to raise `JSONDecodeError`, it'll still work)

m-vdb: changes in ece18bcfe17 and edcd70fae23

rasa/shared/utils/common.py
```diff
@@ -13,24 +17,45 @@ def class_from_module_path(
...
+        module_path: either an absolute path to a Python class,
+                     or the name of the class in the local / global scope.
+        lookup_path: a path where to load the class from, if it cannot
+                     be found in the local / global scope.
+
+    Returns:
+        a Python class
+
+    Raises:
+        ImportError, in case the Python class cannot be found.
```
wochinge, line 32: technically also a `TypeError` in case `import_module` fails

tests/shared/utils/test_common.py
wochinge: thanks for adding the new tests 💯 
Using `locals()` as fallback doesn't make any sense in my opinion. Also `globals` can't work as it will only have the globals [of the module](https://docs.python.org/3/library/functions.html#globals) which the `class_from_module_path`  is defined in. As this is a pure utils class there is nothing of interest in the `globals`. In my opinion two things make sense:
1. Remove this part (preferred option) 
2. Make it possible to pass the callers `globals` (less preferred as no code is currently using this)

What do you think?

wochinge (approved):
Thanks for improving the reporting and doing some maintenance work at the same time 💯

tests/shared/utils/test_common.py
m-vdb: I think removing it is best 👍🏻 if we need the 2. at some point, we can always implement it later (which I doubt...)

rasa/shared/utils/common.py
m-vdb: not sure I follow 🤔 doesn't it raise a `ModuleNotFoundError` which inherits `ImportError`?

wochinge: `import_module` in 3.8.6 also raises a `TypeError` which inherits from `Exception`

```python
def import_module(name, package=None):
    """Import a module.

    The 'package' argument is required when performing a relative import. It
    specifies the package to use as the anchor point from which to resolve the
    relative import to an absolute import.

    """
    level = 0
    if name.startswith('.'):
        if not package:
            msg = ("the 'package' argument is required to perform a relative "
                   "import for {!r}")
            raise TypeError(msg.format(name))
        for character in name:
            if character != '.':
                break
            level += 1
    return _bootstrap._gcd_import(name[level:], package, level)
```

m-vdb: 🤔 yeah ok. It sounds a bit more like a detail. Fine if I keep it like it is?

joejuzl (approved):

rasa/shared/utils/common.py
wochinge: It would be more transparent for the caller as they'd know which errors to handle without having to dive deeper into the code, but up to you as it's a detail as you said.

m-vdb:
finallyyyyyyyyyyy merged!

Issue #7301, opened 2020-11-17
Contradicting rules or stories found


pheel:
Sending the following e2e file off to regular training

```yaml
rules:
  - rule: Thank you
    steps:
      - intent: thankyou
        user: thanks!!!
      - action: utter_noworries
```

causes a **Contradicting rules or stories found** error. Using the same files for training and testing is a convenience. I suggest to ignore the `user` key when e2e mode if off.

**Rasa version**: 2.0.x

sara-tagger:
Thanks for the issue, @TyDunn will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

samsucik:
@pheel have you checked that there are actually no contradicting rule or story parts in your training data? What exactly does the error say and what do your training stories/rules look like?

I should also mention that there have been some bugs uncovered recently in `RulePolicy` that were causing this kind of error to be raised when using a slot with `initial_value` set while also using a rule with `conversation_start: true` (see #7235 ), but that might not affect you of course :-)

pheel:
@samsucik Thanks for getting back to me! Yes I did come upon that PR when researching, but it doesn't seem related. Here's the config and error. Removing the `user` key in the intent step makes it train fine.

**data.yml**
```yaml
rules:
  - rule: Thank you
    steps:
      - intent: thankyou
        user: thanks!!!
      - action: utter_noworries
```

**domain.yml**
```yaml
intents:
  - thankyou

responses:
  utter_noworries:
  - text: no worries!
```

**config.yml**
```yaml
...
policies:
  - name: RulePolicy
  - name: TEDPolicy
```

**error**
```
InvalidRule:
Contradicting rules or stories found 🚨

- the prediction of the action 'utter_greet' in rule 'Thank you' is contradicting with rule(s) ''.
- the prediction of the action 'action_listen' in rule 'Thank you' is contradicting with rule(s) ''.
Please update your stories and rules so that they don't contradict each other.
You can find more information about the usage of rules at https://rasa.com/docs/rasa/rules.
```

samsucik:
Thanks a lot @pheel! I've managed to reproduce it. Will dig deeper into it. While the `user` key looks weird to be included in a rule, it's also true that rules and stories shouldn't really differ -- and in this case including the key breaks things when in a rule but doesn't when in a training story...

samsucik:
[...]

PR #7302, opened 2020-11-17
Fix core training with user key present 

pheel:
Fixes https://github.com/RasaHQ/rasa/issues/7301.

### DIFF
--- a/changelog/7301.bugfix.mdx
+++ b/changelog/7301.bugfix.mdx
@@ -1,0 +1,2 @@
+Fix a bug where stories and rules with `user` keys (used in e2e evaluation)
+would be rejected and throw an error if used in training.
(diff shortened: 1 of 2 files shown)


sara-tagger:

samsucik:
Hi @pheel, could you clarify the purpose here a bit: You want test stories which contain both `intent` and `user` keys to be used in training but you want to ignore the `user` information at training time, right? I guess you've written some stories in the test-story format and just want to use some of those for training?

Using the current `master` version, I tried training `TEDPolicy` on a simple `yaml` story that contains both `intent` and `user` keys for one user turn. I didn't get any errors. Could you share the errors that you're getting (ideally using 2.1.0 or the current `master`) and also some minimalistic setup that leads to those errors? Then I could try to reproduce the errors so I better understand what's going on here. Thanks!

samsucik:
@pheel this will require a different approach, and it should be partly fixed soon as part of a new feature that can use stories/rules with `user` features for training. In any case, the YAML reader should be agnostic to the training vs testing phase and should in both cases read and pass on all the information. I've also created a new issue: #7360.

PR #7308, opened 2020-11-18
Fix `black` to be the same version in the pre-commit config as it's in the pyproject config

samsucik:
**Proposed changes**:
In `.pre-commit-config.yaml` we used `stable` version of `black` but in `pyproject.toml` we had a fixed major version. Now the version should is fixed inside `.pre-commit-config.yaml` as well.

I've checked locally and the updated config installs the right version of `black` into my pre-commit hook.


m-vdb (approved):
thanks @samsucik 🚀

PR #7316, opened 2020-11-19
core featurizers should use regex interpreter if core was trained separately

Ghostvv:
**Proposed changes**:
- fix issue when separate core and nlu training and combining models manually broke core featurizers

### DIFF
--- a/rasa/core/featurizers/single_state_featurizer.py
+++ b/rasa/core/featurizers/single_state_featurizer.py
@@ -34,10 +34,11 @@ class SingleStateFeaturizer:
     """
 
     def __init__(self) -> None:
+        """Initialize the single state featurizer."""
         # rasa core can be trained separately, therefore interpreter during training
         # will be `RegexInterpreter`. If the model is combined with a rasa nlu model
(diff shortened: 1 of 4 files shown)


Ghostvv:
should be merged after https://github.com/RasaHQ/rasa/pull/7235

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -34,15 +34,27 @@ class SingleStateFeaturizer:
     """
 
     def __init__(self) -> None:
+        # rasa core can be trained separately, therefore interpreter during training
+        # will be `RegexInterpreter` but during prediction it'll be
+        # interpreter based on some trained nlu model
```
tabergma, line 39:
```suggestion
        # rasa core can be trained separately, therefore interpreter during training
        # will be `RegexInterpreter`. If the model is combined with a rasa nlu model
        # during prediction the interpreter might be different. If that is the case, we need
        # to make sure to "reset" the interpreter.
```

tabergma (approved):
Just tested it, works as expected 💯  Thanks for fixing this. 

Please add a changelog entry before merging.

Issue #6410, opened 2020-08-14
E2E actions should result in `BotUtterance`s


wochinge:
With Rasa Open Source 2.0 we introduce end-to-end actions. End-to-end actions don't have a label (like `utter_greet`) but contain the actual user utterance, e.g.:

```yaml
stories:
- story: My end-to-end story
  steps:
  - intent: greet
  - bot: Hey, how are you? 
```

Todos:
- [x] make sure that an executed `Action` which has an end-to-end bot utterance, creates a `BotUttered` event
- [x] think about how to distinguish "End-to-end actions" from regular actions in the domain

wochinge:
close in favor of https://github.com/RasaHQ/rasa/pull/7325

PR #7325, opened 2020-11-20
utter end-to-end responses

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6410
- deprecate `random_template_for` as it's unused
- rename `Domain.action_names` to `Domain.action_names_or_texts`. Added a property with deprecation message for `Domain.action_names`
- add docstrings
- utter end-to-end bot responses if predicted
- log end-to-end action correctly (`action_text` property is set)
- added method `event_for_successful_execution` to `Action` to encapsulate action event with the action which ran it

### DIFF
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -102,7 +102,9 @@ def action_for_index(
             f"Domain has {domain.num_actions} actions."
         )
 
-    return action_for_name_or_text(domain.action_names[index], domain, action_endpoint)
+    return action_for_name_or_text(
+        domain.action_names_or_texts[index], domain, action_endpoint
+    )
 
 
 def is_retrieval_action(action_name: Text, retrieval_intents: List[Text]) -> bool:
@@ -141,7 +143,7 @@ def action_for_name_or_text(
     Returns:
         The instantiated action.
     """
-    if action_name_or_text not in domain.action_names:
+    if action_name_or_text not in domain.action_names_or_texts:
         domain.raise_action_not_found_exception(action_name_or_text)
 
     defaults = {a.name(): a for a in default_actions(action_endpoint)}
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -6,7 +6,6 @@ from typing import List, Text, Optional, Dict, Any, Set, TYPE_CHECKING
 import aiohttp
 
 import rasa.core
-from rasa.core.policies.policy import PolicyPrediction
 
 from rasa.shared.core import events
 from rasa.core.constants import DEFAULT_REQUEST_TIMEOUT
@@ -79,15 +78,14 @@ def default_actions(action_endpoint: Optional[EndpointConfig] = None) -> List["A
     ]
 
 
-def action_for_prediction(
-    prediction: PolicyPrediction,
-    domain: Domain,
-    action_endpoint: Optional[EndpointConfig],
+def action_for_index(
+    index: int, domain: Domain, action_endpoint: Optional[EndpointConfig]
 ) -> "Action":
-    """Gets an instantiated `Action` based on the winning policy's prediction.
+    """Get an action based on its index in the list of available actions.
 
     Args:
-        prediction: The prediction for the next action.
+        index: The index of the action. This is usually used by `Policy`s as they
+            predict the action index instead of the name.
         domain: The `Domain` of the current model. The domain contains the actions
             provided by the user + the default actions.
         action_endpoint: Can be used to run `custom_actions`
@@ -97,26 +95,17 @@ def action_for_prediction(
         The instantiated `Action` or `None` if no `Action` was found for the given
         index.
     """
-    index = prediction.max_confidence_index
     if domain.num_actions <= index or index < 0:
         raise IndexError(
             f"Cannot access action at index {index}. "
             f"Domain has {domain.num_actions} actions."
         )
 
-    return action_for_name(
-        domain.action_names[index],
-        domain,
-        action_endpoint,
-        prediction.is_end_to_end_prediction,
-    )
+    return action_for_name(domain.action_names[index], domain, action_endpoint)
 
 
 def action_for_name(
-    action_name: Text,
-    domain: Domain,
-    action_endpoint: Optional[EndpointConfig],
-    is_end_to_end_prediction: bool = False,
+    action_name: Text, domain: Domain, action_endpoint: Optional[EndpointConfig]
 ) -> "Action":
     """Create an `Action` object based on the name of the `Action`.
 
@@ -126,7 +115,6 @@ def action_for_name(
             provided by the user + the default actions.
         action_endpoint: Can be used to run `custom_actions`
             (e.g. using the `rasa-sdk`).
-        is_end_to_end_prediction: `True` if it is an end-to-end prediction.
 
     Returns:
         The instantiated `Action` or `None` if no `Action` was found for the given
@@ -136,9 +124,7 @@ def action_for_name(
     if action_name not in domain.action_names:
         domain.raise_action_not_found_exception(action_name)
 
-    return action_from_name(
-        action_name, domain, action_endpoint, is_end_to_end_prediction
-    )
+    return action_from_name(action_name, domain, action_endpoint)
 
 
 def is_retrieval_action(action_name: Text, retrieval_intents: List[Text]) -> bool:
@@ -162,10 +148,7 @@ def is_retrieval_action(action_name: Text, retrieval_intents: List[Text]) -> boo
 
 
 def action_from_name(
-    name: Text,
-    domain: Domain,
-    action_endpoint: Optional[EndpointConfig],
-    is_end_to_end_prediction: bool = False,
+    name: Text, domain: Domain, action_endpoint: Optional[EndpointConfig]
 ) -> "Action":
     """Retrieves an action by its name.
 
@@ -173,7 +156,6 @@ def action_from_name(
         name: The name of the action.
         domain: The current model domain.
         action_endpoint: The endpoint to execute custom actions.
-        is_end_to_end_prediction: `True` if it is an end-to-end prediction.
 
     Returns:
         The instantiated action.
@@ -188,7 +170,7 @@ def action_from_name(
     ):
         return ActionRetrieveResponse(name)
 
-    if name.startswith(UTTER_PREFIX) or is_end_to_end_prediction:
+    if name.startswith(UTTER_PREFIX) or name in domain.action_texts:
         return ActionUtterTemplate(name)
 
     is_form = name in domain.form_names
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -158,7 +158,7 @@ def action_for_name_or_text(
         return ActionRetrieveResponse(action_name_or_text)
 
     if action_name_or_text in domain.action_texts:
-        return EndToEndAction(action_name_or_text)
+        return ActionEndToEndResponse(action_name_or_text)
 
     if action_name_or_text.startswith(UTTER_PREFIX):
         return ActionUtterTemplate(action_name_or_text)
@@ -288,7 +288,7 @@ class ActionUtterTemplate(Action):
         return "ActionUtterTemplate('{}')".format(self.name())
 
 
-class EndToEndAction(Action):
+class ActionEndToEndResponse(Action):
     """Action to utter end-to-end responses to the user."""
 
     def __init__(self, action_text: Text) -> None:
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -301,7 +301,8 @@ class EndToEndAction(Action):
 
     def name(self) -> Text:
         """Returns action name."""
-
+        # In case of an end-to-end action there is no label (aka name) for the action.
+        # We fake a name by returning the text which we respond to the user.
         return self.action_text
 
     async def run(
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -170,7 +170,7 @@ def action_from_name(
     ):
         return ActionRetrieveResponse(name)
 
-    if name.startswith(UTTER_PREFIX):
+    if name.startswith(UTTER_PREFIX) or name in domain.action_texts:
         return ActionUtterTemplate(name)
 
     is_form = name in domain.form_names
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -137,10 +137,10 @@ def action_for_name_or_text(
 
     Raises:
         ActionNotFoundException: If action not in current domain.
+
     Returns:
         The instantiated action.
     """
-
     if action_name_or_text not in domain.action_names:
         domain.raise_action_not_found_exception(action_name_or_text)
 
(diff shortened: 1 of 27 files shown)


wochinge:
@Ghostvv how can I test this? I tried using the `e2e` example bot but it only predicts regular actions 🤯

rasa/core/actions/action.py
```diff
@@ -170,7 +170,7 @@ def action_from_name(
     ):
         return ActionRetrieveResponse(name)
 
-    if name.startswith(UTTER_PREFIX):
+    if name.startswith(UTTER_PREFIX) or name in domain.action_names:
```
joejuzl, line 173: Won't this prevent remote actions?

rasa/shared/core/domain.py
```diff
@@ -521,12 +536,15 @@ def __init__(
...
 
         self.slots = slots
-        self.templates = templates
-        self.action_texts = action_texts or []
         self.session_config = session_config
 
         self._custom_actions = action_names
 
+        self.action_texts = action_texts or []
+        self._user_responses = templates
```
joejuzl, line 544: Does this mean responses _to_ the user?

rasa/shared/importers/importer.py
```diff
@@ -464,7 +464,7 @@ async def _get_domain_with_e2e_actions(self) -> Domain:
             [],
             {},
             action_names=[],
-            forms=[],
+            forms={},
```
joejuzl, line 467: What's the reason for this change?

rasa/core/actions/action.py
wochinge: omg, yes of course. That was a bling auto complete 🙈

rasa/shared/importers/importer.py
wochinge: That's related to [this change](https://github.com/RasaHQ/rasa/pull/7171). Seems something went wrong when merging `master` into this branch.

rasa/shared/core/domain.py
wochinge: renamed it. Is it clearer now?

wochinge: The difference is that these responses are defined in the domain by the developer. The bot will send them to the user. The `action_texts` are from stories or rules and will also be sent to the user. However, the developer defines them in an end-to-end fashion within the stories / rules instead of the domain. Does that make sense?

Ghostvv:
>  I tried using the e2e example bot but it only predicts regular actions 

strange, it predicts text for me, if I enter `Hello` or smth

wochinge:
I'll try again later 👍 🤔

wochinge:
@Ghostvv The tests fail due to story printing, correct?

Ghostvv:
yes, this are the tests I was talking about earlier

wochinge:
@joejuzl @Ghostvv Ready for another review. It works locally. Also fixed how the action is logged (we previously logged the action text in the action name property).

Testing with the example bot worked now 🤷

joejuzl:
LGTM

rasa/core/actions/action.py
```diff
@@ -95,17 +97,26 @@ def action_for_index(
...
+        prediction.is_end_to_end_prediction,
+    )
 
 
 def action_for_name(
-    action_name: Text, domain: Domain, action_endpoint: Optional[EndpointConfig]
+    action_name: Text,
+    domain: Domain,
+    action_endpoint: Optional[EndpointConfig],
+    is_end_to_end_prediction: bool = False,
```
Ghostvv, line 119: in `Prediction` class `is_end_to_end_prediction` means that a prediction was made using user text, is it the same here?
In this case it shouldn't affect whether it is `action_text` or `action_name`, It's domain's job to differentiate based on given index

rasa/core/policies/policy.py
```diff
@@ -460,6 +460,30 @@ def max_confidence(self) -> float:
...
+            action_name: The name of the predicted action. If it was an end-to-end
+                prediction the `action_name` is actually the predicted text.
+
+        Returns:
+            The events which should be logged.
+        """
+        action_text = None
+        if self.is_end_to_end_prediction:
+            action_text = action_name
+            action_name = None
```
Ghostvv, line 476: this is incorrect. Whether it is `action_text` or `action_name` decides domain.
`self.is_end_to_end_prediction` only indicates what type of user input (text or intent) was used

Ghostvv (changes requested):

rasa/core/policies/policy.py
wochinge: Yes, of course 😬

wochinge: Too much refactoring - I could luckily just revert the commit 😁

rasa/core/policies/policy.py
```diff
@@ -16,7 +16,7 @@
     TYPE_CHECKING,
 )
 import numpy as np
-from rasa.shared.core.events import Event
+from rasa.shared.core.events import Event, ActionExecuted
```
Ghostvv, line 19: I think `ActionExecuted` will be unused export?

rasa/core/processor.py
```diff
@@ -829,15 +829,26 @@ def _log_action_on_tracker(
...
+            tracker.update(self._action_event_for_prediction(action_name, prediction))
 
         logger.debug(f"Action '{action_name}' ended with events '{events}'.")
         tracker.update_with_events(events, self.domain)
 
+    def _action_event_for_prediction(
+        self, action_name: Text, prediction: PolicyPrediction
+    ) -> ActionExecuted:
+        action_text = None
+        if action_name in self.domain.action_texts:
```
Ghostvv, line 841: I would prefer modifying the part of the code (maybe inside the `domain.py`) that transforms predicted index into an action name, to return a tuple `(action_name, action_text)` one of which would be `None`. Then all of these method should take both name and text and pass it along. this way we decouple the logic of text vs name and contain it only within domain file

wochinge: We only pass around the `Action` object which we retrieve [here](https://github.com/RasaHQ/rasa/blob/d43364627ce27b6c7048a7e460a0087a84447f3e/rasa/core/processor.py#L357). I think passing around additional information via tuples rather increases complexity than to help with it. What I could do is to set a flag `is_end_to_end_action` on the `Action` object [here](https://github.com/RasaHQ/rasa/blob/8f60f83d089e0e4dcbd79900c7598935f50119e6/rasa/core/actions/action.py#L173) or even add method `action_executed_for_event` to `Action` which returns the correct `ActionExecuted` event.

Ghostvv: I think `action_for_index` is a perfect place for that. Instead of doing it inside `action_for_name`, create method `action_for_text` and differentiate between them inside `action_for_index`. What do you think?

wochinge: I don't see any advantage of doing this. What are you to accomplish? If you want to isolate determining whether it's an end-to-end action, then I'd go for the described approach above.

Ghostvv: because the name `action_for_name` is confusing. Name doesn't decide the action anymore, it is the index that decides that. Why do we have 2 identical methods on the first place?

Ghostvv: I want to avoid putting `action_text` into variable called `action_name` in a couple of month it'll create terrible confusion

wochinge: > Why do we have 2 identical methods on the first place?

Which two methods are you referring to? Index and name? There are some cases where have a name and want to get the action based on that (e.g. for the `action_session_started)

> I want to avoid putting action_text into variable called action_name in a couple of month it'll create terrible confusion

I see that. As only `_log_action_on_tracker` is having this variable I still suggest doing what I suggested [in my earlier. comment.](https://github.com/RasaHQ/rasa/pull/7325#discussion_r530490382).

Ghostvv: I'm not sure I fully follow your proposal, but if you think it is better let's go for it. I'd prefer the implementation where there is no need for additional flags

wochinge: I agree on the flag thing and I've an idea about that. Can hopefully do a quick draft later which we can then discuss.

rasa/core/actions/action.py
```diff
@@ -147,41 +125,53 @@ def is_retrieval_action(action_name: Text, retrieval_intents: List[Text]) -> boo
...
-    if name.startswith(UTTER_PREFIX) and is_retrieval_action(
-        name, domain.retrieval_intents
+    if action_name_or_text.startswith(UTTER_PREFIX) and is_retrieval_action(
+        action_name_or_text, domain.retrieval_intents
     ):
-        return ActionRetrieveResponse(name)
+        return ActionRetrieveResponse(action_name_or_text)
+
+    if action_name_or_text in domain.action_texts:
+        return EndToEndAction(action_name_or_text)
```
wochinge, line 161: this is the essential part 1 of the change

rasa/core/actions/action.py
```diff
@@ -244,6 +234,21 @@ async def run(
     def __str__(self) -> Text:
         return "Action('{}')".format(self.name())
 
+    def event_for_successful_execution(
```
wochinge, line 237: this is the essential part 2 of the change

rasa/core/actions/action.py
```diff
@@ -283,6 +288,51 @@ def __str__(self) -> Text:
         return "ActionUtterTemplate('{}')".format(self.name())
 
 
+class EndToEndAction(Action):
```
wochinge, line 291: this is the essential part 3 of the change

rasa/core/processor.py
wochinge: @Ghostvv Updated it. Way cleaner solution now. I was able to get rid of all domain changes. Haven't polished the PR yet because I wanted to share it with you first, so please comment on the higher level things

rasa/core/actions/action.py
```diff
@@ -147,41 +125,53 @@ def is_retrieval_action(action_name: Text, retrieval_intents: List[Text]) -> boo
...
         domain: The current model domain.
         action_endpoint: The endpoint to execute custom actions.
 
+    Raises:
+        ActionNotFoundException: If action not in current domain.
     Returns:
         The instantiated action.
     """
+
+    if action_name_or_text not in domain.action_names:
```
Ghostvv, line 144: is it still `domain.action_names`? Can we rename this one as well?

rasa/core/actions/action.py
Ghostvv: I'd say logically it should go before checking `.startswith(UTTER_PREFIX)` on Line 155

rasa/core/actions/action.py
Ghostvv: shall we name it in the spirit of other actions? Smth like `ActionEndToEndText`? or smth similar

Ghostvv (commented):
I like your approach, I agree it makes things clean

rasa/core/actions/action.py
wochinge: Mhm, one could argue that having it after it is better for performance as the `utter_` case should be way more common.

rasa/core/actions/action.py
wochinge: We could shorten it to just `actions`, what do you think? I'd probably keep the old one and deprecate it's usage.
We might want to consider doing this change once everything is merged to master to not increasing the diff unnecessarily.

Ghostvv: `actions` would be confusing as well, since `actions` represent actual actions `Action` class or `ActionExecuted` event

Ghostvv: at this point I wouldn't worry about the diff, it is skyrocket anyhow 🙈

rasa/core/actions/action.py
Ghostvv: but what if your action_text is "utter_goodbye should be called when user says goodbye"?

Ghostvv:
@wochinge is it ready for final review?

wochinge:
@Ghostvv My calendar is evil today. Won't get to it until tomorrow 😬

rasa/core/actions/forms.py
```diff
@@ -550,13 +551,16 @@ async def _ask_for_slot(
...
-        action_to_ask_for_next_slot = action.action_from_name(
-            self._name_of_utterance(domain, slot_name), domain, self.action_endpoint
-        )
-        events_to_ask_for_next_slot = await action_to_ask_for_next_slot.run(
-            output_channel, nlg, tracker, domain
-        )
-        return events_to_ask_for_next_slot
+        action_to_ask_for_next_slot = self._name_of_utterance(domain, slot_name)
+
+        if action_to_ask_for_next_slot:
```
wochinge, line 556: this change was necessary to avoid running an non-existing action (it would fail silently before)

rasa/core/actions/action.py
```diff
@@ -147,41 +125,53 @@ def is_retrieval_action(action_name: Text, retrieval_intents: List[Text]) -> boo
...
-        name, domain.retrieval_intents
+    if action_name_or_text.startswith(UTTER_PREFIX) and is_retrieval_action(
+        action_name_or_text, domain.retrieval_intents
     ):
-        return ActionRetrieveResponse(name)
+        return ActionRetrieveResponse(action_name_or_text)
 
-    if name.startswith(UTTER_PREFIX):
-        return ActionUtterTemplate(name)
+    if action_name_or_text in domain.action_texts:
```
wochinge, line 160: also added a test for this to ensure nobody thinks about switching these :-D

wochinge:
@Ghostvv Ready for review

rasa/core/actions/action.py
Ghostvv: but I'd move this `if` to line 156, do you think this check will hinder the performance significantly? The win is that it'll be less error prone

Ghostvv (commented):
left a comment about the order of checks. Otherwise, looks good from me. But since it introduces a lot of changes to `shared`, I think it is better if someone from backend approves it

rasa/core/actions/action.py
wochinge: tbh, I don't think the performance here matters too much. That should literally be a few milliseconds while e.g. some ML policies perform duplicate steps which are way more cost intense. Or we also fetch the tracker from the database more often than we need too which is in the region of ~100ms

wochinge: (recently started exploring prediction performance in https://github.com/RasaHQ/rasa/pull/7205)

wochinge:
@joejuzl Do you want to have another look or can I merge?

joejuzl (approved):

rasa/core/actions/action.py
Ghostvv: so why didn't you move it?

Ghostvv: it should be the second check before `utter_` prefix check

Ghostvv: @wochinge do we have a check that `action_text` is not exactly the same as some action name, then I believe will have a problem that `domain.action_name_or_text` contain non unique items

wochinge: Yes, `Domain` will check it 👍  Is your message missing a `not have a problem` or is it on purpose? You mean in case a message is coming up in several stories? We de-duplicate them before ([here](https://github.com/RasaHQ/rasa/blob/49bac016e71c8a053fceeb772f0d7f8764888d07/rasa/shared/importers/importer.py#L449))

Ghostvv: without the check there would be a problem, but if we have the check it should be good

PR #7336, opened 2020-11-22
Fixed a bug

shamikan:
**Proposed changes**:
- update the language tokenizer file in rasa/nlu to resolve location conflict while running language model from config.yml

- [ ] renamed the lm_tokenizer.py to languagemodel_tokenizer.py


sara-tagger:
Thanks for submitting a pull request 🚀 @amn41 will take a look at it as soon as possible ✨

dakshvar22:
Hi @shamikan thanks for opening the PR. Can you give more details on why do you want to rename the file? What kind of location conflict arises? Would be helpful if you can share the config where the problem arises.

shamikan:
Hello sir,

Thank you for your response and consideration. Talking about my PR request,
after installation of rasa on my Mac, I couldn't make use of language model
tokenizer in config.yml for my chatbot. However, I followed the
documentation given on the website, yet didn't help. The model could use
other tokenizer like whitespace etc but specifically couldn't use
languagemodel tokenizer and showed error; and the error mentioned that in
the tokenizer file(inside nlu file) no such languagemodel tokenizer is
available. But it was present by the name of lm_tokenizer, however changing
the name to languagemodel_tokenizer made it work. So, I found making the
change to orginal GitHub repo would be much help and hence generated the PR.

Please look into this and let me know if my changing the name aligns with
the website documentation.

Thank you.

Best,
Shamik

dakshvar22:
@shamikan We'll need to reproduce the error. Can you please share the config that lead to the error and also the error that you see? Changing the file name means changing the API which is something we would want to avoid as much as possible.

shamikan:
I understand the difficulty associated with renaming but I haven't had
taken the screenshots earlier. But I will share with you by tomorrow.

tmbo:
@shamikan any updates on the error?

tmbo:
I'll close this for now until we have more clarity why we would need this change.

PR #7337, opened 2020-11-23
Add example of entity button payload

JEM-Mosig:
**Proposed changes**:
- Add example of entity button payload
- Related to https://github.com/RasaHQ/rasa/issues/6616#issue-696827333

### DIFF
--- a/docs/docs/responses.mdx
+++ b/docs/docs/responses.mdx
@@ -167,8 +167,10 @@ Each button in the list of `buttons` should have two keys:
 :::note bypass nlu with buttons
 It's common to use buttons as a shortcut to bypass the machine-learning-based NLU interpreter.
 Messages starting with `/` are sent straight to the
-`RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format.
-In the example above, if the user clicks a button, the user input
+`RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format, e.g.
+`payload: "/inform{{\"planet\": \"saturn\"}}"`.
+
+In the example of this section, if the user clicks a button, the user input
 will be directly classified as either the `mood_great` or `mood_sad` intent.
 :::
 
--- a/docs/docs/responses.mdx
+++ b/docs/docs/responses.mdx
@@ -168,7 +168,9 @@ Each button in the list of `buttons` should have two keys:
 It's common to use buttons as a shortcut to bypass the machine-learning-based NLU interpreter.
 Messages starting with `/` are sent straight to the
 `RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format, e.g.
-`payload: "/inform{{\"planet\": \"saturn\"}}"`.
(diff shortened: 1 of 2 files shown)


docs/docs/responses.mdx
```diff
@@ -167,8 +167,10 @@ Each button in the list of `buttons` should have two keys:
 :::note bypass nlu with buttons
 It's common to use buttons as a shortcut to bypass the machine-learning-based NLU interpreter.
 Messages starting with `/` are sent straight to the
-`RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format.
-In the example above, if the user clicks a button, the user input
+`RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format, e.g.
+`payload: "/inform{{\"planet\": \"saturn\"}}"`.
```
tmbo, line 171: might look better as a code block? and do we really need the double brackets?

JEM-Mosig: I can make it a code block. Yes, this fails without the double brackets, which is what triggered me to write this.

docs/docs/responses.mdx
```diff
@@ -167,8 +167,12 @@ Each button in the list of `buttons` should have two keys:
 :::note bypass nlu with buttons
 It's common to use buttons as a shortcut to bypass the machine-learning-based NLU interpreter.
 Messages starting with `/` are sent straight to the
-`RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format.
-In the example above, if the user clicks a button, the user input
+`RegexInterpreter`, which expects NLU input in a shortened `/intent{entities}` format, e.g.
+```yaml-rasa {4-8}
+payload: "/inform{{\"planet\": \"saturn\"}}"
```
tmbo, line 172: this is what gets rendered for me:
Bildschirmfoto 2020-12-07 um 11 41 16

Do we really need the double `{{`?

JEM-Mosig: Yes, again, if you don't write the double `{{`, then it won't work.

JEM-Mosig: Has something to do with how regex processes it. I had to find that out the hard way when I built a bot.

JEM-Mosig: This is actually why I created this PR in the first place, so people don't have to search on the forum.

joejuzl:
I wonder, should the docs be updated or is this just a bug? Is there a specific reason we have to have the double curly braces?

JEM-Mosig:
> I wonder, should the docs be updated or is this just a bug? Is there a specific reason we have to have the double curly braces?

I think this should be improved with https://github.com/RasaHQ/rasa/issues/6616#issue-696827333. I just wanted to quickly get this PR for the docs through so people stop asking about it on the Forum. I thought we can update the docs again when the issue is fixed. Though given that it's been half a year now and neither issue/PR is closed, it is probably better to just focus on the other issue.

PR #7339, opened 2020-11-23
spaCy 3.0 support

koaning:
This is a work in progress PR. The idea is to see what breaks if we switch to the nightly version of spaCy. 

The docs for this version can be found [here](https://nightly.spacy.io/usage/v3) and the pypi can be found [here](https://pypi.org/project/spacy-nightly/#history).

The main reason for adding this is to keep things maintainable but also because there's support for even more languages *and* new transformer-based models.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ exclude = "((.eggs | .git | .pytest_cache | build | dist))"
 
 [tool.poetry]
 name = "rasa"
-version = "2.1.0"
+version = "2.1.1"
 description = "Open source machine learning framework to automate text- and voice-based conversations: NLU, dialogue management, connect to Slack, Facebook, and more - Create chatbots and voice assistants"
 authors = [ "Rasa Technologies GmbH <hi@rasa.com>",]
 maintainers = [ "Tom Bocklisch <tom@rasa.com>",]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -155,7 +155,7 @@ pytest-timeout = "^1.4.2"
 mypy = "^0.790"
 
 [tool.poetry.extras]
-spacy = [ "spacy",]
+spacy = [ "spacy-nightly",]
 jieba = [ "jieba",]
 transformers = [ "transformers",]
 full = [ "spacy", "transformers", "jieba",]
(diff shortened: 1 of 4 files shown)


koaning:
Installing spaCy in CI is a bother so instead, I'll just share my local observations. 

@tabergma it seems like the standard `spacy link` command is getting [deprecated](https://nightly.spacy.io/usage/v3#incompat-api) so that means that we need to require users to explicitly share their model. That mainly means that we'll need to explicitly refer to a model in our unit tests as well. I'm wondering if running `python -m spacy download en_core_web_sm`is a good idea in CI. Got an opinion on this?

tabergma:
We are also downloading other models for our tests as far as I know. But don't have a strong opinion on this. Maybe @tczekajlo can give more details.

pyproject.toml
```diff
@@ -190,4 +190,4 @@ optional = true
 
 [tool.poetry.dependencies.pymongo]
 version = ">=3.8,<3.11"
-extras = [ "tls", "srv",]
+extras = [ "tls", "srv",]
```
tczekajlo, line 193: No newline at end of file 😄

tczekajlo:
> We are also downloading other models for our tests as far as I know. But don't have a strong opinion on this. Maybe @tczekajlo can give more details.

For the Model Regression Tests workflow we do something like this:

```
          poetry run python -m spacy download de_core_news_md
          poetry run python -m spacy link --force de_core_news_md de
```
I guess it's something that we have to remember if we want to migrate to spacy 3.0

I think downloading the model in CI is ok (I'm open to other propositions). @koaning you mentioned that you build a small custom model in whatlies, can you give me a link to a place where you do this? I'd like to take a look at it 😄

koaning:
Ahhh for real? That was it? Argh! I gotta give it a try again later today. Thanks for the ping @tczekajlo!

koaning:
@tczekajlo the thing I do with spaCy is I just create a small model locally. See example [here](https://github.com/RasaHQ/whatlies/blob/master/tests/scripts/prepare_spacy_tests.py). This will need to be revamped for spaCy 3.0 too 😅 .

pyproject.toml
koaning: That's whats causing the CI errors?

koaning:
This branch was exploring `spacy-nightly` which was a rabbit hole to configure. 

So I started a new PR [here](https://github.com/RasaHQ/rasa/pull/7869).

Issue #7277, opened 2020-11-13
Implement `train_chunk` for all nlu featurizers


tabergma:
related to #6836

For partial dataset loading we want to featurize only a part of the data at a time. Implement the method `train_chunk` for all featurizers. The method will get a `TrainingDataChunk` that contains a list of messages, a subset of all training data.

Most likely you can just call the normal `train` method for the given examples.

**Note**: This feature should branch off from `partial-dataset-loading` and should be squashed and merged back into that branch.

wj-Mcat:
This is not been linked with any pull request, can I have a try to implement it?

tabergma:
@wj-Mcat Thanks for your interest in this! As this is part of a larger project that we are still defining, we cannot open up this issue to the community. Sorry. Take a look at https://github.com/orgs/RasaHQ/projects/23 for open issues that you can work on.

Ghostvv:
@tabergma is it core or nlu featurizers? or both?

tabergma:
we should have a separate issue for core, so just NLU.

Issue #7274, opened 2020-11-13
Implement prepare_partial_training on NLU components


tabergma:
related to #6836

Before the NLU components can be trained in chunks some of them need to be prepared. Each component should extract all information from the `TrainingData` needed inside the `prepare_partial_training` method, so that it can be used for partial training later on. 

For example:
- `CountVectorsFeaturizer` needs to build up a vocabulary from the complete training data. The training data should be tokenized already.
- `RegexFeaturizer` should build up the regexes from the regexes and lookup tables present in the training data.
- `RegexEntityExtractor` should build up the regexes from the regexes and lookup tables present in the training data.
- `EntitySynonymMapper` should build up the synonyms from the synonyms present in the training data.

Note: They might be more components that need to implement this method. Make sure to check all of them.

**Note**: This feature should branch off from `partial-dataset-loading` and should be squashed and merged back into that branch.

tabergma:
@Ghostvv As we spoke about this issue this morning. I guess it might make sense to tackle this together with the implementation of `trian_chunk` for the featurizers. I'll take over one of the other issues instead. Let me know if you think otherwise.

Ghostvv:
@tabergma it seems to me that `Component.prepare_partial_training` is not called anywhere. As part of this issue, should I only implement this method in all subclasses, or also call it in appropriate place?

tabergma:
it will be called once [this PR](https://github.com/RasaHQ/rasa/pull/7298) gets approved, unfortunately still waiting for a review

PR #7346, opened 2020-11-23
implement train chunk and prepare for nlu components except extractors and classifiers

Ghostvv:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/7274
- Closes https://github.com/RasaHQ/rasa/issues/7277


Ghostvv:
@tabergma since it moving code around, not sure if I should any tests?

rasa/nlu/classifiers/keyword_intent_classifier.py
```diff
@@ -38,13 +38,27 @@ def __init__(
...
+        self,
+        training_data_chunk: TrainingDataChunk,
+        config: Optional[RasaNLUModelConfig] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Train this component on the given chunk.
+
+        See parent class for more information.
+        """
+        raise Exception(
```
tabergma, line 51: Would not raise `Exception` here but rather a more specific one. What about creating a `RasaTrainException` in https://github.com/RasaHQ/rasa/blob/master/rasa/shared/exceptions.py, subclassing our `RasaException`?

tabergma (commented):
Looks great 💯 Just left one comment around the exception.

Regarding tests: We could maybe add a tests for tokenizers and maybe some of the classifiers, that checks if `train_chunk` raises an exception. And maybe a test that checks that `prepare_partial_training` is not modifying the training data.

Ghostvv:
> And maybe a test that checks that prepare_partial_training is not modifying the training data.

how can I check that?

tabergma:
Daksh implement a test like this [in his PR](https://github.com/RasaHQ/rasa/pull/7352/files#diff-f1bd9df57cdd694505dea844b8abfca6447e62187de92fa7a56d1f4623b9110bR642).

rasa/nlu/classifiers/keyword_intent_classifier.py
Ghostvv: created `RasaTrainChunkException`

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -712,6 +712,16 @@ def __init__(
...
 
+    @lazy_property
+    def nlu_examples(self) -> List[Message]:
+        """Filter training examples that didn't come from nlu data."""
+        return [ex for ex in self.training_examples if not ex.is_core_message()]
+
+    @lazy_property
+    def entity_examples(self) -> List[Message]:
+        """Filter nlu examples that don't contain entities."""
+        return [ex for ex in self.nlu_examples if ex.get(ENTITIES)]
```
tabergma, line 723: I think we can remove those again as `TrainingDataChunk` now inherits from `TrainingData` once you merge `partial-dataset-loading` branch into yours.

tabergma (approved):
Thanks for adding tests 💯

Ghostvv:
@tabergma @dakshvar22 the tests fail due to timeout on `LanguageModelFeaturizer` and `HFTransformersNLP` should I exclude them form the test?

tabergma:
Yeah, I think Felicia is looking into it. But let's add a TODO, so that we don't forget about those. (And I think you can squash merge right after you update the code, don't think you need to wait until the tests pass again as everything else worked.)

Issue #7273, opened 2020-11-13
Divide TrainingData into smaller chunks


tabergma:
related to #6836

In order to do partial dataset training, we need to divide the provided training data into smaller chunks. We need to make sure that one featurized chunk fits into memory. 

Idea: One chunk should have a fixed size, e.g. 1 MB. This value should be configurable. Add a method to `TrainingData`, called `divide_into_chunks(size_in_mb: int)` that returns a list of `TrainingDataChunks`.  

Note: We should make sure that every chunk is a good representation of the complete dataset. E.g. it should not only contain examples of one intent, but rather match the distribution of the complete dataset.

**Note**: This feature should branch off from `partial_dataset_loading` and should be squashed and merged back into that branch.

PR #7352, opened 2020-11-24
Chunk TrainingData into smaller chunks

dakshvar22:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/7273
- Instead of chunk memory, the method receives number of chunks to be returned. Computing the total memory of a complex object like `TrainingData` is [tricky](https://stackoverflow.com/questions/449560/how-do-i-determine-the-size-of-an-object-in-python).
- Make `TrainingDataChunk` a sub-class of `TrainingData`

### DIFF
--- a/rasa/shared/nlu/training_data/training_data.py
+++ b/rasa/shared/nlu/training_data/training_data.py
@@ -745,7 +745,6 @@ class TrainingDataChunk(TrainingData):
         responses: Optional[Dict[Text, List[Dict[Text, Any]]]] = None,
     ) -> None:
         """Initialize a training data chunk."""
-
         if entity_synonyms or regex_features or lookup_tables:
             raise RasaException(
                 f"{self.__class__} cannot have entity synonyms, "
--- a/rasa/shared/nlu/training_data/training_data.py
+++ b/rasa/shared/nlu/training_data/training_data.py
@@ -744,6 +744,7 @@ class TrainingDataChunk(TrainingData):
         lookup_tables: Optional[List[Dict[Text, Any]]] = None,
         responses: Optional[Dict[Text, List[Dict[Text, Any]]]] = None,
     ) -> None:
+        """Initialize a training data chunk."""
 
         if entity_synonyms or regex_features or lookup_tables:
             raise RasaException(
--- a/rasa/shared/nlu/training_data/training_data.py
+++ b/rasa/shared/nlu/training_data/training_data.py
@@ -678,8 +678,6 @@ class TrainingData:
         ]
         return not any([len(lst) > 0 for lst in lists_to_check])
 
-    # def _sizeof__(self):
-
     def divide_into_chunks(self, num_chunks: int) -> List["TrainingDataChunk"]:
         """Divides the training data into smaller chunks.
 
--- a/rasa/shared/nlu/training_data/training_data.py
+++ b/rasa/shared/nlu/training_data/training_data.py
@@ -61,13 +61,6 @@ class TrainingData:
 
         self._fill_response_phrases()
 
-    def __sizeof__(self) -> int:
-        total_size = 0
-        if self.training_examples:
-            for message in self.training_examples:
-                total_size += sys.getsizeof(message)
-        return total_size
-
     def fingerprint(self) -> Text:
         """Fingerprint the training data.
 
@@ -688,7 +681,7 @@ class TrainingData:
 
     # def _sizeof__(self):
 
-    def divide_into_chunks(self, num_chunks: int = 2) -> List["TrainingDataChunk"]:
+    def divide_into_chunks(self, num_chunks: int) -> List["TrainingDataChunk"]:
         """Divides the training data into smaller chunks.
 
         Each chunk should be a good representation of the complete dataset. E.g. it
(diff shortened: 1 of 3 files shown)


rasa/shared/nlu/training_data/training_data.py
```diff
@@ -679,38 +678,54 @@ def can_train_nlu_model(self) -> bool:
...
         Args:
-            max_size: The maximum size (in MB) of one chunk.
+            num_chunks: The total number of chunks into which the training data should be broken.
 
         Returns:
             A list of all training data chunks.
         """
-        pass
+        all_chunks = []
+        data_to_chunk = copy.deepcopy(self)
```
tabergma, line 695: deepcopy doubles the memory as far as I know. Can we somehow avoid it?

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -679,38 +678,54 @@ def can_train_nlu_model(self) -> bool:
...
+                1 - chunk_size_fraction
+            )
 
-    def __init__(
-        self,
-        training_examples: List[Message],
-        responses: Optional[Dict[Text, List[Dict[Text, Any]]]] = None,
-    ) -> None:
-        """Initialize a training data chunk.
+            if chunk_index == num_chunks - 1:
```
tabergma, line 704: I guess we could move this check up and avoid calling `split_nlu_examples` on the last chunk. We could just take all remaining examples directly.

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -679,38 +678,54 @@ def can_train_nlu_model(self) -> bool:
...
+        return all_chunks
 
-        Args:
-            training_examples: List of messages representing the training data.
-            responses: List of responses.
-        """
-        self.training_examples = training_examples
-        self.responses = responses
+
+class TrainingDataChunk(TrainingData):
```
tabergma, line 727: Can we overwrite the `__init__` method to make sure this object has no regexes, lookup tables, etc. set?

tests/shared/nlu/training_data/test_training_data.py
```diff
@@ -633,3 +634,50 @@ def test_fingerprint_is_same_when_loading_data_again():
     td1 = training_data_from_paths(files, language="en")
     td2 = training_data_from_paths(files, language="en")
     assert td1.fingerprint() == td2.fingerprint()
+
+
+@pytest.mark.parametrize(
+    "intent_frequencies, num_chunks", [([100, 82, 63, 43], 8), ([15, 12, 10, 7], 4)]
+)
+def test_divide_training_data_chunks(intent_frequencies, num_chunks):
```
tabergma, line 642: please add types

tests/shared/nlu/training_data/test_training_data.py
```diff
@@ -633,3 +634,50 @@ def test_fingerprint_is_same_when_loading_data_again():
...
+                for ex_index in range(intent_count)
+            ]
+        )
+    training_data = TrainingData(all_messages)
+    original_fingerprint = training_data.fingerprint()
+    chunks = training_data.divide_into_chunks(num_chunks=num_chunks)
+    new_fingerprint = training_data.fingerprint()
+
+    # Original training data shouldn't be modified
+    assert original_fingerprint == new_fingerprint
```
tabergma, line 661: Why? It should not be needed anymore after we created the chunks.

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -679,38 +678,54 @@ def can_train_nlu_model(self) -> bool:
...
-        pass
+        all_chunks = []
+        data_to_chunk = copy.deepcopy(self)
 
+        for chunk_index in range(num_chunks):
 
-class TrainingDataChunk:
-    """Holds a portion of the complete TrainingData."""
+            chunk_size_fraction = 1 / (num_chunks - chunk_index)
+            current_chunk, leftover_examples = data_to_chunk.split_nlu_examples(
```
tabergma, line 700: `split_nlu_examples` might result in a split where one example is in both chunks (`current_chunk, leftover_examples`). I guess that is not a problem.
Also would it be a problem if we just have one example for an intent in a chunk? I guess not as we will train over all chunks in the end.
 Just want to be sure 🙂

tabergma (commented):
Thanks for implementing this 👍  Left a couple comments.

rasa/shared/nlu/training_data/training_data.py
dakshvar22: Would throwing a warning if they are set enough?

federicotdn, line 727: I am super out of context for this unfortunately, but I'm interested into why a `TrainingDataChunk` is a subclass of `TrainingData` rather than `TrainingData` having a reference to a list of `TrainingDataChunk`.

dakshvar22: I am still not super clear on what's the downside if `TrainingDataChunk` have regexes, lookup tables, etc. Is it that methods like `persist_chunk` and `load_chunk` will need to factor them in as well? If yes then sub-classing also doesn't make sense because it's very unusual for a sub-class to not have attributes of the parent class. What do you think? 😅

tabergma: Is more about the memory used by this object. Lookup tables etc. can also take up quite some memory and we don't need them anymore at the point where we split the training data into smaller chunks. So we don't want to store them anymore. I am just afraid, that if we keep the parameters, that people think that they can use them, which we don't want.

dakshvar22: Hmm, okay. Then I'll raise an exception in the overloaded constructor if they are set by the parent constructor. Does that sound good?

tests/shared/nlu/training_data/test_training_data.py
dakshvar22: This is to be sure that we haven't modified the original `TrainingData` instance in the process of dividing into chunks. I would say even if it's not needed, it should be still preserved. What do you think?

rasa/shared/nlu/training_data/training_data.py
dakshvar22: `TrainingDataChunk` will hold a subset of `training_examples` and `responses` on which a lot of helper methods that `TraningData` class has can be used. For e.g., lazy properties.

rasa/shared/nlu/training_data/training_data.py
dakshvar22: > split_nlu_examples might result in a split where one example is in both chunks (current_chunk, leftover_examples)

This shouldn't happen ideally because it is used inside `train_test_split` as well. It would result in train test leakage.

> Also would it be a problem if we just have one example for an intent in a chunk? I guess not as we will train over all chunks in the end.

A chunk can def have one example for an intent but it's only problematic if we plan to call `train` method of an ML component like DIET for each chunk separately. I don't think we intend to do that, right?

tests/shared/nlu/training_data/test_training_data.py
tabergma: ok, makes sense

rasa/shared/nlu/training_data/training_data.py
tabergma: > I don't think we intend to do that, right?

yes, that is not planned. just wanted to make sure that we don't overlook something.

rasa/shared/nlu/training_data/training_data.py
dakshvar22: Removed the copy. `split_nlu_examples` doesn't modify the original data and hence we don't need the copy anymore.

rasa/shared/nlu/training_data/training_data.py
dakshvar22: Removed 👍

rasa/shared/nlu/training_data/training_data.py
dakshvar22: Added an exception before the parent constructor is even called.

tabergma (approved):
Thanks for addressing all my comments 💯 Looks great!

Issue #7328, opened 2020-11-20
Implement Command-Line interface for finetuning models


wochinge:
Implement the command-line interface for https://github.com/RasaHQ/rasa/issues/6971.

- [x] Implement parsing of `rasa train --finetune --model <path to model to finetune>`
  * `--finetune` is `False` by default
  * `--model` is optional. In case it's not given, Rasa Open Source will try to use the latest model from the `models` directory
- [x] track in telemetry event if `--finetune` was used
- [x] pass necessary information down to the training of the core model and the NLU model

wochinge:
@dakshvar22 Two ideas
* How about using something like `rasa train --finetune <model path>`? Not sure if we can make it work if users don't specify `model path` (e.g. if they want to use the latest model), but we can at least try? 
* Do we need a parameter to quickly override the epochs? E.g. something like `--epoch-fraction`? It seems annoying to me to require two equal configs or change it in between. It also seems to be nicer for the usage with the auto config feature .

dakshvar22:
@wochinge
1. I like the idea if we are able to make it work without users having to specify the model path as well.
2. I like that idea from a UX perspective but I am not sure if a single `epoch-fraction` works well for all components(`DIET`, `ResponseSelector` etc.). I would still say let's add it, if it works it is very useful indeed.

wochinge:
Ok 👍 What should we choose as default for the epoch fraction? We can always add a third argument for the entity epoch fraction in case we see that users are running into issues. We could e.g. have a logic like this then:
```
intent_epoch_fraction = args.epoch_fraction
entity_epoch_fraction = args.epoch_fraction
if args.entity_epoch_fraction:
   entity_epoch_fraction = args.entity_epoch_fraction

```

dakshvar22:
Setting a default here is tricky since we don't know how much amount of data got added. Do we currently/ Should we store the number of data points a particular component was trained on? If yes, then we would know how much extra data each component received and set a default for the epoch fraction inside the component.

Otherwise, we should only use this parameter if the user defined a value for it in the CLI.

PR #7358, opened 2020-11-24
CLI for incremental training

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7328
- implement CLI for incremental training
  - `rasa train --finetune` : will try to finetune model based on latest model in `models`
  - `rasa train --finetune file`: Will try to finetune model based on specified model
- pass `agent` / `interpreter` to Core / NLU training
- track in telemetry if model is trained by finetuning another model
- Others
  - fix deprecated usages of `asyncio.coroutine` in `test_train`
  - move `test_models` module to right package


rasa/model.py
```diff
@@ -515,3 +519,31 @@ async def update_model_with_new_domain(
...
+
+    if previous_model_file is None or not Path(previous_model_file).is_file():
+        return None, None
+
+    with unpack_model(previous_model_file) as unpacked:
+        core, nlu = None, None
+        try:
+            core = Agent.load(unpacked)
+        except ModelNotFound:
+            pass
```
wochinge, line 541: add tests for exception cases

rasa/cli/arguments/train.py
```diff
@@ -10,8 +10,11 @@
 )
 from rasa.shared.constants import DEFAULT_CONFIG_PATH, DEFAULT_DATA_PATH
 
+USE_LATEST_MODEL_FOR_FINE_TUNING = "use latest model for finetuning"
```
wochinge, line 13: How about assigning the value of `out` to the finetune argument right away to avoid having this constant spread around?

rasa/model.py
```diff
@@ -515,3 +515,100 @@ async def update_model_with_new_domain(
...
+    """Retrieves loaded models which can be used for finetuning.
+
+    Args:
+        previous_model_file: Path to model file which should be used for finetuning or
+            a directory in case the latest trained model should be used.
+
+    Returns:
+        Loaded Core model and NLU model.
+    """
+    from rasa.core.interpreter import RasaNLUInterpreter
```
wochinge, line 532: using local import to avoid import cycles

tests/test_model.py
```diff
@@ -418,3 +420,44 @@ async def get_domain() -> Domain:
     actual = Domain.load(tmpdir / DEFAULT_CORE_SUBDIRECTORY_NAME / DEFAULT_DOMAIN_PATH)
```
wochinge, line 420: moved this file because it was in wrong python package

tests/test_train.py
```diff
@@ -358,3 +367,347 @@ def test_train_nlu_autoconfig(
     mocked_get_configuration.assert_called_once()
     _, args, _ = mocked_get_configuration.mock_calls[0]
     assert args[1] == autoconfig.TrainingType.NLU
+
+
+def test_model_finetuning(
```
wochinge, line 372: these tests might be extended for actually testing finetuning 🤔

rasa/cli/arguments/train.py
joejuzl, line 13: Could be `True`?

rasa/cli/train.py
```diff
@@ -88,12 +88,33 @@ def train(args: argparse.Namespace) -> Optional[Text]:
...
+
+
 def train_core(
     args: argparse.Namespace, train_path: Optional[Text] = None
 ) -> Optional[Text]:
+    """Trains a Rasa Core model only.
+
+    Args:
+        args: Command-line arguments to configure training.
+        train_path: Path were trained model but unzipped model should be stored.
```
joejuzl, line 113:
```suggestion
        train_path: Path where trained model but not unzipped model should be stored.
```

...if I understand correclty.

rasa/cli/arguments/train.py
```diff
@@ -147,9 +163,35 @@ def add_model_name_param(parser: argparse.ArgumentParser):
...
+def add_finetune_params(
+    parser: Union[argparse.ArgumentParser, argparse._ActionsContainer]
+) -> None:
+    """Adds parameters for model finetuning."""
+    parser.add_argument(
+        "--finetune",
+        nargs="?",
+        # If the user doesn't specify `--finetune` at all
+        default=None,
+        # If the user user only specifies `--finetune` without an additional path
```
joejuzl, line 184:
```suggestion
        # If the user only specifies `--finetune` without an additional path
```

rasa/model.py
```diff
@@ -515,3 +515,104 @@ async def update_model_with_new_domain(
     domain = await importer.get_domain()
 
     domain.persist(model_path / DEFAULT_DOMAIN_PATH)
+
+
+def get_models_for_finetuning(
+    previous_model_file: Optional[Union[Path, Text]]
+) -> Tuple[Optional["Agent"], Optional["Interpreter"]]:
```
joejuzl, line 522: All the other similar functions in this module return paths rather than actual models, does it make sense to keep it consistent?

joejuzl:
Should this target the integration branch we mentioned before?

tests/test_train.py
```diff
@@ -358,3 +367,347 @@ def test_train_nlu_autoconfig(
...
+    monkeypatch: MonkeyPatch,
+    default_domain_path: Text,
+    default_stories_file: Text,
+    default_stack_config: Text,
+    default_nlu_data: Text,
+    trained_rasa_model: Text,
+):
+    mocked_nlu_training = Mock()
+
+    async def mocked_train_nlu(*args: Any, **kwargs: Any) -> Text:
```
joejuzl, line 383: This is repeated in all the tests so could be extracted into a helper. I did something similar (based on this) in https://github.com/RasaHQ/rasa/pull/7382

tests/test_train.py
```diff
@@ -358,3 +367,347 @@ def test_train_nlu_autoconfig(
...
+        domain=default_domain_path,
+        output=str(tmp_path),
+        model_to_finetune=str(Path(trained_rasa_model).parent),
+        finetuning_epoch_fraction=1,
+    )
+
+    mocked_nlu_training.assert_called_once()
+
+
+def test_model_finetuning_with_latest_model_core(
```
joejuzl, line 560: I guess the tests up to hear are similar enough they could be paramatarized?

joejuzl (changes requested):

rasa/cli/arguments/train.py
```diff
@@ -20,52 +23,63 @@ def set_train_arguments(parser: argparse.ArgumentParser):
...
 
     add_nlu_data_param(parser, help_text="File or folder containing your NLU data.")
 
-    add_num_threads_param(parser)
+    _add_num_threads_param(parser)
 
-    add_model_name_param(parser)
+    _add_model_name_param(parser)
     add_persist_nlu_data_param(parser)
+    add_finetune_params(parser)
```
dakshvar22, line 65: @wochinge Just so I understand correctly - This line ensures that `rasa train nlu finetune` will also work in addition to `rasa train finetune`(currently mentioned in the PR), right?

docs/docs/command-line-interface.mdx
```diff
@@ -80,6 +80,10 @@ these, `rasa train` will fall back to one of these commands by default.
 The name of the model by default is `<timestamp>.tar.gz`. If you want to name your model differently,
 you can specify the name using the `--fixed-model-name` flag.
 
+Use `rasa train finetune` if the model should not be trained from scratch but if you
```
dakshvar22, line 83: Also explain the usage of `--epoch-fraction`.

dakshvar22 (commented):
Took a quick look. Looks good 👍

rasa/cli/arguments/train.py
wochinge: yep, I covered all three cases (`rasa train`, `rasa train core`, `rasa train nlu`)

rasa/model.py
wochinge: Mhm, but then we'd load things repeatedly (e.g. the `interpreter` would be loaded for the core training and for NLU training). What I saw from testing it, is that loading the entire pipeline etc. takes quite a while, that's why I opted for doing it as early as possible.

wochinge: Will do this later 👍

wochinge: done

rasa/model.py
```diff
@@ -515,3 +513,32 @@ async def update_model_with_new_domain(
...
+    """Gets validated path for model to finetune.
+
+    Args:
+        previous_model_file: Path to model file which should be used for finetuning or
+            a directory in case the latest trained model should be used.
+
+    Returns:
+        Path to model archive. `None` if there is no model.
+    """
+    if Path(previous_model_file).is_dir():
```
joejuzl, line 530: `get_latest_model` can handle a file too right?

rasa/model.py
```diff
@@ -515,3 +513,32 @@ async def update_model_with_new_domain(
...
+        logger.debug(
+            f"Trying to load latest model from '{previous_model_file}' for "
+            f"finetuning."
+        )
+        return get_latest_model(previous_model_file)
+
+    if previous_model_file and Path(previous_model_file).is_file():
```
joejuzl, line 537: Should this check if it is _not_ a file or dir?

tests/conftest.py
```diff
@@ -309,3 +321,29 @@ def __init__(
         endpoints_path: Text = "",
     ) -> None:
         super().__init__(tracker_store, event_broker, endpoints_path)
+
+
+def mock_async(
+    monkeypatch: MonkeyPatch, target: Any, name: Text, return_value: Any = None
```
joejuzl, line 327: Just a thought, but this could be a fixture and `monkeypatch` would be injected automatically. Don't feel strongly though.

wochinge: mhm, but that then we couldn't set the `return_value` on the fly 😬

rasa/model.py
wochinge: yep, but then will try to find the latest model in the directory of this file.

tests/conftest.py
joejuzl: I think the pattern would be to have the fixture return a function that already has `monkeypatch` in its closure.

rasa/model.py
wochinge: 👍

tests/conftest.py
wochinge: Mhm, or something like that: https://stackoverflow.com/a/32498408/3429596 . Seems even simpler

wochinge: used the things from above

wochinge:
Ok, everything should work, but we get weird timing problems in the CI 😬

joejuzl (approved):

tests/conftest.py
```diff
@@ -309,3 +321,10 @@ def __init__(
         endpoints_path: Text = "",
     ) -> None:
         super().__init__(tracker_store, event_broker, endpoints_path)
+
+
+class AsyncMock(Mock):
```
joejuzl, line 326: https://docs.python.org/3.8/library/unittest.mock.html#unittest.mock.AsyncMock 😁 
I guess we have to support pre-3.8 though?

wochinge: Exactly 😬

wochinge:
Sorry for not having this merged yet, but things keep on failing the CI which don't do locally 🙈

wochinge:
New but different issues 🤯 This PR is my nemesis 😄 
Screenshot 2020-12-03 at 18 09 20

wochinge:
@joejuzl @dakshvar22 Merged for now so that you can branch off from it. The tests are still flaky in parts which are not related to the changes due to random timeouts. We might need to change the timeout for these test 😬

PR #7364, opened 2020-11-25
 Run training fingerprinting via the CLI

### LINKED ISSUE #6729: Run training fingerprinting via the CLI
**Description of Problem**:

Request to be able to create `fingerprint.json` file via the CLI without having to run `rasa train`. Feature will make it easier to run CI checks against Rasa assistant code changes. 

**Overview of the Solution**:

Exposing a CLI command to generate `fingerprint.json`, something along the lines of `rasa fingerprint` or `rasa train --fingerprint`

**Examples** (if relevant):

**Blockers** (if relevant):

**Definition of Done**:

alwx:
**Proposed changes**:
- adds `rasa train fingerprint` command (part of #6729)


rasa/core/train.py
```diff
@@ -70,6 +70,44 @@ async def train(
     return agent
 
 
+async def save_fingerprint(
+    story_file_path: Text,
```
joejuzl, line 74: what about the other data files e.g. `nlu.yaml`?

alwx: The existing fingerprint function uses the `TrainingDataImporter` to get all the rest, and story file, domain file, and config are required only to load the importer. 

For more details, see 
1) `model_fingerprint` function in `rasa.model` 
2) `load_from_config` in `rasa.shared.importers.importer`

joejuzl: yeah but as far as I am aware `TrainingDataImporter` will only be able to provide `get_nlu_data()` if it was passed an NLU data file to `load_from_config`.

rasa/cli/arguments/train.py
```diff
@@ -57,7 +57,28 @@ def set_train_nlu_arguments(parser: argparse.ArgumentParser):
     add_persist_nlu_data_param(parser)
 
 
-def add_force_param(parser: Union[argparse.ArgumentParser, argparse._ActionsContainer]):
+def set_train_fingerprint_arguments(parser: argparse.ArgumentParser) -> None:
```
wochinge, line 60: Maybe @b-quachtran can shed some light on this:
* the current PR adds a command to create a fingerprint and share it. Users have to manage the persisted fingerprints manually and also compare them manually
* from what I understand, the underlying motivation for this feature is to determine whether a model should be re-trained. From that perspective I'd favor something like a `--dry-run` flag for `rasa train` because
  - less additional code
  - We can actually do the comparison for them as long as they have the last model in the `models` (`out` param) directory and e.g. give different exit codes + a description as output. E.g. exit code `0` means no training necessary, `1` means everything needs to be re-trained, `2` means only `core` needs training and `3` only NLU needs training (this case is gonna be removed with end-to-end)
   - There are some discussions between Ben and me in the [issue](https://github.com/RasaHQ/rasa/issues/6729#issuecomment-729168231) which also would make favor the `--dry-run` solution

what do you think?

alwx: @wochinge  you mean something like `rasa train --only-if-needed` (the name is here only to illustrate the idea)? 

We might do that, either additionally to what's done here or instead of the fingerprint command. It's sort of confusing for me that the description of the issue itself is different but it's not a big deal — we can adapt the code here to the new conditions. However, I would like to also hear @b-quachtran's opinion on that.

wochinge: Nope, rather a `--dry-run` which never trains but only indicates if a training would be performed if the command would be run without `--dry-run

rasa/core/train.py
alwx: That's interesting because the existing fingerprint command doesn't use the NLU data, and the importer that's currently used for fingerprints is always the `CoreDataImporter`. 

But you might be true, and it actually should. Let's return to this once we finish discussing what to do with this command :)

rasa/cli/arguments/train.py
alwx: `FingerprintComparisonResult.is_training_required` also returns true if the responses in the domain should be updated (`FingerprintComparisonResult.nlg` indicates that). Should we also return an error code in this case?
I drafted a PR for it (https://github.com/RasaHQ/rasa/pull/7397), just want to know your opinion before I can add tests and update the code style.

wochinge: > Should we also return an error code in this case?

Yes 👍

wochinge: @b-quachtran Could you please have a short look as well?

alwx: @wochinge @joejuzl nothing is really happening with this PR but I already made a new one: https://github.com/RasaHQ/rasa/pull/7397 could you please check it as well? :)

alwx:
@wochinge @joejuzl nothing is really happening with this PR cause we still got no response from @b-quachtran but I already made another one for `dry-run`, can you please check it as well? https://github.com/RasaHQ/rasa/pull/7397

Let's also decide what option we find better and which one we want to merge — maybe it even makes sense to merge both and give users two options instead of just one.

rasa/cli/arguments/train.py
b-quachtran: This would be for use within a CI pipeline so that a fingerprint can be outputted and diff-ed against previous fingerprints. This customer is running training outside of the runner and they want a way to validate if training needed ahead of that external call so `rasa train --only-if-needed` wouldn't work

b-quachtran:
> @wochinge @joejuzl nothing is really happening with this PR cause we still got no response from @b-quachtran but I already made another one for `dry-run`, can you please check it as well? #7397
> 
> Let's also decide what option we find better and which one we want to merge — maybe it even makes sense to merge both and give users two options (especially since they both can actually be useful)

@alwx Sorry about the delay, just responded in the thread above.

rasa/cli/arguments/train.py
alwx: @wochinge so, do you think we should close this one in favour of https://github.com/RasaHQ/rasa/pull/7397?

wochinge: I'd say so. Or can you think of a good use case for it?

PR #7365, opened 2020-11-25
Update tests for hf_transformers and lm_featurizer

koernerfelicia:
**Proposed changes**:
Fixes RasaHQ/research#96

Proposed solution:
Skip the tests with large model downloads on CI. As of now, this is `bert` with weights `rasa/LaBSE`. For every `bert` test with weights `rasa/LaBSE`, there is a `bert` test with `bert-base-uncased`, which **can** be run on CI. If we decide to use our [own runners in future](https://github.com/RasaHQ/rasa-x/issues/3071) this could change and should be revisited.  **Important**: the skipped tests will still be run locally. Wherever possible in tests, we can skip model load by passing: `skip_model_load=True` to the `LanguageModelFeaturizer` constructor.

- [ ] added some tests for the functionality (N/A)
- [ ] updated the documentation (N/A)


koernerfelicia:
Hi @dakshvar22, could you also let me know whether `bugfix` is the correct changelog item? Really wasn't sure about this, maybe it should be `misc`? AFAIK the windows test failed randomly so I'm rerunning them, but I think they should be fine

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -737,7 +437,7 @@ def test_log_deprecation_warning_with_old_config(text: str, caplog: LogCaptureFi
     assert "deprecated component HFTransformersNLP" in caplog.text
 
 
-@pytest.mark.skip(reason="Results in random crashing of github action workers")
+@skip_on_CI
 def test_preserve_sentence_and_sequence_features_old_config():
     attribute = "text"
     message = Message.build("hi there")
```
dakshvar22, line 443: Not sure, but it could be these kind of tests where multiple models are loaded in the same test which is breaking the CI.

tests/nlu/utils/test_hf_transformers.py
```diff
@@ -132,330 +132,3 @@ def test_attention_mask(
 
     assert np.all(mask_ones == 1)
     assert np.all(mask_zeros == 0)
-
-
-# TODO: need to fix this failing test
-@pytest.mark.skip(reason="Results in random crashing of github action workers")
```
dakshvar22, line 138: Correct me if I am wrong but this test shouldn't actually be removed but decorated with `skip_on_ci` decorated and refactored to match the test cases you have constructed for `test_lm_featurizer_shape_values` inside `test_lm_featurizer.py`.

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -21,8 +22,12 @@
 from rasa.nlu.utils.hugging_face.hf_transformers import HFTransformersNLP
 from rasa.shared.nlu.constants import TEXT, INTENT
 
+skip_on_CI = pytest.mark.skipif(
+    bool(os.environ.get("CI")),
+    reason="Downloading model crashes github action" "workers",
```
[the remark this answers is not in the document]
dakshvar22: I think the decorator can be moved to `tests/conftest.py` since it's a general decorator that can be used by other tests too.

dakshvar22 (changes requested):
Thanks a lot for looking into this ✨  Added some suggestions.
Definitely fits into the `misc` category of PRs IMO.
Also, before going ahead with the suggested fix of skipping only on CI, can we try out one more approach of getting around the problem - What if we create separate tests for each model architecture instead of having one test which loads up multiple models as part of different test cases. So for example, have an individual test for BERT based featurizer named `test_bert_featurizer_shape_values` and this test only has test cases for BERT as the featurizer. My hunch is that the CI doesn't clear up the memory across different test cases of the same pytest and hence the machine runs out of memory(7GB is still a good amount of RAM so there is definitely some memory leakage somewhere 😅 ). Since, we have a finite number of model architectures this is still feasible from an implementation standpoint. What do you think?

tests/nlu/featurizers/test_lm_featurizer.py
koernerfelicia: Ahh that's not a bad thought. I will try and figure that out

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -324,6 +332,7 @@ def test_input_padding(
     ],
 )
 @pytest.mark.skip_on_windows
```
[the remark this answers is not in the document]
koernerfelicia: I opted to keep it because I'm not sure why it's there in the first place. Do you know? Figured if Windows user runs it locally it should maybe still be skipped? Not sure

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -366,342 +375,30 @@ def test_attention_mask(
...
+        ("gpt", "sentence embeddings", [1, 2]),
+        ("gpt", "this is a test", [1, 1, 1, 1]),
+        ("gpt2", "sentence embeddings", [2, 3]),
+        ("gpt2", "this is a test", [1, 1, 1, 1]),
+        ("xlnet", "sentence embeddings", [1, 3]),
+        ("xlnet", "this is a test", [1, 1, 1, 1]),
+        ("distilbert", "sentence embeddings", [1, 4]),
+        ("distilbert", "this is a test", [1, 1, 1, 1]),
+        ("roberta", "sentence embeddings", [2, 3]),
+        ("roberta", "this is a test", [1, 1, 1, 1]),
```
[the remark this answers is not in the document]
koernerfelicia: Oh! I think this was a misunderstanding on my part -- I removed `test_lm_featurizer_edge_cases` because I thought you said it should be removed in the original problem description:

> Now that our tokenizers don't split tokens into sub-tokens, the test in (1) is redundant.

I thought that was weird but did it anyways.. I should've checked with you, sorry.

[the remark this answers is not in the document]
koernerfelicia: So I added this back, but the only thing I think we really care about is the sub-token count. Correct me if I'm wrong? Should I be checking the actual values as well? Seems like this is covered by `shape_values`, but I don't know if these edge cases can cause weird behaviour.

koernerfelicia:
```
sqlite> select ITEM_VARIANT, MEM_USAGE from TEST_METRICS;
test_lm_featurizer_shape_values[bert, texts0, expected_shape0, expected_sequence_vec0, expected_cls_vec0]|5126.85546875
test_lm_featurizer_shape_values[gpt, texts1, expected_shape1, expected_sequence_vec1, expected_cls_vec1]|1374.2734375
test_lm_featurizer_shape_values[gpt2, texts2, expected_shape2, expected_sequence_vec2, expected_cls_vec2]|1852.0703125
test_lm_featurizer_shape_values[xlnet, texts3, expected_shape3, expected_sequence_vec3, expected_cls_vec3]|1845.73046875
test_lm_featurizer_shape_values[distilbert, texts4, expected_shape4, expected_sequence_vec4, expected_cls_vec4]|1419.85546875
test_lm_featurizer_shape_values[roberta, texts5, expected_shape5, expected_sequence_vec5, expected_cls_vec5]|2082.59375
test_sequence_length_overflow_train[20, bert, False]|2075.87109375
test_sequence_length_overflow_train[1000, bert, True]|2075.87109375
test_sequence_length_overflow_train[1000, xlnet, False]|2075.87109375
test_long_sequences_extra_padding[sequence_embeddings0, actual_sequence_lengths0, bert, True]|2075.87109375
test_long_sequences_extra_padding[sequence_embeddings1, actual_sequence_lengths1, xlnet, False]|2075.875
test_long_sequences_extra_padding[sequence_embeddings2, actual_sequence_lengths2, bert, False]|2075.875
test_input_padding[token_ids0, 512, 512, True]|2075.890625
test_input_padding[token_ids1, 512, 512, False]|2075.890625
test_input_padding[token_ids2, 200, 200, False]|2075.890625
test_log_longer_sequence[1000, bert, bert, base, uncased, True]|2076.17578125
test_log_longer_sequence[256, bert, bert, base, uncased, False]|2089.38671875
test_attention_mask[256, 512, 256]|2089.38671875
test_attention_mask[700, 700, 700]|2089.38671875
test_attention_mask[700, 512, 512]|2089.38671875
test_lm_featurizer_edge_cases[bert, None, texts0, expected_number_of_sub_tokens0]|6515.74609375
test_lm_featurizer_edge_cases[bert, bert, base, chinese, texts1, expected_number_of_sub_tokens1]|2109.19921875
test_lm_featurizer_edge_cases[gpt, None, texts2, expected_number_of_sub_tokens2]|1879.0859375
test_lm_featurizer_edge_cases[gpt2, None, texts3, expected_number_of_sub_tokens3]|1858.4609375
test_lm_featurizer_edge_cases[xlnet, None, texts4, expected_number_of_sub_tokens4]|1852.390625
test_lm_featurizer_edge_cases[distilbert, None, texts5, expected_number_of_sub_tokens5]|1647.9921875
test_lm_featurizer_edge_cases[roberta, None, texts6, expected_number_of_sub_tokens6]|2246.171875
test_lm_featurizer_number_of_sub_tokens[bert, sentence embeddings, expected_number_of_sub_tokens0]|6734.82421875
test_lm_featurizer_number_of_sub_tokens[gpt, sentence embeddings, expected_number_of_sub_tokens1]|2251.71875
test_lm_featurizer_number_of_sub_tokens[gpt2, sentence embeddings, expected_number_of_sub_tokens2]|2235.625
test_lm_featurizer_number_of_sub_tokens[xlnet, sentence embeddings, expected_number_of_sub_tokens3]|2230.9296875
test_lm_featurizer_number_of_sub_tokens[distilbert, sentence embeddings, expected_number_of_sub_tokens4]|1908.8984375
test_lm_featurizer_number_of_sub_tokens[roberta, sentence embeddings, expected_number_of_sub_tokens5]|2558.09765625
test_log_deprecation_warning_with_old_config[hi there]|2550.8671875
test_preserve_sentence_and_sequence_features_old_config|2621.8671875
```
Update: did a test with [`pytest-monitor`](https://pytest-monitor.readthedocs.io/en/latest/exploitation.html?highlight=MEM_USAGE#metrics) locally, and found that `bert` seems to be the most expensive (`MEM_USAGE` is in megabytes).. Not sure what to do with this, but maybe it's good to know. Might pull those tests out into their own.

tests/nlu/featurizers/test_lm_featurizer.py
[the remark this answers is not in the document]
dakshvar22: `transformers` is runnable on windows too, so I don't see why they should be skipped. We should remove it and see what happens maybe?

tests/nlu/featurizers/test_lm_featurizer.py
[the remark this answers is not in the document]
dakshvar22: I think testing for sub-token count is good enough.

dakshvar22:
That memory profiling is really cool! Can you also do one with model weights for bert set to `bert-base-uncased`? While I agree that these are not the default model weights and we should test for default model weights, I think for a lot of tests(tokenization based) it doesn't matter which model weights we use. Hence, we might be able to do away with the memory problem partially with that. What do you think?

koernerfelicia:
@dakshvar22 I can do this, but actually I am back to thinking that the error is due to a timeout.
Here's why:
I think the above profiling only really explains why we see this warning from `tensorflow` about the 10% of RAM being used for allocation, but doesn't explain why the tests fail. If I delete my hf cache locally (something like: 
`rm -rf ~/.cache/huggingface/transformers/*`) these tests also fail due to a timeout - downloading bert alone takes ~10mins on my laptop (timeout is set to 120s). Working on a solution for this at the moment, my proposed approach is to download the models in a setup fixture with a longer timeout, but there's at least two kinks to be worked out:

1. tests run in parallel will start the fixture in parallel, which is actually problematic because of competing writes to the cache directory. (I have a [fix for this](https://github.com/pytest-dev/pytest/issues/252#issuecomment-370688629) but want to check that the fix makes sense, and there's not a better one).
2. the fixture takes a long time to run, and will take this long anytime the CI tests are run because the runners get wiped. This may prove to be really annoying to developers.

Once I've verified that this works okay, I'll talk to Mariana and Tomasz about what we can do about these points.

dakshvar22:
@koernerfelicia Sounds good 👍 
FWIW we can definitely increase the timeout of a few tests separately as well. I remember this was already done for some of the tests somewhere 🤔

koernerfelicia:
@tczekajlo could you take a look at this one? I think that the `prepare-transformers` target might be partially failing sometimes (as in, only some models are downloaded successfully), leading to flaky tests

tczekajlo:
> @tczekajlo could you take a look at this one? I think that the `prepare-transformers` target might be partially failing sometimes (as in, only some models are downloaded successfully), leading to flaky tests

It looks good to me, in the logs, you can see that everything was downloaded and stored in the `/home/runner/.cache/torch/transformers` directory or the `C:/Program Files/Git/.cache/torch/transformers/` on Window, e.g.

`C:/Program Files/Git/.cache/torch/transformers/642cba239b8eca22b702e71e92d507b8af47ddd2df74dc7751e2a4f65d8d434c.f26918df904593cca2dbd78a3bb760376f4f8ce1ce8d3b13bb6ab28228ee65fc.h5`

Are you sure that the tests look for data in the correct place?

koernerfelicia:
@tczekajlo, that's a good point, thank you! I thought the default would be that location, but I'm pointing to them explicitly now, we'll see if that works.

Where in the logs do you see that the models got downloaded? I'm not quite sure where to look

tczekajlo:
> @tczekajlo, that's a good point, thank you! I thought the default would be that location, but I'm pointing to them explicitly now, we'll see if that works.
> 
> Where in the logs do you see that the models got downloaded? I'm not quite sure where to look

The 'Install Dependencies` step includes the logs, so you can find them there. You can use "Search logs" in the CI workflows to find a part that you're looking for

Might be helpful: https://docs.github.com/en/actions/managing-workflow-runs/using-workflow-run-logs

koernerfelicia:
Hi @dakshvar22, I know it's been forever, but I think I've finally found an acceptable solution:
1. Models are downloaded in the "Install Dependencies" step, with the `prepare-transformers` target. This did not work for `bert` with `LaBSE`, I think because it's very large, but seems to work for all the others.
2. The tests with `bert` and `LaBSE` are skipped on CI, but run locally.

dakshvar22:
@koernerfelicia Thanks a lot for putting in so much effort into this!
> This did not work for bert with LaBSE, I think because it's very large, but seems to work for all the others.

I haven't looked in the code yet, but can we use a different set of model weights other than LabSE for bert model? Since we know that the model architecture is the same it's okay to test it with different model weights for the sake of tests. What do you think?

koernerfelicia:
@dakshvar22 yeah, that's a good idea, we're downloading `bert-base-uncased` for some tests anyways, I can use those

koernerfelicia:
@dakshvar22 I added those tests, and updated the changelog. I think this is pretty much ready to go.

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -224,6 +260,9 @@ def test_lm_featurizer_shape_values(
         assert np.allclose(
             computed_sentence_vec[0][:5], expected_cls_vec[index], atol=1e-5
         )
+        """
+        print(computed_sequence_vec[: len(expected_sequence_vec[index]), 0])
+        print(computed_sentence_vec[0][:5])
```
dakshvar22, line 265:
```suggestion
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -214,6 +249,7 @@ def test_lm_featurizer_shape_values(
         assert computed_sentence_vec.shape[1] == expected_shape[index][1]
 
         # Look at the value of first dimension for a few starting timesteps
+        """
```
dakshvar22, line 252: This part should be uncommented I guess?

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -624,84 +496,76 @@ def test_attention_mask(
...
+                "crashing of GH action workers."
+            )
     else:
         model_weights_config = {"model_weights": model_weights}
-    transformers_config = {**{"model_name": model_name}, **model_weights_config}
+    transformers_config = {
+        **{"model_name": model_name},
+        **model_weights_config,
+        "cache_dir": HF_TEST_CACHE_DIR,
+    }
```
dakshvar22, line 534: Should we move this logic into a separate function since it's reused across multiple tests?

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -624,84 +496,76 @@ def test_attention_mask(
...
+
+        assert [
+            t.get(NUMBER_OF_SUB_TOKENS) for t in message.get(TOKENS_NAMES[TEXT])
+        ] == expected_number_of_sub_tokens[i]
 
 
 @pytest.mark.parametrize(
-    "text, expected_number_of_sub_tokens",
-    [("sentence embeddings", [1, 4]), ("this is a test", [1, 1, 1, 1])],
+    "model_name, text, expected_number_of_sub_tokens",
```
dakshvar22, line 552: Add `model_weights` here too where in one test it is set to `None` and in another it is set to `bert-base-uncased`?

tests/nlu/utils/test_hf_transformers.py
```diff
@@ -134,328 +138,236 @@ def test_attention_mask(
...
 
-    if model_weights is None:
-        model_weights_config = {}
-    else:
-        model_weights_config = {"model_weights": model_weights}
-    transformers_config = {**{"model_name": model_name}, **model_weights_config}
+    config = {"model_name": model_name, "cache_dir": HF_TEST_CACHE_DIR}
+
+    if model_weights:
+        config["model_weights"] = model_weights
```
dakshvar22, line 339: Same, could be separated out in a function.

dakshvar22 (commented):
We are getting close 🚀  A few code refactoring comments.

data/test/hf_transformers_models.txt
```diff
@@ -4,31 +4,51 @@ https://s3.amazonaws.com/models.huggingface.co/bert/rasa/LaBSE/special_tokens_ma
 99497d78492c90ab7d824d695b9a8d043369fbc2bf1112dcc7cdef9a6c4fa691.275045728fbf41c11d3dae08b8742c054377e18d92cc7b72b6351152a99b64e4
 https://s3.amazonaws.com/models.huggingface.co/bert/rasa/LaBSE/tokenizer_config.json
 527f618330e845c9d31826e7d9ce983aa816fafcf4f29f8c52f8ae1fdd097219.1c61d5d3dc67d88e0c74c64cda9b17bc30bdbd1c373cceeb740b9953729709aa
+https://s3.amazonaws.com/models.huggingface.co/bert/rasa/LaBSE/config.json
+90984a8da5021905af8679644b61bc5428ef16e9a307469152c163ec873db240.f1ba7080a92fc164a144311742f36dfb6a724bc9da532264b30d87040e15cc9d
```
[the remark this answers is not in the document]
koernerfelicia: I was unsure about this, I felt like it would be misleading to not have them in there since people might run the `prepare-transformers` target **locally** and expect it to prepare for all of the tests. I've adjusted the target so it doesn't download those files on the CI

rasa/nlu/constants.py
```diff
@@ -82,3 +83,5 @@
 COMPONENT_INDEX = "index"
 MIN_ADDITIONAL_REGEX_PATTERNS = 10
 MIN_ADDITIONAL_CVF_VOCABULARY = 1000
+
+HF_TEST_CACHE_DIR = f"{os.environ['HOME']}/.cache/torch/transformers"
```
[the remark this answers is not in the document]
koernerfelicia: Yes, at least it is on OSX. I thought this was causing problems on CI, so I named it explicitly. Now I'm not sure if that was the problem. I'll test it again.

tests/nlu/featurizers/test_lm_featurizer.py
koernerfelicia: Yes 🤦‍♀️

tests/nlu/utils/test_hf_transformers.py
koernerfelicia: I separated this out, but since it's really only relevant for this and `LMfeaturizer` I didn't put it into conftest. I know this is unconventional, but I'd like to avoid confusion. Once `HFTransformer` is deprecated, the method will definitely belong just in `LMFeaturizer`

koernerfelicia: Open for counter arguments, though :)

rasa/nlu/constants.py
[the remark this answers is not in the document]
koernerfelicia: Looks like we can scratch this!

tests/nlu/utils/test_hf_transformers.py
dakshvar22: Makes perfect sense 👍

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -366,10 +391,33 @@ def test_attention_mask(
...
 
 
-# TODO: need to fix this failing test
-@pytest.mark.skip(reason="Results in random crashing of github action workers")
+def skip_on_CI(model_name, model_weights):
+    # This only applies when skip_model_load=False
+    return model_name == "bert" and bool(os.environ.get("CI")) and not model_weights
+
+
+def create_config(model_name, model_weights):
```
dakshvar22, line 399: `create_config` is a bit general. Could we rename this to something like `create_config_for_lm_featurizer`?

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -366,10 +391,33 @@ def test_attention_mask(
     assert np.all(mask_zeros == 0)
 
 
-# TODO: need to fix this failing test
-@pytest.mark.skip(reason="Results in random crashing of github action workers")
+def skip_on_CI(model_name, model_weights):
```
dakshvar22, line 394: argument types and return type missing

tests/nlu/featurizers/test_lm_featurizer.py
dakshvar22: Also return types and argument types

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -366,10 +391,33 @@ def test_attention_mask(
...
+
+    Args:
+        model_name: model name
+        model_weights: model weights name
+    """
+    if skip_on_CI(model_name, model_weights):
+        pytest.skip(
+            "Reason: this model is too large, loading it results in"
+            "crashing of GH action workers."
+        )
```
dakshvar22, line 412: Will calling this in a separate function also skip the pytest?

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -624,85 +504,60 @@ def test_attention_mask(
...
-        tokens = whitespace_tokenizer.tokenize(message, TEXT)
-        message.set(TOKENS_NAMES[TEXT], tokens)
-        lm_featurizer.process(message)
 
-        assert [t.text for t in tokens] == gt_tokens
-        assert [t.start for t in tokens] == [i[0] for i in gt_indices]
-        assert [t.end for t in tokens] == [i[1] for i in gt_indices]
+        td = TrainingData([message])
+        whitespace_tokenizer.train(td)
+        lm_featurizer.train(td)
```
dakshvar22, line 538: Better to may be test both `train` and `process`?

dakshvar22 (commented):
Types to be added and one additional comment which we can discuss.

tests/nlu/featurizers/test_lm_featurizer.py
koernerfelicia: You're right, it is vague! I'm not sure about tying it to `lm_featurizer` since it's also used (for now) by `HFTransformersNLP`. Do you thunk that's confusing or maybe not such a big deal? What do you think of `create_transformers_config`?

tests/nlu/featurizers/test_lm_featurizer.py
koernerfelicia: Yep! I was sceptical of this too, but you can see in the logs four tests are skipped (three for `LanguageModelFeaturizer` and one for `HFTransformersNLP`). You can also test this locally for yourself if you just add the line `os.environ["CI"] = "true"`

tests/nlu/featurizers/test_lm_featurizer.py
koernerfelicia: Yes, you're right it is vague! I'm not sure about tying it to `LMFeaturizer`, I wonder if that'd be confusing for the time it's used in `HFTransformersNLP`? Or maybe it's not a big deal? I think maybe `create_pretrained_transformers_config` could work. It's long but since it's just a test helper function..

tests/nlu/featurizers/test_lm_featurizer.py
koernerfelicia: Oh boy you're gonna regret asking for that one ;). I think I found an okay way to do it that doesn't result in too much code duplication but it's a bit of a refactor and the diff looks even more so

dakshvar22: Haha! Thanks for looking into it. I'll review it end of today or maybe Monday morning latest.

koernerfelicia: Okay, no rush!

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -184,57 +270,257 @@
...
+        [
+            ("bert", None, "sentence embeddings", [1, 2]),
+            ("bert", "bert-base-uncased", "sentence embeddings", [1, 4]),
+            ("gpt", None, "sentence embeddings", [1, 2]),
+            ("gpt2", None, "sentence embeddings", [2, 3]),
+            ("xlnet", None, "sentence embeddings", [1, 3]),
+            ("distilbert", None, "sentence embeddings", [1, 4]),
+            ("roberta", None, "sentence embeddings", [2, 3]),
+        ],
+    )
```
dakshvar22, line 464: I know this is ancient code, but how about we merge the `general_case_args` and `edge_case_args` into one argument set? I don't see a huge gain in separating out the two. Moreover, it duplicates the code below. What do you think?

dakshvar22: Also, then we can just parameterize the class, rather than parameterizing each function separately.

dakshvar22: Also, then we can just parameterize the class, rather than parameterizing each function separately.

dakshvar22 (commented):
🚀  Sorry, the refactor pointed out one more thing that we could clean up. Lmk what you think 😬

tests/nlu/featurizers/test_lm_featurizer.py
koernerfelicia: Sounds good

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -243,7 +529,7 @@ def test_lm_featurizer_shape_values(
 )
 def test_sequence_length_overflow_train(
     input_sequence_length: int, model_name: Text, should_overflow: bool
-):
+) -> None:
```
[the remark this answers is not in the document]
koernerfelicia: Good point!

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -184,57 +270,257 @@
...
-    for text in texts:
-        messages.append(Message.build(text=text))
-    td = TrainingData(messages)
+class TestShapeValuesTrainAndProcess:
+    def evaluate_message_shape_values(
+        self,
+        messages: List[Message],
+        expected_shape: List[tuple],
+        expected_sequence_vec: List[List[float]],
+        expected_cls_vec: List[List[float]],
```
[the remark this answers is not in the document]
koernerfelicia: Yep, and so can all the other evaluation methods!

dakshvar22 (approved):
Thanks for putting so much effort into this and addressing all my comments! ✨

PR #7368, opened 2020-11-25
Slack init message

koszta5:
**Proposed changes**:
- Allow slack channel to support welcome message pattern by forwarding to _init_ intent automatically when a user interaction with chatbot starts

- [] added some tests for the functionality
- [✅] updated the documentation
- [✅ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [✅] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

### DIFF
--- a/changelog/7366.feature.md
+++ b/changelog/7366.feature.md
@@ -1,0 +1,2 @@
+Allow to reaction of rasa chatbot to slack opening / switching window of conversation. This allows the chatbot to provide 
+a "welcome message" to the user.
--- a/docs/docs/connectors/slack.mdx
+++ b/docs/docs/connectors/slack.mdx
@@ -143,7 +143,6 @@ your bot and tell you about new messages. If you are running locally, you can
    time).
 
 
-
 Your bot is now ready to go and will receive webhook notifications about new messages.
 ## Optional: Interactive Components
 After you've completed [Sending Messages](./slack.mdx#sending-messages) and
--- a/docs/docs/connectors/slack.mdx
+++ b/docs/docs/connectors/slack.mdx
@@ -136,6 +136,13 @@ your bot and tell you about new messages. If you are running locally, you can
    sending of messages, you'll be prompted to **reinstall your app** which you will
    need to do. Otherwise, Slack will confirm your change with a **Success!**)
 
+## Optional: Initial message
+   If you want to have the bot greet message appear every time when user opens its chat (e.g. for the first time) you want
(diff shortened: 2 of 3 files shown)


sara-tagger:

tmbo (commented):
Thanks a lot for your contribution. After thinking about it, I am not sure we should put the burden of figuring out if this really is an "init" on the user. 

We already have the `session_started` event that we emit if a new conversation is started which you can react to in a custom action and take things from there - I am not sure what different use case this would be used for, can you add some thoughts?

koszta5:
hi @tmbo - yes I m aware of that. But you are kind of proving my point by your reply.
> we have `session_started` event that we emit if a new conversation is started
If you read the article that I provided (and I can provide more as a proof) the good default behavior of chatbot should be that it **actively** send the 1st message to user.

Which current solution using `session_started` is incapable of. This solution will only react to user's first message. Which is putting the burden on user having him/her understand what the chatbot can do or having to ask _the obvious_ question _what can you do for me?_

The burden of having to figure out the "init" of the user is not that hard. Bottom line - this behavior is optional and this fact is clearly documented in changed doc files.

akelad:
@tmbo any updates on this? :D

tmbo:
I agree that there is a use case for this and that it is earlier than what we currently call "SessionStarted". I do think that we need a more general approach to this though, that works across channels.

I don't think using a special intent is the right approach for this for different reasons:
- you'd be able to create stories that start with `init` and some that don't, but what is the difference?
- to generalize this and to be able to get access to as much information about the user as possible, you'd want to have access to more meta information when writing this initial message. You might want to look up the user in a DB and respond with user-specific information, e.g. "Hey there Sara, you got 3 bills awaiting payment, do you want me to make a transfer?"
- backwards-incompatible (at least in its current implementation)

I think we should at least think about more generalisable/alternative approaches to support this use case, e.g.
- add an event that gets triggered before a user initializes a conversation, on chat window open
- move the sessionstarted event to get triggered earlier

koszta5:
@tmbo - I like your ideas - they make sense to me.

> I think we should at least think about more generalisable/alternative approaches to support this use case, e.g.
> - add an event that gets triggered before a user initializes a conversation, on chat window open
> -  move the sessionstarted event to get triggered earlier

which of those 2 makes more sense from your point of view? I could understand both but maybe for the purpose of compatibility there could be a new event called sessioninitiated? 

Something like that? When I get some time I would look into it and try to implement this (I hope that I find the right place in code :P )

Issue #7130, opened 2020-10-28
Make sure the e2e stories are printed correctly


Ghostvv:
with e2e we brought one more complexity to stories. We need to make sure that e2e stories are printed correctly for tracker from real conversations. By correctly, I mean that `UserUttered` featurization flag and `ActionExecuted` `action_name` or `action_text` should be taken into account when printing stories

wochinge:
seems @tabergma fixed large parts of the printing already. However, I didn't find any tests and a couple of edge cases weren't covered. There are also weird things regarding end-to-end **test** stories in the YAML writer which confuse me. My current assumption is that the implementation for end-to-end tests on master is buggy / broken.

PR #7388, opened 2020-11-26
E2e story printing

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7130
- fix printing of end-to-end bot actions in `rasa interactive`
- fix failing tests due to the `DEFAULT_STORIES_FILE` changed to be yaml
- fix reading and writing of yaml stories with end-to-end tests and end-to-end training data
- raise exception if end-to-end `UserUttered` / `ActionExecuted` events are tried to be printed to markdown
- change fingerprinting to use yaml to make sure it works with end-to-end training data (fingerprinting is still not stable for checkpoints - however, should be fixed on `master` and not here
- change story structure validator to use `str` to represent actions instead of markdown representation which uses `as_story_string`

### DIFF
--- a/rasa/core/training/story_conflict.py
+++ b/rasa/core/training/story_conflict.py
@@ -196,6 +196,7 @@ def _find_conflicting_states(
     # represented by its hash
     state_action_mapping = defaultdict(list)
     for element in _sliced_states_iterator(trackers, domain, max_history):
+        # TODO: Hash or just store as dict
         hashed_state = element.sliced_states_hash
         if element.event.as_story_string() not in state_action_mapping[hashed_state]:
             state_action_mapping[hashed_state] += [element.event.as_story_string()]
--- a/rasa/core/training/story_conflict.py
+++ b/rasa/core/training/story_conflict.py
@@ -196,10 +196,10 @@ def _find_conflicting_states(
     # represented by its hash
     state_action_mapping = defaultdict(list)
     for element in _sliced_states_iterator(trackers, domain, max_history):
-        # TODO: Hash or just store as dict
         hashed_state = element.sliced_states_hash
-        if element.event.as_story_string() not in state_action_mapping[hashed_state]:
-            state_action_mapping[hashed_state] += [element.event.as_story_string()]
+        current_hash = hash(element.event)
+        if current_hash not in state_action_mapping[hashed_state]:
+            state_action_mapping[hashed_state] += [current_hash]
 
     # Keep only conflicting `state_action_mapping`s
     return {
@@ -239,8 +239,7 @@ def _build_conflicts_from_states(
                 conflicts[hashed_state] = StoryConflict(element.sliced_states)
 
             conflicts[hashed_state].add_conflicting_action(
-                action=element.event.as_story_string(),
-                story_name=element.tracker.sender_id,
+                action=str(element.event), story_name=element.tracker.sender_id,
             )
 
     # Return list of conflicts that arise from unpredictable actions
--- a/rasa/shared/core/events.py
+++ b/rasa/shared/core/events.py
@@ -410,7 +410,7 @@ class UserUttered(Event):
 
     def __hash__(self) -> int:
         """Returns unique hash of object."""
-        return hash((self.text, self.intent_name, jsonpickle.encode(self.entities)))
+        return hash(json.dumps(self.as_sub_state()))
 
     @property
     def intent_name(self) -> Optional[Text]:
--- a/rasa/shared/core/events.py
+++ b/rasa/shared/core/events.py
@@ -434,9 +434,19 @@ class UserUttered(Event):
 
     def __str__(self) -> Text:
         """Returns text representation of event."""
+        entities = ""
+        if self.entities:
+            entities = [
+                f"{entity[ENTITY_ATTRIBUTE_VALUE]} "
+                f"(Type: {entity[ENTITY_ATTRIBUTE_TYPE]}, "
+                f"Role: {entity.get(ENTITY_ATTRIBUTE_ROLE)}, "
+                f"Group: {entity.get(ENTITY_ATTRIBUTE_GROUP)}"
+                for entity in self.entities
+            ]
+            entities = f", entities: {', '.join(entities)}"
+
         return (
-            f"UserUttered(text: {self.text}, intent: {self.intent}, "
-            f"entities: {self.entities})"
+            f"UserUttered(text: {self.text}, intent: {self.intent_name}" f"{entities}))"
         )
 
     @staticmethod
@@ -513,7 +523,7 @@ class UserUttered(Event):
         except KeyError as e:
             raise ValueError(f"Failed to parse bot uttered event. {e}")
 
-    def _entity_string(self):
+    def _entity_string(self) -> Text:
         if self.entities:
             return json.dumps(
                 {
--- a/rasa/shared/core/events.py
+++ b/rasa/shared/core/events.py
@@ -515,6 +515,7 @@ class UserUttered(Event):
         return f"{self.intent_name or ''}{self._entity_string()}"
 
     def apply_to(self, tracker: "DialogueStateTracker") -> None:
+        """Applies event to tracker. See docstring of `Event`."""
         tracker.latest_message = self
         tracker.clear_followup_action()
 
@@ -1220,10 +1221,15 @@ class ActionExecuted(Event):
 
         super().__init__(timestamp, metadata)
 
-    def __str__(self) -> Text:
+    def __repr__(self) -> Text:
+        """Returns event as string for debugging."""
         return "ActionExecuted(action: {}, policy: {}, confidence: {})".format(
             self.action_name, self.policy, self.confidence
         )
+
+    def __str__(self) -> Text:
+        """Returns event as human readable string."""
+        return self.action_name or self.action_text
 
     def __hash__(self) -> int:
         """Returns unique hash for action event."""
@@ -1241,6 +1247,7 @@ class ActionExecuted(Event):
             return equal
 
     def as_story_string(self) -> Text:
+        """Returns event in Markdown format."""
         if self.action_text:
             raise UnsupportedFeatureException(
                 "Printing end-to-end bot utterances is not supported in the "
--- a/rasa/shared/core/events.py
+++ b/rasa/shared/core/events.py
@@ -433,6 +433,7 @@ class UserUttered(Event):
         )
 
     def __str__(self) -> Text:
+        """Returns text representation of event."""
         return (
             f"UserUttered(text: {self.text}, intent: {self.intent}, "
             f"entities: {self.entities})"
@@ -653,29 +654,34 @@ class BotUttered(Event):
         )
 
     def __hash__(self) -> int:
+        """Returns unique hash for event."""
         return hash(self.__members())
 
     def __eq__(self, other) -> bool:
+        """Compares object with other object."""
         if not isinstance(other, BotUttered):
             return NotImplemented
 
         return self.__members() == other.__members()
 
     def __str__(self) -> Text:
+        """Returns text representation of event."""
         return "BotUttered(text: {}, data: {}, metadata: {})".format(
             self.text, json.dumps(self.data), json.dumps(self.metadata)
         )
 
     def __repr__(self) -> Text:
+        """Returns text representation of event for debugging."""
         return "BotUttered('{}', {}, {}, {})".format(
             self.text, json.dumps(self.data), json.dumps(self.metadata), self.timestamp
         )
 
     def apply_to(self, tracker: "DialogueStateTracker") -> None:
-
+        """Applies event to current conversation state."""
         tracker.latest_bot_utterance = self
 
     def as_story_string(self) -> None:
+        """Skips representing the event in stories."""
         return None
 
     def message(self) -> Dict[Text, Any]:
@@ -697,6 +703,7 @@ class BotUttered(Event):
 
     @staticmethod
     def empty() -> "BotUttered":
+        """Creates an empty bot utterance."""
         return BotUttered()
 
     def as_dict(self) -> Dict[Text, Any]:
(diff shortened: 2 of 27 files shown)


tests/shared/core/training_data/story_writer/test_yaml_story_writer.py
```diff
@@ -180,3 +183,116 @@ def test_yaml_writer_stories_to_yaml(default_domain: Domain):
     assert isinstance(result, OrderedDict)
     assert "stories" in result
     assert len(result["stories"]) == 1
+
+
+def test_writing_end_to_end_stories(default_domain: Domain):
```
wochinge, line 188: do the same for md

wochinge: we don't have to care about reading
writing should work

wochinge: actually writing e2e for md should raise

rasa/shared/core/events.py
```diff
@@ -489,17 +489,32 @@ def _entity_string(self):
...
+        """Return event as string for Markdown training format.
+
+        Args:
+            e2e: `True` if the the event should be printed in the format for
+                end-to-end conversation tests.
+
+        Returns:
+            Event as string.
+        """
+        if self.use_text_for_featurization:
```
wochinge, line 501: @Ghostvv I think printing an end-to-end `UserUttered` event in the conversation test format will just make things worse and more complicated. I'd vote to fail early. What do you think? 
I checked the usage and we should be fine except some stuff for the story validation tool which for some reason uses this.

tests/shared/core/training_data/story_writer/test_yaml_story_writer.py
wochinge, line 188: test roundtrip

rasa/shared/core/events.py
Ghostvv: is it used only in markdown, then ok

Ghostvv: I'd prefer yaml writer to also use method inside the event then everything is in one place

wochinge, line 501: fix fingerprinting by using yaml story reader

wochinge:
Rasa X story printing will be broken

rasa/shared/core/events.py
wochinge: Joe and I just had quite a session about it and the fact that `as_story_string` is used for varying use cases (fingerprinting, printing stories, story validation) actually makes changing code quite complicated. I prefer having the representation of an event in a specific format separated out. How an event is featurized / unique represent should be part of the `Event` , but it shouldn't be related to any markdown stuff.

rasa/shared/core/events.py
```diff
@@ -1211,15 +1221,22 @@ def __init__(
...
-    def __str__(self) -> Text:
+    def __repr__(self) -> Text:
+        """Returns event as string for debugging."""
         return "ActionExecuted(action: {}, policy: {}, confidence: {})".format(
             self.action_name, self.policy, self.confidence
         )
 
+    def __str__(self) -> Text:
+        """Returns event as human readable string."""
+        return self.action_name or self.action_text
```
wochinge, line 1232: background: [str vs repo](https://stackoverflow.com/a/2626364/3429596)

My idea is that we can use `str` to get a human readable output for things like the story validator and that we limit `as_story_string` to it's usage for dumping markdown.

rasa/test.py
```diff
@@ -137,6 +137,7 @@ def test_core(
             "Rasa model and provide it via the '--model' argument."
         )
 
+    # TODO: test for classifier
```
wochinge, line 140: better comment would have been: `test for real classifier` as we currently get the `KeywordClassifier`

rasa/shared/core/events.py
joejuzl: Sounds good to me.

Also, on a separate topic, taking a quote from that article:

> This means, in simple terms: almost every object you implement should have a functional \_\_repr\_\_ that’s usable for understanding the object.

I think this ^ should be a python code convention.

rasa/shared/core/events.py
```diff
@@ -1211,15 +1223,22 @@ def __init__(
...
     def __hash__(self) -> int:
-        return hash(self.action_name)
+        """Returns unique hash for action event."""
+        return hash(self.action_name or self.action_text)
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: Any) -> bool:
+        """Checks if object is equal to another."""
         if not isinstance(other, ActionExecuted):
             return False
```
joejuzl, line 1243: I know not changed in this PR, but as we were discussing dunder methods above...
When the other is a different class we should actually _return_ `NotImplemented` (not raise it, or return `False`).
This allows the interpreter to do clever things like try the reflected \_\_eq\_\_ operation.

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -231,7 +232,7 @@ def parse_e2e_message(line: Text, is_used_for_training: bool = True) -> Message:
         intent = match.group(2)
         message = match.group(4)
         example = entities_parser.parse_training_example(message, intent)
-        if not is_used_for_training:
+        if not is_used_for_training and not self.use_e2e:
```
joejuzl, line 235: when is `parse_e2e_message` used but `elf.use_e2e` `False`?

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -389,7 +372,7 @@ def _user_intent_from_step(
     ) -> Tuple[Text, Optional[Text]]:
         user_intent = step.get(KEY_USER_INTENT, "").strip()
 
-        if not user_intent:
+        if not user_intent and KEY_USER_MESSAGE not in step:
```
joejuzl, line 375: Does the rest of this method gracefully handle there being no intent?

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
```diff
@@ -216,12 +211,20 @@ def process_user_utterance(
...
         if user_utterance.text and (
+            # We only print the utterance text if it was an end-to-end prediction
             user_utterance.use_text_for_featurization
-            or user_utterance.use_text_for_featurization is None
+            # or if we want to print a conversation test story.
+            or is_test_story
         ):
-            result[KEY_USER_MESSAGE] = user_utterance.as_story_string()
+            result[KEY_USER_MESSAGE] = LiteralScalarString(
+                rasa.shared.core.events.md_format_message(
```
joejuzl, line 220: `md` stands for markdown?

rasa/shared/core/training_data/structures.py
```diff
@@ -407,7 +407,11 @@ def fingerprint(self) -> Text:
         Returns:
             fingerprint of the stories
         """
-        self_as_string = self.as_story_string()
+        from rasa.shared.core.training_data.story_writer.yaml_story_writer import (
+            YAMLStoryWriter,
+        )
+
+        self_as_string = YAMLStoryWriter().dumps(self.story_steps)
```
joejuzl, line 414: So this will still not be stable with `or` steps right? But we will fix on master?

tests/shared/core/test_events.py
```diff
@@ -507,3 +509,20 @@ def test_events_begin_with_session_start(
         rasa.shared.core.events.do_events_begin_with_session_start(test_events)
         == begin_with_session_start
     )
+
+
+@pytest.mark.parametrize(
+    "end_to_end_event",
+    [
+        ActionExecuted(action_text="I insist on using Markdown"),
```
joejuzl, line 517: 😂

rasa/shared/core/events.py
wochinge: thanks! TIL!

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
wochinge: We have tests for it and they pass 🤷‍♂️

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
wochinge: when converting stories from Markdown to YAML `core.training.converters.test_story_markdown_to_yaml_converter.test_test_stories` was failing

joejuzl:
LGTM

wochinge:
@Ghostvv Would you mind reviewing?

rasa/core/training/story_conflict.py
```diff
@@ -197,8 +197,9 @@ def _find_conflicting_states(
     state_action_mapping = defaultdict(list)
     for element in _sliced_states_iterator(trackers, domain, max_history):
         hashed_state = element.sliced_states_hash
-        if element.event.as_story_string() not in state_action_mapping[hashed_state]:
-            state_action_mapping[hashed_state] += [element.event.as_story_string()]
+        current_hash = hash(element.event)
```
Ghostvv, line 200: if you use `str()` as a substitute for `element.event.as_story_string()` should it be here `str` as well instead of hash?

rasa/shared/core/events.py
```diff
@@ -377,27 +409,31 @@ def _from_parse_data(
...
+        ) == (
+            other.text,
+            other.intent_name,
+            [jsonpickle.encode(ent) for ent in other.entities],
+        )
 
     def __str__(self) -> Text:
+        """Returns text representation of event."""
         return (
             f"UserUttered(text: {self.text}, intent: {self.intent}, "
```
Ghostvv, line 438: I think it should be
```suggestion
            f"UserUttered(text: {self.text}, intent: {self.intent_name}, "
```
otherwise you get all the confidences etc

rasa/shared/core/events.py
```diff
@@ -489,23 +525,34 @@ def _entity_string(self):
         return ""
 
     def as_story_string(self, e2e: bool = False) -> Text:
```
Ghostvv, line 527: should here be deprecation warning?

Ghostvv (commented):
looks good, but it is better if someone who knows `yaml` writing will take a look

rasa/shared/core/events.py
```diff
@@ -377,27 +409,31 @@ def _from_parse_data(
...
+            other.text,
+            other.intent_name,
+            [jsonpickle.encode(ent) for ent in other.entities],
+        )
 
     def __str__(self) -> Text:
+        """Returns text representation of event."""
         return (
             f"UserUttered(text: {self.text}, intent: {self.intent}, "
             f"entities: {self.entities})"
```
Ghostvv, line 439: that will contain a dict with all the confidences, extractors, etc, should we filter only entity names here?

rasa/core/training/story_conflict.py
wochinge: `as_story_string()` is used as hashable, unique event representation to uniquely de-duplicate / identify events. Directly using `hash` makes the intended behavior clearer in my opinion.

rasa/shared/core/events.py
wochinge: I agree. We should revisit this on `master` once we decided on a Python code convention. Let's focus on getting `e2e` merged and not to fix all problems in the code.

rasa/shared/core/events.py
wochinge: I'd deprecate MD as a whole. I also re-started the discussion on this in the `product` channel in Slack.

rasa/shared/core/events.py
wochinge: see my answer to https://github.com/RasaHQ/rasa/pull/7388#discussion_r537461938

wochinge:
> looks good, but it is better if someone who knows yaml writing will take a look

@Ghostvv I think the changes re YAML are very small in this PR. I think it'd be better to merge this now and then have @degiz review the entire changes on the `e2e` branch. Ok?

Can I merge then?

rasa/core/training/story_conflict.py
Ghostvv: makes sense, but why do we need `str` then? why not use `hash` everywhere?

wochinge: Sometimes we actually want to give the users a nice presentation of the events (e.g. when telling the users which actions are conflicting).

rasa/shared/core/events.py
Ghostvv: what do you mean? I don't think its convention problem. If you use `str` as hash method, you'd get different events all the time, because confidences will be different

wochinge: We don't use `str` as `hash` here. The `hash` actually uses the `intent_name`. `__hash__` is currently not used for hashing on predictions / only on training data.

Ghostvv: then `hash` is wrong, because it contains `self.entities` which contain confidences and extractors etc, which I don't think should be a part of the hash

rasa/shared/core/events.py
```diff
@@ -377,27 +409,31 @@ def _from_parse_data(
         )
 
     def __hash__(self) -> int:
+        """Returns unique hash of object."""
         return hash((self.text, self.intent_name, jsonpickle.encode(self.entities)))
```
Ghostvv, line 413: I don't think `self.entities` here is correct, since it contains a lot of irrelevant unnecessary info

Ghostvv:
I'd add a `TODO`/create separate issue about `self.entities` in `hash` and we can merge it

Ghostvv (approved):

rasa/shared/core/events.py
wochinge: fixed it by using `self.substate` for hashing

rasa/shared/core/events.py
```diff
@@ -377,30 +409,44 @@ def _from_parse_data(
         )
 
     def __hash__(self) -> int:
-        return hash((self.text, self.intent_name, jsonpickle.encode(self.entities)))
+        """Returns unique hash of object."""
+        return hash(json.dumps(self.as_sub_state()))
```
Ghostvv, line 413: ❤️

rasa/shared/core/events.py
wochinge: fixed representation and hashing

PR #7390, opened 2020-11-26
Merge resposnes from NLU and Domain when there're no retrieval intents

degiz:
**Proposed changes**:
- Make sure the `responses` are synced between NLU data and the Domain even if there're no retrieval intents in the training data.


rasa/shared/importers/rasa.py
```diff
@@ -65,7 +65,6 @@ async def get_domain(self) -> Domain:
             return domain
         try:
             domain = Domain.load(self._domain_path)
-            domain.check_missing_templates()
```
dakshvar22, line 68: Shouldn't this check be done somewhere if not here?

dakshvar22 (commented):
Question: Using the `RetrievalModelDataImporter` to achieve this without having any retrieval intents seems weird. Shouldn't the logic of syncing all responses defined in NLU data(irrespective of them being response templates for retrieval intents) be done inside `CombinedDataImporter` itself and then `RetrievalModelDataImporter` adds retrieval intents to domain and fetches the response templates from domain. What do you think?

rasa/shared/importers/rasa.py
degiz: I've added it [here](https://github.com/RasaHQ/rasa/pull/7390/files#diff-663c0521028355e3657687cc698278f0d680619c6897bbe0762fe0e40213143eR328)

dakshvar22:
@degiz I also do not have very strong opinions but renaming it more appropriately(something like - `NLUDomainSyncImporter`) seems like a better approach I think.

degiz:
@dakshvar22 I've change the name! Do you have time to re-review?

dakshvar22 (approved):
Looks good. Thanks for fixing this 💯

PR #7392, opened 2020-11-27
Persist training data chunk

### LINKED ISSUE #7276: Implement 'persist_chunk' in TrainingDataChunk
related to #6836

In order to do partial dataset training we need to write down the featurized chunks. Implement the method `persist_chunk` in `TrainingDataChunk`. The method should write all examples, e.g. labels and features, of the chunk to a [TFRecord](https://www.tensorflow.org/tutorials/load_data/tfrecord#writing_a_tfrecord_file) file. You need to convert each `Message` to a [tf.train.Example](https://tensorflow.google.cn/tutorials/load_data/tfrecord#tftrainexample) and afterwards [write](https://tensorflow.google.cn/tutorials/load_data/tfrecord#tfrecord_files_in_python) them down.

Note: We need to figure out how to handle sparse features.

**Note**: This feature should branch off from `partial_dataset_loading` and should be squashed and merged back into that branch.

tabergma:
**Proposed changes**:
closes #7276 

Implement persist and load of `TrainingDataChunk`.


rasa/shared/nlu/training_data/training_data.py
```diff
@@ -755,6 +759,73 @@ def __init__(
...
+    def _int_feature(array: Union[Tuple[int], np.ndarray]) -> tf.train.Feature:
+        return tf.train.Feature(int64_list=tf.train.Int64List(value=array))
+
+    @staticmethod
+    def _construct_key(
+        attribute: Text, feature_type: Text, origin: Text, is_dense: bool
+    ) -> Text:
+        prefix = (
+            f"{attribute}{TFRECORD_KEY_SEPARATOR}{feature_type}"
+            f"{TFRECORD_KEY_SEPARATOR}{origin}{TFRECORD_KEY_SEPARATOR}"
```
dakshvar22, line 779: Do we need the last `TFRECORD_KEY_SEPARATOR`?

dakshvar22: Ahh I see you are adding a key later. Please ignore the comment.

tabergma: yes, otherwise it would result in something like `text#sequence#spacydense`. much easier to convert it back if it is `text#sequence#spacy#dense`

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -755,6 +759,73 @@ def __init__(
...
+        )
+
+        if is_dense:
+            return f"{prefix}dense"
+        return f"{prefix}sparse"
+
+    @staticmethod
+    def _deconstruct_key(key: Text) -> Tuple[Text, Text, Text, bool, Text]:
```
dakshvar22, line 787: Rename the function to `_deconstruct_tf_record_key` or similar. `_deconstruct_key` is slightly ambiguous.

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -755,6 +759,73 @@ def __init__(
...
+        if isinstance(value, type(tf.constant(0))):
+            value = value.numpy()
+        return tf.train.Feature(bytes_list=tf.train.BytesList(value=[value]))
+
+    @staticmethod
+    def _int_feature(array: Union[Tuple[int], np.ndarray]) -> tf.train.Feature:
+        return tf.train.Feature(int64_list=tf.train.Int64List(value=array))
+
+    @staticmethod
+    def _construct_key(
```
dakshvar22, line 774: Same as below, rename the function to `_construct_tf_record_key`

tests/shared/nlu/training_data/test_training_data.py
```diff
@@ -637,6 +641,36 @@ def test_fingerprint_is_same_when_loading_data_again():
...
+                ),
+            ]
+        )
+    ]
+
+    training_data_chunk = TrainingDataChunk(messages)
+    original_fingerprint = training_data_chunk.fingerprint()
+    file_path = training_data_chunk.persist_chunk(str(tmp_path), "test.tfrecord")
+
+    loaded_training_data_chunk = TrainingDataChunk.load_chunk(file_path)
```
dakshvar22, line 665: Optional: Should we test for the expected keys as well in the persisted tf record file?

dakshvar22 (changes requested):
Looks great 👍 Just a couple of suggestions.
I also haven't worked for a long time on the TFRecord stuff so assuming that's correct, the structure looks good 💯

tests/shared/nlu/training_data/test_training_data.py
tabergma: any suggestion on how we should do that? we cannot just read the content of the file into a string and check if the expected keys exists. also this is kind of already checked as loading would fail if the keys are not present.

dakshvar22: > this is kind of already checked as loading would fail if the keys are not present.

Makes sense, but just to be sure maybe add tests for just `_deconstruct_tf_record_key` and `_construct_tf_record_key` functions since they are static functions?

dakshvar22 (approved):
Thanks for addressing all the comments 💯

PR #7397, opened 2020-11-27
rasa train --dry-run

alwx:
**Proposed changes**:
- adds `rasa train --dry-run` command (part of #6729)


rasa/train.py
```diff
@@ -157,6 +188,18 @@ async def _train_async_internal(
         file_importer.get_stories(), file_importer.get_nlu_data()
     )
 
+    # get the fingerprint
+    new_fingerprint = await model.model_fingerprint(file_importer)
+    old_model = model.get_latest_model(output_path)
+
+    if dry_run:
+        fingerprint_comparison = model.should_retrain(
```
wochinge, line 196: In this case we can drop calling this again in `230` (if `force_training` is `True`, I'd probably return a `3` as exit code?)

wochinge (commented):
looks good 👍

rasa/train.py
```diff
@@ -125,10 +132,32 @@ async def handle_domain_if_not_exists(
...
+def dry_run_result(fingerprint_comparison: FingerprintComparisonResult) -> Tuple[int, Text]:
+    """Returns a result of fingerprint comparison.
+
+    Args:
+        fingerprint_comparison: A result of fingerprint comparison operation.
+
+    Returns:
+        A tuple of code and human-readable text.
+    """
+    if not fingerprint_comparison.nlu and not fingerprint_comparison.core:
```
joejuzl, line 144: There is another scenario where responses are changed in the domain, so the domain will be updated but there will be no training (i.e. just `should_retrain_nlg`). I wonder if it's worth handling this case too?

alwx: Yes, we've been talking about it in another PR already. Gonna be fixed.

rasa/train.py
alwx: Done!

alwx:
Going to add one more test to it, but in general it's done and ready to be reviewed.

rasa/train.py
```diff
@@ -157,6 +209,20 @@ async def _train_async_internal(
...
+
+    fingerprint_comparison = model.should_retrain(
+        new_fingerprint, old_model, train_path, force_training
+    )
+
+    if dry_run:
+        code, texts = dry_run_result(fingerprint_comparison)
+        for text in texts:
+            print_warning(text) if code > 0 else print_success(text)
+        sys.exit(code)
```
joejuzl, line 224: This might be a real PITA, but I wonder if the code should be passed up to `rasa/cli/train.py` and exiting handled there? Just as `rasa/cli/` where we tend to handle things like this.

joejuzl, line 224: This might be a real PITA, but I wonder if the code should be passed up to rasa/cli/train.py and exiting handled there? Just as rasa/cli/ where we tend to handle things like this.
WDYT?

rasa/train.py
```diff
@@ -183,14 +249,7 @@ async def _train_async_internal(
...
 
-    new_fingerprint = await model.model_fingerprint(file_importer)
-    old_model = model.get_latest_model(output_path)
-
-    if not force_training:
-        fingerprint_comparison = model.should_retrain(
-            new_fingerprint, old_model, train_path
-        )
-    else:
+    if force_training:
```
joejuzl, line 252: Isn't this case handled in `should_retrain` already?

tests/cli/test_rasa_train.py
```diff
@@ -194,6 +199,75 @@ def test_train_force(run_in_simple_project_with_model: Callable[..., RunResult])
...
+    with open(os.path.join(temp_dir, "domain.yml"), "w") as f:
+        f.write(domain)
+
+    output = run_in_simple_project_with_model("train", "--dry-run")
+    printed_output = set(output.outlines)
+
+    assert not any([s for s in printed_output if "No training required." in s])
+    assert (
+        (
+            output.ret & CODE_CORE_NEEDS_TO_BE_RETRAINED
```
joejuzl, line 240: this is pretty cool. Is this a conventional way of handling shell exit codes?

alwx: Well, afaik the standard exit code is 0 for success and any number from 1 to 255 for anything else. 
I thought that it might be a good idea to just use bit masks to encode all the information we need because we have 4 booleans in `FingerprintComparisonResult` (which gives us 2^4 possible results).

rasa/train.py
alwx: It's definitely not a PITA, and might actually be a good idea :) 
However, not sure if I have time to check it today — might only check it next Wednesday in the worst case.

rasa/cli/arguments/train.py
```diff
@@ -101,6 +102,24 @@ def add_compare_params(
     )
 
 
+def add_dry_run_param(
+    parser: Union[argparse.ArgumentParser, argparse._ActionsContainer]
+):
```
wochinge, line 107:
```suggestion
) -> None:
```

rasa/cli/arguments/train.py
```diff
@@ -101,6 +102,24 @@ def add_compare_params(
...
+    Args:
+        parser: An instance of `ArgumentParser` or `_ActionsContainer`.
+    """
+    parser.add_argument(
+        "--dry-run",
+        default=False,
+        action="store_true",
+        help="If enabled, not actual training will be performed. Instead, "
+        "it will be determined whether a model should be re-trained "
+        "and this information will be printed as an output.",
```
wochinge, line 119: Should we explain the different exit codes?

rasa/train.py
```diff
@@ -157,6 +209,20 @@ async def _train_async_internal(
         file_importer.get_stories(), file_importer.get_nlu_data()
     )
 
+    # get the fingerprint
```
wochinge, line 212: I think the comment doesn't add much, so I'd omit it

tests/cli/test_rasa_train.py
```diff
@@ -194,6 +199,75 @@ def test_train_force(run_in_simple_project_with_model: Callable[..., RunResult])
...
+        and (output.ret & CODE_FORCED_TRAINING != CODE_FORCED_TRAINING)
+    )
+
+
+def test_train_dry_run_force(
+    run_in_simple_project_with_model: Callable[..., RunResult]
+):
+    temp_dir = os.getcwd()
+
+    assert os.path.exists(os.path.join(temp_dir, "models"))
```
wochinge, line 260: this means that the test is dependent on the previous one which doesn't seem ideal

alwx: Not really — it depends on `run_in_simple_project_with_model` which is specified in `conftest.py`

rasa/cli/arguments/train.py
```diff
@@ -101,6 +102,30 @@ def add_compare_params(
     )
 
 
+def add_dry_run_param(
+    parser: Union[argparse.ArgumentParser, argparse._ActionsContainer]
+) -> None:
+    """Add `--dry-run` argument to a specified `parser`.
```
wochinge, line 108: Because we agreed to use the descriptive style
```suggestion
    """Adds `--dry-run` argument to a specified `parser`.
```

rasa/cli/arguments/train.py
```diff
@@ -101,6 +102,30 @@ def add_compare_params(
...
+
+    Args:
+        parser: An instance of `ArgumentParser` or `_ActionsContainer`.
+    """
+    parser.add_argument(
+        "--dry-run",
+        default=False,
+        action="store_true",
+        help="If enabled, not actual training will be performed. Instead, "
```
wochinge, line 117:
```suggestion
        help="If enabled, no actual training will be performed. Instead, "
```

rasa/cli/arguments/train.py
```diff
@@ -101,6 +102,30 @@ def add_compare_params(
...
+        "it will be determined whether a model should be re-trained "
+        "and this information will be printed as the output. The return "
+        "code is a 4-bit bitmask that can also be used to determine what exactly needs "
+        "to be retrained:\n"
+        "- 0b0001 means Core needs to be retrained\n"
```
wochinge, line 122: Let's just tell them the integer which the bitmask translates to. No need to refresh their knowledge on binary numbers 😂

rasa/train.py
```diff
@@ -20,32 +20,53 @@
...
     DEFAULT_NLU_SUBDIRECTORY_NAME,
 )
 
+CODE_CORE_NEEDS_TO_BE_RETRAINED = 0b0001
+CODE_NLU_NEEDS_TO_BE_RETRAINED = 0b0010
+CODE_NLG_NEEDS_TO_BE_RETRAINED = 0b0100
+CODE_FORCED_TRAINING = 0b1000
+
+
+class TrainingResult:
```
wochinge, line 37: ?
```suggestion
class TrainingResult(NamedTuple):
```

rasa/train.py
```diff
@@ -125,38 +150,91 @@ async def handle_domain_if_not_exists(
...
+
+    Returns:
+        A tuple where the first element is the result code and the second
+        is the list of human-readable texts that need to be printed to the end user.
+    """
+    code = 0
+    texts = []
+
+    if fingerprint_comparison.core:
+        code += CODE_CORE_NEEDS_TO_BE_RETRAINED
```
wochinge, line 169: that's just a really really neat idea 💪

tests/cli/test_rasa_train.py
```diff
@@ -194,6 +199,75 @@ def test_train_force(run_in_simple_project_with_model: Callable[..., RunResult])
...
+    assert output.ret == 0
+
+
+def test_train_dry_run_failure(
+    run_in_simple_project_with_model: Callable[..., RunResult]
+):
+    temp_dir = os.getcwd()
+
+    domain = (
+        "version: '2.0'\n"
```
wochinge, line 222: How about using the constant for 2.0 to avoid that this test causes a warning somewhen in the future?

tests/cli/test_rasa_train.py
```diff
@@ -194,6 +199,75 @@ def test_train_force(run_in_simple_project_with_model: Callable[..., RunResult])
...
+        "actions:\n"
+        "- utter_greet\n"
+        "- utter_cheer_up"
+    )
+
+    with open(os.path.join(temp_dir, "domain.yml"), "w") as f:
+        f.write(domain)
+
+    output = run_in_simple_project_with_model("train", "--dry-run")
+    printed_output = set(output.outlines)
```
wochinge, line 235: I recently debugged one of these tests and the `set` makes it very hard as it shuffles the lines. Hence, I'd prefer keeping the `list` 🙈

changelog/7397.improvement.md
```diff
@@ -0,0 +1,2 @@
+Add `rasa train --dry-run` command that allows to check if training needs to be performed
```
wochinge, line 1: For consistency it might be an idea to also add this to `rasa train core` & `rasa train nlu`, what do you think?

alwx: Well, we can do that but I think it makes very little sense.
In addition, the results of `rasa train core --dry-run` will be different and will basically be just answering a question of "does the core model need to be retrained?" in this case. 

But I don't have a strong opinion on that — on one hand, I understand your argument about the consistency (however, in this case we might also consider adding `--force` argument to `rasa train nlu`) but on the other hand I don't understand what problem does it solve and how it can be useful :)

wochinge: Yeah, I was wondering the same 👍 Let's not cover them then.

rasa/train.py
```diff
@@ -125,38 +164,90 @@ async def handle_domain_if_not_exists(
...
+        texts.append("Core model should be retrained.")
+
+    if fingerprint_comparison.core:
```
wochinge, line 186: 🚨 
```suggestion
    if fingerprint_comparison.nlu:
```

rasa/train.py
```diff
@@ -125,38 +164,90 @@ async def handle_domain_if_not_exists(
...
+        code += CODE_NLU_NEEDS_TO_BE_RETRAINED
+        texts.append("NLU model should be retrained.")
+
+    if fingerprint_comparison.nlg:
+        code += CODE_NLG_NEEDS_TO_BE_RETRAINED
+        texts.append("Responses in the domain should be updated.")
+
+    if fingerprint_comparison.force_training:
+        code += CODE_FORCED_TRAINING
```
wochinge, line 195: doesn't make a big difference, but would that be safer?
```suggestion
        code = CODE_FORCED_TRAINING
```

wochinge (approved):
Please check the comment with the 🚨

rasa/train.py
alwx: Oh wow, thanks for noticing!

joejuzl (approved):
🚀

PR #7399, opened 2020-11-27
IntentTEDPolicy

dakshvar22:
**Proposed changes**:
- Closes https://github.com/RasaHQ/research/issues/122


dakshvar22:
@Ghostvv Do you mind taking a look at the implementation here before I proceed further with the remaining tasks [here](https://github.com/RasaHQ/research/issues/122)? There is still scope for cleaning up parts of code but that's part of polishing the code further.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -375,3 +375,30 @@ def __init__(self, *args, **kwargs) -> None:
...
+
+    def encode_all_labels(
+        self,
+        domain: Domain,
+        interpreter: NaturalLanguageInterpreter,
+        label_tags: List[Text],
+    ) -> List[Dict[Text, List["Features"]]]:
+        """Encode all intents from the domain using the given interpreter.
+        Args:
+            domain: The domain that contains the actions.
```
Ghostvv, line 397:
```suggestion
            domain: The domain that contains the intents.
```

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,857 @@
...
+            [FeatureArray(np.expand_dims(label_ids, -1), number_of_dimensions=2)],
+        )
+
+        return label_data, encoded_all_labels
+
+    @staticmethod
+    def _get_label_key_to_ids_map(labels: List[Text]) -> Dict[Text, int]:
+        return {label: index for index, label in enumerate(labels)}
+
+    def train(
```
Ghostvv, line 321: why do we need to override this?

rasa/utils/tensorflow/layers.py
```diff
@@ -949,3 +949,232 @@ def call(
...
+        label_ids: tf.Tensor,
+        all_labels_embed: tf.Tensor,
+        all_label_ids: tf.Tensor,
+    ) -> Tuple[tf.Tensor, tf.Tensor, tf.Tensor, tf.Tensor]:
+        """Sample negative examples."""
+
+        pos_inputs_embed = tf.expand_dims(
+            inputs_embed, axis=-2, name="expand_pos_input"
+        )
+        # Take only the first label as positive, others may or may not be picked during random sampling
```
Ghostvv, line 980: why only the first one?

Ghostvv (commented):
the overall structure is good, maybe creating a super class IntentPolicy as abstracted class would make sense, than the check in ensemble can be made more general, there are code repetitions in tracker featurizers, not sure how ready the code is.

Why `IntentTEDPolicy` overrides so many methods? can't it only override the featurizer, model loss and prediction?

rasa/utils/tensorflow/layers.py
dakshvar22: The others(some of them) will be picked up during negative sampling.
So, the resultant `pos_neg_labels`(at the end of the function) will contain some 1s and some 0s depending on which data points were sampled during negative sampling.

rasa/core/policies/intent_ted_policy.py
dakshvar22: The threshold calculation is done at the end of this method. But, I think I'll push it to a separate method called `post_training_tasks()` or something which would be inside `TEDPolicy` and overridden by `IntentTEDPolicy`

rasa/utils/tensorflow/layers.py
Ghostvv: but negative sampling is independent, no? We just later create a mask that would set loss from same labels to 0, or do I miss anything?

dakshvar22: We don't set the loss to 0, but actually set the correct label for it.

So, initially if the label_ids passed to this function were - 
```
[[2,3,0,0],
[1,0,0,0],
[3,1,2,0]]
```
(Here `0` is the padding label ID)

and if `neg_ids` are(`num_negs=2`) - 
```
[[1,2],
[0,1],
[2,0]]
```

then `neg_labels_ids` would be - 
```
[[1,1],
[2,1],
[1,2]]
```

and `pos_neg_labels` will become - 
```
[[0,0],
[0,1],
[1,1]]
```

Now, the final `label_ids` that are constructed [here](https://github.com/RasaHQ/rasa/pull/7399/files#diff-dd1545c3aa7b4edad5ad9b0e3f9090a5413e0309af2831e0bfe25b50f644888eR1057) and get passed to `sigmoid_cross_entropy_with_logits` are - 
```
[[1,0,0],
[1,0,1],
[1,1,1]]
```

As you can see the final `label_ids` do not have the same number of positive labels as they were present in the data but the model will definitely see all positive labels in some other batch, due to random sampling. So, the sampling that we do here is not strictly "negative". Does that help clarify?

Ghostvv: hm, I need to spend some time on it 🙈

PR #7402, opened 2020-11-28
Enable convert to yml

koszta5:
**Proposed changes**:
- Add support to enable conversion to YML for NLU file

- [✅ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [✅] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

### DIFF
--- a/changelog/7401.feature.md
+++ b/changelog/7401.feature.md
@@ -1,0 +1,1 @@
+Implementing support for conversion of NLU file to YML
--- a/rasa/nlu/convert.py
+++ b/rasa/nlu/convert.py
@@ -23,10 +23,13 @@ def convert_training_data(
     elif output_format == "md":
         td = rasa.shared.nlu.training_data.loading.load_data(data_file, language)
         output = td.nlu_as_markdown()
+    elif output_format == "yml":
+        td = rasa.shared.nlu.training_data.loading.load_data(data_file, language)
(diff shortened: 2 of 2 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @tabergma will take a look at it as soon as possible ✨

tabergma:
Thanks for submitting the PR @koszta5! We are actually already supporting this, see [here](https://github.com/RasaHQ/rasa/blob/master/rasa/cli/data.py#L305). What is the reason adding this to the other method as well?

koszta5:
OK @tabergma - that makes sense to me, now what does not make sense to me is having 2 methods for the same thing in different places :P 
The place where I made the edit - I originally used this script externally from bash to convert chatito generated json file to MD. I confess I didnt know about the functionality in cli/data.py. 

But that being said - my PR allows json -> yaml conversion. While if using the cli/data.py you would need to do JSON -> MD -> YAML

Care to make any decision how it taken from here? Do you want me to keep both places? Please provide more isight :) I m more than open to adjust things how you think they should be.

degiz:
Hey @koszta5 

Do you mean that `rasa data convert nlu -f yaml` won't convert the `JSON` directly to the `YAML`?

PR #7417, opened 2020-11-30
Correctly set slots in domain when just updating nlg.

joejuzl:
**Proposed changes**:
`Domain.compare_with_specification` was failing when running `rasa train` after only changing NLG data. This is because the categorical slots were not having the default `__other__` value added when the domain was updated using `update_model_with_new_domain` - so it ended up being inconsistent. 
Solution was to also set up the default slots in this method.


Ghostvv (commented):
just checked, works for me

alwx (approved):
Looks good!

PR #7418, opened 2020-11-30
Preserve domain slot ordering and multistrings while dumping

degiz:
Fix for https://github.com/RasaHQ/rasa-x/issues/3900 and https://github.com/RasaHQ/rasa-x-demo/issues/27

**Proposed changes**:
- Preserve ordering of domain slots while dumping
- Preserve multi line strings in responses `text` key

### DIFF
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -52,6 +52,7 @@ KEY_RESPONSES = "responses"
 KEY_ACTIONS = "actions"
 KEY_FORMS = "forms"
 KEY_E2E_ACTIONS = "e2e_actions"
+KEY_RESPONSES_TEXT = "text"
 
 ALL_DOMAIN_KEYS = [
     KEY_SLOTS,
@@ -286,8 +287,8 @@ class Domain:
         slots = []
         # make a copy to not alter the input dictionary
         slot_dict = copy.deepcopy(slot_dict)
-        # Sort the slots so that the order of the slot states is consistent
-        for slot_name in sorted(slot_dict):
+        # Don't sort the slots, see https://github.com/RasaHQ/rasa-x/issues/3900
+        for slot_name in slot_dict:
             slot_type = slot_dict[slot_name].pop("type", None)
             slot_class = Slot.resolve_by_type(slot_type)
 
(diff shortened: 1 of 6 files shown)


rasa/shared/core/domain.py
```diff
@@ -3,6 +3,7 @@
 import json
 import logging
 import os
+import textwrap
```
wochinge, line 6:
```suggestion
```

rasa/shared/core/domain.py
```diff
@@ -1045,12 +1047,39 @@ def as_dict(self) -> Dict[Text, Any]:
...
             KEY_ACTIONS: self._custom_actions,  # class names of the actions
             KEY_FORMS: self.forms,
             KEY_E2E_ACTIONS: self.action_texts,
         }
 
+    @staticmethod
+    def get_responses_with_multilines(
+        responses: Dict[Text, List[Dict[Text, Any]]]
+    ) -> Dict[Text, List[Dict[Text, Any]]]:
+        """Return `responses` with preserved multilines in the `text` key.
```
wochinge, line 1060:
```suggestion
        """Returns `responses` with preserved multilines in the `text` key.
```

rasa/shared/core/domain.py
```diff
@@ -1045,12 +1047,39 @@ def as_dict(self) -> Dict[Text, Any]:
             KEY_INTENTS: self._transform_intents_for_file(),
             KEY_ENTITIES: self._transform_entities_for_file(),
             KEY_SLOTS: self._slot_definitions(),
-            KEY_RESPONSES: self.templates,
+            KEY_RESPONSES: self.get_responses_with_multilines(self.templates),
```
wochinge, line 1050: I think we should call this in `as_yaml`. What do you think? Feels a bit weird in the regular `as_dict` function which is not necessarily used for dumping.

tests/shared/core/test_domain.py
```diff
@@ -1142,3 +1145,63 @@ def test_valid_slot_mappings(domain_as_dict: Dict[Text, Any]):
 def test_form_invalid_mappings(domain_as_dict: Dict[Text, Any]):
     with pytest.raises(InvalidDomain):
         Domain.from_dict(domain_as_dict)
+
+
+def test_slot_order_is_preserved():
```
wochinge, line 1150: Can we please also add a test where we merge domains and then make sure that the slot order still makes sense? Merging is usually the part where things go wrong in this module 🙈

wochinge (commented):
Nice, concise fix! 

Should this rather go into `2.1.x`?

rasa/shared/core/domain.py
degiz: I need to make sure to read our style guide  every two weeks 🤦

wochinge (approved):

Issue #7421, opened 2020-12-01
Lookup table doesn't support Chinese and other un-whitespace-splittable languages correctly


howl-anderson:
**Rasa version**: 2.1.1

**Python version**: 3.7.6

**Operating system** (windows, osx, ...): Ubuntu 18.04.5 LTS

**Issue**:
When the user provides a lookup table, such as: 

```text
上海
北京
```

It will turn to the regex expression `(\\b上海\\b|\\b北京\\b)` eventually. 

Unfortunately, Chinese and other un-whitespace-splittable languages usually don't have whitespace inside the text, so `\\b` usually doesn't match.

### Indeed it works when there is a whitespace inside
```python
result = list(re.finditer("(\\b上海\\b|\\b北京\\b)", "明天 上海 的天气", flags=0))
result
```
it returns:
```text
[<_sre.SRE_Match object; span=(3, 5), match='上海'>]
```

### But it doesn't work when there is no whitespace inside
```python
result = list(re.finditer("(\\b上海\\b|\\b北京\\b)", "明天上海的天气", flags=0))
result
```
it returns:
```text
[]
```

I will submit a PR to fix this.

sara-tagger:
Thanks for raising this issue, @wochinge will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

wochinge:
fixed by https://github.com/RasaHQ/rasa/pull/7422

Thanks @howl-anderson 🚀

Issue #7091, opened 2020-10-23
Chinese or other non-whitespace(maybe) language to use regex ,such as lookup tables


w7lau:
https://github.com/RasaHQ/rasa/blob/6fbee0dbd67abeccb5544b4f3d53d26f636c5c0c/rasa/nlu/extractors/regex_entity_extractor.py#L89
change to
`matches = re.finditer(pattern["pattern"].replace('\\b',''), message.get(TEXT), flags=flags)`

https://github.com/RasaHQ/rasa/blob/6fbee0dbd67abeccb5544b4f3d53d26f636c5c0c/rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py#L131
change to
`matches = re.finditer(pattern["pattern"].replace('\\b',''), message.get(TEXT), flags=flags)`

I test in windows 10,python 3.7.6,rasa 2.0.2
maybe when get the config language=='zh' can ues this?

sara-tagger:
Thanks for submitting this feature request 🚀 @TyDunn will get back to you about it soon! ✨

tabergma:
@w7lau Thanks for raising the issue! I think you might be right and we should add the word boundary to the regex for non-whitespace tokenized language, such as Chinese. 

I think we should modify the code in https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/utils/pattern_utils.py. We could maybe pass on the language and modify the regex depending on the language given. 

@w7lau Do you have an example sentence and regex to test the behaviour? Do you want to submit a PR for this?

w7lau:
> @w7lau Thanks for raising the issue! I think you might be right and we should not add the word boundary to the regex for non-whitespace tokenized language, such as Chinese.
> 
> I think we should modify the code in https://github.com/RasaHQ/rasa/blob/master/rasa/nlu/utils/pattern_utils.py. We could maybe pass on the language and modify the regex depending on the language given.
> 
> @w7lau Do you have an example sentence and regex to test the behaviour? Do you want to submit a PR for this?

Ok，I will try to do this.

tabergma:
@w7lau Great! Let me know if you need help.

w7lau:
@tjpalmer Is example correct like https://github.com/w7lau/rasa-Chinese-test ?
And the pr at https://github.com/RasaHQ/rasa/pull/7107 
But I don't know why the pr checks is failed...

wochinge:
fixed by https://github.com/RasaHQ/rasa/pull/7422

Issue #5508, opened 2020-03-27
[Diet Classifier] ValueError: Number of examples should be the same for all data.


robinsongh381:
**Rasa version**: 1.9.2

**Rasa SDK version** (if used & relevant):

**Rasa X version** (if used & relevant):

**Python version**:
3.6
**Operating system** (windows, osx, ...):
linux
**Issue**:
When training rasa nlu (i.e. rasa nlu train) there is an error from rasa/utils/tensorflow/model_data.py line 107

**Error (including full traceback)**:
```
2020-03-27 13:13:06 INFO     rasa.nlu.model  - Starting to train component tokenizer_whitespace
2020-03-27 13:13:09 INFO     rasa.nlu.model  - Finished training component.
2020-03-27 13:13:09 INFO     rasa.nlu.model  - Starting to train component RegexFeaturizer
2020-03-27 13:13:09 INFO     rasa.nlu.model  - Finished training component.
2020-03-27 13:13:09 INFO     rasa.nlu.model  - Starting to train component CountVectorsFeaturizer
2020-03-27 13:13:09 INFO     rasa.nlu.model  - Finished training component.
2020-03-27 13:13:09 INFO     rasa.nlu.model  - Starting to train component DIETClassifier
Traceback (most recent call last):
  File "/home/gunsu/diet/bin/rasa", line 8, in <module>
    sys.exit(main())
  File "/home/gunsu/diet/lib/python3.6/site-packages/rasa/__main__.py", line 91, in main
    cmdline_arguments.func(cmdline_arguments)
  File "/home/gunsu/diet/lib/python3.6/site-packages/rasa/cli/train.py", line 140, in train_nlu
    persist_nlu_training_data=args.persist_nlu_data,
  File "/home/gunsu/diet/lib/python3.6/site-packages/rasa/train.py", line 414, in train_nlu
    persist_nlu_training_data,
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/home/gunsu/diet/lib/python3.6/site-packages/rasa/train.py", line 445, in _train_nlu_async
    persist_nlu_training_data=persist_nlu_training_data,
  File "/home/gunsu/diet/lib/python3.6/site-packages/rasa/train.py", line 474, in _train_nlu_with_validated_data
    persist_nlu_training_data=persist_nlu_training_data,
  File "/home/gunsu/diet/lib/python3.6/site-packages/rasa/nlu/train.py", line 86, in train
[...]

Issue #1515, opened 2018-11-07
Rasa_nlu returns intent as null for training samples using tensorflow classifier(Chinese) 


pfZhu:
**Rasa NLU version**: 0.13.0

**Operating system** (windows, osx, ...): ubuntu 18.04

**Content of model configuration file**:
```yml
language: "zh"
project: "ivr_nlu"
fixed_model_name: "demo"
path: "models"
pipeline:
- name: "nlp_mitie"
  model: "data/total_word_feature_extractor_zh.dat"
- name: "tokenizer_jieba"
  dictionary_path: "data/userdict.txt"
- name: "intent_featurizer_count_vectors"
- name: "intent_classifier_tensorflow_embedding"
- name: "ner_mitie"
- name: "ner_synonyms"
```

**Issue**:
The language of my system is Chinese. Most of the training samples are classified correctly, but a few samples return an intent as null. The model failed to classify the sample “在吗” which means “are you there” and “是” which means “yes”. Could the model “data/total_word_feature_extractor_zh.dat” affect intent classification? Besides, it didn't happen when I used spacy_sklearn as intent classifier. What is wrong with tf embedding or my configuration?

akelad:
Thanks for raising this issue, @MetcalfeTom  will get back to you about it soon.

pfZhu:
> Hi @pfZhu,
> 
> Null intents are a result of out-of-vocabulary words when using the tensorflow embeddings. Are these samples in your training data?
> 
> In general, I have a couple architecture questions - are you following [this repo's advice?](https://github.com/crownpku/Rasa_NLU_Chi) I would recommend using `intent_classifier_mitie` for this purpose because `total_word_feature_extractor_zh.dat` was trained under the MITIE framework. It is possible it has incompatible features with tensorflow which are causing your null intents.



> Hi @pfZhu,
> 
> Null intents are a result of out-of-vocabulary words when using the tensorflow embeddings. Are these samples in your training data?
> 
> In general, I have a couple architecture questions - are you following [this repo's advice?](https://github.com/crownpku/Rasa_NLU_Chi) I would recommend using `intent_classifier_mitie` for this purpose because `total_word_feature_extractor_zh.dat` was trained under the MITIE framework. It is possible it has incompatible features with tensorflow which are causing your null intents.

Thanks @MetcalfeTom ,
[...]

PR #7422, opened 2020-12-01
Add use_word_boundaries option for RegexFeaturizer and RegexEntityExtractor

howl-anderson:
**Proposed changes**:
- Add use_word_boundaries option for RegexFeaturizer to control the lookup table regex generation

**Proposal**

Need the user specifies the option (`use_word_boundaries=False`) manually, it will equal to:

```python
result = list(re.finditer("(上海|北京)", "明天上海的天气", flags=0))
print(result)
```

result:
```text
[<_sre.SRE_Match object; span=(2, 4), match='上海'>]
```

fixs #7421 
fixs #7091

### DIFF
--- a/docs/docs/components.mdx
+++ b/docs/docs/components.mdx
@@ -2064,6 +2064,9 @@ The `SpacyEntityExtractor` extractor does not provide a `confidence` level and w
   Make the entity extractor case sensitive by adding the `case_sensitive: True` option, the default being
   `case_sensitive: False`.
 
+  For correctly process languages such as Chinese that don't use whitespace for word separation,
+  user need add the `use_word_boundaries: False` option, the default being `use_word_boundaries: True`.
+
   ```yaml-rasa
       pipeline:
       - name: RegexEntityExtractor
@@ -2073,6 +2076,8 @@ The `SpacyEntityExtractor` extractor does not provide a `confidence` level and w
         use_lookup_tables: True
         # use regexes to extract entities
         use_regexes: True
+        # use match word boundaries for lookup table
+        "use_word_boundaries": True
   ```
 
 
--- a/docs/docs/components.mdx
+++ b/docs/docs/components.mdx
@@ -824,8 +824,8 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
   Make the featurizer case insensitive by adding the `case_sensitive: False` option, the default being
   `case_sensitive: True`.
 
-  For correctly process languages such as Chinese that don't use whitespace for word separation,
-  user need add the `use_word_boundaries: False` option, the default being `use_word_boundaries: True`.
+  To correctly process languages such as Chinese that don't use whitespace for word separation,
+  the user needs to add the `use_word_boundaries: False` option, the default being `use_word_boundaries: True`.
 
   ```yaml-rasa
   pipeline:
@@ -2064,8 +2064,8 @@ The `SpacyEntityExtractor` extractor does not provide a `confidence` level and w
   Make the entity extractor case sensitive by adding the `case_sensitive: True` option, the default being
   `case_sensitive: False`.
 
-  For correctly process languages such as Chinese that don't use whitespace for word separation,
-  user need add the `use_word_boundaries: False` option, the default being `use_word_boundaries: True`.
+  To correctly process languages such as Chinese that don't use whitespace for word separation,
+  the user needs to add the `use_word_boundaries: False` option, the default being `use_word_boundaries: True`.
 
   ```yaml-rasa
       pipeline:
--- a/docs/docs/components.mdx
+++ b/docs/docs/components.mdx
@@ -824,11 +824,16 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
   Make the featurizer case insensitive by adding the `case_sensitive: False` option, the default being
   `case_sensitive: True`.
 
+  For correctly process languages such as Chinese that don't use whitespace for word separation,
+  user need add the `use_word_boundaries: False` option, the default being `use_word_boundaries: True`.
+
   ```yaml-rasa
   pipeline:
   - name: "RegexFeaturizer"
     # Text will be processed with case sensitive as default
     "case_sensitive": True
+    # use match word boundaries for lookup table
+    "use_word_boundaries": True
   ```
 
 
--- a/rasa/nlu/extractors/regex_entity_extractor.py
+++ b/rasa/nlu/extractors/regex_entity_extractor.py
@@ -33,6 +33,8 @@ class RegexEntityExtractor(EntityExtractor):
         "use_lookup_tables": True,
         # use regexes to extract entities
         "use_regexes": True,
+        # use match word boundaries for lookup table
+        "use_word_boundaries": True,
     }
 
     def __init__(
@@ -40,6 +42,7 @@ class RegexEntityExtractor(EntityExtractor):
         component_config: Optional[Dict[Text, Any]] = None,
         patterns: Optional[List[Dict[Text, Text]]] = None,
     ):
+        """Extracts entities using the lookup tables and/or regexes defined in the training data."""
         super(RegexEntityExtractor, self).__init__(component_config)
 
         self.case_sensitive = self.component_config["case_sensitive"]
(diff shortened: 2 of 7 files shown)


sara-tagger:

akelad:
@tabergma figured you knows this area of the codebase better

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
```diff
@@ -42,14 +42,16 @@ def required_components(cls) -> List[Type[Component]]:
         "use_lookup_tables": True,
         # use regexes to generate features
         "use_regexes": True,
+        # use match word boundaries for lookup table
+        "use_word_boundaries": True,
```
tabergma, line 46: Should we also add this to the `RegexEntityExtractor`?

rasa/nlu/utils/pattern_utils.py
```diff
@@ -108,8 +120,9 @@ def extract_patterns(
     use_lookup_tables: bool = True,
     use_regexes: bool = True,
     use_only_entities: bool = False,
+    use_word_boundaries: bool = True,
 ) -> List[Dict[Text, Text]]:
-    """Extract a list of patterns from the training data.
+    r"""Extract a list of patterns from the training data.
```
tabergma, line 125:
```suggestion
    """Extract a list of patterns from the training data.
```

rasa/nlu/utils/pattern_utils.py
```diff
@@ -21,19 +26,23 @@ def _convert_lookup_tables_to_regex(
...
         patterns.append(lookup_regex)
     return patterns
 
 
-def _generate_lookup_regex(lookup_table: Dict[Text, Union[Text, List[Text]]]) -> Text:
-    """Creates a regex pattern from the given lookup table.
+def _generate_lookup_regex(
+    lookup_table: Dict[Text, Union[Text, List[Text]]], use_word_boundaries: bool = True
+) -> Text:
+    r"""Creates a regex pattern from the given lookup table.
```
tabergma, line 38:
```suggestion
    """Creates a regex pattern from the given lookup table.
```

rasa/nlu/utils/pattern_utils.py
```diff
@@ -6,13 +6,18 @@
 
 
 def _convert_lookup_tables_to_regex(
-    training_data: TrainingData, use_only_entities: bool = False
+    training_data: TrainingData,
+    use_only_entities: bool = False,
+    use_word_boundaries: bool = True,
 ) -> List[Dict[Text, Text]]:
-    """Convert the lookup tables from the training data to regex patterns.
+    r"""Convert the lookup tables from the training data to regex patterns.
```
tabergma, line 13:
```suggestion
    """Convert the lookup tables from the training data to regex patterns.
```

tabergma (changes requested):
Thanks for submitting the PR 🚀 

Didn't you tackle this already in the past? Or was that only for 1.10.x?

Code looks good 💯 Left a couple of comments. However, we need to update the [documentation](https://github.com/RasaHQ/rasa/blob/master/docs/docs/components.mdx), add a changelog entry and also it would be great if you could add a test.

howl-anderson:
> Didn't you tackle this already in the past? Or was that only for 1.10.x?

@tabergma I didn't deal with this issue before (this is the first time that I review the code of lookup table), but indeed I dealt with a few issues that similar to this but the root cause is not the same (#5508, #1515).

rasa/nlu/utils/pattern_utils.py
howl-anderson: We need add a prefix `r` here, because without it, `make lint` will report `D301 Use r""" if any backslashes in a docstring`

rasa/nlu/utils/pattern_utils.py
howl-anderson: ditto

rasa/nlu/utils/pattern_utils.py
howl-anderson: ditto

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
howl-anderson: Yes, you are right, I missing this part, I will add more commits for it.

howl-anderson:
> Code looks good 100 Left a couple of comments. However, we need to update the documentation, add a changelog entry and also it would be great if you could add a test.

Yes, I will update the documentation, changelog entry and test cases in later commmits.

rasa/nlu/utils/pattern_utils.py
tabergma: ok, wasn't aware of that

docs/docs/components.mdx
```diff
@@ -824,11 +824,16 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
...
 
+  To correctly process languages such as Chinese that don't use whitespace for word separation,
+  the user needs to add the `use_word_boundaries: False` option, the default being `use_word_boundaries: True`.
+
   ```yaml-rasa
   pipeline:
   - name: "RegexFeaturizer"
     # Text will be processed with case sensitive as default
     "case_sensitive": True
+    # use match word boundaries for lookup table
```
tabergma, line 835:
```suggestion
    # use match word boundaries for regexes and lookup table
```

rasa/nlu/extractors/regex_entity_extractor.py
```diff
@@ -33,13 +33,16 @@ class RegexEntityExtractor(EntityExtractor):
         "use_lookup_tables": True,
         # use regexes to extract entities
         "use_regexes": True,
+        # use match word boundaries for lookup table
```
tabergma, line 36:
```suggestion
        # use match word boundaries for lookup tables and/or regexes
```

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
```diff
@@ -42,14 +42,16 @@ def required_components(cls) -> List[Type[Component]]:
         "use_lookup_tables": True,
         # use regexes to generate features
         "use_regexes": True,
+        # use match word boundaries for lookup table
```
tabergma, line 45:
```suggestion
        # use match word boundaries for regexes and lookup tables

```

docs/docs/components.mdx
```diff
@@ -2068,6 +2076,8 @@ The `SpacyEntityExtractor` extractor does not provide a `confidence` level and w
         use_lookup_tables: True
         # use regexes to extract entities
         use_regexes: True
+        # use match word boundaries for lookup table
```
tabergma, line 2079:
```suggestion
        # use match word boundaries for lookup tables and/or regexes
```

tabergma (approved):
Great work 🚀 Thanks for tackling all my comments 💯

koernerfelicia:
Hi @howl-anderson, thank you for your great work on this! However, I think we do not need the word boundary at all (even for whitespace-tokenized languages), and this would fix the original problem (sorry that the timing is slightly unfortunate). I think your tests look good, though, and of course I'd like you to get credit for them.

It would be great if we can either pull them into a new PR in which the word boundary is scrapped, or you adjust this one. What do you think?

docs/docs/components.mdx
howl-anderson: If the `regexes` in here means `training_data.regex_features`, as far as I know, `RegexFeaturizer` and `RegexEntityExtractor` use `training_data.regex_features` without modification, if so, word boundaries option will not affect the `regexes`.

koernerfelicia:
@howl-anderson the changes I've proposed would look like [so](https://github.com/RasaHQ/rasa/pull/7455), except I would like to also include your tests.

howl-anderson:
Hi @koernerfelicia, I am not sure it's a perfect idea that we drop the word boundary at all, maybe we need more time to discuss it and make sure there are no drawbacks. My suggestion is to merge this PR first since it is a bugfix and it is ready. Then when we decide that we need to drop the word boundary, I am happy to do some works for that change, we can release that feature in a later version.

koernerfelicia:
>My suggestion is to merge this PR first since it is a bugfix and it is ready.

Okay, sounds good!

>I am not sure it's a perfect idea that we drop the word boundary at all

If you have thoughts on this I'd be interested to hear them.

howl-anderson:
Hi @koernerfelicia 

> If you have thoughts on this I'd be interested to hear them.

I'm not a linguist, but I am wondering why the initial implementation using the `\b`, surely the implementation tries to fix something that can not be done without it. Maybe usually people want to match "A B C" instead of "A BC" by using "A B" as the pattern? You know I'm not a native speaker of English,  that is a bit of thought from my humble opinion.

PR #7423, opened 2020-12-01
Use response selector keys for confusion matrix labelling

hotzenklotz:
**Proposed changes**:
- At the moment, when running an NLU evaluation for response selectors the plotted confusion matrix with use the training data utterances as plot labels. This small PR uses the response selector keys and predicted (sub)-intents as labels instead. This make the confusion matrix a lot more readable and useful.
- Similarly, the `response_selection_report.json` uses the same labels leading to better comprehension.

**Examples** 
Before 
response_selection_confmat_old

After

response_selection_confmat copy

### DIFF
--- a/rasa/nlu/test.py
+++ b/rasa/nlu/test.py
@@ -31,6 +31,7 @@ from rasa.nlu.constants import (
     ENTITY_ATTRIBUTE_TYPE,
     ENTITY_ATTRIBUTE_GROUP,
     ENTITY_ATTRIBUTE_ROLE,
+    RESPONSE_KEY_ATTRIBUTE
 )
 from rasa.model import get_model
 from rasa.nlu import config, training_data, utils
@@ -62,7 +63,7 @@ IntentEvaluationResult = namedtuple(
 
 ResponseSelectionEvaluationResult = namedtuple(
     "ResponseSelectionEvaluationResult",
-    "intent_target " "response_target " "response_prediction " "message " "confidence",
+    "intent_target response_key response_target response_prediction_full_intent response_prediction message confidence",
 )
 
 EntityEvaluationResult = namedtuple(
@@ -373,7 +374,7 @@ def evaluate_response_selections(
     )
 
     target_responses, predicted_responses = _targets_predictions_from(
-        response_selection_results, "response_target", "response_prediction"
+        response_selection_results, "response_key", "response_prediction_full_intent"
     )
 
     if report_folder:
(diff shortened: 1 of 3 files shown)


sara-tagger:

rasa/nlu/test.py
```diff
@@ -1050,12 +1051,24 @@ def get_eval_data(
...
             ).get(OPEN_UTTERANCE_PREDICTION_KEY, {})
 
+            response_prediction_full_intent = selector_properties.get(
+                response_prediction_key, {}
+            ).get("full_retrieval_intent", {})
+
+            if isinstance(response_prediction_full_intent, str):
+                response_prediction_full_intent = response_prediction_full_intent.split(
+                    "/"
+                )[1]
```
dakshvar22, line 1061: I would suggest not splitting the predicted `retrieval_intent/sub_intent` because you could have same `sub_intent` under multiple retrieval intents. So, let's compare for e.g. `faq/ask_name` to `faq/ask_weather` and not `ask_name` to `ask_weather`.

rasa/nlu/test.py
```diff
@@ -1050,12 +1051,24 @@ def get_eval_data(
...
+
             response_target = example.get("response", "")
+            response_key = example.get(RESPONSE_KEY_ATTRIBUTE, "")
```
dakshvar22, line 1064: Following from the above comment, this would then change to `example.get_combined_intent_response_key()`

rasa/nlu/test.py
```diff
@@ -62,7 +63,7 @@
 
 ResponseSelectionEvaluationResult = namedtuple(
     "ResponseSelectionEvaluationResult",
-    "intent_target " "response_target " "response_prediction " "message " "confidence",
+    "intent_target response_key response_target response_prediction_full_intent response_prediction message confidence",
```
dakshvar22, line 66: It looks like with this change we can also get rid of `intent_target`, `response_target` and `response_prediction`. Can you scrub those off as well? There will be some more places where you will have to make changes.

dakshvar22 (changes requested):
Thanks for fixing this on `1.10.x` branch 👍  I suggested a few changes and based on them there are a few more places where you will have to make changes. For example, line 412-414(return actual and predicted labels instead of actual and predicted texts), line 1488(use actual and predicted labels instead of actual and predicted texts to compute metrics), line 241(not filter response examples based on empty texts).
All of these changes are actually implemented on master `already` if you would like to view them for reference.

hotzenklotz:
@dakshvar22 Thanks for the feedback. I applied your suggestions. Please see commit 021bcb1

hotzenklotz:
@dakshvar22 Is anything else required for this PR?

dakshvar22 (approved):
Thanks for fixing this and addressing all my comments. ✨

hotzenklotz:
@dakshvar22 To my dismay I discovered that my refactoring reintroduced a bug in my original PR. Unfortunately, it was captured by any of the unit the tests. I came across it when running a new response selector evaluation today. Please see  this commit for a fix: https://github.com/hotzenklotz/rasa/commit/fd037beac900a4a47c00cfb071f4563d72b44fd1

In short, the default value for a response selector key was wrong. It needs to default to `None` for all "regular" intent examples or otherwise those won't be filtered out further down the pipeline and crash the sklearn reports.

cc @tmbo

PR #7427, opened 2020-12-01
7330 Add functionality to check if a model is fine-tunable

joejuzl:
**Proposed changes**:
Adds a new method `rasa.model.can_fine_tune` which uses fingerprinting to determine if a model can be fine-tuned.

### DIFF
--- a/rasa/model.py
+++ b/rasa/model.py
@@ -10,6 +10,9 @@ import typing
 from pathlib import Path
 from typing import Any, Text, Tuple, Union, Optional, List, Dict, NamedTuple
 
+from packaging import version
+
+from rasa.constants import MINIMUM_COMPATIBLE_VERSION
 import rasa.shared.utils.io
 import rasa.utils.io
 from rasa.cli.utils import create_output_path
@@ -42,6 +45,7 @@ FINGERPRINT_FILE_PATH = "fingerprint.json"
 FINGERPRINT_CONFIG_KEY = "config"
 FINGERPRINT_CONFIG_CORE_KEY = "core-config"
 FINGERPRINT_CONFIG_NLU_KEY = "nlu-config"
+FINGERPRINT_CONFIG_WITHOUT_EPOCHS_KEY = "config-without-epochs"
 FINGERPRINT_DOMAIN_WITHOUT_NLG_KEY = "domain"
 FINGERPRINT_NLG_KEY = "nlg"
 FINGERPRINT_RASA_VERSION_KEY = "version"
@@ -79,6 +83,14 @@ SECTION_NLU = Section(
     ],
 )
 SECTION_NLG = Section(name="NLG templates", relevant_keys=[FINGERPRINT_NLG_KEY])
+
+SECTION_FINE_TUNE = Section(
+    name="Fine-tune",
+    relevant_keys=[
+        FINGERPRINT_CONFIG_WITHOUT_EPOCHS_KEY,
+        FINGERPRINT_DOMAIN_WITHOUT_NLG_KEY,
+    ],
+)
 
 
 class FingerprintComparisonResult:
@@ -327,6 +339,9 @@ async def model_fingerprint(file_importer: "TrainingDataImporter") -> Fingerprin
         FINGERPRINT_CONFIG_NLU_KEY: _get_fingerprint_of_config(
             config, include_keys=CONFIG_KEYS_NLU
         ),
+        FINGERPRINT_CONFIG_WITHOUT_EPOCHS_KEY: _get_fingerprint_of_config_without_epochs(
+            config
+        ),
         FINGERPRINT_DOMAIN_WITHOUT_NLG_KEY: domain.fingerprint(),
         FINGERPRINT_NLG_KEY: rasa.shared.utils.io.deep_container_fingerprint(responses),
         FINGERPRINT_PROJECT: project_fingerprint(),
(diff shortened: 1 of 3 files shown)


rasa/model.py
```diff
@@ -80,6 +84,14 @@ class Section(NamedTuple):
 )
 SECTION_NLG = Section(name="NLG templates", relevant_keys=[FINGERPRINT_NLG_KEY])
 
+SECTION_FINE_TUNE = Section(
+    name="Fine-tune",
+    relevant_keys=[
+        FINGERPRINT_CONFIG_WITHOUT_EPOCHS_KEY,
+        FINGERPRINT_DOMAIN_WITHOUT_NLG_KEY,
```
joejuzl, line 91: Is this correct to ensure labels have not changed? Maybe there is a more looser definition e.g. `Domain.input_states`?

rasa/model.py
```diff
@@ -468,6 +499,26 @@ def should_retrain(
         return fingerprint_comparison
 
 
+def can_fine_tune(last_fingerprint: Fingerprint, new_fingerprint: Fingerprint) -> bool:
+    """Check which components of a model should be retrained.
```
dakshvar22, line 503:
```suggestion
    """Check if components of a model can be finetuned with incremental training.
```

tests/core/test_model.py
```diff
@@ -233,6 +266,44 @@ async def test_fingerprinting_changed_response_text(project: Text):
...
+async def test_fingerprinting_changing_config_epochs(project: Text):
+    importer = _project_files(project)
+
+    old_fingerprint = await model_fingerprint(importer)
+    config = await importer.get_config()
+
+    for key in ["pipeline", "policies"]:
+        for p in config[key]:
+            if "epochs" in p:
+                p["epochs"] += 10
```
dakshvar22, line 278: Q: Should we create test config files and pass them to this test rather than explicitly increasing the epochs inside the test? Don't have a hard opinion but just wondering what would be the best way to test it.

rasa/model.py
dakshvar22: I think it should be correct. I was first confused by the naming of `FINGERPRINT_DOMAIN_WITHOUT_NLG_KEY` as I thought this is the fingerprint of the domain without the `responses:` key but looks like they are also included inside `domain.fingerprint()`.
Maybe @wochinge has a better opinion on it.

dakshvar22 (changes requested):
Looking good 👍 Please add types for pytests.

rasa/model.py
wochinge: I'm positively surprised how useful the `Domain` is 😄 In my opinion it's fine!

tests/core/test_model.py
```diff
@@ -178,6 +189,28 @@ def test_nlu_fingerprint_changed(fingerprint2, changed):
...
+        (_fingerprint(domain=["other"]), True),
+        (_fingerprint(rasa_version="100"), False),
+        (_fingerprint(config_core=["other"]), False),
+        (_fingerprint(config_nlu=["other"]), False),
+        (_fingerprint(nlu=["other"]), False),
+        (_fingerprint(nlg=["other"]), False),
+        (_fingerprint(stories=["other"]), False),
+    ],
+)
+def test_fine_tune_fingerprint_changed(fingerprint2, changed):
```
wochinge, line 206: types please

tests/core/test_model.py
```diff
@@ -418,3 +489,22 @@ async def get_domain() -> Domain:
...
 
     assert actual.is_empty()
+
+
+@pytest.mark.parametrize(
+    "min_compatible_version, old_model_version, can_tune",
+    [("2.1.0", "2.1.0", True), ("2.0.0", "2.1.0", True), ("2.1.0", "2.0.0", False),],
+)
+async def test_can_fine_tune_min_version(
+    project: Text, monkeypatch, old_model_version, min_compatible_version, can_tune
```
wochinge, line 499: types please

tests/core/test_model.py
```diff
@@ -233,6 +266,44 @@ async def test_fingerprinting_changed_response_text(project: Text):
...
+
+    old_fingerprint = await model_fingerprint(importer)
+    config = await importer.get_config()
+
+    for key in ["pipeline", "policies"]:
+        for p in config[key]:
+            if "epochs" in p:
+                p["epochs"] += 10
+
+    importer.get_config = asyncio.coroutine(lambda: config)
```
wochinge, line 280: `asyncio.coroutine` is deprecated 😬

wochinge (commented):
production code looks great! Need to have another look at the tests later 👍

tests/core/test_model.py
joejuzl: Good question. The problem I see with that is someone could change the number of epochs in one of the files, which would not make this test fail, but mean it wasn't actually testing anything.

dakshvar22: Hmm, that makes sense. I think the place where having it in a file would be useful is when you are writing a test to check if the right number of epochs are passed to the ML components during loading. I guess that would be part of your other PR or maybe my PR.

wochinge: I think we could simplify the test by doing something like:

1. Define config as `Dict`
2. Write to a file
3. `TrainingDataImporter.load_from_config`
4. Test fingerprint

wochinge (approved):
Nice 👍

wochinge:
@dakshvar22 is the `changes request` still valid?

dakshvar22 (approved):
LGTM 👍

PR #7434, opened 2020-12-02
download mitie from github instead of s3

tmbo:
**Proposed changes**:
- download mitie from github instead of s3


tmbo:
@tczekajlo found another location where we are using the download from S3 😅  I've used the same command that you added to the docker file to replace S3

tczekajlo (approved):
LGTM 🚀

PR #7438, opened 2020-12-02
Implement train_step instead of overwriting fit

tabergma:
**Proposed changes**:
- related to https://github.com/RasaHQ/research/issues/97
- (Training is currently not working due to https://github.com/tensorflow/tensorflow/issues/41019 - we have an open PR for tensorflow. We probably copy over some code in the meantime.)


rasa/core/policies/ted_policy.py
```diff
@@ -1524,9 +1578,6 @@ def batch_loss(
 
     # ---PREDICTION---
 
-    def prepare_for_predict(self) -> None:
-        _, self.all_labels_embed = self._create_all_labels_embed()
```
Ghostvv, line 1528: why do did you remove this? who will create `self.all_labels_embed`

tabergma: Seems to be not needed anymore, but I am a bit confused why the model is actually training. It is on my list to take a closer look at this.

Ghostvv: but it is not for training, it is for prediction

tabergma: true 🙈 did not run prediction yet

tabergma:
```yml
include:
 - dataset: ["all"]
   config: ["all"]
```

tabergma:
/modeltest
```yml
include:
 - dataset: ["all"]
   config: ["all"]
```

rasa/utils/tensorflow/callback.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _does_model_improve(self, current_results: Dict[Text, Any]) -> bool:
+        # Initialize best_metrics_so_far with the first results
+        if not self.best_metrics_so_far:
+            keys = filter(
+                lambda k: True
+                if (k.startswith("val") and (k.endswith("_acc") or k.endswith("_f1")))
+                else False,
+                current_results.keys(),
+            )
```
Ghostvv, line 47: would it be simpler to create 2 dict inside the model? One for simply outputting metrics and one for validation purposes?

rasa/utils/tensorflow/data_generator.py
```diff
@@ -0,0 +1,116 @@
+from typing import List, Union, Text, Optional, Any, Tuple
+
+import numpy as np
+import tensorflow as tf
+
+from rasa.utils.tensorflow.constants import SEQUENCE, BALANCED
+from rasa.utils.tensorflow.model_data import RasaModelData
+
+
+class IncreasingBatchSizeDataGenerator(tf.keras.utils.Sequence):
```
Ghostvv, line 10: is it tight in to `RasaModelData`? Does it make sense to move generator related methods fully to simplify `RasaModelData` class?

rasa/utils/tensorflow/models.py
```diff
@@ -367,180 +432,14 @@ def load(
...
-            batch_size=1,
-            evaluate_every_num_epochs=0,
-            evaluate_on_num_examples=0,
-            batch_strategy=SEQUENCE,
-            silent=True,  # don't confuse users with training output
-            loading=True,  # don't use tensorboard while loading
-            eager=True,  # no need to build tf graph, eager is faster here
-        )
+        model.compile()
+        model.fit(model_data_example.as_tf_dataset(1), epochs=1, verbose=0)
```
Ghostvv, line 436: we can pass generator or dataset?

rasa/utils/tensorflow/models.py
```diff
@@ -585,88 +486,11 @@ def batch_to_model_data_format(
...
-        raise Exception(
-            "This method should neither be called nor implemented in our code."
-        )
-
-    def call(self, *args, **kwargs) -> None:
-        raise Exception(
-            "This method should neither be called nor implemented in our code."
-        )
+    def call(self, inputs, training=None, mask=None):
+        pass
```
Ghostvv, line 490: what is this method?

rasa/utils/tensorflow/models.py
tabergma: yes, you can pass all kinds of things :)

rasa/utils/tensorflow/models.py
tabergma: training fails if we don't include it, wanted to check why, but forgot to do it. will do and add a comment

rasa/utils/tensorflow/callback.py
tabergma: We actually have an issue to update our metrics. I think Thomas is working on it. Would do this change along the metric update and not inside this PR.

rasa/utils/tensorflow/models.py
```diff
@@ -117,29 +118,216 @@ def __init__(
...
-    def _set_up_tensorboard_writer(self) -> None:
-        if self.tensorboard_log_dir is not None:
-            if self.tensorboard_log_level not in TENSORBOARD_LOG_LEVELS:
-                raise ValueError(
-                    f"Provided '{TENSORBOARD_LOG_LEVEL}' ('{self.tensorboard_log_level}') "
-                    f"is invalid! Valid values are: {TENSORBOARD_LOG_LEVELS}"
-                )
-            self.tensorboard_log_on_epochs = self.tensorboard_log_level == "epoch"
+    @training.enable_multi_worker
+    def fit(
```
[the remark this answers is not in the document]
tabergma: Yes, this is temporary, added a comment somewhere below. 
So you mean something like `tf.keras.models.Model` -> `RasaKerasModel` (implements temporary custom fit) -> `RasaModel` (implements train step etc.)?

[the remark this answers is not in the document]
Ghostvv: yes, or even `TempKerasModel`, because middle one is not actually rasa

rasa/core/policies/ted_policy.py
```diff
@@ -490,13 +496,29 @@ def train(
...
+        callbacks = train_utils.create_common_callbacks(
+            self.config[EPOCHS],
+            self.config[TENSORBOARD_LOG_DIR],
+            self.config[TENSORBOARD_LOG_LEVEL],
+            self.tmp_checkpoint_dir,
+        )
+
+        self.model.compile(run_eagerly=False)
+        self.model.fit(
+            data_generator,
```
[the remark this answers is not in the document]
tabergma: Not sure I understand. We are parsing the `data_generator` and the `validation_data_generator` as the batch sizes increase.

rasa/utils/tensorflow/data_generator.py
tabergma: You mean removing the method `as_tf_dataset` and everything related to that? We could. We could also add a `FixBatchSizeGenerator` that we could use during prediction so that we don't rely on `tf.data.Dataset.from_generator` anymore.

rasa/utils/tensorflow/models.py
tabergma: > raise NotImplementedError('When subclassing the `Model` class, you should '
                              'implement a `call` method.')

The method seems to be called during `build`. We don't need it, but we nevertheless need to overwrite it.

Ghostvv: it seems it is the main way to subclass keras model, I guess it is called in `train_step` and `predict_step`, but since you overwrote them, it is not used effectively. however, will the model be still called in build?

Ghostvv:
@tabergma the auto slot filling doesn't work fo these entities?

rasa/utils/tensorflow/data_generator.py
Ghostvv: yes, because I think `RasaModel` contains so many things, it became incomprehensible.
Let's remove `tf.data.Dataset.from_generator` isn't it deprecated?

tabergma: Already on it. Becomes much cleaner 🎉 
`tf.data.Dataset.from_generator` is not deprecated but has its limitations.

rasa/utils/tensorflow/models.py
```diff
@@ -69,7 +67,223 @@
 
 
 # noinspection PyMethodOverriding
-class RasaModel(tf.keras.models.Model):
+class TmpKerasModel(tf.keras.models.Model):
```
Ghostvv, line 70: let's put it in a separate file to complete the separation. I think we can put it to the same file as data_adapter and maybe name it temp_keras.py or smth

rasa/utils/train_utils.py
```diff
@@ -186,3 +192,99 @@ def check_deprecated_options(config: Dict[Text, Any]) -> Dict[Text, Any]:
     # note: call _replace_deprecated_option() here when there are options to deprecate
 
     return config
+
+
+def create_data_generators(
```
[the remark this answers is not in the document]
tabergma: would keep it there

tabergma:
The results of the model regression tests does not match with the results when I test locally. Need to investigate that further.

tabergma:
/modeltest
```yml
include:
 - dataset: ["Private 3"]
   config: ["all"]
```

tabergma:
/modeltest
```yml
include:
 - dataset: ["Private 3"]
   config: ["BERT + DIET(bow) + ResponseSelector(bow)"]
```

tabergma:
It seems like the tensorflow changes, e.g. the temp keras modules, are ignored. The model is not trained on the number of epochs specified. Looking into it now.

tabergma:
/modeltest
```yml
include:
 - dataset: ["all"]
   config: ["all"]
```

tabergma:
/modeltest
```yml
include:
 - dataset: ["all"]
   config: ["all"]
```

rasa/utils/tensorflow/models.py
```diff
@@ -320,226 +170,64 @@ def train_on_batch(
...
-            evaluation_dataset_function,
-            self._get_tf_call_model_function(
-                evaluation_dataset_function, self._total_batch_loss, eager, "evaluation"
-            ),
-        )
+    def predict_step(
+        self, batch_in: Union[Tuple[tf.Tensor], Tuple[np.ndarray]]
+    ) -> Dict[Text, tf.Tensor]:
+        """Predict on batch."""
+        self.prepare_for_predict()
```
tabergma, line 190: `rasa shell` does not work. It has to be related to this line. Strangely `rasa shell nlu` works just fine. Unfortunately the error is not really helpful:
```
InvalidArgumentError: Unsuccessful TensorSliceReader constructor: Failed to get matching files on /var/folders/xk/kzf9l4h53k759yd46hv6qz140000gn/T/tmp9qwd4y25/nlu/component_7_DIETClassifier.tf_model: Not found: /var/folders/xk/kzf9l4h53k759yd46hv6qz140000gn/T/tmp9qwd4y25/nlu; No such file or directory [Op:RestoreV2]
```

Ghostvv: does returning to old style `if is not None` work?
my hunch is that `self.prepare_for_predict()` should be part of `build` but only when it is compiled for prediction

Ghostvv: or maybe part of `compile` not sure what is the difference to `build`

tabergma: > does returning to old style if is not None work?

No 😞 

As far as I understand `compile` is just needed before training. But need to take a closer look at all those methods again on Monday.

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -243,7 +244,7 @@ def required_components(cls) -> List[Type[Component]]:
         TENSORBOARD_LOG_DIR: None,
         # Define when training metrics for tensorboard should be logged.
         # Either after every epoch or for every training step.
-        # Valid values: 'epoch' and 'minibatch'
+        # Valid values: 'epoch' and 'batch'
         TENSORBOARD_LOG_LEVEL: "epoch",
```
dakshvar22, line 248: Same comment for changelog and migration guide

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -832,7 +854,7 @@ def _predict_label(
         if predict_out is None:
             return label, label_ranking
 
-        message_sim = predict_out["i_scores"].numpy()
+        message_sim = predict_out["i_scores"]
```
dakshvar22, line 857: For my understanding: What changed because of which we do not need to convert from `tf.Tensor` to numpy equivalent tensors?

rasa/utils/tensorflow/data_generator.py
```diff
@@ -0,0 +1,475 @@
...
+        shuffle: bool = True,
+    ):
+        """Initializes the increasing batch size data generator.
+
+        Args:
+            model_data: The model data to use.
+            batch_size: The batch size.
+            epochs: The total number of epochs.
+            batch_strategy: The batch strategy.
+            shuffle: If 'Ture', data will be shuffled.
```
dakshvar22, line 354:
```suggestion
            shuffle: If 'True', data will be shuffled.
```

rasa/utils/tensorflow/data_generator.py
```diff
@@ -0,0 +1,475 @@
...
+                    # filter all the "fake" inputs, we know the input is "fake",
+                    # when sequence dimension is `0`
+                    list(filter(lambda x: x.shape[0] > 0, array_of_features))
+                    for array_of_features in array_of_array_of_features
+                ],
+            )
+        )
+
+
+class IncreasingBatchSizeDataGenerator(RasaDataGenerator):
```
dakshvar22, line 336: Q: Do we need an extra class for `IncreasingBatchSizeDataGenerator`? Can't we do something like - if passed `batch_size` is a `list` rather than `int`, we call `linearly_increasing_batch_size` to get the next batch size?

rasa/utils/tensorflow/temp_keras_modules.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+# noinspection PyMethodOverriding
+class TmpKerasModel(tf.keras.models.Model):
+    """Temporary solution. Keras model that uses a custom data adapter inside fit."""
+
+    # TODO
+    #  we don't need this anymore once
+    #  https://github.com/tensorflow/tensorflow/pull/45338
+    #  is merged and released
```
dakshvar22, line 29: Can we also add an attribution comment something like - `This code is adapted from <link>`

tests/core/test_policies.py
```diff
@@ -407,35 +408,43 @@ async def test_gen_batch(self, trained_policy: TEDPolicy, default_domain: Domain
             training_data, label_ids, entity_tags, all_labels
         )
         batch_size = 2
-        print(len(next(model_data._gen_batch(batch_size=batch_size))))
-        print()
```
dakshvar22, line 411: cc @Ghostvv So that the `e2e` PR is not merged with these print statements.

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -402,7 +402,8 @@ async def test_train_tensorboard_logging(component_builder, tmpdir):
     assert tensorboard_log_dir.exists()
 
     all_files = list(tensorboard_log_dir.rglob("*.*"))
-    assert len(all_files) == 3
+    print(all_files)
```
dakshvar22, line 405:
```suggestion
```

rasa/core/policies/ted_policy.py
```diff
@@ -255,7 +258,7 @@ class TEDPolicy(Policy):
         TENSORBOARD_LOG_DIR: None,
         # Define when training metrics for tensorboard should be logged.
         # Either after every epoch or for every training step.
-        # Valid values: 'epoch' and 'minibatch'
+        # Valid values: 'epoch' and 'batch'
```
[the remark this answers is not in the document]
dakshvar22: Also, should we add/modify a test to use `batch` instead of `minibatch`? However, not sure what would the test do since it's hard to actually check the logs.

dakshvar22 (changes requested):
Wow! Looks like a lot of clean up 🚀  Left some suggestions.

rasa/utils/tensorflow/models.py
tabergma: Issue seems to be resolved. Just merged in `e2e`. Not sure exactly why it is now working, but maybe it was related to the way how the interpreter was load inside core.

tabergma: (Was on the wrong branch before - still not working.)

After further investigation I think the error isn't related to `self.prepare_for_predict()`. `rasa shell nlu` works fine. Also, when you parse a method as soon as the `Interpreter` is created inside `rasa shell` everything works just fine. However, as soon as the server is up the path to the model file seems to be not valid anymore. We need to check why the model file is needed once the model is already loaded.

tabergma: fixed it, we needed to keep the it on one example during loading to actually build the model and load the weights

rasa/core/policies/ted_policy.py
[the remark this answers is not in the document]
tabergma: We automatically update `minibatch` to `batch` and print a warning. But will add it to the changelog and migration guide. Will take a look at our tests.

rasa/core/policies/ted_policy.py
```diff
@@ -497,13 +504,29 @@ def train(
...
+        )
+
+        self.model.compile()
+        self.model.fit(
+            data_generator,
+            epochs=self.config[EPOCHS],
+            validation_data=validation_data_generator,
+            validation_freq=self.config[EVAL_NUM_EPOCHS],
+            callbacks=callbacks,
+            verbose=False,
```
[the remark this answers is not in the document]
tabergma: No, would keep it silent. The default keras logger cannot handle our dynamic batch size very well. The output will assume you have the same step size every epoch, which is not true for us. So might actually be more confusing for users.
We could add some point make the callbacks itself configurable, e.g. users can list a callback in their config file which should be used during training, but that is something for later.

rasa/utils/tensorflow/data_generator.py
tabergma: Yeah, I guess we could do that, like the idea actually 👍

rasa/nlu/classifiers/diet_classifier.py
tabergma: The outputs are no `tf.Tensor` anymore. Not sure where exactly the outputs are converted (tried to find it inside `fit`, but could not find it), but we already have numpy arrays here.

rasa/utils/tensorflow/data_generator.py
tabergma: Should we keep the name `IncreasingBatchSizeDataGenerator` or would you go with something else?

tests/core/test_policies.py
Ghostvv: already fixed in `e2e`

rasa/utils/tensorflow/data_generator.py
Ghostvv: I'd still use increasing, sometimes it can be increasing by `0` 😂

dakshvar22: I would suggest just having one class named `BatchDataGenerator`. Increasing batch is a property of that class depending on the value of `batch_size`. What do you think?

Ghostvv: let's then add `Rasa` so that we know that it uses custom rasa data structure and that it is not general class. Similar separation we have with our models and layers, where layers without `Rasa...` are general reusable layers by any model

tabergma: Let's keep the `RasaDataGenerator` with the more generic methods as I will create a second data generator that is able to load data from files. 
I merged the `FixedBatchSizeDataGenerator` and the `IncreasingBatchSizeDataGenerator` into one class. It is currently still called `IncreasingBatchSizeDataGenerator`. 
`BatchDataGenerator` seems to be too generic I think. In the end all data generators create some kind of batches.

tabergma: updated the name to `RasaBatchDataGenerator`

tabergma:
TODO: As soon as end-to-end is merged and the release is out, update the migration guide & update the minimal compatible version.

rasa/utils/tensorflow/models.py
```diff
@@ -223,8 +227,14 @@ def load(cls, model_file_name: Text, *args, **kwargs) -> "RasaModel":
...
         # create empty model
         model = cls(*args, **kwargs)
+        # need to train on 1 example to build weights of the correct size
+        model.compile(run_eagerly=True)
+        data_generator = RasaBatchDataGenerator(model_data_example, batch_size=1)
+        model.fit(data_generator, verbose=False)
         # load trained weights
         model.load_weights(model_file_name)
+        # prepare the model for prediction
+        model.prepare_for_predict()
```
dakshvar22, line 237: The call to `prepare_for_predict` is already happening inside `build_for_predict` which is being called by the `_load_model()` of `DIETClassifier` and `TEDPolicy`. I think you should skip making the call here.

dakshvar22: Ahh no, i now see we have remove `build_for_predict` in this PR.

dakshvar22: Okay just so you know, we should merge `e2e`/ `master` carefully into this since `model.load()` needs to take care of loading the model in finetune mode or in predict model.

dakshvar22:
@tabergma Should we run the model regression tests again to see if the performance drop has improved or do you plan to do it after some more modifications?

tabergma:
I want to wait until `e2e` is merged into master and this branch is updated. As soon as that is done I run the tests again.

rasa/nlu/classifiers/diet_classifier.py
dakshvar22: Interesting 🤔

dakshvar22 (approved):
LGTM. Thanks for addressing all my comments ✨

rasa/utils/tensorflow/data_generator.py
```diff
@@ -0,0 +1,428 @@
...
+
+        self.on_epoch_end()
+
+    def __len__(self) -> int:
+        """Number of batches in the Sequence.
+
+        Returns:
+            The number of batches in the Sequence.
+        """
+        pass
```
Ghostvv, line 46: can this generator be used on itself? should it and other `pass` be `raise NotImplementedError`?

rasa/utils/tensorflow/models.py
```diff
@@ -58,31 +45,23 @@
...
 logger = logging.getLogger(__name__)
 
 
-TENSORBOARD_LOG_LEVELS = ["epoch", "minibatch"]
-
-
 # noinspection PyMethodOverriding
-class RasaModel(tf.keras.models.Model):
+class RasaModel(TmpKerasModel):
     """Completely override all public methods of keras Model.
```
Ghostvv, line 59: Could you please rephrase docstring

rasa/utils/tensorflow/models.py
```diff
@@ -58,31 +45,23 @@
...
 
     Cannot be used as tf.keras.Model
```
Ghostvv, line 61: this is still true, right?

rasa/utils/tensorflow/models.py
```diff
@@ -180,114 +117,12 @@ def batch_predict(
...
-        self._training = None  # training phase should be defined when building a graph
-        if not disable:
-            logger.info("Finished training.")
-
-    def train_on_batch(
+    def train_step(
         self, batch_in: Union[Tuple[tf.Tensor], Tuple[np.ndarray]]
-    ) -> None:
+    ) -> Dict[Text, float]:
         """Train on batch."""
```
Ghostvv, line 123: could you please update docstring with Args

rasa/utils/tensorflow/models.py
```diff
@@ -320,47 +155,58 @@ def train_on_batch(
...
-            logger.debug("There is no tensorflow prediction graph.")
-            self.build_for_predict(predict_data)
+        return self._get_metric_results()
 
-        # Prepare a single batch of the size of the input
-        batch_in = predict_data.prepare_batch()
+    def test_step(
+        self, batch_in: Union[Tuple[tf.Tensor], Tuple[np.ndarray]]
+    ) -> Dict[Text, float]:
+        """Train on batch."""
```
Ghostvv, line 165: could you please update docstring

Ghostvv: this is for validation?

rasa/utils/tensorflow/models.py
```diff
@@ -320,47 +155,58 @@ def train_on_batch(
...
-        checkpoint_path = Path(checkpoint_directory)
+        return self._get_metric_results()
 
-        # Copy all tf2 model files from the temp location to the final destination
-        for f in checkpoint_path.glob(f"{checkpoint_file}*"):
-            shutil.move(str(f.absolute()), model_file_name + f.suffix)
+    def predict_step(
+        self, batch_in: Union[Tuple[tf.Tensor], Tuple[np.ndarray]]
+    ) -> Dict[Text, tf.Tensor]:
+        """Predict on batch."""
```
Ghostvv, line 180: could you please update docstring with Args

rasa/utils/tensorflow/models.py
```diff
@@ -320,47 +155,58 @@ def train_on_batch(
...
+            # after training, we need to prepare the model for prediction once
+            self.prepare_for_predict()
+            self.prepared_for_prediction = True
 
-        # Remove the old file
-        checkpoint_path.joinpath("checkpoint").unlink()
+        return self.batch_predict(batch_in)
+
+    def _get_metric_results(self, prefix: Optional[Text] = None) -> Dict[Text, float]:
+        """Get the metrics results."""
```
Ghostvv, line 192: could you please update docstring with args, or remove it, since it is private method

rasa/utils/tensorflow/models.py
```diff
@@ -388,180 +234,22 @@ def load(
...
+        model.compile(run_eagerly=False if finetune_mode else True)
+        data_generator = RasaBatchDataGenerator(model_data_example, batch_size=1)
+        model.fit(data_generator, verbose=False)
         # load trained weights
         model.load_weights(model_file_name)
 
+        if not finetune_mode:
+            # prepare the model for prediction
+            model.prepare_for_predict()
+            model.prepared_for_prediction = True
```
Ghostvv, line 248: I think the model also have to be run on prediction example to build the prediction graph, or did it somehow changed?

rasa/utils/tensorflow/models.py
```diff
@@ -595,97 +289,48 @@ def batch_to_model_data_format(
                         idx += 3
                     else:
                         if isinstance(batch[idx], tf.Tensor):
+                            if number_of_dimensions > 1 and (
+                                batch[idx].shape is None or batch[idx].shape[-1] is None
+                            ):
+                                shape = [None] * (number_of_dimensions - 1) + [
+                                    feature_dimension
+                                ]
+                                batch[idx].set_shape(shape)
```
Ghostvv, line 298: do you thin it makes sense to move it to a helper method?

Ghostvv (approved):
looks awesome, left mostly style comments and one about building prediction graph

rasa/utils/tensorflow/models.py
tabergma: yes, will update the comment

rasa/utils/tensorflow/models.py
tabergma: yes

Ghostvv:
@tabergma does it make sense to rebase to master? so that e2e commits are not the part of this branch?

tabergma:
I would squash merge this PR. Rebase will not work.

tabergma:
/modeltest

```
include:
- dataset: ["all"]
    config: ["all"]
```

rasa/utils/tensorflow/models.py
tabergma: It's not needed anymore. The prediction graph don't uses any new layers. All layers are already build on the `fit` call above. However, it might be a bit faster during inference if we predict already one example during loading. Will check that.

tabergma: Ok, predicting the first example using DIET takes 0.9635 sec vs 0.0632 sec. So I guess we should predict one example already during loading to speed things up.

Ghostvv: yeah, I think it's not about building actual layers, but rather converting python code into c++ executable tf graph

rasa/utils/train_utils.py
```diff
@@ -335,3 +342,97 @@ def override_defaults(
...
+
+    data_generator = RasaBatchDataGenerator(
+        model_data,
+        batch_size=batch_sizes,
+        epochs=epochs,
+        batch_strategy=batch_strategy,
+        shuffle=True,
+    )
+
+    return data_generator, validation_data_generator
```
alwx, line 389: Sort of related to the comment below (so it's better to read the one below first).

In this case it might also be a good idea to create two separate factory functions like `RasaBatchDataGenerator.new_test_generator` and `RasaBatchDataGenerator.new_generator` instead — just for the sake of having those functions in the right place (and not in `utils`) and making them easier to read + easier to test.

rasa/utils/tensorflow/temp_keras_modules.py
```diff
@@ -0,0 +1,265 @@
...
+from tensorflow.python.keras.engine import training_utils
+from tensorflow.python.keras.engine import data_adapter
+from tensorflow.python.keras.utils import version_utils
+from tensorflow.python.eager import context
+from tensorflow.python.keras.engine.data_adapter import DataHandler
+
+
+# noinspection PyMethodOverriding
+class TmpKerasModel(tf.keras.models.Model):
+    """Temporary solution. Keras model that uses a custom data adapter inside fit."""
```
alwx, line 24: Variable names like `x` and `y` in this class are weird and say nothing at all, it's constructor is too long, etc etc but I guess it doesn't matter much since this code is temporary and will be removed later anyway?

rasa/utils/tensorflow/models.py
```diff
@@ -574,118 +301,95 @@ def batch_to_model_data_format(
...
-            "This method should neither be called nor implemented in our code."
-        )
+        Returns:
+            A tensor if there is a single output, or
+            a list of tensors if there are more than one outputs.
+        """
+        # This method needs to be implemented, otherwise the super class is raising a
+        # NotImplementedError('When subclassing the `Model` class, you should
+        #   implement a `call` method.')
+        pass
```
alwx, line 392: What do you think about explicitly raising `NotImplementedError` instead of this comment? (basically the same way as we do it for other methods in `Model` class and its subclasses)

rasa/utils/tensorflow/models.py
```diff
@@ -320,53 +168,80 @@ def train_on_batch(
...
 
-        # Generate the tf2 checkpoint file, copy+replace to ensure consistency
-        destination_path, destination_file = os.path.split(model_file_name)
-        with open(os.path.join(checkpoint_directory, "checkpoint")) as in_file, open(
-            os.path.join(destination_path, "checkpoint"), "w"
-        ) as out_file:
-            for line in in_file:
-                out_file.write(line.replace(checkpoint_file, destination_file))
+    def _get_metric_results(self, prefix: Optional[Text] = None) -> Dict[Text, float]:
+        prefix = prefix or ""
```
alwx, line 221: Seems like the default can just be `""` and not `None` in this case.

rasa/utils/tensorflow/models.py
```diff
@@ -95,55 +80,13 @@ def __init__(
...
-
-            train_log_dir = (
-                f"{self.tensorboard_log_dir}/{class_name}/{current_time}/train"
-            )
-            test_log_dir = (
-                f"{self.tensorboard_log_dir}/{class_name}/{current_time}/test"
-            )
+        random.seed(self.random_seed)
+        tf.random.set_seed(self.random_seed)
+        np.random.seed(self.random_seed)
```
alwx, line 87: I'd say it's better to move this three function calls out of constructor to a separate function.

rasa/utils/tensorflow/data_generator.py
```diff
@@ -0,0 +1,428 @@
...
+
+import rasa.shared.utils.io
+from rasa.utils.tensorflow.constants import SEQUENCE, BALANCED
+from rasa.utils.tensorflow.model_data import RasaModelData, Data, FeatureArray
+
+logger = logging.getLogger(__name__)
+
+
+class RasaDataGenerator(tf.keras.utils.Sequence):
+    """Abstract data generator."""
```
alwx, line 16: Nice to see this one being extracted from the `Model`!

rasa/utils/tensorflow/callback.py
```diff
@@ -0,0 +1,101 @@
...
+                self.checkpoint_file, overwrite=True, save_format="tf"
+            )
+
+    def _does_model_improve(self, current_results: Dict[Text, Any]) -> bool:
+        # Initialize best_metrics_so_far with the first results
+        if not self.best_metrics_so_far:
+            keys = filter(
+                lambda k: True
+                if (k.startswith("val") and (k.endswith("_acc") or k.endswith("_f1")))
+                else False,
```
alwx, line 80: Lambda here can be simplified to just `lambda k: k.startswith("val") and (k.endswith("_acc") or k.endswith("_f1")`

alwx (changes requested):

rasa/utils/tensorflow/temp_keras_modules.py
Ghostvv: this is a copy paste from keras, we're not changing variables in there. Besides `x` and `y` are perfect names here

rasa/utils/tensorflow/models.py
Ghostvv: why would we raise NotImplementedError? it is specifically overwritten not to raise this error

rasa/utils/train_utils.py
```diff
@@ -335,3 +342,97 @@ def override_defaults(
...
+
+
+def create_common_callbacks(
+    epochs: int,
+    tensorboard_log_dir: Optional[Text] = None,
+    tensorboard_log_level: Optional[Text] = None,
+    checkpoint_dir: Optional[Path] = None,
+) -> List[tf.keras.callbacks.Callback]:
```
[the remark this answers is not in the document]
tabergma: As this method will be called for all our custom models, currently three, I just wanted to extract the code in a central place so that we don't repeat the logic three times.

rasa/utils/train_utils.py
tabergma: Would not create factory methods for this. `RasaBatchDataGenerator.new_test_generator` and `RasaBatchDataGenerator.new_generator` would be the same method, the same as the constructor, as I would not put the logic 
```
    if eval_num_examples > 0:
        model_data, evaluation_model_data = model_data.split(
            eval_num_examples, random_seed,
        )
```
inside the `RasaBatchDataGenerator`. This logic rather belongs to the model itself. I just extracted it to a separate function as we use this logic across all our custom models.

rasa/utils/tensorflow/temp_keras_modules.py
tabergma: Yeah, would not modify this file as it is not "our" code. As mentioned in the docstring this is just a temporary solution as long as our PR on tensorflow side is not merged. The code is directly taken from tensorflow, just two lines are adapted.

rasa/utils/tensorflow/models.py
tabergma: We cannot raise a `NotImplementedError` as training would fail in that case. We explicitly want to do nothing here. That is the reason we overwrite this method.

rasa/utils/train_utils.py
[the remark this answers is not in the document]
tabergma: So you are basically saying to just remove the function here and copy over the logic to the model itself when it creates the callbacks?

rasa/utils/train_utils.py
tabergma: Should I move the code to the model itself again?

I think once we refactor the whole structure around ML components, this can be improved further. But that is something for later.

rasa/utils/tensorflow/temp_keras_modules.py
alwx: Ok, no problem :)

rasa/utils/tensorflow/models.py
alwx: Oh okay, that makes sense then 👍

rasa/utils/train_utils.py
alwx: I think it can certainly become a factory method inside `RasaBatchDataGenerator` but if you say that the refactoring is going to continue further then it might be fine to skip it for now as well.

rasa/utils/train_utils.py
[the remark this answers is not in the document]
alwx: So far I see that it's called only once in `diet_classifier.py` but maybe I missed something. 
In this case, it might even be better to place this logic somewhere in the classifier itself (`DIETClassifier` or its parents if it makes sense) — sorry, I am sort of trying to avoid files like `*_utils.py` as much as I can especially when the logic is appicable to one particular component.

Feel free to skip it if my assumption is incorrect and if you see this particular code being in use in many different places, not just classifiers.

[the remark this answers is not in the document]
Ghostvv: it is also called in `TEDPolicy` so we cannot make it part of the diet classifier

rasa/utils/train_utils.py
Ghostvv: same argument here as with the other function. it is used in both TED and DIET. creating validation dataset is external logic towards `RasaBatchDataGenerator`

alwx (approved):

rasa/utils/tensorflow/data_generator.py
```diff
@@ -0,0 +1,428 @@
...
+
+        # return input and target data, as our target data is inside the input
+        # data return None for the target data
+        return self.prepare_batch(self.model_data.data, start, end), None
+
+    def on_epoch_end(self) -> None:
+        """Update the data after every epoch."""
+        self.current_epoch += 1
+        if self.increase_batch_size:
+            self.current_batch_size = self.linearly_increasing_batch_size(
```
dakshvar22, line 399: Probably doesn't matter in terms of performance but to replicate the results of master exactly, the call to `self.linearly_increasing_batch_size` should be inside `on_epoch_begin` instead of `on_epoch_end`.

Ghostvv: yeah, but they don't seem to have `on_epoch_begin` in keras

Ghostvv: looking at it, it seems to be equivalent to calling on the first epoch, except the call in init

Ghostvv: slightly refactored, to make it a bit more clear

joejuzl:
I profiled the memory usage while running `tests/core/test_policies.py` on master, and on this branch.

### master:
policy_memory_profile_master2

### train_step (commit 4ea8d8 - current head )
policy_memory_profile_train_step2

### train_step (commit 07d63c - before keras clear_session)
policy_memory_profile_train_step_before_keras_clear


## Observations:
- Takes much longer to run on `train_step`.
- Way higher memory usage on `train_step`, even at the same relative point in time. 
- The Keras clear_session seemed to help somewhat.
- All show a continuous increase in memory usage, suggesting a memory leak.

Ghostvv:
@joejuzl could you please give it a final review?

.github/workflows/continous-integration.yml
```diff
@@ -191,7 +191,7 @@ jobs:
     needs: [changes]
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
```
joejuzl, line 194: can we add windows back now?

rasa/core/policies/ted_policy.py
```diff
@@ -644,11 +642,7 @@ def predict_action_probabilities(
             tracker, domain, interpreter
         )
         model_data = self._create_model_data(tracker_state_features)
-
-        data_generator = RasaBatchDataGenerator(
-            model_data, batch_size=len(tracker_state_features)
-        )
-        output = self.model.predict(data_generator)
```
joejuzl: is `self.model.predict` ever still used?

Ghostvv: it can be used, we don't use it, but you would in your custom code if you want to connect some callbacks

.github/workflows/continous-integration.yml
Ghostvv: yes, I think this is @tczekajlo addition?

Ghostvv: done

joejuzl (approved):

joejuzl:
#8085 should help to reduce the memory usage of the policy tests and stop the CI tests failing.

PR #7443, opened 2020-12-03
Add entities to UserUttered event if they are predicted via a policy

tabergma:
**Proposed changes**:
- `TEDPolicy` add a `DefinePrevUserUtteredEntities` to the policy prediction in case of entities 
- Update last `UserUttered` event if entities are predicted

related to #6670

### DIFF
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -34,7 +34,6 @@ from rasa.shared.core.events import (
     ActionExecutionRejected,
     ActionExecuted,
     DefinePrevUserUtteredFeaturization,
-    DefinePrevUserUtteredEntities,
 )
 from rasa.core.exceptions import UnsupportedDialogueModelError
 from rasa.core.featurizers.tracker_featurizers import MaxHistoryTrackerFeaturizer
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -743,14 +743,6 @@ class SimplePolicyEnsemble(PolicyEnsemble):
                 winning_prediction.events.append(
                     DefinePrevUserUtteredFeaturization(True)
                 )
-                if winning_prediction.entities is not None:
-                    logger.debug(
-                        f"Added `DefinePrevUserUtteredEntities("
-                        f"{winning_prediction.entities})` event."
-                    )
-                    winning_prediction.events.append(
-                        DefinePrevUserUtteredEntities(winning_prediction.entities)
-                    )
             else:
                 logger.debug("Made prediction using user intent.")
                 logger.debug("Added `DefinePrevUserUtteredFeaturization(False)` event.")
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -34,6 +34,7 @@ from rasa.shared.core.events import (
     ActionExecutionRejected,
     ActionExecuted,
     DefinePrevUserUtteredFeaturization,
+    DefinePrevUserUtteredEntities,
 )
 from rasa.core.exceptions import UnsupportedDialogueModelError
 from rasa.core.featurizers.tracker_featurizers import MaxHistoryTrackerFeaturizer
@@ -562,6 +563,7 @@ class SimplePolicyEnsemble(PolicyEnsemble):
             best_prediction.policy_priority,
             policy_events,
             is_end_to_end_prediction=best_prediction.is_end_to_end_prediction,
+            entities=best_prediction.entities,
         )
 
     def _best_policy_prediction(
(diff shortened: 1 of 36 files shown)


tabergma:
@Ghostvv Just wanted to check if I am on the right path. Can you take a quick look, no full review needed yet. Do we already have an e2e test that I can use to test the functionality?

rasa/core/policies/ensemble.py
```diff
@@ -562,6 +563,7 @@ def _pick_best_policy(
             best_prediction.policy_priority,
             policy_events,
             is_end_to_end_prediction=best_prediction.is_end_to_end_prediction,
+            entities=best_prediction.entities,
```
Ghostvv, line 566: there is no need to add entities to `PolicyPrediction`, you can add `DefinePrevUserUtteredEntities` directly in TED as optional event. Similar like it is done in `RulePolicy` with `LoopInterrupted` except that `RulePolicy` adds it as `events` (applicable regardless of which policy wins), while TED should add it as `optional_events` (applicable only if TED wins)

tabergma: ah, thanks! will update the code

tabergma:
The tests seem to already fail on the `e2e` branch.

rasa/shared/core/events.py
```diff
@@ -559,18 +568,85 @@ def _from_parameters(cls, parameters) -> "DefinePrevUserUtteredFeaturization":
...
+            entities: the entities of a previous user uttered event
+            timestamp: the timestamp
+            metadata: some optional metadata
+        """
+        self.entities = entities
+        super().__init__(timestamp, metadata)
+
+    def __str__(self) -> Text:
+        """Returns the string representation of the event."""
+        return f"DefinePrevUserUtteredEntities({self.entities})"
```
Ghostvv, line 605: that will contain a dict with all the confidences, etc, should we filter only entity names here?

rasa/shared/core/trackers.py
```diff
@@ -475,10 +480,21 @@ def _define_user_featurization(future_events: List[Event]) -> bool:
             elif isinstance(future_event, DefinePrevUserUtteredFeaturization):
                 return future_event.use_text_for_featurization
 
+    @staticmethod
+    def _define_user_entities(
+        future_events: List[Event],
+    ) -> Optional[List[Dict[Text, Any]]]:
+        for future_event in future_events:
+            if isinstance(future_event, DefinePrevUserUtteredEntities):
+                return future_event.entities
```
Ghostvv, line 489: we have to search only until next `AtionExecuted` because the search should happen only within one dialogue turn. See `_define_user_featurization` for reference

rasa/shared/core/trackers.py
```diff
@@ -459,6 +460,10 @@ def applied_events(self) -> List[Event]:
                 )
                 if event.use_text_for_featurization is None:
                     event.use_text_for_featurization = use_text_for_featurization
+                # update event's entities based on the future event
+                entities = self._define_user_entities(events_as_list[i + 1 :])
+                if event.entities is None:
```
Ghostvv, line 465: not sure about that one. Because most probably `event.entities` is already set with smth from nlu pipeline. I'd say we have to update `entities` dict treating `TEDPolicy` as another extractor. I'm not sure though how the historical `UserUttered` events are persisted. We need to avoid updating entities every time

Maybe we should do it `if event.use_text_for_featurization is None:` but not sure if it is a correct condition, it needs to be tested

tabergma: I updated the check. Currently checking if the entity is not yet inside the entities dict. If not, it is added.

tabergma:
@Ghostvv Updated the code and tested it locally on e2e bot. Everything seems to work as expected.

rasa/core/policies/ted_policy.py
```diff
@@ -579,10 +583,58 @@ def predict_action_probabilities(
...
+        if ENTITY_ATTRIBUTE_TYPE not in predicted_tags:
+            # no entities detected
+            return None
+
+        # find last user uttered event as the predicted entities belong to
+        # that utterance
+        if tracker.latest_action_name == ACTION_LISTEN_NAME:
+            last_user_utterance = tracker.latest_message
+        else:
+            return None
```
Ghostvv, line 622: I think
```
if tracker.latest_action_name == ACTION_LISTEN_NAME:
    return
```
should be the first check to save calculation and extraction of entity labels etc
actually, in this case TED itself should return smth like `None`?

rasa/shared/core/trackers.py
```diff
@@ -459,6 +460,12 @@ def applied_events(self) -> List[Event]:
                 )
                 if event.use_text_for_featurization is None:
                     event.use_text_for_featurization = use_text_for_featurization
+                # update event's entities based on the future event
+                entities = self._define_user_entities(events_as_list[i + 1 :])
+                if entities is not None:
+                    for entity in entities:
+                        if entity not in event.entities:
```
Ghostvv, line 467: entity is a full dict here?

Ghostvv (approved):
left a couple of comments

Ghostvv:
@tabergma the auto slot filling doesn't work fo these entities?

rasa/shared/core/trackers.py
tabergma: yes

rasa/core/policies/ted_policy.py
tabergma: I guess you mean `!=` instead of `==`.

tabergma:
Yes, slot filling only works for entities coming from user messages aka NLU directly. But I guess that is something for a different PR. Also testing entities in TED is not yet implemented, e.g. via `rasa test`.

rasa/core/policies/ted_policy.py
Ghostvv: yes

Issue #7395, opened 2020-11-27
Store and load DialogueStateTrackers


tabergma:
related to #6836 

When we create `DialogueStateTrackers` from stories, we need to resolve checkpoints and augment stories. This can potentially lead to out of memory issues.
We should make sure to write `DialogueStateTrackers` to disk as soon as they are complete/finished. We would also need a method that allows us to load a `DialogueStateTracker` from disk again given a file name.

This should only happen if we train in multiple chunks.

PR #7447, opened 2020-12-03
Implement tracker persist and load

JEM-Mosig:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/7395

### DIFF
--- a/rasa/shared/core/trackers.py
+++ b/rasa/shared/core/trackers.py
@@ -64,6 +64,7 @@ import rasa.utils.common
 import pathlib
 import rasa.shared
 import jsonpickle
+import rasa.exceptions
 
 if TYPE_CHECKING:
     from rasa.shared.core.training_data.structures import Story
@@ -833,8 +834,8 @@ class DialogueStateTracker:
 
     @classmethod
     def load_tracker(
-        cls: Type[rasa.utils.common.T], file_path: Union[Text, pathlib.Path]
-    ) -> rasa.utils.common.T:
+        cls, file_path: Union[Text, pathlib.Path]
+    ) -> "DialogueStateTracker":
         """Loads a tracker from a file.
 
         Args:
@@ -845,7 +846,9 @@ class DialogueStateTracker:
         """
         tracker_file = pathlib.Path(file_path)
         if not tracker_file.is_file():
-            raise FileNotFoundError(f"Could not load tracker from '{file_path}'.")
+            raise rasa.exceptions.TrackerFileNotFound(
+                f"Could not load tracker from '{file_path}'."
+            )
 
         return jsonpickle.decode(rasa.shared.utils.io.read_file(tracker_file))
 
--- a/rasa/shared/core/trackers.py
+++ b/rasa/shared/core/trackers.py
@@ -60,6 +60,10 @@ from rasa.shared.core.events import (
 )
 from rasa.shared.core.domain import Domain, State
 from rasa.shared.core.slots import Slot
+import rasa.utils.common
+import pathlib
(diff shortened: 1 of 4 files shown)


rasa/shared/core/trackers.py
```diff
@@ -809,6 +813,42 @@ def latest_action_name(self) -> Optional[Text]:
...
+        rasa.shared.utils.io.create_directory_for_file(tracker_file)
+
+        # noinspection PyTypeChecker
+        rasa.shared.utils.io.write_text_file(str(jsonpickle.encode(self)), tracker_file)
+
+        return str(tracker_file)
+
+    @classmethod
+    def load_tracker(
+        cls: Type[rasa.utils.common.T], file_path: Union[Text, pathlib.Path]
```
tabergma, line 836:
```suggestion
        cls, file_path: Union[Text, pathlib.Path]
```
I don't think you need to specify the type for `cls`.

rasa/shared/core/trackers.py
```diff
@@ -809,6 +813,42 @@ def latest_action_name(self) -> Optional[Text]:
...
+
+        Args:
+            file_path: Full path and name of the file.
+
+        Returns:
+            The loaded DialogueStateTracker.
+        """
+        tracker_file = pathlib.Path(file_path)
+        if not tracker_file.is_file():
+            raise FileNotFoundError(f"Could not load tracker from '{file_path}'.")
```
tabergma, line 848: I think we should use a `RasaException` here. You might want to create a new one in https://github.com/RasaHQ/rasa/blob/master/rasa/exceptions.py

tabergma (approved):
💯 Looks good. Please take a look at the two comments I left.

Please make sure to squash and merge the PR once it is done. You cannot use the label `ready-to-be-merged` for that. You need to do it manually. Thanks.

rasa/shared/core/trackers.py
```diff
@@ -809,6 +814,44 @@ def latest_action_name(self) -> Optional[Text]:
...
+        Returns:
+            Full path of the stored tracker file.
+        """
+        tracker_file = pathlib.Path(path) / filename
+        rasa.shared.utils.io.create_directory_for_file(tracker_file)
+
+        # noinspection PyTypeChecker
+        rasa.shared.utils.io.write_text_file(str(jsonpickle.encode(self)), tracker_file)
+
+        return str(tracker_file)
```
degiz, line 833: Do we have a plan about what to do with the result value? I mean, does it have to be `string` (if we just log it). or should we rather return `Path`, in case we want to i.g. concat it later.

degiz: It's a question, not asking for a change 😄

tests/shared/core/test_trackers.py
```diff
@@ -1242,3 +1242,30 @@ def test_trackers_for_conversation_sessions(
     subtrackers = trackers_module.get_trackers_for_conversation_sessions(tracker)
 
     assert len(subtrackers) == n_subtrackers
+
+
+def test_tracker_persist_and_load_empty_tracker(tmp_path: Path, default_domain: Domain):
```
degiz, line 1247: Should we have one more test that will check that the exception is raised?

degiz (approved):
@JEM-Mosig  I'm approving to unblock you, please check the comments! 👍

rasa/shared/core/trackers.py
JEM-Mosig: It should be used exactly as in the test. I suppose a Path object would be more general, though, so I'll change it to that.

PR #7451, opened 2020-12-04
core featurizers should use regex interpreter if core was trained separately

Ghostvv:
**Proposed changes**:
- apply changes from https://github.com/RasaHQ/rasa/pull/7316 to `2.1.x`


samsucik (approved):
Approving as it's the same changes that were approved previously for `2.0.x`.

Issue #7329, opened 2020-11-20
Load models in finetune mode based on command line parameters.


wochinge:
[Load models for incremental training.](https://github.com/RasaHQ/rasa/issues/6971)

Depends on https://github.com/RasaHQ/rasa/issues/7328.

- [ ] load NLU pipeline in fine-tune mode as shown here
- [ ] load Core policies in fine-tune mode as show here
- [ ] pass in new number of epochs

dakshvar22:
@wochinge @joejuzl Do you folks know how the `load` method for each ML component will be called during finetuning? For example, would there be an extra parameter passed to `load` function of the components to load them in "finetune" mode and also pass other configurations parameters like number of epochs? [Here's](https://github.com/RasaHQ/rasa/blob/incremental_training/rasa/nlu/model.py#L176) how I had done it in the working version branch but I am assuming it wouldn't be exactly the same. If we can plan this ahead, it would unblock me on some of the implementation due for my [PR](https://github.com/RasaHQ/rasa/issues/7413). 🙏

joejuzl:
@dakshvar22 I haven' started looking at this area of the code yet (working on https://github.com/RasaHQ/rasa/issues/7330 first) so it's hard for me to say. @wochinge Any opinions/thoughts?

dakshvar22:
@joejuzl @wochinge Making a proposal to see if we can reach to a consensus quickly on the above question:

1. New number of epochs get changed inside the `meta` parameter that is passed to the `load` method of all components.
2. Add a boolean parameter to `load` method - `finetune_mode` which is set to `True` if the component is loaded in finetune mode.

What do you folks think?

joejuzl:
@dakshvar22 So the solution we have come up with (today!) is as follows:

For NLU:
- In `Interpreter.load` we pass in the new config along with the old model. 
- `Interpreter.load` updates the old metadata with the new epochs and calls `Interpreter.create` with a flag `should_finetune`.
- ` Interpreter.create` sets `should_finetune` in the context which gets passed to each component.

dakshvar22:
@joejuzl Perfect! Small clarification: 
> Interpreter.create sets `should_finetune` in the context which gets passed to each component.

The context dictionary is what will be passed as part of the `kwargs` argument of `load()` methods of each component?

joejuzl:
> The context dictionary is what will be passed as part of the kwargs argument of load() methods of each component?

Yes exactly, via `component_builder.load_component`

[...]

PR #7456, opened 2020-12-04
#7329 load models in finetune mode nlu

joejuzl:
**Proposed changes**:

Load NLU model in fine-tuneing mode

### DIFF
--- a/rasa/nlu/model.py
+++ b/rasa/nlu/model.py
@@ -323,7 +323,7 @@ class Interpreter:
         model_metadata = Metadata.load(model_dir)
 
         if new_config:
-            Interpreter._update_epochs_from_new_config(
+            Interpreter._update_metadata_epochs(
                 model_metadata, new_config, finetuning_epoch_fraction
             )
 
@@ -342,11 +342,11 @@ class Interpreter:
         return get_component_class(name).defaults[key]
 
     @staticmethod
-    def _update_epochs_from_new_config(
+    def _update_metadata_epochs(
         model_metadata: Metadata,
         new_config: Optional[Dict] = None,
         finetuning_epoch_fraction: float = 1.0,
-    ):
+    ) -> Metadata:
         for old_component_config, new_component_config in zip(
             model_metadata.metadata["pipeline"], new_config["pipeline"]
         ):
@@ -360,6 +360,7 @@ class Interpreter:
                 old_component_config[EPOCHS] = ceil(
                     new_epochs * finetuning_epoch_fraction
                 )
+        return model_metadata
 
     @staticmethod
     def create(
--- a/rasa/nlu/model.py
+++ b/rasa/nlu/model.py
@@ -138,6 +138,7 @@ class Trainer:
         cfg: RasaNLUModelConfig,
         component_builder: Optional[ComponentBuilder] = None,
         skip_validation: bool = False,
+        model_to_finetune: Optional["Interpreter"] = None,
     ):
 
         self.config = cfg
@@ -154,8 +155,10 @@ class Trainer:
         if not self.skip_validation:
             components.validate_requirements(cfg.component_names)
 
-        # build pipeline
-        self.pipeline = self._build_pipeline(cfg, component_builder)
+        if model_to_finetune:
+            self.pipeline = model_to_finetune.pipeline
+        else:
+            self.pipeline = self._build_pipeline(cfg, component_builder)
 
     def _build_pipeline(
         self, cfg: RasaNLUModelConfig, component_builder: ComponentBuilder
@@ -297,6 +300,8 @@ class Interpreter:
         model_dir: Text,
         component_builder: Optional[ComponentBuilder] = None,
         skip_validation: bool = False,
+        new_config: Optional[Dict] = None,
+        finetuning_epoch_fraction: float = 1.0,
     ) -> "Interpreter":
         """Create an interpreter based on a persisted model.
 
@@ -307,25 +312,62 @@ class Interpreter:
             model_dir: The path of the model to load
             component_builder: The
                 :class:`rasa.nlu.components.ComponentBuilder` to use.
+            new_config: Optional new config to use for the new epochs.
+            finetuning_epoch_fraction: Value to multiply all epochs by.
 
         Returns:
             An interpreter that uses the loaded model.
         """
-
         model_metadata = Metadata.load(model_dir)
 
+        if new_config:
+            Interpreter._update_epochs_from_new_config(
+                model_metadata, new_config, finetuning_epoch_fraction
+            )
+
         Interpreter.ensure_model_compatibility(model_metadata)
-        return Interpreter.create(model_metadata, component_builder, skip_validation)
+        return Interpreter.create(
+            model_metadata,
+            component_builder,
+            skip_validation,
+            should_finetune=new_config is not None,
+        )
+
+    @staticmethod
+    def _update_epochs_from_new_config(
+        model_metadata: Metadata,
+        new_config: Optional[Dict] = None,
+        finetuning_epoch_fraction: float = 1.0,
+    ):
+        for p1, p2 in zip(model_metadata.metadata["pipeline"], new_config["pipeline"]):
+            assert p1.get("name") == p2.get("name")
+            if "epochs" in p1:
+                p1["epochs"] = (
+                    p2.get("epochs", p1["epochs"]) * finetuning_epoch_fraction
+                )
 
     @staticmethod
     def create(
         model_metadata: Metadata,
         component_builder: Optional[ComponentBuilder] = None,
         skip_validation: bool = False,
+        should_finetune: bool = False,
     ) -> "Interpreter":
-        """Load stored model and components defined by the provided metadata."""
+        """Create model and components defined by the provided metadata.
 
-        context = {}
+        Args:
+            model_metadata: The metadata describing each component.
+            component_builder: The
+                :class:`rasa.nlu.components.ComponentBuilder` to use.
+            skip_validation: If set to `True`, does not check that all
+                required packages for the components are installed
+                before loading them.
+            should_finetune: Indicates if the model components will be fine-tuned.
+
+        Returns:
+            An interpreter that uses the created model.
+        """
+        context = {"finetune_mode": should_finetune}
 
         if component_builder is None:
             # If no builder is passed, every interpreter creation will result
--- a/rasa/nlu/model.py
+++ b/rasa/nlu/model.py
@@ -6,6 +6,7 @@ import os
 from typing import Any, Dict, List, Optional, Text
 
 import rasa.nlu
+# from rasa.nlu.registry import get_component_class
 from rasa.shared.exceptions import RasaException
 import rasa.shared.utils.io
 import rasa.utils.io
@@ -336,6 +337,11 @@ class Interpreter:
         )
 
     @staticmethod
+    def _get_default_value_for_component(name: Text, key: Text) -> Any:
+        # return get_component_class(name).defaults[key]
+        return 300
+
+    @staticmethod
     def _update_epochs_from_new_config(
         model_metadata: Metadata,
         new_config: Optional[Dict] = None,
@@ -345,9 +351,14 @@ class Interpreter:
             model_metadata.metadata["pipeline"], new_config["pipeline"]
         ):
             if EPOCHS in old_component_config:
+                new_epochs = new_component_config.get(
+                    EPOCHS,
+                    Interpreter._get_default_value_for_component(
+                        old_component_config["class"], EPOCHS
+                    ),
+                )
                 old_component_config[EPOCHS] = ceil(
-                    new_component_config.get(EPOCHS, old_component_config[EPOCHS])
-                    * finetuning_epoch_fraction
+                    new_epochs * finetuning_epoch_fraction
                 )
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


rasa/nlu/model.py
```diff
@@ -307,25 +313,63 @@ def load(
...
+                :class:`rasa.nlu.components.ComponentBuilder` to use.
+            skip_validation: If set to `True`, does not check that all
+                required packages for the components are installed
+                before loading them.
+            should_finetune: Indicates if the model components will be fine-tuned.
+
+        Returns:
+            An interpreter that uses the created model.
+        """
+        context = {"finetune_mode": should_finetune}
```
dakshvar22, line 372: Can we change this to `{"should_finetune": should_finetune}`? I already implemented it that way in all components. Will be easier to change here I guess 😅  (if you are okay with the variable name being `should_finetune` and not `finetune_mode`)

rasa/nlu/model.py
```diff
@@ -307,25 +313,63 @@ def load(
...
+        )
+
+    @staticmethod
+    def _update_epochs_from_new_config(
+        model_metadata: Metadata,
+        new_config: Optional[Dict] = None,
+        finetuning_epoch_fraction: float = 1.0,
+    ):
+        for p1, p2 in zip(model_metadata.metadata["pipeline"], new_config["pipeline"]):
+            if not p1.get("name") == p2.get("name"):
```
dakshvar22, line 344: Just checking `name` is not enough. Other parameters should also match exactly(except `epochs`).

dakshvar22: Also would be useful if we can mention in the raised exception which parameter does not match.

dakshvar22: Looking at other [PR](https://github.com/RasaHQ/rasa/pull/7427/files), it seems like you already check it there. Do we still need the check here then?

rasa/train.py
```diff
@@ -674,19 +675,18 @@ async def _train_nlu_with_validated_data(
         return _train_path
 
 
-def _nlu_model_for_finetuning(model_to_finetune: Text) -> Optional[Interpreter]:
-    from rasa.core.interpreter import RasaNLUInterpreter
-
+def _nlu_model_for_finetuning(
+    model_to_finetune: Text, new_config: Dict, finetuning_epoch_fraction: float = 1.0,
```
dakshvar22, line 679: Is it possible to add a more specific type for `new_config`?

rasa/nlu/model.py
```diff
@@ -307,25 +313,63 @@ def load(
...
+                raise InvalidModelError("Inconsistent config for model to fine-tune.")
+            if "epochs" in p1:
```
dakshvar22, line 346: Optional: Would it be more robust if we defined a list containing names of all components which have `epochs` as a parameter and use that for comparison here? Or would it add a dependency which people might forget to update. In that case, the list can also be auto-populated by looking at the defaults of each component.

dakshvar22 (changes requested):
Looking good ✨  Added a few suggestions/comments.

rasa/nlu/model.py
```diff
@@ -307,25 +313,63 @@ def load(
...
+                p1["epochs"] = ceil(
+                    p2.get("epochs", p1["epochs"]) * finetuning_epoch_fraction
```
dakshvar22, line 348: We always want the epochs mentioned in the `new_config` to be multiplied by the `epoch_fraction`. The way it's implemented currently would be problematic if `epochs` is not defined explicitly in the config.

dakshvar22: cc @wochinge Related to our discussion on the slack [thread](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1607103644498300)

rasa/nlu/model.py
joejuzl: So you are right that if anything apart from the epochs change we will not be able to fine-tune (due to the fingerprint checks). I included this name check as a sanity check - but maybe it's unnecessary.

rasa/nlu/model.py
joejuzl: What is the desired behaviour if epochs are not defined in the new config? With this implementation it would multiply the epochs in the previous config by `epoch_fraction`, is that not what we want?

rasa/nlu/model.py
dakshvar22: Yeah, I think they are a bit redundant. The fingerprint checks preserve the order of components in the pipeline as well right?

rasa/nlu/model.py
dakshvar22: Since the component would have used the default number of epochs in that case, we should multiply the `epoch_fraction` with the default number of epochs IMO. That would make the behaviour consistent with the idea of `epoch_fraction` - 
> We will always multiply the epochs based on the provided config

rasa/nlu/model.py
```diff
@@ -138,6 +139,7 @@ def __init__(
         cfg: RasaNLUModelConfig,
         component_builder: Optional[ComponentBuilder] = None,
         skip_validation: bool = False,
+        model_to_finetune: Optional["Interpreter"] = None,
     ):
```
wochinge, line 143:
```suggestion
    ) -> None:
```

rasa/nlu/model.py
```diff
@@ -307,25 +313,63 @@ def load(
...
+            skip_validation,
+            should_finetune=new_config is not None,
+        )
+
+    @staticmethod
+    def _update_epochs_from_new_config(
+        model_metadata: Metadata,
+        new_config: Optional[Dict] = None,
+        finetuning_epoch_fraction: float = 1.0,
+    ):
```
wochinge, line 342:
```suggestion
    ) -> None:
```

wochinge: optional: The code would feel a bit more intuitive to me if the function would return the updated `Metadata` object as it would put the focus on what part is updated exactly.
```suggestion
    ) -> Metadata:
```

rasa/nlu/model.py
wochinge: No strong feelings about this duplicate check.

rasa/nlu/model.py
```diff
@@ -307,25 +313,63 @@ def load(
...
+        for p1, p2 in zip(model_metadata.metadata["pipeline"], new_config["pipeline"]):
+            if not p1.get("name") == p2.get("name"):
+                raise InvalidModelError("Inconsistent config for model to fine-tune.")
```
wochinge, line 345: We need to agree on a consistent spelling of `fine[-]tuning` 😁

rasa/nlu/model.py
```diff
@@ -307,25 +313,63 @@ def load(
...
+            should_finetune=new_config is not None,
+        )
+
+    @staticmethod
+    def _update_epochs_from_new_config(
+        model_metadata: Metadata,
+        new_config: Optional[Dict] = None,
+        finetuning_epoch_fraction: float = 1.0,
+    ):
+        for p1, p2 in zip(model_metadata.metadata["pipeline"], new_config["pipeline"]):
```
wochinge, line 343: what do you think?
```suggestion
        for old_component_config, new_component_config in zip(model_metadata.metadata["pipeline"], new_config["pipeline"]):
```

rasa/nlu/model.py
wochinge: What exactly do you think would be more robust then?

rasa/nlu/model.py
wochinge: Would be great to have constant for `epochs`. This would also ensure that other components don't name the `epochs` param differently for some reason.

rasa/train.py
```diff
@@ -674,19 +675,20 @@ async def _train_nlu_with_validated_data(
...
+def _nlu_model_for_finetuning(
+    model_to_finetune: Text,
+    new_config: Dict[Text, Any],
+    finetuning_epoch_fraction: float = 1.0,
+) -> Optional[Interpreter]:
     path_to_archive = model.get_model_for_finetuning(model_to_finetune)
     if not path_to_archive:
         return None
 
-    try:
```
wochinge, line 684: I thought it's might be a good idea to just train from scratch in case anything goes wrong. On the other hand this can hide issues. If we remove it, we should do so, too, for the Core model loading.

tests/test_train.py
```diff
@@ -430,30 +431,44 @@ def test_model_finetuning_nlu(
...
+    monkeypatch.setattr(Interpreter, "create", mock_interpreter_create)
+
     (tmp_path / "models").mkdir()
     output = str(tmp_path / "models")
 
     if use_latest_model:
-        trained_rasa_model = str(Path(trained_rasa_model).parent)
+        trained_moodbot_path = str(Path(trained_moodbot_path).parent)
+
+    old_config = rasa.shared.utils.io.read_yaml_file("examples/moodbot/config.yml")
```
wochinge, line 449: Can we please add a comment why we are updating the config?

tests/test_train.py
```diff
@@ -430,30 +431,44 @@ def test_model_finetuning_nlu(
     monkeypatch: MonkeyPatch,
     default_domain_path: Text,
     default_nlu_data: Text,
-    default_stack_config: Text,
-    trained_rasa_model: Text,
+    trained_moodbot_path: Text,
```
wochinge, line 434: I believe we should also update the test for `test_model_finetuning`

tests/test_train.py
```diff
@@ -430,30 +431,44 @@ def test_model_finetuning_nlu(
...
+
     mocked_nlu_training.assert_called_once()
     _, kwargs = mocked_nlu_training.call_args
-    assert isinstance(kwargs["model_to_finetune"], Interpreter)
+    model_to_finetune = kwargs["model_to_finetune"]
+    assert isinstance(model_to_finetune, Interpreter)
+
+    new_diet_metadata = model_to_finetune.model_metadata.metadata["pipeline"][-1]
+    assert new_diet_metadata["name"] == "DIETClassifier"
+    assert new_diet_metadata["epochs"] == 5
```
wochinge, line 471: Can we please also test that the components were loaded in finetune mode?

wochinge (commented):
Left a couple of comments, but looking very good 💪

rasa/nlu/model.py
joejuzl: ok then will remove.

rasa/nlu/model.py
joejuzl: > Since the component would have used the default number of epochs in that case, we should multiply the epoch_fraction with the default number of epochs IMO. That would make the behaviour consistent with the idea of epoch_fraction

So if the model is trained first with config1 which has 10 epochs, and then is fine-tuned with config2 that has no epochs defined and epoch_fraction=0.5, then we ignore the 10 from the old config and instead multiply the default number of epochs by 0.5?

dakshvar22: Yes. We basically consider the current config as the source of truth. @wochinge Just making sure, you also agree with this?

rasa/nlu/model.py
joejuzl: I personally try to avoid methods which both mutate and return a value. I find it makes it unclear whether mutation is occurring. I could change the method to create a new metadata, although maybe that would be a pain. WDYT?

rasa/train.py
joejuzl: I would say better to fail fast and explicitly - especially as it's an experimental feature.

tests/test_train.py
joejuzl: Yes good point.

tests/test_train.py
joejuzl: I thought it would make sense to do it once both nlu and core were done.

rasa/nlu/model.py
joejuzl: it's a little harder to implement but it makes sense to me conceptually.

joejuzl: I am getting circular dependencies if I try to use `rasa.nlu.registry.get_component_class` to get the default epochs for a component... Another alternative is to pass `finetuning_epoch_fraction` into each component using the `context`. However this will require supporting it in all the components.

rasa/train.py
wochinge: Good point!

tests/test_train.py
wochinge: Ok, then let's add a `TODO` so we don't forget

rasa/nlu/model.py
wochinge: Makes sense. I'd expect that this is a very rare case though (e.g. the auto config feature suggests a config which includes a specified number of epochs).

wochinge: > Another alternative is to pass finetuning_epoch_fraction into each component using the context. However this will require supporting it in all the components.

Mhm, but `epochs` doesn't make sense for some components (e.g. the `CountVectorizer`). 

> I am getting circular dependencies if I try to use rasa.nlu.registry.get_component_class to get the default epochs for a component.

How about a local import?

rasa/nlu/model.py
joejuzl: Yes - I don't have strong feelings either way.

dakshvar22 (approved):
LGTM 💯

rasa/nlu/model.py
wochinge: > I could change the method to create a new metadata, although maybe that would be a pain. WDYT?

Mhm, it would be probably a `deepcopy` 🤔  Do you think naming it `_update_epochs_in_new_config` or `_update_config_with_epochs` would make the inplace change more explicit?

wochinge (approved):
💯

wochinge:
Great work and congrats to the merge @joejuzl 🎉

Issue #7408, opened 2020-11-30
Create endpoint for cross-validation


wochinge:
Create an endpoint for running a cross-validation as described in the implementation proposal for intent insights [here](https://www.notion.so/rasa/Implementation-Proposal-ff4f5cb9c37e43edac4552647af7dbc2#c060672e87cd4fd891cb1643b76b96a1).

**Todos**:
- [ ] Update API spec
- [ ] add endpoint to run cross-validation
   - by default it's a synchronous call and the response / error is returned in the response payload
   - If a callback URL is passed as query param, then the server should return immediately. The result (and potential error!) of the cross-validation is passed to the caller by calling the callback
   - We need to make sure that the synchronous cross-validation doesn't block other calls. We can e.g. do this by using `run_in_executor` as we do it for the training
- [ ] consider adding a start / end datetime to the cross-validation result. This would be useful for Rasa X (especially when reading this files from disk in local mode), to see on which data the cross-validation results are based upon

PR #7457, opened 2020-12-04
Cross val endpoint

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7408
- add option to use cross-validation to existing endpoint for intent evaluation `/model/test/intents`
- add option to run evaluation asynchronously by supporting a `callback url`
- correctly clean up used temporary directories (we didn't do this before)
- enable a solution for https://github.com/RasaHQ/rasa-x/issues/4083 by adding support for async `train` requests in the HTTP API
- deduplicate code in the `rasa.nlu.test` module


changelog/7408.improvement.md
```diff
@@ -0,0 +1,3 @@
+[`POST /model/test/intents`](/pages/http-api#operation/testModelIntent) now returns
```
wochinge, line 1: technically breaking, but the `report` was previously unusable.

wochinge:
I tested the callback using https://mockoon.com/

wochinge:
Uff, just discovered that the intent reports etc. actually override each other on disk for the different folds. I assume that's not what we want, right @dakshvar22 ?

wochinge:
@joejuzl Seems I need to refactor more stuff tomorrow. Still means you could review everything except the actual cross-validation

wochinge:
Deleted my messages. Overlooked some indent

wochinge:
@joejuzl Please review everything except the call to the cross-validation itself. Currently refactoring the `nlu.test.module`

wochinge:
Done with the refactoring.

docs/static/spec/rasa.yml
```diff
@@ -546,22 +549,48 @@ paths:
...
       requestBody:
         required: true
         content:
-          text/yml:
+          application/x-yaml:
             schema:
-              $ref: '#/components/schemas/NLUTrainingData'
+              type: string
+              description: >-
+                NLU training data and model configuration. The model configuration is
```
joejuzl, line 573: So the config and nlu data are provided together in one file?

changelog/7408.improvement.md
joejuzl: You would be surprised at the extents some people go to use "unusable" things...

rasa/nlu/test.py
```diff
@@ -201,16 +209,31 @@ def write_intent_successes(
...
+    if errors:
+        rasa.shared.utils.io.dump_obj_as_json_to_file(errors_filename, errors)
+        logger.info(f"Incorrect {error_type} predictions saved to {errors_filename}.")
+        logger.debug(
+            f"\n\nThese {error_type} examples could not be classified "
+            f"correctly: \n{errors}"
+        )
+    else:
+        if error_type.endswith("y"):
+            error_type_plural = f"{error_type[:-1]}ies"
```
joejuzl, line 229: this seems a bit hacky, I've used https://pypi.org/project/inflect/ in the past.

rasa/nlu/test.py
```diff
@@ -664,9 +633,53 @@ def evaluate_intents(
...
+        )
+        report = _add_confused_labels_to_report(
+            report,
+            confusion_matrix,
+            labels,
+            exclude_labels=[exclude_label] if exclude_label else [],
+        )
+
+        report_filename = os.path.join(output_directory, filename)
+        rasa.shared.utils.io.dump_obj_as_json_to_file(report_filename, report)
```
joejuzl, line 667: feels to me like file dumping shouldn't be part of the method `_calculate_report`.

rasa/nlu/test.py
```diff
@@ -664,9 +633,53 @@ def evaluate_intents(
...
 
+def _calculate_report(
+    filename: Text,
+    output_directory: Optional[Text],
+    targets: Iterable[Any],
+    predictions: Iterable[Any],
+    report_as_dict: bool = False,
+    exclude_label: Optional[Text] = None,
+) -> Tuple[Union[Text, Dict], float, float, float, np.ndarray, List[Text]]:
+    from rasa.test import get_evaluation_metrics
```
joejuzl, line 648: Do we have general guidelines or rules for when we do local imports?

rasa/nlu/test.py
```diff
@@ -664,9 +633,53 @@ def evaluate_intents(
...
+    import sklearn.metrics
+    import sklearn.utils.multiclass
+
+    confusion_matrix = sklearn.metrics.confusion_matrix(targets, predictions)
+    labels = sklearn.utils.multiclass.unique_labels(targets, predictions)
+
+    if output_directory:
+        report, precision, f1, accuracy = get_evaluation_metrics(
+            targets, predictions, output_dict=True, exclude_label=exclude_label
```
joejuzl, line 657: This is weird behaviour because providing an `report_as_dict` forces `output_dict` to be `True` even if it was specified as `False`.

rasa/server.py
```diff
@@ -476,6 +479,124 @@ async def hello(request: Request):
...
+
+            except Exception as e:
+                if not isinstance(e, ErrorResponse):
+                    logger.error(e)
+                    e = ErrorResponse(
+                        HTTPStatus.INTERNAL_SERVER_ERROR,
+                        "UnexpectedError",
+                        f"An unexpected error occurred. Error: {e}",
+                    )
+                # If an error happens, we sent the error payload to the `callback_url`
```
joejuzl, line 526:
```suggestion
                # If an error happens, we send the error payload to the `callback_url`
```

rasa/server.py
```diff
@@ -476,6 +479,124 @@ async def hello(request: Request):
...
+        # Only process request asynchronously if the user specified a `callback_url`
+        # query parameter.
+        if not callback_url:
+            return await f(request, *args, **kwargs)
+
+        async def wrapped() -> None:
+            try:
+                result: HTTPResponse = await f(request, *args, **kwargs)
+                payload = dict(
+                    data=result.body, headers={"Content-Type": result.content_type}
```
joejuzl, line 511: Any other headers we may want to forward?

rasa/server.py
```diff
@@ -476,6 +479,124 @@ async def hello(request: Request):
...
+        # Run the request in the background on the event loop
+        request.app.add_task(wrapped())
+
+        # The incoming request will return immediately with a 204
+        return response.empty()
+
+    return decorated_function
+
+
+def heavy_computation(f: Callable[..., Coroutine]) -> Callable:
```
joejuzl, line 544: This name feels ambiguous to me, better I think to give it a name which describes what it does rather than what it's used for.

rasa/server.py
```diff
@@ -476,6 +479,124 @@ async def hello(request: Request):
...
+    """Decorator which runs request on a separate thread.
+
+    Some requests (e.g. training or cross-validation) are computional intense requests.
+    This means that they will block the event loop and hence the processing of other
+    requests. This decorator can be used to process these requests on a separate thread
```
joejuzl, line 549: Won't this be affected by the GIL? Any reason not to use multi processing instead?

rasa/server.py
```diff
@@ -476,6 +479,124 @@ async def hello(request: Request):
...
+            # This is a new thread, so we need to create and set a new event loop
+            thread_loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(thread_loop)
+
+            try:
+                return thread_loop.run_until_complete(f(request, *args, **kwargs))
+            finally:
+                thread_loop.close()
+
+        # Run decorated function in thread (`None` will use the `ThreadPoolExecutor`)
```
joejuzl, line 575: Better to explicitly set it than rely on a comment I think.

rasa/server.py
```diff
@@ -951,19 +1072,61 @@ async def evaluate_stories(request: Request) -> HTTPResponse:
...
+            test_coroutine = _cross_validate(
+                test_data, config_file, int(cross_validation_folds)
+            )
+        elif cross_validation_folds:
+            raise ErrorResponse(
+                HTTPStatus.BAD_REQUEST,
+                "TestingError",
+                "Cross-validation is only supported for YAML data.",
+            )
+        else:
```
joejuzl, line 1109: Since all these `if`, `elif` and `else`s are referring to 2 flags (`cross_validation_folds`, and `is_yaml_payload`) a nested `if` might be clearer?

rasa/server.py
```diff
@@ -984,16 +1147,50 @@ async def evaluate_intents(request: Request) -> HTTPResponse:
...
-                "TestingError",
-                f"An unexpected error occurred during evaluation. Error: {e}",
-            )
+        return run_evaluation(
+            data_path, nlu_model, disable_plotting=True, report_as_dict=True
+        )
+
+    async def _cross_validate(data_file: Text, config_file: Text, folds: int) -> Dict:
+        importer = TrainingDataImporter.load_from_dict(
+            None, config_path=config_file, training_data_paths=[data_file]
```
joejuzl, line 1156: At this point might make sense to use kwargs for all args.

rasa/server.py
```diff
@@ -1277,36 +1474,30 @@ def _validate_json_training_payload(rjs: Dict):
...
-    temp_dir = tempfile.mkdtemp()
-    training_data = Path(temp_dir) / "data.yml"
+    training_data = temp_dir / "data.yml"
     rasa.shared.utils.io.write_text_file(decoded, training_data)
 
-    model_output_directory = _model_output_directory(
-        request.args.get("save_to_default_model_directory", True)
-    )
+    model_output_directory = str(temp_dir)
+    if request.args.get("save_to_default_model_directory", True):
```
joejuzl, line 1489: In my opinion request args should all be extracted at the top level.

joejuzl (changes requested):

joejuzl:
The decorators are very cool 😎

rasa/server.py
```diff
@@ -476,6 +479,124 @@ async def hello(request: Request):
         return response.text("Hello from Rasa: " + rasa.__version__)
 
 
+def async_if_callback_url(f: Callable[..., Coroutine]) -> Callable:
```
joejuzl, line 482: We don't have any equivalent of integration tests for rasa OSS where we could test these right?

wochinge: Nope. However, I added a test unit test for the cross-validation which mocks the callback URL and checks if the request was done as expected.

docs/static/spec/rasa.yml
wochinge: yep, one long yaml string. Since 2.0 everything can be in one file so we don't really separate anymore.

changelog/7408.improvement.md
wochinge: would you consider this as blocking?

rasa/server.py
wochinge: we rely on the Python default in that case 😄  Will make it explicit

rasa/nlu/test.py
wochinge: In this case it avoid a cyclic import and sklearn is a very heavy dependency which means the local import avoid long loading times of the CLI (or better: avoids even longer loading times). 

But no, we don't have them.

rasa/nlu/test.py
wochinge: done

rasa/server.py
wochinge: you mean right in the request handler instead of submethod?

rasa/server.py
wochinge: Mhm, good question. I'd expect that most numpy / TensorFlow / sklearn things run in C(++) anyway and the GIL wouldn't kick in. No strong opinion though. In addition it's a bit more convenient in combination with a local function 😁

rasa/server.py
wochinge: Mhm, everything else should be set automatically from scratch. We just need to encoding as only obtain the already encoded data. I also can't see other default from Sani when we call `response.json` than setting the content type

rasa/nlu/test.py
wochinge: got rid of it by getting rid of the plural 😂

changelog/7408.improvement.md
wochinge: We could make it a query arg, but I'd go for the risk here. The other format was really not usable.

joejuzl: I guess the only risk was someone was parsing it somehow, or directly displaying it.

rasa/server.py
joejuzl: Would be good to (stress)test at some point - but this seems fine to begin with at least.

joejuzl (approved):
Looks good!

rasa/server.py
wochinge: So far it worked fine for the training. Let's see how this develops as the models grow / get more complex

PR #7458, opened 2020-12-04
#7329 load models in finetune mode core

joejuzl:
**Proposed changes**:
Load the core models in fine-tuning mode.

### DIFF
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -7,7 +7,7 @@ import sys
 from collections import defaultdict
 from datetime import datetime
 from pathlib import Path
-from typing import Text, Optional, Any, List, Dict, Tuple, Union
+from typing import Text, Optional, Any, List, Dict, Tuple, Type, Union
 
 import rasa.core
 import rasa.core.training.training
@@ -42,6 +42,7 @@ from rasa.core.policies.rule_policy import RulePolicy
 from rasa.shared.core.trackers import DialogueStateTracker
 from rasa.shared.core.generator import TrackerWithCachedStates
 from rasa.core import registry
+from rasa.utils.tensorflow.constants import EPOCHS
 
 logger = logging.getLogger(__name__)
 
@@ -304,11 +305,19 @@ class PolicyEnsemble:
             )
 
     @staticmethod
-    def _get_updated_epochs(config_for_policy, finetuning_epoch_fraction):
-        epochs = None
-        if "epochs" in config_for_policy:
-            epochs = math.ceil(config_for_policy["epochs"] * finetuning_epoch_fraction)
-        return epochs
+    def _get_updated_epochs(
+        policy_cls: Type[Policy],
+        config_for_policy: Dict[Text, Any],
+        finetuning_epoch_fraction: float,
+    ):
+        if EPOCHS in config_for_policy:
+            epochs = config_for_policy[EPOCHS]
+        else:
+            try:
+                epochs = policy_cls.defaults[EPOCHS]
+            except (KeyError, AttributeError):
+                return None
+        return math.ceil(epochs * finetuning_epoch_fraction)
 
     @classmethod
     def load(
@@ -339,7 +348,7 @@ class PolicyEnsemble:
 
                 config_for_policy = new_config["policies"][i]
                 epochs = cls._get_updated_epochs(
-                    config_for_policy, finetuning_epoch_fraction
+                    policy_cls, config_for_policy, finetuning_epoch_fraction
                 )
                 policy = policy_cls.load(
                     policy_path, should_finetune=True, epoch_override=epochs
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -309,7 +309,7 @@ class PolicyEnsemble:
         policy_cls: Type[Policy],
         config_for_policy: Dict[Text, Any],
         finetuning_epoch_fraction: float,
-    ):
+    ) -> Optional[int]:
         if EPOCHS in config_for_policy:
             epochs = config_for_policy[EPOCHS]
         else:
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -1,12 +1,13 @@
 import importlib
 import json
 import logging
+from math import ceil
 import os
 import sys
 from collections import defaultdict
 from datetime import datetime
 from pathlib import Path
-from typing import Text, Optional, Any, List, Dict, Tuple, NamedTuple, Union
+from typing import Text, Optional, Any, List, Dict, Tuple, Union
 
 import rasa.core
 import rasa.core.training.training
@@ -303,9 +304,13 @@ class PolicyEnsemble:
             )
 
     @classmethod
-    def load(cls, path: Union[Text, Path]) -> "PolicyEnsemble":
-        """Loads policy and domain specification from storage"""
-
+    def load(
+        cls,
+        path: Union[Text, Path],
+        new_config: Optional[Dict] = None,
+        finetuning_epoch_fraction: float = 1.0,
+    ) -> "PolicyEnsemble":
+        """Loads policy and domain specification from storage."""
         metadata = cls.load_metadata(path)
         cls.ensure_model_compatibility(metadata)
         policies = []
@@ -313,7 +318,27 @@ class PolicyEnsemble:
             policy_cls = registry.policy_from_module_path(policy_name)
             dir_name = f"policy_{i}_{policy_cls.__name__}"
             policy_path = os.path.join(path, dir_name)
-            policy = policy_cls.load(policy_path)
+
+            if new_config:
+                if "should_finetune" not in rasa.shared.utils.common.arguments_of(
+                    policy_cls.load
+                ):
+                    raise UnsupportedDialogueModelError(
+                        f"{policy_cls.__name__} does not support fine-tuning"
+                    )
+
+                config_for_policy = new_config["policies"][i]
+                epochs = None
+                if "epochs" in config_for_policy:
+                    epochs = ceil(
+                        config_for_policy["epochs"] * finetuning_epoch_fraction
+                    )
+                policy = policy_cls.load(
+                    policy_path, should_finetune=True, epoch_override=epochs
+                )
+            else:
+                policy = policy_cls.load(policy_path)
+
             cls._ensure_loaded_policy(policy, policy_cls, policy_name)
             policies.append(policy)
         ensemble_cls = rasa.shared.utils.common.class_from_module_path(
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -349,16 +349,16 @@ class PolicyEnsemble:
             if "kwargs" not in rasa.shared.utils.common.arguments_of(policy_cls.load):
                 if context:
                     raise UnsupportedDialogueModelError(
-                        f"`{policy_cls.__name__}.load` does not accept **kwargs. "
-                        f"Attempting to pass {context} to the policy. "
-                        f"This argument should be added to all policies by "
+                        f"`{policy_cls.__name__}.{policy_cls.load.__name__}` does not "
+                        f"accept `**kwargs`. Attempting to pass {context} to the "
+                        f"policy. `**kwargs` should be added to all policies by "
                         f"Rasa Open Source 3.0.0."
                     )
                 else:
                     rasa.shared.utils.io.raise_deprecation_warning(
-                        f"{policy_cls.__name__} `load` method does not "
-                        f"accept **kwargs. This is required for contextual"
-                        f" information e.g. the flag `should_finetune`.",
+                        f"`{policy_cls.__name__}.{policy_cls.load.__name__}` does not "
+                        f"accept `**kwargs`. `**kwargs` are required for contextual "
+                        f"information e.g. the flag `should_finetune`.",
                         warn_until_version="3.0.0",
                     )
 
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -350,9 +350,23 @@ class PolicyEnsemble:
                 epochs = cls._get_updated_epochs(
                     policy_cls, config_for_policy, finetuning_epoch_fraction
                 )
-                policy = policy_cls.load(
-                    policy_path, should_finetune=True, epoch_override=epochs
-                )
+                if epochs:
+                    if "epoch_override" not in rasa.shared.utils.common.arguments_of(
+                        policy_cls.load
+                    ):
+                        raise UnsupportedDialogueModelError(
+                            f"{policy_cls.__name__} does not support fine-tuning. "
+                            f"To support fine-tuning the `load` method must have the "
+                            f"argument `epoch_override` if the policy uses epochs."
+                            f"This argument should be added to all policies by "
+                            f"Rasa Open Source 3.0.0."
+                        )
+
+                    policy = policy_cls.load(
+                        policy_path, should_finetune=True, epoch_override=epochs
+                    )
+                else:
+                    policy = policy_cls.load(policy_path, should_finetune=True)
             else:
                 policy = policy_cls.load(policy_path)
 
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -326,7 +326,7 @@ class PolicyEnsemble:
         new_config: Optional[Dict] = None,
         finetuning_epoch_fraction: float = 1.0,
     ) -> "PolicyEnsemble":
-        """Loads policy and domain specification from storage."""
+        """Loads policy and domain specification from disk."""
         metadata = cls.load_metadata(path)
         cls.ensure_model_compatibility(metadata)
         policies = []
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -1,7 +1,7 @@
 import importlib
 import json
 import logging
-from math import ceil
+import math
 import os
 import sys
 from collections import defaultdict
@@ -303,6 +303,13 @@ class PolicyEnsemble:
                 "".format(policy_name)
             )
 
+    @staticmethod
+    def _get_updated_epochs(config_for_policy, finetuning_epoch_fraction):
+        epochs = None
+        if "epochs" in config_for_policy:
+            epochs = math.ceil(config_for_policy["epochs"] * finetuning_epoch_fraction)
+        return epochs
+
     @classmethod
     def load(
         cls,
@@ -324,15 +331,16 @@ class PolicyEnsemble:
                     policy_cls.load
                 ):
                     raise UnsupportedDialogueModelError(
-                        f"{policy_cls.__name__} does not support fine-tuning"
+                        f"{policy_cls.__name__} does not support fine-tuning. "
+                        f"To support fine-tuning the `load` method must have the "
+                        f"argument `should_finetune`. This argument should be added to "
+                        f"all policies by Rasa Open Source 3.0.0."
                     )
 
                 config_for_policy = new_config["policies"][i]
-                epochs = None
-                if "epochs" in config_for_policy:
-                    epochs = ceil(
-                        config_for_policy["epochs"] * finetuning_epoch_fraction
-                    )
+                epochs = cls._get_updated_epochs(
+                    config_for_policy, finetuning_epoch_fraction
+                )
                 policy = policy_cls.load(
                     policy_path, should_finetune=True, epoch_override=epochs
                 )
--- a/rasa/core/policies/ensemble.py
+++ b/rasa/core/policies/ensemble.py
@@ -335,43 +335,37 @@ class PolicyEnsemble:
             dir_name = f"policy_{i}_{policy_cls.__name__}"
             policy_path = os.path.join(path, dir_name)
 
+            context = {}
             if new_config:
-                if "should_finetune" not in rasa.shared.utils.common.arguments_of(
-                    policy_cls.load
-                ):
-                    raise UnsupportedDialogueModelError(
-                        f"{policy_cls.__name__} does not support fine-tuning. "
-                        f"To support fine-tuning the `load` method must have the "
-                        f"argument `should_finetune`. This argument should be added to "
-                        f"all policies by Rasa Open Source 3.0.0."
-                    )
+                context["should_finetune"] = True
 
                 config_for_policy = new_config["policies"][i]
                 epochs = cls._get_updated_epochs(
                     policy_cls, config_for_policy, finetuning_epoch_fraction
                 )
                 if epochs:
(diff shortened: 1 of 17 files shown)


dakshvar22 (changes requested):
Haven't reviewed extensively yet, but an initial observation.

rasa/core/policies/policy.py
```diff
@@ -285,11 +296,24 @@ def load(cls, path: Union[Text, Path]) -> "Policy":
...
             if (Path(path) / FEATURIZER_FILE).is_file():
                 featurizer = TrackerFeaturizer.load(path)
                 data["featurizer"] = featurizer
 
+            # TODO: figure out this mess
+            constructure_args = rasa.shared.utils.common.arguments_of(cls)
+            if "kwargs" not in constructure_args and (
+                (should_finetune and "should_finetune" not in constructure_args)
+                or (epoch_override or "epochs" not in constructure_args)
+            ):
```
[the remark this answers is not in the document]
joejuzl: It's ended up being simpler now we explicitly required the argument.

rasa/core/policies/ensemble.py
```diff
@@ -303,17 +304,41 @@ def _ensure_loaded_policy(cls, policy, policy_cls, policy_name: Text):
...
         cls.ensure_model_compatibility(metadata)
         policies = []
         for i, policy_name in enumerate(metadata["policy_names"]):
             policy_cls = registry.policy_from_module_path(policy_name)
             dir_name = f"policy_{i}_{policy_cls.__name__}"
             policy_path = os.path.join(path, dir_name)
-            policy = policy_cls.load(policy_path)
+
+            if new_config:
+                if "should_finetune" not in rasa.shared.utils.common.arguments_of(
```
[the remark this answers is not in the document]
joejuzl: I have extracted part of it.

rasa/core/policies/ensemble.py
```diff
@@ -302,18 +304,72 @@ def _ensure_loaded_policy(cls, policy, policy_cls, policy_name: Text):
...
+                if epochs:
+                    if "epoch_override" not in rasa.shared.utils.common.arguments_of(
+                        policy_cls.load
+                    ):
+                        raise UnsupportedDialogueModelError(
+                            f"{policy_cls.__name__} does not support fine-tuning. "
+                            f"To support fine-tuning the `load` method must have the "
+                            f"argument `epoch_override` if the policy uses epochs."
+                            f"This argument should be added to all policies by "
+                            f"Rasa Open Source 3.0.0."
```
dakshvar22, line 362: Should this be a warning rather than an Exception? Reason being that users may run `rasa train --finetune` with custom policies and the command will just fail even if they had compatible NLU components. Since we are saying that this argument should be added to all policies by 3.0.0, I think we should make a call to `load` with `policy_cls.load(policy_path)` and output a warning saying `training may fail/may be incorrect because the load method does not contain these arguments. See migration guide on how to migrate custom policies for incremental training`. What do you think?

rasa/core/policies/ensemble.py
```diff
@@ -302,18 +304,72 @@ def _ensure_loaded_policy(cls, policy, policy_cls, policy_name: Text):
...
+                        )
+
+                    policy = policy_cls.load(
+                        policy_path, should_finetune=True, epoch_override=epochs
+                    )
+                else:
+                    policy = policy_cls.load(policy_path, should_finetune=True)
```
dakshvar22, line 369: I guess this should also be covered in the above `if` construct which should throw a warning for unsupported policy otherwise for custom policies with custom method for `load`, this line will error out?

rasa/core/policies/policy.py
```diff
@@ -285,6 +293,15 @@ def load(cls, path: Union[Text, Path]) -> "Policy":
...
             data = json.loads(rasa.shared.utils.io.read_file(metadata_file))
+            if should_finetune:
+                if "should_finetune" not in rasa.shared.utils.common.arguments_of(cls):
+                    raise UnsupportedDialogueModelError(
+                        f"{cls.__name__} does not support fine-tuning. "
+                        f"To support fine-tuning the `__init__` method must have the "
+                        f"argument `should_finetune`. This argument should be added to "
+                        f"all policies by Rasa Open Source 3.0.0."
+                    )
+                data["should_finetune"] = should_finetune
```
dakshvar22, line 304: I am not sure if adding this to the base class makes a lot of sense. Ability to finetune a model depends on the type of the model. While deep learning based models need a specific implementation to finetune themselves, other simple ML models do not and it's easier to just "train" them on the complete new data. So, I would not make `should_finetune` a parameter of the constructor of `Policy`. It depends on the model of the `Policy` if it needs to be there.
The idea is similar to what we do for NLU components. We haven't made `should_finetune` a parameter of the `Component` class but a parameter for specific ML components which require it.

dakshvar22: If we remove this argument from the constructor, then custom policies which do not differ in implementation for training the policy from scratch and training the policy in finetune mode will not have to be refactored at all.

rasa/core/policies/ensemble.py
wochinge: I'd vote to fail early. They shouldn't use `--finetune` if it's not supported. Training a policy which does't support finetuning might lead to weird and unexpected behavior which seems more problematic than giving them an explicit signal.

rasa/core/policies/ensemble.py
```diff
@@ -302,18 +304,72 @@ def _ensure_loaded_policy(cls, policy, policy_cls, policy_name: Text):
...
+                    policy_cls, config_for_policy, finetuning_epoch_fraction
+                )
+                if epochs:
+                    if "epoch_override" not in rasa.shared.utils.common.arguments_of(
+                        policy_cls.load
+                    ):
+                        raise UnsupportedDialogueModelError(
+                            f"{policy_cls.__name__} does not support fine-tuning. "
+                            f"To support fine-tuning the `load` method must have the "
+                            f"argument `epoch_override` if the policy uses epochs."
```
wochinge, line 360:
```suggestion
                            f"argument `epoch_override` if the policy uses epochs. "
```

rasa/core/policies/memoization.py
```diff
@@ -80,8 +81,8 @@ def __init__(
             max_history: maximum history to take into account when featurizing trackers
             lookup: a dictionary that stores featurized tracker states and
                 predicted actions for them
+            should_finetune: Indicates if the model components will be fine-tuned.
```
wochinge, line 84:
```suggestion
            should_finetune: Indicates if the model components will be finetuned.
```

rasa/core/policies/policy.py
```diff
@@ -110,12 +113,16 @@ def __init__(
         self,
         featurizer: Optional[TrackerFeaturizer] = None,
         priority: int = DEFAULT_POLICY_PRIORITY,
+        should_finetune: bool = False,
     ) -> None:
+        """Construct a new Policy object."""
```
wochinge, line 118:
```suggestion
        """Constructs a new Policy object."""
```

rasa/core/policies/policy.py
```diff
@@ -272,11 +279,12 @@ def persist(self, path: Union[Text, Path]) -> None:
         rasa.shared.utils.io.dump_obj_as_json_to_file(file, self._metadata())
 
     @classmethod
-    def load(cls, path: Union[Text, Path]) -> "Policy":
+    def load(cls, path: Union[Text, Path], should_finetune: bool = False,) -> "Policy":
         """Loads a policy from path.
 
         Args:
             path: Path to load policy from.
+            should_finetune: Indicates if the model components will be fine-tuned.
```
wochinge, line 287:
```suggestion
            should_finetune: Indicates if the model components will be finetuned.
```

rasa/core/policies/policy.py
wochinge: I see your argument. What about having a `context` dictionary as argument for the constructor for **all** policies? That would also solve the `epoch_override` problem neatly. Using a `context` argument we'd have 
1) A common interface for all policies
2) Policies can decide themselves if they want to handle it
3) Similar approach to NLU which makes the codebase quicker to understand

rasa/core/policies/rule_policy.py
```diff
@@ -124,6 +125,11 @@ def __init__(
                 if no rule matched.
             enable_fallback_prediction: If `True` `core_fallback_action_name` is
                 predicted in case no rule matched.
+            restrict_rules: If `True` rules are restricted to contain a maximum of 1
+                user message. This is used to avoid that users build a state machine
+                using the rules.
+            check_for_contradictions: Check for contradictions.
+            should_finetune: Indicates if the model components will be fine-tuned.
```
wochinge, line 132:
```suggestion
            should_finetune: Indicates if the model components will be finetuned.
```

rasa/core/policies/sklearn_policy.py
```diff
@@ -84,8 +87,8 @@ def __init__(
                 *inverse_transform* method.
             shuffle: Whether to shuffle training data.
             zero_state_features: Contains default feature values for attributes
+            should_finetune: Indicates if the model components will be fine-tuned.
```
wochinge, line 90:
```suggestion
            should_finetune: Indicates if the model components will be finetuned.
```

rasa/core/policies/ensemble.py
dakshvar22: Fair point, agreed. 👍

rasa/core/policies/policy.py
dakshvar22: Yes, having a context dictionary sounds like a good idea 👍

joejuzl: There are two interfaces to a `Policy`, `load` and `__init__`.
`load` is called from the ensemble, and currently we are supporting:
```
def load( cls, path: Union[Text, Path], should_finetune: bool = False, epoch_override: Optional[float] = None)
def load( cls, path: Union[Text, Path], should_finetune: bool = False)
def load( cls, path: Union[Text, Path])
```
The next interface, `__init__` is called from the base class `Policy.load`, and currently we are supporting:
```
def __init__(self, args, should_finetune: bool = False)
def __init__(self, args)
```
Policies like `TEDPolicy` have a custom load, so they then handle passing `epochs` into their constructor.

So, should we have a `context` (essentially like the `kwargs` in my first draft version) in both?
And do we remove all warnings related to arguments? Or only warn if `context` is not there?

rasa/core/policies/ensemble.py
```diff
@@ -302,20 +304,68 @@ def _ensure_loaded_policy(cls, policy, policy_cls, policy_name: Text):
...
 
-    @classmethod
-    def load(cls, path: Union[Text, Path]) -> "PolicyEnsemble":
-        """Loads policy and domain specification from storage"""
+    @staticmethod
+    def _get_updated_epochs(
+        policy_cls: Type[Policy],
+        config_for_policy: Dict[Text, Any],
+        finetuning_epoch_fraction: float,
+    ):
```
dakshvar22, line 312: return type
```suggestion
    ) -> int:
```

rasa/core/policies/policy.py
```diff
@@ -386,7 +412,7 @@ def __init__(
                 you return as they can potentially influence the conversation flow.
             is_end_to_end_prediction: `True` if the prediction used the text of the
                 user message instead of the intent.
-        """
+         """
```
dakshvar22, line 415:
```suggestion
        """
```

rasa/core/policies/sklearn_policy.py
```diff
@@ -321,10 +323,15 @@ def load(cls, path: Union[Text, Path]) -> Policy:
         meta = json.loads(rasa.shared.utils.io.read_file(meta_file))
         zero_state_features = io_utils.pickle_load(zero_features_filename)
 
+        data = {"should_finetune": kwargs.get("should_finetune", False)}
+        if "epoch_override" in kwargs:
+            data[EPOCHS] = kwargs["epoch_override"]
+
```
dakshvar22, line 329: We don't need this as `SklearnPolicy` does not use `epochs`.

rasa/core/policies/ted_policy.py
```diff
@@ -500,6 +500,10 @@ def load(cls, path: Union[Text, Path]) -> "TEDPolicy":
         )
         model.build_for_predict(predict_data_example)
 
+        meta["should_finetune"] = kwargs.get("should_finetune", False)
+        if "epoch_override" in kwargs:
+            meta[EPOCHS] = kwargs["epoch_override"]
+
```
dakshvar22, line 506: This needs a larger change which i'll address in my PR as part of making core policies finetunable. It's okay to keep as it is here.

dakshvar22 (approved):
Looking great 👍

rasa/core/policies/ensemble.py
wochinge: edited the suggestion :-)

rasa/core/policies/policy.py
```diff
@@ -290,6 +296,26 @@ def load(cls, path: Union[Text, Path]) -> "Policy":
                 featurizer = TrackerFeaturizer.load(path)
                 data["featurizer"] = featurizer
 
+            if "should_finetune" in kwargs:
+                data["should_finetune"] = kwargs["should_finetune"]
```
wochinge, line 300: shouldn't we do something like `data.update(kwargs)` to include all kwargs into `data`?

rasa/core/policies/policy.py
```diff
@@ -290,6 +296,26 @@ def load(cls, path: Union[Text, Path]) -> "Policy":
...
+                if set(data.keys()).issubset(set(constructor_args)):
+                    rasa.shared.utils.io.raise_deprecation_warning(
+                        f"{cls.__name__} `__init__` method does not accept **kwargs "
+                        f"This is required for contextual information e.g. the flag "
+                        f"`should_finetune`.",
+                        warn_until_version="3.0.0",
+                    )
+                else:
+                    raise UnsupportedDialogueModelError(
+                        f"{cls.__name__} `__init__` method does not accept **kwargs. "
```
wochinge, line 313:
```suggestion
                        f"`{cls.__name__}.__init__` method does not accept `**kwargs`. "
```

rasa/core/policies/policy.py
```diff
@@ -290,6 +296,26 @@ def load(cls, path: Union[Text, Path]) -> "Policy":
...
                 data["featurizer"] = featurizer
 
+            if "should_finetune" in kwargs:
+                data["should_finetune"] = kwargs["should_finetune"]
+
+            constructor_args = rasa.shared.utils.common.arguments_of(cls)
+            if "kwargs" not in constructor_args:
+                if set(data.keys()).issubset(set(constructor_args)):
+                    rasa.shared.utils.io.raise_deprecation_warning(
+                        f"{cls.__name__} `__init__` method does not accept **kwargs "
```
wochinge, line 306:
```suggestion
                        f"`{cls.__name__}.__init__` method does not accept *`*kwargs`. "
```

rasa/core/policies/rule_policy.py
```diff
@@ -108,6 +108,7 @@ def __init__(
         enable_fallback_prediction: bool = True,
         restrict_rules: bool = True,
         check_for_contradictions: bool = True,
+        **kwargs,
```
wochinge, line 111:
```suggestion
        **kwargs: Any,
```

rasa/core/policies/sklearn_policy.py
```diff
@@ -302,7 +303,8 @@ def persist(self, path: Union[Text, Path]) -> None:
             )
 
     @classmethod
-    def load(cls, path: Union[Text, Path]) -> Policy:
+    def load(cls, path: Union[Text, Path], **kwargs,) -> Policy:
```
wochinge, line 306:
```suggestion
    def load(cls, path: Union[Text, Path], **kwargs: Any,) -> Policy:
```

rasa/core/policies/sklearn_policy.py
```diff
@@ -302,7 +303,8 @@ def persist(self, path: Union[Text, Path]) -> None:
...
+        """See the docstring for `Policy.Load`."""
```
wochinge, line 307:
```suggestion
        """See the docstring for `Policy.load`."""
```

rasa/core/policies/sklearn_policy.py
```diff
@@ -321,10 +323,15 @@ def load(cls, path: Union[Text, Path]) -> Policy:
         meta = json.loads(rasa.shared.utils.io.read_file(meta_file))
         zero_state_features = io_utils.pickle_load(zero_features_filename)
 
+        data = {"should_finetune": kwargs.get("should_finetune", False)}
```
wochinge, line 326: we can access the `should_fine` directly, no?

rasa/core/policies/ted_policy.py
```diff
@@ -437,8 +436,9 @@ def persist(self, path: Union[Text, Path]) -> None:
         )
 
     @classmethod
-    def load(cls, path: Union[Text, Path]) -> "TEDPolicy":
+    def load(cls, path: Union[Text, Path], **kwargs,) -> "TEDPolicy":
```
wochinge, line 439:
```suggestion
    def load(cls, path: Union[Text, Path], **kwargs: Any,) -> "TEDPolicy":
```

rasa/core/policies/sklearn_policy.py
wochinge: (same for the others)

rasa/core/policies/two_stage_fallback.py
```diff
@@ -62,10 +62,12 @@ def __init__(
         fallback_core_action_name: Text = ACTION_DEFAULT_FALLBACK_NAME,
         fallback_nlu_action_name: Text = ACTION_DEFAULT_FALLBACK_NAME,
         deny_suggestion_intent_name: Text = USER_INTENT_OUT_OF_SCOPE,
+        **kwargs,
```
wochinge, line 65:
```suggestion
        **kwargs: Any,
```

tests/core/test_ensemble.py
```diff
@@ -83,16 +118,17 @@ def __init__(
         is_end_to_end_prediction: bool = False,
         events: Optional[List[Event]] = None,
         optional_events: Optional[List[Event]] = None,
+        **kwargs,
```
wochinge, line 121:
```suggestion
        **kwargs: Any,
```

tests/core/test_policies.py
```diff
@@ -1229,3 +1231,34 @@ def test_get_training_trackers_for_policy(
...
+
+class PolicyWithoutInitKwargs(Policy):
+    def __init__(self, *args):
+        pass
+
+    def persist(self, _) -> None:
+        pass
+
+    @classmethod
+    def _metadata_filename(cls):
```
wochinge, line 1244:
```suggestion
    def _metadata_filename(cls) -> Text:
```

tests/core/test_policies.py
```diff
@@ -1229,3 +1231,34 @@ def test_get_training_trackers_for_policy(
 def test_deprecation_warnings_for_old_rule_like_policies(policy: Type[Policy]):
     with pytest.warns(FutureWarning):
         policy(None)
+
+
+class PolicyWithoutInitKwargs(Policy):
+    def __init__(self, *args):
```
wochinge, line 1237:
```suggestion
    def __init__(self, *args: Any) -> None:
```

tests/core/test_policies.py
```diff
@@ -1229,3 +1231,34 @@ def test_get_training_trackers_for_policy(
...
+
+    def persist(self, _) -> None:
+        pass
+
+    @classmethod
+    def _metadata_filename(cls):
+        return "no_finetune_policy"
+
+
+def test_loading_policy_with_no_constructor_kwargs(tmp_path: Path):
```
wochinge, line 1248: Should we move these two tests closer to the similar tests for `load`?

tests/test_train.py
```diff
@@ -518,15 +572,13 @@ def test_model_finetuning_nlu_with_default_epochs(
     assert new_diet_metadata[EPOCHS] == DIETClassifier.defaults[EPOCHS] * 0.5
 
 
-@pytest.mark.parametrize("model_to_fine_tune", ["invalid-path-to-model", "."])
```
wochinge, line 521: mhm, can we still test the case that we have a directory without model files, please?

wochinge (commented):
🚀

tests/core/test_ensemble.py
joejuzl: We should enforce typing in the CI

rasa/core/policies/policy.py
joejuzl: Are we sure that every `kwarg` given to `load` should be forwarded to `__init__`? 
I just worried that we could end up passing stuff to the policy constructor which we didn't want to, and which ends up being automatically incorporated into the policy config (like in the TED policy).

rasa/core/policies/sklearn_policy.py
joejuzl: We are purposefully not passing `should_finetune` in as a kwarg if it is not `True` so as not to break old policies.

tests/core/test_policies.py
joejuzl: The other load tests are implemented on the test class so that it runs for each of the existing policies. Whereas these are a little different as they test the behaviour when using custom policy classes.

tests/test_train.py
joejuzl: I thought it was no longer valid because of removing the `try` but actually it was just breaking because I had an empty model in `.`. So I will add them back.

dakshvar22:
@joejuzl Can you add the required changes to custom policies in the migration guide as part of this PR itself?

tests/core/test_policies.py
wochinge: ok

rasa/core/policies/sklearn_policy.py
wochinge: Mhm, but that's the specific `load` impl of the `SklearnPolicy`. I don't see what would break here?

rasa/core/policies/policy.py
wochinge: I'd argue that if somebody has a `load` function which uses `context` which isn't used by the `constructur` they should 1) `pop` it and 2) they would define the constructor call manually anyway. What do you think?

wochinge:
I think a constant for `should_finetune` would be great considering the many locations where we use this. Can in my opinion also be a separate PR then.

joejuzl:
@dakshvar22 I think a separate PR for the migration guide would make more sense so we can start using this code easier in other PRs.

rasa/core/policies/sklearn_policy.py
joejuzl: oh you mean making "should_finetune" a kwarg in the method definition?

wochinge: Yes

rasa/core/policies/policy.py
joejuzl: Yeah that seems reasonable.

PR #7479, opened 2020-12-08
updated dependencies

tmbo:
**Proposed changes**:
- checked changelogs and updated dependencies
- Socket IO has a version breaking change - but this is more something for the user to be aware of, as the version they should choose for `python-socketio` depends on the version they are choosing for the socketio clients they are using


joejuzl:
I think the `pytest` update is causing errors in the CI.

joejuzl (approved):

PR #7483, opened 2020-12-08
consider a story file as testfile if in test directory

wochinge:
**Proposed changes**:
- also accept a story file as test file if it's the default tests directory


kalkbrennerei:
Hi @wochinge! I think there was a specific reason why we decided against the folder structure when we changed the format to yaml? If I remember correctly we wanted to accept test stories only with the `test_` prefix.

Issue #7487, opened 2020-12-08
Add Team ID to Metadata for Slack Channel


kearnsw:
**Description of Problem**: The current slack channel does not extract the team id from the slack event into the UserMessage metadata.

First, it is currently not possible to know from which workspace a request is coming. This is an issue when it is required by any custom code in the action server.

Second, I think this also raises a larger issue that the sender_id as currently defined may not be unique across multiple workspaces. As discussed in [this forum post](https://forum.rasa.com/t/how-to-pass-metadata-in-post-request-and-then-extract-it/15092/2), we may need to update the sender_id to be a concatenation of the user_id and the team_id.

**Overview of the Solution**: The initial pull request will address the first issue of adding the team_id to the metadata using the get_metadata function of the slack channel. **The solution to the second issue of the uniqueness of the sender_id requires further discussion.**

sara-tagger:
Thanks for submitting this feature request 🚀 @joejuzl will get back to you about it soon! ✨

PR #7488, opened 2020-12-08
Feat #7487: Extract team id from slack event metadata and add to user…

kearnsw:
Changes are described in issue #7487.

- [N/A] updated the documentation


sara-tagger:

erohmensing (commented):
Feature makes sense to me, however these specific tests aren't passing -- I think you might need to add team to each of the responses? Not exactly sure. 

@b-quachtran as I'm going offline do you mind helping Will out?

kearnsw:
Thanks, Ella. That's odd, the tests passed on my local machine. I do see they failed on the requests that did not have a team field, but they only failed at the point which the metadata was referenced on the UserMessage. This means the get_metadata function successfully returned None for team_id. Are None valued metadata fields filtered out before being added to the UserMessage?

tests/core/channels/test_slack.py
```diff
@@ -143,19 +145,21 @@ def test_slack_form_metadata():
     assert metadata["out_channel"] == channel
     assert metadata["users"][0] == authed_user
     assert metadata["thread_id"] == ts
+    assert metadata["team_id"] is None
```
b-quachtran, line 148: This appears to be one of the failing cases, line 431 in slack.py needs to be updated to return the "team_id" parameter

b-quachtran (changes requested):

tests/core/channels/test_slack.py
kearnsw: Right, I'm not clear though on why this is the case. Can you check my comment above? I wonder if there is an issue deeper in the code base that doesn't allow None values to be added to the metadata. If that's the intended behavior, I can just remove that line from the test case. @b-quachtran

b-quachtran: The assert statement is checking that the "team_id" field exists in the metadata and that the value for that field is `None`. It's failing because this response does not include "team_id" in the JSON response as a field: 

```
        if content_type == "application/x-www-form-urlencoded":
            # if URL-encoded message is received
            output = request.form
            payload = json.loads(output["payload"][0])
            message = payload.get("message", {})
            thread_id = message.get("thread_ts", message.get("ts"))

            users = []
            if payload.get("user", {}).get("id"):
                users.append(payload.get("user", {}).get("id"))

            return {
                "out_channel": payload.get("channel", {}).get("id"),
                "thread_id": thread_id,
                "users": users,
            }
```

You'll need to update it to:
```
            return {
                "out_channel": payload.get("channel", {}).get("id"),
                "thread_id": thread_id,
                "users": users,
                "team_id": message.get("team_id"),
            }
```

kearnsw: Ah, thank you. Looks like I overlooked the extraction in the case it has the "application/x-www-form-urlencoded" header. I checked the Slack API reference documentation here: https://api.slack.com/reference/interaction-payloads/block-actions and updated the form object accordingly to include this piece of metadata. It's slightly different than the above.

kearnsw:
Looks like the error on Windows Python 3.8 was not a result of this PR.

b-quachtran (approved):
🚀 Awesome! Thanks @kearnsw for this Slack improvement

PR #7495, opened 2020-12-09
Deprecate Markdown Training & Test Data

wochinge:
**Proposed changes**:
- deprecate Markdown Training Data (reading and writing)
 - Story reading / writing
 - NLU reading writing
 - Response Selector reading / writing
- skip printing deprecation warnings when converting training data

### DIFF
--- a/tests/core/training/converters/test_story_markdown_to_yaml_converter.py
+++ b/tests/core/training/converters/test_story_markdown_to_yaml_converter.py
@@ -32,11 +32,11 @@ def test_converter_filters_correct_files(training_data_file: Text, should_filter
     )
 
 
-async def test_stories_are_converted(tmpdir: Path):
-    converted_data_folder = tmpdir / "converted_data"
+async def test_stories_are_converted(tmp_path: Path):
+    converted_data_folder = tmp_path / "converted_data"
     os.mkdir(converted_data_folder)
 
-    training_data_folder = tmpdir / "data/core"
+    training_data_folder = tmp_path / "data/core"
     os.makedirs(training_data_folder, exist_ok=True)
     training_data_file = Path(training_data_folder / "stories.md")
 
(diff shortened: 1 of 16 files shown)


tests/core/training/converters/test_story_markdown_to_yaml_converter.py
```diff
@@ -32,11 +32,11 @@ def test_converter_filters_correct_files(training_data_file: Text, should_filter
     )
 
 
-async def test_stories_are_converted(tmpdir: Path):
-    converted_data_folder = tmpdir / "converted_data"
+async def test_stories_are_converted(tmp_path: Path):
```
wochinge, line 35: replaced `tmpdir` as `tmpdir` is not a `Path` but of type `py.path.local `

wochinge:
@edmacovaz @degiz Re Rasa X:

This change will lead to warnings in the terminal when using Rasa X with Markdown format. I added switches so that it's possible to suppress them, but the question is whether this actually needed. What do you think? 
I think the more important `TODO` item would be to have a warning for the future deprecation in the Rasa X UI.

degiz (commented):
Looks great, thanks a lot for working on that! 🚀 🚀 🚀 
Not putting `Approve` because of the docs, I'm not sure if the deprecation is communicated well enough.

> I added switches so that it's possible to suppress them, but the question is whether this actually needed

I think yes. The way it's used in this PR makes sense to me, because we def. don't want these deprecation warnings when `rasa data convert` is being used.

> future deprecation in the Rasa X UI

This will be more tricky for sure. The closest thing we have now are the warning for the inconsistencies in the `domain`, and that's a separate endpoint. Probably will have to do something similar with the training data.

wochinge:
> because we def. don't want these deprecation warnings when rasa data convert is being used.

Sorry, I meant if we should use the switches when reading data within Rasa X (e.g. for Integrated Version Control). The deprecation warnings would "only" show up in the server logs which means that only developer can see them. If we use the switches in Rasa X we also have to consider that users might use Rasa X with older Rasa Open Source version where the readers / writer don't have the switch.

erohmensing:
I think the docs look okay, @degiz what would you add to clarify further?

degiz:
@erohmensing I don't have anything specific on mind, just wanted anyone from CSE to confirm 👍

degiz (approved):

wochinge:
@edmacovaz what do you think about the Rasa X implications?

wochinge:
@edmacovaz ?

edmacovaz:
@wochinge Sorry, I missed that one. I'll try to find some time to think about this once the squad organisation stuff has settled a little. I guess this still falls with you and the Enable team (where I think there's an open question about where you get your design help from). Does that work for you?

wochinge:
Mhm, could also be the CDD squad 🤔 Works for me 👍🏻

PR #7502, opened 2020-12-09
updated build backend

tmbo:
**Proposed changes**:
- updated pep-517 build backend according to https://python-poetry.org/blog/announcing-poetry-1-1-0.html
- change is not strictly necessary, but the old package name is deprecated since poetry moved their pip builder to a separate package. so sooner or later it is necessary


tmbo:
🛑   this can't be merged yet, build system requires the poetry version which isn't mentioned anywhere after switching the builder...https://github.com/python-poetry/poetry/issues/3469

joejuzl (approved):

scripts/poetry-version.sh
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env python
 
-"""Extract the poetry version from pyproject.toml. Used in e.g. github workflows."""
+"""Extract the poetry version from .github/poetry_version.txt. Used in e.g. github workflows."""
 
+import pathlib
```
tczekajlo, line 5: Does it require python 3?

tczekajlo: This should help

```
  docker:
    name: Build Docker
    runs-on: ubuntu-20.04
```

joejuzl: in `continous-integration.yml`?

tczekajlo: Yes :)

tczekajlo (approved):
LGTM 🎖️

PR #7503, opened 2020-12-09
update user messages in apply_to of define events

Ghostvv:
**Proposed changes**:
- use `apply_to` methods instead of `tracker.applied_events`

### DIFF
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -512,8 +512,9 @@ class TEDPolicy(Policy):
         domain: Domain,
         interpreter: NaturalLanguageInterpreter,
     ) -> List[List[Dict[Text, List["Features"]]]]:
-        # Construct two examples in the batch to be fed to the model - 
-        # One by featurizing last user text and second an optional one(see conditions below).
+        # construct two examples in the batch to be fed to the model -
+        # one by featurizing last user text
+        # and second - an optional one (see conditions below),
         # the first example in the constructed batch either does not contain user input
         # or uses intent or text based on whether TED is e2e only.
         tracker_state_features = self.featurizer.create_state_features(
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -526,7 +526,7 @@ class TEDPolicy(Policy):
 
     def _pick_confidence(
         self, confidences: np.ndarray, similarities: np.ndarray
-    ) -> Tuple[np.ndarray, bool]:
+    ) -> Tuple[int, bool]:
         # the confidences and similarities have shape (batch-size x number of actions)
         # batch-size can only be 1 or 2;
         # in the case batch-size==2, the first example contain user intent as features,
@@ -545,12 +545,12 @@ class TEDPolicy(Policy):
                 # TODO maybe compare confidences is better
                 and np.max(similarities[1]) > np.max(similarities[0])
             ):
-                return confidences[1], True
+                return 1, True
 
-            return confidences[0], False
+            return 0, False
 
         # by default the first example in a batch is the one to use for prediction
-        return confidences[0], self.only_e2e
+        return 0, self.only_e2e
 
     def predict_action_probabilities(
         self,
@@ -578,13 +578,16 @@ class TEDPolicy(Policy):
         similarities = output["similarities"].numpy()[:, -1, :]
         confidences = output["action_scores"].numpy()[:, -1, :]
         # take correct prediction from batch
-        confidence, is_e2e_prediction = self._pick_confidence(confidences, similarities)
+        prediction_index, is_e2e_prediction = self._pick_confidence(
+            confidences, similarities
+        )
+        confidence = confidences[prediction_index]
 
         if self.config[LOSS_TYPE] == SOFTMAX and self.config[RANKING_LENGTH] > 0:
             confidence = train_utils.normalize(confidence, self.config[RANKING_LENGTH])
 
         optional_events = self._create_optional_event_for_entities(
-            output, interpreter, tracker
+            output, prediction_index, is_e2e_prediction, interpreter, tracker
         )
 
         return self._prediction(
@@ -596,12 +599,15 @@ class TEDPolicy(Policy):
     def _create_optional_event_for_entities(
         self,
         prediction_output: Dict[Text, tf.Tensor],
+        prediction_index: int,
+        is_e2e_prediction: bool,
         interpreter: NaturalLanguageInterpreter,
         tracker: DialogueStateTracker,
     ) -> Optional[List[Event]]:
-        if tracker.latest_action_name != ACTION_LISTEN_NAME:
-            # entities belong to the last user message
-            # a user message is always followed by action listen
+        if tracker.latest_action_name != ACTION_LISTEN_NAME or not is_e2e_prediction:
+            # entities belong only to the last user message
+            # and only if user text was used for prediction,
+            # a user message is always followed by the action listen
             return
 
         if not self.config[ENTITY_RECOGNITION]:
@@ -612,7 +618,7 @@ class TEDPolicy(Policy):
             predicted_tags,
             confidence_values,
         ) = rasa.utils.train_utils.entity_label_to_tags(
-            prediction_output, self._entity_tag_specs
+            prediction_output, self._entity_tag_specs, prediction_index=prediction_index
         )
 
         if ENTITY_ATTRIBUTE_TYPE not in predicted_tags:
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -602,11 +602,11 @@ class TEDPolicy(Policy):
         if tracker.latest_action_name != ACTION_LISTEN_NAME:
             # entities belong to the last user message
             # a user message is always followed by action listen
-            return None
+            return
 
         if not self.config[ENTITY_RECOGNITION]:
             # entity recognition is not turned on, no entities can be predicted
-            return None
+            return
 
         (
             predicted_tags,
@@ -617,7 +617,7 @@ class TEDPolicy(Policy):
 
         if ENTITY_ATTRIBUTE_TYPE not in predicted_tags:
             # no entities detected
-            return None
+            return
 
         # entities belong to the last message of the tracker
         # convert the predicted tags to actual entities
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -524,7 +524,7 @@ class TEDPolicy(Policy):
             )
         return tracker_state_features
 
-    def _pick_confidence(
+    def _pick_prediction_index(
         self, confidences: np.ndarray, similarities: np.ndarray
     ) -> Tuple[int, bool]:
         # the confidences and similarities have shape (batch-size x number of actions)
@@ -578,7 +578,7 @@ class TEDPolicy(Policy):
         similarities = output["similarities"].numpy()[:, -1, :]
         confidences = output["action_scores"].numpy()[:, -1, :]
         # take correct prediction from batch
-        prediction_index, is_e2e_prediction = self._pick_confidence(
+        prediction_index, is_e2e_prediction = self._pick_prediction_index(
             confidences, similarities
         )
         confidence = confidences[prediction_index]
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -524,9 +524,9 @@ class TEDPolicy(Policy):
             )
         return tracker_state_features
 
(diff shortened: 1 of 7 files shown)


rasa/core/policies/ted_policy.py
```diff
@@ -602,11 +602,11 @@ def _create_optional_event_for_entities(
         if tracker.latest_action_name != ACTION_LISTEN_NAME:
             # entities belong to the last user message
             # a user message is always followed by action listen
-            return None
+            return
```
Ghostvv, line 605: @tabergma in this method `prediction_output` will contain the batch of 2 inputs. I'd say we should add entities only if e2e prediction is chosen. Do we do that? Sorry, I missed it in your PR

tabergma: Inside TED I only check if `self.config[ENTITY_RECOGNITION]` is set to `True`. So no, we don't have that right now.

Ghostvv: hm, that needs to be fixed then, how does it extract entities from the batch of size 2?

tabergma: Not sure what you mean. If the model was trained with entity recognition, it will automatically predict entities on incoming user utterances. We extract any entities if they are present.

Ghostvv: yes, but at prediction time due to e2e, the output will contain affectively 2 user utterance. And we need to predict entities only if e2e won

Ghostvv: I updated it in this PR, could you please take a look?

rasa/core/policies/ted_policy.py
```diff
@@ -578,13 +578,16 @@ def predict_action_probabilities(
         similarities = output["similarities"].numpy()[:, -1, :]
         confidences = output["action_scores"].numpy()[:, -1, :]
         # take correct prediction from batch
-        confidence, is_e2e_prediction = self._pick_confidence(confidences, similarities)
+        prediction_index, is_e2e_prediction = self._pick_confidence(
```
tabergma, line 581: Should we rename `_pick_confidence` to `_pick_prediction_index`?

tabergma (approved):
🚀

rasa/shared/core/events.py
```diff
@@ -631,6 +632,24 @@ def as_dict(self) -> Dict[Text, Any]:
         d.update({USE_TEXT_FOR_FEATURIZATION: self.use_text_for_featurization})
         return d
 
+    def apply_to(self, tracker: "DialogueStateTracker") -> None:
```
wochinge, line 635: so clean 😻

rasa/shared/core/events.py
```diff
@@ -631,6 +632,24 @@ def as_dict(self) -> Dict[Text, Any]:
...
+            tracker: The current conversation state.
+        """
+        if tracker.latest_action_name != ACTION_LISTEN_NAME:
+            # featurization belong only to the last user message
+            # a user message is always followed by action listen
+            return
+
+        # update previous user message's featurization based on this event
+        if tracker.latest_message.use_text_for_featurization is None:
+            # only update it if it was not defined
```
wochinge, line 648: I think this kind of comments is not particularly. helpful as they just repeat the code. The question (and I'm actually having it 😆 ): Why do we need to safeguard against this and when could this actually happen that we try to set it when it was already set previously?

Ghostvv:
@tabergma I was wrong, sorry. Things are a bit more complicated 🙈: The batch dimension of entity prediction is not of batch size, but of number of last text inputs in the batch

rasa/shared/core/domain.py
```diff
@@ -930,10 +930,13 @@ def _get_user_sub_state(
 
         # filter entities based on intent config
         # sub_state will be transformed to frozenset therefore we need to
-        # convert the list to the tuple
+        # convert the set to the tuple
         # sub_state is transformed to frozenset because we will later hash it
         # for deduplication
-        entities = tuple(self._get_featurized_entities(latest_message))
+        entities = tuple(
+            self._get_featurized_entities(latest_message)
```
wochinge, line 937: Nearly there but not entirely. 

This still fails because `self.latest_message` is just a pointer to the event.
```python
def test_tobi():
    tracker = DialogueStateTracker.from_events(
        "Vova",
        evts=[
            ActionExecuted(ACTION_LISTEN_NAME),
            UserUttered("hi", intent={"name": "greet"}),
            DefinePrevUserUtteredFeaturization(True),
            DefinePrevUserUtteredEntities(
                entities=[{"entity": "entity1", "value": "value1"}]
            ),
        ],
    )

    user = list(tracker.events)[1]
    assert isinstance(user, UserUttered)

    # Fails
    assert not user.entities
```

wochinge (changes requested):

rasa/shared/core/domain.py
Ghostvv: it should fail

Ghostvv: we want user uttered to have entities

rasa/shared/core/events.py
Ghostvv: every time we apply the events from tracker to tracker, which we do multiple times

rasa/shared/core/domain.py
wochinge: but this means that when we store the tracker, we store the updated version of `UserUttered` which would change history.

rasa/shared/core/events.py
wochinge: Mhm, and applying it twice is bad because of performance or why?

Ghostvv: I don't know 😅

Ghostvv: actually, can we guarantee that `UserUttered` events will be persisted with all the updates? Then I can remove `Define` events from applied events

rasa/shared/core/domain.py
Ghostvv: that's fine

Ghostvv: featurizers take care of it

rasa/shared/core/events.py
Ghostvv: I removed safeguard, I have don't know whether it is possible that we would redefine featurization

rasa/shared/core/domain.py
wochinge: In my opinion I don't think it's fine. It means that the `TrackerStore` is no longer a 1:1 presentation of what happened. This break the events' property of immutability.

rasa/shared/core/events.py
wochinge: >  Then I can remove Define events from applied events

That's actually the thing I'm trying to get you to fix in [here](https://github.com/RasaHQ/rasa/pull/7503#discussion_r540236892)

rasa/shared/core/domain.py
Ghostvv: It depends when you write events to the tracker store. I think we already discussed it, and I thought you fixed it. We should only persist `UserUttered` event after policy prediction happened and all the policy events are applied

wochinge: There is no need to rely on a fragile thing such as the time of persistence. 

> We should only persist UserUttered event after policy prediction happened and all the policy events are applied

There is no need for that as we have the two additional / new events. We will just replay events and come back to the current state. When we discussed the timing of persistence, it was when we considered if we need the additional events after all.

Ghostvv: yeah, but we use them differently

Ghostvv: >  It means that the TrackerStore is no longer a 1:1 presentation of what happened.

which is good, since we effectively made `UserUttered` a dynamic event

Ghostvv:
@wochinge since we established that the problem exists regardless of this change, I'd like to merge this PR, because it contains other fixes

rasa/shared/core/domain.py
wochinge: This is good for the ML models, but it's not good for the system itself. Breaking the immutability property is opening the box of the pandora in my opinion.

rasa/shared/core/events.py
```diff
@@ -1404,7 +1439,7 @@ def __repr__(self) -> Text:
 
     def __str__(self) -> Text:
         """Returns event as human readable string."""
-        return self.action_name or self.action_text
+        return f"ActionExecuted({self.action_name or self.action_text})"
```
wochinge, line 1442: why? Pretty sure we are using `str` to make it a human readable version (e.g. in the story validator). I think this rather fits `__repr__`

wochinge (commented):
I left another comment on the `ActionExecuted` class which should be considered.
We can continue the discussion for the mutating the `UserUttered` event on the other branch.

rasa/shared/core/domain.py
Ghostvv: I saw the other comment, I see your point

rasa/shared/core/events.py
Ghostvv: just to make it consistent with the way we present `UserUttered`, explicitly having the name of event make it more human readable

Ghostvv: if you look at other events, literally all of them have their class names in `str`

wochinge: It makes it easier to debug (--> `__repr__`), not necessarily easier to read. E.g. in the story validator there is absolutely no reason to prefix this with `ActionExecuted`. I think it makes sense to prefix certain events with their class name (e.g. ones which usually don't show up in stories like `LoopInterrupted`, but I don't think that makes sense for actions (nor for `UserUttered`). 

If you absolutely want to change this back, you have to make sure that you don't introduce regression (e.g. in the way how `rasa data validate stories` warns about clashing actions).

Ghostvv: then `str` is wrongly used. It should be consistent along various events and independent of the event itself. I'll change it back in order not to postpone this PR

Ghostvv: ready to merge?

Ghostvv:
@wochinge do you have an idea how the changes in this PR break the tests that fail. I don't understand why they're dependent

wochinge (approved):

rasa/shared/core/trackers.py
```diff
@@ -461,46 +458,11 @@ def applied_events(self) -> List[Event]:
...
-                # update event's entities based on the future event
-                entities = self._define_user_entities(events_as_list[i + 1 :])
-                if entities is not None:
-                    for entity in entities:
-                        if entity not in event.entities:
-                            event.entities.append(entity)
-
-                applied_events.append(event)
-            else:
+            elif not isinstance(event, DefinePrevUserUttered):
```
Ghostvv, line 461: @wochinge for some reason this change breaks everything. If I don't include `Define...` events in `applied_events` sometimes they're not applied. I'll remove it from here. Anyhow, this should be redesigned

wochinge: Mhm, then we should investigate why this breaks everything. That can only mean that events are not applied properly to the tracker.

wochinge: Maybe the problem is that your `apply_to` implementations manipulate `last_user_message` and `last_action` instead of the events `tracker.events` 🤔 They should refer to the same things, but maybe some stuff is actually deepcopying it.

wochinge: ah, now I understand it! Yes, you definitely have to still add the event as applied events.

wochinge: If you don't add it, then the events are not applied in the new tracker (as it doesn't get them) and everything will break

Ghostvv: but... even if I add them, there is still problems during evaluation. I patched it in featurizers. But, basically pandora has already opened. Let's carefully redesign it after release

Ghostvv: > then the events are not applied in the new tracker

but the new tracker should already get new updated `UserUttered`

wochinge: > but the new tracker should already get new updated UserUttered

There is no need to. The new tracker will call `apply_to` on each new event

Ghostvv: but then we do it million of times, doesn't feel efficient for me

wochinge: We actually need to do this for all of the events (e.g. `restarted` needs to reset a lot of state). I think the problem is rather in the featurization

Ghostvv: what do you mean? in core featurizers? Theoretically they should always receive `UserUttered` events where their featurization is decided

wochinge: I'm not sure about this. Anyway, let's close this discussion and have it in comment of the large PR

PR #7504, opened 2020-12-09
Integrate the finetune fingerprint checks into the train commands.

joejuzl:
**Proposed changes**:
Integrate the finetune fingerprint checks into the train commands.

### DIFF
--- a/rasa/model.py
+++ b/rasa/model.py
@@ -50,6 +50,7 @@ FINGERPRINT_NLG_KEY = "nlg"
 FINGERPRINT_RASA_VERSION_KEY = "version"
 FINGERPRINT_STORIES_KEY = "stories"
 FINGERPRINT_NLU_DATA_KEY = "messages"
+FINGERPRINT_NLU_LABELS_KEY = "nlu_labels"
 FINGERPRINT_PROJECT = "project"
 FINGERPRINT_TRAINED_AT_KEY = "trained_at"
 
@@ -81,14 +82,6 @@ SECTION_NLU = Section(
     ],
 )
 SECTION_NLG = Section(name="NLG templates", relevant_keys=[FINGERPRINT_NLG_KEY])
-
-SECTION_FINE_TUNE = Section(
-    name="Fine-tune",
-    relevant_keys=[
-        FINGERPRINT_CONFIG_WITHOUT_EPOCHS_KEY,
-        FINGERPRINT_DOMAIN_WITHOUT_NLG_KEY,
-    ],
-)
 
 
 class FingerprintComparisonResult:
@@ -346,6 +339,7 @@ async def model_fingerprint(file_importer: "TrainingDataImporter") -> Fingerprin
         FINGERPRINT_NLG_KEY: rasa.shared.utils.io.deep_container_fingerprint(responses),
         FINGERPRINT_PROJECT: project_fingerprint(),
         FINGERPRINT_NLU_DATA_KEY: nlu_data.fingerprint(),
+        FINGERPRINT_NLU_LABELS_KEY: nlu_data.label_fingerprint(),
         FINGERPRINT_STORIES_KEY: stories.fingerprint(),
         FINGERPRINT_TRAINED_AT_KEY: time.time(),
         FINGERPRINT_RASA_VERSION_KEY: rasa.__version__,
@@ -376,7 +370,7 @@ def _get_fingerprint_of_config_without_epochs(
     copied_config = copy.deepcopy(config)
 
     for key in ["pipeline", "policies"]:
-        if key in copied_config:
+        if key in copied_config and copied_config[key]:
             for p in copied_config[key]:
                 if "epochs" in p:
                     del p["epochs"]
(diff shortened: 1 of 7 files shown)


rasa/model.py
```diff
@@ -376,7 +370,7 @@ def _get_fingerprint_of_config_without_epochs(
     copied_config = copy.deepcopy(config)
 
     for key in ["pipeline", "policies"]:
-        if key in copied_config:
+        if key in copied_config and copied_config[key]:
```
wochinge, line 373: ? 
```suggestion
        if copied_config.get(key):
```

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -79,6 +79,14 @@ def fingerprint(self) -> Text:
 
         return rasa.shared.utils.io.deep_container_fingerprint(relevant_attributes)
 
+    def label_fingerprint(self) -> Text:
+        """Fingerprint the labels in the training data.
```
wochinge, line 83:
```suggestion
        """Fingerprints the labels in the training data.
```

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -79,6 +79,14 @@ def fingerprint(self) -> Text:
...
+
+        Returns:
+            hex string as a fingerprint of the training data labels.
+        """
+        return rasa.shared.utils.io.deep_container_fingerprint(self.intents)
```
wochinge, line 88: In my opinion this should include entity names as well

rasa/train.py
```diff
@@ -491,19 +491,27 @@ async def _train_core_with_validated_data(
...
     finetuning_epoch_fraction: float = 1.0,
 ) -> Optional[Agent]:
     path_to_archive = model.get_model_for_finetuning(model_to_finetune)
     if not path_to_archive:
         return None
 
     with model.unpack_model(path_to_archive) as unpacked:
+        new_fingerprint = await model.model_fingerprint(file_importer)
+        old_fingerprint = model.fingerprint_from_path(unpacked)
+        if not model.can_finetune(old_fingerprint, new_fingerprint, core=True):
```
wochinge, line 506: technically we only need to do this when `train_path is None`, correct?

rasa/train.py
```diff
@@ -639,12 +647,14 @@ async def _train_nlu_with_validated_data(
         )
 
         if model_to_finetune:
-            model_to_finetune = _nlu_model_for_finetuning(
-                model_to_finetune, config, finetuning_epoch_fraction
+            model_to_finetune = await _nlu_model_for_finetuning(
+                model_to_finetune,
+                file_importer,
+                finetuning_epoch_fraction,
+                also_train_core=train_path is not None,
```
wochinge, line 654: let's add a comment why this works. I doubt that our future selves will remember how this plays together. Or even better sth like `called_from_combined_training = train_path is None`

rasa/train.py
```diff
@@ -491,19 +491,27 @@ async def _train_core_with_validated_data(
...
     path_to_archive = model.get_model_for_finetuning(model_to_finetune)
     if not path_to_archive:
         return None
 
     with model.unpack_model(path_to_archive) as unpacked:
+        new_fingerprint = await model.model_fingerprint(file_importer)
+        old_fingerprint = model.fingerprint_from_path(unpacked)
+        if not model.can_finetune(old_fingerprint, new_fingerprint, core=True):
+            rasa.shared.utils.cli.print_error_and_exit(
+                "NLU model can not be finetuned."
```
wochinge, line 508: optional: That might be a bit weird in case `Core` changes are actually the reason for this

rasa/train.py
```diff
@@ -683,20 +693,34 @@ async def _train_nlu_with_validated_data(
         return _train_path
 
 
-def _nlu_model_for_finetuning(
+async def _nlu_model_for_finetuning(
     model_to_finetune: Text,
-    new_config: Dict[Text, Any],
+    file_importer: TrainingDataImporter,
     finetuning_epoch_fraction: float = 1.0,
+    also_train_core: bool = False,
```
wochinge, line 700: ?
```suggestion
    trained_before_core_model: bool = False,
```

tests/test_train.py
```diff
@@ -480,12 +483,68 @@ def test_model_finetuning_core_with_default_epochs(
...
 
 
+def test_model_finetuning_core_new_domain_label(
+    tmp_path: Path, monkeypatch: MonkeyPatch, trained_moodbot_path: Text,
+):
+    mocked_core_training = AsyncMock()
+    monkeypatch.setattr(rasa.core, rasa.core.train.__name__, mocked_core_training)
+
+    (tmp_path / "models").mkdir()
+    output = str(tmp_path / "models")
```
wochinge, line 493: optional: just do the concatenation once

tests/test_train.py
```diff
@@ -480,12 +483,68 @@ def test_model_finetuning_core_with_default_epochs(
...
+
```
wochinge, line 494: ?
```suggestion
    # Simulate addition to training data
```

wochinge (commented):
💯

rasa/shared/nlu/training_data/training_data.py
dakshvar22: It should include `entities`, `entity_roles` and `entity_groups`.

rasa/train.py
joejuzl: Technically yes - although maybe it would be a bit confusing? Also maybe a bit brittle as it relies on the fact we check NLU first? I'm not sure WDYT?

rasa/train.py
joejuzl: this was a copy&paste error.

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -181,9 +196,14 @@ def intents(self) -> Set[Text]:
         """Returns the set of intents in the training data."""
         return {ex.get(INTENT) for ex in self.training_examples} - {None}
 
+    @lazy_property
+    def actions(self) -> Set[Text]:
```
dakshvar22, line 200: Can we rename it to `action_names`? Primarily because end-to-end is going to get it `action_texts` as well and then `actions` as a property will be ambiguous.

rasa/train.py
```diff
@@ -683,20 +693,37 @@ async def _train_nlu_with_validated_data(
...
+        new_fingerprint = await model.model_fingerprint(file_importer)
+        old_fingerprint = model.fingerprint_from_path(unpacked)
+        if not model.can_finetune(
+            old_fingerprint,
+            new_fingerprint,
+            nlu=True,
+            core=called_from_combined_training,
+        ):
+            rasa.shared.utils.cli.print_error_and_exit(
+                "NLU model can not be finetuned."
```
dakshvar22, line 718: Is it possible to point to the possible reason? I think we can definitely distinguish between the three cases right?

dakshvar22 (commented):
Awesome testing I think 💯 . Just a comment about distinguishing between different reasons why finetuning may not be possible.

rasa/train.py
joejuzl: Yes, would required a bit of refactoring - but definitely possible.

dakshvar22: If it's major refactoring, we could also create an issue and tackle it after the release.

joejuzl: Actually it will currently log it to `info` (in `did_section_fingerprint_change`). Maybe that is enough for now?

dakshvar22: Tried it out. The explanation is not very helpful but I'd say it's enough for now. Can you create an enhancement issue to make these logs more verbose so that the user understands exactly where the problem lies?

rasa/train.py
wochinge: Both things are fine for a first iteration in my opinion

joejuzl: Oops sorry forgot this was still outstanding.

wochinge: no worries. If it'd been more important, I would have made it clearer

Issue #7507, opened 2020-12-10
[Windows unit-test] Windows installation times out after 5 hours


m-vdb:
It happened [here](https://github.com/RasaHQ/rasa/pull/7268/checks?check_run_id=1525103814) and [here](https://github.com/RasaHQ/rasa/pull/7471/checks?check_run_id=1525006234) and also [here](https://github.com/RasaHQ/rasa/pull/7284/checks?check_run_id=1529366468).

It times out at the step `poetry run python -m pip install -U pip`

cc @koernerfelicia @tczekajlo

tczekajlo:
Poetry since >= 1.1.0b switched to `vritualenv` (https://github.com/python-poetry/poetry/releases/tag/1.1.0b1). The default version of `vritualenv` used by poetry is 20.0.31, update virtualenv to the latest version helps only partially (I assume this fixed some things https://github.com/pypa/virtualenv/pull/1976), there is still a problem with python 3.6 though.

In the https://github.com/RasaHQ/rasa/pull/7542 I switched back to `venv` that works for Windows and Linux.

PR #7509, opened 2020-12-10
[Unit tests] Don't load Poetry cache for Windows

tczekajlo:
**Proposed changes**:
- Use the same key name to restore a cache
- Don't load Poetry cache for Windows
- Include poetry version in a cache key

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -228,7 +228,7 @@ jobs:
         poetry config virtualenvs.in-project true
 
     - name: Load Poetry Cached Libraries ⬇
-      if: needs.changes.outputs.backend == 'true'
+      if: needs.changes.outputs.backend == 'true' && matrix.os != 'windows-latest'
       uses: actions/cache@v1
       with:
         path: .venv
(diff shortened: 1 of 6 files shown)


tczekajlo:
@m-vdb Can you 👀  and ✅ ? It's a temporary fix for Windows - https://github.com/RasaHQ/rasa/issues/7507

.github/workflows/continous-integration.yml
```diff
@@ -228,12 +228,11 @@ jobs:
...
 
     - name: Load Poetry Cached Libraries ⬇
-      if: needs.changes.outputs.backend == 'true'
+      if: needs.changes.outputs.backend == 'true' && matrix.os != 'windows-latest'
       uses: actions/cache@v1
       with:
         path: .venv
-        key: ${{ runner.os }}-poetry-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
-        restore-keys: ${{ runner.os }}-poetry-${{ matrix.python-version }}
+        key: ${{ runner.os }}-poetry-${{ env.POETRY_VERSION }}-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
```
m-vdb, line 235: do we want to include the `POETRY_VERSION` for the cache key of other workflows?

m-vdb (approved):
🥥

.github/workflows/continous-integration.yml
tczekajlo: Hmm, yes! Thanks for pointing it out.

PR #7522, opened 2020-12-11
Add migration guide for policies

joejuzl:
**Proposed changes**:
Add migration steps for rasa 2.1 to 2.2 caused by continuous training

### DIFF
--- a/docs/docs/migration-guide.mdx
+++ b/docs/docs/migration-guide.mdx
@@ -10,6 +10,15 @@ description: |
 This page contains information about changes between major versions and
 how you can migrate from one version to another.
 
+## Rasa 2.1 to Rasa 2.2
+
+### Policies
+
+[Policies](./polcies.mdx) now require a `**kwargs` argument in their constructor and `load` method.
+Policies without `**kwargs` will be supported until Rasa version `3.0.0`.
+However when using [incremental training](./command-line-interface.mdx#incremental-training)
+`**kwargs` *must* be included.
+
 ## Rasa 2.0 to Rasa 2.1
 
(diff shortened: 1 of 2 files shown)


docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,15 @@ description: |
 This page contains information about changes between major versions and
 how you can migrate from one version to another.
 
+## Rasa 2.1 to Rasa 2.2
+
+### Policies
+
+[Policies](./policies.mdx) now require a `**kwargs` argument in their constructor and `load` method.
```
wochinge, line 17: we also have a changelog for this, right?

wochinge (approved):
🚀

docs/docs/migration-guide.mdx
joejuzl: hmm I don't think so. Will we have one changelog for the whole of continuous training?

wochinge: we should also add a changelog mentioning this change and then e.g. point to the migration guide.

> Will we have one changelog for the whole of continuous training?

I think this change is somewhat unrelated to the training (we needed it, but we could have also done the policy interface change on its own) so I think it deserves it's own changelog. Especially because it's a deprecation.

joejuzl: ok I can add it in this PR then. What issue number should we use?

wochinge: mhm, either the number of this PR or probably better the one from the PR which introduced the change.

wochinge (approved):

Issue #7525, opened 2020-12-11
Slack channel silently dropping long-running custom actions.


nathan-chappell:
**Rasa version**: 2.2.0a1

**Rasa SDK version** (if used & relevant): Not relevant

**Rasa X version** (if used & relevant): Not used

**Python version**: 3.8.6

**Operating system** (windows, osx, ...): Linux-4.15.0-47-generic-x86_64-with-glibc2.2.5

**Issue**: I have a custom action that takes some time to complete. It takes long enough that slack will close its connection with the rasa server, and send a retry request. When this happens, the action fails silently.

**Error (including full traceback)**: The only clue as to the error comes from opening up the debugger in the middle of the server, e.g. calling `breakpoint()` [here](https://github.com/RasaHQ/rasa/blob/28f3ad70b3006147150223b9f1c347a3a0c2e5e3/rasa/core/channels/slack.py#L525). Then, if you wait a moment, then try to step over the `return await self.process_message(...`, you will get a stack trace and a line which reads:

 asyncio.exceptions.CancelledError

**Command or request that led to error**: Described above.

**Content of configuration file (config.yml)** (if relevant): Not relevant - other than having a slack channel configured.

**Content of domain file (domain.yml)** (if relevant): Not relevant

**NOTE** I have a fix, and intend to discuss further issues not directly related to this in comments to my PR.

sara-tagger:
Thanks for the issue, @samsucik will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

wochinge:
@samsucik Could you validate this problem?

samsucik:
@wochinge in all honesty I'm quite busy and want to close some existing issues & projects first... Do you think that someone else could have a look?

wochinge:
Mhm, you can ask in the product-support channel but I guess everyone is busy atm.

melindaloubser1:
@nathan-chappell  re. 
>I have a fix, and intend to discuss further issues not directly related to this in comments to my PR.
what is your fix? Is it something that would be generally applicable, or was it only for a specific use case?

nathan-chappell:
[...]

PR #7526, opened 2020-12-11
Slack channel async fix

nathan-chappell:
**Proposed changes**:
- Handle messages on slack channel in `Task` as opposed to in `await` statement
- Catch `asyncio.CancelledError` along with `Exception`

This fixes the problem of silently dropped actions discussed in [issue #7525](https://github.com/RasaHQ/rasa/issues/7525).  First I'll describe how I found the bug, then I'll describe some problems with the fix.

## Bug description

My custom action takes a few seconds to complete.  In this time, Slack gets impatient and closes its connection, and sends a retry request.  I was frustrated, because the custom action would work fine in `rasa shell` as well as with the facebook connection.  Furthermore, the logs from the action server were showing that my custom action would continue to do what it should, the Rasa server seemed to be simply ignoring it.  If I were to debug the Rasa server at the right place and time, I could observe an `asyncio.CancelledError` seem to appear out of thin air.  I finally resorted to observing the TCP traffic to determine why it wasn't working for Slack.  I finally observed this behavior:

### Observed behavior

    Connection Slack <-> Rasa opened                  # user message
    Connection Rasa  <-> Action server opened    # custom action
    Connection Slack <-> Rasa closed                    # Slack impatient
    Connection Rasa  <-> Action server closed      # custom action dies... SILENTLY
    Connection Slack <-> Rasa opened                  # retry request
    # this connection stays open for a long time...
    Connection Slack <-> Rasa closed                    # dies of natural causes

The hypothesis was that the connection from Slack <-> Rasa closing was causing the connection from Rasa <-> Action server to be closed, and that the underlying connections (from either `asyncio` or `sanic`) were just throwing away whatever the results were hoped to be.  Upon digging into `Rasa core,` I was happy to find this all to be true.  The `await on_new_message(user_msg)` code was creating an unecessary dependecy on the connection Slack <-> Rasa for Rasa <-> Action server.  By getting the event loop and creating a task, instead of just awaiting the coroutine, an "independent" connection Rasa <-> Action server is created, and Rasa can respond to events as expected.

## Open questions

* Is this an `asyncio` issue?  Should one task being cancelled close connection for tasks that are running "underneath" it?  I think this is probably the only reasonably simple implementation, to propogate "cancel" events to tasks like this, but I don't know all the details.
* Is this a `sanic` issue?  I observed that a `FIN` being sent Slack -> Rasa would trigger a `FIN` Rasa -> Action server, but as far as I know a `FIN` doesn't require the entire connection being closed.  I don't know enough about the standards, protocols, or implementation to assert this is undesired behavior, but maybe this could be handled differently by `sanic`.
* Is there a Slack issue?  Slack gets impatient, and sends a `FIN` to Rasa, ignoring any `keep-alive` sent by Rasa (well, `sanic`).  I understand this, and it's (somewhat) documented, but then they will respect the `keep-alive` when they send there retry-request (this is the last line of the observed behavior described above)...
* Is there another Rasa issue?  Rasa "gracefully" handles these retry requests by logging the message to the user and responding with `None` [see here](https://github.com/RasaHQ/rasa/blob/28f3ad70b3006147150223b9f1c347a3a0c2e5e3/rasa/core/channels/slack.py#L356) (I believe that `sanic` prefers you to send an empty string rather than `None`...).  But this request should not have any `keep-alive` set.  It does not seem easy to indicate that a specific response should have not `keep-alive` header associated with it, this is determined by `sanic` at a lower level by referring to the parser, and sending the server's default `keep-alive` value.

## Problems with my fix

If the event loop should stop before the tasks put into it in my code finish, then `asyncio` will complain quite loudly.  This is because these tasks are left on their own, and are not taken care of appropriately (they should be `awaited` somewhere I guess...)  I personally think this is still a better option than killing everything due to a possible network issue.

## Better fix

My fix is pragmatic and simple, and can be applied to other connections if they have the same problem.  I've variously seen recommendations to work around this issue as "trigger an external intent."  I think that this fix is far more inline with the semantics of Rasa and what users expect the behavior to be.  It does seem that some new abstraction may benefit Rasa for managing such connections though (assuming this is not the only place this issue will ever occur).  For example, a type of action class `LongRunningAction` may be useful to explicitly indicate that the action may take longer than what is typical, and connection failures between Rasa and whatever other side of the connection should not cause the action to fail.  Maybe in some cases you do want the action to fail, and maybe in some cases you want a way to explicitly cancel some `LongRunningAction`.  In all these cases, a separate `RasaConnectionManager` type of service could be useful.  Instead of letting the underlying network determine the semantics of Rasa conversations, these events, and the desires of the user, could be taken into account in a principled way.  Of course, this is probably a pain in the ass to implement, and if it would not actually be beneficial outside of this singular case then it should not be done (I wouldn't be able to determine this...)

## Final note on asyncio.CancelledError

This makes me want to throw a wrench at somebody.  I have no idea why this "exception" does not derive from `Exception`, but rather `BaseException`.  It means that the standard python idiom of logging exceptions fails when `asyncio` cancels a task.  Even if you don't like shifting the action handling mechanism to as `Task` and letting it try to survive the cruel asynchronous world all on its own, dealing with the `CancelledError` exception explicitly is a must - that way the user will know the reason why Rasa is ignoring its action server.

- [*] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)


sara-tagger:
Thanks for submitting a pull request 🚀 @tmbo will take a look at it as soon as possible ✨

melindaloubser1:
https://github.com/RasaHQ/rasa/issues/7525

melindaloubser1:
@nathan-chappell Thanks for the detailed description! 

> dealing with the CancelledError exception explicitly is a must - that way the user will know the reason why Rasa is ignoring its action server.

Do you think it would be possible to do what you've mentioned here without turning to a `task`? 
Do you know when (under what circumstances) the event loop would stop before the task was completed?

nathan-chappell:
@melindaloubser1 I don't think that the issue here is avoidable without creating task as I've described...  The only other solution would be to deal with this issues at a lower level, like asyncio or perhaps sanic...  Even if you don't use a task, you'll need some sort of abstraction that says "run this function at a later time, independently of the one currently running."
I'm pretty sure the only way that the task loop would stop before the task is completed is by the user killing the process under which sanic is running.  I'm not positive, but I'd guess that when sanic is interrupted it tries to finish all tasks it has pending before closing the loop, but by repeated pressing ctrl-c or sending a kill signal you'll probably bypass that...

melindaloubser1:
@wochinge wdyt of the proposed solution?

rasa/core/channels/slack.py
```diff
@@ -375,8 +377,8 @@ async def process_message(
                 metadata=metadata,
             )
 
-            await on_new_message(user_msg)
-        except Exception as e:
+            asyncio.get_event_loop().create_task(on_new_message(user_msg))
+        except (Exception, asyncio.CancelledError) as e:
```
wochinge, line 381: > asyncio.CancelledError

Can this error even happen as the `try .. except` block will already be left when this happens?

wochinge:
@melindaloubser1 I like the solution. @nathan-chappell Very detailed and helpful description of a complex problem 🚀 

Some ideas around this:
* I think it would be cleaner if we pass in the Sanic loop to the channels explicitly instead of having to refer to `get_event_loop`. This is related to https://github.com/RasaHQ/rasa/issues/7789
* This change makes sense for a couple of other channels as well (basically every channel which we create a new connection to send responses instead of returning the response with the request)

nathan-chappell:
@wochinge Thanks.  I 100% agree that passing in Sanic's event loop is the right move (I don't know much about Sanic).  It looks like #7789 also had the idea to create a higher-level connection management system.  Sounds like a bit of work, but it could take care of all similar problems in a principled way.

wochinge:
@nathan-chappell Would you be up for adapting the PR with a more general solution? We can also iterate on it together, e.g. you start by making some comments where you'd change stuff and then I give feedback, and so on?

wochinge:
@nathan-chappell Please feel free to re-open if you have time to continue

PR #7529, opened 2020-12-11
Default slots

wochinge:
**Proposed changes**:
- add default slots upon domain creation to ensure fingerprints are created correctly (previously there was problem because we compared a domain without default slots with a domain which already contained the default slots - the test added to `test_model` visualizes this)
- deprecate `add_categorical_slot_default_value`, `add_requested_slot`, `add_knowledge_base_slots`. Their internal versions are automatically called when the domain is created
- don't persist the default value for categorical slots. We add it when the domain is loaded.


data/test_config/stack_config.yml
```diff
@@ -8,3 +8,4 @@ pipeline:
 # https://rasa.com/docs/rasa/policies
 policies:
   - name: MemoizationPolicy
+  - name: RulePolicy
```
wochinge, line 11: this was needed to work with the form which I added to the default domain.

wochinge: the presence of the `RulePolicy` in turn led to fallback, which is why I had to add the new story

rasa/shared/core/slots.py
```diff
@@ -298,8 +298,15 @@ def add_default_value(self) -> None:
...
     def persistence_info(self) -> Dict[Text, Any]:
+        """Returns serialized slot."""
         d = super().persistence_info()
-        d["values"] = self.values
+        d["values"] = [
+            value
+            for value in self.values
+            # Don't add default slot when persisting it.
+            # We'll re-add it on the fly when creating the domain.
+            if value != rasa.shared.core.constants.DEFAULT_CATEGORICAL_SLOT_VALUE
```
joejuzl, line 308: Isn't this handed by the change in `_slot_definitions`?

tests/core/test_model.py
```diff
@@ -347,6 +348,33 @@ def test_should_retrain(
...
+    story_file.write_text(
+        """
+stories:
+- story: test_story
+  steps:
+  - intent: greet
+  - action: utter_greet
+    """
+    )
+    trained_model = await train_core_async(
```
joejuzl, line 363: I wonder if this test should be in `test_train` as it calls the train methods?

tests/core/test_model.py
```diff
@@ -424,7 +452,7 @@ async def test_update_with_new_domain_preserves_domain(
     tmpdir: Path, domain_with_categorical_slot_path
 ):
     domain = Domain.load(domain_with_categorical_slot_path)
-    domain.setup_slots()
+    domain._add_default_slots()
```
joejuzl, line 455: doesn't this happen in the `__init__` now?

tests/shared/core/test_domain.py
```diff
@@ -886,7 +913,7 @@ def test_add_knowledge_base_slots(default_domain: Domain):
     assert SLOT_LAST_OBJECT not in slot_names
     assert SLOT_LAST_OBJECT_TYPE not in slot_names
 
-    test_domain.add_knowledge_base_slots()
+    test_domain._add_knowledge_base_slots()
```
joejuzl, line 916: Again, does this not happen in the constructor now?

joejuzl (changes requested):

tests/core/test_model.py
wochinge: Will move it! I came for a fingerprinting perspective that's why I added it here

wochinge: Actually I'll keep it there. It tests `should_retrain` which is part of `model`. the training is just preparing the test data.

rasa/shared/core/slots.py
wochinge: No. The `default_slot_value` is a **value** and not an entire slot. Means we can't exclude the `entire` slot, however, we can exclude the `default` and then re-add it later.

joejuzl (approved):

joejuzl:
Looks good! 🚀

wochinge:
@joejuzl I suspect merge conflicts when merging 2.1.x to `master`. How about I handle the `2.1.x` to `master` merge (will request your review) which you'd have done otherwise for the release?

joejuzl:
@wochinge that would be great - thank you

joejuzl (approved):

PR #7535, opened 2020-12-13
add example showing e2e functionality

amn41:
**Proposed changes**:
- updates the example/e2ebot to show how e2e works and motivate why you would want it. 

NB I had to finesse the `config.yml` to get it to work as intended (basically just tweaking the number of epochs for DIET and TED so that e2e would kick in when I wanted it to). We definitely need a better way for users to do this. 


examples/e2ebot/config.yml
```diff
@@ -10,7 +10,8 @@ pipeline:
     min_ngram: 1
     max_ngram: 4
   - name: DIETClassifier
-    epochs: 200
+    epochs: 10
```
Ghostvv, line 13: let's keep `200`. Otherwise, our losses doesn't have a chance to train

examples/e2ebot/config.yml
```diff
@@ -10,7 +10,8 @@ pipeline:
     min_ngram: 1
     max_ngram: 4
   - name: DIETClassifier
-    epochs: 200
+    epochs: 10
 policies:
+- name: MemoizationPolicy
 - name: TEDPolicy
-  epochs: 200
+  epochs: 50
```
Ghostvv, line 17: let's keep `200`. Otherwise, our losses doesn't have a chance to train

examples/e2ebot/config.yml
```diff
@@ -10,7 +10,8 @@ pipeline:
     min_ngram: 1
     max_ngram: 4
   - name: DIETClassifier
-    epochs: 200
+    epochs: 10
 policies:
+- name: MemoizationPolicy
```
Ghostvv, line 15: let's not use it, if you want some defined behavior, please use rules

examples/e2ebot/config.yml
amn41: thanks - can confirm that with this additional NLU example, still works if I turn off memo and increase epochs back to 200

examples/e2ebot/config.yml
```diff
@@ -13,4 +13,4 @@ pipeline:
     epochs: 200
 policies:
 - name: TEDPolicy
-  epochs: 200
+  epochs: 200
```
Ghostvv, line 16: not sure what are the guidelines, but all other `yaml` files have blank line in the end
```suggestion
  epochs: 200

```

examples/e2ebot/data/nlu.yml
```diff
@@ -61,44 +36,13 @@ nlu:
...
-    - are you a bot?
-    - are you a human?
-    - am I talking to a bot?
-    - am I talking to a human?
+    - I'm looking for some food
+    - show me a place to eat
+    - where should I eat tonight?
+    - list restaurants
+    - food
+    - I'm hungry
```
Ghostvv, line 48:
```suggestion
    - I'm hungry

```

examples/e2ebot/domain.yml
```diff
@@ -1,14 +1,23 @@
...
+responses:
+  utter_greet:
+  - text: "hi!"
+  utter_suggest_cuisine:
+  - text: "how about greek food?"
+  - text: "how about chinese food?"
+  - text: "how about italian food?"
+  - text: "how about sushi?"
+  utter_search_restaurants:
+  - text: "great! here's what I found ..."
```
Ghostvv, line 23:
```suggestion
  - text: "great! here's what I found ..."

```

examples/e2ebot/data/stories.yml
```diff
@@ -1,30 +1,40 @@
...
+  - action: utter_suggest_cuisine
+  - user: "I can always go for sushi"
+  - action: utter_search_restaurants
+
+- story: rejects suggestion, indirectly
+  steps:
+  - intent: search_restaurant
+  - action: utter_suggest_cuisine
+  - user: "I had that yesterday"
+  - action: utter_suggest_cuisine
```
Ghostvv, line 40:
```suggestion
```

Ghostvv (approved):
great, thank you for updating it

Ghostvv (changes requested):
Sorry forgot. Could you please add stories where bot also answers smth directly in the story

examples/e2ebot/data/stories.yml
```diff
@@ -1,30 +1,38 @@
...
   - intent: greet
   - action: utter_greet
-  - user: "I'm not happy"
-  - action: utter_cheer_up
-  - action: utter_did_that_help
-  - user: "Not at all"
-  - bot: "Goodbye!"
-
+  - intent: search_restaurant
+  - bot: "how about sushi?"
```
Ghostvv, line 29: this would be a contradicting story to `accepts suggestion` and also "how about sushi?" is already part of the template, I'm afraid this example introduces more confusion on to when use e2e bot utterances

Ghostvv (changes requested):

examples/e2ebot/data/stories.yml
amn41: 1. why is it contradicting? it's the same story, except that the user says 'affirm' in an usual way
2. I included this because it highlights how the exact text of the bot response can also influence the meaning of the user response (e.g. if the user had said "I can always go for Italian", then they would be saying something different). 

I am totally fine with removing this, though. I think for now the main use case for full end-to-end is to train models on raw transcripts. I don't see an obvious reason for just replacing action names with bot text in your stories.

Ghostvv: > why is it contradicting?
you have the following steps:
```
- intent: greet
- action: utter_greet
 - intent: search_restaurant
```
in both stories and then either `utter_suggest_cuisine` or `"how about sushi?"` the bot would know which one to predict

Ghostvv: > I think for now the main use case for full end-to-end is to train models on raw transcripts. I don't see an obvious reason for just replacing action names with bot text in your stories.

I agree. But I thought in this example bot, we could show the raw functionality

Ghostvv: > I included this because it highlights how the exact text of the bot response can also influence the meaning of the user response

If I understood correctly you wanted rather to condition on bot utterance randomly picked from template response, this wouldn't work. We treat `utter_suggest_cuisine` and `"how about sushi?"` as two separate actions

amn41: ah yes, I forgot that adding the bot utterance in this story would treat "how about sushi" as a new action rather than as a particular instance of an existing action. Something we can keep in mind for the future

b-quachtran (changes requested):
Can an example of an entirely e2e training story be added? I think it'd be helpful to showcase that in the bot.

amn41:
I don't think we should provide an example of a fully end-to-end story without a clear use case of why you would want to use that. 
Fully end-to-end training is a useful thing, but I don't see why you would have full e2e stories in your training data for a bot you're developing.

Ghostvv:
I agree with @b-quachtran I'd keep this bot simply as example of functionality. We don't recommend using e2e in production, but if we want users to experiment with it and give us as much feedback as possible, I'd showcase here everything. Maybe make a disclaimer that this example doesn't represent best practices

amn41:
I think it's a good idea to explain to users that full e2e exists, but that should be in the docs somewhere, not in the example bot. 
The use case is a different one and having a random single e2e story in the example bot is more confusing than helpful

Ghostvv:
I don't see how it is confusing. The whole bot is artificial anyhow. @akelad what do you think

amn41:
Agree that it's artificial, but the pattern "replace intents with the actual text in stories" is one that users can try out for themselves. We can explain the benefit of using that pattern in the context of building a bot with CDD (see blog post).

For replacing the action label with the bot utterance, I don't know how users can take that pattern and apply it to their own use cases. Except in the case of training a fully e2e model on human-human transcripts, but that's a different example from an example bot.

akelad (approved):
I'm here to resolve the discussion 💁🏻‍♀️ I agree that the example in it's current state makes sense, given that this is our current best guess of how people should use e2e. we can always add full e2e examples later once we know more about the feature

PR #7539, opened 2020-12-14
Commented the old loading of data and trying to use TrainingDataImporter

Imod7:
**Proposed changes**:
- ...


sara-tagger:
Thanks for opening a draft pull request 🚀If you have any questions, you can direct them to @dakshvar22 ✨

rasa/test.py
```diff
@@ -151,10 +152,11 @@ def test_core(
         additional_arguments, test, ["stories", "agent"]
     )
 
-    rasa.utils.common.run_in_loop(test(stories, _agent, out_directory=output, **kwargs))
+    # rasa.utils.common.run_in_loop(test(stories, _agent, out_directory=output, **kwargs))
```
wochinge, line 155: Good catch! The error is raised as this line (currently commented out) creates a second loop although the one which you create in `rasa.cli.test` is already running. So having `test_core` being `async` is actually a very good idea 👍🏻

rasa/cli/test.py
```diff
@@ -167,11 +168,17 @@ def run_nlu_test(args: argparse.Namespace) -> None:
...
-    run_core_test(args)
-    run_nlu_test(args)
+    await run_core_test(args)
+    await run_nlu_test(args)
+
+
+def test(
+    args: argparse.Namespace, loop: Optional[asyncio.AbstractEventLoop] = None,
+):
+    return rasa.utils.common.run_in_loop(test_async(args), loop,)
```
wochinge, line 184: as discussed in our call I'd move `run_in_loop` further down to something like a `test_nlu_async` and `test_perform_nlu_cross_validation_async`. 

This gives us the benefit of reusable `async` functions. Functions with the `argparse` arguments are not really reusable as only `argparse` passes the arguments within the `argparse.Namespace` object. This should also help you to get rid of the error which you had [here](https://github.com/RasaHQ/rasa/pull/7539/files#r545251989)

wochinge: I'd imagine something like 

test -> run_nlu_test ->  -> `test_nlu` (and keep that async) . Then we actually don't even need something like a `test_nlu_async`

wochinge: You would probably do the `run_in_loop` [here](https://github.com/RasaHQ/rasa/blob/b00ab8e1b0c048811b6a4ebee5b589161b1cd72c/rasa/cli/test.py#L171) then 🤔

rasa/test.py
```diff
@@ -122,6 +122,7 @@ def test_core(
 
     try:
         unpacked_model = rasa.model.get_model(model)
+        unpacked_model = rasa.model.get_model(model)
```
wochinge, line 125: btw, this line is duplicate

rasa/cli/test.py
```diff
@@ -167,7 +168,10 @@ def run_nlu_test(args: argparse.Namespace) -> None:
             args.model, "model", DEFAULT_MODELS_PATH
         )
 
-        test_nlu(model_path, nlu_data, output, vars(args))
+        # test_nlu(model_path, nlu_data, output, vars(args))
+        rasa.utils.common.run_in_loop(
```
wochinge, line 172: it seems the import for `import rasa.utils.common` is missing.

rasa/test.py
```diff
@@ -154,7 +155,7 @@ def test_core(
     rasa.utils.common.run_in_loop(test(stories, _agent, out_directory=output, **kwargs))
 
 
-def test_nlu(
+async def test_nlu(
```
wochinge, line 158: this function is also called from [here](https://github.com/RasaHQ/rasa/blob/bc76915a8fbd40f505f75c8246f03f26caaa0323/rasa/test.py#L104). Seems we need another `rasa.utils.common.run_in_loop` here

rasa/nlu/test.py
```diff
@@ -1909,7 +1913,7 @@ def compare_nlu(
                 model_path = os.path.join(get_model(model_path), "nlu")
 
                 output_path = os.path.join(model_output_path, f"{model_name}_report")
-                result = run_evaluation(
+                result = run_evaluation_async(
```
wochinge, line 1916: `run_evaluation_async` doesn't exist. Did you mean this?
```suggestion
                result = run_evaluation(
```

rasa/nlu/test.py
```diff
@@ -1454,7 +1454,7 @@ def remove_pretrained_extractors(pipeline: List[Component]) -> List[Component]:
     return pipeline
 
 
-def run_evaluation(
+async def run_evaluation(
```
wochinge, line 1457: It's also called from [here](https://github.com/RasaHQ/rasa/blob/5c1e92d1eaaa005e2a6f56db37abcf3e77110237/rasa/server.py#L988). We need to add an `await` there so that our `run_evaluation` coroutine is actually executed

wochinge: Some tests (e.g. [this](https://github.com/RasaHQ/rasa/blob/0426f1558f49b019d402b082507ec0a0ee7a0ef3/tests/nlu/test_evaluation.py#L361)) also use it. I can highly recommend using Pycharms `Find Usages` to see where a function / method is used in the code.

rasa/nlu/test.py
```diff
@@ -1820,7 +1824,7 @@ def compute_metrics(
     )
 
 
-def compare_nlu(
+async def compare_nlu(
```
wochinge, line 1827: This is also called from [here](https://github.com/RasaHQ/rasa/blob/bc76915a8fbd40f505f75c8246f03f26caaa0323/rasa/test.py#L220). I suggest to also add a `rasa.utils.common.run_in_loop` there

wochinge (commented):
Nice, I think we're nearly there 🚀 

Re the failing tests: 
* Lines like [these](https://github.com/RasaHQ/rasa/runs/1576391757?check_suite_focus=true#step:11:649) suggest that we missed an `await` for one of the functions we changed to be `async`.
* [The Code Quality check](https://github.com/RasaHQ/rasa/runs/1576391557?check_suite_focus=true) fails as `run_evaluation_async` is not defined (I also commented on this)

wochinge:
@Imod7 Can we close this PR in favor of https://github.com/RasaHQ/rasa/pull/7674 ?

PR #6753, opened 2020-09-23
[bugfix][WIP] Fix the bug of Chinese tokenization in 2.0.0rc1

howl-anderson:
**Proposed changes**:
- Add a character-based tokenizer for Chinese and other languages not suit for whitespace tokenizer
- Conditionally use character-based tokenizer in HFTransformersNLP

fix #6752 

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] updated the documentation
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #7550, opened 2020-12-15
[WIP][Feature] HuggingFace transformers based tokenizer

howl-anderson:
**Proposed changes**:
- Add HuggingFace transformers based tokenizer

For Chinese and other whitespace non-splittable languages, incorrectly tokenized examples will cause problems: entity misaligned at training, incorrect at inference.

For example, for Chinese text "今天天气" (in English: "today's weather”, "今天" equals to "today", "天气" equals to "weather”), the Jieba tokenizer incorrectly split the text into tokens list ["今天天气"], the correct one should be ["今天", "天气"], the consequence of this mistake is that the entity (`[今天](date)天气`) misaligned and will be dropped.

By using transformers based tokenizer, there is no such mistake anymore.

**Before transformers in rasa upgrade to 3.x, this PR can not be merged.**


sara-tagger:
Thanks for submitting a pull request 🚀 @m-vdb will take a look at it as soon as possible ✨

m-vdb:
Hi @howl-anderson 👋🏻 thanks a lot for your contribution. Such feature could have implications for other languages and impact entity predictions, as mentioned by one of my colleagues [here](https://github.com/RasaHQ/rasa/pull/6753#issuecomment-727819280); so at the moment, we cannot merge this in. Would using a custom tokenizer class be an option for you?

howl-anderson:
Hi @m-vdb, yes, this feature is not intended to be merged at this moment (as I mentioned before: before transformers in rasa upgrade to 3.x, this PR can not be merged). This feature is an option for users, so it doesn't have implications for other languages or impact entity predictions if users don't use this tokenizer in their pipeline. As far as I know, this new tokenizer is the best solution to fix the misaligned entity issues (4 of 196 training examples got this error in my weather bot project, many Chinese rasa users complain such a thing to me) caused by poor performance tokenizers (such as Jieba for Chinese) for Chinese and other whitespace non-splittable languages. In fact, no tokenizer can segment words perfectly, because Chinese does not have a unified word segmentation standard, and there is inherent ambiguity. This is why BERT and recent DNN papers&applications use character-based embedding instead of word-based. What https://github.com/RasaHQ/rasa/pull/6753#issuecomment-727819280 worried, it indeed happens when the model is poorly trained, but it only influences the languages that using sub-word tokenization, for Chinese language (maybe also works for other whitespace non-splittable languages) there is no such issue.

dakshvar22:
@howl-anderson Thanks for the explanation. We really appreciate you putting in the effort to propose a new component which would be better for the Chinese rasa users and possibly other whitespace non-splittable languages too. All your experiments and feedback has made us rethink how we should handle these cases better. I would definitely like to have this inside Rasa Open Source but once we see more use cases/instances of tokenizers such as Jieba failing terribly for a large vocabulary set. Also the problem that https://github.com/RasaHQ/rasa/pull/6753#issuecomment-727819280 mentions can still happen for proper nouns in Chinese I believe since the the vocabulary of the pre-trained models can always be limiting. Until then, I would suggest having this component inside our [`nlu-examples repo`](https://github.com/RasaHQ/rasa-nlu-examples). Feel free to make a pull request there.
cc @koaning

howl-anderson:
Hi, @m-vdb @dakshvar22 thank you for your comments. I moved my HuggingFace transformers based tokenizer to repo https://github.com/howl-anderson/rasa_chinese which is a Chinese specific additional package of Rasa. When this feature proves to be very useful and important for Chinese users in the future, I will try to migrate it to the official repo.

Issue #7394, opened 2020-11-27
Implement train_in_chunks on classifiers


tabergma:
related to #6836 

Implement the method `train_chunk` on all classifiers and entity extractors. 

Make sure to raise an error when `train_chunk` on components is called that cannot handle training in chunks, e.g. `CRFEntityExtractor`. This happened already for most components, but should be double checked.

Most likely we also need to update the `RasaModel` interface for training as we don't want to have the train loop inside the fit method but rather outside as one `TrainingDataChunk` should correspond to one batch.

tabergma:
Implementation is done and can be found on `file-loader-data-generator`. NLU training works. However, the implementation depends on https://github.com/RasaHQ/rasa/pull/7438. Will open a PR once https://github.com/RasaHQ/rasa/pull/7438 is merged.

PR #7551, opened 2020-12-15
Train NLU pipeline using chunks

tabergma:
**Proposed changes**:
closes https://github.com/RasaHQ/rasa/issues/7394

- Add a new data generator that loads data from files
- Update the `train_chunk` interface for classifiers and entity extractors 
- Update the train loop, training NLU models in chunks is working


dakshvar22 (commented):
Looks good! ✨  Sorry for the delay in review.
Should we add a test which checks that components are trained in the correct order for different configuration of pipelines? For example, `train` for `DIET` is not called before featurizers are trained on all chunks.

rasa/train.py
```diff
@@ -48,7 +48,7 @@ def train(
     domain: Text,
     config: Text,
     training_files: Union[Text, List[Text]],
-    output_path: Text = DEFAULT_MODELS_PATH,
+    output: Text = DEFAULT_MODELS_PATH,
```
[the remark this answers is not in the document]
tabergma: Yeah, also recognized that and undid the change. I guess I forgot some parts.

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -820,7 +838,7 @@ def _to_tf_features(self, features: List[Features]) -> Dict[Text, tf.train.Featu
             if feature.is_dense():
                 tf_features[key] = self._bytes_feature(feature.features)
             else:
-                data = feature.features.data
+                data = feature.features.data.astype(np.int64)
```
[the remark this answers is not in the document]
tabergma: for sparse features our data are always integer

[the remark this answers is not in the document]
dakshvar22: Ahh right, I missed the if else construct above.

rasa/nlu/model.py
```diff
@@ -264,24 +266,42 @@ def train_in_chunks(
...
 
+        logger.info(f"Divide training data into {number_of_chunks} chunks.")
         training_data_chunks = working_training_data.divide_into_chunks(
             number_of_chunks
         )
+        data_chunk_files = []
 
         # perform featurization
         for i, data_chunk in enumerate(training_data_chunks):
+            logger.info(f"Starting to train all featurizers on chunk {i}.")
```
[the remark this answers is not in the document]
tabergma: We currently also inform the user about the status of the training. E.g. for each component we log 
"started training" and "finished training". Don't want to log this for every featurizer as they are called multiple times, so I just inform the user that a particular chunk is featurized.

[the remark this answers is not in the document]
dakshvar22: Right, I see your point 👍

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -781,6 +789,64 @@ def preprocess_train_data(self, training_data: TrainingData) -> RasaModelData:
...
+        # TODO:
+        #  we need to make sure that we can actually train on this data, e.g. see
+        #  checks inside train().
+        #  Do we have any intent examples to train on?
+        #  Do we have any response examples to train on?
+        #  Do we have any entity examples to train on?
+        #  Do we have enough training examples, e.g. enough labels?
+
+        def _load_data_func(file_path: Path) -> RasaModelData:
+            training_data_chunk = TrainingDataChunk.load_chunk(file_path)
```
joejuzl, line 816: Is this the right place to load the training data from files? I wonder if the data should be loaded before it's passed to the classifiers.

tabergma:
@dakshvar22 @joejuzl FYI: I discovered a couple of more edge cases that we need to fix before we can merge this PR. Will try to wrap up as much as I can today and will then handover the PR to @Ghostvv. So no need to review for now in detail I think.

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -39,23 +54,50 @@ def prepare_partial_training(
...
         **kwargs: Any,
     ) -> None:
-        """Train this component on the given chunk.
+        """Trains this component using the list of data chunk files.
 
         See parent class for more information.
         """
-        for example in training_data_chunk.entity_examples:
+        for data_chunk in data_chunk_files:
+            training_data_chunk = TrainingDataChunk.load_chunk(data_chunk.file_path)
```
joejuzl, line 80: This means we are loading the same files multiple times right? As we also do this is the classifier?

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -861,6 +880,51 @@ def _to_tf_features(self, features: List[Features]) -> Dict[Text, tf.train.Featu
 
         return tf_features
 
+    @staticmethod
+    def _relevant_message_data_keys() -> List[Text]:
```
joejuzl, line 884: Any reason this is a method?

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -861,6 +880,51 @@ def _to_tf_features(self, features: List[Features]) -> Dict[Text, tf.train.Featu
...
+        return [
+            TEXT,
+            INTENT,
+            RESPONSE,
+            INTENT_RESPONSE_KEY,
+            ENTITIES,
+            TOKENS_NAMES[TEXT],
+        ]
+
+    def _encode_message_data(
```
joejuzl, line 896: Is this the right place for this? Where do we normally encode training data?

rasa/nlu/classifiers/diet_classifier.py
tabergma: Maybe some background information: We currently are not able to train on large datasets due to out of memory issues. That's why we are introducing a new command called `rasa train --number-of-chunks <n>`. This divides the training data into multiple smaller chunks and featurizes them separately. During training we need to train on all data, e.g. one epoch of training should see all training data. As the training data at this point is saved in multiple files on disk, as we cannot have everything in memory at the same time, we pass the data files to the training method. During training we load one data file at a time to make sure we are not running out of memory. This has the benefit that we can (hopefully) train on larger datasets, but has the downside that training will take longer as we need to load data multiple times from files.

So to answer your question, yes this is the right place as we cannot load all the data before training.

rasa/nlu/extractors/entity_synonyms.py
tabergma: yes, see comment above

rasa/shared/nlu/training_data/training_data.py
tabergma: We did not encode the message data before. (Maybe encode is not the best word.) We need to write down the training data incl. features to files. As we also need to have the text and labels of the message, we need to "encode" the data into tf.Features.

rasa/nlu/extractors/entity_synonyms.py
tabergma: Ah no, did not see that this referred to entity synonyms. In this case each file is just loaded once.

tabergma:
@Ghostvv Training the rasa init project is working again, e2e and incremental training messed up some things. I also now build up the label index mapping during preparation in DIET, so that we can actually check if the model can be trained (especially important for Response Selector). However, some things still not work as expected:
- training our e2e example bot fails
- testing does not work
It might be that some featurizers do not work as expected as the testing pipeline is exactly the same as before. I don't think I find the time to investigate those two issues further.

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -767,6 +774,16 @@ def has_e2e_examples(self):
         return any(message.is_e2e_message() for message in self.training_examples)
 
 
+RELEVANT_MESSAGE_KEYS = [
```
tabergma, line 777: We most likely need to add ACTION_TEXT and ACTION_NAME for e2e training.

Ghostvv: @dakshvar22 do you remember why we need that list of constants?

rasa/nlu/utils/hugging_face/hf_transformers.py
```diff
@@ -684,9 +688,9 @@ def _get_docs_for_batch(
 
         return batch_docs
 
-    def train_chunk(
+    def train(
         self,
-        training_data_chunk: TrainingDataChunk,
+        training_data: TrainingDataFull,
```
Ghostvv, line 693: @dakshvar22 should this be `train_chunk`?

rasa/nlu/utils/spacy_utils.py
```diff
@@ -234,10 +234,10 @@ def train_chunk(
         See parent class for more information.
         """
         attribute_docs = self._docs_for_training_examples(
-            training_data_chunk.training_examples
+            training_data.training_examples
```
Ghostvv, line 237: @dakshvar22 it seems to me that these `docs` already contain dense features. I think it should be `train_chunk` then?

rasa/shared/nlu/training_data/training_data.py
Ghostvv: I think adding `ACTION_TEXT, ACTION_NAME` is enough. I think we need `TOKEN_NAMES[TEXT]` for entities

rasa/nlu/classifiers/keyword_intent_classifier.py
```diff
@@ -41,11 +42,11 @@ def __init__(
...
-        training_data_chunk: TrainingDataChunk,
+        data_chunk_files: List[DataChunkFile],
         config: Optional[RasaNLUModelConfig] = None,
         **kwargs: Any,
     ) -> None:
-        """Train this component on the given chunk.
+        """Trains this component using the list of data chunk files.
 
         See parent class for more information.
         """
```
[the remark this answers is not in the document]
Ghostvv: @dakshvar22 I think it should be done in the separate PR

rasa/nlu/utils/hugging_face/hf_transformers.py
dakshvar22: Hmm, I think because components like `HFTransformersNLP`, and `SpacyNLP` go from creating tokens to features inside one component, it doesn't need a separate `train_chunk` method explicitly because we are passing the complete training data [here](https://github.com/RasaHQ/rasa/pull/7551/files#diff-178f448a45faaee012886c9ac7564a55639e9ff7383e2c68ae4c1210020276daR285). But then those components should be doing the chunking themselves and featurizing chunks? These components are really an exception for the logic in the component class and break the abstraction between `train` and `train_chunk` I think.

rasa/nlu/utils/spacy_utils.py
dakshvar22: This one is also like the above point made about `HFTransformers`.

rasa/nlu/utils/hugging_face/hf_transformers.py
Ghostvv: but the problem is that they will create dense features for the whole training data, which will blow up in memory. I think the problem is that they are applied before tokenizers

dakshvar22: Yes, that's what I meant by these components being an exception for the logic in `component.py`

Ghostvv: let's add them to the to do list, and tackle it in the separate PR

Ghostvv:
@joejuzl @dakshvar22 the tests are passing, could you please review?

examples/e2ebot/data/nlu.yml
```diff
@@ -1,48 +0,0 @@
-version: "2.0"
```
joejuzl, line 1: why is this being deleted?

rasa/shared/nlu/training_data/message.py
```diff
@@ -127,7 +127,9 @@ def fingerprint(self) -> Text:
         Returns:
             Fingerprint of the message.
         """
-        return rasa.shared.utils.io.deep_container_fingerprint(self.data)
+        # ensure the same order of keys inside data
+        data = {k: v for k, v in sorted(self.data.items(), key=lambda item: item[0])}
```
joejuzl, line 131:
```suggestion
        data = dict(sorted(data.items()))
```

rasa/shared/nlu/training_data/tokens.py
```diff
@@ -0,0 +1,93 @@
...
+        Args:
+            token_dict: the dictionary that contains all properties of a token.
+
+        Returns:
+            A token.
+        """
+        return Token(
+            token_dict["text"],
+            token_dict["start"],
+            token_dict["end"] if "end" in token_dict else None,
```
joejuzl, line 65:
```suggestion
            token_dict.get("end"),
```

rasa/shared/nlu/training_data/tokens.py
```diff
@@ -0,0 +1,93 @@
...
+            token_dict["data"] if "data" in token_dict else None,
```
joejuzl, line 66:
```suggestion
            token_dict.get("data"),
```

rasa/shared/nlu/training_data/tokens.py
```diff
@@ -0,0 +1,93 @@
...
+            token_dict["lemma"] if "lemma" in token_dict else None,
```
joejuzl, line 67:
```suggestion
            token_dict.get("lemma")
```

rasa/shared/nlu/training_data/tokens.py
```diff
@@ -0,0 +1,93 @@
...
+            other.lemma,
+        )
+
+    def __lt__(self, other):
+        if not isinstance(other, Token):
+            return NotImplemented
+        return (self.start, self.end, self.text, self.lemma) < (
+            other.start,
+            other.end,
+            other.text,
```
joejuzl, line 87: does it makes sense to compare `text` for `<`?

joejuzl (changes requested):

examples/e2ebot/data/nlu.yml
Ghostvv: no idea 😅

Ghostvv: I guess artifact of testing something

rasa/shared/nlu/training_data/tokens.py
Ghostvv: hm, that's strange. Do you know the way to figure out where it is used?

Ghostvv: modified to only compare start and end

joejuzl: It will be used whenever the object is used with the infix operator `<`.
As for finding where it's used, you could pop a debugger/breakpoint in the method and run the tests and see if it catches.
My guess would be it's to do with sorting a list of these things.

joejuzl: hmm actually is that is the case, maybe we do want to use `text` in the comparison...

Ghostvv: yeah, I don't know it is implemented that way in `main`, let's not touch it

Ghostvv: this piece of code is just copy pasted to `shared`

joejuzl (commented):
The (albeit small) "engineering" changes look fine to me. Just checking though, the things that Tanja mentioned were broken:
> - training our e2e example bot fails
> - testing does not work

are now fixed?

I will wait for @dakshvar22 to give actual approval as he can review all the ml changes.

Ghostvv:
> training our e2e example bot fails

nlu part works, it was related to the fact that core messages were not included into chunks

> testing does not work

hm, I thought she meant tests. I'll check `rasa test`

Ghostvv:
`rasa test` and `rasa test nlu` seems to be working for me

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -91,6 +98,17 @@ def nlu_examples(self) -> List[Message]:
             ex for ex in self.training_examples if not ex.is_core_or_domain_message()
         ]
 
+    @lazy_property
+    def core_examples(self) -> List[Message]:
+        """Return examples which have come from NLU training data.
+
+        E.g. If the example came from a story or domain it is not included.
+
+        Returns:
```
dakshvar22, line 107: I think it's the opposite of what the function does.

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -382,6 +447,19 @@ def has_e2e_examples(self):
         """Checks if there are any training examples from e2e stories."""
         return any(message.is_e2e_message() for message in self.training_examples)
 
+    def filter_training_examples(
+        self, condition: Callable[[Message], bool]
+    ) -> "NLUPipelineTrainingData":
```
dakshvar22, line 452: Use `TrainingDataFull` as return type

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -884,6 +970,18 @@ def __init__(self, *args, **kwargs):
         )
 
 
+RELEVANT_MESSAGE_KEYS = [
```
dakshvar22, line 973: Can we move this to the constants file?

rasa/utils/tensorflow/models.py
```diff
@@ -71,11 +71,29 @@ class RasaModel(TmpKerasModel):
...
+        (without its trained weights) from this configuration.
+
+        The config of a layer does not include connectivity
+        information, nor the layer class name. These are handled
+        by `Network` (one layer of abstraction above).
+
+        Returns:
+            Python dictionary.
+        """
+        # TODO: figure out whether it can/needs to be implemented
```
dakshvar22, line 89: This can be removed I suppose?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -492,13 +495,15 @@ async def test_set_random_seed(component_builder, tmpdir):
         data=DEFAULT_DATA_PATH,
         component_builder=component_builder,
     )
+    print("----")
```
dakshvar22, line 498:
```suggestion
```

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -492,13 +495,15 @@ async def test_set_random_seed(component_builder, tmpdir):
...
     # second run
     (trained_b, _, persisted_path_b) = await train(
         _config,
         path=tmpdir.strpath + "_b",
         data=DEFAULT_DATA_PATH,
         component_builder=component_builder,
     )
+    print("----")
```
dakshvar22, line 506:
```suggestion
```

dakshvar22 (commented):
Some minor comments 🚀 
In addition, can you please make follow up issues for handling training in chunks for keyword intent classifier and language model featurizer. Also, can you please create a changelog already listing the changes like `TrainingData` is deprecated, etc. It will be easier to collect all the changes in the big PR.

I haven't tried training with this yet. Will try it later today. Also, since we refactored some bits of DIET and featurizer, we should run the model regression tests.

Ghostvv:
> In addition, can you please make follow up issues for handling training in chunks for keyword intent classifier and language model featurizer.

already in the list, let's create actual issues when we plan the next steps

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -596,27 +623,20 @@ def _create_label_data(
...
         """
-        # Collect one example for each label
-        labels_idx_examples = []
-        for label_name, idx in label_id_dict.items():
-            label_example = self._find_example_for_label(
-                label_name, training_data.intent_examples, attribute
-            )
-            labels_idx_examples.append((idx, label_example))
+        if not self._label_attribute:
+            return RasaModelData()
```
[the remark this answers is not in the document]
Ghostvv: when DIET performs only entity classification

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -749,99 +765,68 @@ def _add_label_features(
...
-        # keep one example for persisting and loading
-        self._data_example = model_data.first_data_example()
+    def _load_data_func(self, file_path: Path) -> RasaModelData:
+        training_data_chunk = TrainingDataChunk.load_chunk(file_path)
+        return self._preprocess_train_data(training_data_chunk)
 
+    def _train_model(
+        self,
+        model_data: RasaModelData,
+        data_generator: RasaDataGenerator,
```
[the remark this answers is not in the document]
Ghostvv: what do you mean? we train model in this method

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -859,6 +844,103 @@ def train(
...
+        data_generator = RasaDataChunkFileGenerator(
+            data_chunk_files,
+            self._load_data_func,
+            batch_size=self.component_config[BATCH_SIZES],
+            epochs=self.component_config[EPOCHS],
+            shuffle=True,
+            random_seed=self.component_config[RANDOM_SEED],
+        )
+
+        self._train_model(sample_model_data, data_generator)
```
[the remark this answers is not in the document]
Ghostvv: but we run validation in the end of each epoch, not batch

rasa/nlu/classifiers/diet_classifier.py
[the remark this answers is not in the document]
dakshvar22: Yes, but the third argument is `validation_data_generator` and hence I thought we can be a bit more explicit by renaming `data_generator` to `train_data_generator`.

rasa/nlu/classifiers/diet_classifier.py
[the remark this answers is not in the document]
dakshvar22: Yes, over here as well `data_generator` contains data for whole epoch and not one batch. Since we don't pass any `validation_data_generator` there won't be a validation run. All I'm saying is that we can possibly create a validation data generator as well when training on chunks and pass it to `_train_model` just like we do for usual training on line 926.

rasa/nlu/classifiers/diet_classifier.py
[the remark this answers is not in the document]
Ghostvv: aaah, I thought you suggest to rename the method itself

rasa/shared/nlu/training_data/training_data.py
Ghostvv: moved it inside `TrainingDataChunk` class

rasa/utils/tensorflow/models.py
Ghostvv: no, this is smth I don't understand still

rasa/nlu/classifiers/diet_classifier.py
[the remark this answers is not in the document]
Ghostvv: I suppose, but then we'll need to use a separate chunk for it. Let's do it in a separate PR, I'll add it to the list

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -215,7 +221,7 @@ def _load_model_instance(self, skip_model_load: bool) -> None:
 
     @classmethod
     def cache_key(
-        cls, component_meta: Dict[Text, Any], model_metadata: Metadata
+        cls, component_meta: Dict[Text, Any], model_metadata: "Metadata"
```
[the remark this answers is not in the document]
Ghostvv: I think there is circular imports, test complains

Ghostvv:
@dakshvar22 I answered your comments. ready to merge?

dakshvar22:
> Also, since we refactored some bits of DIET and featurizer, we should run the model regression tests.

Do you want to do this or are you confident that the changes shouldn't bring a regression for now?

Ghostvv:
/modeltest
```
include:
   - dataset: ["all"]
     config: ["all"]
```

Ghostvv:
@dakshvar22 regression tests look good to me. The training on carbon-bot is unstable on majority of PRs

dakshvar22:
Right, good to merge 👍

PR #7553, opened 2020-12-15
don't persist changed entities + autofill slots for policy entities

wochinge:
**Proposed changes**:
- don't persist `UserUttered` attribute which were changed by further events
- The tracker we return in the API / send to the action server now contains the combined `UserUttered.parse_data` in its field `latest_message`.`Combined` means that it contains the entities of NLU predictions and entities predicted by the policies. This allows users to access the entities as usual.
- support autofilling slots with entities predicted by policies
- rename `DefinePrevUserUtteredEntities` to `EntitiesAdded`


tests/shared/core/test_trackers.py
```diff
@@ -36,6 +36,8 @@
     LegacyForm,
     LegacyFormValidation,
     LoopInterrupted,
+    DefinePrevUserUtteredFeaturization,
```
wochinge, line 39: @Ghostvv @tmbo Brought up if we can come up with better names / names which generalize better:

How about 
* `UserUtteredFeaturizated`
* `EntitiesAdded`

Ghostvv: In the redesigned system, when/if we have them as separate events, I think it is good names. In the current system, I'd keep `Define` ones. But if you think it is better to rename now, please go for it

Ghostvv:
I'm not sure I can understand. So entities will be added every time in applied events? So that users will see "normal" `UserUttered` in custom actions?

Ghostvv:
if this is the case, I believe the `if` on line 728 in `events.py` should be removed

wochinge:
> I'm not sure I can understand. So entities will be added every time in applied events? So that users will see "normal" UserUttered in custom actions?

Entities are added via `applied_to` as you implemented it. However, any changes to `UserUttered.entities` won't affect the serialized version of the event. This gives you the updated `entities` for the featurization while preserving the original event.
A user would only see the original `UserUttered` in a custom action (without entities added by policies).

tests/shared/core/test_trackers.py
wochinge: @joejuzl What do you think?

joejuzl: If these are the only two which behave in this way I can see an argument for having them have a common prefix.

wochinge: Tom also meant the point that all events are using the past tense to stress the fact that they mirror something in the past. So I think it makes sense to at least take that into account.

wochinge: And events are so essential that they'll basically stay around forever one they were added.  Hence, more general names could help us to adapt in the future.

Ghostvv:
> A user would only see the original UserUttered in a custom action (without entities added by policies).}

then it's not very useful, this was the reason for updating `UserUttered` event, so that these entities are perceived by users exactly the same as extracted by TED

> This gives you the updated entities for the featurization while preserving the original event.

we actually don't use e2e entities as features

wochinge:
> we actually don't use e2e entities as features

So only entities / roles / groups from the NLU classification are considered as features?

> then it's not very useful, this was the reason for updating UserUttered event, so that these entities are perceived by users exactly the same as extracted by TED

This needs to be done in a different place then. We can't mutate the event. What specific locations are you thinking of when you mean they should be perceived the same as extracted by TED? It should work out of the box for everything except for custom actions. 

We could use `applied_events` instead combined with a `copy.deepcopy` but then we'd need to make sure we'd call that whenever exposing the tracker to the user. Currently we're e.g. not calling `applied_events` before sending them to the action server.

Ghostvv:
> So only entities / roles / groups from the NLU classification are considered as features?

yes

tests/shared/core/test_trackers.py
Ghostvv: taking into account new development around e2e entities, we can safely rename them

wochinge: Do you mean we should rename them because a lot of things around entities are gonna change or did you mean there is no need to rename as things are gonna stay the same? 🙈

wochinge:
@Ghostvv @joejuzl Updated according to what we discussed. How you can test this: 
1. Use the `e2e` example bot
2. train it
3. Send the message `Hello`
4. `GET <host>:<port>:/conversations/<conversation ID>/tracker` will return a tracker where `latest_message` contains the entities from the `TEDPolicy` while the actual `events` are untouched.

wochinge:
@Ghostvv We should in my opinion also updated [this check](https://github.com/RasaHQ/rasa/blob/12672a7da8298eac897bebb17253be484d90ac39/rasa/shared/core/trackers.py#L610) to work for the new event as well, shouldn't we?

joejuzl:
@wochinge 

> GET <host>:<port>:/conversations/<conversation ID>/tracker will return a tracker where latest_message contains the entities from the TEDPolicy while the actual events are untouched.

So the `DefinePrevUserUttered...` events now only exist in the tracker within the ensemble?

wochinge:
> So the DefinePrevUserUttered... events now only exist in the tracker within the ensemble?

Nope, it's a regular event as the others as well. We just changed that it does not longer affect the persisted version of `UserUttered`. In addition we changed the `latest_message` property of the tracker to make it easier for the users to access the combined entities (e.g. in custom actions)

joejuzl (approved):
I've tested it an the entities show as expected.

wochinge:
@joejuzl @Ghostvv Added support for autofilling slots based on entities predicted by polices. Please(re-)review

tests/shared/core/test_trackers.py
Ghostvv: I meant we can rename them because for entities we're not "defining" previous useruttered entities anymore, so the name `DefinePrevUserUtteredFeaturization` is not accurate. And since you pointed out that deprecating event names is a pain, we might go with the names we like from the start

Ghostvv (changes requested):
This solution seems to me very convoluted. Why do we update previous `UserUttered` with entities? Then we use a lot of different `.copy()` to mitigate that. Isn't it simpler to store `e2e_entities` separate on the tracker and add them to `latest_message` data directly?

wochinge:
>  Then we use a lot of different .copy() to mitigate that.

actually just one 😆 

> Isn't it simpler to store e2e_entities separate on the tracker and add them to latest_message data directly?

The nice thing about having it in `entities` is that story printing would work as usual. It allows to interact with the event as it'd be one thing.

rasa/shared/core/trackers.py
```diff
@@ -600,9 +623,15 @@ def update(self, event: Event, domain: Optional[Domain] = None) -> None:
...
-        if domain and isinstance(event, UserUttered):
-            # store all entities as slots
-            for e in domain.slots_for_entities(event.parse_data["entities"]):
+        if domain and isinstance(event, (UserUttered, DefinePrevUserUtteredEntities)):
+            entities = event.entities
+            if isinstance(event, UserUttered):
+                # Rather get entities from `parse_data` as
+                # `DefinePrevUserUtteredEntities` might have already affected the
+                # `UserUttered.entities` attribute
+                entities = event.parse_data["entities"]
```
Ghostvv, line 632: could you please unpack this convoluted `if`, I think checks for `UserUttered` and `DefinePrevUserUtteredEntities` can be done in parallel

Ghostvv (approved):
ok, if you think it is better 🚀

what about renaming? also left one refactoring comment

tests/shared/core/test_trackers.py
wochinge: Ok, so how about `EntitiesAdded`?

rasa/shared/core/trackers.py
Ghostvv: actually, wouldn't `event.parse_data["entities"] == event.entities` for `DefinePrevUserUtteredEntities`?

also, I think we have a constant for `"entities"`, no?

tests/shared/core/test_trackers.py
Ghostvv: let's add `e2e` in there? for explicitness

rasa/shared/core/trackers.py
Ghostvv: `# DefinePrevUserUtteredEntities might have already affected the UserUttered.entities attribute`? what does it mean? why behavior is not deterministic?

wochinge: > could you please unpack this convoluted if, I think checks for UserUttered and DefinePrevUserUtteredEntities can be done in parallel

I made it more explicit. 

> actually, wouldn't event.parse_data["entities"] == event.entities for DefinePrevUserUtteredEntities?

`DefinePrevUserUtteredEntities` doesn't have a `parse_data`?

> also, I think we have a constant for "entities", no?
 
done

rasa/shared/core/trackers.py
```diff
@@ -624,12 +625,13 @@ def update(self, event: Event, domain: Optional[Domain] = None) -> None:
...
         if domain and isinstance(event, (UserUttered, DefinePrevUserUtteredEntities)):
-            entities = event.entities
             if isinstance(event, UserUttered):
                 # Rather get entities from `parse_data` as
                 # `DefinePrevUserUtteredEntities` might have already affected the
                 # `UserUttered.entities` attribute
-                entities = event.parse_data["entities"]
+                entities = event.parse_data[ENTITIES]
+            else:
+                entities = event.entities
```
Ghostvv: what about:
```
if domain:
   if isinstance(event, UserUttered):
   if isinstance(event, DefinePrevUserUtteredEntities):
```
?

tests/shared/core/test_trackers.py
wochinge: In theory it doesn't really matter where entities are coming from, no?

rasa/shared/core/trackers.py
Ghostvv: > DefinePrevUserUtteredEntities doesn't have a parse_data?

ok

what about `might` why it is not deterministic?

tests/shared/core/test_trackers.py
wochinge: Eg. if we split up events in something like 

* `UserInputEvent` -> pure unchanged user input
* `NLUIntentEvent`
* `NLUEntityEvent`
* `PolicyPrediction`

then we could actually use `EntityAdded` both for NLU and Policy prediction

Ghostvv: sounds good

rasa/shared/core/trackers.py
wochinge: I think that's style dependent and should be optional to address.

Ghostvv: I mean... this is colossal `if domain and isinstance(event, (UserUttered, DefinePrevUserUtteredEntities)):` and then the solitary `else` is confusing

Ghostvv: it takes more than a second to understand this `if` for someone else

Ghostvv: I don't think it is style dependent, we should make sure that the code is as simple as possible to grasp by looking at it

tests/shared/core/test_trackers.py
wochinge: done

rasa/shared/core/events.py
```diff
@@ -686,21 +688,22 @@ def __eq__(self, other) -> bool:
         return self.use_text_for_featurization == other.use_text_for_featurization
 
 
-class DefinePrevUserUtteredEntities(DefinePrevUserUttered):
-    """Event that is used to set entities on a previous user uttered event."""
+class EntitiesAdded(DefinePrevUserUttered):
```
Ghostvv, line 691: I'd say the super class `DefinePrevUserUttered` should be removed, what do you think?

rasa/shared/core/trackers.py
wochinge: I'm trying to play it save. Pretty sure that this would happen for  the `InMemoryTrackerStore`

rasa/shared/core/trackers.py
wochinge: I still think it's clear. Happy to hear @joejuzl 's opinion on it.

rasa/shared/core/events.py
wochinge: kept it but made it a more general approach to deduplicate this behavior for other events as well

rasa/shared/core/events.py
```diff
@@ -345,6 +345,18 @@ def __eq__(self, other: Any) -> bool:
         return True
 
 
+class SkipEventInMDStoryMixin(Event, ABC):
```
Ghostvv, line 348: do we need that? let's just add `as_story_string` to all the methods

Ghostvv: @joejuzl what do you think. this method seems to me redundant but also increases cognitive complexity of events

joejuzl (approved):

rasa/shared/core/events.py
joejuzl: I think it's fine. It means you can easily identify events which won't be in the story rather than having to look at the implementation of the method in each event.

Ghostvv: to me it contradicts `Occam’s razor` principle, but after all it's your part of code

rasa/shared/core/trackers.py
Ghostvv: ok, could you please add comment, so it'll be easier to modify this piece in the future

rasa/shared/core/events.py
Ghostvv: for me personally, it makes it harder to understand, I see the method - I know how it is represented, looking at superclass requires more cognitive function

joejuzl: I see arguments for both ways - but in the interest of speed (and since it's already done like this) - I think we should keep it as is.

rasa/shared/core/trackers.py
wochinge: As you already merged, should I directly push the comment to `e2e` or are we good?

Ghostvv: sorry, forgot about this one. as it is just a comment could you please directly merge to `e2e`.

wochinge: done

PR #7554, opened 2020-12-15
Merge master into e2e

dakshvar22:
**Proposed changes**:
- 


rasa/shared/core/domain.py
```diff
@@ -714,13 +717,13 @@ def is_retrieval_intent_template(
...
         """Sets up the default slots and slot values for the domain."""
-        self.add_requested_slot()
-        self.add_knowledge_base_slots()
-        self.add_categorical_slot_default_value()
+        self._add_requested_slot()
+        self._add_knowledge_base_slots()
+        self._add_categorical_slot_default_value()
 
-    def add_categorical_slot_default_value(self) -> None:
+    def _add_categorical_slot_default_value(self) -> None:
```
dakshvar22, line 726: @wochinge This part seems to be part of what you fixed in the domain regarding default slots being added while persisting the domain. Do you think the merge is correctly done then?

wochinge: looks good!

tests/shared/core/training_data/story_writer/test_markdown_story_writer.py
```diff
@@ -17,4 +17,5 @@ async def test_tracker_dump_e2e_story(default_agent: Agent):
         "* greet: /greet",
         "    - utter_greet",
         "* goodbye: /goodbye",
+        "    - utter_goodbye",
```
dakshvar22, line 20: @wochinge Is this correct? For some reason the exported story had `utter_goodbye` as well as the last action but I am wondering why was it not there on the right hand side in the first place.

wochinge: I added a new story for goodbye. Previously `AugementedMemo` wouldn't know what to do and predict `action_listen`. When I decided that the `default_domain` should contain a `form`, I had to add the `RulePolicy`, which predicts a fallback in case it doesn't know what to do. To get everybody back on track I had to add story for `goodbye` so that `AugmentedMemoizationPolicy` can be in charge once more.

dakshvar22: Ahh okay so then it is needed indeed for this test. Thanks for the explanation!

PR #7558, opened 2020-12-15
Review comments2

wochinge:
**Proposed changes**:
- force implementation of `__eq__` for Events
- add a helper class for Events without attributes to avoid duplicating their `__eq__` implementation
- mark `Event` class as abstract

### DIFF
--- a/rasa/shared/core/events.py
+++ b/rasa/shared/core/events.py
@@ -329,7 +329,7 @@ class EventWithoutExtraAttributes(Event, ABC):
     """Class to deduplicate common behavior for events without additional attributes."""
 
     def __eq__(self, other: Any) -> bool:
-        """Returns text representation of event."""
+        """Compares object with other object."""
         if not isinstance(other, self.__class__):
             return NotImplemented
 
--- a/rasa/shared/core/events.py
+++ b/rasa/shared/core/events.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import re
+from abc import ABC
 
 import jsonpickle
 import time
@@ -314,14 +315,23 @@ class Event:
 
     def __eq__(self, other: Any) -> bool:
         """Compares object with other object."""
-        if not isinstance(other, self.__class__):
-            return NotImplemented
-
-        return True
+        # Every class should implement this
+        raise NotImplementedError()
 
     def __str__(self) -> Text:
         """Returns text representation of event."""
         return f"{self.__class__.__name__}()"
+
+
+class EventWithoutExtraAttributes(Event, ABC):
+    """Class to deduplicate common behavior for events without additional attributes."""
+
+    def __eq__(self, other: Any) -> bool:
+        """Returns text representation of event."""
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+
+        return True
 
 
 # noinspection PyProtectedMember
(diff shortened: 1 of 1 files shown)


rasa/shared/core/events.py
```diff
@@ -314,21 +317,30 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
 
     def __eq__(self, other: Any) -> bool:
         """Compares object with other object."""
-        if not isinstance(other, self.__class__):
-            return NotImplemented
-
-        return True
+        # Every class should implement this
```
joejuzl, line 320: Does this mean that `Event` should be an abstract class? (And this an abstract method?)

rasa/shared/core/events.py
```diff
@@ -314,21 +317,30 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
...
         """Returns text representation of event."""
         return f"{self.__class__.__name__}()"
 
 
-# noinspection PyProtectedMember
+class EventWithoutExtraAttributes(Event, ABC):
+    """Class to deduplicate common behavior for events without additional attributes."""
+
+    def __eq__(self, other: Any) -> bool:
+        """Returns text representation of event."""
```
joejuzl, line 332: This comment seems wrong.

rasa/shared/core/events.py
```diff
@@ -314,21 +317,30 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
...
+        # Every class should implement this
+        raise NotImplementedError()
 
     def __str__(self) -> Text:
         """Returns text representation of event."""
         return f"{self.__class__.__name__}()"
 
 
-# noinspection PyProtectedMember
+class EventWithoutExtraAttributes(Event, ABC):
```
joejuzl, line 328: I find this name is confusing and likely to form a grab bag. Maybe something along the lines of `AlwaysEqualEventMixin`?

joejuzl (changes requested):

rasa/shared/core/events.py
wochinge: good catch! Should have been an abstract class all along

rasa/shared/core/events.py
wochinge: Mhm, it's not always equal. It's only equal if it has the same type 🤔 In my opinion it'd actually be ok if it ends up as grab bag, e.g. I see that most of this classes have `as_story_string` implemented with `return self.type_name`

wochinge: actually I like your name suggestion, will go with it 👍🏻

joejuzl: I do see your point though! If we add more functionality to it we can change the name again to fit.

wochinge: Yep 👍🏻 Let's keep it narrow for now. We can't fix all problems in the `e2e` branch anyway

rasa/shared/core/events.py
```diff
@@ -314,21 +317,30 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
 
     def __eq__(self, other: Any) -> bool:
```
joejuzl, line 318: We can also make this an `abstractmethod` now right?

joejuzl (approved):
Looks good. Just one final point.

rasa/shared/core/events.py
wochinge: actually didn't know about this decorator. Added it 👍🏻

PR #7570, opened 2020-12-16
Prepare release 2.2.0

joejuzl:


wochinge (approved):
Thanks for coordinating this release so patiently! 💯

m-vdb:
@joejuzl I've just pushed a commit to fix the bandit check, they updated a few things in version 1.7 and it's only coming up now on this PR 🤷🏻

m-vdb:
about the timeout issues, it's possible to control the timeouts on particular tests using `@pytest.mark.timeout(300)` for instance. Default timeout is [120 second](https://github.com/RasaHQ/rasa/blob/master/pytest.ini)

joejuzl:
> about the timeout issues, it's possible to control the timeouts on particular tests using @pytest.mark.timeout(300) for instance. Default timeout is 120 second

We probably don't want tests taking more than 120 seconds though to be honest. Increasing the timeout seems like fixing the wrong problem.

PR #7575, opened 2020-12-17
Amendmend to PR #7423

hotzenklotz:
**Proposed changes**:
@dakshvar22 To my dismay I discovered that my refactoring reintroduced a bug in my original PR #7423. Unfortunately, it was not captured by any of the unit the tests. I came across it when running a new response selector evaluation today. This PR should fix it.

In short, the default value for a response selector key was wrong. It needs to default to `None` for all "regular" intent examples or otherwise those won't be filtered out further down the pipeline and crash the sklearn reports.

cc @tmbo 

### DIFF
--- a/changelog/7575.bugfix.rst
+++ b/changelog/7575.bugfix.rst
@@ -1,0 +1,1 @@
+Fix scikit-learn crashing during response selector evaluation.
--- a/changelog/7575.bugfix.rst
+++ b/changelog/7575.bugfix.rst
@@ -1,1 +1,1 @@
-Fix scikit-learn crashing during response selector evaluation.
+Fix scikit-learn crashing during evaluation of `ResponseSelector` predictions.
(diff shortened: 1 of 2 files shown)


dakshvar22:
Thanks @hotzenklotz. Can you please also post what the output now looks like after running a fresh evaluation on the trained response selector model?

hotzenklotz:
It looks pretty much the same as before. Only now the response selector labels include their parents intent name as well, e.g. `ask_faq/question_1`.

image

dakshvar22:
Thanks! Can you please also add a unit test which would check if training examples with no response keys are correctly filtered during evaluation?

hotzenklotz:
There is already a unit test for that: https://github.com/RasaHQ/rasa/blob/1.10.x/tests/nlu/test_evaluation.py#L582

dakshvar22 (approved):

dakshvar22:
@hotzenklotz I just pushed a changelog to indicate this is a bugfix.

PR #7585, opened 2020-12-17
correctly replace database name in query

wochinge:
**Proposed changes**:
- correctly interpolate text by using the `sqlalchemy.text` construct
- make sure the database creation is only used for postgresql database

### DIFF
--- a/rasa/core/tracker_store.py
+++ b/rasa/core/tracker_store.py
@@ -917,33 +917,44 @@ class SQLTrackerStore(TrackerStore):
         )
 
     def _create_database_and_update_engine(self, db: Text, engine_url: "URL"):
-        """Create databse `db` and update engine to reflect the updated `engine_url`."""
-
+        """Creates database `db` and updates engine accordingly."""
         from sqlalchemy import create_engine
+
+        if not self.engine.dialect.name == "postgresql":
+            rasa.shared.utils.io.raise_warning(
+                "The parameter 'login_db' can only be used with a postgres database.",
+            )
+            return
 
         self._create_database(self.engine, db)
         engine_url.database = db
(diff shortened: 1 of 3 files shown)


rasa/core/tracker_store.py
```diff
@@ -917,33 +917,44 @@ def get_db_url(
         )
 
     def _create_database_and_update_engine(self, db: Text, engine_url: "URL"):
-        """Create databse `db` and update engine to reflect the updated `engine_url`."""
-
+        """Creates database `db` and updates engine accordingly."""
         from sqlalchemy import create_engine
 
+        if not self.engine.dialect.name == "postgresql":
```
wochinge, line 923: we could also use https://sqlalchemy-utils.readthedocs.io/en/latest/_modules/sqlalchemy_utils/functions/database.html#create_database but that only supports mysql in addition which doesn't seem worth to include it as extra library (`login_db`  doesn't make sense for sqlite)

rasa/core/tracker_store.py
```diff
@@ -917,33 +917,44 @@ def get_db_url(
...
+            rasa.shared.utils.io.raise_warning(
+                f"The parameter 'login_db' can only be used with a postgres database.",
```
degiz, line 925: Seems a bit a wrong place to warn about `login_db`, since it was passed as `login_db` in the constructor of `SQLTrackerStore`.

degiz (approved):
Thanks a lot for a super quick fix 🙏

rasa/core/tracker_store.py
wochinge: It's called from the constructor, no? And we can only safely determine the dialect once we're connected 🤷🏻

PR #7615, opened 2020-12-21
E2e rec at 20

koernerfelicia:
Note: this won't be merged, just for e2e tests on Advising and MultiWOZ


rasa/core/policies/ted_policy.py
```diff
@@ -602,12 +603,28 @@ def predict_action_probabilities(
 
         output = self.model.predict(model_data)
 
+        num_labels = len(domain.action_names_or_texts)
+        indices_to_choose_from = np.concatenate(
+            (np.arange(action_index), np.arange(action_index + 1, num_labels))
+        )
```
Ghostvv, line 609: this seems to me exactly the same as `np.arange(num_labels)`

Ghostvv (commented):
left one comment, otherwise looks good

rasa/core/policies/ted_policy.py
koernerfelicia: I think it's not the same, because `indices_to_choose_from` should not include `action_index` (because we shuffle `indices_to_choose_from` and only keep the first `19` after shuffling).

koernerfelicia: Then `action_index` is added back to the `19` shuffled indices to make for `20` total that definitely includes the `action_index`

Ghostvv: aaah, yes, good point, arrange doesn't include it

PR #7616, opened 2020-12-21
Add sigmoid to softmax loss

dakshvar22:
**Proposed changes**:

- Constrain similarity values to an approximate range in `DotProductLoss` by applying sigmoid over them during training.
- Also, added an option `model_confidence` to each ML component. It affects how model's confidence for each label is computed during inference. It can take three values -
1. `softmax` - Similarities between input and label embeddings are post-processed with a softmax function, as a result of which confidence for all labels sum up to 1.
2. `cosine` - Cosine similarity between input label embeddings. Confidence for each label is in the range `[-1,1]`.
3. `inner` - Dot product similarity between input and label embeddings. Confidence for each label in in an unbounded range.

Change autoconfig to use `constrain_similarities=True` and `model_confidence=cosine`.


dakshvar22:
/modeltest

```yml
dataset_branch: "hint3"
include:
 - dataset: ["curekart_full", "curekart_subset", "powerplay11_full", "powerplay11_subset", "sofmattress_full", "sofmattress_subset"]
   config: ["BERT + DIET(bow) + ResponseSelector(bow)", "BERT + DIET(seq) + ResponseSelector(t2t)", "Sparse + BERT + DIET(bow) + ResponseSelector(bow)", "Sparse + BERT + DIET(seq) + ResponseSelector(t2t)", "Sparse + DIET(bow) + ResponseSelector(bow)", "Sparse + DIET(seq) + ResponseSelector(t2t)", "ConveRT + DIET(seq) + ResponseSelector(t2t)", "Sparse + ConveRT + DIET(seq) + ResponseSelector(t2t)", "ConveRT + DIET(bow) + ResponseSelector(bow)", "Sparse + ConveRT + DIET(bow) + ResponseSelector(bow)"]
```

changelog/7616.improvement.md
```diff
@@ -0,0 +1,6 @@
+Constrain similarity values to an approximate range in `DotProductLoss` by applying sigmoid over them during training.
+
+This affects the default behaviour of the loss function(`loss_type=softmax`) inside machine learning(ML) components - `DIETClassifier`, `ResponseSelector` and `TEDPolicy`.
```
Ghostvv, line 3: I'd suggest to rename `softmax` to cross-entropy loss

rasa/core/policies/ted_policy.py
```diff
@@ -317,11 +324,25 @@ def __init__(
         self._label_data: Optional[RasaModelData] = None
         self.data_example: Optional[Dict[Text, List[np.ndarray]]] = None
 
+    def _check_similarity_confidence_setting(self) -> None:
```
Ghostvv, line 327: this check also should be done for DIET. Let's move it into `utils`

rasa/core/policies/ted_policy.py
```diff
@@ -609,10 +630,14 @@ def predict_action_probabilities(
         confidence, is_e2e_prediction = self._pick_confidence(confidences, similarities)
 
         if self.config[LOSS_TYPE] == SOFTMAX and self.config[RANKING_LENGTH] > 0:
-            confidence = rasa.utils.train_utils.normalize(
+            confidence = rasa.utils.train_utils.sort_and_rank(
                 confidence, self.config[RANKING_LENGTH]
             )
 
+            if self.config[RELATIVE_CONFIDENCE]:
```
Ghostvv, line 637: `cosine` similarity should not be normalized?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -283,13 +290,27 @@ def _check_share_hidden_layers_sizes(self) -> None:
                     f"{HIDDEN_LAYERS_SIZES} must coincide."
                 )
 
+    def _check_similarity_confidence_setting(self) -> None:
```
Ghostvv, line 293: similar as with ted, let's move this out as it is general

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -854,10 +875,15 @@ def _predict_label(
...
         ):
-            message_sim = train_utils.normalize(
+            message_sim = train_utils.sort_and_rank(
                 message_sim, self.component_config[RANKING_LENGTH]
             )
 
+            if self.component_config[RELATIVE_CONFIDENCE]:
+                # Normalize the values if returned probabilities are from
+                # softmax(hence relative to each other).
+                message_sim = train_utils.normalize(message_sim)
```
Ghostvv, line 885: similar here, I don't think cosine should be normalized

rasa/utils/tensorflow/layers.py
```diff
@@ -580,7 +586,10 @@ def __init__(
         name: Optional[Text] = None,
         parallel_iterations: int = 1000,
         same_sampling: bool = False,
+        constrain_similarities: bool = True,
+        relative_confidence: bool = True,
     ) -> None:
+        """Declare instance variables with default values."""
```
Ghostvv, line 592: please add Args

rasa/utils/tensorflow/layers.py
```diff
@@ -731,14 +742,26 @@ def sim(a: tf.Tensor, b: tf.Tensor, mask: Optional[tf.Tensor] = None) -> tf.Tens
...
+            sim: Computed similarities
+            similarity_type: Similarity function to use - COSINE, INNER, AUTO.
+
+        Returns:
+            Confidences corresponding to each similarity value.
+        """
         if similarity_type == COSINE:
             # clip negative values to zero
             return tf.nn.relu(sim)
         else:
```
Ghostvv, line 758: you don't need `else` if you `return`

rasa/utils/tensorflow/layers.py
```diff
@@ -852,18 +875,50 @@ def _loss_softmax(
...
+                    tf.zeros_like(sigmoid_logits[..., 1:], tf.float32),
+                ],
+                axis=-1,
+            )
+
+            sigmoid_loss = tf.nn.sigmoid_cross_entropy_with_logits(
+                labels=sigmoid_labels, logits=sigmoid_logits
+            )
+
+            loss += tf.reduce_mean(sigmoid_loss, axis=-1)
```
Ghostvv, line 920: this mean is over number of logits?

rasa/core/policies/ted_policy.py
```diff
@@ -272,6 +274,11 @@ class TEDPolicy(Policy):
         FEATURIZERS: [],
         # If set to true, entities are predicted in user utterances.
         ENTITY_RECOGNITION: True,
+        # if 'True' applies sigmoid on all similarity terms and adds it to the loss function to
+        # ensure that similarity values are approximately bounded. Used inside softmax loss only.
+        CONSTRAIN_SIMILARITIES: True,
+        # Return softmax based probabilities during prediction.
+        RELATIVE_CONFIDENCE: True,
```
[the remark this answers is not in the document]
akelad: what do you mean by "how should we change default parameters?"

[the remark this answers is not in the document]
Ghostvv: if we change defaults right now, the users will have different behavior which I think contradicts our policy for minor releases. So my question is do we drag it till `3.0`?

rasa/utils/tensorflow/layers.py
dakshvar22: Yes

rasa/core/policies/ted_policy.py
dakshvar22: Good point! I added a check for this but I also added a warning if the user has configured to use `cosine` with `cross_entropy` loss or `inner` with `margin` loss.

rasa/nlu/classifiers/diet_classifier.py
dakshvar22: Added a check 👍

rasa/utils/tensorflow/layers.py
Ghostvv: could you please add a comment

changelog/7616.improvement.md
```diff
@@ -0,0 +1,6 @@
+Constrain similarity values to an approximate range in `DotProductLoss` by applying sigmoid over them during training.
```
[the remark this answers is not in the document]
dakshvar22: rephrased

changelog/7616.improvement.md
dakshvar22: Deprecated the value `softmax` for `loss_type` in favour of `cross_entropy` with a deprecation warning.

rasa/core/policies/ted_policy.py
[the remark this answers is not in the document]
akelad: Yep I would guess so. You can add warnings saying we'll be changing the defaults, and that they should give this a try though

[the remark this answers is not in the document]
akelad: Though, actually not 100% sure what our guidelines are wrt to minors vs majors and different model behaviours

rasa/utils/train_utils.py
```diff
@@ -58,14 +80,35 @@ def update_similarity_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
...
-        if config[LOSS_TYPE] == SOFTMAX:
+        if config[LOSS_TYPE] == CROSS_ENTROPY:
             config[SIMILARITY_TYPE] = INNER
         elif config[LOSS_TYPE] == MARGIN:
             config[SIMILARITY_TYPE] = COSINE
 
     return config
 
 
+def update_loss_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
```
Ghostvv, line 91: we already have a function for deprecating warnings: `check_core_deprecated_options`, `check_deprecated_options`. I'd suggest to reuse them

rasa/core/policies/ted_policy.py
Ghostvv: I guess this can be removed then?

rasa/utils/tensorflow/layers.py
```diff
@@ -566,6 +572,12 @@ class DotProductLoss(tf.keras.layers.Layer):
             to run in parallel.
         same_sampling: Boolean, if 'True' sample same negative labels
             for the whole batch.
+        constrain_similarities: Boolean, if 'True' applies sigmoid on all
+            similarity terms and adds to the loss function to
+            ensure that similarity values are approximately bounded.
+            Used inside _loss_softmax() only.
```
Ghostvv, line 578:
```suggestion
            Used inside _loss_cross_entropy() only.
```

rasa/utils/tensorflow/layers.py
```diff
@@ -580,7 +592,40 @@ def __init__(
...
         same_sampling: bool = False,
+        constrain_similarities: bool = True,
+        relative_confidence: bool = True,
     ) -> None:
+        """Declare instance variables with default values.
+
+        Args:
+            num_neg: Positive integer, the number of incorrect labels;
+                the algorithm will minimize their similarity to the input.
+            loss_type: The type of the loss function, either 'softmax' or 'margin'.
```
Ghostvv, line 603:
```suggestion
            loss_type: The type of the loss function, either 'cross_entropy' or 'margin'.
```
same in the class description

rasa/utils/tensorflow/layers.py
```diff
@@ -580,7 +592,40 @@ def __init__(
...
+                the confidence of the correct prediction.
+            name: Optional name of the layer.
+            parallel_iterations: Positive integer, the number of iterations allowed
+                to run in parallel.
+            same_sampling: Boolean, if 'True' sample same negative labels
+                for the whole batch.
+            constrain_similarities: Boolean, if 'True' applies sigmoid on all
+                similarity terms and adds to the loss function to
+                ensure that similarity values are approximately bounded.
+                Used inside _loss_softmax() only.
```
Ghostvv, line 625:
```suggestion
                Used inside _loss_cross_entropy() only.
```

rasa/utils/tensorflow/layers.py
Ghostvv: It seems you didn't push the change

rasa/utils/tensorflow/layers.py
dakshvar22: Removing the args from the class description since we added to the constructor.

rasa/utils/train_utils.py
dakshvar22: Those functions raise a warning if the key in the configuration dict has changed. Here the value is changed.

Ghostvv: oh, yes, sorry. missed that

changelog/7616.improvement.md
```diff
@@ -0,0 +1,8 @@
+Added cross-entropy loss over sigmoid of all similarity values to constrain them in an approximate range in `DotProductLoss`.
+
+This affects the default behaviour of the loss function(`loss_type=cross_entropy`) inside machine learning (ML) components - `DIETClassifier`, `ResponseSelector` and `TEDPolicy`.
```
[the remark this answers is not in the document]
Ghostvv: Please also mention in the migration guide that fallback thresholds might be affected by this change (regardless of the option really, because we add sigmoids by default)

Ghostvv:
@dakshvar22 since you introduced `inner` product, did you fix printing histograms?

dakshvar22:
@Ghostvv Yes, the histogram plotting now takes into consideration the range of values passed to it. It does not assume that values will be in the range `[0,1]`.

Ghostvv:
@dakshvar22 could you run regression tests with `cosine` as confidences, I quite consistently get 1% worse accuracy

dakshvar22:
@Ghostvv Results seem fine to me, do you see anything suspicious?

dakshvar22:
@Ghostvv Made the changes to defaults as discussed on slack.
@wochinge Can you please review parts related to autoconfig?

Ghostvv:
the latest results are results with cosine as confidence?

dakshvar22:
@Ghostvv Yes

rasa/utils/tensorflow/layers.py
```diff
@@ -685,24 +697,49 @@ def _sample_negatives(
...
         )
 
-    @staticmethod
-    def sim(a: tf.Tensor, b: tf.Tensor, mask: Optional[tf.Tensor] = None) -> tf.Tensor:
+    def sim(
+        self, a: tf.Tensor, b: tf.Tensor, mask: Optional[tf.Tensor] = None
+    ) -> tf.Tensor:
         """Calculate similarity between given tensors."""
-
+        if self.similarity_type == COSINE:
```
Ghostvv, line 704: for `margin` loss, it seems to me you'd apply double normalization during prediction, I think it'll be clearer if you make `similarity_type` an input argument to this method

rasa/utils/tensorflow/layers.py
```diff
@@ -815,19 +852,52 @@ def _loss_softmax(
...
+                [sim_pos, sim_neg_il, sim_neg_ll, sim_neg_ii, sim_neg_li], axis=-1
+            )
+
+            sigmoid_labels = tf.concat(
+                [
+                    tf.expand_dims(
+                        tf.ones_like(sigmoid_logits[..., 0], tf.float32), -1
+                    ),
+                    tf.zeros_like(sigmoid_logits[..., 1:], tf.float32),
+                ],
```
Ghostvv, line 890: this can be simplified, also you don't need explicit `dtype` since `tf._like` will pick a type automatically the same as for inputs
```suggestion
                [tf.ones_like(sigmoid_logits[..., :1]), tf.zeros_like(sigmoid_logits[..., 1:])],
```

rasa/utils/tensorflow/models.py
```diff
@@ -790,6 +791,9 @@ def _prepare_dot_product_loss(
             scale_loss,
             # set to 1 to get deterministic behaviour
             parallel_iterations=1 if self.random_seed is not None else 1000,
```
Ghostvv, line 793: I think this parameter is not used anymore? Could you please remove it?

rasa/utils/train_utils.py
```diff
@@ -37,19 +40,38 @@
...
+    """
+    new_values = values.copy()
+
+    if np.sum(new_values) > 0:
+        new_values = new_values / np.sum(new_values)
+
+    return new_values
+
+
+def filter_top_k(values: np.ndarray, ranking_length: Optional[int] = 0) -> np.ndarray:
```
Ghostvv, line 60: there is no need for this method as a separate, the filtering was done only for normalization

Ghostvv (changes requested):

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -367,6 +369,72 @@ async def test_softmax_normalization(
...
+    output_length: int,
+    monkeypatch: MonkeyPatch,
+):
+    pipeline = as_pipeline(
+        "WhitespaceTokenizer", "CountVectorsFeaturizer", "DIETClassifier"
+    )
+    assert pipeline[2]["name"] == "DIETClassifier"
+    pipeline[2].update(classifier_params)
+
+    _config = RasaNLUModelConfig({"pipeline": pipeline})
```
wochinge, line 404: can we limit the epochs please to speed up the tests?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -367,6 +369,72 @@ async def test_softmax_normalization(
     assert parse_data.get("intent") == intent_ranking[0]
 
 
+@pytest.mark.parametrize(
```
wochinge, line 372: don't we also need tests for the response selector / `TEDPolicy`?

wochinge (changes requested):

changelog/7616.improvement.md
```diff
@@ -0,0 +1,16 @@
+Added an option `constrain_similarities` which adds sigmoid cross-entropy loss on all similarity values to constrain them to an approximate range in `DotProductLoss`.
+
+This affects the behaviour of the loss function(`loss_type=cross_entropy`) inside machine learning (ML) components - `DIETClassifier`, `ResponseSelector` and `TEDPolicy`.
+By default, the parameter is set to `False` but users are encouraged to set it to `True` and re-train their assistants as it will be set to `True` by default, Rasa Open Source 3.0 onwards.
```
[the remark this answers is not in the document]
dakshvar22: Wouldn't they be more encouraged if we use `logger.warning`? `logger.info` log could just blend in with other logs and will be easy to miss IMO.

rasa/utils/tensorflow/layers.py
```diff
@@ -851,12 +921,12 @@ def _chosen_loss(self) -> Callable:
 
         if self.loss_type == MARGIN:
             return self._loss_margin
-        elif self.loss_type == SOFTMAX:
-            return self._loss_softmax
+        elif self.loss_type == CROSS_ENTROPY:
+            return self._loss_cross_entropy
         else:
             raise ValueError(
```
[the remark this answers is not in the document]
Ghostvv: this code is meant to be rasa independent. Let's keep `ValueError`

rasa/utils/tensorflow/layers.py
```diff
@@ -591,6 +595,14 @@ def __init__(
         self.scale_loss = scale_loss
         self.parallel_iterations = parallel_iterations
         self.same_sampling = same_sampling
+        self.constrain_similarities = constrain_similarities
+        self.model_confidence = model_confidence
+        self.similarity_type = similarity_type
+        if self.similarity_type and self.similarity_type not in {COSINE, INNER}:
+            raise ValueError(
```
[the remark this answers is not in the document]
Ghostvv: this code is meant to be rasa independent. Let's keep ValueError. If you use rasa that cannot happen, because we check for it in diet and ted

rasa/utils/tensorflow/layers.py
```diff
@@ -288,20 +287,16 @@ class Embed(tf.keras.layers.Layer):
...
 
-        self.similarity_type = similarity_type
-        if self.similarity_type and self.similarity_type not in {COSINE, INNER}:
-            raise ValueError(
-                f"Wrong similarity type '{self.similarity_type}', "
-                f"should be '{COSINE}' or '{INNER}'."
-            )
+        Args:
+            embed_dim: Dimensionality of the output space.
+            reg_lambda: Regularization factor.
```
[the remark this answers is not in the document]
Ghostvv: because regularization constant is almost always called `lambda`, `lambda` is typical name for a `float`

rasa/core/policies/ted_policy.py
```diff
@@ -610,11 +624,20 @@ def predict_action_probabilities(
...
         confidence, is_e2e_prediction = self._pick_confidence(confidences, similarities)
 
-        if self.config[LOSS_TYPE] == SOFTMAX and self.config[RANKING_LENGTH] > 0:
-            confidence = rasa.utils.train_utils.normalize(
+        if self.config[LOSS_TYPE] == CROSS_ENTROPY and self.config[RANKING_LENGTH] > 0:
+            confidence = rasa.utils.train_utils.filter_top_k(
                 confidence, self.config[RANKING_LENGTH]
             )
 
+            if (
```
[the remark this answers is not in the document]
Ghostvv: I wouldn't do that, especially since normalization will be deprecated

rasa/utils/tensorflow/layers.py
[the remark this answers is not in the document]
wochinge: I think we should be consistent. It doesn't make the code any more or less dependent on Rasa by using that exception type. Also this code sits in the `rasa` repository, so it has to integrate.

changelog/7616.improvement.md
```diff
@@ -0,0 +1,16 @@
...
+Configuration option `loss_type=softmax` is now deprecated. Use `loss_type=cross_entropy` instead.
+
+Also, added an option `model_confidence` to each ML component. It affects how m mponent. It affects how model's confidence for each label is computed during inference. It can take three values -
+1. `softmax` - Similarities between input and label embeddings are post-processed with a softmax function, as a result of which confidence for all labels sum up to 1.
+2. `cosine` - Cosine similarity between input label embeddings. Confidence for each label will be in the range `[-1,1]`.
+3. `inner` - Dot product similarity between input and label embeddings. Confidence for each label will be in an unbounded range.
+
+The default value is `softmax`, but we recommend using `cosine` as that will be the new default value, Rasa Open Source 3.0 onwards. The value of this option does not affect how confidences are computed for entity predictions in `DIETClassifier` and `TEDPolicy`.
+
+The default auto-configuration is changed to use `constrain_similarities=True` and `model_confidence=cosine` in ML components so that new users start with the recommended configuration.
```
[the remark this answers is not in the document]
dakshvar22: @wochinge Do you have a link? I couldn't find a section specifically for auto config in the docs.

changelog/7616.improvement.md
```diff
@@ -0,0 +1,16 @@
+Added an option `constrain_similarities` which adds sigmoid cross-entropy loss on all similarity values to constrain them to an approximate range in `DotProductLoss`.
```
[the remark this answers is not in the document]
dakshvar22: Added links for each of the ML components where configuration is available for each component.

docs/docs/components.mdx
```diff
@@ -1531,10 +1531,12 @@ However, additional parameters exist that can be adapted.
...
 |                                 |                  | or 'inner'.                                                  |
 +---------------------------------+------------------+--------------------------------------------------------------+
-| loss_type                       | "softmax"        | The type of the loss function, either 'softmax' or 'margin'. |
+| loss_type                       | "cross_entropy"  | The type of the loss function, either 'cross_entropy'        |
+|                                 |                  | or 'margin'.                                                 |
 +---------------------------------+------------------+--------------------------------------------------------------+
-| ranking_length                  | 10               | Number of top actions to normalize scores for loss type      |
-|                                 |                  | 'softmax'. Set to 0 to turn off normalization.               |
+| ranking_length                  | 10               | Number of top intents to normalize scores for. Applicable    |
+|                                 |                  | with loss type 'cross_entropy'. Set to 0 to disable          |
```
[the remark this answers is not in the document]
dakshvar22: Yes, that's correct. Adding that in.

rasa/utils/tensorflow/layers.py
dakshvar22: Adding as an input to the function would mean we pass the value of the class's own instance variable `similarity_type` to its own function. It seemed very weird to do that and moreover, we will have to make sure that we pass the correct value in all parts where `.sim` is called.
I think it's fine that we double normalize in this case because the output would still be correct and refactoring it for `margin` loss seems an overkill as we will get rid of it eventually.
Once we get rid of margin loss we should do a clean up since there will only be one similarity type `inner` during training and different types of similarity during inference which  should be cleaner.
Wdyt?

Ghostvv: maybe then moving normalization to `margin` loss?

dakshvar22: In that case we will have to restrict `cosine` similarity only to `margin` loss. Currently we just raise a warning when `similarity_type=cosine` and `loss_type=cross_entropy/softmax` is used.
I guess we can't hard deprecate the combination in a minor?

Ghostvv: aaah let's leave it then. Could you create an issue about deprecating `margin` loss? Maybe some of the users will chime in

dakshvar22: Done - https://github.com/RasaHQ/rasa/issues/7891

rasa/utils/tensorflow/layers.py
dakshvar22: Ofcourse, don't know why I complicated this op 🙈

tests/nlu/classifiers/test_diet_classifier.py
dakshvar22: The epochs is indeed set to 1 in the parameters of the test. Do you mean to set it somewhere else?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -850,13 +865,23 @@ def _predict_label(
         label_ids = message_sim.argsort()[::-1]
 
         if (
-            self.component_config[LOSS_TYPE] == SOFTMAX
+            self.component_config[LOSS_TYPE] == CROSS_ENTROPY
             and self.component_config[RANKING_LENGTH] > 0
         ):
-            message_sim = train_utils.normalize(
+            message_sim = train_utils.filter_top_k(
```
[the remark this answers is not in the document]
dakshvar22: > filtering was needed only for renormalization. Number of top k prediction is regulated later in lines 896-902
👍 

> I don't think normalization should be applied for `inner`

You mean `similarity_type=inner` or `model_confidence=`inner`. Currently, I am changing the if condition to this - 

```
if (
            self.component_config[LOSS_TYPE] == CROSS_ENTROPY
            and self.component_config[RANKING_LENGTH] > 0
            and self.component_config[SIMILARITY_TYPE] == INNER
            and self.component_config[MODEL_CONFIDENCE] == SOFTMAX
        ):
```
This would ensure that normalization is done only for older settings. In any of the new setting it won't be done.

wochinge:
@dakshvar22 Should this go in the 2.3 milestone?
+ is this model breaking?

dakshvar22:
@wochinge yes it should go in the milestone. It isn't model breaking as you can still load a previously trained model and expect the output to be the same.

rasa/core/policies/ted_policy.py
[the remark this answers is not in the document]
dakshvar22: I agree with vova here. It's code that will be deprecated, lets not spend too much effort in refactoring it.

rasa/nlu/classifiers/diet_classifier.py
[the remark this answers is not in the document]
Ghostvv: oh, did I misread `SIMILARITY_TYPE` and `MODEL_CONFIDENCE`? don't we have a check when loading params that `MODEL_CONFIDENCE] == SOFTMAX` only if other conditions meet? then I think here we can only check for `MODEL_CONFIDENCE] == SOFTMAX`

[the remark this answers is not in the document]
dakshvar22: Do you mean that `model_confidence` should be softmax only if - 
1. `loss_type` is `cross entropy`
2. `similarity_type` is `inner`.

We don't make these checks explicitly right now. I think it makes sense to add. Will do 👍

rasa/core/policies/ted_policy.py
```diff
@@ -340,6 +349,11 @@ def _load_params(self, **kwargs: Dict[Text, Any]) -> None:
         self.config = rasa.utils.train_utils.override_defaults(
             self.defaults, new_config
         )
+
+        rasa.utils.train_utils._check_loss_setting(self.config)
```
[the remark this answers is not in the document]
dakshvar22: Moved all calls to a public method in `rasa.utils.train_utils.`

rasa/utils/tensorflow/layers.py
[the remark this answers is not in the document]
dakshvar22: Changed to `RasaException` for now. Any other inherited exception would have been too Rasa specific.

rasa/utils/tensorflow/layers.py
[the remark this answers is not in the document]
dakshvar22: Changed to `RasaException` for the same reason as above.

tests/nlu/classifiers/test_diet_classifier.py
dakshvar22: Added 👍

rasa/utils/plotting.py
```diff
@@ -133,21 +133,36 @@ def plot_histogram(
     # Wine-ish colour for the confidences of hits.
     # Blue-ish colour for the confidences of misses.
     colors = ["#009292", "#920000"]
-    bins = [0.05 * i for i in range(1, 21)]
+    n_bins = 25
+    max_value = max(
+        [max(hist_data[0], default=0), max(hist_data[1], default=0)], default=0
+    )
```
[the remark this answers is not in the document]
dakshvar22: @Ghostvv I tried another strategy on sara's data, how about these - 
`cosine`(left), `inner`(right)
intent_histogram
intent_histogram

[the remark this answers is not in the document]
dakshvar22: Alternatively, if we feel the bar thickness is distracting we could keep the plotting strategy as before for `softmax` and `cosine` because most users will use those two options and use the above one for `inner`.

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,25 @@ description: |
...
+components - `DIETClassifier`, `ResponseSelector` and `TEDPolicy`. These include:
+1. Configuration option `loss_type=softmax` is now deprecated. Use `loss_type=cross_entropy` instead.
+2. The default loss function (`loss_type=cross_entropy`) adds an optional sigmoid cross-entropy loss of all similarity values to constrain
+them to an approximate range. You can turn on this option by setting `constrain_similarities=True`.
+
+Also, a new option `model_confidence` has been added to each ML component. It affects how model's confidence for each label is computed during inference. It can take three values -
+1. `softmax` - Similarities between input and label embeddings are post-processed with a softmax function, as a result of which confidence for all labels sum up to 1.
+2. `cosine` - Cosine similarity between input label embeddings. Confidence for each label will be in the range `[-1,1]`.
+3. `inner` - Dot product similarity between input and label embeddings. Confidence for each label will be in an unbounded range.
+The default value is `softmax`, but we recommend using `cosine` as that will be the new default value, Rasa Open Source 3.0 onwards.
```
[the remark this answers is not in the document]
dakshvar22: Added below

rasa/utils/plotting.py
```diff
@@ -128,26 +128,59 @@ def plot_histogram(
...
+                [
+                    (binned_data_set != 0).argmax(axis=0)
+                    for binned_data_set in binned_data_sets
+                ]
+            )
+        ]
+        # - bins[0]
+    )
+
+    # max_ylim = max(bins) + bin_width
```
dakshvar22, line 180: Experimental, decide on it based on above discussion.

dakshvar22:
@melindaloubser1 We wanted someone from CSE to take a quick look at the docs part of the PR. Could you please take a look?

rasa/utils/train_utils.py
```diff
@@ -63,14 +67,36 @@ def update_similarity_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
...
-        if config[LOSS_TYPE] == SOFTMAX:
+        if config[LOSS_TYPE] == CROSS_ENTROPY:
             config[SIMILARITY_TYPE] = INNER
         elif config[LOSS_TYPE] == MARGIN:
             config[SIMILARITY_TYPE] = COSINE
 
     return config
 
 
+def update_loss_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
```
wochinge, line 78: what do you think? 
```suggestion
def update_deprecated_loss_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
```

rasa/utils/train_utils.py
```diff
@@ -342,6 +368,94 @@ def override_defaults(
...
+        component_config: Configuration to validate.
+    """
+    _check_loss_setting(component_config)
+    _check_confidence_setting(component_config)
+    _check_similarity_loss_setting(component_config)
+
+
+def _check_confidence_setting(component_config: Dict[Text, Any]) -> None:
+    if component_config[MODEL_CONFIDENCE] == SOFTMAX:
+        rasa.shared.utils.io.raise_warning(
```
wochinge, line 406:
```suggestion
        rasa.shared.utils.io.raise_deprecation_warning(
```

rasa/utils/train_utils.py
```diff
@@ -342,6 +368,94 @@ def override_defaults(
...
+            f"{MODEL_CONFIDENCE} is set to `softmax`. It is recommended "
+            f"to set it to `cosine`. It will be set to `cosine` by default, "
+            f"Rasa Open Source 3.0 onwards.",
```
wochinge, line 409:
```suggestion
            f"Rasa Open Source 3.0.0 onwards.",
```

rasa/utils/train_utils.py
```diff
@@ -342,6 +368,94 @@ def override_defaults(
...
+                f"only with {SIMILARITY_TYPE}={INNER}."
+            )
+
+
+def _check_loss_setting(component_config: Dict[Text, Any]) -> None:
+    if not component_config[CONSTRAIN_SIMILARITIES] and component_config[LOSS_TYPE] in [
+        SOFTMAX,
+        CROSS_ENTROPY,
+    ]:
+        rasa.shared.utils.io.raise_warning(
```
wochinge, line 433:
```suggestion
        rasa.shared.utils.io.raise_deprecation_warning(
```

rasa/utils/train_utils.py
```diff
@@ -342,6 +368,94 @@ def override_defaults(
...
+
+
+def _check_similarity_loss_setting(component_config: Dict[Text, Any]) -> None:
+    if (
+        component_config[SIMILARITY_TYPE] == COSINE
+        and component_config[LOSS_TYPE] == CROSS_ENTROPY
+        or component_config[SIMILARITY_TYPE] == INNER
+        and component_config[LOSS_TYPE] == MARGIN
+    ):
+        rasa.shared.utils.io.raise_warning(
```
wochinge, line 448:
```suggestion
        rasa.shared.utils.io.raise_deprecation_warning(
```

wochinge: Is this a regular warning or a deprecation warning? It's not deprecated it should be `UserWarning`

tests/nlu/classifiers/test_diet_classifier.py
wochinge: nope, all good then

tests/nlu/selectors/test_selectors.py
```diff
@@ -315,3 +323,126 @@ async def test_process_gives_diagnostic_data(trained_response_selector_bot: Path
...
+    mock.normalize.assert_not_called()
+
+
+@pytest.mark.parametrize(
+    "classifier_params", [({LOSS_TYPE: "margin", RANDOM_SEED: 42, EPOCHS: 1})],
+)
+async def test_margin_loss_is_not_normalized(
+    monkeypatch: MonkeyPatch,
+    component_builder: ComponentBuilder,
+    tmpdir: Path,
```
wochinge, line 387: `tmpdir` has a different type.
```suggestion
    tmp_path: Path,
```

rasa/utils/train_utils.py
```diff
@@ -342,6 +368,94 @@ def override_defaults(
...
+
+    Args:
+        component_config: Configuration to validate.
+    """
+    _check_loss_setting(component_config)
+    _check_confidence_setting(component_config)
+    _check_similarity_loss_setting(component_config)
+
+
+def _check_confidence_setting(component_config: Dict[Text, Any]) -> None:
```
wochinge, line 404: do we have tests for this?

changelog/7616.improvement.md
```diff
@@ -0,0 +1,24 @@
...
+With both the above recommendations, users should configure their ML component, e.g. `DIETClassifier`, as -
+```
+- name: DIETClassifier
+  model_confidence: cosine
+  constrain_similarities: True
+  ...
+```
+Once the assistant is re-trained with the above configuration, users should also tune fallback confidence thresholds.
+
+Configuration option `loss_type=softmax` is now deprecated. Use `loss_type=cross_entropy` instead.
```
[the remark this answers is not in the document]
dakshvar22: It is added to the migration guide too. Will add that it will be done Rasa Open Source 3.0 onwards.

docs/docs/components.mdx
```diff
@@ -2814,6 +2836,22 @@ However, additional parameters exist that can be adapted.
 |                                 |                   | Requires `evaluate_on_number_of_examples > 0` and            |
 |                                 |                   | `evaluate_every_number_of_epochs > 0`                        |
 +---------------------------------+-------------------+--------------------------------------------------------------+
+| constrain_similarities          | False             | If `True`, applies sigmoid on all similarity terms and adds  |
+|                                 |                   | it to the loss function to ensure that similarity values are |
+|                                 |                   | approximately bounded. Used only if `loss_type=cross_entropy`|
++---------------------------------+-------------------+--------------------------------------------------------------+
+| model_confidence                | "softmax"         | Affects how model's confidence for each response label       |
+|                                 |                   | is computed. It can take three values -                      |
```
dakshvar22, line 2844:
```suggestion
|                                 |                   | is computed. It can take three values                        |
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,34 @@ description: |
...
+With both the above recommendations, users should configure their ML component, e.g. `DIETClassifier`, as -
+```
+- name: DIETClassifier
+  model_confidence: cosine
+  constrain_similarities: True
+  ...
+```
+Once the assistant is re-trained with the above configuration, users should also tune fallback confidence thresholds.
+
+Configuration option `loss_type=softmax` is also deprecatedin all ML components. Use `loss_type=cross_entropy` instead.
```
dakshvar22, line 39:
```suggestion
```

docs/docs/policies.mdx
```diff
@@ -268,10 +268,12 @@ However, additional parameters exist that can be adapted.
...
 +---------------------------------------+------------------------+--------------------------------------------------------------+
-| loss_type                             | "softmax"              | The type of the loss function, either 'softmax' or 'margin'. |
+| loss_type                             | "cross_entropy"        | The type of the loss function, either 'cross_entropy'        |
+|                                       |                        | or 'margin'.                                                 |
 +---------------------------------------+------------------------+--------------------------------------------------------------+
-| ranking_length                        | 10                     | Number of top actions to normalize scores for loss type      |
-|                                       |                        | 'softmax'. Set to 0 to turn off normalization.               |
+| ranking_length                        | 10                     | Number of top actions to normalize scores for. Applicable    |
+|                                       |                        | only with loss type 'cross_entropy' and 'softmax'            |
```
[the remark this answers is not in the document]
dakshvar22: It isn't the start of a new sentence. Maybe you missed that there is `Applicable` before `only`?

rasa/core/policies/ted_policy.py
```diff
@@ -212,10 +216,10 @@ class TEDPolicy(Policy):
         NUM_NEG: 20,
         # Type of similarity measure to use, either 'auto' or 'cosine' or 'inner'.
         SIMILARITY_TYPE: AUTO,
-        # The type of the loss function, either 'softmax' or 'margin'.
-        LOSS_TYPE: SOFTMAX,
-        # Number of top actions to normalize scores for loss type 'softmax'.
-        # Set to 0 to turn off normalization.
+        # The type of the loss function, either 'cross_entropy' or 'margin'.
+        LOSS_TYPE: CROSS_ENTROPY,
```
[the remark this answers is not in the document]
dakshvar22: I don't think so. We do accept `softmax` also and change it to `cross_entropy` automatically. It's the same as how other deprecated parameters are handled as of now.

docs/docs/policies.mdx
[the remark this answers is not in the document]
wochinge: Fair. All others start with a capital letter, too, don't they?

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,34 @@ description: |
...
 how you can migrate from one version to another.
 
+## Rasa 2.2 to Rasa 2.3
+
+### Machine Learning Components
+
+Few changes have been made to the loss function inside machine learning (ML)
+components - `DIETClassifier`, `ResponseSelector` and `TEDPolicy`. These include:
+1. Configuration option `loss_type=softmax` is now deprecated and will be removed in Rasa Open Source 3.0. Use `loss_type=cross_entropy` instead.
+2. The default loss function (`loss_type=cross_entropy`) adds an optional sigmoid cross-entropy loss of all similarity values to constrain
```
Ghostvv, line 20:
```suggestion
2. The default loss function (`loss_type=cross_entropy`) can add an optional sigmoid cross-entropy loss of all similarity values to constrain
```

rasa/utils/tensorflow/layers.py
```diff
@@ -589,8 +591,15 @@ def __init__(
         self.use_max_sim_neg = use_max_sim_neg
         self.neg_lambda = neg_lambda
         self.scale_loss = scale_loss
-        self.parallel_iterations = parallel_iterations
         self.same_sampling = same_sampling
+        self.constrain_similarities = constrain_similarities
+        self.model_confidence = model_confidence
+        self.similarity_type = similarity_type
+        if self.similarity_type and self.similarity_type not in {COSINE, INNER}:
```
Ghostvv, line 598: what happens `if not self.similarity_type`?

rasa/utils/tensorflow/layers.py
```diff
@@ -806,7 +840,7 @@ def _loss_margin(
 
         return loss
 
-    def _loss_softmax(
+    def _loss_cross_entropy(
```
Ghostvv, line 843: could you please separate this methods into helpers?

rasa/utils/plotting.py
```diff
@@ -133,21 +134,37 @@ def plot_histogram(
...
 
-    min_ylim = bins[
-        min(
-            [
-                (binned_data_set != 0).argmax(axis=0)
-                for binned_data_set in binned_data_sets
-            ]
-        )
-    ]
+    min_ylim = (
```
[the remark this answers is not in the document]
dakshvar22: Hmm, then the bin calculation will have to be placed outside of this function. I'd do it in a separate PR I think.

rasa/utils/tensorflow/layers.py
```diff
@@ -589,8 +591,15 @@ def __init__(
         self.use_max_sim_neg = use_max_sim_neg
         self.neg_lambda = neg_lambda
         self.scale_loss = scale_loss
-        self.parallel_iterations = parallel_iterations
         self.same_sampling = same_sampling
+        self.constrain_similarities = constrain_similarities
+        self.model_confidence = model_confidence
+        self.similarity_type = similarity_type
+        if self.similarity_type and self.similarity_type not in {COSINE, INNER}:
+            raise RasaException(
```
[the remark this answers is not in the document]
dakshvar22: I don't think it should be `InvalidConfigException`. As vova had mentioned in the previous comment, this code can be used outside of regular Rasa usage where this TF layer is created without a usual rasa config. `RasaException` seems fine because of that.

changelog/7616.improvement.md
```diff
@@ -0,0 +1,24 @@
...
+By default, the parameter is set to `False` to preserve the old behaviour, but users are encouraged to set it to `True` and re-train their assistants as it will be set to `True` by default, Rasa Open Source 3.0 onwards.
+
+Parameter `model_confidence` affects how model's confidence for each label is computed during inference. It can take three values -
+1. `softmax` - Similarities between input and label embeddings are post-processed with a softmax function, as a result of which confidence for all labels sum up to 1.
+2. `cosine` - Cosine similarity between input label embeddings. Confidence for each label will be in the range `[-1,1]`.
+3. `inner` - Dot product similarity between input and label embeddings. Confidence for each label will be in an unbounded range.
+
+Setting `model_confidence=cosine` should help users tune the fallback thresholds of their assistant better. The default value is `softmax` to preserve the old behaviour, but we recommend using `cosine` as that will be the new default value, Rasa Open Source 3.0 onwards. The value of this option does not affect how confidences are computed for entity predictions in `DIETClassifier` and `TEDPolicy`.
```
melindaloubser1, line 11:
```suggestion
Setting `model_confidence=cosine` should help users tune the fallback thresholds of their assistant better. The default value is `softmax` to preserve the old behaviour, but we recommend using `cosine` as that will be the new default value from Rasa Open Source 3.0 onwards. The value of this option does not affect how confidences are computed for entity predictions in `DIETClassifier` and `TEDPolicy`.
```

rasa/utils/tensorflow/layers.py
```diff
@@ -851,12 +916,12 @@ def _chosen_loss(self) -> Callable:
 
         if self.loss_type == MARGIN:
             return self._loss_margin
-        elif self.loss_type == SOFTMAX:
-            return self._loss_softmax
+        elif self.loss_type == CROSS_ENTROPY:
+            return self._loss_cross_entropy
         else:
-            raise ValueError(
+            raise RasaException(
```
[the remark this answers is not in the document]
dakshvar22: Same as above. All these classes can be used outside of regular rasa usage.

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,34 @@ description: |
 This page contains information about changes between major versions and
 how you can migrate from one version to another.
 
+## Rasa 2.2 to Rasa 2.3
+
+### Machine Learning Components
+
+Few changes have been made to the loss function inside machine learning (ML)
```
melindaloubser1, line 17:
```suggestion
A few changes have been made to the loss function inside machine learning (ML)
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,34 @@ description: |
...
+components - `DIETClassifier`, `ResponseSelector` and `TEDPolicy`. These include:
```
melindaloubser1, line 18:
```suggestion
components `DIETClassifier`, `ResponseSelector` and `TEDPolicy`. These include:
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,34 @@ description: |
...
+1. Configuration option `loss_type=softmax` is now deprecated and will be removed in Rasa Open Source 3.0. Use `loss_type=cross_entropy` instead.
+2. The default loss function (`loss_type=cross_entropy`) adds an optional sigmoid cross-entropy loss of all similarity values to constrain
+them to an approximate range. You can turn on this option by setting `constrain_similarities=True`. This should help the models to perform better on real world test sets.
+
+Also, a new option `model_confidence` has been added to each ML component. It affects how model's confidence for each label is computed during inference. It can take three values -
```
melindaloubser1, line 23:
```suggestion
Also, a new option `model_confidence` has been added to each ML component. It affects how model's confidence for each label is computed during inference. It can take one of three values:
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,34 @@ description: |
...
+1. `softmax` - Similarities between input and label embeddings are post-processed with a softmax function, as a result of which confidence for all labels sum up to 1.
+2. `cosine` - Cosine similarity between input label embeddings. Confidence for each label will be in the range `[-1,1]`.
+3. `inner` - Dot product similarity between input and label embeddings. Confidence for each label will be in an unbounded range.
+The default value is `softmax`, but we recommend using `cosine` as that will be the new default value, Rasa Open Source 3.0 onwards.
```
melindaloubser1, line 27:
```suggestion
The default value is `softmax`, but we recommend using `cosine` as that will be the new default value from Rasa Open Source 3.0 onwards.
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,34 @@ description: |
...
+The value of this option does not affect how confidences are computed for entity predictions in `DIETClassifier` and `TEDPolicy`.
+
+With both the above recommendations, users should configure their ML component, e.g. `DIETClassifier`, as -
```
melindaloubser1, line 30:
```suggestion
With both the above recommendations, users should configure their ML component, e.g. `DIETClassifier`, as:
```

melindaloubser1 (changes requested):

docs/docs/policies.mdx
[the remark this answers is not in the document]
dakshvar22: Not sure what you mean. I see a capital letter being used only when there is a `.` in other parts too.

rasa/utils/tensorflow/layers.py
dakshvar22: Good point, used it as it was before. Changed it now to `if not self.similarity_type or self.similarity_type not in {COSINE, INNER}`

dakshvar22: Good point, used it as it was before. Changed it now to `if not self.similarity_type or self.similarity_type not in {COSINE, INNER}`

Ghostvv: I suggest to remove `Optional` from input type and don't check for `None`. We cannot check every argument

rasa/utils/tensorflow/layers.py
dakshvar22: Done

Ghostvv (approved):
ML part looks good to me

rasa/utils/train_utils.py
dakshvar22: Changed 👍

rasa/utils/train_utils.py
dakshvar22: It's a regular warning, changing it to `UserWarning`

dakshvar22: Do you have a good pointer on when to use `UserWarning` v/s `FutureWarning`. I seem to be missing it always. My understanding was that if we are describing a behaviour that is expected to take place in the future, it should be `FutureWarning`.

rasa/utils/train_utils.py
dakshvar22: This is also not a deprecation but rather a behaviour expected in future.

rasa/utils/train_utils.py
dakshvar22: Added for exceptions raised.

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,33 @@ description: |
...
+
+A few changes have been made to the loss function inside machine learning (ML)
+components `DIETClassifier`, `ResponseSelector` and `TEDPolicy`. These include:
+1. Configuration option `loss_type=softmax` is now deprecated and will be removed in Rasa Open Source 3.0.0. Use `loss_type=cross_entropy` instead.
+2. The default loss function (`loss_type=cross_entropy`) can add an optional sigmoid cross-entropy loss of all similarity values to constrain
+them to an approximate range. You can turn on this option by setting `constrain_similarities=True`. This should help the models to perform better on real world test sets.
+
+Also, a new option `model_confidence` has been added to each ML component. It affects how model's confidence for each label is computed during inference. It can take one of three values:
+1. `softmax` - Similarities between input and label embeddings are post-processed with a softmax function, as a result of which confidence for all labels sum up to 1.
+2. `cosine` - Cosine similarity between input label embeddings. Confidence for each label will be in the range `[-1,1]`.
```
dakshvar22, line 25:
```suggestion
2. `cosine` - Cosine similarity between input and label embeddings. Confidence for each label will be in the range `[-1,1]`.
```

rasa/utils/tensorflow/layers.py
```diff
@@ -845,18 +879,69 @@ def _loss_softmax(
...
+        # create label_ids for softmax
+        softmax_label_ids = tf.zeros_like(softmax_logits[..., 0], tf.int32)
+        softmax_loss = tf.nn.sparse_softmax_cross_entropy_with_logits(
+            labels=softmax_label_ids, logits=softmax_logits
+        )
+        return softmax_loss
+
     @property
     def _chosen_loss(self) -> Callable:
         """Use loss depending on given option."""
```
dakshvar22, line 936:
```suggestion
```

docs/docs/policies.mdx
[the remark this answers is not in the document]
wochinge: sorry, double checked it and it was some confusion of mine with the Git diff

rasa/utils/plotting.py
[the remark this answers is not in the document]
wochinge: Ok. I agree that the code shouldn't have gone in untested in the  first place. Can you please make sure we don't loose track of this?

rasa/utils/tensorflow/layers.py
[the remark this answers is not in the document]
wochinge: In that case I think you should still create a subclass for `RasaException` which is more specific, e.g. something like a `LayerConfigException`. `raise RasaException` is basically like a `raise Exception` and too broad

rasa/utils/train_utils.py
wochinge: Anything which is just incorrect / not suggested behavior: `UserWarning`
If something is deprecated and will be removed in the future: `FutureWarning`

Does that make sense?

> a behaviour that is expected to take place in the future,

What do you mean by that?

rasa/utils/plotting.py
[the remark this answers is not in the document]
dakshvar22: Created an issue https://github.com/RasaHQ/rasa/issues/7912

rasa/utils/train_utils.py
wochinge: Mhm, I think I get your troubles now based on [this](https://github.com/RasaHQ/rasa/pull/7616#discussion_r572203976) comment.

You mean that the parameter isn't deprecated and won't be deprecated, but the default will change, right?

wochinge: then I'd say it's a regular `UserWarning`

dakshvar22: > mean that the parameter isn't deprecated and won't be deprecated, but the default will change, right?
Yes, alright then all of these are `UserWarning`. Thanks 👍

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,33 @@ description: |
...
+
+### Machine Learning Components
+
+A few changes have been made to the loss function inside machine learning (ML)
+components `DIETClassifier`, `ResponseSelector` and `TEDPolicy`. These include:
+1. Configuration option `loss_type=softmax` is now deprecated and will be removed in Rasa Open Source 3.0.0. Use `loss_type=cross_entropy` instead.
+2. The default loss function (`loss_type=cross_entropy`) can add an optional sigmoid cross-entropy loss of all similarity values to constrain
+them to an approximate range. You can turn on this option by setting `constrain_similarities=True`. This should help the models to perform better on real world test sets.
+
+Also, a new option `model_confidence` has been added to each ML component. It affects how model's confidence for each label is computed during inference. It can take one of three values:
```
melindaloubser1, line 23:
```suggestion
Also, a new option `model_confidence` has been added to each ML component. It affects how a model's confidence for each label is computed during inference. It can take one of three values:
```
Either that or `affects how model confidence for each label...`

melindaloubser1 (approved):

rasa/utils/tensorflow/layers.py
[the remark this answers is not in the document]
dakshvar22: Added `TFLayerConfigException`.

rasa/utils/tensorflow/layers.py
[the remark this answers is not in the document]
dakshvar22: Added `TFLayerConfigException`

dakshvar22:
@wochinge I've added `TFLayerConfigException` instead of `RasaException`. So, the PR should be good to merge. Can you flip your review? I can't merge otherwise :)

wochinge:
> So, the PR should be good to merge. Can you flip your review? I can't merge otherwise :)

Sure, let me just give it a final glance 👍🏻 Sorry, had a busy day and only had time for the review now.

rasa/shared/exceptions.py
```diff
@@ -111,3 +111,7 @@ class ConnectionException(RasaException):
     It's used by our broker and tracker store classes, when
     they can't connect to services like postgres, dynamoDB, mongo.
     """
+
+
+class TFLayerConfigException(RasaException):
```
wochinge, line 116: @dakshvar22 Can we move this to `rasa.utils.tensorflow.layers`? There is no need to have this in the shared package.

rasa/utils/train_utils.py
```diff
@@ -342,6 +368,94 @@ def override_defaults(
     return config
 
 
+def update_confidence_type(component_config: Dict[Text, Any]) -> Dict[Text, Any]:
```
wochinge, line 371: I believe this is not covered by a test. Could you please add one?

rasa/utils/train_utils.py
```diff
@@ -342,6 +368,94 @@ def override_defaults(
...
+    """Set model confidence to cosine if margin loss is used.
+
+    Args:
+        component_config: model configuration
+
+    Returns:
+        updated model configuration
+    """
+    # TODO: Remove this once model_confidence is set to cosine by default.
```
wochinge, line 380: Shouldn't we still warn about invalid config in this case? We should maybe replace it with a `raise`  then.

wochinge (approved):
Please check the 2 todos. I'm good otherwise 👍🏻

rasa/shared/exceptions.py
dakshvar22: Should we have an `exceptions.py` inside `rasa.utils.tensorflow` and place it inside that?

wochinge: up to you 👍🏻

rasa/utils/train_utils.py
dakshvar22: This not invalid config. I expect this to happen when users who had trained their assistant with margin loss in the past update to the new version and just re-train without changing anything in the config. Since the default value of `model_confidence` is set to `softmax`(as we believe most users use `cross_entropy` loss), this is an attempt to update the value of `model_confidence` for them automatically. If users follow the advice on the changelog, they will anyways change it to `cosine` themselves. ;)

rasa/utils/train_utils.py
dakshvar22: Added a test

Issue #6939, opened 2020-10-06
add possibility to define optional slots for forms


Ghostvv:
we could add `optional` key when define slot mappings in the domain, so that slot_mappings for extra slots could be defined, but the form shouldn't ask these slots

domsammut:
Agreed. This is a pretty big regression from the forms in 1.0. With the move to yaml structure, this forces you to move into python and add custom code to perform slot mappings for anything that is either optional or conditional. Previously you could map intents/entities/text to any slots regardless of if they were required, optional or conditional:

```python
def slot_mappings(self) -> Dict[Text, Union[Dict, List[Dict]]]:
    return {
        'test_slot': [
            self.from_intent(intent='test_intent', value=True)
            self.from_entity(entity='test_entity', value=True),
            self.from_text()
        ]
    }
```

rgstephens:
@m-vdb Where was this feature added? I haven't noticed anything new that addresses the comment from @domsammut or the original idea of an `optional` keyword in the domain.

rgstephens:
Optional slots are a common use case and request from users. It would be nice to specify a condition associated with a slot to save writing python in simple cases.

m-vdb:
what part of your use case isn't possible today with the [current Global Slot mappings functionality](https://rasa.com/docs/rasa/domain/#slot-mappings)?

totalhack:
Hey there, I am newer to Rasa and was curious about the same thing. I took a look at the docs @m-vdb linked to but it's unclear how it would be accomplished with that. Perhaps you could provide a simple example of how you think optional slots are possible with the functionality there?

rgstephens:
@totalhack You use the `required_slots` method which is documented under [dynamic form behavior](https://rasa.com/docs/rasa/forms/#dynamic-form-behavior).

totalhack:
Thanks @rgstephens, that's helpful. 

So it seems there is no way to do this directly in the yaml files, even for simple cases, correct?

Also in those examples, aren't you really just dynamically adding another *required* slot, rather than truly having an optional slot? Is there some other way to do the latter that I missed? Basically have a form pick up and fill a slot if the user happened to also provide that information?

rgstephens:
> So it seems there is no way to do this directly in the yaml files, even for simple cases, correct?

Correct, that's what this issue is for and why I'd like it to stay open.

> Also in those examples, aren't you really just dynamically adding another required slot, rather than truly having an optional slot?

[...]

PR #7625, opened 2020-12-22
Fix #7624: only fill slots where auto_fill is true in extract_other_s…

### LINKED ISSUE #7624: Auto-fill not respected by FormAction
**Rasa version**: 2.2.2

**Python version**: 3.6

**OS**: Mac

**Issue**:

*   The form action `extract_other_slots` is not respecting the `auto_fill` property of the slots in the domain file. So they are being overwritten.

There is a bug report on the forum here and also here: [https://forum.rasa.com/t/slots-overwritten-in-forms-with-duckling-entities/38197/5](https://forum.rasa.com/t/slots-overwritten-in-forms-with-duckling-entities/38197/5)

I was able to resolve the issue through a modification of this function. Will open a pull request.

kearnsw:
**Proposed changes**:
Check auto_fill property before extracting slot in `extract_other_slots`, only fill the slot if auto_fill = True (default value)

- [N/A] updated the documentation


sara-tagger:

wochinge (commented):
@kearnsw Thanks for your contribution. I'm not sure if I understand the issue completely: why would you specify a slot mapping in the domain if this slot should not be filled by the form?

ma7555:
> @kearnsw Thanks for your contribution. I'm not sure if I understand the issue completely: why would you specify a slot mapping in the domain if this slot should not be filled by the form?

@kearnsw 
please see this https://forum.rasa.com/t/slots-overwritten-in-forms-with-duckling-entities/38197/5

kearnsw:
Thanks for linking that @ma7555. @wochinge, the thread linked above explains the issue. It's possible I'm going about the solution wrong, but I was able to solve that user's problem by the changes in this PR.

kearnsw:
@wochinge, I came across this issue (#6939) as well which is related to the question of when you would want to define a slot mapping, but not fill the slot. Although, I'm not sure that this PR will necessarily resolve that as I didn't test for that specifically.

wochinge:
@kearnsw To me it seems they should rather validate the [extracted slots](https://rasa.com/docs/rasa/forms#validating-form-input) or making the slot mappings more narrow (e.g. excluding an intent). What do you think?

kearnsw:
> @kearnsw To me it seems they should rather validate the [extracted slots](https://rasa.com/docs/rasa/forms#validating-form-input) or making the slot mappings more narrow (e.g. excluding an intent). What do you think? 

@wochinge, Validation may not work if there is no pattern that the slot can be validated against and/or two slots have overlapping entity types. 

Narrowing the scope of the mapping is a possibility in some situations, however often intents are not easily distinguished between slots so are combined into a single `inform` intent. Also, as we move toward e2e models this will no longer be an option.

wochinge:
> Validation may not work if there is no pattern that the slot can be validated against and/or two slots have overlapping entity types.

Makes sense - but how does this relate to auto slot filling? Sorry for not getting it 🙈 
I think I'm not understanding the motivation for this correctly: Forms are meant to fill slots based on the users input. If a slot mapping is wide / shouldn't apply in certain conditions then a custom slot mapping should be used.

wochinge:
@kearnsw Please feel free to re-open if you have time to continue

PR #7656, opened 2020-12-31
fix migration guide so it's in line with formbot example

amn41:
**Proposed changes**:
- fix typo


wochinge (approved):
🚀

Issue #5986, opened 2020-06-10
rasa test should use `TrainingDataImporter`


wochinge:
**Current State**
When running `rasa train` it uses the `TrainingDataImporter` interface to load and provide data to the underlying functions. 
`rasa test` doesn't use this interface and calls [this](https://github.com/RasaHQ/rasa/blob/73c16d1a0e4ae5bb0a9d241031dbaa93042e48bc/rasa/cli/test.py#L70) directly. 

**Solution**
Whenever we import training data we should use the `TrainingDataImporter` interface. 

**Todo**:
- [ ] Make sure `rasa test` uses the `TrainingDataImporter` (I believe [here](https://github.com/RasaHQ/rasa/blob/14fae60ef0b49fdaae553d58785f272eb9cd2d1e/rasa/core/test.py#L328) and [here](https://github.com/RasaHQ/rasa/blob/7a634f0e9c1fd001ee4fcd9460b875c4802036f7/rasa/nlu/test.py#L1485)

wochinge:
@Imod7 I'll re-open this as the Core part is still missing.

PR #7674, opened 2021-01-03
Implementing the use of TrainingDataImporter in rasa test 

Imod7:
Implementing the 2nd Part of the changes described in [Issue 5986](https://github.com/RasaHQ/rasa/issues/5986) related to `run_evaluation`.
_The current PR is a continuation of this one [#7539](https://github.com/RasaHQ/rasa/pull/7539) - in the 7539 I had issues with the CLA certificate and then squash and rebase became very difficult because of merged commits so I decided to create a new one, the current one._

**Proposed changes**:
- Loading `test data` using `TrainingDataImporter` in `run_evaluation` function.
- Changed the `Optional` parameters in `load_from_dict` of TrainingDataImporter to make it work with the data we have in `run_evaluation`.
- Instead of having 1 function, the `run_nlu_test` now we have 2 functions described below : 
   1. `run_nlu_test` that runs `run_nlu_test_async` in the event loop.
   2. `run_nlu_test_async`
- Passing directly the specific arguments to `run_nlu_test_async` instead of `argparse.Namespace` thus making it more reusable.
- Made async the following functions : 
  1. `run_evaluation`
  2. `compare_nlu`
  3. `test_nlu`
  4. `compare_nlu_models`
  5. `train_nlu_async` 
  6. `test_nlu_comparison`
- Making `train_nlu_async` function public so we can call it from other modules.

### DIFF
--- a/rasa/cli/test.py
+++ b/rasa/cli/test.py
@@ -1,7 +1,7 @@
 import argparse
 import logging
 import os
-from typing import List
+from typing import List, Optional, Text, Dict, Union, Any
 
 from rasa.cli import SubParsersAction
 import rasa.shared.data
@@ -20,6 +20,7 @@ from rasa.shared.constants import (
 from rasa.core.test import FAILED_STORIES_FILE
 import rasa.shared.utils.validation as validation_utils
 import rasa.cli.utils
+import rasa.utils.common
 
 logger = logging.getLogger(__name__)
 
@@ -115,29 +116,54 @@ def run_core_test(args: argparse.Namespace) -> None:
     )
 
 
-def run_nlu_test(args: argparse.Namespace) -> None:
-    """Run NLU tests."""
+async def run_nlu_test_async(
+    config: Optional[Union[Text, List[Text]]],
+    data_path: Text,
+    models_path: Text,
+    output_dir: Text,
+    cross_validation: bool,
+    percentages: List[int],
+    runs: int,
+    no_errors: bool,
+    all_args: Dict[Text, Any],
+) -> None:
+    """Runs NLU tests.
+
+    Args:
+        all_args: all arguments gathered in a Dict so we can pass it as one argument
+                  to other functions.
+        config: it refers to the model configuration file. It can be a single file or
+                a list of multiple files or a folder with multiple config files inside.
+        data_path: path for the nlu data.
+        models_path: path to a trained Rasa model.
+        output_dir: output path for any files created during the evaluation.
+        cross_validation: indicates if it should test the model using cross validation
+                          or not.
+        percentages: defines the exclusion percentage of the training data.
+        runs: number of comparison runs to make.
+        no_errors: indicates if incorrect predictions should be written to a file
+                   or not.
+    """
     from rasa.test import compare_nlu_models, perform_nlu_cross_validation, test_nlu
 
-    nlu_data = rasa.cli.utils.get_validated_path(args.nlu, "nlu", DEFAULT_DATA_PATH)
+    nlu_data = rasa.cli.utils.get_validated_path(data_path, "nlu", DEFAULT_DATA_PATH)
     nlu_data = rasa.shared.data.get_nlu_directory(nlu_data)
-    output = args.out or DEFAULT_RESULTS_PATH
-    args.errors = not args.no_errors
-
+    output = output_dir or DEFAULT_RESULTS_PATH
+    all_args["errors"] = not no_errors
     rasa.shared.utils.io.create_directory(output)
 
-    if args.config is not None and len(args.config) == 1:
-        args.config = os.path.abspath(args.config[0])
-        if os.path.isdir(args.config):
-            args.config = rasa.shared.utils.io.list_files(args.config)
+    if config is not None and len(config) == 1:
+        config = os.path.abspath(config[0])
+        if os.path.isdir(config):
+            config = rasa.shared.utils.io.list_files(config)
 
-    if isinstance(args.config, list):
+    if isinstance(config, list):
         logger.info(
             "Multiple configuration files specified, running nlu comparison mode."
         )
 
         config_files = []
-        for file in args.config:
+        for file in config:
             try:
                 validation_utils.validate_yaml_schema(
                     rasa.shared.utils.io.read_file(file), CONFIG_SCHEMA_FILE,
@@ -148,26 +174,46 @@ def run_nlu_test(args: argparse.Namespace) -> None:
                     f"Ignoring file '{file}' as it is not a valid config file."
                 )
                 continue
-
-        compare_nlu_models(
+        await compare_nlu_models(
             configs=config_files,
             nlu=nlu_data,
             output=output,
-            runs=args.runs,
-            exclusion_percentages=args.percentages,
+            runs=runs,
+            exclusion_percentages=percentages,
         )
-    elif args.cross_validation:
+    elif cross_validation:
         logger.info("Test model using cross validation.")
         config = rasa.cli.utils.get_validated_path(
-            args.config, "config", DEFAULT_CONFIG_PATH
+            config, "config", DEFAULT_CONFIG_PATH
         )
-        perform_nlu_cross_validation(config, nlu_data, output, vars(args))
+        perform_nlu_cross_validation(config, nlu_data, output, all_args)
     else:
         model_path = rasa.cli.utils.get_validated_path(
-            args.model, "model", DEFAULT_MODELS_PATH
+            models_path, "model", DEFAULT_MODELS_PATH
         )
 
-        test_nlu(model_path, nlu_data, output, vars(args))
+        await test_nlu(model_path, nlu_data, output, all_args)
+
+
+def run_nlu_test(args: argparse.Namespace) -> None:
+    """Runs NLU tests.
+
+    Args:
+        args: the parsed CLI arguments for 'rasa test nlu'.
+    """
+    rasa.utils.common.run_in_loop(
+        run_nlu_test_async(
+            args.config,
+            args.nlu,
+            args.model,
+            args.out,
+            args.cross_validation,
+            args.percentages,
+            args.runs,
+            args.no_errors,
+            vars(args),
+        )
+    )
 
 
 def test(args: argparse.Namespace):
--- a/rasa/test.py
+++ b/rasa/test.py
@@ -101,7 +101,9 @@ def test(
         additional_arguments = {}
 
     test_core(model, stories, output, additional_arguments)
-    test_nlu(model, nlu_data, output, additional_arguments)
+    rasa.utils.common.run_in_loop(
+        test_nlu(model, nlu_data, output, additional_arguments)
+    )
 
 
 def test_core(
@@ -110,6 +112,7 @@ def test_core(
     output: Text = DEFAULT_RESULTS_PATH,
     additional_arguments: Optional[Dict] = None,
 ) -> None:
+    """Tests a trained Core model against a set of test stories."""
     import rasa.model
     from rasa.shared.nlu.interpreter import RegexInterpreter
     from rasa.core.agent import Agent
@@ -154,12 +157,13 @@ def test_core(
     rasa.utils.common.run_in_loop(test(stories, _agent, out_directory=output, **kwargs))
 
 
-def test_nlu(
+async def test_nlu(
     model: Optional[Text],
     nlu_data: Optional[Text],
     output_directory: Text = DEFAULT_RESULTS_PATH,
     additional_arguments: Optional[Dict] = None,
 ):
+    """Tests the NLU Model."""
     from rasa.nlu.test import run_evaluation
     from rasa.model import get_model
 
@@ -180,7 +184,9 @@ def test_nlu(
         kwargs = rasa.shared.utils.common.minimal_kwargs(
             additional_arguments, run_evaluation, ["data_path", "model"]
         )
-        run_evaluation(nlu_data, nlu_model, output_directory=output_directory, **kwargs)
+        await run_evaluation(
+            nlu_data, nlu_model, output_directory=output_directory, **kwargs
+        )
     else:
         rasa.shared.utils.cli.print_error(
             "Could not find any model. Use 'rasa train nlu' to train a "
@@ -188,7 +194,7 @@ def test_nlu(
         )
 
 
-def compare_nlu_models(
+async def compare_nlu_models(
     configs: List[Text],
     nlu: Text,
     output: Text,
@@ -214,7 +220,7 @@ def compare_nlu_models(
         model_name: [[] for _ in range(runs)] for model_name in model_names
     }
 
-    training_examples_per_run = compare_nlu(
+    training_examples_per_run = await compare_nlu(
         configs,
         data,
         exclusion_percentages,
--- a/rasa/train.py
+++ b/rasa/train.py
@@ -698,7 +698,7 @@ def train_nlu(
 
     """
     return rasa.utils.common.run_in_loop(
-        _train_nlu_async(
+        train_nlu_async(
             config,
             nlu_data,
             output,
@@ -713,7 +713,7 @@ def train_nlu(
     )
 
 
-async def _train_nlu_async(
+async def train_nlu_async(
     config: Text,
     nlu_data: Text,
     output: Text,
@@ -725,6 +725,7 @@ async def _train_nlu_async(
     model_to_finetune: Optional[Text] = None,
     finetuning_epoch_fraction: float = 1.0,
 ) -> Optional[Text]:
+    """Trains an NLU model asynchronously."""
     if not nlu_data:
         rasa.shared.utils.cli.print_error(
             "No NLU data given. Please provide NLU data in order to train "
--- a/rasa/nlu/test.py
+++ b/rasa/nlu/test.py
@@ -56,6 +56,7 @@ from rasa.nlu.model import Interpreter, Trainer, TrainingData
 from rasa.nlu.components import Component
 from rasa.nlu.tokenizers.tokenizer import Token
 from rasa.utils.tensorflow.constants import ENTITY_RECOGNITION
+from rasa.shared.importers.importer import TrainingDataImporter
 
 logger = logging.getLogger(__name__)
 
@@ -1415,7 +1416,7 @@ def remove_pretrained_extractors(pipeline: List[Component]) -> List[Component]:
     return pipeline
 
 
-def run_evaluation(
+async def run_evaluation(
     data_path: Text,
     model_path: Text,
     output_directory: Optional[Text] = None,
@@ -1448,9 +1449,10 @@ def run_evaluation(
     interpreter = Interpreter.load(model_path, component_builder)
 
     interpreter.pipeline = remove_pretrained_extractors(interpreter.pipeline)
-    test_data = rasa.shared.nlu.training_data.loading.load_data(
-        data_path, interpreter.model_metadata.language
+    test_data_importer = TrainingDataImporter.load_from_dict(
+        training_data_paths=[data_path]
     )
+    test_data = await test_data_importer.get_nlu_data()
 
     result: Dict[Text, Optional[Dict]] = {
         "intent_evaluation": None,
@@ -1822,7 +1824,7 @@ def compute_metrics(
     )
 
 
-def compare_nlu(
+async def compare_nlu(
     configs: List[Text],
     data: TrainingData,
     exclusion_percentages: List[int],
@@ -1850,7 +1852,7 @@ def compare_nlu(
     Returns: training examples per run
     """
 
-    from rasa.train import train_nlu
+    from rasa.train import train_nlu_async
 
     training_examples_per_run = []
 
@@ -1895,7 +1897,7 @@ def compare_nlu(
                 )
 
                 try:
-                    model_path = train_nlu(
+                    model_path = await train_nlu_async(
                         nlu_config,
                         train_split_path,
                         model_output_path,
@@ -1911,7 +1913,7 @@ def compare_nlu(
                 model_path = os.path.join(get_model(model_path), "nlu")
 
                 output_path = os.path.join(model_output_path, f"{model_name}_report")
-                result = run_evaluation(
+                result = await run_evaluation(
                     test_path, model_path, output_directory=output_path, errors=True
                 )
 
--- a/rasa/server.py
+++ b/rasa/server.py
@@ -1151,7 +1151,7 @@ def create_app(
         model_directory = eval_agent.model_directory
         _, nlu_model = model.get_model_subdirectories(model_directory)
 
-        return run_evaluation(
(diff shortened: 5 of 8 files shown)


rasa/train.py
```diff
@@ -697,19 +697,17 @@ def train_nlu(
...
-            output,
-            train_path,
-            fixed_model_name,
-            persist_nlu_training_data,
-            additional_arguments,
-            domain=domain,
-            model_to_finetune=model_to_finetune,
-            finetuning_epoch_fraction=finetuning_epoch_fraction,
-        )
+    return await _train_nlu_async(
```
wochinge, line 700: why was this change required?

rasa/cli/test.py
```diff
@@ -167,7 +168,16 @@ def run_nlu_test(args: argparse.Namespace) -> None:
             args.model, "model", DEFAULT_MODELS_PATH
         )
 
-        test_nlu(model_path, nlu_data, output, vars(args))
+        await test_nlu(model_path, nlu_data, output, vars(args))
+
+
+def run_nlu_test(args: argparse.Namespace) -> None:
```
wochinge, line 174: As discussed in the other PR, I think it would be better to pass the arguments directly to `run_nlu_test_async` to make `run_nlu_test_async` more reusable. Our command-line parser library `argparse` is the only thing which encapsulates arguments into the `arparse.Namespace` type, while other code would rather pass the arguments directly.

What do you think? 

That's a short and not complete example of what I mean:

```python
def run_nlu_test(args: Argparse) -> None:
   rasa.utils.common.run_in_loop(run_nlu_test_async(args.nlu, args.out, args.config, ...))

async def run_nlu_test_async(data_path: Optional[Text], output_dir: Optional[Text], config: Optional[Text], ...) -> None:
    # implementation
```

wochinge (commented):
Great iteration! 💯 Left a few comments

wochinge:
@Imod7 I don't understand this part in your proposed changes, what do you mean?

> Making async a list of functions and then adding the corresponding await statements when they are called.

rasa/train.py
Imod7: Because I had errors of type `RuntimeError: this event loop is already running.`. 

I just checked it again and for example for these tests  : 
```
FAILED tests/test_train.py::test_train_nlu_temp_files - RuntimeError: this event loop is already running.
FAILED tests/test_train.py::test_train_nlu_wrong_format_error_message - RuntimeError: this event loop is alre...
FAILED tests/test_train.py::test_train_nlu_with_responses_no_domain_warns - RuntimeError: this event loop is ...
FAILED tests/test_train.py::test_train_nlu_with_responses_and_domain_no_warns - RuntimeError: this event loop...
FAILED tests/test_train.py::test_train_nlu_no_nlu_file_error_message - RuntimeError: this event loop is alrea...
FAILED tests/test_train.py::test_train_nlu_autoconfig - RuntimeError: this event loop is already running.
```
I get the above errors when I have the `run_in_loop`. When I change to `await` the tests pass.

wochinge: Mhm, interesting 🤔  Why did this error happen? What's the stack trace? Could it be that [this](https://github.com/RasaHQ/rasa/blob/a7387f445604be5d123b776a2f4d1521990b8453/rasa/nlu/test.py#L1900) should rather call the `async` version of `train_nlu` (which is `_train_nlu_async`)?

rasa/cli/test.py
Imod7: > @Imod7 I don't understand this part in your proposed changes, what do you mean?
> 
> > Making async a list of functions and then adding the corresponding await statements when they are called.

@wochinge For someone who doesn't know the issue at all, I just thought it would be good to mention that I added `async` and `await` statements in the definition and the calling of a list of functions. Like a summary of some of the changes I did instead of mentioning the name of the functions one by one and where those changes took place. They can also look at the files that were changed but I thought it is helpful for someone to get a quick idea before checking the files.

Imod7: > As discussed in the other PR, I think it would be better to pass the arguments directly to `run_nlu_test_async` to make `run_nlu_test_async` more reusable. Our command-line parser library `argparse` is the only thing which encapsulates arguments into the `arparse.Namespace` type, while other code would rather pass the arguments directly.
> 
> What do you think?
> 
> That's a short and not complete example of what I mean:
> 
> ```python
> def run_nlu_test(args: Argparse) -> None:
>    rasa.utils.common.run_in_loop(run_nlu_test_async(args.nlu, args.out, args.config, ...))
> 
> async def run_nlu_test_async(data_path: Optional[Text], output_dir: Optional[Text], config: Optional[Text], ...) -> None:
>     # implementation
> ```

@wochinge Yes! I wanted to ask you about this one because I didn't know how to change it, like what type of arguments to pass. So it is better to pass explicitly the arguments that I will use inside the function (data_path, output_dir, config)? Can I not use something like **kwargs? Because if I pass them explicitly and at one point we add an extra option then we have to add it in the parameters also.

wochinge: > So it is better to pass explicitly the arguments that I will use inside the function (data_path, output_dir, config)?

Yes 👍🏻 This has the following advantage:
* it's clear to the caller which arguments are expected
* We can type annotate each parameter which then our CI can check which in turn makes our code more robust

> Because if I pass them explicitly and at one point we add an extra option then we have to add it in the parameters also.

I'd argue that it is the responsibility of the person adding the extra option to make sure it's used correctly. If you'd use `kwargs` then we'd still have to extra the value at some point in the code. This means the person adding the extra option would have to change other code anyway. 
Or maybe from a different perspective: If all our code just passes around some dictionaries then we end up with a giant black box. When you want to understand whether / how something is at some point a provided parameters in such a codebase you'd have to understand a lot of prior code. Does that make sense?

> didn't know how to change it, like what type of arguments to pass.

To see which parameters you need, you can look at the function and see where we call `args.<some value name>`. To see which type they should have you can look at the functions which our function calls and see how these values are annotated there. Consider the following example:

```python
def foo(args: argparse.Namespace) -> None:
   my_param = args.test
   bar(my_param)
   bar2(args.output_directory)
   
def bar(value: Text, some_value: Optional[float] = None) -> None:
   pass

def bar(dir: Optional[Text] = None) -> None:
   pass
```

By looking at the implementation of `foo` you can see that there are two CLI parameters `test` and `output_directory`. After looking at the type annotation in `bar` and `bar2` you can see that they have type `Text` and type `Optional[Text]`. The explicit version of `foo` would hence look like:

```python
def foo_explicit(my_param: Text, output_directory: Optional[Text) -> None:
   bar(my_param)
   bar2(output_directory)
```

Imod7: @wochinge Changed the parameters of `run_nlu_test_async` so it's called with the specific arguments and not with `argparse.Namespace` anymore. 
Some thoughts : 
- Maybe passing `vars(args)` as a first argument (which goes to `all_args: Optional[Dict]`) is not a good idea and I should : 
   - construct the `Dict` inside the `run_nlu_test_async` function by using all the available parameters or
   - construct the `Dict` inside the `run_nlu_test`, then passing it as a Dict argument in `run_nlu_test_async`
That way I avoid any black boxes that you mention in your comments.
- Default values are set in the parameters. But every time someone decides to change the default values in general, then someone also has to change the default values in `run_nlu_test_async`. Is this a good practice? Maybe these default values are set in some global variables somewhere? 
- I added the `docstring` from my understanding of the use of each function parameter.

rasa/train.py
Imod7: @wochinge Hey Tobias! 
Here are the error messages [trace.txt](https://github.com/RasaHQ/rasa/files/5768791/trace.txt)
I can check and try to call `_train_nlu_async`

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
     )
 
 
-def run_nlu_test(args: argparse.Namespace) -> None:
-    """Run NLU tests."""
+async def run_nlu_test_async(
+    all_args: Optional[Dict] = None,
```
wochinge, line 120: I'd suggest to make this the last argument (similar as `additional_arguments` in `perform_nlu_cross_validation` or `test_nlu`

rasa/cli/test.py
wochinge: > Maybe passing vars(args) as a first argument (which goes to all_args: Optional[Dict]) is not a good idea and I should 

You have a very good point there. This whole `vars` thing is in my opinion not clean at all. I'm a bit worried though that this PR is getting to big, so I'd actually keep your solution for now. If you want you can tackle this in a follow-up PR?

> Default values are set in the parameters. 

I'd omit them for now as they shouldn't be required. 

> I added the docstring from my understanding of the use of each function parameter.
 Write
 
 Sounds good. Will review them in the final review 👍🏻

rasa/train.py
wochinge: I double checked for `test_train_nlu_temp_files` and if you don't make the test function `async`, you can use `rasa.utils.common.run_in_loop` as before. 

The issue is that when you define a test function via

```python
async def test_something():
   pass
```
it will start a event loop. When your test code starts another event loop via `rasa.utils.common.run_in_loop` you get the error above. As you don't need the `async` for anything in this test, you can remove it safely.

rasa/__main__.py
```diff
@@ -113,7 +114,10 @@ def main() -> None:
             set_log_and_warnings_filters()
             rasa.telemetry.initialize_telemetry()
             rasa.telemetry.initialize_error_reporting()
-            cmdline_arguments.func(cmdline_arguments)
+            if inspect.iscoroutinefunction(cmdline_arguments.func):
+                rasa.utils.common.run_in_loop(cmdline_arguments.func(cmdline_arguments))
+            else:
+                cmdline_arguments.func(cmdline_arguments)
```
wochinge, line 120: this is no longer needed, no?

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
     )
 
 
-def run_nlu_test(args: argparse.Namespace) -> None:
-    """Run NLU tests."""
+async def run_nlu_test_async(
+    config: Optional[Union[Text, Dict]],
```
wochinge, line 120: according to your docstring it should be this, no?
```suggestion
    config: Optional[Union[Text, List[Text]],
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+    cross_validation: Optional[bool],
+    percentages: Optional[List[int]],
+    runs: Optional[int],
+    no_errors: Optional[bool],
+    all_args: Optional[Dict[Text, Any]],
+) -> None:
+    """Runs NLU tests.
+
+    Args:
+        all_args: all arguments gathered in a Dict so we can pass them to 'perform_nlu_cross_validation' and
```
wochinge, line 133: can please wrap the docstring so that it doesn't exceed our 88 char limit?

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+                  'test_nlu' as parameter.
+        config: config file or a list of multiple config files.
+        data_path: path for the nlu data.
+        models_path: the path for the nlu data.
```
wochinge, line 137:
```suggestion
        models_path: Path to a trained Rasa model..
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+        output_dir: the directory for the results to be saved.
+        cross_validation: boolean value that indicates if it should test the model using cross validation or not.
```
wochinge, line 139: I suggest not repeating the types in docstrings as we have the type annotations for this.
```suggestion
        cross_validation: indicates if it should test the model using cross validation or not.
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+        percentages: defines the exclusion percentage of the training data.
+        runs: used in 'compare_nlu_models' and indicates the number of comparison runs.
```
wochinge, line 141:
```suggestion
        runs: Number of comparison runs to make.
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+        no_errors: boolean value that indicates if incorrect predictions should be written to a file or not.
```
wochinge, line 142: I suggest not repeating the types in docstrings as we have the type annotations for this.

I suggest not repeating the types in docstrings as we have the type annotations for this.


```suggestion
        no_errors: indicates if incorrect predictions should be written to a file or not.
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
 
-def run_nlu_test(args: argparse.Namespace) -> None:
-    """Run NLU tests."""
+async def run_nlu_test_async(
+    config: Optional[Union[Text, Dict]],
+    data_path: Optional[Text],
+    models_path: Optional[Text],
+    output_dir: Optional[Text],
+    cross_validation: Optional[bool],
+    percentages: Optional[List[int]],
```
wochinge, line 125: As `compare_nlu_models` only accepts a `List[int]` this should also only be able to accept `List[int]`
```suggestion
    percentages:List[int],
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
 
 
-def run_nlu_test(args: argparse.Namespace) -> None:
-    """Run NLU tests."""
+async def run_nlu_test_async(
+    config: Optional[Union[Text, Dict]],
+    data_path: Optional[Text],
+    models_path: Optional[Text],
+    output_dir: Optional[Text],
+    cross_validation: Optional[bool],
```
wochinge, line 124: As the our `argparse` config specifies a default, `None` can't happen
```suggestion
    cross_validation:bool,
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+    percentages: Optional[List[int]],
+    runs: Optional[int],
```
wochinge, line 126: As the our `argparse` config specifies a default, `None` can't happen

```suggestion
    runs: int,
```

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+    no_errors: Optional[bool],
+    all_args: Optional[Dict[Text, Any]],
```
wochinge, line 128: How can this parameter become `None` in the existing calls?
```suggestion
    all_args: Dict[Text, Any],
```

rasa/cli/test.py
```diff
@@ -148,26 +171,46 @@ def run_nlu_test(args: argparse.Namespace) -> None:
...
-            args.model, "model", DEFAULT_MODELS_PATH
+            models_path, "model", DEFAULT_MODELS_PATH
         )
 
-        test_nlu(model_path, nlu_data, output, vars(args))
+        await test_nlu(model_path, nlu_data, output, all_args)
+
+
+def run_nlu_test(args: argparse.Namespace) -> None:
+    """Adding this function layer to be able to run run_nlu_test_async in the event loop.
```
wochinge, line 196: can you please update the docstring?

rasa/test.py
```diff
@@ -110,6 +112,7 @@ def test_core(
     output: Text = DEFAULT_RESULTS_PATH,
     additional_arguments: Optional[Dict] = None,
 ) -> None:
+    """Checks if models are present and then runs the test function in the event loop."""
```
wochinge, line 115: I think it's better to describe what and why things are happening in a docstring instead of describing how things are done. 
```suggestion
    """Tests a trained Core model against a set of test stories."""
```

wochinge (commented):
Wuhuu, we're nearly there 💯 Awesome work with disentangling all the `async` things 🚀

rasa/cli/test.py
Imod7: Maybe this one 
```
"""Runs NLU tests by calling run_nlu_test_async in the event loop.

    Args:
        args: all arguments that were given and parsed.
    """
```

wochinge: The `args` part feels a bit long. How about `the parsed CLI arguments for 'rasa test nlu'`?

Imod7:
I think I did all the last changes that you mentioned and I also removed `Optional` from these parameters in `run_nlu_test_async` :
```
data_path: Text,
models_path: Text,
output_dir: Text,
no_errors: bool,
```
because as you showed yesterday I checked the functions related to `set_test_nlu_arguments` and : 
- in `add_nlu_data_param` it adds default nlu/data path `data`
- in `add_model_param` it adds the default models path/directory `models`
- in `add_test_nlu_argument_group`  it sets the default output directory `results`
- in `add_errors_success_params` it sets default no_errors `False`

Changed the docstring again of `run_nlu_test_async`  and `run_nlu_test` (edited)

rasa/__main__.py
```diff
@@ -3,6 +3,7 @@
 import os
 import platform
 import sys
+import inspect
```
wochinge, line 6: this import is unused

rasa/cli/test.py
```diff
@@ -115,29 +116,54 @@ def run_core_test(args: argparse.Namespace) -> None:
     )
 
 
-def run_nlu_test(args: argparse.Namespace) -> None:
-    """Run NLU tests."""
+async def run_nlu_test_async(
+    config: Optional[Union[Text, List[Text]]],
+    data_path: Text,
+    models_path: Text,
```
wochinge, line 122: that can be `None`, no?
```suggestion
    models_path: Optional[Text],
```

rasa/cli/test.py
```diff
@@ -115,29 +116,54 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+    output_dir: Text,
```
wochinge, line 123: this can also be `None`, no?
```suggestion
    output_dir: Optional[Text],
```

wochinge (commented):
Two more comments. Can you please also update the pull request description and convert your PR from a draft to a actual pull request?

rasa/cli/test.py
Imod7: Same as above

rasa/cli/test.py
Imod7: So I thought I should remove the `Optional` because I saw that it always gets a default value so it will never be `None` as long as it is run through the command line. 
`_StoreAction(option_strings=['-m', '--model'], dest='model', nargs=None, const=None, default='models'`

wochinge: Very good point! That can indeed not be `None`. Sorry for the confusion!

rasa/cli/test.py
wochinge: 👍🏻

rasa/__main__.py
wochinge: @Imod7 This is still here?

Imod7: Yes, I was waiting for feedback on the other 2 comments and then I would push. I will do a final check and push now.

wochinge: 👍🏻 Could please also resolve the comments further up in the PR which you've worked in? Makes it easier to see which changes weren't addressed yet

Imod7: Yes! I will change the description of the PR and convert it from draft also.

wochinge:
Just re-request review once you're done and I have a final look + approve 🎉

rasa/cli/test.py
```diff
@@ -115,29 +116,51 @@ def run_core_test(args: argparse.Namespace) -> None:
...
+async def run_nlu_test_async(
+    all_args: Optional[Dict] = None,
+    config: Optional[Union[Text, Dict]] = None,
+    data_path: Optional[Text] = "data",
+    models_path: Optional[Text] = "models",
+    output_dir: Optional[Text] = "results",
+    cross_validation: Optional[bool] = False,
+    percentages: Optional[List[int]] = [0, 25, 50, 75],
+    runs: Optional[int] = 3,
+    no_errors: Optional[bool] = False,
```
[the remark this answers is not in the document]
Imod7: I removed all default values from the parameters of run_nlu_test_async

rasa/cli/test.py
Imod7: Moved `all_args` as the last parameter in `run_nlu_test_async`.

rasa/__main__.py
Imod7: True! I removed the coroutine check and it calls directly `cmdline_arguments.func(cmdline_arguments)` as before.

rasa/cli/test.py
Imod7: True! I changed it as you suggested.

rasa/cli/test.py
Imod7: I did! I also changed the setting in Pycharm so that the wrapping line is at 88 characters and this hopefully does not happen again.

rasa/cli/test.py
Imod7: Thank you for the suggestion. Changed it!

rasa/cli/test.py
Imod7: True! I removed all types from doc strings.

rasa/cli/test.py
Imod7: Changed it !

rasa/cli/test.py
wochinge: slightly related and nice to know: https://en.wikipedia.org/wiki/Hungarian_notation

rasa/cli/test.py
Imod7: Changed it!

rasa/cli/test.py
Imod7: True! Changed it!

rasa/cli/test.py
Imod7: True! Changed it!

rasa/cli/test.py
Imod7: I adjusted it!

rasa/test.py
Imod7: Changed it! I will keep this advice also for all the future docstrings.

wochinge (approved):
Congrats to your first Rasa Open Source change 🚀 

Two notes:
* It's (in my opinion) ok to stay a little bit more high level in your pull request description. I can see the details in the code. In my opinion it's more important to explain why certain things where changed (e.g. why did did you have to make things `async`). No need to change anything in this PR, rather a suggestion for the future).
* Either add the label `ready-to-merge` to the PR so that the PR is automatically merged or press the green `merge` button yourself once all checks passed 🎉 (which I recommend for your first PR as pressing a button is way more fun and satisfying 😄 )

Issue #7686, opened 2021-01-06
Edited message on telegram raises an exception


tranchitella:
**Rasa version**: 2.0.0

**Python version**: 3.8

**Operating system**: Ubuntu 20.04.1 LTS

**Issue**: If a user edits a message over Telegram, the Rasa server raises an exception and returns an error in response to the Telegram webhook. From that point, the chatbot stops to work.

Already reported here, but closed because of missing activity: [https://github.com/RasaHQ/rasa/issues/5654](https://github.com/RasaHQ/rasa/issues/5654)

**Error (including full traceback)**:

 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: Traceback (most recent call last):
 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: File "/home/ubuntu/rasa-libya/venv/lib/python3.8/site-packages/sanic/app.py", line 938, in handle_request
 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: response = await response
 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: File "/home/ubuntu/rasa-libya/venv/lib/python3.8/site-packages/rasa/core/channels/telegram.py", line 217, in message
 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: if self._is_user_message(msg):
 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: File "/home/ubuntu/rasa-libya/venv/lib/python3.8/site-packages/rasa/core/channels/telegram.py", line 177, in _is_user_message
 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: return message.text is not None
 Jan 6 10:57:26 ip-10-50-188-7 rasa<span class="error">[215589]</span>: AttributeError: 'NoneType' object has no attribute 'text'

**Command or request that led to error**: Edit a message in Telegram

tranchitella:
PR: https://github.com/RasaHQ/rasa/pull/7687

sara-tagger:
Thanks for raising this issue, @desmarchris will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

superDPgg:
I have the same problem for telegram！

PR #7687, opened 2021-01-06
Do not break if update.message is None

tranchitella:
If the update doesn't contain a message, which is true for edited
messages, return a successful response instead of raising an exception
and returning an error to Telegram.

Signed-off-by: Fabio Tranchitella <fabio@tranchitella.eu>

Issue: https://github.com/RasaHQ/rasa/issues/7686

### DIFF
--- a/rasa/core/channels/telegram.py
+++ b/rasa/core/channels/telegram.py
@@ -212,9 +212,9 @@ class TelegramInput(InputChannel):
                     text = update.callback_query.data
                 else:
                     msg = update.message
-                    if self._is_user_message(msg):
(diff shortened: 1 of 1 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @tmbo will take a look at it as soon as possible ✨

tmbo (commented):
Looks good :+1: thanks a lot for that telegram fix. The code lgtm. Do you mind adding a test to avoid a regression in the future?

tranchitella:
Superseded by https://github.com/RasaHQ/rasa/pull/8719

Issue #7226, opened 2020-11-10
Unfeaturize rule snippets for inference through TED


dakshvar22:
**Description of Problem**:
Currently, during inference to predict the next bot action, the first check is to see if a rule can be triggered. If not, then the prediction is done through ML policies. However, this can potentially cause a problem because these rule snippets may not be part of any story during training and hence during inference the ML policy may be confused by their presence in the tracker. 

**Overview of the Solution**:
We can unfeaturize/revert conversation turns in the tracker which **only** appear as a rule in the training data. That way any rule which was never a part of a training story but is triggered at inference time is not part of the tracker when an ML policy like TED makes a prediction.

For example, we can have a rule -
```
intent: greet
action: utter_greet
```

Let’s say this rule doesn’t appear in any stories. The conversation may proceed as -
```
User: greet
Bot: utter_greet
User: mood_happy
Bot: ...
```
When the bot has to predict the next action after `mood_happy`, we unfeaturize the states corresponding to the conversation turn involving `greet` and `utter_greet`.

One way to do this is to add new event called `RuleReverted` which contains appropriate number of dialogue turns to revert. When `RulePolicy` is trained, it "learns" deterministically from data whether or not to predict this event during inference.

Ghostvv:
I just understood, you cannot do it in `tracker.applied_events`, because rule policy should see this events. Rather, you should create a temporary tracker inside predict ted, and rule policy should "learn" whether this predicted action should be unfeaturized or not. This information (rather logical not) can be extracted from checking rule predictions on stories. Then `RulePolicy` can release optional event

PR #7701, opened 2021-01-08
hide only rule dialogue turns

Ghostvv:
**Proposed changes**:
- closes https://github.com/RasaHQ/rasa/issues/7226

originally I had 2 ideas: operate on states or use forgetting functionality in tracker.applied_events through specialized events. In the end I used the mix of these approaches.


Ghostvv:
tests, docs and changelog are to come. But could you please check logical correctness

docs/docs/rules.mdx
```diff
@@ -47,6 +47,14 @@ rules:
 This example rule applies at the start of conversation as well as when the user decides
 to a send a message with an intent `greet` in the middle of an ongoing conversation.
 
+Dialogue turns that only appear as a rule in the training data and do not appear in stories
+will be hidden in the tracker states for ML only policies like `TEDPolicy`.
```
erohmensing, line 51:
```suggestion
Dialogue turns that only appear as rules in the training data and do not appear in stories
will be ignored by ML-only policies like `TEDPolicy` at prediction time.
```

erohmensing: Is this still a valid explanation?

changelog/7701.improvement.md
```diff
@@ -0,0 +1,3 @@
+Hide dialogue turns predicted by `RulePolicy` in the tracker states
+for ML only policies like `TEDPolicy`
+if dialogue turns only appear as a rule in the training data and does not appear in stories.
```
erohmensing, line 3:
```suggestion
Hide dialogue turns predicted by `RulePolicy` in the tracker states
for ML-only policies like `TEDPolicy`
if those dialogue turns only appear as rules in the training data and do not appear in stories.
```

docs/docs/rules.mdx
```diff
@@ -47,6 +47,14 @@ rules:
...
+For example if you have a greet rule above and don't add it to any of your stories then
+when during prediction `RulePolicy` predicts `utter_greet`, afterwards `TEDPolicy` will
+see dialogue turns as if user never input intent `greet`.
+
```
erohmensing, line 55:
```suggestion
For example if you define the greeting rule as above and don't add it to any of your stories,
after `RulePolicy` predicts `utter_greet`, `TEDPolicy` will
make predictions as if the `greet, utter_greet` turn did not occur.
```

Ghostvv:
@wochinge could you please take a look if we can remove forgetting of `TwoStageFallback` rules

tests/core/policies/test_rule_policy.py
```diff
@@ -1928,3 +1932,277 @@ def test_predict_nothing_if_fallback_disabled():
...
+        UserUttered("haha", {"name": some_intent}),
+    ]
+    prediction = policy.predict_action_probabilities(
+        DialogueStateTracker.from_events(
+            "casd", evts=conversation_events, slots=domain.slots
+        ),
+        domain,
+        RegexInterpreter(),
+    )
+    assert_predicted_action(prediction, domain, some_action)
```
dakshvar22, line 2097: `some_slot` should also be reset in the states for ml policy right? Can we check that just as how we check inside `test_hide_rule_turn`

dakshvar22 (approved):
Amazing job on this§1 💯  Just a couple of minor comments from my side.

rasa/core/policies/ted_policy.py
```diff
@@ -568,16 +576,30 @@ def _pick_confidence(
...
+            logger.debug(f"User text lead to '{e2e_action_name}'.")
             if (
                 np.max(confidences[1]) > self.config[E2E_CONFIDENCE_THRESHOLD]
                 # TODO maybe compare confidences is better
                 and np.max(similarities[1]) > np.max(similarities[0])
             ):
+                logger.debug(f"TED predicted '{e2e_action_name}' based on user text.")
                 return confidences[1], True
 
+            logger.debug(f"TED predicted '{non_e2e_action_name}' based on user intent.")
```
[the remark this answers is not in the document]
Ghostvv: yes, I was changing tracker features and I thought it is innocent change not to create separate PR for this

rasa/core/policies/rule_policy.py
```diff
@@ -621,11 +698,43 @@ def train(
...
+        Args:
+            training_trackers: The list of the trackers.
+            domain: The domain.
+            interpreter: Interpreter which can be used by the polices for featurization.
+        """
+        # only consider original trackers (no augmented ones)
+        training_trackers = [
+            t
+            for t in training_trackers
+            if not hasattr(t, "is_augmented") or not t.is_augmented
```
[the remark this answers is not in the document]
Ghostvv: it is the difference between `TrackerWithCachedStates` and `DialogueStateTracker` you can actually pass both to train methods. I'd say refactoring it is out of scope of this PR

rasa/shared/core/domain.py
```diff
@@ -1068,15 +1068,90 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
...
+        last_action_is_ml_action = True
+        for tr in tracker.generate_all_prior_trackers():
+            if for_only_ml_policy:
+                # remember previous ml action based on the last non hidden turn
+                # we need this to override previous action in the ml state
+                if last_action_is_ml_action:
+                    last_ml_action_sub_state = self._get_prev_action_sub_state(tr)
+                last_action_is_ml_action = not tr.hide_rule_turn
+
+                if tr.hide_rule_turn:
```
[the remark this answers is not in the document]
Ghostvv: when the dialogue turn is predicted by ML policy

tests/core/policies/test_rule_policy.py
Ghostvv: updated the test

wochinge (commented):
I need a bit more time to digest this so here is my input on this so far:
* If users turn of `_check_for_contradictions` then this wouldn't work only longer, correct? 
* Wouldn't it be better to warn the user that they should add a story so that the ML policies can learn to perform the transition after a rule application instead of completely blanking out the rule? 
* Could we also to this without the additional event and simply remove all events before an `ActionExecuted` with `policy=="RulePolicy"` until the previous action?

Ghostvv:
>Wouldn't it be better to warn the user that they should add a story so that the ML policies can learn to perform the transition after a rule application instead of completely blanking out the rule?

no, because you'd need to add combinatorial number of rules into stories, which is unnecessary if you can hide rules

> Could we also to this without the additional event and simply remove all events before an ActionExecuted with policy=="RulePolicy" until the previous action?

no, because not all rules should be ignored, only the ones that are not present in the stories, also not all slot sets should be ignored

rasa/core/policies/rule_policy.py
```diff
@@ -571,29 +650,20 @@ def _find_contradicting_rules(
             )
 
         logger.debug("Found no contradicting rules.")
+        all_rules = (
+            set(self._rules_sources.keys())
+            | self._default_sources()
+            | self._handling_loop_sources(domain)
+        )
+        # set is not json serializable, so convert to list
+        return list(all_rules - rules_used_in_stories)
```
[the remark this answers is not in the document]
Ghostvv: before it was even worse, it was double usage of `_check_prediction`. Would renaming `_find_contradicting_rules` help?

rasa/core/policies/rule_policy.py
```diff
@@ -725,26 +834,30 @@ def _get_possible_keys(
...
         default_action_name = DEFAULT_ACTION_MAPPINGS.get(
             tracker.latest_message.intent.get(INTENT_NAME_KEY)
         )
 
         if default_action_name:
             logger.debug(f"Predicted default action '{default_action_name}'.")
+            return (
+                default_action_name,
+                DEFAULT_RULES + tracker.latest_message.intent.get(INTENT_NAME_KEY),
+            )
```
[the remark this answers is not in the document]
Ghostvv: what do you mean?

Ghostvv:
> If users turn of _check_for_contradictions then this wouldn't work only longer, correct?
yes

Ghostvv:
@ArjaanBuijk did you have a chance to try this out?

ArjaanBuijk:
> @ArjaanBuijk did you have a chance to try this out?

I will test it today. This is a great enhancement that will likely allow cleanup of financial-demo, where we had introduced logic in custom actions to keep the tracker clean.

ArjaanBuijk:
@Ghostvv ,

The logic works great, but this update will break bots that currently rely on a `UserUtteranceReverted` event sent from a custom action.

It seems that the bot is rewinding beyond the current utterance, if the utterance is a rule only dialogue turn.

tests/core/policies/test_rule_policy.py
```diff
@@ -1936,3 +1940,374 @@ def test_predict_nothing_if_fallback_disabled():
...
+    )
+    states = tracker.past_states(domain, ignore_rule_only_turns=True)
+    assert states == [
+        {},
+        {USER: {INTENT: chitchat}, PREVIOUS_ACTION: {ACTION_NAME: ACTION_LISTEN_NAME}},
+        {USER: {INTENT: chitchat}, PREVIOUS_ACTION: {ACTION_NAME: action_chitchat}},
+    ]
+
+
+def test_hide_rule_turn_with_loops():
```
wochinge, line 2206: can we please also assert in a test that there is no `HideRuleTurn` event for loops which are also in stories?

rasa/shared/core/domain.py
```diff
@@ -1068,15 +1068,90 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
...
+                    continue
+
+            state = self.get_active_states(tr)
+
+            if ignore_rule_only_turns:
```
wochinge, line 1129: can we concat this with the previous `if ignore_rule_only_turns`?

rasa/shared/core/domain.py
```diff
@@ -1068,15 +1068,90 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
         }
         return self._clean_state(state)
 
+    @staticmethod
+    def _remove_rule_only_features(
+        state: State, tracker: "DialogueStateTracker"
+    ) -> None:
+        # remove only rule slots
```
wochinge, line 1075: let's be a bit verbose here so that it gets easier to understand 
```suggestion
        # remove slots which only occur in rules but not in stories
```

rasa/shared/core/domain.py
```diff
@@ -1068,15 +1068,90 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
...
+        for slot in tracker.rule_only_slots:
+            if state.get(rasa.shared.core.constants.SLOTS, {}).get(slot):
+                del state[rasa.shared.core.constants.SLOTS][slot]
+        # remove active loop feature if it is only rule loop
```
wochinge, line 1079:
```suggestion
        # remove active loop which only occur in rules but not in stories
```

rasa/core/policies/rule_policy.py
wochinge, line 719: why do we need `hasattr`?

rasa/core/policies/rule_policy.py
```diff
@@ -896,33 +1015,38 @@ def predict_action_probabilities(
...
-            return self._prediction(
+            return self._rule_prediction(
                 self._prediction_result(loop_happy_path_action_name, tracker, domain),
                 is_no_user_prediction=True,
             )
 
         # predict rules from text first
         if rules_action_name_from_text:
-            return self._prediction_with_unhappy_path(
+            self._prediction_source = prediction_source_from_text
```
wochinge, line 1048: We can't set any policy state during prediction time. This would e.g. blow up if a second user message is handled in parallel by the policy.

rasa/core/policies/rule_policy.py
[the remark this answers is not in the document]
wochinge: We do the string concatenation to rebuild the string key which we created in `_default_sources()`, right? In my opinion this connection should be clarified in the code.

rasa/core/policies/rule_policy.py
```diff
@@ -449,22 +487,51 @@ def _check_prediction(
...
+            DEFAULT_RULES + default_intent
+            for default_intent in DEFAULT_ACTION_MAPPINGS.keys()
+        }
+
+    @staticmethod
+    def _handling_loop_sources(domain: Domain) -> Set[Text]:
+        loop_sources = set()
+        for loop_name in domain.form_names:
+            loop_sources.add(LOOP_RULES + loop_name)
+            loop_sources.add(LOOP_RULES + loop_name + " - " + ACTION_LISTEN_NAME)
```
wochinge, line 524: `" - "` should be a constant

rasa/core/policies/rule_policy.py
```diff
@@ -887,7 +1006,7 @@ def predict_action_probabilities(
         """Predicts the next action (see parent class for more information)."""
         (
             rules_action_name_from_text,
-            self._prediction_source,
+            prediction_source_from_text,
```
wochinge, line 1009: The name seems quite confusing, what about something like:
```suggestion
            prediction_source_if_not_loop_prediction,
```

rasa/core/policies/rule_policy.py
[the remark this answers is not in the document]
wochinge: What would you suggest?

I don't think it would help much. In my opinion this mixing multiple things which make the code hard to understand:
* finding rules which are also in stories
* finding conflicts

wochinge (commented):
I'm still a bit worried about frequence in which we create new event types 😬 
Marking start / end of "rule-like" turns is also covered by the `LoopEvent` for example. Could we unify `Loop` and `HideRuleTurn`? 

Also, the new event rather seems like a workaround to add static `RulePolicy` data to the tracker (e.g. `only_rule_slots`, `only_rule_loops` are always the same once the `RulePolicy is trained). Would we rather need a mechanism to let policies set some initial tracker state before doing a prediction / at the beginning of a session?

rasa/core/policies/rule_policy.py
Ghostvv: this is an old code. it happens because right now 2 types of trackers are not interchangeable because of this param. It needs to be fixed, but it is out of scope of this PR

rasa/core/policies/rule_policy.py
Ghostvv: It is like this in `master`, but I don't think is actually used during prediction time

rasa/core/policies/rule_policy.py
Ghostvv: but it is not prediction source if not loop prediction, why do you want to rename to that?

rasa/core/policies/rule_policy.py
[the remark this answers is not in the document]
Ghostvv: I think it's totally fine it does it in one go over training trackers

Ghostvv:
> I'm still a bit worried about frequence in which we create new event types 😬

I told you that would happen long time ago 😉

> Marking start / end of "rule-like" turns is also covered by the LoopEvent for example. Could we unify Loop and HideRuleTurn?

I don't understand what do you mean. `LoopEvent` is used to mark the current active loop, how can it be used for any generic rule?

> Also, the new event rather seems like a workaround to add static RulePolicy data to the tracker (e.g. only_rule_slots, only_rule_loops are always the same once the `RulePolicy is trained). Would we rather need a mechanism to let policies set some initial tracker state before doing a prediction / at the beginning of a session?

yes we would, we had this discussion in slack

rasa/core/policies/rule_policy.py
Ghostvv: great point, totally slipped out of my mind. Thank you for spotting, it saved us a lot of trouble!

rasa/core/policies/rule_policy.py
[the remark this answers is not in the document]
Ghostvv: maybe some more generic name like analyze rules or smth?

Ghostvv:
> I'm still a bit worried about frequence in which we create new event types 😬

why do you think it is a problem?

rasa/core/policies/rule_policy.py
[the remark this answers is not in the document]
wochinge: > maybe some more generic name like analyze rules or smth?

Ok 👍🏻 In combination with a good docstring that could work. I'd also rename / split `_run_prediction_on_trackers` to get rid of the `collect_sources` flag (kinda doing [this](https://martinfowler.com/bliki/FlagArgument.html) refactoring)

rasa/core/policies/rule_policy.py
wochinge: Is there an issue which describes the problem? And which two types of tracker?

rasa/core/policies/rule_policy.py
wochinge: nevermind 👍🏻

wochinge:
> why do you think it is a problem?

* Because they will hang around in the tracker stores / databases forever which means we have to maintain them forever
* The affect all down stream dependencies: each change here affects the SDK and Rasa X

> But also we still need this event, because it's presence or absence is a marker

Isn't the name of the policy marker enough? 

> I don't understand what do you mean. LoopEvent is used to mark the current active loop, how can it be used for any generic rule?

My question was whether we can make `Loop` a more generic marker event.

Ghostvv:
> Isn't the name of the policy marker enough? 

no, because the fact that `RulePolicy` predicted an action, doesn't mean it should be hidden.
In principle we could use `PolicyPrediction` object to communicate this information but then it clashes with the way we featurize our trackers. When we create states from intermediate trackers up to `ActionExecuted` so that action itself is not part of the featurization process.

> My question was whether we can make Loop a more generic marker event.

I don't know. Maybe, but for me they are completely independent from each other

rasa/shared/core/trackers.py
```diff
@@ -205,6 +207,9 @@ def __init__(
         self.latest_bot_utterance = None
         self._reset()
         self.active_loop: Dict[Text, Union[Text, bool, Dict, None]] = {}
+        self.hide_rule_turn = False
+        self.rule_only_slots = []
+        self.rule_only_loops = []
```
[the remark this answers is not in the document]
Ghostvv: I don't know. I'm not familiar with `Tracker.current_state`, how is it used?

tests/core/policies/test_rule_policy.py
```diff
@@ -1936,3 +1940,374 @@ def test_predict_nothing_if_fallback_disabled():
     )
 
     assert prediction.max_confidence == 0
+
+
+def test_hide_rule_turn():
```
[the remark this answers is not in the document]
Ghostvv: do you have an idea, how to do that?

tests/core/policies/test_rule_policy.py
```diff
@@ -1936,3 +1940,374 @@ def test_predict_nothing_if_fallback_disabled():
...
+    ]
+    tracker = DialogueStateTracker.from_events(
+        "casd", evts=conversation_events, slots=domain.slots
+    )
+    states = tracker.past_states(domain, ignore_rule_only_turns=True)
+    assert states == [
+        {},
+        {
+            USER: {TEXT: "haha", INTENT: some_other_action},
+            PREVIOUS_ACTION: {ACTION_NAME: ACTION_LISTEN_NAME},
```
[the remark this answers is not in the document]
Ghostvv: not sure what do you mean. Why do you think it should be in the states?

tests/core/policies/test_rule_policy.py
```diff
@@ -1936,3 +1940,374 @@ def test_predict_nothing_if_fallback_disabled():
...
+        - {chitchat}
+        - {activate_another_form}
+        actions:
+        - {UTTER_GREET_ACTION}
+        - {action_chitchat}
+        slots:
+          {REQUESTED_SLOT}:
+            type: unfeaturized
+        forms:
+        - {form_name}
```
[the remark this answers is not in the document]
Ghostvv: should it be
```
        forms:
          {form_name}:
```
?

rasa/shared/core/domain.py
Ghostvv: what do you mean? state need to be calculated in between

rasa/core/policies/rule_policy.py
Ghostvv: I don't think there is an issue.

> And which two types of tracker?

`TrackerWithCachedStates` and `DialogueStateTracker`

rasa/shared/core/events.py
```diff
@@ -1875,3 +1883,77 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
...
         tracker._reset()
+
+
+class HideRuleTurn(SkipEventInMDStoryMixin, AlwaysEqualEventMixin):
+    """Emulate undoing of the last dialogue turn.
+
+    The bot reverts everything until before the most recent action.
+    This includes the action itself, as well as any events that
+    action created, like set slot events, unless they are in a list
+    to reapply them.
```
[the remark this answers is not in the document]
Ghostvv: rephrased

rasa/shared/core/trackers.py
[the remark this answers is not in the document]
Ghostvv: I'd prefer to set `self.rule_only_slots` and `self.rule_only_loops` as dialogue turn independent properties directly from `RulePolicy` during loading, but I don't know how to do that

tests/core/policies/test_rule_policy.py
```diff
@@ -1936,3 +1940,374 @@ def test_predict_nothing_if_fallback_disabled():
     )
 
     assert prediction.max_confidence == 0
+
+
+def test_hide_rule_turn():
+    chitchat = "chitchat"
+    action_chitchat = "action_chitchat"
+    domain = Domain.from_yaml(
```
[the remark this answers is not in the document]
Ghostvv: sure. Would it be:
```
f"""
version: "2.0"

intents:
```
?

wochinge:
> no, because the fact that RulePolicy predicted an action, doesn't mean it should be hidden.

Ok. How about adding the information about used rule and whether this rule is covered by a story as well in an `ActionExecuted` property (either a new one or we use `metadata` for this)?

wochinge:
@Ghostvv Sorry, I know it's somewhat my fault because I didn't come back to this PR for a while but I think this needs some more time for discussions and hence won't be ready to into 2.3 tomorrow. I'd rather take several extra refactoring steps than having to deal with the legacy of an extra event type

Ghostvv:
> hence won't be ready to into 2.3 tomorrow

totally agree, just wanted to suggest that 😉

Ghostvv:
> How about adding the information about used rule and whether this rule is covered by a story as well in an ActionExecuted property

I don't think used rule makes sense. I'll create a property whether this action is only rule action. It'll make creating states, a bit tricky, but nothing we cannot handle.

How do we pass `only_rule_slots` and `only_rule_loops` to trackers?

wochinge:
>  I'll create a property whether this action is only rule action. It'll make creating states, a bit tricky, but nothing we cannot handle.

Nice 👍🏻 

> How do we pass only_rule_slots and only_rule_loops to trackers?

How about doing this during the training data importing time and adding it as a propery to the domain?

Ghostvv:
> How about doing this during the training data importing time and adding it as a propery to the domain?

hm, this is not a property of the domain, we need a trained `RulePolicy` to create these lists

Ghostvv:
@wochinge actually, I don't think we need to pass this info till `trackers`, we can keep it as `self.` variables in the `ensemble.py`

Ghostvv:
@wochinge I removed `HideRuleTurn` event. Also, could you please take a look at my answers to your comments

wochinge:
@Ghostvv I try to have another look today but can't promise as it's full of meetings 😬 FYI: I'm on vacation until next Friday then.

rasa/core/policies/ensemble.py
```diff
@@ -69,6 +70,21 @@ def __init__(
...
+        rule_policy = next(
+            (policy for policy in self.policies if isinstance(policy, RulePolicy)),
+            None,
+        )
+        rule_only_slots = (
+            rule_policy.lookup.get(RULE_ONLY_SLOTS, []) if rule_policy else []
+        )
+        rule_only_loops = (
+            rule_policy.lookup.get(RULE_ONLY_LOOPS, []) if rule_policy else []
+        )
```
wochinge, line 85: can we make a public method on `RulePolicy` to get these (e.g. `get_slots_used_only_in_rules`)? Directly accessing these seems a bit fragile to me.

rasa/shared/core/domain.py
wochinge: can't we calculate the state before the previous `if`?

rasa/shared/core/events.py
```diff
@@ -1511,13 +1518,20 @@ def as_dict(self) -> Dict[Text, Any]:
         confidence = None
         if hasattr(self, "confidence"):
             confidence = self.confidence
+        action_text = None
+        if hasattr(self, "action_text"):
+            action_text = self.action_text
```
wochinge, line 1523: I think all his `hasattr` stuff is old and we don't need that anymore.

wochinge (commented):
Nice! 💯 Way better than the additional event in my opinion. Didn't have time for a detailed review, but I think the overall approach is good now and we can merge after another (more code) focused review.

@joejuzl Could. you maybe take over as I'll be offline till next Friday? Note that the parts around `ActionExecuted` and setting data there via `PolicyPrediction` are exactly what we discussed today as well.

rasa/shared/core/domain.py
Ghostvv: then it is unnecessary state calculation. Previous if contains `continue`

docs/docs/rules.mdx
```diff
@@ -47,6 +47,13 @@ rules:
 This example rule applies at the start of conversation as well as when the user decides
 to a send a message with an intent `greet` in the middle of an ongoing conversation.
 
+Dialogue turns that only appear as rules in the training data and do not appear in stories
+will be ignored by ML-only policies like `TEDPolicy` at prediction time.
+For example if you define the greeting rule as above and don't add it to any of your stories,
```
wochinge, line 52: Let's give actual rules / stories as example to make it less abstract for users.

rasa/core/policies/ensemble.py
```diff
@@ -69,6 +70,15 @@ def __init__(
         self._check_priorities()
         self._check_for_important_policies()
 
+        self._rule_only_slots, self._rule_only_loops = self._get_rule_only_slots_loops()
+
+    def _get_rule_only_slots_loops(self) -> Tuple[List[Text], List[Text]]:
```
wochinge, line 75: I'd either add an `and` or make it a more general name: 
```suggestion
    def _get_rule_only_features(self) -> Tuple[List[Text], List[Text]]:
```

rasa/core/policies/ensemble.py
```diff
@@ -69,6 +70,15 @@ def __init__(
         self._check_priorities()
         self._check_for_important_policies()
 
+        self._rule_only_slots, self._rule_only_loops = self._get_rule_only_slots_loops()
```
wochinge, line 73: out of the scope of this PR: In my opinion all the content of this `__init__` should go to the constructor of `SimplePolicyEnsemble`. In my opinion there is too much specific behavior in the abstract interface.

rasa/core/policies/ensemble.py
```diff
@@ -204,6 +214,10 @@ def train(
             self.action_fingerprints = rasa.core.training.training.create_action_fingerprints(
                 training_trackers, domain
             )
+            (
+                self._rule_only_slots,
+                self._rule_only_loops,
+            ) = self._get_rule_only_slots_loops()
```
wochinge, line 220: why do we need to do this after the training?

rasa/core/policies/ensemble.py
```diff
@@ -705,7 +719,11 @@ def _get_prediction(
             and "interpreter" in arguments
         ):
             prediction = policy.predict_action_probabilities(
-                tracker, domain, interpreter
+                tracker,
+                domain,
+                interpreter,
+                rule_only_slots=self._rule_only_slots,
+                rule_only_loops=self._rule_only_loops,
```
wochinge, line 726: we should mention these two new fields in the changelog

rasa/core/policies/memoization.py
```diff
@@ -195,8 +195,22 @@ def _recall_states(self, states: List[State]) -> Optional[Text]:
         return self.lookup.get(self._create_feature_key(states))
 
     def recall(
-        self, states: List[State], tracker: DialogueStateTracker, domain: Domain
+        self,
+        states: List[State],
+        tracker: DialogueStateTracker,
+        domain: Domain,
+        **kwargs: Any,
```
wochinge, line 202: Our out-of-the-box policies should unwrap default `kwargs` and make them explicit as soon as possible (which would be `policy.predict_action_probabilities`. This makes the code more explicit and also enables type checking.

rasa/core/policies/policy.py
```diff
@@ -195,6 +201,70 @@ def featurize_for_training(
 
         return state_features, label_ids, entity_tags
 
+    def prediction_states(
+        self,
+        tracker: DialogueStateTracker,
+        domain: Domain,
+        use_text_for_last_user_input: bool = False,
+        **kwargs: Any,
```
wochinge, line 209: Our out-of-the-box policies should unwrap default `kwargs` and make them explicit as soon as possible (which would be policy.predict_action_probabilities. This makes the code more explicit and also enables type checking

rasa/core/policies/policy.py
```diff
@@ -195,6 +201,70 @@ def featurize_for_training(
...
+        return self.featurizer.prediction_states(
+            [tracker],
+            domain,
+            use_text_for_last_user_input=use_text_for_last_user_input,
+            ignore_rule_only_turns=self.supported_data() == SupportedData.ML_DATA,
+            rule_only_slots=kwargs.get("rule_only_slots"),
+            rule_only_loops=kwargs.get("rule_only_loops"),
+        )[0]
+
+    def featurize_for_prediction(
```
wochinge, line 231: this can be `protected`, no?
```suggestion
    def _featurize_for_prediction(
```

rasa/shared/core/events.py
```diff
@@ -723,7 +723,10 @@ def __hash__(self) -> int:
 
     def __eq__(self, other: Any) -> bool:
         """Compares this event with another event."""
-        return isinstance(other, EntitiesAdded)
+        if not isinstance(other, EntitiesAdded):
```
wochinge, line 726: 🚀

rasa/shared/core/generator.py
```diff
@@ -107,7 +107,15 @@ def _unfreeze_states(frozen_states: Deque[FrozenState]) -> List[State]:
             for frozen_state in frozen_states
         ]
 
-    def past_states(self, domain: Domain) -> List[State]:
+    def past_states(self, domain: Domain, **kwargs: Any) -> List[State]:
```
wochinge, line 110: Please also document `kwargs`

wochinge: and ideally also unwrap `kwargs` to its single arguments.

rasa/shared/core/trackers.py
```diff
@@ -273,16 +274,30 @@ def freeze_current_state(state: State) -> FrozenState:
...
 
         Args:
-            domain: a :class:`rasa.shared.core.domain.Domain`
+            domain: a :class:`rasa.shared.core.domain.Domain`.
+            ignore_rule_only_turns: If True ignore dialogue turns that are present
+                only in rules.
+            rule_only_slots: The list of slot names,
+                which only occur in rules but not in stories.
+            rule_only_loops: The list of loop names,
+                which only occur in rules but not in stories.
```
wochinge, line 293: optional: Let's avoid encoding the types (e.g. `list`) in the docstring. The type annotations do a better job at this and don't outdate when we e.g. change this to a `Set`. To indicate a collection I'd use the plural.

rasa/shared/core/trackers.py
```diff
@@ -556,7 +574,9 @@ def _undo_till_previous_loop_execution(
             if isinstance(e, ActionExecuted) and e.action_name == loop_action_name:
                 break
 
-            if isinstance(e, (ActionExecuted, UserUttered)):
+            if isinstance(
+                e, (ActionExecuted, UserUttered, DefinePrevUserUtteredFeaturization,),
```
wochinge, line 578: Great catch! Can we please cover this with a test?

tests/core/policies/test_rule_policy.py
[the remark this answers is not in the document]
wochinge: yep

tests/core/policies/test_rule_policy.py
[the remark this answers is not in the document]
wochinge: yep

tests/core/policies/test_rule_policy.py
```diff
@@ -1936,3 +1944,554 @@ def test_predict_nothing_if_fallback_disabled():
...
+        },
+    ]
+
+
+def test_hide_rule_turn_with_slots():
+    some_action = "some_action"
+    some_other_action = "some_other_action"
+    some_intent = "some_intent"
+    some_other_intent = "some_other_intent"
+    some_slot = "some_slot"
```
wochinge, line 2027: optional: Very verbose, but makes the test easier to read, what do you think? 
```suggestion
    slot_which_is_only_in_rule = "some_slot"
```

tests/core/policies/test_rule_policy.py
```diff
@@ -1936,3 +1944,554 @@ def test_predict_nothing_if_fallback_disabled():
...
+    some_slot_value = "value1"
+    some_other_slot = "some_other_slot"
```
wochinge, line 2029: optional: Very verbose, but makes the test easier to read, what do you think?

```suggestion
    slot_which_is_also_in_story = "some_other_slot"
```

tests/core/policies/test_rule_policy.py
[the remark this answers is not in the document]
wochinge: How about:
1. Train policies 
2. Put them in a `Ensemble`
3. Create processor
4. Call `processor.predict_next_with_tracker` and make sure ML policy gets it right?

rasa/core/policies/rule_policy.py
```diff
@@ -524,27 +595,78 @@ def _run_prediction_on_trackers(
...
+        and creates the lookup for rules that are not present in the stories.
+
+        Args:
+            rule_trackers: The list of the rule trackers.
+            all_trackers: The list of all trackers.
+            domain: The domain.
+            interpreter: Interpreter which can be used by the polices for featurization.
+
+        Returns:
+             The list of rules that are not present in the stories.
```
wochinge, line 668: let's avoid duplicating the type in the docstring
```suggestion
             Rules that are not present in the stories.
```

rasa/core/policies/rule_policy.py
```diff
@@ -761,14 +920,22 @@ def _find_action_from_loop_happy_path(
...
             logger.debug(
                 f"Predicted '{ACTION_LISTEN_NAME}' after loop '{active_loop_name}'."
             )
-            return ACTION_LISTEN_NAME
+            return (
+                ACTION_LISTEN_NAME,
+                LOOP_RULES
+                + active_loop_name
+                + LOOP_RULES_SEPARATOR
+                + ACTION_LISTEN_NAME,
```
wochinge, line 935: how about using an f-string?

rasa/core/policies/rule_policy.py
```diff
@@ -987,3 +1174,70 @@ def _metadata(self) -> Dict[Text, Any]:
...
+        Returns:
+            A tuple of lists of slots and loops that are used only in rule data.
+        """
+        return (
+            self.lookup.get(RULE_ONLY_SLOTS, []),
+            self.lookup.get(RULE_ONLY_LOOPS, []),
+        )
+
+
+class RulePolicyPrediction(PolicyPrediction):
```
wochinge, line 1190: why do we need this?

rasa/shared/core/domain.py
```diff
@@ -1074,15 +1074,109 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
...
+            if not last_ml_state.get(rasa.shared.core.constants.USER) and state.get(
+                rasa.shared.core.constants.USER
+            ):
+                del state[rasa.shared.core.constants.USER]
+            elif last_ml_state.get(rasa.shared.core.constants.USER):
+                state[rasa.shared.core.constants.USER] = last_ml_state[
+                    rasa.shared.core.constants.USER
+                ]
+
     def states_for_tracker_history(
```
wochinge, line 1110: can we please add some tests for this in `test_domain`?

rasa/core/policies/rule_policy.py
Ghostvv: I didn't want to add prediction source to general `PolicyPrediction` object, since it is only used inside `RulePolicy`

rasa/core/policies/ensemble.py
Ghostvv: added `and`

rasa/shared/core/domain.py
Ghostvv: we can, but effectively, they will duplicate the tests in `test_rule_policy.py`

rasa/core/policies/policy.py
```diff
@@ -226,6 +302,9 @@ def predict_action_probabilities(
             domain: the :class:`rasa.shared.core.domain.Domain`
             interpreter: Interpreter which may be used by the policies to create
                 additional features.
+        Keyword Args:
```
wochinge, line 305: let's add these as proper parameters in the method signature
```suggestion
```

rasa/core/policies/memoization.py
```diff
@@ -222,12 +239,25 @@ def predict_action_probabilities(
         interpreter: NaturalLanguageInterpreter,
         **kwargs: Any,
     ) -> PolicyPrediction:
+        rule_only_slots = kwargs.get("rule_only_slots")
```
wochinge, line 242: let's use proper parameters instead of getting this from kwargs

rasa/core/policies/sklearn_policy.py
```diff
@@ -284,7 +284,15 @@ def predict_action_probabilities(
         interpreter: NaturalLanguageInterpreter,
         **kwargs: Any,
     ) -> PolicyPrediction:
-        X = self.featurizer.create_state_features([tracker], domain, interpreter)
+        rule_only_slots = kwargs.get("rule_only_slots")
```
wochinge, line 287: let's use proper parameters instead of gettings from the kwargs

rasa/core/policies/ted_policy.py
```diff
@@ -606,12 +632,15 @@ def predict_action_probabilities(
 
         See the docstring of the parent class `Policy` for more information.
         """
+        rule_only_slots = kwargs.get("rule_only_slots")
```
wochinge, line 635: let's use proper parameters instead of gettings from the kwargs

rasa/shared/core/domain.py
```diff
@@ -1074,15 +1074,107 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
...
+    def _remove_rule_only_features(
+        state: State,
+        rule_only_slots: Optional[List[Text]],
+        rule_only_loops: Optional[List[Text]],
+    ) -> None:
+        # remove slots which only occur in rules but not in stories
+        if rule_only_slots:
+            for slot in rule_only_slots:
+                if state.get(rasa.shared.core.constants.SLOTS, {}).get(slot):
+                    del state[rasa.shared.core.constants.SLOTS][slot]
```
wochinge, line 1087:
```suggestion
                state.get(rasa.shared.core.constants.SLOTS, {}).pop(slot, None)
```

rasa/core/policies/rule_policy.py
Ghostvv: ok

rasa/core/policies/policy.py
Ghostvv: then pycharm complains

rasa/core/policies/memoization.py
Ghostvv: but then I'll change the signature of `predict_action_probabilities`? wouldn't it be breaking change?

wochinge: Nope, as they should already have the `kwargs` param in place which would catch the new param

tests/core/training/test_interactive.py
```diff
@@ -582,14 +582,15 @@ async def test_write_domain_to_file_with_form(tmp_path: Path):
...
-    - {form_name}
-    intents:
-    - greet
-    """
+        version: "2.0"
+        actions:
+        - utter_greet
+        - utter_goodbye
+        forms:
+          {form_name}:
```
wochinge, line 590: thank you!

wochinge (commented):
Nice work 💯 

One more small things: Can you please adapt the remaining policies to have the new params in `predict_action_probabilities`. As this is the official codebase we shouldn't rely on the `kwarg` mechanism in my opinion but have the exactly same method signature as in the parent class.

Ghostvv:
> Can you please adapt the remaining policies to have the new params in predict_action_probabilities

I could, but the problem is that passing `rule_only_slots, rule_only_loops` to predict feels like they don't belong there, it feels like hack, because it is very much kind of single use case rules related and doesn't feel general enough to be part of such a high level public API. Do you have an idea how can we pass them differently?

wochinge:
How about defining a new `set_rule_only_data` on `Policy` which is called in the constructor of `PolicyEnsemble`? It would be `pass` by default but any policies using it could do so. The alternative is to store it in the domain which would make sense as the domain already stores loops and slots.

Ghostvv:
I like the idea of having some place for policies shared info. I don't think domain is a good place, since it is domain independent. Same domains can have different only rule slots and loops. Actually, they can be different after each retrain

wochinge:
But the domain currently contains all the shared information already, no?

> Actually, they can be different after each retrain

What do you mean? You mean in case you add another story?

Ghostvv:
> What do you mean? You mean in case you add another story?

yes

Ghostvv:
also, updating domain retrospectively after training doesn't feel right

Ghostvv:
@wochinge could you give it a review? is it ready to merge?

rasa/core/policies/policy.py
```diff
@@ -120,12 +126,17 @@ def __init__(
...
         self.priority = priority
         self.finetune_mode = should_finetune
+        self._rule_only_data = {}
 
     @property
     def featurizer(self):
         """Returns the policy's featurizer."""
         return self.__featurizer
 
+    def set_rule_only_data(self, rule_only_data: Dict[Text, Any]) -> None:
```
wochinge, line 136: how about using something which generalizes better?

```python
def set_shared_policy_state(**kwargs: Any) -> None
```

Advantages:
* no deprecation is required if we want to pass more information / different information
* users can use this for their own policies (e.g. setting some shared state originating from their policies)

Ghostvv: users wouldn't be able to use it, since ensemble will only pass `rule_only_data`

Ghostvv: renamed the method

rasa/core/policies/policy.py
```diff
@@ -120,12 +126,17 @@ def __init__(
...
         self.priority = priority
         self.finetune_mode = should_finetune
+        self._rule_only_data = {}
 
     @property
     def featurizer(self):
         """Returns the policy's featurizer."""
         return self.__featurizer
 
+    def set_shared_policy_states(self, **kwargs: Any) -> None:
```
wochinge, line 136: can we please mention this change in the changelog?

rasa/shared/core/domain.py
```diff
@@ -1131,24 +1131,115 @@ def get_active_states(
...
+                        rasa.shared.core.constants.PREVIOUS_ACTION
+                    ] = last_ml_action_sub_state
+
+            states.append(self._clean_state(state))
+
+        return states
 
     def slots_for_entities(self, entities: List[Dict[Text, Any]]) -> List[SlotSet]:
-        """Returns `SlotSet` events for extracted entities."""
+        """Find which slots should be auto fill with entities.
```
wochinge, line 1235: I actually prefer the old docstring or something like this:
```suggestion
        """Creates slot events for entities if auto-filling is enabled.
```

rasa/shared/core/generator.py
```diff
@@ -107,7 +107,19 @@ def _unfreeze_states(frozen_states: Deque[FrozenState]) -> List[State]:
             for frozen_state in frozen_states
         ]
 
-    def past_states(self, domain: Domain) -> List[State]:
+    def past_states(
+        self, domain: Domain, ignore_rule_only_turns: bool = False
+    ) -> List[State]:
+        """Generate the past states of this tracker based on the history.
```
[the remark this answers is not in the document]
wochinge: @Ghostvv Can you change this back to add the `s`?

rasa/shared/core/trackers.py
```diff
@@ -288,19 +289,30 @@ def freeze_current_state(state: State) -> FrozenState:
...
     def past_states(
-        self, domain: Domain, omit_unset_slots: bool = False,
+        self,
+        domain: Domain,
+        omit_unset_slots: bool = False,
+        ignore_rule_only_turns: bool = False,
+        rule_only_data: Optional[Dict[Text, Any]] = None,
     ) -> List[State]:
-        """Generates the past states of this tracker based on the history.
+        """Generate the past states of this tracker based on the history.
```
wochinge, line 298:
```suggestion
        """Generates the past states of this tracker based on the history.
```

rasa/core/policies/rule_policy.py
```diff
@@ -524,27 +587,67 @@ def _run_prediction_on_trackers(
...
+
+    def _analyze_rules(
         self,
         rule_trackers: List[TrackerWithCachedStates],
         all_trackers: List[TrackerWithCachedStates],
         domain: Domain,
-        interpreter: NaturalLanguageInterpreter,
-    ) -> None:
+    ) -> List[Text]:
+        """Analyze learned rules by running prediction on training trackers.
```
wochinge, line 637:
```suggestion
        """Analyzes learned rules by running prediction on training trackers.
```

rasa/core/policies/policy.py
Ghostvv: do you mean the fact that we added it? it's not really a change, no?

wochinge: it's a change to a public interface of Rasa Open Source so users might also want to implement it for their own `Policy`s

wochinge (approved):
👍🏻

Ghostvv:
@wochinge do you have an idea why test docs fail? I tried restarting the docs several times

wochinge:
due to https://github.com/RasaHQ/rasa/pull/8272

PR #7716, opened 2021-01-13
Add `split_entities_config` to TEDPolicy

### LINKED ISSUE #7707: Add `split_entities_config` to `TEDPolicy` for e2e
**Rasa version**:
2.2.2
**Python version**:
3.7.2
OSX

**Issue**:

`TEDPolicy` needs a `split_entities_config` value, to mirror `CRFEntityExtractor` and `DIET`

When we call `convert_predictions_into_entities`, the value of split_entities_config is `None`, which is only a problem if we then call `EntityExtractor._check_is_single_entity`. I think this is somewhat rare, which is why the bug went unnoticed so far.

`TEDPolicy` should have its own value for this, with the same default as `DIET,` and this parameter should be configurable in the same way.

koernerfelicia:
**Proposed changes**:
Closes #7707


docs/docs/policies.mdx
```diff
@@ -176,6 +176,24 @@ If you want to fine-tune your model, start by modifying the following parameters
   set `weight_sparsity` to 1 as this would result in all kernel weights being 0, i.e. the model is not able
   to learn.
 
+* `split_entities_by_comma`:
```
Ghostvv, line 179: please add this param to the param table as well

rasa/core/policies/ted_policy.py
```diff
@@ -280,6 +286,16 @@ def _standard_featurizer(max_history: Optional[int] = None) -> TrackerFeaturizer
             SingleStateFeaturizer(), max_history=max_history
         )
 
+    def init_split_entities(self, split_entities_config):
```
Ghostvv, line 289: is the same code used in DIET? could you please move it to `train_utils` to avoid code duplication

koernerfelicia:
@tttthomasssss and @Ghostvv, I applied the suggestions. One thing I'm not so sure whether ResponseSelector should have a default value for `split_entities_config`

Ghostvv:
ResponseSelector doesn't extract entities, so I'd say that it shouldn't have it. @tttthomasssss could you please review the PR, since the original config is yours

rasa/nlu/extractors/extractor.py
```diff
@@ -47,17 +70,18 @@ def add_processor_name(self, entity: Dict[Text, Any]) -> Dict[Text, Any]:
...
-        else:
-            split_entities_config[SPLIT_ENTITIES_BY_COMMA] = self.defaults[
-                SPLIT_ENTITIES_BY_COMMA
-            ]
-        return split_entities_config
+        default_value = (
+            self.defaults[SPLIT_ENTITIES_BY_COMMA]
+            if SPLIT_ENTITIES_BY_COMMA in self.defaults
+            else SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE
+        )
```
tttthomasssss, line 81: I guess that could be simplified to:
```
default_value = self.defaults.get(SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE)
```

rasa/utils/train_utils.py
```diff
@@ -335,3 +340,12 @@ def override_defaults(
                 config[key] = custom[key]
 
     return config
+
+
+def init_split_entities(split_entities_config, default_split_entity):
```
tttthomasssss, line 345: Could you add type annotations for the arguments and the return value?

rasa/nlu/extractors/extractor.py
```diff
@@ -47,17 +70,18 @@ def add_processor_name(self, entity: Dict[Text, Any]) -> Dict[Text, Any]:
         return entity
 
     def init_split_entities(self):
```
tttthomasssss, line 72: Could you add a type annotation for the return value?

docs/docs/policies.mdx
```diff
@@ -320,6 +338,15 @@ However, additional parameters exist that can be adapted.
...
+| split_entities_by_comma               | True                   | Splits a list of extracted entities by comma to treat each   |
+|                                       |                        | one of them as a single entity. Can either be `True`/`False` |
+|                                       |                        | globally, or set per entity type, such as:                   |
+|                                       |                        | ```                                                          |
+|                                       |                        | - name: TEDPolicy                                            |
+|                                       |                        |   split_entities_by_comma:                                   |
+|                                       |                        |     address: True                                            |
+|                                       |                        | ```                                                          |
++---------------------------------------+------------------------+--------------------------------------------------------------+
```
tttthomasssss, line 349: I don't think we need the docs in the hidden table, given its already described in the "visible" docs (Neither of the other options in the above list are listed in the table below).

tttthomasssss: Ignore me, I have just seen @Ghostvv's comment above and have checked the table again

docs/docs/policies.mdx
tttthomasssss: Could you add an example for an address in addition to the `banana,apple` example in order to illustrate when splitting an entity does not make sense?

tests/utils/test_train_utils.py
```diff
@@ -26,3 +33,31 @@ def test_align_token_features():
...
+            {"address": False, "ingredients": True},
+            {
+                "address": False,
+                "ingredients": True,
+                SPLIT_ENTITIES_BY_COMMA: SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE,
+            },
+        ),
+    ],
+)
+def test_init_split_entities_config(
```
tttthomasssss, line 55: Is there a small real-life use-case for the functionality with `TEDPolicy` that we could add to the tests?

tttthomasssss:
@koernerfelicia great work 🚀 ! 

I've left a few comments.

tests/utils/test_train_utils.py
koernerfelicia: I think the functionality should be covered by the existing test -- there shouldn't be a difference between the behaviour for `TEDPolicy` or one of the `EntityExtractors`, because we'd use the split configs only in e2e, when `TEDPolicy` extracts entities

tttthomasssss: OK - great!

tttthomasssss (approved):
Awesome, thank you for addressing my comments - LGTM 👍 !

PR #7723, opened 2021-01-13
Adding Sara Chatroom Integration to Docs 

b-quachtran:
**Proposed changes**:
- ...


b-quachtran:
@lunelson I took a stab at creating the custom Layout, but not entirely sure where to go from here. Any pointers on if I'm on the right track?

lunelson:
@b-quachtran Hey, so I've got this working now with their default chatbot, I recommend you try it out and see now if you can connect it to a real rasa bot. 

Right now it just mounts right away, but probably you want to implement something like a button to launch the bot, and have it otherwise be hidden, Also, it should probably be collapsible, which currently it doesn't seem to be. You'll have to look in to the documentation as to how their component can be customized I guess.

lunelson:
Correction: this is only working in development right now, it doesn't seem to come through in the build, I'm not yet sure why

b-quachtran:
@lunelson Awesome, yes the UI is loading for me in docs. I'll work on getting this integrated with the Sara backend and tweaking a few things like the initial payload. Thanks!

lunelson:
@b-quachtran I assume you closed this because we're going to use the `rasa-webchat` package instead...

b-quachtran:
Yes, the `rasa-webchat` is going to be used instead of the Botfront widget

PR #7747, opened 2021-01-18
Docs: Training a Model in Any Language

koaning:
**Proposed changes**:

The docs currently say "Training a Model in Any Language" . This is not correct so in this PR I've changed it to "Training a Model in Many Languages". 

This PR only contains minor changes, my goal is to make more PRs with more elaborate advice for languages that do not work with whitespacetok/spacy. I want to double-check with the research/devrel/product teams first though which tools we would/would not recommend. 

@akelad @melindaloubser1 pinging you as you responded in the slack thread.


melindaloubser1 (approved):

akelad (changes requested):
eh the title "Training a model in Many Languages" sounds kind of weird. Can we not just leave the original title and say that this is the pipeline to use if your model is whitespace tokenizable, and if not then you should provide your own tokenizer and it will also work

akelad (approved):

docs/docs/language-support.mdx
```diff
@@ -12,9 +12,10 @@ from scratch with the data you provide.
 In addition, we also support pre-trained word embeddings such as spaCy. For information on
 what pipeline is best for your use case, check out [choosing a pipeline](./tuning-your-model.mdx#how-to-choose-a-pipeline).
 
-## Training a Model in Any Language
+## Training a Model in Any Languages
```
melindaloubser1, line 15: Is this a typo or is it intentionally "any languages" now?

koaning: We came to the conclusion that "many" sounded a bit strange. Therefore we figured just adding the "whitespace"-tokenizable part is "good enough for now". I'll likely do another iteration on this when the spaCy 3.0 stuff is merged. SpaCy will also start supporting Chinese, Japanese, Thai, Ukranian, Korean and Vietnamese.

Issue #7683, opened 2021-01-06
The format for exporting "any" type slots (in JSON or list) through migration or interactive learning blocks any subsequent training


ganeshviyer:
**Rasa version**: 2.2.2

**Rasa SDK version** (if used & relevant): 2.2

**Rasa X version** (if used & relevant): not relevant here

**Python version**: 3.7.9

**Operating system** (windows, osx, ...): Mac OS Big Sur 11.1

**Issue**: If any of the slots of type "any" are in JSON or in list format, the format in YAML through which the slot values are exported through the automated migration (from 1.x to 2.x) or through Interactive Learning, fails any subsequent training.

The JSON or the list value of the slot is converted to a YAML tree structure that looks something like this under the `slot_was_set` node in stories. The slot `SLOT_NAME` in the example below is used as a JSON during the conversations by custom actions.

*   action: previous_action
*   slot_was_set:
*   SLOT_NAME:
     SLOT_KEY_1:
     slot_property_1: <value>
     slot_property_2: <value>
     slot_property_list_3:
*   slot_property_3_1: <value>
     slot_property_3_2: <value>
*   slot_property_3_1: <value>
     slot_property_3_2: <value>
*   slot_property_3_1: <value>
     slot_property_3_2: <value>
*   slot_property_3_1: <value>
     slot_property_3_2: <value>
     SLOT_KEY_2
     ...

The issue is described in detail here - [https://forum.rasa.com/t/the-slot-type-any-does-not-seem-to-work-correctly/38674](https://forum.rasa.com/t/the-slot-type-any-does-not-seem-to-work-correctly/38674)

**Error (including full traceback)**: The errors I get look like

 Value 'ordereddict([('SLOT_NAME', ordereddict([('SLOT_KEY_1', ordereddict([('slot_property_1, '<value>'), ('slot_property_2', '<value>'), ('slot_property_list_3', [ordereddict(<span class="error">[('slot_property_3_1', '
 <value>
  '), ('slot_property_3_2', '
  <value>
   ')]
  </value>
 </value></span>), ordereddict(<span class="error">[('slot_property_3_1', '
 <value>
  '), ('slot_property_3_2', '
  <value>
   ')]
  </value>
 </value></span>), ordereddict(<span class="error">[('slot_property_3_1', '
 <value>
  '), ('slot_property_3_2', '
  <value>
   ')]
  </value>
 </value></span>), ordereddict(<span class="error">[('slot_property_3_1', '
 <value>
[...]

PR #7748, opened 2021-01-18
Updated docs for slot type any

ganeshviyer:
**Proposed changes**:
- The update is regarding insufficient documentation for slot type "any" where the the slot can carry a JSON or a list value.  This is not compatible when performing the migration or when generating new stories through `rasa interactive`.  The resulting training file (stories.yml) breaks any subsequent training.
- The corresponding issue has been discussed here in the forums - https://forum.rasa.com/t/the-slot-type-any-does-not-seem-to-work-correctly/38674/3
- The issue on Github has been raised here - https://github.com/RasaHQ/rasa/issues/7683 and is still open.

### DIFF
--- a/docs/docs/domain.mdx
+++ b/docs/docs/domain.mdx
@@ -343,8 +343,11 @@ information before it is able to predict the weather.
 * **Description**
 
   Slots of type `any` are always ignored during conversations. The property
-  `influence_conversation` cannot be set to `true` for this slot type. If you want to
-  store a custom data structure which should influence the conversation, use a
+  `influence_conversation` cannot be set to `true` for this slot type. 
+  
+  In stories, while annotating slots of type `any` that are in a list or dictionary format, include the set slot value inside single quotes. See example here in [stories](./writing-stories.mdx).
+  
+  If you want to store a custom data structure which should influence the conversation, use a
   [custom slot type](domain.mdx#custom-slot-types).
 
 #### Custom Slot Types
--- a/docs/docs/writing-stories.mdx
+++ b/docs/docs/writing-stories.mdx
@@ -105,6 +105,7 @@ stories:
    - action: action_check_profile
    - slot_was_set:
      - premium_account: true
+     - account_data: '{"id":"zpNi3JA2", "fullName":"Ada Lovelace", "accountRenewalDate":"October 3, 2021"}'
    - action: utter_welcome_premium
 
 - story: Welcome message, basic user
@@ -113,6 +114,7 @@ stories:
    - action: action_check_profile
    - slot_was_set:
      - premium_account: false
+     - account_data: '{"id":"3JcoaEa4", "fullName":"Jack Harkness"}'
    - action: utter_welcome_basic
    - action: utter_ask_upgrade
 ```


sara-tagger:

docs/docs/domain.mdx
```diff
@@ -343,8 +343,11 @@ information before it is able to predict the weather.
 * **Description**
 
   Slots of type `any` are always ignored during conversations. The property
-  `influence_conversation` cannot be set to `true` for this slot type. If you want to
-  store a custom data structure which should influence the conversation, use a
+  `influence_conversation` cannot be set to `true` for this slot type. 
+  
+  In stories, while annotating slots of type `any` that are in a list or dictionary format, include the set slot value inside single quotes. See example here in [stories](./writing-stories.mdx).
+
```
melindaloubser1, line 349: Slots of type `any` should not be annotated in stories at all, since they are not influencing the conversation

docs/docs/writing-stories.mdx
```diff
@@ -105,6 +105,7 @@ stories:
    - action: action_check_profile
    - slot_was_set:
      - premium_account: true
+     - account_data: '{"id":"zpNi3JA2", "fullName":"Ada Lovelace", "accountRenewalDate":"October 3, 2021"}'
```
melindaloubser1, line 108: Same for this - a slot of type any should not be included in a story, as it won't make any difference

melindaloubser1 (changes requested):

docs/docs/domain.mdx
ganeshviyer: I was told differently in the rasa forums - https://forum.rasa.com/t/the-slot-type-any-does-not-seem-to-work-correctly/38674/3

samsucik: @melindaloubser1 I think this is more of a gray area -- they slots can be there and that shouldn't break anything. It's just that -- for the purpose of training -- the slots are useless.

melindaloubser1: Fair enough, then the documentation can rather reflect this since it will apply to other slots with that format in them e.g. adding this sentence to the top of the slots doc (not under any):

>You can include a slot whose value is in dictionary or list format in stories by wrapping the value in single quotes e.g.
```
story: story with custom slot in dictionary format
steps: 
   - slot_was_set:
     - account_data: '{"id":"zpNi3JA2", "fullName":"Ada Lovelace", "accountRenewalDate":"October 3, 2021"}'
```
> Keep in mind you only need to include slot values in your story if the slot has `influence_conversation` set to `true`.

samsucik: I agree with adding it somewhere more general and also with adding the note about `influence_conversation` (though we have to word it carefully because influencing is often turned on by default).

Also, I think what we need even more is adding under the relevant slot type sections (`list`, `any` and custom slots) some concrete examples of slot values (especially lists/dicts) written in stories. Perhaps this would be useful even for simpler slot types, e.g. to clarify if `float` slot values should be quoted or not...

melindaloubser1: I think we should  do that in a separate PR 👍 @ganehviyer do you want to take care of moving the info to the general section and adding the note?

melindaloubser1:
@ganeshviyer Sorry I couldn't tag you in the thread, not sure why - does the discussion make sense to you? Would you like to take care of moving the info up the page and adding a note?

PR #7749, opened 2021-01-18
fix e2e overriding loop prediction

Ghostvv:
**Proposed changes**:
- fix e2e overriding loop prediction


Ghostvv:
@wochinge @joejuzl do you know why `windows` tests fail? I don't seem to add any changes related to spacy

joejuzl:
> @wochinge @joejuzl do you know why windows tests fail? I don't seem to add any changes related to spacy

I think it must be unrelated to your changes, maybe some network issue when pulling in the windows dependencies.

Ghostvv:
@joejuzl three times in a raw. Do you have an idea how to solve it?

joejuzl:
@Ghostvv this seems to be a problem on master too. I created this [test PR](https://github.com/RasaHQ/rasa/pull/7752) and it's failing in the same way. I suspect it's something to do with pypi.

changelog/7749.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix the bug when `RulePolicy` handling loop predictions are overwritten by e2e `TEDPolicy`
```
wochinge, line 1:
```suggestion
Fix the bug when `RulePolicy` handling loop predictions are overwritten by e2e `TEDPolicy`.
```

rasa/core/policies/policy.py
```diff
@@ -399,7 +400,7 @@ def __init__(
         policy_priority: int = 1,
         events: Optional[List[Event]] = None,
         optional_events: Optional[List[Event]] = None,
-        is_end_to_end_prediction: bool = False,
+        is_end_to_end_prediction: Optional[bool] = False,
```
wochinge, line 403: I like the limited scoped of the fix but I don't like the additional meaning in the boolean value which isn't obvious immediately (the meaning of `is_end_to_end_prediction=None` only becomes obvious by looking onto more code / comments). I'd prefer one of these options
- adding another field `is_loop_prediction` to `PolicyPrediction`
- adding a field `prediction_type` to `PolicyPrediction` with an `Enum` type `PredictionType`. This is more scalable if we add further prioritization of special cases later (e.g. for fallbacks (just an example)) but would require a deprecation of  `is_end_to_end_prediction` which makes it slightly more work.

rasa/core/policies/ensemble.py
```diff
@@ -556,6 +556,20 @@ def _is_form_policy(policy_name: Text) -> bool:
 
         return policy_name.endswith("_" + FormPolicy.__name__)
 
+    @staticmethod
+    def _prediction_type(predictions: Dict[Text, PolicyPrediction]) -> Optional[bool]:
```
wochinge, line 560: a test for this logic would be great

rasa/core/policies/policy.py
Ghostvv: effectively we have `3` types of prediction, regardless of loop of fallback, etc: `without using user input`, `with user intent`, `with user text`. The reason I like reusing the same bool, as both `without using user input` and `with user intent` require adding `DefinePrevUserUtteredFeaturization(False)`. Since it is all about the user I prefer current way, since it precisely represents that, where `None` mean that we don't know

rasa/core/policies/ensemble.py
Ghostvv: added

rasa/core/policies/policy.py
wochinge: @joejuzl @alwx Any strong opinion? I'd still favor explicitness as it makes the logic easier to understand without too much further background.

joejuzl: Do we only ever convert it to a boolean? Or do we ever branch on it being `None`

Ghostvv: in `ensemble.py`, we branch if it is `None`

rasa/core/policies/ensemble.py
```diff
@@ -556,6 +556,20 @@ def _is_form_policy(policy_name: Text) -> bool:
...
+        # No user input prediction overrule all other predictions.
+        if any(
+            prediction.is_end_to_end_prediction is None
```
wochinge, line 563: @joejuzl Here we branch on `None` as `None` means it's a `loop` prediction

joejuzl: ok so `use_only_end_to_end` is:
- `None` if at least one `prediction.is_end_to_end_prediction` is `None`
- `False` if all are `False`
- `True` if at least one is `True` and none are `None`.

Then, yes I do find that hard to follow.

rasa/core/policies/policy.py
wochinge: @Ghostvv Considering [Joe's comment](https://github.com/RasaHQ/rasa/pull/7749#discussion_r563863122) I'd gone with one of my two solutions then . Do you expect further prediction types or or do you expect it to stay these 3 for a while?

Ghostvv: as you know, it's hard to predict 🙈 But I think we can go with three for now

Ghostvv: I'll update

Ghostvv: I added the unnecessary `bool`

rasa/core/policies/policy.py
```diff
@@ -417,13 +421,16 @@ def __init__(
                 you return as they can potentially influence the conversation flow.
             is_end_to_end_prediction: `True` if the prediction used the text of the
                 user message instead of the intent.
+            is_no_user_prediction: `True` if the prediction didn't use neither the text
```
joejuzl, line 424:
```suggestion
            is_no_user_prediction: `True` if the prediction uses neither the text
```
Double negative.

joejuzl (approved):
Looks good 👍

rasa/core/policies/ensemble.py
```diff
@@ -578,16 +578,29 @@ def _pick_best_policy(
 
         form_confidence = None
         form_policy_name = None
-        # End-to-end predictions overrule all other predictions.
-        use_only_end_to_end = any(
+        # different type of predictions have different priorities
```
wochinge, line 581: And here we are again - having different priorities ~per~ within policies 😆

rasa/core/policies/ensemble.py
```diff
@@ -578,16 +578,29 @@ def _pick_best_policy(
...
         )
-        policy_events = []
 
+        policy_events = []
         for policy_name, prediction in predictions.items():
             policy_events += prediction.events
 
-            if prediction.is_end_to_end_prediction != use_only_end_to_end:
+            # No user predictions overrule all other predictions.
+            if prediction.is_no_user_prediction != is_no_user_prediction:
```
wochinge, line 596: seems a tiny bit easier to understand for me, what do you think? 
```suggestion
            if is_no_user_prediction and not prediction.is_no_user_prediction:
```

rasa/core/policies/ensemble.py
```diff
@@ -578,16 +578,29 @@ def _pick_best_policy(
...
+                continue
+
+            # End-to-end predictions overrule all other predictions based on user input.
+            if (
+                not is_no_user_prediction
+                and prediction.is_end_to_end_prediction != is_end_to_end_prediction
```
wochinge, line 602: we don't have to check for `not is_no_user_prediction` again, do we?
```suggestion
                if is_end_to_end_prediction and not prediction.is_end_to_end_prediction:
```

rasa/core/policies/policy.py
```diff
@@ -420,6 +424,8 @@ def __init__(
                 you return as they can potentially influence the conversation flow.
             is_end_to_end_prediction: `True` if the prediction used the text of the
                 user message instead of the intent.
+            is_no_user_prediction: `True` if the prediction uses neither the text
+                of the user message nor the intent.
```
wochinge, line 428: let's please add something along these lines for clarification. I don't think it's clear for everyone otherwise.
```suggestion
                of the user message nor the intent. This is for the example the case for happy loop paths.
```

tests/core/test_ensemble.py
```diff
@@ -527,6 +530,32 @@ def test_end_to_end_prediction_supersedes_others(default_domain: Domain):
...
+    )
+    tracker = DialogueStateTracker.from_events("test", evts=[])
+
+    prediction = ensemble.probabilities_using_best_policy(
+        tracker, default_domain, RegexInterpreter()
+    )
+
+    assert prediction.max_confidence == expected_confidence
+    assert prediction.max_confidence_index == expected_action_index
+    assert prediction.policy_name == f"policy_2_{ConstantPolicy.__name__}"
```
wochinge, line 556: let's also check that `prediction. is_no_user_prediction` is set to `True` and `is_end_to_end_prediction` is `False`.

rasa/core/policies/ensemble.py
```diff
@@ -578,16 +578,29 @@ def _pick_best_policy(
...
             prediction.is_end_to_end_prediction for prediction in predictions.values()
         )
-        policy_events = []
 
+        policy_events = []
         for policy_name, prediction in predictions.items():
             policy_events += prediction.events
 
-            if prediction.is_end_to_end_prediction != use_only_end_to_end:
+            # No user predictions overrule all other predictions.
```
wochinge, line 595: let's add this example. I think the concept of `no user predictions` makes sense but I don't think it's clear to all others immediately without example.
```suggestion
            # No user predictions (e.g. loop predictions) overrule all other predictions.
```

wochinge (commented):
couple of minor comments - ready to go otherwise 💯Thanks for adding the new parameter to clarify the code.

rasa/core/policies/ensemble.py
Ghostvv: yes

rasa/core/policies/ensemble.py
Ghostvv: it doesn't seem to me easier to understand, took me some time to grasp that it would be the same thing

rasa/core/policies/ensemble.py
Ghostvv: we do. It's so happen that right now, `is_end_to_end_prediction` will be `False`. It happens because in reality we need three values rather than 2 booleans, but since you wanted 2 booleans, I'd keep the check here

Ghostvv: I still think having `None` would be significantly more clear implementation

PR #7754, opened 2021-01-19
Backport of aio-pika implementation to 1.10.X

b-quachtran:
**Proposed changes**:
- Backport of aio-pika implementation to 1.10.X
- explicitly cast int to bool in lockstore


wochinge (dismissed):
@alwx @b-quachtran In my opinion this is way beyond the scope of something which we can release in a `micro`. changed interfaces, completely new implementations, etc. I'm in strong favor of having one more look at the old implementation and fixing that instead. It worked for all these years - it must be possible to get that working.

b-quachtran:
@wochinge That makes sense, @alwx do you know if you'll be able to take a look at the old implementation's multi-processing issue? It's on the high priority side for a customer's current production deployment.

alwx:
We were discussing this issue and this PR and decided to reopen it. On one hand, we think that backporting features is what we shouldn't be doing, but in this case it's clear that actually approving this change would save us a lot of time and will allow customers to have a better and more stable implementation of Pika blocker.

b-quachtran:
Sounds good, let me know about the status when you get a chance to review 🙏

rasa/core/exporter.py
```diff
@@ -89,7 +90,10 @@ def publish_events(self) -> int:
                 logger.exception(e)
                 raise PublishingError(current_timestamp)
 
-        self.event_broker.close()
+        if asyncio.iscoroutinefunction(self.event_broker.close):
+            self.event_broker.close()
+        else:
+            await self.event_broker.close()
```
wochinge, line 96: that was actually a bug I just recently fixed 🙈 
```suggestion
        if asyncio.iscoroutinefunction(self.event_broker.close):
            await self.event_broker.close()
        else:
            self.event_broker.close()
```

wochinge (commented):
Can you please also add a changelog @b-quachtran ?

PR looks good! Good decision to remove the deprecation logs 👍🏻 How have you tested this?

wochinge:
Do you have time to do a release tomorrow or should I take over some steps @b-quachtran ?

b-quachtran:
Hey @wochinge, I won't have a chance to review for a release tomorrow so if you could take over some steps that'd be a big help. As far as testing, I updated some of the Pika test cases and verified that it was working with Homebrew's RabbitMQ server, but nothing beyond that.

wochinge:
Ok, I'll give it a quick test run as well and then merge + prepare the release 👍🏻

b-quachtran:
🙏  thanks!

wochinge:
seems the `poetry.lock` file was incorrect. Fixed it. Seems the testing was actually important 😆

wochinge (approved):
tested sending a message + waiting for a longer time and then sending a message to see if heartbeat works 👍🏻

PR #7758, opened 2021-01-20
Using the TrainingDataImporter to load the data in core test

Imod7:
**Proposed changes**:
- Using the TrainingDataImporter to load the data in core test.

**Note**
It is the proposed solution for the 1st part of this Issue [#5986](https://github.com/RasaHQ/rasa/issues/5986).
The 2nd part was solved from this [PR](https://github.com/RasaHQ/rasa/pull/7674).


Imod7:
The result from the tests in CI CD is the same that I have locally. There is one test that fails and it has input a `markdown` file. The passing tests have `yaml` input files. I am not quite sure how to correct this. With the previous method of loading the data, there is a `_get_reader` function [here](https://github.com/RasaHQ/rasa/blob/07fba6dbda7c7ac79409516005a34bf4bba72cef/rasa/shared/core/training_data/loading.py#L29). Maybe we need to make a function like this one for the TrainingDataImporter too or add the check only (if it is `markdown` or a `yaml` file) in one of the existing functions? If it already exists, my apologies I could not find it.

rasa/core/test.py
```diff
@@ -336,11 +337,13 @@ async def _create_data_generator(
...
     from rasa.shared.core.generator import TrainingDataGenerator
 
-    from rasa.core import training
-
-    story_graph = await training.extract_story_graph(
+    test_data_importer = TrainingDataImporter.load_from_dict(
+        training_data_paths=[resource_name]
+    )
+    story_graph = await test_data_importer.get_stories(
         resource_name, agent.domain, use_e2e
```
wochinge, line 344: the parameter usage doesn't look right as PyCharm gives a warning.

wochinge (changes requested):
Nice work in tracing the different test methods down to this isolated location 💯

rasa/core/test.py
Imod7: Ok, I will check!

Imod7: Thank you for pointed out the parameters usage issue. I changed the arguments and left only the `use_e2e` in the `get_stories` call but I still got the same error. 
After checking again the error message, I also saw that when running the `test_source_in_failed_stories` the assert  is related to the path itself. It looks like the difference is that I have the absolute path and the assert compares only with the root path. So I did a small change in the path [here](https://github.com/RasaHQ/rasa/blob/88d97e1d54ce79b9dac2d1091c7db7eb80fc6451/rasa/shared/data.py#L150).
As a note, the old way of loading the data was working because it does not use this function when running this test.  
After doing the change I run all the tests locally and it looks good. In the CI/CD I think the error is not related to the change but I will check again.

wochinge:
@Imod7 Seems something went wrong (seems like during a rebase) and the diff of the PR is huge now. Can you please clean that up? Happy to jump on a call if you need some help.

rasa/core/test.py
wochinge: > As a sidenote, pathlib (new standard library in Python 3.4) looks like a good alternative to the os.path module.

Yep, that's what we are trying to use in any new code 👍🏻 

Unfortunately it's very hard for me to reconstruct your issue due to the wrong diff of the PR 😬

wochinge: Thanks for the detailed explanation 💯

Imod7:
Closing this PR because of wrong diffs. Created a new one [here](https://github.com/RasaHQ/rasa/pull/7674)

PR #7759, opened 2021-01-20
Change cache key for poetry by a secret variable

tczekajlo:
**Proposed changes**:
- The only option to get rid of the old cache is to use a new key. Add the `POETRY_CACHE_VERSION` variable to the poetry cache key in order to control the key without additional steps that require commits


m-vdb (approved):
Smart 🚥

m-vdb:
btw @tczekajlo there might be other workflows with a poetry cache too, right?

tczekajlo:
> btw @tczekajlo there might be other workflows with a poetry cache too, right?

Oh yeah, I'll check it and I'll add changes if there is something else

Issue #7720, opened 2021-01-13
Train finetune flag throws tensorflow error


stefanvantchev:
**Rasa version**:
2.2.3

**Rasa SDK version** (if used & relevant):
2.2.0

**Rasa X version** (if used & relevant):
none

**Python version**:
3.7.3

**Operating system** (windows, osx, ...):
Linux-4.19.0-13-cloud-amd64-x86_64-with-debian-10.7

**Issue**:
```
ValueError: Tensor conversion requested dtype int64 for Tensor with dtype float32: <tf.Tensor 'batch_in_3:0' shape=(None,) dtype=float32>

```
**Error (including full traceback)**:
```
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/__main__.py", line 134, in <module>
    main()
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/__main__.py", line 116, in main
    cmdline_arguments.func(cmdline_arguments)
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/cli/train.py", line 205, in train_nlu
    finetuning_epoch_fraction=args.epoch_fraction,
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/train.py", line 711, in train_nlu
    finetuning_epoch_fraction=finetuning_epoch_fraction,
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/utils/common.py", line 308, in run_in_loop
    result = loop.run_until_complete(f)
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/train.py", line 757, in _train_nlu_async
    finetuning_epoch_fraction=finetuning_epoch_fraction,
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/train.py", line 818, in _train_nlu_with_validated_data
    **additional_arguments,
  File "/home/stefan.vantchev/.local/lib/python3.7/site-packages/rasa/nlu/train.py", line 116, in train
    interpreter = trainer.train(training_data, **kwargs)
[...]

PR #7769, opened 2021-01-21
Fix incremental training when no entities are present in the training data

dakshvar22:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/7720
- The error comes up only when there are no entities in the training data but `DIETClassifier` is configured to perform entity recognition. In such a case, the data signature of instance of `RasaModelData` should not contain `entities` as a key.

### DIFF
--- a/changelog/7720.bugfix.md
+++ b/changelog/7720.bugfix.md
@@ -1,0 +1,3 @@
+Fix incremental training for cases when training data does not contain entities but `DIETClassifier` is configured to perform entity recognition also.
+
+Now, the instance of `RasaModelData` inside `DIETClassifier` does not contain `entities` as a feature for training if there is no training data present for entity recognition.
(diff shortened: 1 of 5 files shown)


changelog/7720.bugfix.md
```diff
@@ -0,0 +1,3 @@
+Fix incremental training for cases when training data does not contain entities but `DIETClassifier` is configured to perform entity recognition also.
```
Ghostvv, line 1: what will happen if new data contain entities?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -648,9 +648,13 @@ def _create_model_data(
...
             attributes_to_consider.append(label_attribute)
-        if training and self.component_config[ENTITY_RECOGNITION]:
-            # we don't have any entity tags during prediction, just add them during
-            # training
+        if (
+            training
+            and self.component_config[ENTITY_RECOGNITION]
+            and self._entity_tag_specs
+        ):
+            # Consider entities as features during training and only if there was
```
Ghostvv, line 656:
```suggestion
            # Add entities as labels only during training and only if there was
```

changelog/7720.bugfix.md
dakshvar22: That's invalid because you cannot add new labels and run incremental training. We throw an appropriate error in that case.

Ghostvv (approved):

Issue #7751, opened 2021-01-19
Core re-asks for slot upon form interruption


humcasma:
**Rasa version**: 2.2.5

**Rasa SDK version** (if used & relevant): 2.2.0

**Python version**: 3.6.9

**Operating system** (windows, osx, ...): Linux-4.19.112+-x86_64-with-Ubuntu-18.04-bionic

**Issue**:
When a form is (temporarily) interrupted as part of handling an unhappy path, rasa core prints the `utter_ask_<form_name>_<slot_name>` message just before executing the actions specified to handle the deviation from the happy path.

An example can be seen executing the [financial demo](https://github.com/RasaHQ/financial-demo/tree/rasa-2-0). Using rasa version 2.0.8, the expected dialogue is generated when trying to switch forms:

```
User: I want to pay my credit
Bot: How much do you want to pay?
User: I want to transfer money
Bot: We haven't completed the credit card payment yet. Are you sure you want to switch to money transfer?
```

However, using version 2.2.5, this is the generated dialogue:

```
User: I want to pay my credit
Bot: How much do you want to pay?
User: I want to transfer money
Bot: How much do you want to pay?  <---- REPEATED!!!
Bot: We haven't completed the credit card payment yet. Are you sure you want to switch to money transfer?
```

**Error (including full traceback)**:
With Rasa version 2.0.8:
```
How much do you want to pay?
Your input ->  I want to transfer money
2021-01-19 10:46:32 DEBUG    rasa.core.processor  - Received user message 'I want to transfer money' with intent '{'id': -971893940168523848, 'name': 'transfer_money', 'confidence': 0.9983439445495605}' and entities '[]'
2021-01-19 10:46:32 DEBUG    rasa.core.processor  - Logged UserUtterance - tracker now has 18 events.
2021-01-19 10:46:32 DEBUG    rasa.core.policies.memoization  - Current tracker state [{}, {'user': {'intent': 'pay_cc'}, 'prev_action': {'action_name': 'action_listen'}}, {'user': {'intent': 'pay_cc'}, 'prev_action': {'action_name': 'cc_payment_form'}, 'active_loop': {'name': 'cc_payment_form'}}, {'user': {'intent': 'transfer_money'}, 'prev_action': {'action_name': 'action_listen'}, 'active_loop': {'name': 'cc_payment_form'}}]
2021-01-19 10:46:32 DEBUG    rasa.core.policies.memoization  - There is a memorised next action 'action_switch_forms_ask'
[...]

Issue #7710, opened 2021-01-12
Form switching fails when requested_slot is the same in both forms


desmarchris:
https://rasa-hq.slack.com/archives/C7FCS16SD/p1608240711266000

**Rasa version**: 2.1.2

**Rasa SDK version** (if used & relevant): 2.1.2

**Python version**: 3.8.6

**Operating system** (windows, osx, ...): osx

**Issue**:
When switching between two forms, if the `requested_slot` is the same in the new and old form, the bot returns an empty response rather than switching to the new form

The issue can be reproduced by pulling [this commit](https://github.com/RasaHQ/retail-demo/tree/4bbd4109a5d7d3fce09fb43a5e8b9a1cfcee3d80) from the retail-demo, and then try:

```bash
Your input ->  order status
I can help you find the status of your order. What's your email address so I can find the order?
Your input ->  return
Your input ->
```

The expected conversation must be:
```bash
Your input ->  order status
I can help you find the status of your order. What's your email address so I can find the order?
Your input ->  return
Let's start that return. To help find your order, can you please provide your email address?
Your input ->
```

TyDunn:
@ArjaanBuijk Can this be closed?

PR #7764, opened 2021-01-20
valide slots when coming back from unhappy paths

nbeuchat:
**Proposed changes**:
- Call the `validate_slots` function even if there are no slots to validate to get which slots to request next
- This fixes a bug where the next slot asked was not consistent when returning to a form from an unhappy path https://github.com/RasaHQ/rasa-sdk/issues/389

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] updated the documentation
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #7771, opened 2021-01-21
Form switching

ArjaanBuijk:
**Proposed changes**:
Fixes for two issues when switching forms:
- #7751 : Core re-asks for slot upon form interruption
- #7710 : Form switching fails when requested_slot is the same in both forms


wochinge (commented):
> Core re-asks for slot upon form interruption

In my opinion this can't happen due to https://github.com/RasaHQ/rasa/blob/cbed5741e16fa087577c16c4279bdecde7e16665/rasa/core/actions/forms.py#L492, no?

ArjaanBuijk:
@wochinge ,

> > Core re-asks for slot upon form interruption
> 
> In my opinion this can't happen due to
> 
> https://github.com/RasaHQ/rasa/blob/cbed5741e16fa087577c16c4279bdecde7e16665/rasa/core/actions/forms.py#L492
> 
> , no?

It actually does happen, because the loop is first done with the original form still active.
- The sentence that will trigger the next form is first validated for the requested slot of the orginal form
- The validation will fail, but the form is not done, and it will ask for the next slot if we don't prevent this

I restored the logic to avoid raising the ActionExecutionRejected to quickly, so other slots set by the custom validation action will be set, but add an if statement that avoids the form asking for the next slot if ActionExecutionRejected was returned.

rasa/core/actions/forms.py
```diff
@@ -428,6 +441,12 @@ def _temporary_tracker(
             slots=domain.slots,
         )
 
+    async def user_rejected_manually(self, validation_events: List[Event]) -> bool:
```
wochinge, line 444: let's make it protected as nothing outside the class calls it 
```suggestion
    async def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
```

rasa/core/actions/forms.py
```diff
@@ -428,6 +441,12 @@ def _temporary_tracker(
...
+        """Returns True if user return ActionExecutionRejected event, else False."""
```
wochinge, line 445: `if user return ActionExecutionRejected event` seems a bit hard to understand. Could please elaborate a bit in the docstring?

rasa/core/actions/forms.py
```diff
@@ -317,13 +317,26 @@ def extract_other_slots(
 
         return slot_values
 
+    def get_slot_to_fill(self, tracker: "DialogueStateTracker") -> Optional[str]:
+        """Get the slot to be filled.
```
wochinge, line 321:
```suggestion
        """Gets the name of the slot which should be filled next.
```

rasa/core/actions/forms.py
```diff
@@ -317,13 +317,26 @@ def extract_other_slots(
...
+
+        When switching to another form, the requested slot setting is still from the
+        previous form and must be ignored.
+
+        Returns:
+        The slot name or None
```
wochinge, line 327:
```suggestion
        The slot name or `None`
```

rasa/core/actions/forms.py
```diff
@@ -317,13 +317,26 @@ def extract_other_slots(
...
+        )
+
     def extract_requested_slot(
         self, tracker: "DialogueStateTracker", domain: Domain
     ) -> Dict[Text, Any]:
-        """Extract the value of requested slot from a user input
-        else return `None`.
-        """
-        slot_to_fill = tracker.get_slot(REQUESTED_SLOT)
+        """Extract the value of requested slot from a user input else return `None`."""
```
wochinge, line 338: [We use the descriptive style for docstrings](https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e)
```suggestion
        """Extracts the value of requested slot from a user input else return `None`."""
```

rasa/core/actions/forms.py
```diff
@@ -317,13 +317,26 @@ def extract_other_slots(
...
+    def get_slot_to_fill(self, tracker: "DialogueStateTracker") -> Optional[str]:
+        """Get the slot to be filled.
+
+        When switching to another form, the requested slot setting is still from the
+        previous form and must be ignored.
+
+        Returns:
+        The slot name or None
+        """
+        return (
```
wochinge, line 329: could you please add a test which tests switching forms?

rasa/core/actions/forms.py
```diff
@@ -679,9 +700,10 @@ async def do(
     ) -> List[Event]:
         events = await self._validate_if_required(tracker, domain, output_channel, nlg)
 
-        events += await self.request_next_slot(
-            tracker, domain, output_channel, nlg, events_so_far + events
-        )
+        if not await self.user_rejected_manually(events):
```
wochinge, line 703: could you please add a test for this logic?

wochinge (commented):
* Can you please add a changelog entry?
* I suggest merging to `2.2.x` as it's a bug fix
* FYI: @nbeuchat is currently fixing similar parts with https://github.com/RasaHQ/rasa/pull/7764

rasa/core/actions/forms.py
ArjaanBuijk: This logic is already tested by this test, which I had to modify after the update, because the next slot is no longer being requested when the user manually rejects:

https://github.com/RasaHQ/rasa/blob/fc674407543df7ca120d3115f5ce46f27df75ada/tests/core/actions/test_forms.py#L295-L302

rasa/core/actions/forms.py
ArjaanBuijk: Ok, let me try to figure out how to create a test that mimics the scenario of https://github.com/RasaHQ/rasa/issues/7710

wochinge:
Just re-request review once you you're done 🙌🏻

rasa/core/actions/forms.py
ArjaanBuijk: I added a test for switching forms. I commented it quite heavily. Let me know if that is OK or a bit too much comments.

rasa/core/actions/forms.py
```diff
@@ -428,6 +441,17 @@ def _temporary_tracker(
             slots=domain.slots,
         )
 
+    async def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
+        """Returns True if slot validation rejected the form execution, else False.
```
wochinge, line 445: optional: at this point it might make more sense to use a fullblown docstring which includes `Args:` and `Returns`
```suggestion
        """Returns `True` if slot validation rejected the form execution, else `False`.
```

rasa/core/actions/forms.py
```diff
@@ -428,6 +441,17 @@ def _temporary_tracker(
             slots=domain.slots,
         )
 
+    async def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
```
wochinge, line 444: there is no need for this to be `async`, no?
```suggestion
    def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
```

rasa/core/actions/forms.py
```diff
@@ -461,17 +485,19 @@ async def validate(
...
         )
-        user_rejected_manually = any(
-            isinstance(event, ActionExecutionRejected) for event in validation_events
-        )
+
         if (
             slot_to_fill
             and not some_slots_were_validated
-            and not user_rejected_manually
+            and not await self._user_rejected_manually(validation_events)
```
wochinge, line 492:
```suggestion
            and not self._user_rejected_manually(validation_events)
```

rasa/core/actions/forms.py
```diff
@@ -679,9 +705,10 @@ async def do(
     ) -> List[Event]:
         events = await self._validate_if_required(tracker, domain, output_channel, nlg)
 
-        events += await self.request_next_slot(
-            tracker, domain, output_channel, nlg, events_so_far + events
-        )
+        if not await self._user_rejected_manually(events):
```
wochinge, line 708:
```suggestion
        if not self._user_rejected_manually(events):
```

tests/core/actions/test_forms.py
```diff
@@ -90,6 +90,80 @@ async def test_activate_with_prefilled_slot():
     ]
 
 
+async def test_switch_forms_with_same_slot():
+    """Tests switching of forms, where the first slot is the same in both forms.
+
+    Tests the fix for issue 7710"""
```
wochinge, line 96:
```suggestion
    """Tests switching of forms, where the first slot is the same in both forms."""
```

tests/core/actions/test_forms.py
```diff
@@ -90,6 +90,80 @@ async def test_activate_with_prefilled_slot():
...
+
+    # Define two forms in the domain, with same first slot
+    slot_a = "my_slot_a"
+
+    form_1 = "my_form_1"
+    text_1a = f"Please provide the value for {slot_a} of form 1"
```
wochinge, line 102: what do you think of naming the `text_xa` items to this?
```suggestion
    utter_ask_form_1 = f"Please provide the value for {slot_a} of form 1"
```

tests/core/actions/test_forms.py
```diff
@@ -90,6 +90,80 @@ async def test_activate_with_prefilled_slot():
...
+      entity: number
+responses:
+    utter_ask_{form_1}_{slot_a}:
+    - text: {text_1a}
+    utter_ask_{form_2}_{slot_a}:
+    - text: {text_2a}
+"""
+    domain = Domain.from_yaml(domain)
+
+    # Driving it like rasa/core/processor._run_py
```
wochinge, line 125: it seems to be safe to create a `Processor` object with your domain and then manually call `_run_action` on it. what do you think?

tests/core/actions/test_forms.py
```diff
@@ -90,6 +90,80 @@ async def test_activate_with_prefilled_slot():
...
+        domain,
+    )
+    tracker.update_with_events(events, domain)  # _log_action_on_tracker in processor.py
+    # verify that `form_1` asks for `slot_a` using the correct response
+    assert events[:-1] == [ActiveLoop(form_1), SlotSet(REQUESTED_SLOT, slot_a)]
+    assert isinstance(events[-1], BotUttered)
+    assert events[-1].text == text_1a
+
+    # Next, bot predicts action_listen
+    tracker.update_with_events([ActionExecuted(action_name="action_listen")], domain)
```
wochinge, line 144: let's use the existing constant
```suggestion
    tracker.update_with_events([ActionExecuted(action_name=ACTION_LISTEN_NAME)], domain)
```

tests/core/actions/test_forms.py
```diff
@@ -90,6 +90,80 @@ async def test_activate_with_prefilled_slot():
...
+
+    # User utters something, with the intent to switch to form_2
```
wochinge, line 146: let's also add this as event to make the test more realistic

tests/core/actions/test_forms.py
```diff
@@ -90,6 +90,80 @@ async def test_activate_with_prefilled_slot():
...
+    events = await action_2.run(
+        CollectingOutputChannel(),
+        TemplatedNaturalLanguageGenerator(domain.templates),
+        tracker,
+        domain,
+    )
+    # verify that `form_2` asks for `slot_a` using the correct response
+    assert events[:-1] == [ActiveLoop(form_2), SlotSet(REQUESTED_SLOT, slot_a)]
+    assert isinstance(events[-1], BotUttered)
+    assert events[-1].text == text_2a
```
wochinge, line 164: let's also assert the currently active form

rasa/core/actions/forms.py
ArjaanBuijk: Created a fullblow docstring which includes Args & Returns.

tests/core/actions/test_forms.py
ArjaanBuijk: I like that idea.  Let me try that.

tests/core/actions/test_forms.py
ArjaanBuijk: Aren't we doing that by asserting `ActiveLoop(form_2)`, or are you thinking of a different parameter?

tests/core/actions/test_forms.py
wochinge: @ArjaanBuijk have you seen this one?

tests/core/actions/test_forms.py
```diff
@@ -90,6 +95,150 @@ async def test_activate_with_prefilled_slot():
...
+            ActionExecuted(ACTION_LISTEN_NAME),
+            UserUttered("order status", {"name": "form_1", "confidence": 1.0}),
+            DefinePrevUserUtteredFeaturization(False),
+        ],
+    )
+    # rasa/core/processor.predict_next_action
+    probabilities = []
+    policy_name = "policy_2_RulePolicy"
+    policy_priority = (6,)
+    policy_events = []
```
wochinge, line 163: in my opinion fine to set these directly in `PolicyPrediction`

tests/core/actions/test_forms.py
```diff
@@ -90,6 +95,150 @@ async def test_activate_with_prefilled_slot():
...
+    prediction = PolicyPrediction(
+        probabilities,
+        policy_name,
+        policy_priority,
+        policy_events,
+        is_end_to_end_prediction=False,
+        is_no_user_prediction=False,
+        diagnostic_data={},
```
wochinge, line 171: no need to set all these default values again

tests/core/actions/test_forms.py
```diff
@@ -90,6 +95,150 @@ async def test_activate_with_prefilled_slot():
...
+    assert events[:-1] == [
+        ActionExecuted(ACTION_LISTEN_NAME),
+        UserUttered("order status", {"name": "form_1", "confidence": 1.0}),
+        DefinePrevUserUtteredFeaturization(False),
+        ActionExecuted(form_1),
+        ActiveLoop(form_1),
+        SlotSet(REQUESTED_SLOT, slot_a),
+    ]
+    assert isinstance(events[-1], BotUttered)
+    assert events[-1].text == utter_ask_form_1
```
wochinge, line 195: How about checking this with the previous `assert`?

tests/core/actions/test_forms.py
```diff
@@ -90,6 +95,150 @@ async def test_activate_with_prefilled_slot():
...
+    # tracker.update(ActionExecutionRejected(action_name=form_1))  # processor._run_action
+    await processor._run_action(
+        action_1,
+        tracker,
+        CollectingOutputChannel(),
+        TemplatedNaturalLanguageGenerator(domain.templates),
+        prediction,
+    )
+    events = tracker.applied_events()
+    assert events[-1] == ActionExecutionRejected(action_name=form_1)
```
wochinge, line 220: How about checking this with the previous assert?

tests/core/actions/test_forms.py
```diff
@@ -90,6 +95,150 @@ async def test_activate_with_prefilled_slot():
...
+    )
+    events = tracker.applied_events()
+    # verify that `form_2` asks for `slot_a` using the correct response
+    assert events[-4:-1] == [
+        ActionExecuted(form_2),
+        ActiveLoop(form_2),
+        SlotSet(REQUESTED_SLOT, slot_a),
+    ]
+    assert isinstance(events[-1], BotUttered)
+    assert events[-1].text == utter_ask_form_2
```
wochinge, line 239: How about checking this with the previous assert?

ArjaanBuijk:
@wochinge ,
I like the way the test for form switching is looking after implementing your suggestions. Let me know what you think.

tests/core/actions/test_forms.py
wochinge: @ArjaanBuijk ?

wochinge (approved):
Looks good 🚀 Nice work! Do you want to target 2.2.x so you can release a micro with the fix? Alternatively we're also releasing a minor on Thursday.

ArjaanBuijk:
> Looks good  Nice work! Do you want to target 2.2.x so you can release a micro with the fix? Alternatively we're also releasing a minor on Thursday.

@wochinge ,
I prefer to release in the minor on Thursday.

wochinge (approved):
🚀

wochinge:
Nice work @ArjaanBuijk ! 💯

Issue #7579, opened 2020-12-17
implement BILOU tags in core featurizers


Ghostvv:
with e2e TED can extract entities using data in stories.
BILOU tags help extract multi-token entities, it is implemented in DIET, port BILOU tags to TED

PR #7785, opened 2021-01-22
add bilou tagging to e2e ted entities

Ghostvv:
**Proposed changes**:
- closes https://github.com/RasaHQ/rasa/issues/7579
- add bilou tagging to e2e ted entities


rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -5,7 +5,9 @@
 from collections import defaultdict
 
 import rasa.shared.utils.io
-from rasa.nlu.constants import TOKENS_NAMES
+from rasa.nlu.classifiers.diet_classifier import EntityTagSpec
```
joejuzl, line 8: should this be moved to a more neutral location?

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -49,32 +50,61 @@ def __init__(self) -> None:
...
+
+        # NO_ENTITY_TAG corresponds to non-entity which should correspond to 0 index
+        # needed for correct prediction for padding
+        tag_id_index_mapping[NO_ENTITY_TAG] = 0
+
+        # TODO
+        #  The entity states used to create the tag-idx-mapping contains the
+        #  entities and the concatenated entity and roles/groups. We do not
+        #  distinguish between entities and roles/groups right now.
+        #  we return a list to anticipate that
```
joejuzl, line 86: would this be a tuple rather than a list?

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -275,40 +309,36 @@ def encode_entities(
...
         #  entities and the concatenated entity and roles/groups. We do not
         #  distinguish between entities and roles/groups right now.
-        # TODO
-        #  Should we support BILOU tagging?
-
-        if TEXT not in entity_data or len(self.entity_tag_id_mapping) < 2:
+        if (
+            not entity_data
+            or not self.entity_tag_specs
+            or self.entity_tag_specs[0].num_tags < 2
```
joejuzl, line 324: Why not just have different instance variables for entities and roles/groups?

tests/core/featurizers/test_single_state_featurizers.py
```diff
@@ -225,6 +225,70 @@ def test_single_state_featurizer_with_entity_roles_and_groups(
     )
 
 
+@pytest.mark.timeout(300)  # these can take a longer time than the default timeout
+def test_single_state_featurizer_with_bilou_entity_roles_and_groups(
+    unpacked_trained_moodbot_path: Text,
+):
+    from rasa.core.agent import Agent
+
+    interpreter = Agent.load(unpacked_trained_moodbot_path).interpreter
```
joejuzl, line 234: do we really need a fully trained model here?

joejuzl (changes requested):

rasa/core/featurizers/single_state_featurizer.py
Ghostvv: good point, I thought about it, but then thought meh 😅

rasa/core/featurizers/single_state_featurizer.py
Ghostvv: because they don't always all appear and not always together

tests/core/featurizers/test_single_state_featurizers.py
Ghostvv: I believe there is no other way currently to get featurizers?

rasa/core/featurizers/single_state_featurizer.py
Ghostvv: in line 87 I return a list, do you think I should return a tuple? I think list is better

rasa/core/featurizers/single_state_featurizer.py
Ghostvv: moved to `extractor.py`

tests/core/featurizers/test_single_state_featurizers.py
Ghostvv: at least other tests get the full model. I just copy pasted 🙈

rasa/core/featurizers/single_state_featurizer.py
joejuzl: If it's always 2 things, then it should be a tuple right? A list is for a collection of varying length.

tests/core/featurizers/test_single_state_featurizers.py
joejuzl: It's not ideal because it means we have to train moodbot just to run this test, and I'm assuming the functionality tested here doesn't require a trained model right?

rasa/core/featurizers/single_state_featurizer.py
Ghostvv: it's not always 2 things. it can be 1 or 2 or 3 things, and 2nd thing is not always the same.
We have entities, roles and groups, and there might be situations:
- nothing
- only entities
- entities and roles
- entities and groups
- entities and roles and groups

tests/core/featurizers/test_single_state_featurizers.py
Ghostvv: it does. Remember we discussed the need for separate featurizers? This is the reason, in order to extract entities we need trained tokenizer, and I don't think there is a way extract only tokenizer. And if there is, I'd say it is out of scope of current PR

joejuzl: ok makes sense

rasa/core/featurizers/single_state_featurizer.py
joejuzl: ok, so three nullable things then? A dictionary would work? Having a mapping from type of thing to index it occurs if not ideal.

Ghostvv: maybe. but it is the way it is done in nlu currently. I'd suggest that it should be done in separate PR, since it'll require refactoring a lot of nlu code

joejuzl (approved):

PR #7786, opened 2021-01-22
Added (try..except) block for is_domain_file

degiz:
Fixes: https://github.com/RasaHQ/rasa-x/issues/4311

**Proposed changes**:
- Added a `try..catch` block in `is_domain_file()`


federicotdn (approved):
Thanks, this will fix our problem. I'm thinking that in the not so distant future we'll want to do something similar for more YAML-loading code paths, as I honestly think that this environment-variable-expanding thing should not prevent Rasa from loading a perfectly valid YAML file.

PR #7795, opened 2021-01-25
Loading the data in core test with TrainingDataImporter interface

Imod7:
**Proposed changes**:
Using the TrainingDataImporter to load the data in core test.

_[The current PR is a continuation of this [PR](https://github.com/RasaHQ/rasa/pull/7758) where I did something wrong with the commits/rebase/merge and it is quite difficult to clean it up so I created a new PR, the current one, with all changes in a new/clean branch]._

**Note**
It is the proposed solution for the 1st part of this Issue [#5986](https://github.com/RasaHQ/rasa/issues/5986).
The 2nd part was solved from this [PR](https://github.com/RasaHQ/rasa/pull/7674).


rasa/shared/data.py
```diff
@@ -147,7 +147,7 @@ def get_data_files(
 
         if is_valid_filetype(path):
             if filter_predicate(path):
-                data_files.add(os.path.abspath(path))
+                data_files.add(path)
```
wochinge, line 150: [Rasa X ](https://github.com/RasaHQ/rasa-x/blob/b03fc7359e29a3c088a92b3dc91a8449c9216bdc/src/community/rasax/community/initialise.py#L357) also this so I'm a bit hesitant to change this as it might break things on Rasa X side. This leaves us with two options:
* changing the test: One could argue that the absolute path is bad UX 🤔 However, I think that's pretty minor
* changing [this](https://github.com/RasaHQ/rasa/blob/954e4734d479924df7211cd3bc5503285d54a66b/rasa/shared/core/trackers.py#L667) to convert the path to a relative path. 

What do you think? Both options are valid for me.

wochinge (commented):
Thanks for creating the new PR!

We also need a changelog entry for this.

rasa/shared/data.py
Imod7: That is so true (about the use from Rasa X). I missed it this time but next time I will be more careful on this. 
So, I changed the `test` (by changing to an absolute path) since I find this the safest solution.
I also did a search on what is the best practice (absolute vs relative path) but there are different opinions based on the use so I can't really express an opinion on that.

Imod7: Also, I added a changelog with type `improvement`.

wochinge: > I missed it this time but next time I will be more careful on this.

No worries. These things are very hard to catch and that's what a reviewer is there for. 

> I find this the safest solution.

Why do you consider this option safer?

changelog/5986.improvement.md
```diff
@@ -0,0 +1 @@
+Using the TrainingDataImporter interface to load the data in core test
```
wochinge, line 1: I'd also mention the path in the failed stories file which is absolute now!
```suggestion
Using the `TrainingDataImporter` interface to load the data in `rasa test core` .
```

wochinge (commented):
seems it's time to convert from `draft` to `ready for review` 💯

rasa/shared/data.py
Imod7: I think that a change in the test has less **direct** impact on the functionality of the code/product. Changing code affects directly how the code works and it can change/break things like you mentioned. However the test can also have impact on the functionality (maybe more **indirect** or **long term**?) in case the change makes us miss things that we should catch or vice versa. 🧐 Let's say that I am more of a fan of first changing/doing more tests in order to analyse the behaviour of the code under different cases/perspectives, understand better the problem it is trying to solve until I am convinced that the design of the code needs improvement or fixing. That way, taking into account different cases and having better understanding, I can also come up with a better code solution/improvement/fix proposal.

changelog/5986.improvement.md
Imod7: Yes I can do that! 👍🏾 To be honest, it is not quite clear for me what to include in the changelog and what not. Based on the very helpful also advice from @alwx  [here](https://github.com/RasaHQ/rasa-x/pull/4239#discussion_r560809197), I included only the change related to the issue itself. 

@wochinge @alwx Do you think having a guideline on how to write a good changelog will be helpful? There are different articles out there explaining that but maybe having one that outlines what Rasa wants to track and easily find ?

wochinge: >  Based on the README, the changelog is for the users

This is absolutely correct 👍🏻 

> the change in the test should not be mentioned right?

You had to change the test because the behavior of the production code changed. Instead of referring to failed test stories via relative path, Rasa Open Source will now refer to them via absolute path. This is a change which will affect the users and hence it should be covered by a changelog. 

> But I thought that in general a changelog is for the developers no?

**Developers** who are **using** our framework. They are using our framework to train and create contextual AI assistants. They are interested what changes in the usage. Imagine you're a **developer** **using** the `numpy` library. Would you care if they add an additional type annotation? I guess, no. Would you care if they rename `numpy.array` to `numpy.as_array`? I hope yes 😆 

> Otherwise how do we keep track of changes like this one ?

If it's impacting the users we should have a changelog. Otherwise we still have Git (blame) + the commit messages and the related PR to track any changes.

rasa/shared/data.py
wochinge: > I think that a change in the test has less direct impact on the functionality of the code/product.

The question is **why** you had to change the test. In your case you had to change the test because you changed some production code. I've a feeling that absolute paths are in general more robust than relatives which could be seen as argument that this way is slightly "safer"

changelog/5986.improvement.md
Imod7: Just to note that this Issue [#5986](https://github.com/RasaHQ/rasa/issues/5986) has 2 parts/2 related PRs, one is the current one and the other is this [PR](https://github.com/RasaHQ/rasa/pull/7674). 
The PR (_Implementing the use of `TrainingDataImporter` in `rasa test`_) did not have a changelog but I guess it is not necessary since it did not affect/changed anything on the usage. Just mentioning it here.

changelog/5986.improvement.md
```diff
@@ -0,0 +1,3 @@
+Using the `TrainingDataImporter` interface to load the data in `rasa test core`.
+
+Referring to failed test stories via absolute path (change in test `test_source_in_failed_stories`).
```
wochinge, line 3: There is no need to point to the actual code change you did. If users want to see that than they can find the PR using the issue number of the changelog. Rasa developers can also find the PR by its commit hash.

```suggestion
Failed test stories are now referenced by their absolute path instead of the relative path.
```

wochinge (approved):
Nice work wrapping this up 💯

Issue #7680, opened 2021-01-05
Add integration tests for Postgres, Rabbit, Redis


wochinge:
## Proposal

Rasa Open Source has multiple integrations with external services, such as databases or event brokers. While it's very hard to test against all of them (e.g. integration tests for all input channels), we should add least have integration tests for the most important ones (e.g. the ones which are used by Rasa X), namely:
* `SQLTrackerStore` with Postgres
* `RedisLockStore`
* `PikaEventBroker` 

### Workflow
- In the `continous-integration.yml` workflow, add a new job called `integration-tests`
- This job will [setup 3 services](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices) using Docker images: `postgres` ([more info](https://docs.github.com/en/actions/guides/creating-postgresql-service-containers)), `redis` ([more info](https://docs.github.com/en/actions/guides/creating-redis-service-containers)) and `rabbitmq` ([more info](https://github.community/t/test-the-rabbitmq-service-with-pika-library-in-github-actions/116599/5))
- Then, this job will run all the integration tests via pytest (we'll need to ignore these tests from the unit and functional tests)
- This job will be required to pass via branch protection, and will prevent releases in case of failure

### Folder structure
- under the `tests/` folder, create an `integration_tests/` folder, where all integration tests will live
- I'd follow the same folder structure than other tests (e.g. replicate the tree structure of the codebase), and create a `conftest.py` file that will contain constants and fixtures for every services

### Seeding the tests
We'll need to write at least one test for each service. Here are simple use cases that we can start with:
- Postgres: `SQLTrackerStore` initialisation with a `login_db` (success + failure paths)
- Redis: `RedisLockStore`, actually we can move [all the tests](https://github.com/RasaHQ/rasa/blob/master/tests/core/test_lock_store.py)  here and make sure they pass without the mocked library
- RabbitMQ: `PikaEventBroker` initialisation + connection

⬆️ if you have any idea on how we can reduce the scope of these, please LMK

kaustuk:
@wochinge is this open for community? If yes can you give me some pointer on what exactly is being expected and what needs to be done. I would like to contribute if possible.

wochinge:
Hey @kaustuk , thanks for the initiative! I'll pull in @tczekajlo and @mprazz from our infrastructure team who'd usually implement / coordinate this within our team.

m-vdb:
[...]

PR #7800, opened 2021-01-25
Integration tests workflow for Postgres, Redis, RabbitMQ

m-vdb:
# Proposal
Fixes https://github.com/RasaHQ/rasa/issues/7680

## Workflow
- In the `continous-integration.yml` workflow, add a new job called `integration-tests`
- This job will [setup 3 services](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices) using Docker images: `postgres` ([more info](https://docs.github.com/en/actions/guides/creating-postgresql-service-containers)), `redis` ([more info](https://docs.github.com/en/actions/guides/creating-redis-service-containers)) and `rabbitmq` ([more info](https://github.community/t/test-the-rabbitmq-service-with-pika-library-in-github-actions/116599/5))
- Then, this job will run all the integration tests via pytest (we'll need to ignore these tests from the unit and functional tests)
- This job will be required to pass via branch protection, and will prevent releases in case of failure

## Folder structure
- under the `tests/` folder, create an `integration_tests/` folder, where all integration tests will live
- I'd follow the same folder structure than other tests (e.g. replicate the tree structure of the codebase), and create a `conftest.py` file that will contain constants and fixtures for every services

## Seeding the tests
We'll need to write at least one test for each service. Here are simple use cases that we can start with:
- Postgres: `SQLTrackerStore` initialisation with a `login_db` (success + failure paths)
- Redis: `RedisLockStore`, actually we can move [all the tests](https://github.com/RasaHQ/rasa/blob/master/tests/core/test_lock_store.py)  here and make sure they pass without the mocked library
- RabbitMQ: `PikaEventBroker` initialisation + connection

## Setup to run tests locally
Nice to have: docker compose setup

# Status (please check what you already did):


tests/integration_tests/core/test_tracker_store.py
```diff
@@ -0,0 +1,124 @@
...
+            database_name=POSTGRES_TRACKER_STORE_DB,
+        )
+        .rowcount
+    )
+    assert matching_rows == 1
+    tracker_store.engine.dispose()
+
+
+@pytest.mark.timeout(10)
+def test_sql_tracker_store_with_login_db_race_condition(
```
m-vdb, line 80: These 3 `SQLTrackerStore` tests cannot be run in parallel because they rely on the state of the database. That's why they're failing in the CI. Possible solutions:
- disable parallel runs for now, and activate when we want too
- use pytest markers to identify tests that should not be ran in parallel, and run integration tests with 2 commands.

thoughts?

mprazz: 2nd option is preferrable from my PoV, as we'll have a lower impact on overall test runtime + it's something we'll probably need to do eventually anyway.

m-vdb: Thanks Mariana! Now you know what I was talking about 😉 

There is also a third option: rewrite the tests so that they can run in parallel. It means that the names of the database vary from a test to another. I'm not a fan of this solution because I have a feeling it could get quite messy

mprazz: I'd say we will always have some test cases that will necessitate some kind of sequential execution 🤔  (I'm thinking of the onboarding tests that I'm currently working on as an example. Since I need to set a specific onboarding state and do consequential actions in the Rasa X instance, running these tests in parallel will be near impossible and likely very messy 😬 )

tczekajlo:
@m-vdb Just wanted to ask if you need help here with anything?

m-vdb:
all good @tczekajlo . It's done on my end, except [the outstanding discussion](https://github.com/RasaHQ/rasa/pull/7800#discussion_r565498115) where I need input from @RasaHQ/enable-squad engineers. But I assume that they'll do that when they have time, e.g. next week!

tests/integration_tests/core/test_lock_store.py
```diff
@@ -0,0 +1,194 @@
...
+            message.sender_id,
+            wait_time_in_seconds=wait_time_in_seconds,
+            lock_lifetime=lock_lifetime,
+        ):
+            # hold up the message processing after the lock has been acquired
+            await asyncio.sleep(holdup)
+
+        return None
+
+    with patch.object(Agent, "handle_message", mocked_handle_message):
```
wochinge, line 183: optional: let's use monkeypatch as in the other tests for consistency

tests/integration_tests/core/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+@pytest.fixture
+async def default_agent(_trained_default_agent: Agent) -> Agent:
+    agent = _trained_default_agent
+    agent.tracker_store = InMemoryTrackerStore(agent.domain)
+    agent.domain.session_config = SessionConfig.default()
+    return agent
+
+
+@pytest.fixture
+def postgres_login_db_connection() -> Iterator[sa.engine.Connection]:
```
joejuzl, line 58: This is great. We should do something like this for mysql in the unit test to force cleanup.
One potential (optional) improvement: We should create the engine and connection in a session wide fixture, and then have a function level fixture which creates a transaction , yields, then rolls the transaction.

tests/integration_tests/core/test_tracker_store.py
joejuzl: If each test runs in a separate transaction (like in my suggestion above) you should be able to run them in parallel.

tests/integration_tests/core/conftest.py
m-vdb: So yes, ultimately, this is something we should do. But in this case, it doesn't help me:
> CREATE DATABASE cannot run inside a transaction block

We cannot leverage transactions for that use case 😫

tests/integration_tests/core/test_tracker_store.py
m-vdb: this is reasonable but [doesn't work in this case](https://github.com/RasaHQ/rasa/pull/7800#discussion_r573085200):
> CREATE DATABASE cannot run inside a transaction block

m-vdb: I gave a go at the second option with 598c16ffff9

tests/integration_tests/core/test_lock_store.py
m-vdb: fixed in d910d53f948; I'm somewhat of a `unittest.patch` guy

.github/workflows/continous-integration.yml
```diff
@@ -30,6 +30,7 @@ env:
   BOTO_CONFIG: /dev/null
   IS_TAG_BUILD: ${{ startsWith(github.event.ref, 'refs/tags') }}
   DOCKERHUB_USERNAME: tmbo
+  DEFAULT_PYTHON_VERSION: '3.7'
```
wochinge, line 33: 🚀

.github/workflows/continous-integration.yml
```diff
@@ -285,6 +286,136 @@ jobs:
...
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+      RABBITMQ_HOST: localhost
+      RABBITMQ_PORT: 5672
+      RABBITMQ_USER: guest
+      RABBITMQ_PASSWORD: guest
+
+    services:
+      redis:
+        image: redis
```
wochinge, line 307: wondering if we should pin the version as using `latest` is against best practices. What do you think?

.github/workflows/continous-integration.yml
```diff
@@ -285,6 +286,136 @@ jobs:
...
+        # Set health checks to wait until redis has started
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          # FIXME: cannot use ${{ env.REDIS_PORT }} here
```
wochinge, line 315: why not?

wochinge: seems weird that it [works for credentials](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservicesservice_idcredentials) but not for ports 🤔

.github/workflows/continous-integration.yml
```diff
@@ -285,6 +286,136 @@ jobs:
...
+      with:
+        poetry-version: ${{ env.POETRY_VERSION }}
+
+    - name: Load Poetry Cached Libraries ⬇
+      id: cache-poetry
+      if: needs.changes.outputs.backend == 'true'
+      uses: actions/cache@v2
+      with:
+        path: .venv
+        key: ${{ runner.os }}-poetry-${{ env.POETRY_VERSION }}-${{ env.DEFAULT_PYTHON_VERSION }}-${{ hashFiles('**/poetry.lock') }}-venv-${{ secrets.POETRY_CACHE_VERSION }}
```
wochinge, line 370: why are we not using the same cache as for the backend unit tests?

.github/workflows/continous-integration.yml
```diff
@@ -285,6 +286,136 @@ jobs:
...
+    - name: Set up virtual environment
+      if: needs.changes.outputs.backend == 'true'
+      # Poetry on Windows cannot pick up the virtual environments directory properly,
+      # and it creates a new one every time the pipeline runs.
+      # This step solves this problem — it tells poetry to always use `.venv` directory inside
+      # the project itself, which also makes it easier for us to determine the correct directory
+      # that needs to be cached.
+      run: poetry config virtualenvs.in-project true
+
+
```
wochinge, line 392:
```suggestion
```

.github/workflows/continous-integration.yml
```diff
@@ -285,6 +286,136 @@ jobs:
         COVERALLS_SERVICE_NAME: github
       run: poetry run coveralls
 
+  integration_test:
```
wochinge, line 289: potentially out of the scope of this PR, but: shouldn't we run the integration tests against different Python versions / Windows as well?

.github/workflows/continous-integration.yml
```diff
@@ -285,6 +286,136 @@ jobs:
...
+      if: needs.changes.outputs.backend == 'true'
+      env:
+        JOBS: 2
+        PYTEST_MARKERS: '"not sequential"'
+        PYTHONIOENCODING: "utf-8"
+      run: |
+        make test-integration
+
+    - name: Test Code with Services 🩺 (sequential)
+      if: needs.changes.outputs.backend == 'true'
```
wochinge, line 411: How about adding a comment which explains `sequential` vs `not sequential`?

README.md
```diff
@@ -185,6 +185,29 @@ JOBS=[n] make test
 
 Where `[n]` is the number of jobs desired. If omitted, `[n]` will be automatically chosen by pytest.
 
+
+### Running the Integration Tests
+
+In order to run the integration tests, make sure that you have the development requirements installed:
```
wochinge, line 191: should we add a link where to get docker-compose / docker?

changelog/7680.improvement.md
```diff
@@ -0,0 +1,4 @@
+We're now running integration tests for Rasa, with initial coverage for `SQLTrackerStore` (with PostgreSQL),
```
wochinge, line 1: + should in my opinion be a `misc` changelog as it's not user facing
```suggestion
We're now running integration tests for Rasa Open Source, with initial coverage for `SQLTrackerStore` (with PostgreSQL),
```

changelog/7680.improvement.md
```diff
@@ -0,0 +1,4 @@
...
+`RedisLockStore` (with Redis) and `PikaEventBroker` (with RabbitMQ). The integration tests are now part of our
+CI, and can also be ran locally using `make test-integration`
+(see [Rasa Open Source README](https://github.com/RasaHQ/rasa) for more information).
```
wochinge, line 4: optional: link to the specific section

deployment/docker-compose.integration.yml
```diff
@@ -0,0 +1,38 @@
+version: "3.4"
```
wochinge, line 1: Should we use a newer version? 3.4 seems quite old

pytest.ini
```diff
@@ -1,5 +1,6 @@
 [pytest]
 markers =
     skip_on_windows: mark a test as a test that shouldn't be executed on Windows.
+    sequential: mark tests than cannot be ran in parallel mode.
```
wochinge, line 4: what do you think? 
```suggestion
    sequential: mark tests than cannot be ran in parallel (e.g. because they need access to the same resource).
```

rasa/core/tracker_store.py
```diff
@@ -939,13 +939,14 @@ def _create_database_and_update_engine(self, db: Text, engine_url: "URL"):
             return
 
         self._create_database(self.engine, db)
+        self.engine.dispose()
```
wochinge, line 942: 🚀

tests/integration_tests/core/conftest.py
wochinge: Usually the docker image already creates a default database . Can't we use that one as `login_db` and then use Joe's s suggestion?

tests/integration_tests/core/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+        _create_login_db(conn)
+        yield conn
+    finally:
+        _drop_db(conn, POSTGRES_LOGIN_DB)
+        _drop_db(conn, POSTGRES_TRACKER_STORE_DB)
+        conn.close()
+        engine.dispose()
+
+
+def _create_login_db(connection: sa.engine.Connection):
```
wochinge, line 81:
```suggestion
def _create_login_db(connection: sa.engine.Connection) -> None:
```

tests/integration_tests/core/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+    connection.execution_options(isolation_level="AUTOCOMMIT").execute(
+        f"CREATE DATABASE {POSTGRES_LOGIN_DB}"
+    )
+
+
+def _drop_db(connection: sa.engine.Connection, database_name: Text):
```
wochinge, line 87:
```suggestion
def _drop_db(connection: sa.engine.Connection, database_name: Text) -> None:
```

tests/integration_tests/core/conftest.py
wochinge, line 58: should the postgres fixtures go in the `test_tracker_store` module?

tests/integration_tests/core/test_lock_store.py
```diff
@@ -0,0 +1,194 @@
...
+    ticket = lock.issue_ticket(0.00001)
+    time.sleep(0.00002)
+    assert ticket == 1
+    assert lock._ticket_for_ticket_number(ticket) is None
+
+    # newly assigned ticket should get number 1 again
+    assert lock.issue_ticket(10) == 1
+
+
+async def test_multiple_conversation_ids(default_agent: Agent):
```
wochinge, line 81: I don't understand this test. Seems to me it tests if `async.gather` works as expected?

tests/integration_tests/core/test_tracker_store.py
```diff
@@ -0,0 +1,139 @@
...
+from .conftest import (
+    POSTGRES_HOST,
+    POSTGRES_PORT,
+    POSTGRES_LOGIN_DB,
+    POSTGRES_TRACKER_STORE_DB,
+    POSTGRES_USER,
+    POSTGRES_PASSWORD,
+)
+
+# NOTE about the timeouts in this file. We want to fail fast
```
wochinge, line 19: 🙌🏻

tests/integration_tests/core/test_tracker_store.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.mark.sequential
+@pytest.mark.timeout(10)
+def test_sql_tracker_store_with_login_db_race_condition(
+    postgres_login_db_connection: sa.engine.Connection,
+    caplog: LogCaptureFixture,
+    monkeypatch: MonkeyPatch,
+):
+    original_execute = sa.engine.Connection.execute
+
+    def mock_execute(self, *args, **kwargs):
```
wochinge, line 97: please add types

wochinge (commented):
Very neat! This gives us a great base to get started 💯 

Todo list (not for this PR but in general)
- [ ] write a `EventBroker` test which tests publising a message
- [ ] write a `EventBroker` test which uses multiple queues
- [ ] write a `EventBroker` test which simulates a connection error
- [ ] write a `TrackerStore` tests which tests retrieving only events from the last session
- [ ] write a `TrackerStore` test which tests all possible events

.github/workflows/continous-integration.yml
m-vdb: yes, really not sure why not 🤷🏻

.github/workflows/continous-integration.yml
m-vdb: 🤔 they're the same? am I missing something?

.github/workflows/continous-integration.yml
m-vdb: Definitely out of scope 👍🏻 but to add to a TODO / future discussions

deployment/docker-compose.integration.yml
m-vdb: I've copied it from Rasa X, but sure

.github/workflows/continous-integration.yml
m-vdb: yes that's better. Wondering if we should pin the exact version or only the major. I'd go with the major (Redis seems to update major tags with the latest micro). WDYT?

wochinge: Sounds good 👍🏻 We are not using any crazy feature so functionality shouldn't break too quickly anyway

tests/integration_tests/core/test_tracker_store.py
m-vdb: to be fair it's a tad hard to find out the return type of `sa.engine.Connection.execute`. Suggestions? Otherwise I'll go with `Any`

tests/integration_tests/core/conftest.py
m-vdb: the problem doesn't come from the `login_db`, but from the other db that is created upon login. We want to test situations where there db exists or doesn't exist yet. So I don't think it fixes the problem 🤔

.github/workflows/continous-integration.yml
m-vdb: I can't pin rabbitmq's version because I'm using a version with custom healthchecks and this repo doesn't push tags 🤦🏻 . I think the right thing to do here is to fork the repo and introduce versions (or ask them to push versioned tags instead of just `latest`). But I don't think I can do this in a reasonable timeframe 😖

tests/integration_tests/core/conftest.py
m-vdb: 🤔 not sure at this stage, I'd rather keep things in conftest.py for now, as maybe some other tests could use it?

tests/integration_tests/core/test_lock_store.py
m-vdb: I think that the agent's `handle_message` leverages the lock store, so this test ensure it works correctly. I copy pasted this from the unit tests, guessing it could be useful to have an integration test for it. WDYT?

wochinge:
just another todo (not for this PR but for the future): Integration tests with the `TrackerStore` and Oracle

deployment/docker-compose.integration.yml
```diff
@@ -0,0 +1,38 @@
...
+    image: postgres:13
+    # Set health checks to wait until postgres has started
+    healthcheck:
+      test: ["CMD", "pg_isready"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    environment:
+      # postgres image requires password to be set
+      PGUSER: ${PGUSER}
```
joejuzl, line 26: where is this var set?

Makefile
```diff
@@ -1,6 +1,8 @@
 .PHONY: clean test lint init docs build-docker build-docker-full build-docker-mitie-en build-docker-spacy-en build-docker-spacy-de
 
 JOBS ?= 1
+INTEGRATION_TEST_FOLDER = tests/integration_tests/
+PYTEST_MARKERS ?= "sequential or not sequential"
```
joejuzl, line 5: `PYTEST_MARKERS` feels a bit generic to me, maybe include integration in the name?

tests/integration_tests/core/brokers/test_pika.py
```diff
@@ -0,0 +1,23 @@
...
+    broker = PikaEventBroker(
+        host=RABBITMQ_HOST,
+        username=RABBITMQ_USER,
+        password=RABBITMQ_PASSWORD,
+        port=RABBITMQ_PORT,
+        queues=[RABBITMQ_DEFAULT_QUEUE],
+    )
+    try:
+        await broker.connect()
+        assert broker._exchange is not None
```
joejuzl, line 21: could check `.is_ready()` here

tests/integration_tests/core/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+DEFAULT_STORIES_FILE = "data/test_yaml_stories/stories_defaultdomain.yml"
+
+
+@pytest.fixture
+def redis_lock_store() -> RedisLockStore:
+    return RedisLockStore(REDIS_HOST, REDIS_PORT)
+
+
+@pytest.fixture(scope="session")
+async def _trained_default_agent(tmpdir_factory: TempdirFactory) -> Agent:
```
joejuzl, line 35: Any reason we are duplicating this?

tests/integration_tests/core/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+POSTGRES_DEFAULT_DB = os.getenv("POSTGRES_DEFAULT_DB", "postgres")
+POSTGRES_TRACKER_STORE_DB = "tracker_store_db"
+POSTGRES_LOGIN_DB = "login_db"
+
+DEFAULT_STORIES_FILE = "data/test_yaml_stories/stories_defaultdomain.yml"
+
+
+@pytest.fixture
+def redis_lock_store() -> RedisLockStore:
+    return RedisLockStore(REDIS_HOST, REDIS_PORT)
```
joejuzl, line 31: Is it worth making a redis fixture that cleans up data?

tests/integration_tests/core/test_lock_store.py
joejuzl: I wonder if there is a way to have these tests in a test class and then just run them in different contexts e.g. inmemory for the unit tests, but with redis here?
Maybe a future improvement...

tests/integration_tests/core/test_tracker_store.py
```diff
@@ -0,0 +1,139 @@
...
+        password=POSTGRES_PASSWORD,
+        db=POSTGRES_TRACKER_STORE_DB,
+        login_db=POSTGRES_LOGIN_DB,
+    )
+
+    matching_rows = (
+        postgres_login_db_connection.execution_options(isolation_level="AUTOCOMMIT")
+        .execute(
+            sa.text(
+                "SELECT 1 FROM pg_catalog.pg_database WHERE datname = :database_name"
```
joejuzl, line 44: can we get this info from the sqlalchemy session?

joejuzl (changes requested):
Do we envision the `deployment/` directory being used for anything other than integration tests? If so I would consider renaming/moving it.

tests/integration_tests/core/test_lock_store.py
m-vdb: 🤔 yes could be a good idea at some point, seems a bit premature to me though.

deployment/docker-compose.integration.yml
m-vdb: good catch, forgot to add it to `.env.example`

tests/integration_tests/core/conftest.py
m-vdb: nope 😬 good one

tests/integration_tests/core/test_tracker_store.py
m-vdb: I'm not sure what you exactly mean here @joejuzl, can you detail a bit more? Are you talking about using `SQLTrackerStore.session_scope()`?

m-vdb:
> Do we envision the deployment/ directory being used for anything other than integration tests? If so I would consider renaming/moving it.

@joejuzl I duplicated what we have on Rasa X side. Not sure how we want to scaffold things ultimately. Do you have a suggestion for the name? `integration`? `services`?

tests/integration_tests/core/conftest.py
m-vdb: yes! and it's dead simple with Redis

tests/integration_tests/core/test_tracker_store.py
joejuzl: I just wondered if it was possible to see what databases existed through an sqlalchemy method rather than having to query sql directly.

joejuzl:
> @joejuzl I duplicated what we have on Rasa X side. Not sure how we want to scaffold things ultimately. Do you have a suggestion for the name? integration? services?

I was thinking that it may be confusing to users if it's called `deployment` but is only for integration tests. Maybe we could have it in the integration tests folder? Or call it something `integration_test_deployment`? (or maybe a bit of a mouth-full 😛 )

m-vdb:
@joejuzl ok 👍🏻 how about `tests_deployment`?

tests/integration_tests/core/test_tracker_store.py
m-vdb: 🤔 I didn't find a way to do that with sqlalchemy

tests/integration_tests/core/test_tracker_store.py
joejuzl: Could `print(type(whatever))` and run it. However I'm not super bothered about types for test code like this tbh.

joejuzl (approved):
Good stuff! 🚀  
I'm looking forward to catching our first bug with them :)

PR #7817, opened 2021-01-26
use simple random

Ghostvv:
**Proposed changes**:
- correct negative sampling requires while_loop implementation with random shuffle inside, however it creates a memory leak. Substitute it with simple random uniform int.


Ghostvv:
/modeltest

```
include:
- dataset: ["all"]
  config: ["all"]
```

rasa/utils/tensorflow/layers.py
```diff
@@ -601,45 +601,9 @@ def _make_flat(x: tf.Tensor) -> tf.Tensor:
...
-            tf.while_loop(
-                cond,
-                body,
-                loop_vars=[idx1, out1],
-                shape_invariants=[idx1.shape, tf.TensorShape([None, self.num_neg])],
-                parallel_iterations=self.parallel_iterations,
-            ),
-        )[1]
+        return tf.random.uniform(
+            shape=(batch_size, self.num_neg), maxval=total_candidates, dtype=tf.int32
```
dakshvar22, line 605: Just to be sure, `total_candidates` is 1-indexed and not 0-indexed, right? Asking bc according to the documentation, range of `tf.random.uniform` is defined by `[minval, maxval)`.

Ghostvv: total_candidates should be 0-indexed, it should be the length of the batch or number of labels

dakshvar22: If it is 0-indexed then we should do `maxval=total_candidates+1`.

dakshvar22: But I don't think that's the case because we were using `tf.range(total_candidates)` earlier which would have had the same effect.

Ghostvv: but to double check, I printed the max value to be sure

dakshvar22:
@Ghostvv Can you add a changelog as well for record purposes?

Ghostvv:
@dakshvar22 I believe regression tests indicate that we're good to go

joejuzl (approved):
No leak and more succinct! Very nice 👍

PR #7819, opened 2021-01-26
Create a Docker base image with poetry

tczekajlo:
**Proposed changes**:
- Create a separate Docker base image with poetry


docker/Dockerfile.full
```diff
@@ -2,11 +2,17 @@
...
 ARG BASE_IMAGE_HASH
 ARG BASE_MITIE_IMAGE_HASH
+ARG POETRY_VERSION
 
 FROM ${IMAGE_BASE_NAME}:base-mitie-${BASE_MITIE_IMAGE_HASH} as mitie
 
 FROM ${IMAGE_BASE_NAME}:base-${BASE_IMAGE_HASH} as builder
 
+# install poetry
+ENV POETRY_VERSION ${POETRY_VERSION}
```
wochinge, line 12: Do you think having a separate `base-builder` image which extends the `base` image with the `poetry` installation is an overkill? Having that would reduce the duplicate steps for installing poetry.

wochinge (commented):
💯

docker/Dockerfile.full
tczekajlo: I like the idea!

wochinge (approved):
💯

Issue #7069, opened 2020-10-20
UnsupportedModelError could be made clearer.


koaning:
I was using Rasa 1.10 to load in a Rasa 2.0 model. This was accidental on my part, but the error message seems to assume that you're loading in an "older" model. It is also throw when you're loading in a newer model.

Here's the error I got. 

```
UnsupportedModelError: The model version is too old to be loaded by this Rasa NLU instance. Either retrain the model, or run with an older version. Model version: 1.10.3 Instance version: 2.0.0
```

Would anybody oppose if we change it to; 

```
UnsupportedModelError: The model was trained using Rasa 1.10.3 and does not match your Rasa installation (2.0.0). This means that you probably want to retrain your model to ensure that the versions match up again. 
```

I'll gladly pick it up when I get a green light.

TyDunn:
@koaning @alwx and I discussed that this makes sense, so I think you can go ahead 👍

PR #7838, opened 2021-01-28
Update UnsupportedModelError

koaning:
As small PR that addresses https://github.com/RasaHQ/rasa/issues/7069. The old error message assumes that you're working on a newer Rasa version while this technically isn't always the case. Hence a small change to the error message.


rasa/nlu/model.py
```diff
@@ -289,11 +289,10 @@ def ensure_model_compatibility(
         model_version = metadata.get("rasa_version", "0.0.0")
         if version.parse(model_version) < version.parse(version_to_check):
             raise UnsupportedModelError(
-                "The model version is too old to be "
-                "loaded by this Rasa NLU instance. "
-                "Either retrain the model, or run with "
-                "an older version. "
-                "Model version: {} Instance version: {}"
+                "The model version is trained using Rasa {}"
```
wochinge, line 292: optional: Let's use f-strings for the win 🥳

wochinge, line 292: also note the addtional space at the end
```suggestion
                "The model version is trained using Rasa Open Source {} "
```

rasa/nlu/model.py
```diff
@@ -289,11 +289,10 @@ def ensure_model_compatibility(
...
+                "and is not compatible with your current installation ({}). "
+                "This means that you probably want to retrain your model "
```
wochinge, line 294: optional: No strong opinion on this, but I don't think we should give instructions using words like `probably`. What do you think? 
```suggestion
                "This means that you either need to retrain your model "
```

wochinge (commented):
💯 Thanks for this! I've seen multiple GitHub issues related to this so I think it will be very helpful. How about targeting 2.2.x as it's a bug fix / usage improvement?

koaning:
@wochinge I can directly point to `2.2.x` now?

wochinge:
>  I can directly point to 2.2.x now?

Yes please, but you'd need to do a rebase (otherwise you'll have all the `main` commits in 2.2.x)

rasa/nlu/model.py
```diff
@@ -289,12 +289,11 @@ def ensure_model_compatibility(
...
-                "loaded by this Rasa NLU instance. "
-                "Either retrain the model, or run with "
-                "an older version. "
-                "Model version: {} Instance version: {}"
-                "".format(model_version, rasa.__version__)
+                f"The model version is trained using Rasa Open Source {model_version} "
+                f"and is not compatible with your current installation ({rasa.__version__}). "
+                "This means that you either need to retrain your model "
+                "or revert back to the Rasa version that trained the model "
+                "to ensure that the versions match up again."
```
wochinge, line 296: I think this is somewhere in our code engineering guidelines. If one part of a string uses an f-string, the other parts should also use it for consistency.
```suggestion
                f"The model version is trained using Rasa Open Source {model_version} "
                f"and is not compatible with your current installation ({rasa.__version__}). "
                f"This means that you either need to retrain your model "
                f"or revert back to the Rasa version that trained the model "
                f"to ensure that the versions match up again."
```

wochinge (commented):
Code looks great. Not approving yet because of the rebase.

koaning:
I am suddenly confronted with the fact that I've never rebased before in my life. 😅 

I'm more comfortable making a new PR that's clean for now, while I read on rebase in the meantime.

koaning:
New PR https://github.com/RasaHQ/rasa/pull/7877

Issue #7250, opened 2020-11-12
Add __repr__ to Token object


koaning:
When I'm debugging what is happening in a pipeline I usually have a look and print the `Message`.

It might look something like;

<pre>> message.as_dict_nlu()
 {'entities': [],
 'intent': 

 {'confidence': 0.8422146439552307, 'id': 5464168520135895892, 'name': 'are_you_real'} 

,
 'intent_ranking': [

 {'confidence': 0.8422146439552307, 'id': 5464168520135895892, 'name': 'are_you_real'} 

,

 {'confidence': 0.04479081928730011, 'id': -5724794541682903534, 'name': 'how_made'} 

,

 {'confidence': 0.021328529343008995, 'id': -6505008832162321521, 'name': 'how_are_you'} 

,

 {'confidence': 0.019935669377446175, 'id': 522343063256866795, 'name': 'what_can_you_do'} 

,

 {'confidence': 0.013976013287901878, 'id': 7022544970669495622, 'name': 'languages'} 

,

 {'confidence': 0.013974795117974281, 'id': -279316034893973609, 'name': 'weather'} 

,

 {'confidence': 0.013861254788935184, 'id': -3284643768121576930, 'name': 'what_time'} 

,

 {'confidence': 0.013021928258240223, 'id': -938916847167484283, 'name': 'who_made_you'} 

,

 {'confidence': 0.011181396432220936, 'id': 2425092467132022535, 'name': 'birthday'} 

,

 {'confidence': 0.005714914295822382, 'id': 4373745406320874474, 'name': 'restaurant'} 

],
 'intent_tokens': [<rasa.nlu.tokenizers.tokenizer.token object at 0x7fdc606328d0>
 ],
 'text': 'are you a real person',
 'text_tokens': [
 <rasa.nlu.tokenizers.tokenizer.token object at 0x7fdc880b1350>
 ,

 <rasa.nlu.tokenizers.tokenizer.token object at 0x7fdc6057ff90>
 ,

 <rasa.nlu.tokenizers.tokenizer.token object at 0x7fdc6057fed0>
 ,

 <rasa.nlu.tokenizers.tokenizer.token object at 0x7fdc6057ff50>
 ,

 <rasa.nlu.tokenizers.tokenizer.token object at 0x7fdc6057ff10>
 ]}
 </rasa.nlu.tokenizers.tokenizer.token>
 </rasa.nlu.tokenizers.tokenizer.token>
[...]

PR #7839, opened 2021-01-28
Add __repr__ to Token

koaning:
Adds a `__repr__` to token. Part of issue https://github.com/RasaHQ/rasa/issues/7250. 

Haven't 100% decided what an ideal `__repr__` of the other message features might be. So I'll do that later in another PR.

### DIFF
--- a/rasa/nlu/tokenizers/tokenizer.py
+++ b/rasa/nlu/tokenizers/tokenizer.py
@@ -60,6 +60,9 @@ class Token:
             other.lemma,
         )
 
+    def __repr__(self):
+        return f"<Token object value='{self.text}' start={self.start} end={self.end} at {hex(id(self))}>"
+
 
 class Tokenizer(Component):
     def __init__(self, component_config: Dict[Text, Any] = None) -> None:


koaning:
It feels like a grand formality, do we really want to add docstrings to dunder-methods like `__repr__`?

dakshvar22:
@koaning The change looks good to me. Regarding your question on docstring, I'd ask @joejuzl or @wochinge on it. Any opinions folks?

joejuzl:
I don't think all dunder methods (especially  `__repr__`) should require docstrings. However I think this would require a change to the docstring CI check.

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -60,6 +60,9 @@ def __lt__(self, other):
             other.lemma,
         )
 
+    def __repr__(self):
```
wochinge, line 63:
```suggestion
    def __repr__(self) -> Text:
```

wochinge:
> I don't think all dunder methods (especially __repr__) should require docstrings. However I think this would require a change to the docstring CI check.

I was also annoyed by this. Will create a survey.

dakshvar22 (approved):
Good to merge once the docstring requirement is relaxed.

m-vdb:
@koaning is this PR still on your radar?

koaning:
It kind of is, but it's very much in the "nice to have"-category as opposed to the "must have"-category. 

Was the docstring style thing resolved already? I assumed since there's no ping on the topic it's still ongoing.

PR #7844, opened 2021-01-29
Fix v1 docs noindex header

maggo:
Proposed changes:

Fix _redirects so that netlify picks up the noindex, nofollow rule

### DIFF
--- a/docs/_redirects
+++ b/docs/_redirects
@@ -1,4 +1,4 @@
 # target all pages
 /*
   # set noindex, no follow
-  X-Robots-Tag: noindex, no follow
+  X-Robots-Tag: noindex, nofollow


sara-tagger:
Thanks for submitting a pull request 🚀 @m-vdb will take a look at it as soon as possible ✨

m-vdb (approved):
🚀

lunelson (commented):
@maggo should this file not have been named _headers rather than _redirects? Also, I think `none` is a more sure value than `noindex, nofollow` and maybe remove the comment...

maggo:
@lunelson oh you're absolutely right. Was so used to the unified config file that I didn't look at the filename. 🤦  Also yeah, we can switch to `X-Robots-Tag: none`

Will open another PR

Issue #7741, opened 2021-01-15
Enforce annotating unit tests that train a model


joejuzl:
**Description of Problem**:
Following on from #7740 we would like to enforce that in the future tests that train a model are annotated using a pytest mark.

**Overview of the Solution**:
Determine if there is a way to enforce annotating unit tests that train. This might be a mix of looking at usage of fixtures and calling of certain methods.

**Definition of Done**:
Either:
- Implement the solution if manageable. Or
-  A new issue containing the proposed approach. Or
- A reason why it is not possible.

joejuzl:
One approach to this would be to call:
```
RAISE_ON_TRAIN=True pytest -m "not trains_model"
```

And then decorate the train methods with a method that raises if `RAISE_ON_TRAIN` is set.

PR #7860, opened 2021-02-01
Mark tests that train a model.

joejuzl:
**Proposed changes**:
This PR marks all tests that call either `nlu.train`, `core.train`, `Agent.train`, or `Trainer.train`. This can be directly, or indirectly e.g. through a fixture or http request.


tests/core/test_agent.py
```diff
@@ -74,6 +74,7 @@ async def test_training_data_is_reproducible():
         assert str(x.as_dialogue()) == str(same_training_data[i].as_dialogue())
 
 
+@pytest.mark.trains_model
```
wochinge, line 77: what about `test_load_agent`?

tests/core/test_agent.py
```diff
@@ -385,6 +390,7 @@ async def test_agent_update_model_none_domain(trained_rasa_model: Text):
     assert tracker.events[3].intent["name"] == "greet"
 
 
+@pytest.mark.trains_model
```
wochinge, line 393: why is this one marked?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -195,6 +195,7 @@ async def _train_persist_load_with_different_settings(
 
 
 @pytest.mark.skip_on_windows
+@pytest.mark.trains_model
```
wochinge, line 198: `test_process_gives_diagnostic_data` also trains, no?

tests/nlu/selectors/test_selectors.py
```diff
@@ -274,6 +275,7 @@ async def _train_persist_load_with_different_settings(
 
 
 @pytest.mark.skip_on_windows
+@pytest.mark.trains_model
```
wochinge, line 278: what about `_train_persist_load_with_different_settings`?

tests/nlu/test_components.py
```diff
@@ -190,6 +191,7 @@ def test_can_handle_language_guard_clause(
     assert expected_exec_msg in str(excinfo.value)
 
 
+@pytest.mark.trains_model
```
wochinge, line 194: mhm, I don't think it makes sense to lable this one.

tests/nlu/test_components.py
```diff
@@ -204,6 +206,7 @@ async def test_validate_requirements_raises_exception_on_component_without_name(
         )
 
 
+@pytest.mark.trains_model
```
wochinge, line 209: same as the comment above.

tests/nlu/test_evaluation.py
```diff
@@ -369,6 +370,7 @@ async def test_run_evaluation(unpacked_trained_moodbot_path: Text):
     assert result.get("intent_evaluation")
 
 
+@pytest.mark.trains_model
```
wochinge, line 373: `test_run_cv_evaluation` and `test_run_cv_evaluation_with_response_selector` also train (even multiple times)

tests/nlu/test_train.py
```diff
@@ -155,6 +156,7 @@ async def test_train_persist_load_parse(language, pipeline, component_builder, t
 @pytest.mark.timeout(600)
 @pytest.mark.parametrize("language, pipeline", pipelines_for_non_windows_tests())
 @pytest.mark.skip_on_windows
+@pytest.mark.trains_model
```
wochinge, line 159: haha, this is such an expensive test 🤯

wochinge: We should also annoate the `Trainer.train` tests. They are such easy wins!

wochinge (commented):
* don't you need to add the marker [here](https://github.com/RasaHQ/rasa/blob/f376df30c807eacb0ccd272885ed419a51118928/pytest.ini#L2)?
* I think we should also handle `Agent.train` , what do you think?
  - https://github.com/RasaHQ/rasa/blob/644ec371c0d6b7d3d3eb3fcf064156e7a2e03dfa/tests/conftest.py#L83
  - https://github.com/RasaHQ/rasa/blob/c62680968d253556c8b7b7fdfb0fda94ab26d362/tests/core/test_evaluation.py#L149
  
 * What about `Trainer.train`? 
   - https://github.com/RasaHQ/rasa/blob/c62680968d253556c8b7b7fdfb0fda94ab26d362/tests/nlu/selectors/test_selectors.py#L64
  - https://github.com/RasaHQ/rasa/blob/a4375463f12d402ff64c1bf66899d738e4905cf4/tests/nlu/test_train.py#L167 (very expensive)
  
 * I think these were not caught as they are done as part of different process
   - https://github.com/RasaHQ/rasa/blob/e69600e81f449e9796ea7bc00b4757c564023466/tests/cli/conftest.py#L55
   - https://github.com/RasaHQ/rasa/blob/3ce26e53241693568b16e9a2af9f78c90c28375e/tests/cli/test_rasa_train.py#L31 (all this module)
   

Wow 💯 What a useful exercise. Especially the NLU pipeline trainings are such easy wins 💯  Do we merge this to `main` or do we use that as target branch for our improvements?

tests/core/test_agent.py
joejuzl: Are you suggesting renaming the marker or the test?

tests/core/test_agent.py
joejuzl: oops was meant to be on `test_load_agent`

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: yeah, not sure why that one didn't get recorded...

tests/nlu/selectors/test_selectors.py
joejuzl: it's not a test, but I've marked the tests that call it.

tests/nlu/test_components.py
joejuzl: Our marking strategy has to be consistent if we want to automate it. And we agreed on any call to `nlu.train` or `code.train`.

tests/nlu/test_evaluation.py
joejuzl: I guess not through `nlu.train`?

joejuzl:
> don't you need to add the marker here?

Ah yeah, I just forgot to commit it. It's been a hectic week.

> I think these were not caught as they are done as part of different process

Hmm it shouldn't matter as I recorded them by writing to file. Maybe it doesn't make it to the train method method?

w.r.t to also catching `Agent.train` and `Trainer.train`, I think it makes sense as it means a whole ensemble or pipeline will be trained. It's pretty time consuming to catch them all though, so maybe as a followup?

> Do we merge this to main or do we use that as target branch for our improvements?

Good question, maybe integration branch works until we have the automatic check in the CI? Also, we probably need to document it somewhere.

tests/core/test_agent.py
wochinge: adding a marker

wochinge: haha, seems resolved 👍🏻

tests/nlu/test_components.py
wochinge: Ok 👍🏻

tests/nlu/test_evaluation.py
wochinge: No 😬  It trains through the `Trainer`.

wochinge:
> Hmm it shouldn't matter as I recorded them by writing to file. Maybe it doesn't make it to the train method method?

mhm, maybe this process doesn't know it's run by `pytest`? The should train through the CLI train methods and hence trigger your checks 🤔 

>  I think it makes sense as it means a whole ensemble or pipeline will be trained. It's pretty time consuming to catch them all though, so maybe as a followup?

Mhm, I think we should at least do the ones which commented on in `tests/nlu/test_train` as they'd be a huge win in terms of performance.

joejuzl:
> Mhm, I think we should at least do the ones which commented on in tests/nlu/test_train as they'd be a huge win in terms of performance.

@wochinge I've added all the tests that call Agent.train, or Trainer.train.

wochinge:
Great work, another review?

> Good question, maybe integration branch works until we have the automatic check in the CI? Also, we probably need to document it somewhere.

Sounds good 👍🏻

joejuzl:
> mhm, maybe this process doesn't know it's run by pytest? The should train through the CLI train methods and hence trigger your checks 🤔

It was because I was using a relative path. Marked them now.

tests/nlu/selectors/test_selectors.py
```diff
@@ -48,6 +48,7 @@
         ],
     ],
 )
+@pytest.mark.trains_model
```
wochinge, line 51: I guess `test_process_gives_diagnostic_data` should also be annotated

wochinge (approved):
💯

tests/nlu/selectors/test_selectors.py
joejuzl: yeah, these are new tests that keep coming 🤦 
We can get all the stragglers once #7741 is done and merged into this branch.

wochinge: Sounds good! Already forcing people to use `epoch: 1` whereever possible :-D

tczekajlo (approved):
LGMT 🎖️

PR #7861, opened 2021-02-01
add missing 'confused with' field to model report

wochinge:
**Proposed changes**:
- include 'confused' with field when doing model testing
- add more logs when doing model testing via server

### DIFF
--- a/rasa/nlu/test.py
+++ b/rasa/nlu/test.py
@@ -665,7 +665,7 @@ def _calculate_report(
             labels,
             exclude_labels=[exclude_label] if exclude_label else [],
         )
-    else:
+    elif not output_directory:
         log_evaluation_table(report, precision, f1, accuracy)
 
     return report, precision, f1, accuracy, confusion_matrix, labels
--- a/rasa/nlu/test.py
+++ b/rasa/nlu/test.py
@@ -654,13 +654,11 @@ def _calculate_report(
     if report_as_dict is None:
         report_as_dict = bool(output_directory)
 
-    if output_directory:
-        report, precision, f1, accuracy = get_evaluation_metrics(
-            targets,
-            predictions,
-            output_dict=report_as_dict,
-            exclude_label=exclude_label,
-        )
+    report, precision, f1, accuracy = get_evaluation_metrics(
+        targets, predictions, output_dict=report_as_dict, exclude_label=exclude_label,
+    )
+
+    if report_as_dict:
         report = _add_confused_labels_to_report(
             report,
             confusion_matrix,
@@ -668,14 +666,7 @@ def _calculate_report(
             exclude_labels=[exclude_label] if exclude_label else [],
         )
     else:
-        report, precision, f1, accuracy = get_evaluation_metrics(
-            targets,
-            predictions,
-            output_dict=report_as_dict,
-            exclude_label=exclude_label,
-        )
-        if isinstance(report, str):
-            log_evaluation_table(report, precision, f1, accuracy)
+        log_evaluation_table(report, precision, f1, accuracy)
 
     return report, precision, f1, accuracy, confusion_matrix, labels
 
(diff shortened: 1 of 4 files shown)


rasa/nlu/test.py
```diff
@@ -654,28 +654,19 @@ def _calculate_report(
...
-            targets,
-            predictions,
-            output_dict=report_as_dict,
-            exclude_label=exclude_label,
-        )
+    report, precision, f1, accuracy = get_evaluation_metrics(
+        targets, predictions, output_dict=report_as_dict, exclude_label=exclude_label,
+    )
+
+    if report_as_dict:
```
joejuzl, line 661: It's not clear to me why we only add confused labels if this is true.
In fact when is this actually `False`?

rasa/nlu/test.py
```diff
@@ -654,28 +654,19 @@ def _calculate_report(
...
     else:
-        report, precision, f1, accuracy = get_evaluation_metrics(
-            targets,
-            predictions,
-            output_dict=report_as_dict,
-            exclude_label=exclude_label,
-        )
-        if isinstance(report, str):
-            log_evaluation_table(report, precision, f1, accuracy)
+        log_evaluation_table(report, precision, f1, accuracy)
```
joejuzl, line 669: Why only log in this case?

rasa/nlu/test.py
wochinge: This is `false` if no output directory was provided **and** it's used from the CLI. If this is `False` then the report will be formatted string (done by sklearn) which would lead to an error when trying to add the `confused_with` labels

rasa/nlu/test.py
wochinge: Whenever we store the report to disk it will be a formatted string report and `report_as_dict` will be `True` (from the CLI). Only if `report_as_dict` is `False` (either specified manually or not specified + no output directory provided), we get the formatted string report which needs to be logged. We could change this to `elif not output_directory` 🤔

rasa/nlu/test.py
joejuzl: Is that even possible? It seems to default to a path in `run_nlu_test_async`

wochinge: True! Could still be used from the Python API though.

rasa/nlu/test.py
wochinge: Changed this to the `elif` part btw.

rasa/nlu/test.py
joejuzl: Having a boolean arg which is passed so far through the code like this, changed by the value of other args, and determining the type of the value being passed back and how it's manipulated it super error prone. (Especially considering it's probably not really needed!)

wochinge: The only other option I see I is to always `report_as_dict` and purge the logging to stdout completely. In my opinion this is nothing we can do in a micro release. Also wondering if the error isn't default in the CLI and `output_directory` should be `None` by default.

joejuzl: Ideally we would have always have a dictionary representation of the report at this level, and would then have different ways of outputting at the interface levels e.g. server, cli, python api. 
But agreed it's too much for now.

wochinge: True. I guess it's a bit difficult as we are currently using `sklearn` to create the formatted report instead of deciding that on our own.

wochinge:
@joejuzl Can I merge this then?

rasa/nlu/test.py
joejuzl: We could still always convert it to a dict when we first generate it.

joejuzl (approved):

rasa/nlu/test.py
wochinge: True. Should I create an issue?

joejuzl: Yeah would be great 👍

Issue #7658, opened 2020-12-31
Rasa 2.2.2 core model predictions off


iszainab:
**Rasa version**: 2.2.2

**Rasa SDK version** (if used & relevant): 2.2.0

**Rasa X version** (if used & relevant): None

**Python version**: 3.6.9

**Operating system** (windows, osx, ...): Ubuntu 18.04

**Issue**:
1. Rasa core model predictions seem off. Everything was fine with RASA 1.10 however I migrated to RASA 2.2.2 following the docs and everything and I made sure not to change any hyper parameters when switching from depreciated policies. I tried debugging and my stories seem to match the tracker states and yet the predicted action is off.

2. Why does it seem like the nlu components are being processed during the core model's evaluation? Is this expected behavior? I can see in the docs that it says that the state featurizer uses the nlu pipeline to featurize states but it doesn't seem right that my custom components that have nothing to do with featurization are being processed. Can there be a better solution for this please?

NOTE: For all practical aspects, my training parameters have not changed at all.

**Content of old configuration file (config.yml)** (if relevant):
```yml
# Configuration for Rasa NLU.
# https://rasa.com/docs/rasa/nlu/components/
language: en

pipeline:
  - name: "mycustomcomponent1.CustomComponent1"
  - name: "mycustomcomponent2.CustomComponent2"
  - name: "SpacyNLP"
  - name: "SpacyTokenizer"
  - name: "SpacyFeaturizer"
  - name: "RegexFeaturizer"
  - name: "LexicalSyntacticFeaturizer"
  - name: "CountVectorsFeaturizer"
    analyzer: "char_wb"
    min_ngram: 1
    max_ngram: 4
  - name: "CRFEntityExtractor"
    BILOU_flag: True
  - name: "EntitySynonymMapper"
  - name: "DIETClassifier"
    intent_classification: True
    entity_recognition: False
    use_masked_language_model: False
    batch_strategy: 'sequence'
    epochs: 300
  - name: "ResponseSelector"
    retrieval_intent: chitchat
    epochs: 100

policies:
  - name: FormPolicy
  - name: FallbackPolicy
    nlu_threshold: 0.3
    ambiguity_threshold: 0.1
    core_threshold: 0.05
    fallback_action_name: 'entity_driven_fallback'
  - name: MappingPolicy
  - name: AugmentedMemoizationPolicy
    featurizer:
      - name: MaxHistoryTrackerFeaturizer
        max_history: 11
        state_featurizer:
[...]

Issue #7221, opened 2020-11-10
Rasa 2.0.x fails to use stories where the previous events in the current conversation are different to the previous events in a story when making predictions


Devocado:
**Rasa version**: 2.0.3

**Python version**: 3.8.2

**Operating system** Ubuntu 20.04

**Issue**: 
It appears that Rasa 2.0.x fails to use stories where the previous events in the current conversation are different to the previous events in the story when making predictions where Rasa 1.10.x does not

have create a demo repository to reproduce the issue here: https://github.com/Devocado/reproduce-stories-bug-rasa2 which demonstrates two instances of this.
- In the first case, circling back to a certain point in a story (demonstrated in the first test) ends in Rasa 2.0.x falling back, while Rasa 1.10.x correctly follows the same path as before.
- In the second case, where a story does not overlap with other stories Rasa 2.0.x will not predict the correct actions when the intent which begins the story is triggered. Again, the second test fails in Rasa 2.0.x while passing in rasa 1.10.x

It also gives the steps to reproduce this behaviour, which for completeness I'll add it here too:

**Steps to reproduce:**
- Train the assistant in Rasa 1.10.14.
- Run the tests

All tests will pass.

- Train the assistant in Rasa 2.0.x
- Run the tests

The first two tests will fail

sara-tagger:
Thanks for raising this issue, @tmbo will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

wochinge:
@Ghostvv Do you have an idea?

Ghostvv:
The only difference to 1.10 that come to mind is that we changed the padding logic from padding from the beginning to padding from the end. The fact that TED cannot learn to predict `create_shipment_form` is not surprising. The surprising part is that 1.10 ted can learn that

Ghostvv:
Hi, I have a theory why performance might drop. Could you please try `mcfly-ted` brach of `rasa` with your data?

PR #7867, opened 2021-02-02
invert dialogue time in ted

Ghostvv:
**Proposed changes**:
- invert dialogue sequence before passing it to the transformer so that the last dialogue input becomes the first and therefore always have the same pos encoding
- fixes: https://github.com/RasaHQ/rasa/issues/7658
- fixes: https://github.com/RasaHQ/rasa/issues/7221
- fixes: https://forum.rasa.com/t/ted-fails-overfitting/39474

Some results on Sara:
main -> conv acc `0.533`, action acc `0.841`
mcfly-ted -> conv acc `0.564`, action acc `0.873`


rasa/core/policies/ted_policy.py
```diff
@@ -191,7 +191,7 @@ class TEDPolicy(Policy):
         # Max position for relative embeddings
         MAX_RELATIVE_POSITION: None,
         # Use a unidirectional or bidirectional encoder.
-        UNIDIRECTIONAL_ENCODER: True,
+        UNIDIRECTIONAL_ENCODER: False,
```
dakshvar22, line 194: Will need a change in the docs too. Can you also mention the change of default value in the changelog?

rasa/core/policies/ted_policy.py
```diff
@@ -925,6 +925,11 @@ def _prepare_layers(self) -> None:
             self.config[TRANSFORMER_SIZE][DIALOGUE],
             self.config[DROP_RATE_DIALOGUE],
             self.config[DROP_RATE_ATTENTION],
+            # use bidirectional transformer, because
+            # we will invert dialogue sequence so that the last turn is located
+            # at the first position and would always have
+            # exactly the same positional encoding
+            unidirectional=not self.use_only_last_dialogue_turns,
```
dakshvar22, line 932: I'd say lets use a variable explicitly to denote when the dialogue transformer is supposed to be bi-directional. Right now, on first look it seems a bit blurry on what `config[UNIDIRECTIONAL]`, `self.use_only_last_dialogue_turns` are intended to do. Specially, now that we are using `config[UNIDIRECTIONAL]` for feature sequence layer transformers as well.

dakshvar22 (changes requested):
Small refactor comments.

rasa/core/policies/ted_policy.py
Ghostvv: basically `config[UNIDIRECTIONAL]` is a parameter from `DIET` and is only for text transformers

Ghostvv: > I'd say lets use a variable explicitly

could you please elaborate, what do you mean?

dakshvar22: How about we create an instance variable in `TED` named `dialogue_bidirectional`(or something similar) which is set in the constructor itself based on `self.use_only_last_dialogue_turns`. Then here and everywhere else where we make an operation because of transformer being bidirectional and input sequence being reversed, we explicitly use that variable instead of `use_only_last_dialogue_turns`. Right now, the usage of `use_only_last_dialogue_turns` is overloaded across a couple of different contexts.

rasa/core/policies/ted_policy.py
Ghostvv: this parameter was never in the docs, because the fact that it was configurable is rather a bug

Ghostvv: created bug fix changelog

rasa/core/policies/ted_policy.py
Ghostvv: > first look it seems a bit blurry on what config[UNIDIRECTIONAL]
I updated the comment in `defaults`

Ghostvv: they have to be linked they are equivalent to `isinstance(self.featurizer, MaxHistoryTrackerFeaturizer)` if one is `True` the other one should be `False` the other combination is not allowed. I can rename `use_only_last_dialogue_turns` into `max_history_featurizer_is_used`

dakshvar22: That would be better I guess, since both are derived from the same information,  it will be more explicit on how they are computed.

Ghostvv: renamed

dakshvar22 (approved):

Issue #7724, opened 2021-01-14
support spaCy 2.3.0+


Aspirinkb:
**Description of Problem**:

The latest rasa(2.2.5) needs spacy < 2.3.0, but there only Chinese models after 2.3.0 in spacy.
**Overview of the Solution**:

Upgrading rasa to support spacy 2.3.0+, so that one can use spacy official Chinese language model.
**Examples** (if relevant):

spaCy official Chinese language model: https://spacy.io/models/zh
**Blockers** (if relevant):

Chinese language.
**Definition of Done**:

sara-tagger:
Thanks for submitting this feature request 🚀 @tttthomasssss will get back to you about it soon! ✨

koaning:
I'm picking this up as part of our upgrade to spaCy 3.0. Note that this should make all available Chinese models available in Rasa too!

Aspirinkb:
Nice to see the upgrade~

Ghostvv:
@koaning should this issue be closed in favor of https://github.com/RasaHQ/rasa/issues/6906?

koaning:
I was thinking of closing this issue once the PR was merged but sure. The ball is already rolling so I'm fine with removing this one.

Issue #6906, opened 2020-10-05
spaCy 3.0 support, is Rasa 2.0 ready? 


koaning:
spaCy 3.0 will be out, probably this month. It makes sense to check compatibility. The release notes on [pypi](https://pypi.org/project/spacy/#history) suggests there's already dev versions.

koaning:
So it seems like spaCy will have transformer support for a lot of Non-English languages. The thing though is that the vectors are retrieved differently. 

```
token_emb, sent_emb = nlp("i duck to hide")._.trf_data.tensors
token_emb.shape, sent_emb.shape
```

For transformer based models you can't fetch the tensors directly via `vector` anymore but via `._.trf_data`. 

Also note that the `token_emb` do not match the shape of the tokens that spaCy finds. More likely these tokens correspond tot he bytepair-something tokens that the internal transformer is using.

Issue #5400, opened 2020-03-10
suggestion for Chinese general entity extractor.


BobCN2017:
**Description of Problem**:
There is currently no general Chinese entity extraction component in the Rasa components. Recently I found a good solution.

**Overview of the Solution**:
Lexical Analysis of Chinese (LAC) is from Baidu. LAC provides a trained model, which can be easily called through paddlehub, which can extract the four main entities of PER-person name, LOC-place name, TIME-time, ORG-organization name, and can also extract 24 types of part-of-speech tags. See details https://github.com/PaddlePaddle/models/tree/develop/PaddleNLP/lexical_analysis#%E4%BB%BB%E5%8A%A1%E5%AE%9A%E4%B9%89%E4%B8%8E%E5%BB%BA%E6%A8%A1

LAC precision and recall about extracting is nearly 90%, It is good enough, I think this will like DucklingHTTPExtractor for chinese.

Based on the above situation, this model can be easily integrated into rasa and used as a general entity extraction component. Refer to spacy_entity_extractor, I written in the chinese_normal_entity-extractor, the detailed code is as follows:

import logging
import time
from typing import Any, Dict, List, Text

from rasa.nlu.extractors import EntityExtractor
from rasa.nlu.training_data import Message

logger = logging.getLogger(__name__)

class ChineseNormalEntityExtractor(EntityExtractor):
    provides = ["entities"]

    defaults = {
        # by default PER LOC TIME ORG dimensions recognized by LAC are returned
        # dimensions can be configured to contain an array of strings
        # with the names of the dimensions to filter for
        "dimensions": ["PER", "LOC", "TIME", "ORG"],
        "rename": {},
    }

    def __init__(self, component_config: Text = None) -> None:
        super(ChineseNormalEntityExtractor, self).__init__(component_config)
        import paddlehub as hub
        self.module = hub.Module(name='lac')

    def process(self, message: Message, **kwargs: Any) -> None:
        start = time.time()
        all_extracted = self.add_extractor_name(self.extract_entities(message.text))
        dimensions = self.component_config["dimensions"]
        extracted = ChineseNormalEntityExtractor.filter_irrelevant_entities(
            all_extracted, dimensions
        )
        message.set(
            "entities", message.get("entities", []) + extracted, add_to_output=True
        )
        logger.debug("LAC cost time:{}".format(time.time() - start))

    def extract_entities(self, text: "Text") -> List[Dict[Text, Any]]:
        entities = []
[...]

PR #7869, opened 2021-02-02
spaCy 3.0

koaning:
This PR should fix https://github.com/RasaHQ/rasa/issues/7724, fix https://github.com/RasaHQ/rasa/issues/6906, fix https://github.com/RasaHQ/rasa/issues/5400 and add support for **many** languages! 

This PR will add basic support for spaCy 3.0.  I'll list a bunch of things that have changed. 

- I ran `poetry update` which caused a lot of changes in `poetry.lock`
- the `spacy link` command has been deprecated. That means that some of the Dockerfiles needed to be rewritten as well as our `Makefile`. This also means that this PR contains breaking changes. You *must* name your spaCy model explicitly in a pipeline from now on. 
- Every reference to a `SpacyNLP` component in the tests now needs a model attached. I've chosen to use `en_core_web_md` (for English) and `de_core_news_sm` (for German). These are already downloaded beforehand in the CI. Note that this means that *loads* of tests required an update.


koaning:
Milestone: the docker contains build!

koaning:
I think, as is, this PR is a nice feature. I haven't check support for the new transformer models though. These will require more work because the API is slightly different. It might be best to delay that feature, but I'm open to suggestions. 

Do we need to maybe do a performance check for models? The embeddings have been retrained if I recall correctly.

rasa/nlu/utils/spacy_utils.py
```diff
@@ -49,11 +48,9 @@ def load_model(spacy_model_name: Text) -> "Language":
...
             raise InvalidModelError(
-                "Model '{}' is not a linked spaCy model.  "
-                "Please download and/or link a spaCy model, "
-                "e.g. by running:\npython -m spacy download "
-                "en_core_web_md\npython -m spacy link "
-                "en_core_web_md en".format(spacy_model_name)
+                "Please confirm that {} is an available spaCy model. "
+                "You may need to download one upfront. For example:\npython -m spacy download "
+                "en_core_web_md".format(spacy_model_name)
             )
```
koaning, line 55: Some extra changes should be made to ensure that a user passes a model name. That means removing lines like [this](https://github.com/RasaHQ/rasa/pull/7869/files#diff-cb47a94f8eba5f3264f3da2701ac4d56dec845bca6a1d92915b7785a9e76be8fR72-R74) and replacing them with a checking mechanism that throws appropriate errors.

wochinge:
@koaning Can we make this not breaking by using the `language` setting from the domain as fallback? E.g. if the language is `en` we automatically use `en_core_web_md` ?

koaning:
> @koaning Can we make this not breaking by using the language setting from the domain as fallback? E.g. if the language is en we automatically use en_core_web_md (+ deprecation warning of course)

I guess we could ... but it gets very tricky. The English models use `en_core_web_md` but, for example, the [French](https://spacy.io/models/fr) models use `fr_core_news_md`. The naming standard `<lang>_core_<src>_<size>` technically cannot be assumed to be consistent across languages and spaCy doesn't support each language equally. We'd also need to assume that the user downloaded `en_core_web_md` (as opposed to `en_core_web_lg`) beforehand. 

From my perspective spaCy has made a very sensible change here, as it forces users to be explicit.

koaning:
@wochinge I'm wondering timeline-wise ... should I merge all of this work on top of the `main` branch or is it best to target a branch with a version? If so, which one might be practical? I'm asking because I imagine this PR will need a substantial amount of reviews.

wochinge:
> should I merge all of this work on top of the main branch or is it best to target a branch with a version?

As it's feature it should go into `main`. Do you expect more changes? Then might make sense to do separate PRs which target this branch.

> From my perspective spaCy has made a very sensible change here, as it forces users to be explicit, which is more maintainable."

I agree. My issues are just that we shouldn't have this sort of breaking changes in a minor. Couldn't we do
* map the main languages to explict model names
* print a big warning in case we have to fallback to this case ("please use the explicit model name")

In this case we would at least try?

koaning:
I'm certainly in favor of raising warnings/errors. I can get started on that right away (and I also need to make changes to the docs to reflect the new situation). 

> map the main languages to explicit model names

I'm uneasy about this because I fear it introduces an inconsistency across languages *and* it adds a legacy to the codebase that may fail implicitly. With that in mind, I'd argue it's better to keep it explicit for all languages and to make sure the warnings/errors are as clear as possible. 

@wochinge Would it be better to raise a warning here or to raise a proper error?

docker/Dockerfile.pretrained_embeddings_spacy_de
```diff
@@ -23,9 +23,8 @@ RUN . /opt/venv/bin/activate && poetry build -f wheel -n && \
 # make sure we use the virtualenv
 ENV PATH="/opt/venv/bin:$PATH"
 
-# spacy link
-RUN python -m spacy download de_core_news_sm && \
-    python -m spacy link de_core_news_sm de
+# spacy link is deprecated in v3.0
```
joejuzl, line 26: Does this comment make sense out of the context of this PR?

tests/nlu/test_train.py
```diff
@@ -26,8 +26,11 @@ def get_test_params_for_component(component: Text) -> Dict[Text, Union[Text, int
     )
 
 
-def as_pipeline(*components):
-    return [{**{"name": c}, **get_test_params_for_component(c)} for c in components]
+def as_pipeline(*components) -> List[Dict[Text, Dict]]:
+    return [
+        {**{"name": c}, **get_test_params_for_component(c)} if isinstance(c, str) else c
```
joejuzl, line 31: I think the first `**` is unneeded and `"name": c` can just be inlined.

data/test_config/config_spacy_entity_extractor.yml
```diff
@@ -1,6 +1,7 @@
 language: en
 pipeline:
     - name: "SpacyNLP"
+      model: "en_core_web_md"
```
joejuzl, line 4: Do we also need to make this change to all the example bots?

joejuzl (changes requested):
I agree with @wochinge that we shouldn't have a breaking change like this in a minor release. Have you thought more about how to make this backwards compatible with old configs?

Also can we have a changelog, and maybe also a migration guide?

data/test_config/config_spacy_entity_extractor.yml
koaning: If they use spaCy, yes. I'm reminded that this would also include all examples in the docs. Note that we're not introducing the `model` variable, we're merely requiring that it's filled in explicitly. 

Just to check, @joejuzl what example bots were you thinking about? Our starter packs?

docker/Dockerfile.pretrained_embeddings_spacy_de
koaning: Fair! I'll drop the comment.

tests/nlu/test_train.py
koaning: This change was generated by black but I agree that we don't need the first `**`.

koaning:
> Have you thought more about how to make this backwards compatible with old configs?

@joejuzl I guess I could add a dictionary that internally links the most popular languages. 

```
en -> en_core_web_md
es -> es_core_news_md
de -> de_core_news_sm
zh -> zh_core_web_md
pt -> pt_core_news_md
nl -> nl_core_news_md
pl -> pl_core_news_md
```

But I would be **strongly** in favour of deprecating this as soon as possible. It's a bad pattern and we don't want to maintain this in a way that is consistent with any new languages that spaCy might start supporting. Note that this could make users confused, especially when they made a link before via `spacy link en_core_web_lg en`. We might grab a model with lower resolution and the users might wonder why their model performance changes. 

Also note that even if we assume `en_core_web_md` to be the standard for `en` that there is no guarantee that `en_core_web_md` is actually downloaded. 

There's something about making this assumption upfront that doesn't feel right to me. If you merely need the tokeniser then it's much better to use `de_core_news_sm` (which is what we use in tests). If you need the word embeddings then suddenly we would recommend `de_core_news_md`.

If you @joejuzl and @wochinge feel very strongly about this then I will implement this. But it really feels like a compromise that's going to cause more headaches in the long run.

data/test_config/config_spacy_entity_extractor.yml
wochinge: I think Joe means the one in https://github.com/RasaHQ/rasa/tree/main/examples

wochinge:
> Note that this could also make users confused, especially when they made a link before via spacy link en_core_web_lg en. We might grab a model with lower resolution and the users might wonder why their model performance changes.

We should print a loud warning for the case that they haven't specified the model and we have to attempt to use our custom language-model-mapping. That should in my opinion make it very clear for users.

> But I would be strongly in favour of deprecating this as soon as possible. 

I'd deprecate asap but we can only remove this behavior in the next Rasa Open Source major. 

>  we can't maintain this in a way that is guaranteed to be consistent with any new languages that spaCy starts supporting

That's fine in my opinion. They will read the updated Rasa Open Source docs and hopefully not even try to link the spacy model via the language. It's more about historic Rasa-Spacy users.

> If you @joejuzl and @wochinge feel very strongly about this then I will implement this. 

I still think we should at least attempt to make this non breaking and have a very explicit deprecation message (or multiple 😆 ) to move them towards explicitly specifying this.

pyproject.toml
```diff
@@ -170,7 +170,7 @@ version = "^0.4.4"
 markers = "sys_platform == 'win32'"
 
 [tool.poetry.dependencies.spacy]
-version = ">=2.1,<2.3"
+version = ">=3.0"
```
wochinge, line 173: we should also update the docs [here](If you @joejuzl and @wochinge feel very strongly about this then I will implement this. )

joejuzl:
> I still think we should at least attempt to make this non breaking and have a very explicit deprecation message (or multiple 😆 ) to move them towards explicitly specifying this.

This. I think we all agree that automatic mapping is not the way forward. We just shouldn't introduce a breaking change right now.

rasa/nlu/utils/spacy_utils.py
```diff
@@ -49,11 +48,9 @@ def load_model(spacy_model_name: Text) -> "Language":
             return spacy.load(spacy_model_name, disable=["parser"])
         except OSError:
             raise InvalidModelError(
-                "Model '{}' is not a linked spaCy model.  "
-                "Please download and/or link a spaCy model, "
-                "e.g. by running:\npython -m spacy download "
-                "en_core_web_md\npython -m spacy link "
-                "en_core_web_md en".format(spacy_model_name)
+                "Please confirm that {} is an available spaCy model. "
```
wochinge, line 51: optional: f-strings for the win.

pyproject.toml
koaning: I think you tried to copy a link but pasted in a sentence instead, what part of the docs require this?

koaning:
All fair feedback. I'll be making a second iteration ASAP. Thanks for the pointers @wochinge @joejuzl!

data/test_config/config_spacy_entity_extractor.yml
koaning: Done. Only `moodbot` and `reminderbot` used spaCy.

koaning:
I've just added warnings. This is the log output from `rasa train` without a model configured for `en`.

### Fallback Logs

```yaml
language: en

pipeline:
   - name: SpacyNLP
```

```
> rasa train --verbose
2021-03-04 11:49:13 INFO     rasa.model  - Data (nlu-config) for NLU model section changed.
Training NLU model...
2021-03-04 11:49:15 WARNING  rasa.nlu.utils.spacy_utils  - SpaCy model is not properly configured! Please add a `model` property to `SpacyNLP`.
2021-03-04 11:49:15 WARNING  rasa.nlu.utils.spacy_utils  - Will try to recover a model based on language name: 'en'.
2021-03-04 11:49:15 WARNING  rasa.nlu.utils.spacy_utils  - Will use 'en_core_web_md' as a fallback spaCy model.
2021-03-04 11:49:15 WARNING  rasa.nlu.utils.spacy_utils  - Please update the configuraton as soon as possible. This fallback will be deprecated in Rasa 3.0
2021-03-04 11:49:15 INFO     rasa.nlu.utils.spacy_utils  - Trying to load spacy model with name 'en_core_web_md'
2021-03-04 11:49:16 INFO     rasa.nlu.components  - Added 'SpacyNLP' to component cache. Key 'SpacyNLP-en_core_web_md'.
...
```

### Proper Settings

When `en_core_web_md` is configured the warnings go away. 

```yaml
language: en

pipeline:
   - name: SpacyNLP
     model: en_core_web_md
```

```
2021-03-04 11:51:35 INFO     rasa.model  - Data (nlu-config) for NLU model section changed.
Training NLU model...
2021-03-04 11:51:37 INFO     rasa.nlu.utils.spacy_utils  - Trying to load spacy model with name 'en_core_web_md'
2021-03-04 11:51:38 INFO     rasa.nlu.components  - Added 'SpacyNLP' to component cache. Key 'SpacyNLP-en_core_web_md'.
2021-03-04 11:51:38 INFO     rasa.shared.nlu.training_data.training_data  - Training data stats:
...

```

### Bad Language Setting 

Here's what happens when I change the language to `xx` and don't have a model selected either.

```yaml
language: xx

pipeline:
   - name: SpacyNLP
```

```
> rasa train --verbose
2021-03-04 11:48:35 INFO     rasa.model  - Data (nlu-config) for NLU model section changed.
Training NLU model...
2021-03-04 11:48:37 WARNING  rasa.nlu.utils.spacy_utils  - SpaCy model is not properly configured! Please add a `model` property to `SpacyNLP`.
2021-03-04 11:48:37 WARNING  rasa.nlu.utils.spacy_utils  - Will try to recover a model based on language name: 'xx'.
  File "/Users/vincent/Development/rasa/venv/bin/rasa", line 5, in <module>
    main()
  File "/Users/vincent/Development/rasa/rasa/__main__.py", line 116, in main
  File "/Users/vincent/Development/rasa/rasa/cli/train.py", line 58, in <lambda>
    train_parser.set_defaults(func=lambda args: train(args, can_exit=True))
  File "/Users/vincent/Development/rasa/rasa/cli/train.py", line 102, in train
    finetuning_epoch_fraction=args.epoch_fraction,
  File "/Users/vincent/Development/rasa/rasa/train.py", line 109, in train
    loop,
  File "/Users/vincent/Development/rasa/rasa/utils/common.py", line 308, in run_in_loop
  File "/Users/vincent/Development/rasa/rasa/train.py", line 174, in train_async
    finetuning_epoch_fraction=finetuning_epoch_fraction,
  File "/Users/vincent/Development/rasa/rasa/train.py", line 353, in _train_async_internal
    finetuning_epoch_fraction=finetuning_epoch_fraction,
  File "/Users/vincent/Development/rasa/rasa/train.py", line 396, in _do_training
    finetuning_epoch_fraction=finetuning_epoch_fraction,
  File "/Users/vincent/Development/rasa/rasa/train.py", line 819, in _train_nlu_with_validated_data
    **additional_arguments,
  File "/Users/vincent/Development/rasa/rasa/nlu/train.py", line 98, in train
    nlu_config, component_builder, model_to_finetune=model_to_finetune
  File "/Users/vincent/Development/rasa/rasa/nlu/model.py", line 163, in __init__
    self.pipeline = self._build_pipeline(cfg, component_builder)
  File "/Users/vincent/Development/rasa/rasa/nlu/model.py", line 174, in _build_pipeline
    component = component_builder.create_component(component_cfg, cfg)
  File "/Users/vincent/Development/rasa/rasa/nlu/components.py", line 852, in create_component
    component_config, Metadata(cfg.as_dict(), None)
  File "/Users/vincent/Development/rasa/rasa/nlu/components.py", line 768, in __get_cached_component
    cache_key = component_class.cache_key(component_meta, model_metadata)
  File "/Users/vincent/Development/rasa/rasa/nlu/utils/spacy_utils.py", line 85, in cache_key
    spacy_model_name = cls.check_model_fallback(component_meta.get("model"), model_metadata.language, warn=True)
  File "/Users/vincent/Development/rasa/rasa/nlu/utils/spacy_utils.py", line 119, in check_model_fallback
    spacy_model_name = fallback_mapping[language_name]
KeyError: 'xx'
```

### Quick Question 

@joejuzl @wochinge do we want to have a unit test for these warnings?

wochinge:
> ny comments on these warnings? Do we also want to have a unit test for these warnings?

Tests would be great 👍🏻 How about adding a links to the docs / migration guide in the warning?

pyproject.toml
wochinge: 🤦🏻 I mean this https://rasa.com/docs/rasa/installation#dependencies-for-spacy

rasa/nlu/utils/spacy_utils.py
```diff
@@ -90,10 +84,57 @@ def cache_key(
...
+    @classmethod
+    def check_model_fallback(cls, spacy_model_name, language_name, warn=False):
+        """
+        This method checks if the `spacy_model_name` is missing and attempts to do a fallback.
+        This feature is a measure to support spaCy 3.0 without breaking on users. In the future
+        spaCy will no longer support `spacy link`.
+        """
+        if not spacy_model_name:
+            if warn:
+                logger.warning(
```
wochinge, line 102: should be a `warnings.warn` as it's user facing (see https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e)

koaning:
> Tests would be great 👍🏻 How about adding a link to the docs/migration guide in the warning?

@wochinge  Yeah that's my next step. But I prefer to do the docs once I'm 100% sure about the features/tests. I'll add these in my next iteration :)

tests/nlu/test_spacy_utils.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+
+from rasa.nlu.model import InvalidModelError
+from rasa.nlu.config import RasaNLUModelConfig
+from rasa.nlu.utils.spacy_utils import SpacyNLP
+
+
+def test_model_fallback_raises_error():
```
koaning, line 8: @wochinge is there a way to flag these tests that they should be removed in Rasa 3.0?

koaning:
@wochinge @joejuzl I've added docs, changelog and links. 

I'm assuming we still need to review a few things but it might be time to ask the question: do we want to pin this on a release? This feature would be something we'd like to brag a bit about to the community so it would help if we could plan it.

wochinge:
Do you think we can make 2.4? Happy to provide reviews - note that Monday is a holiday in Germany. Agree that we want to brag with this 🙂

koaning:
That will depend on your availability next week then. 

One thing, should I worry about those coveralls?

wochinge:
Coveralls is a bit all over the place at the moment (cc @tczekajlo ) 😬

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +83,82 @@ def cache_key(
...
+                "mk": "mk_core_news_md",
+                "nb": "nb_core_news_md",
+                "pl": "pl_core_news_md",
+                "pt": "pt_core_news_md",
+                "ro": "ro_core_news_md",
+                "ru": "ru_core_news_md",
+                "es": "es_core_news_md",
+            }
+            if warn:
+                warnings.warn(
```
wochinge, line 120: lets please use `raise_deprecation_warning`. I'd also suggest to have only one warning instead of 4 which is raised in case of the fallback is attempted

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +83,82 @@ def cache_key(
...
-        spacy_model_name = component_meta.get("model", model_metadata.language)
+        spacy_model_name = cls.check_model_fallback(
+            component_meta.get("model"), model_metadata.language, warn=True
+        )
 
         return cls.name + "-" + spacy_model_name
 
+    @classmethod
+    def check_model_fallback(cls, spacy_model_name, language_name, warn=False):
+        """This method checks if the `spacy_model_name` is missing and attempts to do a fallback.
```
wochinge, line 94: plesae wrap the docstrings after 88 chars to be aligned with our maximum code width.

tests/nlu/test_spacy_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    "pl",
+    "pt",
+    "ro",
+    "ru",
+    "es",
+]
+
+
+@pytest.mark.parametrize("lang", supported_langauges)
+def test_model_fallback_raises_warning(lang):
```
wochinge, line 35: please add type annotations

tests/nlu/test_spacy_utils.py
wochinge: no need in my opinion. The test fails when removing the production code anyway. 
Btw, isn't this test already covered by the test function at the bottom of this module?

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +83,82 @@ def cache_key(
...
-        # as the model name if no explicit name is defined
-        spacy_model_name = component_meta.get("model", model_metadata.language)
+        spacy_model_name = cls.check_model_fallback(
+            component_meta.get("model"), model_metadata.language, warn=True
+        )
 
         return cls.name + "-" + spacy_model_name
 
+    @classmethod
+    def check_model_fallback(cls, spacy_model_name, language_name, warn=False):
```
wochinge, line 93: let's make this protected as it shouldn't be used outside of it.
```suggestion
    def _check_model_fallback(cls, spacy_model_name, language_name, warn=False):
```

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +83,82 @@ def cache_key(
         cls, component_meta: Dict[Text, Any], model_metadata: "Metadata"
     ) -> Optional[Text]:
 
-        # Fallback, use the language name, e.g. "en",
-        # as the model name if no explicit name is defined
-        spacy_model_name = component_meta.get("model", model_metadata.language)
+        spacy_model_name = cls.check_model_fallback(
+            component_meta.get("model"), model_metadata.language, warn=True
```
wochinge, line 87: Shouldn't we always warn except in this case?

koaning: Do we want to raise this warning when a user loads a model from disk? 

In my mind we want to raise this warning when we are training the model and this should only be raised once. The `_check_model_fallback` method is called in `SpacyNLP.create`, `SpacyNLP.load` and `SpacyNLP.cache_key` so I added the flag in order to not overflow the users retina with warnings. I can add it in all places, but the user will see the error multiple times during training.

changelog/7869.feature.md
```diff
@@ -0,0 +1,42 @@
+Upgraded Rasa to be compatible with spaCy 3.0.
```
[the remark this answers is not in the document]
koaning: I just added a migration guide portion with a variant of this information. 

@wochinge I think here I am supposed to use headers. Am I correct?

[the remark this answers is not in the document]
wochinge: so correct!

docs/docs/components.mdx
```diff
@@ -130,12 +127,12 @@ word vectors in your pipeline.
...
-  If you want to incorporate a custom model you've found into spaCy, check out their page on
-  [adding languages](https://spacy.io/usage/adding-languages/). As described in the documentation, you need to
-  register your language model and link it to the language identifier, which will allow Rasa to load and use your new language
-  by passing in your language identifier as the `language` option.
+  attach spaCy models that you've trained yourself.
 
+:::caution Fallback
+Rasa Open Source will try to fallback to a common model on your behalf if you don't pass a `model` setting. This is a
+temporary feature we've introduced as part of the spaCy 3.0 migration but the fallback will be removed in a future release.
+:::
```
wochinge, line 135:
```suggestion
temporary feature we've introduced as part of the spaCy 3.0 migration but the fallback will be removed in Rasa Open Source 3.0.0.
:::
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,51 @@ description: |
...
+pipeline:
+   - name: SpacyNLP
+     model: en_core_web_md
+```
+
+**Fallback**
+
+To make the transition easier, Rasa will try to fall back to a medium spaCy model when-ever
+a compatible language is configured for the entire pipeline in `config.yml` even if you don't
+specify a `model`. This fallback behavior is temporary and will be deprecated in spaCy 3.0.
```
wochinge, line 52: spacy spacy spacy spacy
```suggestion
specify a `model`. This fallback behavior is temporary and will be deprecated in Rasa Open Source 3.0.0.
```

wochinge (commented):
Let me know once you've integrated these and the previous comments, then I'll give it another review.

JEM-Mosig (approved):
Looks good to me! Just a few minor comments on formulation.

koaning:
@wochinge per your advice I replaced the `poetry.lock` file with contents generated from a `poetry lock` command. From my end it seems like I've ticked off all of your comments. @JEM-Mosig has also given this work a review on behalf of the research team. From my end, it feels like it's time for a final review ... or maybe even a merge 😄 !

docs/docs/migration-guide.mdx
wochinge: @koaning I don't think this addressed yet based on the current diff.

pyproject.toml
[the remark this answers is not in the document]
wochinge: @koaning What about this one?

wochinge:
@koaning I see multiple comments which aren't addressed yet 😬 Maybe you missed pushing?

koaning:
@wochinge That's ... strange. Let me double check! I'll ping you once I'm 100% sure.

docs/docs/migration-guide.mdx
koaning: Confirmed now. https://github.com/RasaHQ/rasa/pull/7869/files#diff-144c9993b9753bf4e93513f5c220747c493570dd472eefc19e701791e0da8622R52

docs/docs/components.mdx
koaning: Also confirmed https://github.com/RasaHQ/rasa/pull/7869/files#diff-a1167895182b95a8cde4e98af355375597a90d92e76698823e2a61ac4d5521f8R134

koaning:
@wochinge I *think* everything should now be aligned. I'm still not 100% sure what happened but I can confirm that your comments have been added [here](https://github.com/RasaHQ/rasa/pull/7869/files#diff-4f7520efb414d462bbc72e9b20c6ab289b1237f13c146a1115e6943c1dad513fR36) and [here](https://github.com/RasaHQ/rasa/pull/7869/files#diff-144c9993b9753bf4e93513f5c220747c493570dd472eefc19e701791e0da8622R52). The unit tests/style checks also all look green now.

pyproject.toml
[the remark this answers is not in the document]
wochinge: @koaning This one still doesn't seem to be adressed 🙈

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +83,82 @@ def cache_key(
...
+        )
 
         return cls.name + "-" + spacy_model_name
 
+    @classmethod
```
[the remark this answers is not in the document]
wochinge: @koaning

rasa/nlu/utils/spacy_utils.py
```diff
@@ -49,11 +49,9 @@ def load_model(spacy_model_name: Text) -> "Language":
             return spacy.load(spacy_model_name, disable=["parser"])
         except OSError:
             raise InvalidModelError(
-                "Model '{}' is not a linked spaCy model.  "
-                "Please download and/or link a spaCy model, "
-                "e.g. by running:\npython -m spacy download "
-                "en_core_web_md\npython -m spacy link "
-                "en_core_web_md en".format(spacy_model_name)
+                f"Please confirm that {spacy_model_name} is an available spaCy model. "
+                "You need to download one upfront. For example:\npython -m spacy download "
```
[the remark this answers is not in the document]
wochinge: @koaning

rasa/nlu/utils/spacy_utils.py
wochinge: @koaning

wochinge (changes requested):
Sorry, but they still don't seem to be adressed / discussed 😄

rasa/nlu/utils/spacy_utils.py
koaning: Are you referring to a function called `raise_deprecation_warning`? If so, I couldn't find one in Rasa so I assumed you are referring to raising a `DeprecationWarning` here? I've moved everything into a single warning.

koaning:
@wochinge thanks for pointing out the remaining flaws. 

Once again, I *think* the changes are now properly addressed. I didn't resolve any issues myself to make it easy for you to confirm.

There's a new issue though, it seems like the doc tests aren't passing. Since the issue revolves around dead links on other parts of the repo, can I safely ignore these?

koaning:
Ah. It seems that we're pointing users to a sanic doc that no longer exists. Will investigate.

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +84,71 @@ def cache_key(
...
+            if language_name not in fallback_mapping.keys():
+                raise InvalidModelError(
+                    f"There is no fallback model for language '{language_name}'."
+                    f" Please add a `model` property to `SpacyNLP` manually to prevent this."
+                    f"More informaton can be found on https://rasa.com/docs/rasa/components#spacynlp"
+                )
+
+            spacy_model_name = fallback_mapping[language_name]
+            if warn:
+                warnings.warn(
```
wochinge, line 132: can we please use https://github.com/RasaHQ/rasa/blob/92f5466c3b5c8ff0c4bf2abcf77e3bc091879f83/rasa/shared/utils/io.py#L530

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +84,71 @@ def cache_key(
...
+                "pt": "pt_core_news_md",
+                "ro": "ro_core_news_md",
+                "ru": "ru_core_news_md",
+                "es": "es_core_news_md",
+            }
+            if language_name not in fallback_mapping.keys():
+                raise InvalidModelError(
+                    f"There is no fallback model for language '{language_name}'."
+                    f" Please add a `model` property to `SpacyNLP` manually to prevent this."
+                    f"More informaton can be found on https://rasa.com/docs/rasa/components#spacynlp"
```
wochinge, line 127:
```suggestion
                    f"More informaton can be found on {DOCS_URL_COMPONENTS}#spacynlp"
```

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +84,71 @@ def cache_key(
...
+                )
+
+            spacy_model_name = fallback_mapping[language_name]
+            if warn:
+                warnings.warn(
+                    DeprecationWarning(
+                        f"SpaCy model is not properly configured! Please add a `model` property to `SpacyNLP`."
+                        f"More informaton can be found on https://rasa.com/docs/rasa/components#spacynlp"
+                        f"Will use '{spacy_model_name}' as a fallback spaCy model."
+                        f"This fallback will be deprecated in Rasa 3.0"
```
wochinge, line 137:
```suggestion
                        f"SpaCy model is not properly configured! Please add a `model` property to `SpacyNLP`. "
                        f"More informaton can be found on {DOCS_URL_COMPONENTS}#spacynlp . "
                        f"Will use '{spacy_model_name}' as a fallback spaCy model."
                        f"This fallback will be deprecated in Rasa 3.0"
```

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +84,71 @@ def cache_key(
...
         return {"spacy_nlp": self.nlp}
 
     def doc_for_text(self, text: Text) -> "Doc":
-
+        """Makes a spaCy doc object from a string of text."""
         return self.nlp(self.preprocess_text(text))
 
     def preprocess_text(self, text: Optional[Text]) -> Text:
-
+        """Process the text before it is handled by spaCy."""
```
wochinge, line 151:
```suggestion
        """Processes the text before it is handled by spaCy."""
```

wochinge (commented):
nearly there 💯

rasa/nlu/utils/spacy_utils.py
```diff
@@ -129,13 +131,13 @@ def _check_model_fallback(
...
-                warnings.warn(
-                    DeprecationWarning(
-                        f"SpaCy model is not properly configured! Please add a `model` property to `SpacyNLP`. "
-                        f"More informaton can be found on {DOCS_URL_COMPONENTS}#spacynlp . "
-                        f"Will use '{spacy_model_name}' as a fallback spaCy model."
-                        f"This fallback will be deprecated in Rasa 3.0"
-                    )
+                message = (
+                    f"SpaCy model is not properly configured! Please add a `model` property to `SpacyNLP`. "
+                    f"Will use '{spacy_model_name}' as a fallback spaCy model."
```
wochinge, line 136:
```suggestion
                    f"Will use '{spacy_model_name}' as a fallback spaCy model. "
```

rasa/nlu/utils/spacy_utils.py
```diff
@@ -1,6 +1,7 @@
-import logging
 import typing
-from typing import Any, Dict, List, Optional, Text, Tuple
+import logging
+import warnings
```
wochinge, line 3:
```suggestion
```

rasa/nlu/utils/spacy_utils.py
```diff
@@ -49,11 +51,10 @@ def load_model(spacy_model_name: Text) -> "Language":
...
             raise InvalidModelError(
-                "Model '{}' is not a linked spaCy model.  "
-                "Please download and/or link a spaCy model, "
-                "e.g. by running:\npython -m spacy download "
-                "en_core_web_md\npython -m spacy link "
-                "en_core_web_md en".format(spacy_model_name)
+                f"Please confirm that {spacy_model_name} is an available spaCy model. "
+                f"You need to download one upfront. For example:\npython -m spacy download "
+                f"en_core_web_md\n"
+                f"More informaton can be found on https://rasa.com/docs/rasa/components#spacynlp"
```
wochinge, line 57:
```suggestion
                f"More informaton can be found on {DOCS_URL_COMPONENTS}#spacynlp"
```

rasa/nlu/utils/spacy_utils.py
```diff
@@ -88,21 +86,71 @@ def cache_key(
...
+                "pt": "pt_core_news_md",
+                "ro": "ro_core_news_md",
+                "ru": "ru_core_news_md",
+                "es": "es_core_news_md",
+            }
+            if language_name not in fallback_mapping.keys():
+                raise InvalidModelError(
+                    f"There is no fallback model for language '{language_name}'."
+                    f" Please add a `model` property to `SpacyNLP` manually to prevent this."
+                    f"More informaton can be found on {DOCS_URL_COMPONENTS}#spacynlp"
```
wochinge, line 129:
```suggestion
                    f"There is no fallback model for language '{language_name}'. "
                    f"Please add a `model` property to `SpacyNLP` manually to prevent this. "
                    f"More informaton can be found on {DOCS_URL_COMPONENTS}#spacynlp"
```

wochinge (approved):
Looking good 🚀 

@joejuzl is it okay to dismiss your "Request for changes"?

rasa/nlu/utils/spacy_utils.py
```diff
@@ -9,6 +10,8 @@
 from rasa.shared.nlu.training_data.message import Message
 from rasa.nlu.model import InvalidModelError
 from rasa.nlu.constants import SPACY_DOCS, DENSE_FEATURIZABLE_ATTRIBUTES
+from rasa.shared.utils.io import raise_deprecation_warning
```
wochinge, line 13:
```suggestion
rasa.shared.utils.io
```

According to our Python code conventions we don't import functions directly but reference them by their module.  https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#5b63fd5269514270aaf61ab8303ae319

joejuzl (approved):
Nice job!

koaning:
Booya!

wochinge:
Thanks for this awesome change @koaning ! 💯

Issue #7887, opened 2021-02-04
Reminder Events Don't Acquire a Conversation Lock


b-quachtran:
**Rasa version**:
1.10.X/2.X

**Issue**:

Reminder Events don't acquire a conversation lock when the event is executed. This can cause an issue if the tracker is simultaneously being updated by another endpoint / channel.

Issue #5257, opened 2020-02-18
Refactor Agent / Processor / TrackerStore


wochinge:
## The Problem

*   the `Agent` class is big and hard to maintain
*   new types of models are hard to implement (e.g. the response selector)

## Goals

*   Make it easy to implement different / new types of models without having to squeeze it into the existing code
*   De-couple our classes to make Rasa Open Source more modular and easier adapt to change
*   `Agent`: 
      *   separate model loading logic from class
    *   separate training from agent class
    *   remove duplicate code between `Agent` / `Processor`
*   `TrackerStore`: Separate it from the model specific domain (the `TrackerStore` should not know of the model domain)
*   make this not breaking with the existing Rasa Open Source models

## Implementation Proposal

*   This is about adding a layer before the actual model implementations. So we would e.g. have to adapt our `Processor` class to the `Model` interface but could use this then as before

#### Class hierarchy

[<span class="image-wrap" style="">![](https://user-images.githubusercontent.com/7667273/74769451-b8589900-528a-11ea-8c99-0a2ef575bb67.png)</span>](https://user-images.githubusercontent.com/7667273/74769451-b8589900-528a-11ea-8c99-0a2ef575bb67.png)

#### Running a training

This UML diagram how the workflow is when running `rasa train` It shows how

*   a model is trained using a `TrainingDataImporter`
*   the `ModelFactory` is selected based on a default value / config

[<span class="image-wrap" style="">![](https://user-images.githubusercontent.com/7667273/74769466-be4e7a00-528a-11ea-82cb-9f98371904a9.png)</span>](https://user-images.githubusercontent.com/7667273/74769466-be4e7a00-528a-11ea-82cb-9f98371904a9.png)

#### Inference Mode / Interacting with a running Rasa server

It shows how

*   the interaction works with the changed `TrackerStore` (returns a list of events instead of a DialogueStateTracker
*   getting response from a model which is behind a `Model` interface

[<span class="image-wrap" style="">![](https://user-images.githubusercontent.com/7667273/74769477-c3abc480-528a-11ea-8e6c-a3a667afd668.png)</span>](https://user-images.githubusercontent.com/7667273/74769477-c3abc480-528a-11ea-8e6c-a3a667afd668.png)

#### Starting a Rasa Server and loading a model

This shows how:

*   the right repository / factory classes are selected to load the desired model type
*   a model is loaded from disk
*   how models are re-loaded when using a `ModelServer` (that's the most vague part I'd say)

[...]

PR #7895, opened 2021-02-05
Added Locking Mechanism to Reminders Handler

b-quachtran:
**Proposed changes**:
- Added lock_store as an input to MessageProcessor so that handle_reminder can properly lock the conversation
- Updated handle_reminder to lock the conversation before retrieving the tracker store & executing
- Updated default_processor to pass in InMemoryLockStore as a param for test cases
- Resolves Issue #7887 


b-quachtran:
@wochinge I added lock_store as an input parameter to MessageProcessor so that handle_reminder could lock the conversation, but not entirely sure if this is the best way to do it. Wdyt?

rasa/core/processor.py
```diff
@@ -348,28 +351,31 @@ async def handle_reminder(
...
-            logger.debug(
-                f"Canceled reminder because it is outdated ({reminder_event})."
-            )
-        else:
-            intent = reminder_event.intent
-            entities = reminder_event.entities or {}
-            await self.trigger_external_user_uttered(
-                intent, entities, tracker, output_channel
-            )
+            if not tracker:
```
wochinge, line 359: the tracker can't be `None` in this case

rasa/core/processor.py
```diff
@@ -348,28 +351,31 @@ async def handle_reminder(
     ) -> None:
         """Handle a reminder that is triggered asynchronously."""
 
-        tracker = await self.get_tracker_with_session_start(sender_id, output_channel)
-
-        if not tracker:
-            logger.warning(
-                f"Failed to retrieve tracker for conversation ID '{sender_id}'."
+        async with self.lock_store.lock(sender_id):
```
wochinge, line 354: can you please add a test to avoid regression?

wochinge (commented):
Thanks for providing the fix! Will you also implement that for 2.x?

b-quachtran:
@wochinge Yes, definitely. I'll create a PR for 2.X as well once this fix is good to go

tests/core/test_processor.py
```diff
@@ -181,6 +181,38 @@ async def test_reminder_scheduled(
     assert t.events[-1] == ActionExecuted("action_listen")
 
 
+async def test_reminder_lock(
+    default_channel: CollectingOutputChannel, default_processor: MessageProcessor
+):
+    from io import StringIO
+
+    logger = logging.getLogger("rasa.core.lock_store")
```
wochinge, line 189: please use the caplog fixture instead (e.g. see [here](https://github.com/RasaHQ/rasa/blob/9913c1d53fd8a78b4564ebabb18a0bde9a939611/tests/core/test_tracker_stores.py#L274) on its usage)

rasa/core/agent.py
```diff
@@ -888,6 +888,7 @@ def create_processor(
             self.policy_ensemble,
             self.domain,
             self.tracker_store,
+            self.lock_store,
```
wochinge, line 891: not a fan of handing down intrastructure components to the `processor` but should in my opinion addressed separately (e.g. as part of https://github.com/RasaHQ/rasa/issues/5257)

wochinge (commented):
did you also test this manually? If so, how?

rasa/core/agent.py
b-quachtran: I couldn't think of a better way to do this without requiring re-factoring. Is there a better approach to this than passing it into processor?

b-quachtran:
@wochinge Yes, I tested manually using the ReminderBot with redis-server running locally and the SocketIO channel. Redis logs showed the lock correctly being acquired and deleted when the reminder was handled.

rasa/core/agent.py
wochinge: Also can't think of one for now. I think it's fine for now 👍🏻

changelog/7895.bugfix.md
```diff
@@ -0,0 +1 @@
+Fixed bug where the conversation does not lock before handling a reminder event.
```
wochinge, line 1: this has to be a `rst` changelog for `1.x`

wochinge (approved):

wochinge:
Nice job 💯

PR #7902, opened 2021-02-08
Finetuning (--finetune) with a specific model as an argument

Imod7:
**Proposed changes**:
- Adding an `else` statement in `rasa/cli/train.py._model_for_finetuning`

- [x] added one test for the functionality

### DIFF
--- a/changelog/7829.bugfix.md
+++ b/changelog/7829.bugfix.md
@@ -1,0 +1,1 @@
+Fine-tuning by passing a specific model path with the `--finetune` argument.
--- a/tests/cli/test_rasa_train.py
+++ b/tests/cli/test_rasa_train.py
@@ -572,12 +572,4 @@ def test_train_nlu_finetune_with_model(
     model_name = "models" + os.path.join(files[0])[0] + os.path.basename(files[0])
     output = run_in_simple_project_with_model("train", "nlu", "--finetune", model_name)
 
-    assert [
-        s
-        for s in output.outlines
-        if "Loading NLU model from models"
-        + os.path.join(files[0])[0]
-        + os.path.basename(files[0])
-        + " for finetuning..."
-        in s
-    ]
+    assert any(os.path.basename(files[0]) in line for line in output.outlines)
--- a/tests/cli/test_rasa_train.py
+++ b/tests/cli/test_rasa_train.py
@@ -569,7 +569,7 @@ def test_train_nlu_finetune_with_model(
     files = rasa.shared.utils.io.list_files(os.path.join(temp_dir, "models"))
     assert len(files) == 1
 
-    model_name = "models" + os.path.join(files[0])[0] + os.path.basename(files[0])
+    model_name = os.path.relpath(files[0])
     output = run_in_simple_project_with_model("train", "nlu", "--finetune", model_name)
 
     assert any(os.path.basename(files[0]) in line for line in output.outlines)
--- a/tests/cli/test_rasa_train.py
+++ b/tests/cli/test_rasa_train.py
@@ -570,11 +570,14 @@ def test_train_nlu_finetune_with_model(
     assert len(files) == 1
 
     output = run_in_simple_project_with_model("train", "nlu", "--finetune")
+    breakpoint()
+    assert any(os.path.basename(files[0]) in line for line in output.outlines)
 
     assert [
         s
         for s in output.outlines
-        if "Loading NLU model from models/"
+        if "Loading NLU model from models"
+        + os.path.join(files[0])[0]
         + os.path.basename(files[0])
         + " for finetuning..."
         in s
--- a/tests/cli/test_rasa_train.py
+++ b/tests/cli/test_rasa_train.py
@@ -569,7 +569,8 @@ def test_train_nlu_finetune_with_model(
     files = rasa.shared.utils.io.list_files(os.path.join(temp_dir, "models"))
     assert len(files) == 1
 
-    output = run_in_simple_project_with_model("train", "nlu", "--finetune")
+    model_name = "models" + os.path.join(files[0])[0] + os.path.basename(files[0])
+    output = run_in_simple_project_with_model("train", "nlu", "--finetune", model_name)
 
     assert [
         s
--- a/tests/cli/test_rasa_train.py
+++ b/tests/cli/test_rasa_train.py
@@ -558,3 +558,24 @@ def test_get_valid_config(parameters):
 def test_get_valid_config_with_non_existing_file():
     with pytest.raises(SystemExit):
         _get_valid_config("non-existing-file.yml", CONFIG_MANDATORY_KEYS)
+
+
+def test_train_nlu_finetune_with_model(
+    run_in_simple_project_with_model: Callable[..., RunResult]
+):
+    temp_dir = os.getcwd()
+
+    assert os.path.exists(os.path.join(temp_dir, "models"))
+    files = rasa.shared.utils.io.list_files(os.path.join(temp_dir, "models"))
+    assert len(files) == 1
+
+    output = run_in_simple_project_with_model("train", "nlu", "--finetune")
+
+    assert [
+        s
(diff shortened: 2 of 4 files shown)


joejuzl:
The fix looks good 👍  
Can we test it too please?  e.g. in `tests/cli/test_rasa_train.py`

tests/cli/test_rasa_train.py
```diff
@@ -558,3 +558,24 @@ def test_get_valid_config(parameters):
...
+def test_train_nlu_finetune_with_model(
+    run_in_simple_project_with_model: Callable[..., RunResult]
+):
+    temp_dir = os.getcwd()
+
+    assert os.path.exists(os.path.join(temp_dir, "models"))
+    files = rasa.shared.utils.io.list_files(os.path.join(temp_dir, "models"))
+    assert len(files) == 1
+
+    output = run_in_simple_project_with_model("train", "nlu", "--finetune")
```
Imod7, line 572: Should we also have the same test but without the `nlu` argument here? I already checked manually and it also gets the specified model in that case too.

tests/cli/test_rasa_train.py
```diff
@@ -558,3 +558,24 @@ def test_get_valid_config(parameters):
...
+
+    assert [
+        s
+        for s in output.outlines
+        if "Loading NLU model from models/"
+        + os.path.basename(files[0])
+        + " for finetuning..."
+        in s
+    ]
```
Imod7, line 581: - So here I only assert/check that the message "_Loading NLU model from models/...._" is printed in the standard output.
- I thought of adding the following asserts also (related to the result - if fine-tuning was successful or not-) :
     ```
    assert not any([s for s in output.outlines if
                    "NLU model can not be finetuned." in s])
    assert output.ret == 1
   ```
but I thought it is not necessary since we are only checking here if it can actually load/get the specified model and not if it can train it or not. If it is not the case, let me know and I will add them.

Imod7 (commented):
- Just to mention that this test checks the following case (the file retrieved is not prefixed with the `nlu` word in the beginning) :
`rasa train nlu --finetune models/20210127-140930.tar.gz` 
because in the issue you mention a slightly different case (filename with an `nlu` prefix)
`rasa train nlu --finetune models/nlu-20210127-140930.tar.gz`
but I guess it is ok. I do not know how to specify to `run_in_simple_project_with_model` to generate a file with `nlu` prefix.

- Something that I also noticed but not related directly to this Issue is that if I give : 
  `rasa train --finetune models/nlu-20210208-222928.tar.gz`
I get an error message _"NLU model can not be finetuned."_ which imo is not 100% accurate since the model can be trained but we just need to add the argument `nlu` in the command line. What do you think?

tests/cli/test_rasa_train.py
joejuzl, line 572: The functionality that is broken, and that we are trying to test is when the model name is passed to the argument `--finetune`.

tests/cli/test_rasa_train.py
joejuzl: Yeah I think just checking it loads it is enough.

joejuzl (changes requested):
Can you also add a changelog too please? :)

wochinge:
Would be great to get this into 2.3, what do you think?

Imod7:
@wochinge @joejuzl Yes it would be great to include it in the release. My only issue is that id does not pass the windows tests. Since there is no specific solution for that after my slack question [here](https://rasa-hq.slack.com/archives/C36SS4N8M/p1612878988073700?thread_ts=1607443058.364600&cid=C36SS4N8M), I am thinking of skipping the test in windows. What is your opinion on that ?

tests/cli/test_rasa_train.py
```diff
@@ -558,3 +558,24 @@ def test_get_valid_config(parameters):
...
+    assert os.path.exists(os.path.join(temp_dir, "models"))
+    files = rasa.shared.utils.io.list_files(os.path.join(temp_dir, "models"))
+    assert len(files) == 1
+
+    output = run_in_simple_project_with_model("train", "nlu", "--finetune")
+
+    assert [
+        s
+        for s in output.outlines
+        if "Loading NLU model from models/"
```
wochinge, line 577: i think it might be `models\` in Windows. In my opinion fine to do do something like:


```python
assert any(os.path.basename(files[0]) in line for line in output.outlines)
```

wochinge: I think skipping the test for Windows is nothing we should do as we need to ensure the same behavior on Windows

tests/cli/test_rasa_train.py
joejuzl: Still need to actually pass the model name.

tests/cli/test_rasa_train.py
Imod7: That is so accurate!!! I forgot about the pathnames difference. I will try that. Thank YOU!

tests/cli/test_rasa_train.py
Imod7: Very true!!! I added it just now.

tests/cli/test_rasa_train.py
Imod7: I changed it and now it sets the directory separator that finds in the file path (so if it's ubuntu it will take the slash and if it's windows it will find and take the backslash). Let's see if it works. Otherwise I will use your assert statement.

Imod7: I also tried your assert statement but it fails again with error message :
```
Training NLU model...
No NLU model for finetuning found. Please make sure to either specify a path to a previous model or to have a finetunable model within the directory 'models'.
```

tests/cli/test_rasa_train.py
```diff
@@ -558,3 +558,18 @@ def test_get_valid_config(parameters):
...
+    temp_dir = os.getcwd()
+
+    assert os.path.exists(os.path.join(temp_dir, "models"))
+    files = rasa.shared.utils.io.list_files(os.path.join(temp_dir, "models"))
+    assert len(files) == 1
+
+    model_name = os.path.relpath(files[0])
+    output = run_in_simple_project_with_model("train", "nlu", "--finetune", model_name)
+
+    assert any(os.path.basename(files[0]) in line for line in output.outlines)
```
wochinge, line 575: shouldn't it be like that?
```suggestion
    assert any(model_name in line for line in output.outlines)
```

wochinge: In my opinion we should check the specific model was used for finetuning. This only asserts the directory. What do you think?

tests/cli/test_rasa_train.py
```diff
@@ -558,3 +558,18 @@ def test_get_valid_config(parameters):
...
     with pytest.raises(SystemExit):
         _get_valid_config("non-existing-file.yml", CONFIG_MANDATORY_KEYS)
+
+
+def test_train_nlu_finetune_with_model(
+    run_in_simple_project_with_model: Callable[..., RunResult]
+):
+    temp_dir = os.getcwd()
+
+    assert os.path.exists(os.path.join(temp_dir, "models"))
```
wochinge, line 568: Why do we need this assert? In my opinion this tests only if the fixture `run_in_simple_project_with_model` works and that's not the scope of this test case.

Imod7: It is present in another test (that also does not check if the mentioned fixture works) and to me it looked like a good way to have more specific error messages in case the models folder does not exist. However it is true that we are not checking this functionality at this test so I will remove it.

wochinge (approved):

changelog/7829.bugfix.md
```diff
@@ -0,0 +1 @@
+Fine-tuning by passing a specific model path with the `--finetune` argument.
```
wochinge, line 1: I'll adapt the changelog a bit in the release PR to avoid another CI run.

PR #7911, opened 2021-02-09
7741 enforce marking tests that train

joejuzl:
**Proposed changes**:
This follows on from #7860 and enforces that the mark is added.
It splits the CI tests into two jobs one with and one without training tests.
If a test tries to train in the non-training job, it will fail.

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -276,7 +276,107 @@ jobs:
         JOBS: 2
         PYTHONIOENCODING: "utf-8"
       run: |
-        make test
+        make test-non-training
+
+    - name: Send Coverage Report 📊
+      if: needs.changes.outputs.backend == 'true' && matrix.python-version == 3.6 && matrix.os != 'windows-latest'
+      env:
+        COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        COVERALLS_SERVICE_NAME: github
+      run: poetry run coveralls
+
+  test_training:
+    name: Run Tests
+    runs-on: ${{ matrix.os }}
+    needs: [changes]
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        python-version: [3.6, 3.7, 3.8]
+
+    steps:
+    - name: Checkout git repository 🕝
+      if: needs.changes.outputs.backend == 'true'
+      uses: actions/checkout@v2
+
+    - name: Set up Python ${{ matrix.python-version }} 🐍
+      if: needs.changes.outputs.backend == 'true'
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Read Poetry Version 🔢
+      if: needs.changes.outputs.backend == 'true'
+      run: |
+        echo "POETRY_VERSION=$(scripts/poetry-version.sh)" >> $GITHUB_ENV
+      shell: bash
+
+    - name: Install poetry 🦄
+      if: needs.changes.outputs.backend == 'true'
+      uses: Gr1N/setup-poetry@v4
+      with:
+        poetry-version: ${{ env.POETRY_VERSION }}
+
+    - name: Load Poetry Cached Libraries ⬇
+      id: cache-poetry
+      if: needs.changes.outputs.backend == 'true'
+      uses: actions/cache@v2
+      with:
+        path: .venv
+        key: ${{ runner.os }}-poetry-${{ env.POETRY_VERSION }}-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}-venv-${{ secrets.POETRY_CACHE_VERSION }}
+
+    - name: Clear Poetry cache
+      if: steps.cache-poetry.outputs.cache-hit == 'true' && needs.changes.outputs.backend == 'true' && contains(github.event.pull_request.labels.*.name, 'tools:clear-poetry-cache-unit-tests')
+      run: rm -r .venv
+
+      # Poetry >= 1.1.0b uses virtualenv to create a virtual environment.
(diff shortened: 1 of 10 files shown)


.github/workflows/continous-integration.yml
```diff
@@ -276,7 +276,107 @@ jobs:
...
+        make test-non-training
+
+    - name: Send Coverage Report 📊
+      if: needs.changes.outputs.backend == 'true' && matrix.python-version == 3.6 && matrix.os != 'windows-latest'
+      env:
+        COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        COVERALLS_SERVICE_NAME: github
+      run: poetry run coveralls
+
+  test_training:
```
joejuzl, line 288: Is there an easy way to reuse these duplicate steps?

tczekajlo: No, you can consider using GH action of the composite type but I wouldn't say it's an easy way

https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action

.github/workflows/continous-integration.yml
```diff
@@ -276,7 +276,107 @@ jobs:
...
+    name: Run Tests That Train
+    runs-on: ${{ matrix.os }}
+    needs: [changes]
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        python-version: [3.6, 3.7, 3.8]
```
tczekajlo, line 295: I think we would add here something like this

```suggestion
        python-version: [3.6, 3.7, 3.8]
        test: [test-non-training, test-training]
```

.github/workflows/continous-integration.yml
```diff
@@ -276,7 +276,107 @@ jobs:
...
+        make install-full
+        make prepare-tests-windows
+
+    - name: Test Code 🔍
+      if: needs.changes.outputs.backend == 'true' && !(matrix.python-version == 3.8 && matrix.os == 'windows-latest')
+      env:
+        JOBS: 2
+        PYTHONIOENCODING: "utf-8"
+      run: |
+        make test-training
```
tczekajlo, line 379:
```suggestion
        make ${{ matrix.test }}
```

and then we don't have to copy the entire job

tczekajlo (commented):
I've left some suggestions :D

.github/workflows/continous-integration.yml
joejuzl: nice 👍

rasa/shared/utils/common.py
```diff
@@ -205,3 +206,13 @@ def arguments_of(func: Callable) -> List[Text]:
     import inspect
 
     return list(inspect.signature(func).parameters.keys())
+
+
+def raise_on_unexpected_train(f):
```
wochinge, line 211: Let's please add a docstring which points them to the docs - also worth updating the #product-engineering channel about it

wochinge, line 211:
```suggestion
def raise_on_unexpected_train(f: Callable) -> Any:
```

rasa/core/agent.py
```diff
@@ -707,6 +708,7 @@ async def load_data(
             exclusion_percentage=exclusion_percentage,
         )
 
+    @rasa.shared.utils.common.raise_on_unexpected_train
```
wochinge, line 711: what about doing a session wide fixture which monkeypatches this for tests? That way we don't have to have this in our production code.

wochinge (commented):
Nice work 💯

rasa/core/agent.py
joejuzl: I was thinking maybe it would be useful in other scenarios, but you are right - best keep it out of production code.

rasa/core/agent.py
```diff
@@ -50,6 +50,7 @@
 from rasa.shared.nlu.training_data.training_data import TrainingData
 from rasa.utils.endpoints import EndpointConfig
 import rasa.utils.io
+import rasa.shared.utils.common
```
wochinge, line 53: Can this import be removed? (same for other modules)

rasa/shared/utils/common.py
```diff
@@ -3,6 +3,7 @@
 import importlib
 import inspect
 import logging
+import os
```
wochinge, line 6:
```suggestion
```

tests/conftest.py
```diff
@@ -452,3 +453,31 @@ class AsyncMock(Mock):
 
     async def __call__(self, *args: Any, **kwargs: Any) -> Any:
         return super().__call__(*args, **kwargs)
+
+
+def raise_on_unexpected_train(f) -> Callable:
```
wochinge, line 458:
```suggestion
def raise_on_unexpected_train(f: Callable) -> Callable:
```

tests/conftest.py
```diff
@@ -452,3 +453,31 @@ class AsyncMock(Mock):
...
+    @functools.wraps(f)
+    def decorated(*args, **kwargs):
+        if os.environ.get("RAISE_ON_TRAIN") == "True":
+            raise ValueError("Training called and RAISE_ON_TRAIN is set.")
```
wochinge, line 462: How about adding a link to the readme section and / or telling developers how to fix this?

tests/conftest.py
```diff
@@ -452,3 +453,31 @@ class AsyncMock(Mock):
...
+
+        See "Tests that train" section in rasa/README.md.
+    """
+    import rasa.nlu as nlu
+    import rasa.core as core
+    from rasa.nlu.model import Trainer
+    from rasa.core.agent import Agent
+
+    for training_module in [nlu, core, Trainer, Agent]:
+        training_module.train = raise_on_unexpected_train(training_module.train)
```
wochinge, line 479: how convenient that all methods are called `train` :-D

wochinge (approved):
💯

Issue #7789, opened 2021-01-23
Support user messages from transports other than HTTP(S)


veigaribo:
First of all I'd like to say that this is my first kind of contribution to the project, so it's quite possible for me to have not understood something.

**Description of Problem**:

I would like for my Rasa bot to be able to receive user messages from AMQP, through a messaging system, and not from HTTP(S) with webhooks. But, currently, from what I've checked, that is basically impossible.

My suggestion would be a way to create Custom Connectors that work without relying on Sanic routes.

**Overview of the Solution**:

Actually, it's almost possible to do such a thing now. The only thing we would need to do would be to give the `InputChannel`'s `blueprint` method access to the Sanic `app` object, so that it could register a listener for when the server is ready, and even get access to the event loop that way (that is necessary because, if one tries to call the `handler` / `on_new_message` callback before the Sanic server has been initialized, an error could be raised, since the `app` object will possibly have `None` as the `agent` attribute. Also access to the server's event loop is important to enable proper scheduling with the messages coming from the new non-HTTP(S) channel).

The problem I see is that, currently, the `InputChannel` is way too tied with HTTP(S). The function that creates them is even called `create_http_input_channels`.

So, for the end goal, what I have in mind is:

*   We would probably need to rename the `create_http_input_channels` function, I don't see why it wouldn't be used for other types of channels;
*   We would need to give to either the current `InputChannel` objects or to objects of a newly created class the opportunity to run code after the `app.agent` has been defined (i.e. right after server initialization), while still giving access to the `handler` / `on_new_message` callback. That is what is lacking in the current `InputChannel` implementation, as it assumes the callback will only be called once the server has started, which, in our new case, would not be necessarily true (messages could be received before the server has started).

Another alternative, I suppose, would be to stop using the `agent` from the Sanic `app` object, and instead get it from some other way, so that the callback that is given to the `blueprint` method may be called before the `app.agent` has been set. I haven't looked very much into that possibility, though.

**Examples**:

I have implemented this example: [https://github.com/Veigaribo/example-rasa-amqp-connector/blob/master/connectors/amqp.py](https://github.com/Veigaribo/example-rasa-amqp-connector/blob/master/connectors/amqp.py)

That would work if the Sanic `app` object was passed as an argument to the `blueprint` method.

By the way, about that repository, be warned that cloning its submodules will currently not work. That shouldn't be necessary to experiment, though.

**Definition of Done**:

*   [ ] Tests are added
[...]

PR #7924, opened 2021-02-10
Support transports other than HTTP(S)

Veigaribo:
This implements a solution for the issue #7789.

What I did was the second idea I proposed, where basically we create an intermediary method `setup` (which I referred to as `register` in the issue) to actually register the Channel blueprint to the app. (I give my reasons for it below, but, please, bring it up if you feel like another approach would be better)

What that allows us to do is basically access both the Sanic app and the `on_new_message` callback, which makes it possible for us to register a listener on the app and use the `on_new_message` only when it is ready.

Suppose we have an InputChannel that is supposed to receive it's messages using AMQP:

```python
class AmqpInput(InputChannel):
    # ...

    # override setup
    def setup(self, on_new_message, app, options):
        # we may still register the blueprint for health-check, etc
        super().setup(on_new_message, app, options)

        # defer actual setup for when the server has started
        app.register_listener(
            # when this lambda executes, we may be sure on_new_message is ready to be called
            # so we may start consuming immediately
            lambda _, loop: self._setup_connection(on_new_message, loop),
            "after_server_start",
        )
``` 

Also, I created a class (that could very well be a _dataclass_) to hold extra options, just in case we need to add more. Then we may do it in a backwards-compatible way.

---

The reasons why I chose this approach over the other one (creating a `on_agent_ready` method to run specifically when the server has started) are:
1. It just seems much easier to implement
2. I imagine it will be easier to use this solution to solve #7525 (more below)
3. Basically, compared to the other approach, this seems more future-proof, in the sense that it gives more power to the end-user to do what they want. 

---

Regarding the issue #7525 (and PR #7526), the improvement I imagine this would enable would be to do something like:

```python
class SlackInput(InputChannel):
    # ...

    # override setup
    def setup(self, on_new_message, app, options):
        # do not register the blueprint at first

        # by deferring the setup to just before the server starts, we are able to get access to the event loop
        app.register_listener(
            # then we may give the would-be-blueprint method access to it
            lambda _, loop: self.blueprint_with_loop(on_new_message, loop),
            "before_server_start",
        )

    def blueprint_with_loop(on_new_message, loop):
        # ...

        loop.create_task(on_new_message(user_msg))
```

---

## Questions

- Are there any tests I need to change / add? From what I looked, the `tests/core/test_channels.py`,  `tests/core/test_run.py` and the ones in `tests/core/channels` passed. Does that mean it's ok?

**Status**:

### DIFF
--- a/rasa/core/channels/channel.py
+++ b/rasa/core/channels/channel.py
@@ -78,6 +78,13 @@ class UserMessage:
         self.metadata = metadata
 
 
+class InputChannelSetupOptions:
+    """Parameters used to setup InputChannels"""
+
+    def __init__(self, url_prefix: Optional[Text] = None):
+        self.url_prefix = url_prefix
+
+
 def register(
     input_channels: List["InputChannel"], app: Sanic, route: Optional[Text]
 ) -> None:
@@ -89,7 +96,9 @@ def register(
             p = urljoin(route, channel.url_prefix())
         else:
             p = None
-        app.blueprint(channel.blueprint(handler), url_prefix=p)
+
+        options = InputChannelSetupOptions(p)
+        channel.setup(handler, app, options)
 
     app.input_channels = input_channels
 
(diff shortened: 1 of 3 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @alopez will take a look at it as soon as possible ✨

Veigaribo:
I'm working on another approach at #8031, I suppose only one of the pull requests should be accepted.

PR #7929, opened 2021-02-10
fix regex pattern to match all new micro versions

m-vdb:
**Proposed changes**:
- Fix regex pattern in doc release script to match all micros. Until now it matched all the micros not containing a "0" digit (`2.2.1`, `2.2.7`, `2.2.12`), missing micros like `2.2.10` or `2.2.20`

### DIFF
--- a/scripts/push_docs_to_branch.sh
+++ b/scripts/push_docs_to_branch.sh
@@ -5,7 +5,7 @@ set -Eeuo pipefail
 TODAY=`date "+%Y%m%d"`
 # we build new versions only for minors and majors
 PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.0$"
-PATTERN_FOR_PATCH_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9]+$"
+PATTERN_FOR_PATCH_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9]+[0-9]*$"
 MASTER_REF=refs/heads/master
 VARIABLES_JSON=docs/docs/variables.json
 SOURCES_FILES=docs/docs/sources/
--- a/scripts/push_docs_to_branch.sh
+++ b/scripts/push_docs_to_branch.sh
@@ -15,7 +15,7 @@ set -Eeuo pipefail
 TODAY=`date "+%Y%m%d"`
 # we build new versions only for minors and majors
 PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.0$"
-PATTERN_FOR_MICRO_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9]+[0-9]*$"
+PATTERN_FOR_MICRO_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9][0-9]*$"
 MASTER_REF=refs/heads/master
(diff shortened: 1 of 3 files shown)


scripts/push_docs_to_branch.sh
```diff
@@ -1,11 +1,22 @@
...
+# In case this script fails in our CI workflows, you can run it locally.
+# For instance, if the doc version for 2.2.10 fails to be pushed, you
+# can do:
+#
+# TMP_DOCS_FOLDER=/tmp/documentation DOCS_BRANCH=documentation GITHUB_REF=refs/tags/2.2.10 ./scripts/push_docs_to_branch.sh
+#
+# This command will push the changes to the `documentation` branch of this repository.
+#
+# Please make sure to have the proper dependencies installed with `make install install-docs`
+# and that the docs are pre-built with `make prepare-docs`.
```
m-vdb, line 12: once this is reviewed I'll update the other branches on other repos

scripts/push_docs_to_branch.sh
```diff
@@ -1,11 +1,22 @@
...
+
 set -Eeuo pipefail
 
 TODAY=`date "+%Y%m%d"`
 # we build new versions only for minors and majors
 PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.0$"
-PATTERN_FOR_PATCH_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9]+$"
+PATTERN_FOR_MICRO_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9]+[0-9]*$"
```
wochinge, line 19: Seems a bit clearer to me ("no 0 in the beginning, then all numbers are allowed")
Should we also check the "no 0 in the beginning" for majors / minors?
```suggestion
PATTERN_FOR_MICRO_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9][0-9]*$"
```

scripts/push_docs_to_branch.sh
```diff
@@ -1,11 +1,22 @@
...
 
+# In case this script fails in our CI workflows, you can run it locally.
+# For instance, if the doc version for 2.2.10 fails to be pushed, you
+# can do:
+#
+# TMP_DOCS_FOLDER=/tmp/documentation DOCS_BRANCH=documentation GITHUB_REF=refs/tags/2.2.10 ./scripts/push_docs_to_branch.sh
+#
+# This command will push the changes to the `documentation` branch of this repository.
+#
+# Please make sure to have the proper dependencies installed with `make install install-docs`
```
wochinge, line 11: Should we make this a numerated list to be more concise? E.g.

1. Run `make install install-docs`
2. Run `make prepare-docs`
3. Run `TMP_DOCS...`

scripts/push_docs_to_branch.sh
m-vdb: yeah that's true, we do not need the `+` 👍🏻 . And yes, I think it's more accurate even though it doesn't change the result.

m-vdb: actually no, for minors and majors it's accurate. We want to be able to match `0.0.12`, `2.0.22`, etc...

scripts/push_docs_to_branch.sh
m-vdb: yes, that's a great suggestion

wochinge (approved):

Issue #261, opened 2017-04-04
Extracting Number Entities


james-aiwip:
**rasa NLU version** (e.g. `0.7.3`):

0.7.3

**Used backend / pipeline** (`mitie`, `spacy_sklearn`, ...):

spacy_sklearn

**Operating system** (windows, osx, ...):

osx

**Issue**:

Is it possible to extract number entities ?

tmbo:
Sure thing, you can use the `fine_tune_spacy_ner ` setting to resuse existing spay entities. There is an entity for cardinals (https://spacy.io/docs/usage/entity-recognition#entity-types) which should capture numbers.

james-aiwip:
@tmbo is there an example I can use for the API for this ?

tmbo:
Before training your model you need to add this key to your configuration e.g.:
```json
{
  "backend": "spacy_sklearn",
  "path" : "./models",
  "data" : "./data/examples/rasa/demo-rasa.json",
  "fine_tune_spacy_ner": true
}
```

After training your model you can use the server to shoot queries at it. and it should return your custom entities as well as the pretrained spacy entities.

PR #7931, opened 2021-02-10
Adding a "global" not_intent parameter for slot mappings

Imod7:
**Description**:
This PR can potentially close [#261](https://github.com/RasaHQ/rasa-sdk/issues/261) reported in the rasa-sdk repo. The fix is done in rasa-oss since forms are moved in rasa-oss now.

### Proposed changes:
The changes described below are related to the concept of adding a new feature in [forms](https://rasa.com/docs/rasa/forms/) and more specifically of adding a global parameter in a form that affects/updates the `not_intent` parameter of all the slots present in that form. This change meant the addition of the parameter `global_not_intent` under forms and the addition of the keyword `required_slots` before the slot mappings definition. This resulted also to the deprecation of the lack of the keyword `required_slots` as described in more detail below.

**Changes in code / functionality**
- Added an optional `global_not_intent` parameter under the form's name (in `domain.yml` file) with the format suggested by @akelad [here](https://github.com/RasaHQ/rasa/pull/7931#issuecomment-785808059).
- When the `global_not_intent` parameter is set (in the `domain.yml` file) then all slots parameter `not_intent` are affected/updated on the fly / internally with the value of the `global_not_intent` parameter during the form's initialisation. However, this update that happens internally **does not** change the `domain.yml` file. 
Example
if in the domain.yml file we have the following form example : 
```
forms:
  restaurant_form:
    global_not_intent: [chitchat, inform]
    required_slots:
      cuisine:
      - type: from_entity
        entity: cuisine
```
this will NOT be converted to this
```
forms:
  restaurant_form:
    global_not_intent: [chitchat, inform]
    required_slots:
      cuisine:
      - type: from_entity
        entity: cuisine
        not_intent: [chitchat, inform]
```
but internally yes, the global parameter will affect/update the slot's `not_intent` parameter so the behaviour should be also affected accordingly.

- When the `global_not_intent` keyword is present then also the `required_slots` keyword needs to be added before the `slot_mappings` to denote the start of their definition.
- Added a test function with different parameter cases that checks if the `global_not_intent` parameter affects the individual slot's parameter `not_intent` correctly.

**`required_slots` keyword update**
- Added the keyword `required_slots` in all forms examples found in tests/examples/docs as described [here](https://github.com/RasaHQ/rasa/pull/7931/files#r601229968)

**Exceptions / Warnings**
- Added a test that checks if an InvalidException is raised when `global_not_intent` is present but `required_slots` keyword is missing. Since this results to an empty domain.
- Added a deprecation warning for when the keyword `required_slots` is not used to define slot mappings.

- [x] added two tests (one for the functionality and one to check if exception is raised)
- [x] added a deprecation log (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [ ] added a migration guide item


wochinge:
I'm a bit busy atm - will request Joe instead, ok?

Imod7:
I still need to do quite a few changes so I am not really ready for review at the moment. I just opened the PR to make a start and implement the changes in a more transparent way (by pushing them gradually here).

Imod7:
Some questions after the last changes that I did are the following : 
- During `rasa train` it looks like it picks up from the `domain.yml` the `global_not_intent` variable and its value but then when I run `rasa shell` I don't have it and I don't know why. Maybe I miss a step in the updates that I do during `rasa train`?
- In `forms.py` I did some changes to update the `not_intent` value of a slot with the global param's value and it looks like it updates it correctly when I run and debug. Not sure how to have a more formal way to verify it like a test case. Any ideas?

rasa/core/actions/forms.py
```diff
@@ -64,6 +64,7 @@ def from_entity(
         not_intent: Optional[Union[Text, List[Text]]] = None,
         role: Optional[Text] = None,
         group: Optional[Text] = None,
+        global_not_intent: Optional[Text] = None,
```
joejuzl, line 67: After looking through the code I don't think this method is ever called with `intent` or `not_intent`. I think they must be legacy. It seems like `intent_is_desired` is the method you actually have to look at. 
Also, another option to passing the `global_not_intent` around, is to just update the `not_intent` list for each  `slot_mapping` in the domain when it's initialized.

Imod7:
@joejuzl I guess I will also need to add a test for this so I am looking how to implement one by checking some of these [tests](https://github.com/RasaHQ/rasa/blob/f4af88f0c165c8898b6924b0cbdb0984b8117e63/tests/core/actions/test_forms.py#L744).

joejuzl:
@Imod7 I would start by adding a test to `tests/shared/core/test_domain.py`. Look at `test_valid_slot_mappings` for inspiration. You can test either the contents of `Domain.forms` or the output of `slot_mapping_for_form`. 
It could also be cool to have a test in `rasa/test_forms.py` like you suggested, to test the full functionality.

Imod7:
@joejuzl Hello! Sorry to bother you again but your feedback is super helpful. I have some after thoughts
- I should have the global_not_intent under forms not under sessionconfig
- Any ideas on how to make the code where I update the slots not_intent more …. Elegant? 
- I added the `test_global_not_intent_slot_mappings` in the test_domain and you were right to point me there because it is covering the functionality and it is much easier to make the test. Thank you for this. In the parametric decorator I added different test cases just to be sure that it works correctly. Maybe it is too much? Testing the global_not_intent (last test case) I only have one example, maybe on this one I should add more.

rasa/shared/core/domain.py
```diff
@@ -619,6 +627,41 @@ def _initialize_forms(
...
+            # Updating the `not_intent` parameter of each slot with `global_not_intent`
+            if session_config.global_not_intent is not None:
+                global_params = session_config.global_not_intent
+                if not isinstance(global_params, list):
+                    global_params = [global_params]
+                for global_value in global_params:
+                    # looping through all the available forms, their slots and their
+                    # all their slot's types and updating the `not_intent` key
+                    # with the `global_not_intent` value.
+                    for form_count, _ in enumerate(list(forms.keys())):
```
joejuzl, line 639: rather than using enumerate with the keys why not try iterating over `.items()` this will give you immediate access to both the key and the value.

rasa/shared/core/domain.py
```diff
@@ -619,6 +627,41 @@ def _initialize_forms(
...
+                                            global_value
+                                            not in slot_params[type_count][key]
+                                        ):
+                                            slot_params[type_count][key].append(
+                                                global_value
+                                            )
+                                    else:
+                                        if global_value != slot_params[type_count][key]:
+                                            slot_params[type_count][key] = [
+                                                slot_params[type_count][key],
```
joejuzl, line 660: you reference `lot_params[type_count][key]` many times, so may be worth extracting into a variable.

tests/shared/core/test_domain.py
```diff
@@ -1320,3 +1322,118 @@ def test_is_valid_domain_doesnt_raise_with_invalid_yaml(tmpdir: Path):
         potential_domain_path,
     )
     assert not Domain.is_domain_file(potential_domain_path)
+
+
+@pytest.mark.parametrize(
+    "domain_as_dict, expected_not_intent_value",
```
joejuzl, line 1328: since only `"not_intent"` and `GLOBAL_NOT_INTENT` change, you could just paramatize those values, making the whole thing much more succinct.

joejuzl (changes requested):
I think it's fine to put the `global_not_intent` in the session config. I think the forms section can only contain references to actual forms, not config.

Imod7:
Hello @akelad ! Could you please share the process/steps to follow to update the documentation for a change like this one? Also, let me know if I need to ping someone else and not you for this. Thank you so much in advance! 🙏🏽

Imod7:
Hello @akelad ! I made some draft changes in the domain & forms docs. Could you please take a look and let me know if I should adjust something? Also, please share your thoughts on the implemented functionality in case I need to change something on that too. Thank you so much for your time!

akelad:
Hmmm I just started looking at this, and i'm kind of skeptical that `not_intent` is configured on the session config level and not in the forms. What if you e.g. want an intent to fill slots in one form but not the other? I think we may have gone too global :P

Imod7:
Good point! :) So it is better to define it in the level of forms but the question is where to put it. In the `domain.yml` file I haven't seen anything else being put under the `form_name` except the slots so it might be tricky to add a config parameter there (or maybe not? 🤔). 
Does it make sense to keep it in the `session_config` but define it like this :
```
session_config:
  session_expiration_time: 60  # value in minutes
  carry_over_slots_to_new_session: true
  first_form:
    global_not_intent: chitchat
  second_form:
    global_not_intent: nlu_fallback
```
or maybe I can try use the `config` key in the `domain.yml` file and add something like this :
```
config:
     store_entities_as_slots: false
     first_form:
        global_not_intent: chitchat
     second_form:
        global_not_intent: nlu_fallback
```
or put it under each `form_name` like this : 
```
forms:
  first_form:
    global_not_intent: chitchat
    some_slot:
    ....
  second_form:
    global_not_intent: nlu_fallback
    other_slot:
    ....
```
but then how to differentiate the slot from the config parameter. Also, this might be confusing from a user perspective because the `global_not_intent` looks like a slot.
Do you think any of these solutions (or none is also an option :) ) is acceptable ? Thank you so much for the very good feedback!

akelad:
I guess you could do something more verbose for forms like:

```
forms:
  restaurant_form:
    global_not_intent: intent
    required_slots:
      cuisine:
        - type: from_entity
          entity: cuisine
      num_people:
        - type: from_entity
          entity: number
```

might be worth getting some input from other people as well before deciding on this

Imod7:
Thank you so much Akela for the useful feedback! 
@wochinge @joejuzl @alwx Maybe we can discuss this (_the best approach to have a `global_not_intent` in forms level_) tomorrow during the refinement? Any suggestions on who else to include to decide on this is more than welcome.

wochinge:
I prefer Akela's suggestion.

docs/docs/forms.mdx
```diff
@@ -32,13 +32,42 @@ The following example form `restaurant_form` will fill the slot
...
+            entity: number
+```
+
+There is also an optional parameter that can be defined in each form which is called
+`global_not_intent` parameter. It can be added under the form_name and its value will
+affect all slots that are present in that form. More specifically, it will update/add
+the `not_intent` key of all individual slots with the specified value (or list of values)
+found in the global parameter.
+
+For example, if you want to fill the `not_intent` key of all slots with `intent` then
```
joejuzl, line 51: I think this wording is a little confusing. I would prefer to focus on the fact that it means non of the slots will be filled with the intent `intent` (maybe rename too), rather than the fact it "fills" the key "nont_intent" for each slot.

docs/docs/forms.mdx
```diff
@@ -32,13 +32,42 @@ The following example form `restaurant_form` will fill the slot
...
+    required_slots:
+        cuisine:
+          - type: from_entity
+            entity: cuisine
+        num_people:
+          - type: from_entity
+            entity: number
 ```
+Note that if you decide to use the `global_not_intent` parameter then the keyword
+`required_slots` should precede the definition of slot mappings as shown in the above
```
joejuzl, line 68: I would argue that we always support `required_slots` even if there is no `global_not_intent`. Maybe we also think about deprecating the lack of it? @akelad thoughts?

rasa/shared/core/domain.py
```diff
@@ -619,6 +621,39 @@ def _initialize_forms(
             `FormAction` which is implemented in the Rasa SDK.
         """
         if isinstance(forms, dict):
+            # looping through all the available forms and checking if a
```
joejuzl, line 624: This is getting quite big. I would extract a method which can handle this for a single form, and then iteratively apply it to each form.

rasa/shared/core/domain.py
```diff
@@ -1683,10 +1718,25 @@ def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
     if not isinstance(forms, dict):
         raise InvalidDomain("Forms have to be specified as dictionary.")
 
-    for form_name, slots in forms.items():
-        if slots is None:
+    for form_name, form_data in forms.items():
+        if form_data is None:
             continue
 
+        if hasattr(form_data, "__iter__"):
```
joejuzl, line 1725: what is this checking?

joejuzl (changes requested):

docs/docs/forms.mdx
akelad: I'm pretty sure that leaving required_slots empty is allowed if you have custom slot mappings? See point 2 here: https://rasa.com/docs/rasa/forms#custom-slot-mappings

joejuzl (approved):
Great job! 🚀

docs/docs/forms.mdx
wochinge: > I'm pretty sure that leaving required_slots empty is allowed if you have custom slot mappings?

True

> Maybe we also think about deprecating the lack of it

I would do this in case there are non custom slot mappings to push users to annotate them within `required_slots` and so that we can drop the extra code in 3.0

Imod7 (commented):
Submitting review

docs/docs/forms.mdx
wochinge: I think the current implementation is fine, **but** I'd implement a deprecation warning for the case that users don't use `required_slots` to define slot mappings. I would always raise this warning independent if they use `global_not_intent` or not.

Means this causes a deprecation warning (but continues to work)

```yaml
forms:
  my_form:
    my_slot:
   - type: from_text
```

And these don't:
```yaml
forms:
  my_form:
    required_slots:
      my_slot:
     - type: from_text
```

```yaml
forms:
  my_form:
    global_not_intent:
    - intent1
    - intent2
    required_slots:
      my_slot:
     - type: from_text
```

> and in the function required_slots we define the ones with custom_mappings.

This function is in the SDK and doesn't need to be considered here, **but** it raises an important point:

* from what I understand `slot_mapping_for_form` breaks with this change as it won't return the slot mappings but rather `required_slots` 
* This in turns requires another change in the SDK [here](https://github.com/RasaHQ/rasa-sdk/blob/7f9d7a43a48c9fbec170453c8459c70648cfb8a8/rasa_sdk/forms.py#L820)

rasa/shared/core/domain.py
```diff
@@ -598,9 +600,43 @@ def _collect_overridden_default_intents(
         }
         return sorted(intent_names & set(rasa.shared.core.constants.DEFAULT_INTENTS))
 
-    @staticmethod
+    @classmethod
+    def _update_slots_with_global_not_intent_parameter(
```
wochinge, line 604: I think this method should leave the domain in a consistent state. Currently the slot mappings can either be in `required_slots` or at the level above. For the sake of other components relying on this we should in my opinion make sure that the domain always moves the slots mappings beyind `required_slots`.

rasa/shared/core/domain.py
```diff
@@ -598,9 +600,43 @@ def _collect_overridden_default_intents(
...
+                for slot_key, slot_val in form_data.items():
+                    for slot in slot_val:
+                        key = "not_intent"
+                        # check that `not_intent` param is present in the slot
+                        if key in slot.keys():
+                            # check that the value of `not_intent` is a list
+                            if isinstance(slot[key], list):
+                                # check that `global_not_intent` is in the list
+                                if global_val not in slot[key]:
+                                    slot[key].append(global_val)
```
wochinge, line 629: Now I know what the downside is of handling this in the domain instead of on the fly in the `FormAction` 😬 When we dump the domain then this will currently alter the domain, e.g.:

```yaml
forms:
  my_form:
    global_not_intent: [greet]
    required_slots:
      name:
      - type: from_text
```

would become
```yaml
forms:
  my_form:
    global_not_intent: [greet]
    required_slots:
      name:
      - type: from_text
        not_intent: [greet]
```

For the Rasa Open Source world this isn't a big deal, but for Integrated Version Control in Rasa X this will lead to big diffs and users might be confused by that. In my opinion we either need to revert the domain as done [here](https://github.com/RasaHQ/rasa/blob/b9705f455d05cc619c222d16970f2b412fae3af1/rasa/shared/core/domain.py#L1326) or we add the global ignored intents in the `FormAction` which seems a tad simpler to me.

rasa/shared/core/domain.py
```diff
@@ -1683,10 +1725,39 @@ def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
...
+            elif "global_not_intent" in form_data and "required_slots" not in form_data:
+                raise InvalidDomain(
+                    f"If you use the `global_not_intent` parameter in your form, then"
+                    f" the keyword `required_slots` should precede the definition"
+                    f" of your slot mappings. Please see"
+                    f" {rasa.shared.constants.DOCS_URL_FORMS} for more information."
+                )
+            else:
+                slots = form_data
+        else:
```
wochinge, line 1744: I think we still need to treat the `list` type as valid: https://github.com/RasaHQ/rasa/blob/b9705f455d05cc619c222d16970f2b412fae3af1/rasa/shared/core/domain.py#L635

docs/docs/forms.mdx
```diff
@@ -40,6 +40,32 @@ forms:
         entity: number
 ```
 
+There is also an optional parameter that can be defined in each form which is called
```
wochinge, line 43: aren't there other examples in the docs which we'll have to adapt to the new format as well?

changelog/261.improvement.md
```diff
@@ -0,0 +1,5 @@
+Added an optional `global_not_intent` parameter in forms.
```
wochinge, line 1: we also need a deprecation changelog and a migration guide item.

wochinge (changes requested):
Sorry for all the comments 😬 This turns out to be quite a complex issue.

docs/docs/forms.mdx
Imod7: @wochinge So, I need to update all the examples that include `forms` : 
- in the docs, e.g. [from_entity](https://rasa.com/docs/rasa/forms#from_entity)
- maybe also in the [examples](https://github.com/RasaHQ/rasa/blob/dcaa9210326d41c7569fe5a0478cbc0d2f8ffb27/examples/formbot/domain.yml#L92) folder 
- and in rasa/data folder ? (eg. [default_with_slots.yml](https://github.com/RasaHQ/rasa/blob/dcaa9210326d41c7569fe5a0478cbc0d2f8ffb27/data/test_domains/default_with_slots.yml#L36))
- and wherever I find `forms` related examples

and add the keyword `required_slots` before the definition of slot mappings ?

rasa/shared/core/domain.py
Imod7: @wochinge I agree with the `FormAction` solution! 👌🏾 So, I will revert the changes done in the `domain` and I will start implementing it in the `FormAction`.

wochinge: Sorry, for not having thought of this earlier. We need to keep the `required_slots` stuff in the domain. The "only" thing which would need moving to the `FormAction` is the handling of the `global_not_intent` itself (so you'd remove the part from the domain which inserts the `global_not_intent` into each slot mapping)

wochinge: Slick!

Imod7: I pushed some of the changes just to double check with you that I am on the right track. If you could check the changes in the `domain` and `forms` and give me a feedback if that is what you meant, that would be great. 🙏🏽 However in the current implementation I see some issues, for example :
- In the function `get_mappings_for_slot ` I go through all the forms but I should only update (with global_param) the form where the specific slot is (since `get_mappings_for_slot` addresses only to one slot (`slot_to_fill`)). But how do I know in which form I am ? 🧐
- The `get_mappings_for_slot ` looks like it updates only the `from_entity` type in the slot? But I should update the `not_intent` of all the types right (with `global_not_intent`)?

rasa/shared/core/domain.py
Imod7: So I will remove the warning but if the form is a list does this affect how I retrieve the `slots` or how I check the other parameters (`global_not_intent` and `required_slots`) ?

rasa/shared/core/domain.py
```diff
@@ -598,9 +600,43 @@ def _collect_overridden_default_intents(
...
+            global_params = form_data.get("global_not_intent")
+            # accessing the form slots by using keyword `required_slots`
+            if "required_slots" in form_data:
+                form_data = form_data.get("required_slots")
+            # checking if the global parameter is a list or not and always
+            # making it a list.
+            if not isinstance(global_params, list):
+                global_params = [global_params]
+            # looping through the form's global params.
+            for global_val in global_params:
```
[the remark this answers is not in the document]
Imod7: That is very true! 💯 I changed it and it is now located in the `FormAction`.

docs/docs/forms.mdx
Imod7: - So, I need to check if there are other places in the docs where forms are mentioned and check if I should mention also the `global_not_intent` parameter ? 
- Also, do I have to add the `required_slots` keyword : 
   - in all the `form` related examples in the docs ?
   - Also, in the [examples](https://github.com/RasaHQ/rasa/tree/e2759589e308de1120364035256d6169776c8c55/examples) folder ?
   - Maybe also in the examples in the tests, eg. [here](https://github.com/RasaHQ/rasa/blob/e2759589e308de1120364035256d6169776c8c55/tests/core/actions/test_forms.py#L76) ?

wochinge: I think all docs examples should use the `required_slots` key from now onwards. The old syntax should still work but give deprecation warnings. And yep, that would require changes in all of the mentioned locations 💯

rasa/shared/core/domain.py
wochinge: You can safely remove it 👍🏻 Any forms defined as lists will have no mapped slots anyway. `_initialize_forms` takes care of this

rasa/shared/core/domain.py
wochinge: Are you sure that you've pushed all changes? I Can't see any within `forms`?

rasa/core/actions/forms.py
```diff
@@ -55,7 +55,14 @@ def required_slots(self, domain: Domain) -> List[Text]:
         Returns:
             A list of slot names.
         """
-        return list(domain.slot_mapping_for_form(self.name()).keys())
+        form = domain.slot_mapping_for_form(self.name())
```
wochinge, line 58: I would handle the `required_slots` stuff within the domain and consolidate it there.
I'd imagine the workflow like this:

1. `Domain` is loaded from YAML
2. `Domain.__init__` double checks form format. In case the users didn't use `required_slots` it adjusts it so that the dictionary looks like `{"forms": {"my_form": {"required_slots": {"my_slot": [] }}}}`

rasa/core/actions/forms.py
```diff
@@ -88,17 +94,51 @@ def from_entity(
...
     def get_mappings_for_slot(
         self, slot_to_fill: Text, domain: Domain
     ) -> List[Dict[Text, Any]]:
         """Get mappings for requested slot.
 
         If None, map requested slot to an entity with the same name
         """
-
         requested_slot_mappings = self._to_list(
             domain.slot_mapping_for_form(self.name()).get(
```
wochinge, line 140: I think we can simplify the changes a bit:
1. Adapt `required_slots` as you've done already (no need to consider a missing `required_slot` key as this should've been consolidated by the `Domain` beforehand)
2. Add a new method `get_global_not_intents` which retrieves a domain and returns a list of potentially ignored intents from the domain
3. Make `intent_is_desired` a object method by adding `self` and removing the `staticmethod` decorator
4. make the domain a param of `intent_is_desired`
4. Concatenate `mapping_not_intents` with `self.get_global_not_intents(domain)`

Does that make sense?

tests/core/actions/test_forms.py
```diff
@@ -1334,3 +1358,73 @@ async def test_ask_for_slot_if_not_utter_ask(
...
+        ),
+        (
+            # `global_not_intent` as a list of 2 values and slot's not_intent has one
+            # value that is included also in `global_not_intent`.
+            ["chitchat", "greet"],
+            ["chitchat"],
+        ),
+    ],
+)
+async def test_global_not_intent(
```
wochinge, line 1390: I think we don't need `async` here, do we? 
```suggestion
def test_global_not_intent(
```

wochinge, line 1390: can we please also add tests for the other slot mapping types?

rasa/shared/core/domain.py
```diff
@@ -1774,7 +1763,7 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
         Returns:
             The slot mapping or an empty dictionary in case no mapping was found.
         """
-        return self.forms.get(form_name, {})
+        return self.forms.get(form_name, {})["required_slots"]
```
wochinge, line 1766: How about making `required_slots` a constant? https://deviq.com/antipatterns/magic-strings

rasa/shared/core/domain.py
```diff
@@ -1849,10 +1838,32 @@ def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
     if not isinstance(forms, dict):
         raise InvalidDomain("Forms have to be specified as dictionary.")
 
-    for form_name, slots in forms.items():
-        if slots is None:
+    for form_name, form_data in forms.items():
+        if form_data is None:
             continue
 
+        if isinstance(form_data, Dict):
```
wochinge, line 1845: What else can it be than a `Dict`? Why do we have to check this?

rasa/shared/core/domain.py
```diff
@@ -1849,10 +1838,32 @@ def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
...
+            if "required_slots" in form_data:
+                slots = forms[form_name].get("required_slots")
+            elif "global_not_intent" in form_data and "required_slots" not in form_data:
+                raise InvalidDomain(
+                    f"If you use the `global_not_intent` parameter in your form, then"
+                    f" the keyword `required_slots` should precede the definition"
+                    f" of your slot mappings. Please see"
+                    f" {rasa.shared.constants.DOCS_URL_FORMS} for more information."
+                )
```
wochinge, line 1854: do we need this? Can't we just always warn if `required_slots` is not used?

rasa/shared/core/domain.py
```diff
@@ -1849,10 +1838,32 @@ def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
...
+        if "required_slots" not in form_data:
+            rasa.shared.utils.io.raise_deprecation_warning(
+                f"The definition of slot mappings in your form"
+                f" should always be preceded by the keyword `required_slots`."
+                f" The lack of this keyword will be deprecated in"
+                f" Rasa Open Source 3.0.0. Please see "
+                f"{rasa.shared.constants.DOCS_URL_FORMS} for more information.",
+            )
+
         if not isinstance(slots, Dict):
```
wochinge, line 1867: pycharm gives me a warning that `slots` might be referenced before assignment. We should fix this.

rasa/shared/core/domain.py
```diff
@@ -628,24 +629,12 @@ def _initialize_forms(
             `FormAction` which is implemented in the Rasa SDK.
         """
         if isinstance(forms, dict):
+            for form_name, form_data in forms.items():
```
wochinge, line 632: do we have a test for this behavior?

rasa/core/actions/forms.py
```diff
@@ -165,16 +163,43 @@ def _entity_mapping_is_unique(
...
+        """Returns a list of ignored intents as retrieved from `global_not_intent`.
+
+        Args:
+            domain: The current model domain.
+
+        Returns:
+            The value/s found in `global_not_intent` parameter in the `domain.yml`
+            (under forms).
+        """
+        if "global_not_intent" in domain.__dict__["forms"][self.name()]:
```
wochinge, line 176: Why not this? Accessing stuff via `__dict__` is very unsafe and should only be done in exceptional cases.
```suggestion
        if "global_not_intent" in domain.forms[self.name()]:
```

wochinge, line 176: optional: In my opinion this violates the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). I find it slightly preferable if we ask the domain to give us the `global_not_intent` for the form (e.g. `domain.global_not_intent_for(self.name()`)

rasa/core/actions/forms.py
```diff
@@ -165,16 +163,43 @@ def _entity_mapping_is_unique(
...
+        tracker: "DialogueStateTracker",
+        domain: Domain,
     ) -> bool:
         """Check whether user intent matches intent conditions."""
         mapping_intents = FormAction._to_list(requested_slot_mapping.get("intent", []))
         mapping_not_intents = FormAction._to_list(
             requested_slot_mapping.get("not_intent", [])
         )
 
+        mapping_not_intents = list(
```
wochinge, line 199: what do we need the `list(...)` call for?

examples/formbot/domain.yml
```diff
@@ -89,34 +89,36 @@ actions:
...
+        not_intent: affirm
+      feedback:
+      - type: from_entity
+        entity: feedback
+      - type: from_text
 
 session_config:
   session_expiration_time: 60  # value in minutes
-  carry_over_slots_to_new_session: true
+  carry_over_slots_to_new_session: true
```
wochinge, line 124:
```suggestion
  carry_over_slots_to_new_session: true

```

changelog/261.improvement.md
wochinge: can you please also create an entry in the migration guide for 2.5? 🙌🏻

changelog/261.improvement.md
```diff
@@ -0,0 +1,5 @@
...
+
+- In order to use it, please add in your `domain.yml` file the `global_not_intent` parameter with a value/list of values of your choice after the form's name.
```
wochinge, line 3: I'd link to the documentation section as well

docs/docs/forms.mdx
```diff
@@ -32,14 +32,41 @@ The following example form `restaurant_form` will fill the slot
...
+    required_slots:
+        cuisine:
+          - type: from_entity
+            entity: cuisine
+        num_people:
+          - type: from_entity
+            entity: number
+```
+
+Note that if you use the `global_not_intent` parameter then the keyword `required_slots`
```
wochinge, line 67: They should always use `required_slots`, no?

wochinge (commented):
Looks really good! Thanks for adapting all the examples! 

@federicotdn Heads up that we're changing the form syntax here. From what I know Rasa X doesn't look at the form data in detail but maybe I'm missing sth here.

changelog/261.removal.md
```diff
@@ -0,0 +1 @@
+In forms, the keyword `required_slots` should always precede the definition of slot mappings and the lack of it will be deprecated.
```
wochinge, line 1: Can we please link to the docs for this?
```suggestion
In forms, the keyword `required_slots` should always precede the definition of slot mappings and the lack of it is deprecated.
```

docs/docs/forms.mdx
```diff
@@ -32,12 +32,36 @@ The following example form `restaurant_form` will fill the slot
...
+            entity: cuisine
+        num_people:
+          - type: from_entity
+            entity: number
+```
+
+There is also an optional parameter that can be defined in each form which is called
+`global_not_intent` parameter. It can be added under the form_name and its value will
+affect all slots that are present in that form. More specifically, it will update/add
+the `not_intent` key of all individual slots with the specified value (or list of values)
```
wochinge, line 47: @akelad Can you or somebody from your team please review this one as well? 🙌🏻

docs/docs/forms.mdx
```diff
@@ -32,12 +32,36 @@ The following example form `restaurant_form` will fill the slot
...
+found in the global parameter.
+
+For example, if you do not want any of the slots of the specific form to be filled with
+intent `chitchat` then you would need to define the following (after the form_name and
```
wochinge, line 51:
```suggestion
intent `chitchat` then you would need to define the following (after the form name and
```

docs/docs/forms.mdx
```diff
@@ -32,12 +32,36 @@ The following example form `restaurant_form` will fill the slot
...
+before the definition of the slot mappings):
+
+```yaml-rasa
+forms:
+  restaurant_form:
+    global_not_intent: chitchat
```
wochinge, line 57: does it make sense to call this `ignored_intents`? It's not consistent with the `not_intent` naming but seems more intuitive to me

docs/docs/migration-guide.mdx
```diff
@@ -55,6 +55,15 @@ We've updated our docs to reflect these changes. All examples now show a direct
 correct spaCy model. We've also added a warning to the [SpaCyNLP](components.mdx#spacynlp)
 docs that explains the fallback behavior.
 
+### Forms
+
+#### New `global_not_intent` parameter in Forms
+
+We added a new parameter under Forms with the name `global_not_intent`. This parameter
+can be used in case the user never wants to fill any slots of a form with the specified
+intent, eg. chitchat. Please see https://rasa.com/docs/rasa/forms for examples and more
```
wochinge, line 64: please use relative links (ideally also link to the specific section in the forms docs page)
```suggestion
intent, eg. chitchat. Please see the [Forms documentation](forms.mdx) for examples and more
```

wochinge, line 64: I'd vote for showing an example of the old format and then show how to transition that to the new format

rasa/core/actions/forms.py
```diff
@@ -165,16 +166,46 @@ def _entity_mapping_is_unique(
...
+        """Returns a list of ignored intents as retrieved from `global_not_intent`.
+
+        Args:
+            domain: The current model domain.
+
+        Returns:
+            The value/s found in `global_not_intent` parameter in the `domain.yml`
+            (under forms).
+        """
+        if domain.__dict__["forms"]:
```
wochinge, line 179: why are we accessing the forms attribute via `__dict__`? Can't we just use `domain.forms`? Even cleaner would be something like `domain.get_global_not_intent(self.name()` to avoid breaking the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter)

rasa/shared/core/domain.py
```diff
@@ -20,7 +20,19 @@
     Iterable,
 )
 
-import rasa.shared.constants
+from rasa.shared.constants import (
```
wochinge, line 23: why this change?

rasa/shared/core/domain.py
wochinge: I can't find the tests for it - how is it called?

rasa/shared/core/domain.py
```diff
@@ -1810,7 +1804,10 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
         Returns:
             The slot mapping or an empty dictionary in case no mapping was found.
         """
-        return self.forms.get(form_name, {})
+        if self.forms.get(form_name, {}):
+            return self.forms.get(form_name, {})[REQUIRED_SLOTS_KEY]
+        else:
+            return {}
```
wochinge, line 1810: why do we need this?

Imod7:
@wochinge I checked the CI tests that fail but I can't figure out the errors. Could you take a look? Thank you!!!

wochinge:
@Imod7 Are you referring to `Code Quality` or `Run Tests ...`?

changelog/261.improvement.md
```diff
@@ -0,0 +1,5 @@
+Added an optional `ignored_intents` parameter in forms.
+
+- In order to use it, please add in your `domain.yml` file the `ignored_intents` parameter with a value/list of values of your choice after the form's name. Please see https://rasa.com/docs/rasa/forms for more information.
```
wochinge, line 3: let's use relativ links please
```suggestion
- In order to use it, please add in your `domain.yml` file the `ignored_intents` parameter with a value/list of values of your choice after the form's name. Please see [Forms](forms.mdx) for more information.
```

changelog/261.improvement.md
```diff
@@ -0,0 +1,5 @@
...
+- This parameter will then update/add the `not_intent` key of all individual slots of a form with its specified value (or list of values).
```
wochinge, line 4: Can we explain that in a more user-facing way and less coupled to the implementation? E.g.
```suggestion
- all slot mappings of a form won't apply if the intent of a message is specific in `ignored_intents`
```

changelog/261.removal.md
```diff
@@ -0,0 +1,2 @@
+In forms, the keyword `required_slots` should always precede the definition of slot mappings and the lack of it is deprecated.
+Please see https://rasa.com/docs/rasa/forms for more information.
```
wochinge, line 2:
```suggestion
Please see the [migration guide](migration-guide.mdx) for more information.
```

docs/docs/forms.mdx
wochinge: @melindaloubser1 maybe?

docs/docs/migration-guide.mdx
```diff
@@ -55,6 +55,15 @@ We've updated our docs to reflect these changes. All examples now show a direct
 correct spaCy model. We've also added a warning to the [SpaCyNLP](components.mdx#spacynlp)
 docs that explains the fallback behavior.
 
+### Forms
```
wochinge, line 58: that needs to go in a different section now, doesn't it? 2.6 to 2.5

examples/formbot/domain.yml
```diff
@@ -89,33 +89,35 @@ actions:
...
-    - type: from_intent
-      intent: deny
-      value: no additional preferences
-    - type: from_text
-      not_intent: affirm
-    feedback:
-    - type: from_entity
-      entity: feedback
-    - type: from_text
+    ignored_intents: [chitchat, inform]
```
wochinge, line 92: ok to this for `chitchat`, but I think `inform` breaks most of the bot. Also if you ignore `chitchat` then you don't need to ignore it again in line 97

rasa/core/actions/forms.py
```diff
@@ -165,16 +166,44 @@ def _entity_mapping_is_unique(
         mapping_as_string = json.dumps(slot_mapping, sort_keys=True)
         return mapping_as_string in self._unique_entity_mappings
 
-    @staticmethod
+    def get_ignored_intents(self, domain: Domain) -> List[Text]:
+        """Returns a list of ignored intents as retrieved from `ignored_intents`.
```
wochinge, line 170: what does the `as retrieved from ignored_intents` part mean?

rasa/core/actions/forms.py
```diff
@@ -165,16 +166,44 @@ def _entity_mapping_is_unique(
...
+
+        Args:
+            domain: The current model domain.
+
+        Returns:
+            The value/s found in `ignored_intents` parameter in the `domain.yml`
+            (under forms).
+        """
+        if domain.forms:
```
wochinge, line 179: why do we need the `if domain.forms` check?

rasa/shared/core/domain.py
```diff
@@ -85,17 +97,24 @@ class ActionNotFoundException(ValueError, RasaException):
...
-            rasa.shared.constants.DEFAULT_CARRY_OVER_SLOTS_TO_NEW_SESSION,
+            DEFAULT_SESSION_EXPIRATION_TIME_IN_MINUTES,
+            DEFAULT_CARRY_OVER_SLOTS_TO_NEW_SESSION,
         )
 
     def are_sessions_enabled(self) -> bool:
+        """Returns a boolean value depending on the value of session_expiration_time.
+
+        Returns True if the value of the session_expiration_time variable is greater
+        than zero and False if session_expiration_time is less than/equal to zero."""
```
wochinge, line 117: I think the code is clear enough and more concise than the extra docstring
```suggestion
"""
```

rasa/shared/core/domain.py
```diff
@@ -144,21 +163,19 @@ def from_path(cls, path: Union[Text, Path]) -> "Domain":
 
     @classmethod
     def from_file(cls, path: Text) -> "Domain":
+        """Given a specific filepath (path), returns the domain in a yaml format."""
```
wochinge, line 166:
```suggestion
        """Loads the `Domain` from a YAML file."""
```

rasa/shared/core/domain.py
wochinge: @Imod7

rasa/shared/core/domain.py
wochinge: @Imod7

wochinge (commented):
@Imod7 Could you please resolve all addressed comments and come back to the open ones?

tests/core/actions/test_forms.py
```diff
@@ -1334,3 +1359,218 @@ async def test_ask_for_slot_if_not_utter_ask(
...
+                                    "entity": entity_name,
+                                    "not_intent": slot_not_intent,
+                                }
+                            ],
+                        },
+                    }
+                }
+            }
+        )
+    elif type == "from_text":
```
wochinge, line 1505: why don't we make the form input paramatrizable (the dictionary which is currently mapped to `form_name`? The `if type == ...` make the test hard to read and to understand. An alternative could also be to split the test

rasa/core/actions/forms.py
Imod7: Otherwise, this test `pytest tests/core/actions/test_forms.py::test_extract_requested_slot_default` fails because `domain.forms` is an empty dict `{}`.

tests/core/actions/test_forms.py
Imod7: That is so True! Very good point Tobias!!!! 👌🏾 I am going to split it.

rasa/shared/core/domain.py
Imod7: The same as above. If I do not have the check, the same test `pytest tests/core/actions/test_forms.py::test_extract_requested_slot_default` fails because `domain.forms` is an empty dict `{}`.

Imod7 (commented):
> @Imod7 Are you referring to `Code Quality` or `Run Tests ...`?

@wochinge 
I am referring to `Run Tests` like [here](https://github.com/RasaHQ/rasa/pull/7931/checks?check_run_id=2388821010) where for example the following tests fail : 

```
FAILED tests/core/test_actions.py::test_get_form_action_with_rasa_open_source_1_forms
FAILED tests/core/test_ensemble.py::test_mapping_wins_over_form[events0] - ra...
FAILED tests/core/test_ensemble.py::test_form_wins_over_everything_else[ensemble0]
FAILED tests/core/test_agent.py::test_form_without_form_policy[policy_config0]
FAILED tests/core/test_ensemble.py::test_form_wins_over_everything_else[ensemble1]
FAILED tests/shared/core/test_domain.py::test_merge_domain_with_forms - Asser...
FAILED tests/shared/core/test_domain.py::test_check_domain_sanity_on_invalid_domain
```

rasa/shared/core/domain.py
```diff
@@ -661,28 +672,16 @@ def _initialize_forms(
             `FormAction` which is implemented in the Rasa SDK.
         """
         if isinstance(forms, dict):
```
Imod7, line 674: @wochinge 
In reply to [this](https://github.com/RasaHQ/rasa/pull/7931#discussion_r615950111)

If I do not check if `forms` is a `dict` then this test `pytest tests/shared/core/test_domain.py::test_valid_slot_mappings` fails for the same reason as above, because `forms` is an empty dict, `{}` in this test.

docs/docs/forms.mdx
melindaloubser1: sure

docs/docs/forms.mdx
```diff
@@ -32,12 +32,36 @@ The following example form `restaurant_form` will fill the slot
...
+        num_people:
+          - type: from_entity
+            entity: number
+```
+
+There is also an optional parameter that can be defined in each form which is called
+`ignored_intents` parameter. It can be added under the form name and its value will
+affect all slots that are present in that form. More specifically, it will update/add
+the `not_intent` key of all individual slots with the specified value (or list of values)
+found in the `ignored_intents` parameter.
```
melindaloubser1, line 48:
```suggestion
You can define a list of intents to ignore for the whole form under the 
`ignored_intents` key. Intents listed under `ignored_intents` will be added to the `not_intent` key of each required slot in the form.
```

docs/docs/forms.mdx
```diff
@@ -32,12 +32,36 @@ The following example form `restaurant_form` will fill the slot
...
+
+For example, if you do not want any of the slots of the specific form to be filled with
+intent `chitchat` then you would need to define the following (after the form name and
+before the definition of the slot mappings):
```
melindaloubser1, line 52:
```suggestion
For example, if you do not want any of the required slots of a form to be filled when the intent is `chitchat`, then you would need to define the following (after the form name and
before the definition of the slot mappings):
```
Does the order matter? Why can't `ignored_intents` come after `required_slots`?

docs/docs/forms.mdx
```diff
@@ -32,12 +32,36 @@ The following example form `restaurant_form` will fill the slot
...
+
+```yaml-rasa
+forms:
+  restaurant_form:
+    ignored_intents: chitchat
```
melindaloubser1, line 57:
```suggestion
    ignored_intents: 
    - chitchat
```
If I understand this right this should be a list?

docs/docs/migration-guide.mdx
```diff
@@ -55,6 +55,15 @@ We've updated our docs to reflect these changes. All examples now show a direct
...
+
+#### New `ignored_intents` parameter in Forms
+
+We added a new parameter under Forms with the name `ignored_intents`. This parameter
+can be used in case the user never wants to fill any slots of a form with the specified
+intent, eg. chitchat. Please see the [Forms documentation](forms.mdx) for examples and more
+information on how to use it in your `domain.yml` file.
+
```
melindaloubser1, line 66:
```suggestion
#### New `ignored_intents` parameter in Forms

There is a new parameter under Forms called `ignored_intents`. This parameter
can be used to prevent any required slots in a form from being filled with the specified
intent or intents. Please see the [Forms documentation](forms.mdx) for examples and more
information on how to use it in your `domain.yml` file.

```

melindaloubser1 (changes requested):
Just some wording changes and one question about the format

docs/docs/migration-guide.mdx
Imod7: Moved it! The description is ok?

rasa/shared/core/domain.py
wochinge: The other changes are out of scope for this PR. let's try to keep the diff small.

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,17 @@ description: |
...
 how you can migrate from one version to another.
 
+## Rasa 2.5 to 2.6
+
+### Forms
+
+#### New `ignored_intents` parameter in Forms
+
+We added a new parameter under Forms with the name `ignored_intents`. This parameter
+can be used in case the user never wants to fill any slots of a form with the specified
```
wochinge, line 20: can we please show them an example which illustrates what they have to change exactly (like a "before" and "after" version of the domain)?

rasa/core/actions/forms.py
wochinge: In that case the test `test_extract_requested_slot_default` needs adaption as we shouldn't use an empty domain but actually one which specifies the slot mappings for the form.

rasa/core/actions/forms.py
```diff
@@ -165,16 +166,44 @@ def _entity_mapping_is_unique(
...
+            if IGNORED_INTENTS in domain.forms[self.name()]:
+                ignored_intents = domain.forms[self.name()][IGNORED_INTENTS]
+            else:
+                ignored_intents = []
```
wochinge, line 183: seems simpler, no?
```suggestion
            ignored_intents = domain.forms[self.name()].get(IGNORED_INTENTS, [])
```

rasa/shared/core/domain.py
wochinge: > What behavior do we want to test here?

We should test that the Domain is read correctly if users have a (deprecated) domain like this:

```yaml

forms:
  my_form:
    my_slot:
    - type: from_text
```

rasa/shared/core/domain.py
wochinge: let's fix the test instead as described in the other comment 👍🏻

rasa/core/actions/forms.py
wochinge: It's fine 👍🏻

docs/docs/forms.mdx
wochinge: @melindaloubser1 Is that also cool for you?

rasa/shared/core/domain.py
wochinge: do we have a test that the lack of `required_slots` warns?

examples/formbot/domain.yml
```diff
@@ -89,33 +89,34 @@ actions:
...
-    - type: from_intent
-      intent: deny
-      value: no additional preferences
-    - type: from_text
-      not_intent: affirm
-    feedback:
-    - type: from_entity
-      entity: feedback
-    - type: from_text
+    ignored_intents: chitchat
```
wochinge, line 92: optional: I'd vote to push users towards always using a list 
```suggestion
    ignored_intents: [chitchat]
```

docs/docs/forms.mdx
melindaloubser1: yes, I see it's already been changed elsewhere

wochinge:
> I am referring to Run Tests like here where for example the following tests fail :

What do the logs of tests say? Do they also fail when running these tests locally?

docs/docs/forms.mdx
Imod7: Yes

Imod7 (commented):
> > I am referring to Run Tests like here where for example the following tests fail :
> 
> What do the logs of tests say? Do they also fail when running these tests locally?

There are different errors. Yes they fail also locally. I will check again.

rasa/shared/core/domain.py
Imod7: This behaviour is checked with this test `test_form_with_no_required_slots_keyword`.

rasa/shared/core/domain.py
Imod7: I have a question on that and it is mentioned in the other comment.

tests/core/actions/test_forms.py
```diff
@@ -1334,3 +1359,267 @@ async def test_ask_for_slot_if_not_utter_ask(
...
+            ),
+            ActionExecuted(ACTION_LISTEN_NAME),
+        ],
+    )
+
+    slot_values = form.extract_other_slots(tracker, domain)
+    assert slot_values == {}
+
+
+def test_form_with_no_required_slots_keyword():
```
wochinge, line 1611: can we please move this test to `test_domain`? It doesn't execute any code within forms but purely tests implementation from the `domain` module. 
Can you please also test that this gives a `FutureWarning`?

rasa/core/actions/forms.py
wochinge: > It is actually a domain that has an empty form like shown below :

You can see that in the implementation of `Domain.empty` [here](https://github.com/RasaHQ/rasa/blob/3e7212c1f6e95f518e2749c576676559b1dae756/rasa/shared/core/domain.py#L124)

> We should not keep the test case of empty form since there might be a user who puts an empty form in a domain?

if `domain.forms` is empty then there is also no way that we'd call this form in the first place. 

> as we see an empty form

Note that `if domain.forms` doesn't test if a single form is empty but only whether there are **any** forms given in the domain.

rasa/shared/core/domain.py
Imod7: Adjusted the test by giving a domain with valid slot mappings for the form.

changelog/261.improvement.md
```diff
@@ -0,0 +1,5 @@
+Added an optional `ignored_intents` parameter in forms.
+
+- In order to use it, please add in your `domain.yml` file the `ignored_intents` parameter with a value/list of values of your choice after the form's name. Please see [Forms](forms.mdx) for more information.
+- All slot mappings of a form won't apply if the intent of a message is specific in `ignored_intents`.
+- This can be used in case the user never wants to fill any slots of a form with the specified intent, eg. chitchat.
```
melindaloubser1, line 5:
```suggestion
- To use it, add the `ignored_intents` parameter  in your `domain.yml` file after the forms name and provide a list of intents to ignore. Please see [Forms](forms.mdx) for more information.
- This can be used in case the user never wants to fill any slots of a form with the specified intent, eg. chitchat. 
```

melindaloubser1: Since it's more completely described in the docs, shortened the changelog a bit

docs/docs/forms.mdx
```diff
@@ -32,12 +32,34 @@ The following example form `restaurant_form` will fill the slot
...
+        cuisine:
+          - type: from_entity
+            entity: cuisine
+        num_people:
+          - type: from_entity
+            entity: number
+```
+
+You can define a list of intents to ignore for the whole form under the 
+`ignored_intents` key. Intents listed under `ignored_intents` will be added to the `not_intent` key of each required slot in the form.
```
melindaloubser1, line 45:
```suggestion
`ignored_intents` key. Intents listed under `ignored_intents` will be added to the `not_intent` key of each slot mapping in the form.
```

This is my bad, I suggested `required slot` in the first place, but on re-reading it it actually applies per mapping, not just per slot

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,62 @@ description: |
...
+intent, eg. chitchat. Please see the [Forms documentation](forms.mdx) for examples and
```
melindaloubser1, line 21:
```suggestion
intent, e.g. chitchat. Please see the [Forms documentation](forms.mdx) for examples and
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,62 @@ description: |
...
+more information on how to use it in your `domain.yml` file.
+
+So before, if a user did not want to fill any slots of a form with a specified intent
```
melindaloubser1, line 24:
```suggestion
Before, if a user did not want to fill any slots of a form with a specified intent
```

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,62 @@ description: |
...
+we would have to define it (with the `not_intent` parameter) in every slot
+as shown in the following example :
```
melindaloubser1, line 26:
```suggestion
they would have to define it under the `not_intent` parameter for every slot mapping
as shown in the following example :
```

melindaloubser1 (changes requested):
Looking good! Couple more wording changes suggested.

docs/docs/forms.mdx
Imod7: Thank YOU for re reading it!!! 💯

melindaloubser1 (approved):
🚀

rasa/shared/core/domain.py
```diff
@@ -1875,23 +1890,51 @@ def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
...
+            )
+            slots = form_data
+
+        if IGNORED_INTENTS in form_data and REQUIRED_SLOTS_KEY not in form_data:
+            raise InvalidDomain(
+                f"If you use the {IGNORED_INTENTS} parameter in your form, then"
+                f" the keyword {REQUIRED_SLOTS_KEY} should precede the definition"
+                f" of your slot mappings. Please see"
+                f" {DOCS_URL_FORMS} for more information."
+            )
```
wochinge, line 1931: I think this check should go before the one in line 1913 because if we actually raise an exception then there is no reason to also bother the user with a warning.

wochinge: That's a screenshot which shows both warnings.
Screenshot 2021-04-22 at 14 04 18

wochinge:
A few tiny comments - but otherwise ready-to-go! (except Melinda has more docs feedback) 🚀 
Also tested the change manually and it works like a charm 💪🏻 

@joejuzl Could you please remove the `change requested` as @Imod7 implemented all the changes?

wochinge:
@federicotdn Did you have a chance to have a look? https://github.com/RasaHQ/rasa/pull/7931#pullrequestreview-626378601

rasa/shared/core/domain.py
Imod7: True!

tests/core/actions/test_forms.py
```diff
@@ -1334,3 +1378,250 @@ async def test_ask_for_slot_if_not_utter_ask(
...
+            ["chitchat", "greet"],
+            ["chitchat"],
+            "from_trigger_intent",
+        ),
+    ],
+)
+def test_ignored_intents_with_other_type_of_slots(
+    ignored_intents: Union[Text, List[Text]],
+    slot_not_intent: Union[Text, List[Text]],
+    type: Text,
```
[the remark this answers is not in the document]
Imod7: True! 😱

rasa/shared/core/domain.py
```diff
@@ -1875,23 +1890,51 @@ def _validate_slot_mappings(forms: Union[Dict, List]) -> None:
...
+                f"as '{type(form_data)}'. They need to be specified "
+                f"as dictionary. Please see {DOCS_URL_FORMS} "
+                f"for more information."
+            )
+
+        if REQUIRED_SLOTS_KEY in form_data:
+            slots = forms[form_name].get(REQUIRED_SLOTS_KEY)
+        else:
+            rasa.shared.utils.io.raise_deprecation_warning(
+                f"The definition of slot mappings in your form"
```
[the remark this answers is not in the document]
Imod7: Omg!!!! You noticed!!!!🤩 Ok, that's it. You are my absolute HERO!!! My.Hero.Forever.And.Ever.The.End. I thought about the spaces at one point (start or end...) but then I had other stuff to correct so I did not pay much attention and then I forgot. I will put everywhere the same. 🙂👌🏾

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,62 @@ description: |
...
+      feedback:
+      - type: from_entity
+        entity: feedback
+        not_intent: chitchat
+```
+
+By introducing the `ignored_intents` parameter, we now only need to define it
+in one place and it will affect all the slots of the form :
+
+```yaml
```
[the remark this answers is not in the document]
Imod7: That is awesome! I did not know we could do that!

federicotdn:
@wochinge, sorry, not yet. I can do it tomorrow.

docs/docs/migration-guide.mdx
```diff
@@ -10,6 +10,62 @@ description: |
...
+more information on how to use it in your `domain.yml` file.
+
+Before, if a user did not want to fill any slots of a form with a specified intent
+they would have to define it under the `not_intent` parameter for every slot mapping
+as shown in the following example :
+
+```yaml
+forms:
+  restaurant_form:
+    required_slots:
```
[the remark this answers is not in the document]
Imod7: Yes, I thought about this but I added the `required_slots` in all examples, tests, docs, everywhere so I thought it would be weird to only have this place that `required_slots` keyword is missing. But ok, I removed it.

melindaloubser1 (approved):

wochinge (approved):
Great and patient work on this complex feature! Changes which evolve deprecations, docs changes and also implicate other code bases such as the SDK contain a lot of complexity. Despite the amount of feedback, a change in approach and an underestimation of the issue from me, you showed perseverance and kept a clear head! 💯🥇

wochinge:
(lets quickly wait for Fede's ok before merging it)

federicotdn:
I believe this does not change anything on the Rasa X side. We are currently storing forms as `DomainAction` entities, and the entire JSON content of each individual form is stored inside a `data` column. The contents of this `data` do not matter to Rasa X, and they are included (of course) when a domain needs to be serialized back into file form.

Obviously we should test it anyways (once it's released), but even if we don't, it should make no difference!

PR #7936, opened 2021-02-11
Small code-tick fix. 

koaning:
Part of our [docs](https://rasa.com/docs/rasa/migrate-from/google-dialogflow-to-rasa/) don't render nicely because of a few backticks missing.

This PR fixes that. 

**What it currently looks like**

image


tmbo (approved):
looks good (just saw this and jumped on it to fix it for today's release)

PR #7949, opened 2021-02-12
Avoid validating slot of previous form

ArjaanBuijk:
**Proposed changes**:
Avoid making a validation request to the action server for a slot that does not belong to the active form.

The update of https://github.com/RasaHQ/rasa/pull/7764 enforces a slot validation when returning to a form after an `unhappy path`. However, it is possible that the  `unhappy path` results in another form to be activated, while the REQUESTED_SLOT is still set to a slot of the previous form. This then results in a validation request to the action server for a slot that does not belong to the active form.

This fix sets the REQUESTED_SLOT in the temporary tracker before sending it to the action server.


rasa/core/actions/forms.py
```diff
@@ -437,7 +437,9 @@ def _temporary_tracker(
             current_tracker.events_after_latest_restart()
             # Insert form execution event so that it's clearly distinguishable which
             # events were newly added.
-            + [ActionExecuted(self.name())] + additional_events,
+            + [SlotSet(REQUESTED_SLOT, self.get_slot_to_fill(current_tracker))]
```
Ghostvv, line 440: could you please add the comment here why we reset `requested_slot`

rasa/core/actions/forms.py
```diff
@@ -437,7 +437,9 @@ def _temporary_tracker(
...
+            + [ActionExecuted(self.name())]
```
Ghostvv, line 441:
```suggestion
            # Insert form execution event so that it's clearly distinguishable which
            # events were newly added.
            + [ActionExecuted(self.name())]
```

rasa/core/actions/forms.py
```diff
@@ -437,7 +437,9 @@ def _temporary_tracker(
             current_tracker.events_after_latest_restart()
             # Insert form execution event so that it's clearly distinguishable which
             # events were newly added.
```
Ghostvv, line 439:
```suggestion
```
I think it is better to move this comment after reseting requested slot

Ghostvv (approved):
looks good to me. Could you please add a comment in the code

Ghostvv:
@ArjaanBuijk the tests are failing

ArjaanBuijk:
@Ghostvv , I will fix the failing tests.

PR #7961, opened 2021-02-15
script to update path of sources loaded with remark-source to a versioned path

m-vdb:
**Proposed changes**:

When we update the `documentation` branch, we might be creating or editing the documentation for a specific minor version (e.g. `2.3.x`). Some of the doc files (`.mdx` files) contain codeblocks that are supposed to load files contained in the codebase thanks to `remark-source` (+ a script that copies the sources under the `docs/` folder during the `make prepare-docs` phase). Such instructions look like this:
```
```python (docs/sources/rasa/core/channels/rest.py)
```
What's problematic is that this will always load the file from the `main` version of the docs, not the version-specific doc, and **this is a bug**. One way of fixing it is to replace `docs/sources/` by `versioned_docs/version-2.3.x/sources/` (for version `2.3.x` in this case) for each version of the files.

Alternatively, I considered updating `remark-sources` plugin (with a fork), but this plugin doesn't have knowledge of the underlying file structure, only the content it's being passed for processing. So we're stuck with a custom step in our script.

I've run the changes I made in this PR in another branch created from `documentation` and pushed [this commit](https://github.com/RasaHQ/rasa/commit/9f0560ba2e98b0bb86ebcb561cce94a13a1a3983) that shows the result. After this branch is merged into `documentation`, I'll run the script directly on the branch and commit the changes.

💡 no need to update Rasa X it isn't using the `remark-source` plugin. We'll need to update Rasa SDK though.


scripts/push_docs_to_branch.sh
```diff
@@ -70,6 +70,12 @@ else
     fi
 fi
 
+CURRENTLY_EDITING_VERSION=${EXISTING_VERSION:-$NEW_VERSION}
+if [ ! -z "$CURRENTLY_EDITING_VERSION" ]
```
wochinge, line 74:
```suggestion
if [ -n "$CURRENTLY_EDITING_VERSION" ]
```

docs/plugins/included_source.js
```diff
@@ -69,11 +72,49 @@ async function getIncludedSources(options) {
...
-    // finally, write all the source files in the `sourceDir`
+    // finally, write all the source files in the source directory
     return await Promise.all(sourceFiles.map(async ([sourceFile, content]) => {
-        return await outputFile(`${sourceDir}/${sourceFile}`, content);
+        return await outputFile(`${docsDir}/${relativeSourceDir}/${sourceFile}`, content);
     }));
 };
 
 
+/**
```
wochinge, line 82: Can we please add a short sentence about what this is doing?

wochinge (commented):
I don't understand everything, but seemse to make sense. Thanks for the explanation and the example commit 🙌🏻 Should maybe Lu have another look?

docs/plugins/included_source.js
m-vdb: ah yeah, totally forgot about this, thanks for the heads up

lunelson (commented):
Looks good. It's too bad there isn't a way to have the remark plugin be aware of the path of the file being transformed 🤷🏻 so they could use relative paths, or a way to specify a reference directory so they could just refer to `source/..` and it would resolve; it worries me a little bit to bind the content of all these docs to a specific file structure but it seems like the best way to go 👍🏻

lunelson (approved):

PR #7965, opened 2021-02-16
validate model config

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7893
- validate model configurations with a schema file
- validate config files (`endpoints.yml`, `credentials.yml`) using a schema instead of a Python check
- move schema files which are used by `rasa.shared` code but aren't in `rasa.shared` to `rasa.shared`
- make `language` not a required key after double checking with  Ty


rasa/shared/utils/schemas/model_config.yml
```diff
@@ -0,0 +1,33 @@
+allowempty: True
+mapping:
+  version:
+    type: "str"
+    required: False
+    allowempty: False
+  language:
+    type: "str"
+    required: True
```
wochinge, line 9: double check if this is really required

rasa/shared/utils/io.py
```diff
@@ -518,26 +521,56 @@ def raise_deprecation_warning(
...
-        )
+
+def read_model_configuration(filename: Union[Path, Text]) -> Dict[Text, Any]:
+    """Parses a model configuration file.
+
+    Args:
+        filename: The path to the file which should be read.
+
+    Raises:
+        YamlValidationException: In case the model configuration doesn't match the
```
wochinge, line 567: that's a different one than before but in my opinion the correct one. Double checked the code and nothing depends on the exact exception type.

rasa/shared/utils/schemas/model_config.yml
wochinge: `language` [should not be required](https://rasa-hq.slack.com/archives/C36SS4N8M/p1613480592285800)

alwx (approved):

PR #7944, opened 2021-02-11
profile training data loading

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7892

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #7968, opened 2021-02-16
YAML speed up

### LINKED ISSUE #7952: speed up `is_key_in_yaml`
**Problem**
We currently use the [function `is_key_in_yaml`](https://github.com/RasaHQ/rasa/blob/954e4734d479924df7211cd3bc5503285d54a66b/rasa/shared/core/training_data/story_reader/yaml_story_reader.py#L167) to check if a file is a nlu or story file. This reads and parses the entire yaml file just to see if a key is present. 

This makes the training data reading [slow](https://github.com/RasaHQ/rasa/issues/7892#issuecomment-778302866) because:
* we read the entire file although this information might be right in the beginning
* we parse the entire file although we don't need to

**Proposed Solution**:
* Something like this: https://github.com/RasaHQ/rasa/pull/7944/files#r575331363

### LINKED ISSUE #7953: speed up yaml parsing
**Problem**
When parsing yaml files we currently look for environment variables to interpolate in every single line. While this is a useful feature for configuration files (`endpoints.yml`, `config.yml`, `credentials.yml`) and e.g. also used for Rasa X, this makes the training data reading very slow as stated [here](https://github.com/RasaHQ/rasa/issues/7892#issuecomment-778302866).

Note that removing this would be a breaking change as some users might rely on it. We need some sort of smooth transition for it.

Note that I've not investigated how much [this line](https://github.com/RasaHQ/rasa-x/blob/f5b89e565161f80307c524138df32553a63e80fc/src/community/rasax/community/utils/yaml.py#L17) also contributes to the slowness

**Ideas for solutions**:
* disable environment variable interpolation for stories / nlu files by default 
* add a switch to disable it 

**Definition of Done**
- [ ] Environment variable interpolation no longer is a blocking factor when reading big training data files
- [ ] investigate if [this line](https://github.com/RasaHQ/rasa-x/blob/f5b89e565161f80307c524138df32553a63e80fc/src/community/rasax/community/utils/yaml.py#L17) is still needed / what it's impact is on training data loading time
- [ ] document usage / migration path for environment variable interpolation

alwx:
**Proposed changes**:
- fix #7952 
- fix #7953


rasa/shared/utils/io.py
```diff
@@ -318,25 +319,99 @@ def env_var_constructor(loader, node):
...
         content: A text containing yaml content.
         reader_type: Reader type to use. By default "safe" will be used
+        replace_env_vars: Specifies if environment variables need to be replaced
 
     Raises:
         ruamel.yaml.parser.ParserError: If there was an error when parsing the YAML.
     """
-    fix_yaml_loader()
-
-    replace_environment_variables()
```
alwx, line 333: In fact, the performance drop was also caused by this call being made every time we tried to read YAML. Multiple `SafeConstructor`s and implicit resolvers were being added, and every attempt to read YAML was making a situation even worse and making the whole process of env interpolation even slower.

wochinge:
@alwx Can you please request from @joejuzl ? Looking at the time today and my calandar tomorrow I won't get to it before my vacation.

wochinge:
@alwx How long does loading multiwoz with this take? You can re-use the test from my investigation PR https://github.com/RasaHQ/rasa/pull/7944

joejuzl (changes requested):

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -175,11 +176,18 @@ def is_key_in_yaml(cls, file_path: Union[Text, Path], *keys: Text) -> bool:
...
-            YamlException: if the file seems to be a YAML file (extension) but
-                can not be read / parsed.
+            FileNotFoundException: if the file cannot be found.
         """
-        content = rasa.shared.utils.io.read_yaml_file(file_path)
-        return any(key in content for key in keys)
+        try:
+            with open(file_path) as file:
+                return any(
+                    all(line.startswith(f"{key}:") for key in keys) for line in file
```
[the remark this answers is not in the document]
alwx: That's correct. Fixed it and added one more test for this.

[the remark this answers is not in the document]
alwx: You're right — it would not be found in this case, but I would prefer to keep it that way because this is how it worked before. This function is a part of `rasa.shared` and it means it's better to not change the behaviour.

However, I updated the docstring.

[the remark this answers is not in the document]
joejuzl: ok I see - it's top level keys only.

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -175,11 +176,18 @@ def is_key_in_yaml(cls, file_path: Union[Text, Path], *keys: Text) -> bool:
...
-            YamlException: if the file seems to be a YAML file (extension) but
-                can not be read / parsed.
+            FileNotFoundException: if the file cannot be found.
         """
-        content = rasa.shared.utils.io.read_yaml_file(file_path)
-        return any(key in content for key in keys)
+        try:
+            with open(file_path) as file:
+                return all(
+                    any(line.startswith(f"{key}:") for line in file) for key in keys
```
joejuzl, line 184: Optional: would be faster to do one pass of the lines and only check remaining keys.
Also if we change the behaviour back to `any`, then it would be much faster.

tests/shared/core/training_data/story_reader/test_yaml_story_reader.py
```diff
@@ -204,6 +204,25 @@ async def test_is_yaml_file(file: Text, is_yaml_file: bool):
...
+        ("data/test_yaml_stories/stories.yml", ["stories"], True),
+        ("data/test_yaml_stories/stories.yml", ["stories", "something_else"], False),
+        (
+            "data/test_domains/default_retrieval_intents.yml",
+            ["intents", "responses"],
+            True,
+        ),
+        ("data/test_yaml_stories/rules_without_stories.yml", ["rules"], True),
+        ("data/test_yaml_stories/rules_without_stories.yml", ["stories"], False),
+        ("data/test_stories/stories.md", ["something"], False),
```
joejuzl, line 219: This used to throw an error. Do we want to keep that behaviour?

rasa/shared/utils/io.py
```diff
@@ -321,25 +322,106 @@ def env_var_constructor(loader, node):
...
+    # because the set of env variables might change
+    if replace_env_vars:
+        return YAMLParser(reader_type=reader_type, replace_env_vars=True)
+
+    key = frozenset(reader_type) if isinstance(reader_type, list) else reader_type
+    if key in _parsers:
+        return _parsers[key]
+
+    new_parser = YAMLParser(reader_type=reader_type)
+    _parsers[key] = new_parser
```
joejuzl, line 410: we could use `setdefault` to set and get the parser in one line.

rasa/shared/utils/io.py
```diff
@@ -321,25 +322,106 @@ def env_var_constructor(loader, node):
...
+    return new_parser
+
+
 def read_yaml(content: Text, reader_type: Union[Text, List[Text]] = "safe") -> Any:
     """Parses yaml from a text.
 
     Args:
         content: A text containing yaml content.
         reader_type: Reader type to use. By default "safe" will be used
+        replace_env_vars: Specifies if environment variables need to be replaced
```
joejuzl, line 420: this arg does not seem to exist for this method.

rasa/shared/utils/io.py
```diff
@@ -321,25 +322,106 @@ def env_var_constructor(loader, node):
...
+            replace_environment_variables()
+        else:
+            # we need to do this because `yaml` adds contains a global `implicit_resolvers` list
+            # and the resolves from there cannot be de-registered
+            yaml.SafeConstructor.add_constructor("!env_var", lambda _, node: node.value)
+
+        self._save_modified_yaml_parameters()
+        self.restore_default_yaml_parameters()
+
+    def _save_default_yaml_parameters(self) -> None:
```
joejuzl, line 358: Do we need a copy of the default saved in each parser object if they are always the same?

rasa/shared/utils/io.py
```diff
@@ -321,25 +322,106 @@ def env_var_constructor(loader, node):
     yaml.SafeConstructor.add_constructor("!env_var", env_var_constructor)
 
 
+class YAMLParser(yaml.YAML):
+    """A custom parser for YAML.
+
+    It's used because `ruamel.yaml` stores all its information globally.
+    Which means it's impossible to create different parsers with different constructors
+    and resolvers without remembering the previous configuration and restoring it once
```
joejuzl, line 330: Is this threadsafe? Unfortunately we spawn new threads for some API endpoints...

joejuzl (changes requested):

tests/shared/core/training_data/story_reader/test_yaml_story_reader.py
alwx: That seems to be almost impossible with the current way of checking since we do not convert to yaml anymore. We can add `is_likely_yaml_file` check here but the behaviour would not be the same anyway.

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -175,11 +176,18 @@ def is_key_in_yaml(cls, file_path: Union[Text, Path], *keys: Text) -> bool:
...
         Raises:
-            YamlException: if the file seems to be a YAML file (extension) but
-                can not be read / parsed.
+            FileNotFoundException: if the file cannot be found.
         """
-        content = rasa.shared.utils.io.read_yaml_file(file_path)
-        return any(key in content for key in keys)
+        try:
+            with open(file_path) as file:
+                return all(
```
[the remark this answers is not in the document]
alwx: Updated.

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
alwx: Updated

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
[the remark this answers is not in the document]
alwx: Sorry, I got confused by the docstring that was here before — it was clearly saying "all of the keys" while in fact it worked differently.

rasa/shared/utils/io.py
alwx: I updated and simplified it — now it's thread-safe as well.

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -165,21 +166,28 @@ def is_stories_file(cls, file_path: Union[Text, Path]) -> bool:
...
-            YamlException: if the file seems to be a YAML file (extension) but
-                can not be read / parsed.
+            FileNotFoundException: if the file cannot be found.
         """
-        content = rasa.shared.utils.io.read_yaml_file(file_path)
-        return any(key in content for key in keys)
+        try:
+            with open(file_path) as file:
+                return any(
+                    any(line.lstrip().startswith(f"{key}:") for key in keys) for line in file
```
alwx, line 184: I've also added `.lstrip()` here because apparently this is a valid YAML:
```
        stories:
        - story: simple_story_without_checkpoint
          steps:
          - intent: simple
          - action: utter_default
          - action: utter_greet
```

rasa/shared/utils/io.py
joejuzl: bump :)

joejuzl (approved):
Looks good! But would be good to also see some numbers for how much it's sped things up.

alwx:
The duration of Multiwoz test from @wochinge's PR:

**Before changes:**
More than 1h and still running :(

**After changes:**
2059.36s call     tests/shared/importers/test_importer.py::test_profile_training_data_loading
(it's 0:34:24)

wochinge:
Thanks for sharing the reslts @alwx ! Is there a bigger speedup if we don't interpolate env variables or did your resolver fix already address things?

alwx:
It seems like the biggest speedup is caused by this: https://github.com/RasaHQ/rasa/pull/7968/files#diff-2725ed843eaab36adb512c85fbed96f62150901754e5fcae30c182d54d80fa5eR184 and this one simple line: https://github.com/RasaHQ/rasa/pull/7968/files#diff-d0c59274cb6aad666a5addf9f758510ba9a2ecdcb48b08a16b904f90c980e017R326. 
By doing this, we're not adding multiple resolvers, and by doing the first thing we're making the parsing much faster.

Issue #7892, opened 2021-02-05
Investigate rasahq/rasa#7272


TyDunn:
Investigate #7272

Background: We have the datasets. Tanja gave us some tips. Some code changes might be needed as part of this investigation (commenting out things / if a hack solves it).

**Definition of done**
- [x] Determine how long it takes to load MultiWOZ in [huggingface datasets](https://huggingface.co/docs/datasets/loading_datasets.html)
- [ ] Determine what needs to be done for #7272
- [ ] Create tickets for what plan to solve #7272

TyDunn:
@wochinge I updated the definition of done for this issue. Please let me know if there is anything else you need :)

wochinge:
News so far:
* Loading the multiwoz dataset using Huggingface on my machine takes ~ 1 second
* Loading the multiwoz dataset on a 4CPU, 16GiB Gcloud VM using Rasa hasn't terminated yet after > 12 hours

wochinge:
@TyDunn As stated I don't we should compare ourselves to Huggingface because our code does more (e.g. like fingerprinting the model). We can of course focus solely on the importing but than the entire process will still be very slow so I think it makes sense to focus upon the process from reading until (excluded) training. Considering that my tests still not terminated I would aim for results in the range of minutes rather than seconds (probably everything < 20min is already a huge win.

TyDunn:
@wochinge I completely agree. Now that we know huggingface loading of multiwoz takes seconds, while we take hours. I have updated the definition of done in #7272

wochinge:
I brought it down to 220 seconds. It's very hacky as I just commented out some things, but I guess something in this range is realistic

TyDunn:
That's awesome. I would guess that something around that would be good enough

wochinge:
Suggested next steps based on [PR](https://github.com/RasaHQ/rasa/pull/7944/files) and [the comment](https://github.com/RasaHQ/rasa/issues/7892#issuecomment-778069664) from above (this is a order of priority) @RasaHQ/enable-squad

**Where we started:**: Rasa with multiwoz on a gcloud machine with 4 cpus and 16 GiB memory didn't finish loading the training data within 18 hours.

1. Replace the implementation to look for keys in a yaml file with this (trivial): https://github.com/RasaHQ/rasa/pull/7944/files#r575331363

[...]

PR #7970, opened 2021-02-16
load schema files for `pykwalify` to avoid global `yaml` usage

wochinge:
**Proposed changes**:
- `pykwalify` uses a global `ruamel.yaml` instance. This instance sets object attributes while reading files (e.g. the YAML schema files itself!) . This can lead to problem when we  have multiple concurrent yaml validations.  Fix: Either load the yaml files for pykwalify (what I did in my PR) or provide the schema files in json

### DIFF
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -154,9 +154,15 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
         PACKAGE_NAME, SCHEMA_EXTENSIONS_FILE
     )
 
+    # Load schema content using our YAML loader as `pykwalify` uses a global instance
+    # which can fail when used concurrently
+    schema_content = rasa.shared.utils.io.read_yaml_file(schema_file)
+    schema_utils_content = rasa.shared.utils.io.read_yaml_file(schema_utils_file)
+    schema_content = dict(schema_content, **schema_utils_content)
+
     c = Core(
         source_data=source_data,
-        schema_files=[schema_file, schema_utils_file],
+        schema_data=schema_content,
         extensions=[schema_extensions],
     )
 
--- a/changelog/7970.bugfix.md
+++ b/changelog/7970.bugfix.md
@@ -1,0 +1,3 @@
+Fixed a YAML validation error which happened when executing multiple validations
+concurrently. This could e.g. happen when sending concurrent requests to server
+endpoints which process YAML training data.
--- a/tests/shared/utils/test_validation.py
+++ b/tests/shared/utils/test_validation.py
@@ -1,3 +1,5 @@
+from threading import Thread
+
 import pytest
 
 from pep440_version_utils import Version
@@ -12,6 +14,7 @@ from rasa.shared.constants import (
     DOMAIN_SCHEMA_FILE,
     LATEST_TRAINING_DATA_FORMAT_VERSION,
 )
+from rasa.shared.nlu.training_data.formats.rasa_yaml import NLU_SCHEMA_FILE
 from rasa.shared.utils.validation import KEY_TRAINING_DATA_FORMAT_VERSION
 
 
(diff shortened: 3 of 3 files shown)


rasa/shared/utils/validation.py
```diff
@@ -154,9 +154,15 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
         PACKAGE_NAME, SCHEMA_EXTENSIONS_FILE
     )
 
+    # Load schema content using our YAML loader as `pykwalify` uses a global instance
+    # which can fail when used concurrently
+    schema_content = rasa.shared.utils.io.read_yaml_file(schema_file)
+    schema_utils_content = rasa.shared.utils.io.read_yaml_file(schema_utils_file)
```
wochinge, line 160: alternatively we can convert all schema files to `json` as this would achieve the same effect and might be better re performance

wochinge: Or even better: define the schemas as Python objects - this saves us reading things entirely. What do you think?

wochinge: (this is related to my insights in https://github.com/RasaHQ/rasa/issues/7892#issuecomment-778302866)

wochinge:
@alwx Not ready for review yet, I rather want your input before I start making bigger changes

rasa/shared/utils/validation.py
alwx: I think it might make sense to define schemas as Python objects (don't see any need to go with JSON in this case) — but we can also create an issue and do it later.

alwx (approved):

rasa/shared/utils/validation.py
wochinge: True. I'll add a changelog, create a new issue + merge 👍🏻

wochinge: https://github.com/RasaHQ/rasa/issues/7977

joejuzl:
I wonder if this is the best fix? Maybe it would be safer going forward to just stop using multithreading? We could easily run into more issues like this in the future unless we ensure the threadsafety of all the code we write, and all the libraries we use!  We could use multiprocessing instead (should be a very easy change).

wochinge:
@joejuzl Great suggestion! Unfortunately this is not as easy expected. 

1. Pickle error for the local function `run`: That one was easy to fix
2. Pickle error for the decorared function as it's a local function as well (all part of `create_app`)🤯 
3. Pickle error for the `request` object.

That's the code I used:

```python
def _run_in_process(f, request: Request, *args: Any, **kwargs: Any) -> None:
    # This is a new thread, so we need to create and set a new event loop
    thread_loop = asyncio.new_event_loop()
    asyncio.set_event_loop(thread_loop)

    try:
        return thread_loop.run_until_complete(f(request, *args, **kwargs))
    finally:
        thread_loop.close()


def run_in_thread(f: Callable[..., Coroutine]) -> Callable:
    """Decorator which runs request on a separate thread.

    Some requests (e.g. training or cross-validation) are computional intense requests.
    This means that they will block the event loop and hence the processing of other
    requests. This decorator can be used to process these requests on a separate thread
    to avoid blocking the processing of incoming requests.

    Args:
        f: The request handler function which should be decorated.

    Returns:
        The decorated function.
    """

    @wraps(f)
    async def decorated_function(
        request: Request, *args: Any, **kwargs: Any
    ) -> HTTPResponse:
        with concurrent.futures.ProcessPoolExecutor() as pool:
            return await request.app.loop.run_in_executor(
                pool, functools.partial(_run_in_process, f, request, *args, **kwargs)
            )

    return decorated_function
```

Do you have an idea? Otherwise I'd suggest to go with the current solution for now. We can still iterate on the solution in case we see threading errors elsewhere

joejuzl:
@wochinge ahaa - that's annoying 🤦 
Then yeah, let's fix it like this and create an issue to followup.

wochinge:
it was a very nice bug 😄 And now I have more ideas / opinions about our YAML validation.

Follow up issue https://github.com/RasaHQ/rasa/issues/7980

Issue #7753, opened 2021-01-19
RasaYAMLWriter renders multiline NLU examples incorrectly


chdorner:
**Rasa version**: 2.2.5 (`master` at 5549b07d70a39819bf793017de3badd7ea903bc6)

**Rasa SDK version** (if used & relevant): –

**Rasa X version** (if used & relevant): –

**Python version**: 3.8.6

**Operating system** (windows, osx, ...): Mac OS 11

**Issue**: 
YAML output of `RasaYAMLWriter` for multiline examples is not identical to the original YAML input.

**Error (including full traceback)**:
–

**Command or request that led to error**:
```python
from rasa.shared.nlu.training_data.formats.rasa_yaml import RasaYAMLReader, RasaYAMLWriter
yaml_string = f"""version: "2.0"
nlu:
- intent: greet
  examples:
  - text: |
      Hello,
      My name is Christof
  - text: |
      Hello,
      You can call me Christof
"""
training_data = RasaYAMLReader().reads(yaml_string)
print(RasaYAMLWriter().dumps(training_data))
```

Outputs:
```yaml
version: "2.0"
nlu:
- intent: greet
  examples: |
    - Hello,\nMy name is Christof
    - Hello,\nYou can call me Christof
```

**Content of configuration file (config.yml)** (if relevant):
–

**Content of domain file (domain.yml)** (if relevant):
–

twerkmeister:
Okay, I looked into this a bit, and one could argue that the issue is not necessarily that the input and output isn't identical, since both variants might be valid. Also, there could be a mix of variants within a single document for different intents, so we would need to track these layout features about the data which we would need a strong use case for to justify IMO.

However, the output differs in the sense that it actually outputs `\\n` which is an escaped backslash plus the letter n. Re-parsing this output as training data results in slightly different messages being parsed.

On top of that, re-parsing the string (note: which is different from the actual output)
```
version: "2.0"
nlu:
- intent: greet
  examples: |
    - Hello,\nMy name is Christof
    - Hello,\nYou can call me Christof
``` 
verbatim fails as the parser does not like the explicit line breaks in the multi line example format. 

These two problems are definitely not great.

It seems to me that a sensible solution would be to use the dictionary output format either by default or whenever the examples contain line breaks. @chdorner @wochinge What do you think?

@chdorner can you give us some insight into the value these line breaks have for you? In most cases these whitespace characters will be discarded by the tokenizer, unless you are very intentional about them.

[...]

PR #7761, opened 2021-01-20
Support for writing NLU intent/example metadata to YAML

chdorner:
**Proposed changes**:
This adds support to the `RasaYAMLWriter` for writing intent and example metadata back into YAML files.
The parsing functionality has been added in #5989.

Fixes #7731 

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- ~~[ ] updated the documentation~~
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #7978, opened 2021-02-17
consistently reading and writing multiline intent examples

twerkmeister:
**Proposed changes**:
- #7753 
- Whenever there are escape characters, e.g. \n or \r we use the "as object" yaml style
- In the as object yaml style, we do not encode escape characters anymore (previously \n would become \\\\n)

**Consequences/Things to keep in mind**
- If someone changes a single intent example in rasa x to use multiline for an intent that didn't have any metadata examples before, all examples of that same intent will now use the dictionary format, resulting in a big change of the yaml and a big git diff. However, same has been the case before when some metadata was added

**Possible further work**
- In case someone had multiline examples before, these would have been transformed to the single line notation with encoded characters. One of the problems was that these were never decoded again when reading the file. With this change, we stopped creating this problem, but we are not recovering the past problem. I could add code that decodes encoded escape chars when reading the examples.


sara-tagger:

rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -23,6 +23,7 @@
 
 from rasa.shared.nlu.training_data.training_data import TrainingData
 from rasa.shared.nlu.training_data.message import Message
+from rasa.shared.nlu.training_data.util import has_string_escape_chars
```
wochinge, line 26: minor: We have a Python Code convention to not import functions from modules but rather their modules (see [here](https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#900c90130b4e4fe6b688f30c3826e764))

tests/shared/nlu/training_data/formats/test_rasa_yaml.py
```diff
@@ -519,3 +521,20 @@ def test_responses_text_multiline_is_preserved():
...
     assert dumped == RasaYAMLWriter().dumps(dumped_result)
+
+
+def test_intent_examples_multiline_consistency():
+    """Test that multiline examples are written back as multiline examples."""
+
+    training_data_file = "data/test_multiline_intent_examples_yaml/nlu.yml"
+    training_data_from_disc = RasaYAMLReader().read(filename=training_data_file)
+
+    tmpfile = tempfile.NamedTemporaryFile(delete=False)
```
wochinge, line 532: I'd suggest to use the `tmp_path` fixture from pytest. This why everything is cleaned up correctly after the test.

wochinge: nit picking: `pathlib` offers you a `Path.read_text(...)` which would make the code a tiny bit more concise

tests/shared/nlu/training_data/test_util.py
```diff
@@ -0,0 +1,16 @@
...
+    "s, result",
+    [
+        ("Hey,\nmy name is Christof", True),
+        ("Howdy!", False),
+        ("A\tB", True),
+        ("Hey,\rmy name is Thomas", True),
+        ("Hey, my name is Thomas", False),
+    ],
+)
+def test_has_string_escape_chars(s, result):
```
wochinge, line 15:
```suggestion
def test_has_string_escape_chars(s: Text, result: bool):
```

wochinge, line 15:
```suggestion
def test_has_string_escape_chars(s: Text, has_escaped_char: bool):
```

rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -484,8 +485,12 @@ def process_training_examples_by_key(
             if intent_metadata:
                 intent[KEY_METADATA] = intent_metadata
 
-            render_as_objects = any(KEY_METADATA in ex for ex in converted)
-            if render_as_objects:
+            examples_have_metadata = any(KEY_METADATA in ex for ex in converted)
+            example_texts_have_escape_chars = any(
+                has_string_escape_chars(ex.get(KEY_INTENT_TEXT, "")) for ex in converted
```
wochinge, line 490: optional: should it be `has_string_escaped_chars`?

data/test_multiline_intent_examples_yaml/nlu.yml
```diff
@@ -0,0 +1,14 @@
+version: "2.0"
+nlu:
+- intent: greet
+  examples:
+  - text: |
+      Hello,
+      My name is Christof
+  - text: |
+      Hello,
+      You can call me Christof
```
wochinge, line 10: how about also having a test for an example which is something like `- Hello,\nMy name is Christof`?

wochinge (commented):
Looks good! 💯 I'm on vacation from tomorrow until next Friday so I'd leave the final review to @chdorner . @melindaloubser1 Could you please add Thomas to the team member lest for Sara tagger 🙌🏻

rasa/shared/nlu/training_data/formats/rasa_yaml.py
twerkmeister: These are actually called escape characters :) https://en.wikipedia.org/wiki/Escape_character

data/test_multiline_intent_examples_yaml/nlu.yml
twerkmeister: TLDR: I think we should discourage using the \n notation in the training file and instead have people use literal line breaks if they want them. Having a uniform notation will lead to less unexpected automatic changes of the file structure. We could potentially add a warning if people use examples with escape chars

Long answer: This is a good question and relates to my thoughts I noted under *further work*. If you would read in a file containing - Hello,\nMy name is Christof? What you would actually be reading in is `\\n` which is an escaped backslash plus n. A real line break would just be a line break in the file. We could decode `\\n` to turn into `\n`, but then when it's rewritten, there isn't a simple way to know which notation was used originally, so we would probably not encode it and instead write a literal line break, thus leading to a change in the file. Even worse, an example might use both literal line breaks and `\n`, so you would need to track their positions if we wanted to make sure there are no changes, which seems way out of proportion to me.

twerkmeister:
Sorry for the mess, now it's appropriately forked off 2.3.x and going back into that.

rasa/shared/nlu/training_data/formats/rasa_yaml.py
```diff
@@ -541,9 +549,7 @@ def _render_training_examples_as_objects(examples: List[Dict]) -> List[Dict]:
 
         def render(example: Dict) -> Dict:
             text = example[KEY_INTENT_TEXT]
-            example[KEY_INTENT_TEXT] = LiteralScalarString(
-                TrainingDataWriter.generate_string_item(text)
-            )
+            example[KEY_INTENT_TEXT] = LiteralScalarString(text + "\n")
```
chdorner, line 552: I've refactored the `TrainingDataWriter.generate_list_item` only for this line ([here](https://github.com/RasaHQ/rasa/pull/7761/files#diff-cd73e9fa63d1bf46bf4e4ce0c49e5f063894ddd9c6ad227a1ce06ff16f839d97)). It might make sense to move undo that change as well and remove the now unused `TrainingDataWriter.generate_string_item`?

tests/shared/nlu/training_data/test_util.py
```diff
@@ -0,0 +1,18 @@
+from rasa.shared.nlu.training_data.util import has_string_escape_chars
+import pytest
+from typing import Text
+
+
+@pytest.mark.parametrize(
+    "s, result",
+    [
+        ("Hey,\nmy name is Christof", True),
```
chdorner, line 9: Yes! I made it into the test data 🎉

data/test_multiline_intent_examples_yaml/nlu.yml
chdorner: I personally agree that we should set some expectations and not try to support every possible way that people could write YAML. I guess that is pretty much the discussion in the "Revisit training data + IVC" ice box item.

chdorner (commented):
Looking really good! What do you think about undoing that refactor that I did previously so that we can get rid of the now unused `TrainingDataWriter.generate_string_item` function?

rasa/shared/nlu/training_data/formats/rasa_yaml.py
wochinge: Isn't the `/` the "escape character" and something like `\n` is an "escape sequence"?

data/test_multiline_intent_examples_yaml/nlu.yml
wochinge: Sounds good. Should we create an issue to raise a warning for the `\n` case? 🤔

chdorner (approved):
🍡

PR #7984, opened 2021-02-17
Making `rasa data convert config` update `config.yml` and `data/rules.yml` correctly with Mapping Policy

Imod7:
**Proposed changes in Mapping Policy & Model Config Migration**:
- When migrating the `MappingPolicy` we append the `RulePolicy` in `config.yml` even if new_rules are empty.
- When migrating the model config, we create the `rules.yml` file even if it is empty.
- Added the test case of having only one policy (the Mapping Policy) and no rules.

- [x] added a test case for the functionality


Imod7:
Hey @joejuzl while checking the CI warnings, I bumped into this one :

_UserWarning: Found a rule-based policy in your pipeline but no rule-based training data. Please add rule-based stories to your training data or remove the rule-based policy (`RulePolicy`) from your your pipeline._

So based on this warning I should not create a `rules.yml` file (if there are no rules available) neither append the `RulePolicy` in the config file. I guess the only thing that I should do is remove the `MappingPolicy` from the config file during migration and change the success message. What do you think?

rasa/cli/data.py
```diff
@@ -472,7 +472,7 @@ def _migrate_model_config(args: argparse.Namespace) -> None:
     if fallback_rule:
         new_rules.append(fallback_rule)
 
-    if new_rules:
+    if new_rules or model_configuration["policies"]:
```
joejuzl, line 475: So the behaviour we want is to always create a rules file even if empty?

rasa/cli/data.py
```diff
@@ -643,15 +643,23 @@ def _backup(path: Path) -> None:
...
+        suffix = "rule"
+        verb = "was"
 
     rasa.shared.utils.cli.print_success(
-        f"Finished migrating your policy configuration 🎉.\n"
-        f"The migration generated {len(new_rules)} {suffix} which {verb} added to "
-        f"'{output_file}'."
+        "Finished migrating your policy configuration 🎉."
     )
+    if len(new_rules) == 0:
```
joejuzl, line 656: Can we just have one set of `if` and `else`/`elif`s

rasa/cli/data.py
```diff
@@ -643,15 +643,23 @@ def _backup(path: Path) -> None:
...
 def _print_success_message(new_rules: List["StoryStep"], output_file: Path) -> None:
-    if len(new_rules) > 1:
-        suffix = "rule"
-        verb = "was"
-    else:
+    if len(new_rules) > 1 or len(new_rules) == 0:
         suffix = "rules"
         verb = "were"
+    else:
+        suffix = "rule"
```
joejuzl, line 650: I know this was already like this, but `suffix` seems like the wrong name

rasa/core/config.py
```diff
@@ -216,7 +216,7 @@ def migrate_mapping_policy_to_rules(
     # finally update the policies
     policies = _drop_policy(MappingPolicy.__name__, policies)
 
-    if new_rules and not has_rule_policy:
+    if not has_rule_policy:
```
joejuzl, line 219: Do we always want the rule policy?

joejuzl (changes requested):

rasa/cli/data.py
Imod7: I am not sure what is the best approach so what I do now is creating the `rules.yml` file even if empty. However, after reading the warning above I think I should not create an empty `rules.yml` file. Also, I should change the implementation to the following :
- If there are no rules, then the `rules.yml` file should not be created.
- Adjust the success message to say something like "_The migration generated 0 rules so the file 'data/rules.yml' was not created._"
- The migration should edit the `config.yml` file by removing the MappingPolicy but since there are no rules, it should not append the RulePolicy.
So, I should actually make the opposite changes from the ones that I have now. What is happening right now is described in the description. Maybe @koaning has a suggestion on this one? Thank you!

koaning: I don't see the harm in creating an empty `rules.yml` file or a `RulePolicy` that doesn't influence the conversation. But this is a *highly personal opinion that is created by thinking outloud*. One thing that I do not understand though, is `rasa data convert config` a new command? I couldn't find it on the docs.

Imod7: So, looking at the codebase : 
- it will raise a warning [here](https://github.com/RasaHQ/rasa/blob/b13e70bbbb8b774fb66340af96914da24d244465/rasa/core/policies/ensemble.py#L168) if there is a `RulePolicy` in the pipeline but no rules available so I guess until now the best practice was to **not** have an empty `rules.yml` file. 
- Also, this [test](https://github.com/RasaHQ/rasa/blob/b13e70bbbb8b774fb66340af96914da24d244465/tests/core/test_ensemble.py#L471) is there to check that the warning is raised correctly.

The `rasa data convert config` command is mentioned in the docs [here](https://rasa.com/docs/rasa/migration-guide) where it also says 
> _It will also add a rules.yml **_if necessary_**._

So if it is not necessary (eg. no rules available), it will not be created.

So, it looks like everything is aligned to not having the empty `rules.yml` file but of course the current state of the code is not necessarily the best practice and that it should not be questioned. I agree that it is no harm at all to have an empty rules.yml file or a RulePolicy that is not affecting the conversation. 

However, correct me if I am wrong but I think the goal of a migration is to convert what the user **already** has to the corresponding format of the new Rasa version and not what might be useful for the user to have. So based on that I would stick with what the current logic is and not create an empty rules file nor having a `RulePolicy` if there are no rules available. So in this PR I am in favour of : 
- only adjusting the success message to reflect that the `rules.yml` is not created and 
- make sure that the MappingPolicy is removed from the pipeline.
What do you think ?

koaning: I'm less comfortable with having a strong opinion here because this is an area of the codebase that I'm unfamiliar with. That said, I might prefer having a `rules.yml` without any rules in it. The reason is that practically, a user will use rules somewhere in their assistant at some point. Might as well have a file/settings around that's easy to add rules to.

rasa/core/config.py
Imod7: From the last comment of @koaning with the logic of which I agree, yes, we will now always have the `RulePolicy` when we are migrating the `MappingPolicy` (even if we have no rules). From what I have seen this function will only be called if the `MappingPolicy` is present in the `config.yml` file so it doesn’t affect other cases.

rasa/cli/data.py
Imod7: Not really because we have different criteria in each check. More specifically : 
- the variables 'rules_text' and 'verb' are changing if the rules are >1 or equal to 0 **vs** we have only 1 rule (else case). 
- but then the actual success message is different only in the case that we have 0 rules (_...so **no** rules are added...._) **otherwise** the syntax/text is the same (case of having >1 rules or only 1 rule).

Imod7 (commented):
@koaning I agree with the logic you proposed since from a user's perspective is more convenient to already have a `rules.yml` file even if empty. Maybe it makes also a better user experience because if the file is not there and they need rules, they might not be sure if they should create it and where exactly (so they will have to look in the docs or in the examples) or they might feel they have something wrong in their setup if they don't see it there, etc.

So I kept this logic, to create during the `MappingPolicy` migration an empty `rules.yml` file and add a `RulePolicy` in the `config.yml` file even if it is not necessarily used. 

But I also left the warning check (and the related test) as it is right now. So we will still have the following warnings raised :
- If there is a RulePolicy and no rule data, it raises a warning that we should add rule data.
- If there are rule data and no RulePolicy, it raises a warning that we should add the policy.

I think it is important to leave these warnings (I am actually referring to the first one) since it is something that the user should take note of. 

Also, the success message now changes to the following : 
- if there is a MappingPolicy and no rules, the message is 
"_The migration generated 0 rules so no rules were added to 'data/rules.yml'._"

- if there is a MappingPolicy and at least one rule, the message is
"_The migration generated 1 rule which was added to 'data/rules.yml'._"

rasa/cli/data.py
Imod7: @koaning I agree with the logic you proposed since from a user's perspective is more convenient to already have a `rules.yml` file even if empty. Maybe it makes also a better user experience because if the file is not there and they need rules, they might not be sure if they should create it and where exactly (so they will have to look in the docs or in the examples) or they might feel they have something wrong in their setup if they don't see it there, etc.

rasa/cli/data.py
Imod7: Changed it to `rules_text`.

joejuzl (approved):

Issue #7578, opened 2020-12-17
Enable rasa test usage for e2e


Ghostvv:
Users need `rasa test` to validate and inform their decisions to update their training data. Without it, they cannot have a functioning live project using e2e, which in turns means that we wouldn't be able to get valuable feedback on e2e.

With `e2e` TED can extract entities using data in stories. Currently `rasa test` only runs NLU to compare entities.
We need to update the `rasa test` command so that the entities extracted by TED would be compared with entity labels in stories.

# Enable

**Definition of done (what needs to be true to close this issue)**
- [ ] If TED is configured to extract entities (when using e2e), `rasa test` should compare predictions to golden labels in the stories

PR #7995, opened 2021-02-19
Rasa test evaluates TED e2e entity extraction

joejuzl:
**Proposed changes**:
Fixes: #7578 
Adds functionality to `rasa test` to evaluate the entities TED extracts.


rasa/core/test.py
```diff
@@ -7,7 +7,9 @@
 
 from rasa import telemetry
 from rasa.core.policies.policy import PolicyPrediction
+from rasa.nlu.test import EntityEvaluationResult, evaluate_entities
```
joejuzl, line 10: should these be moved to shared?

rasa/core/test.py
```diff
@@ -481,13 +489,38 @@ def emulate_loop_rejection(partial_tracker: DialogueStateTracker) -> None:
...
+):
+    previous_event = tracker.events[-1]
+    if isinstance(previous_event, (UserUttered)):
+        entity_predictions = [
+            entity
+            for prediction_event in prediction.events
+            if isinstance(prediction_event, EntitiesAdded)
+            for entity in prediction_event.entities
+        ]
+        entity_targets = previous_event.entities
```
Ghostvv, line 505: how will it work in case of e2e testing? because I believe it'll contain both nlu entities and e2e entities

rasa/core/test.py
```diff
@@ -802,11 +860,21 @@ async def test(
...
             logger.info(f"Stories report saved to {report_filename}.")
+
         else:
             report, precision, f1, accuracy = get_evaluation_metrics(
                 targets, predictions, output_dict=True
             )
 
+        evaluate_entities(
+            entity_results,
+            {"TEDPolicy"},
```
Ghostvv, line 871: can we extract the name from entities? I think they contain the name of the extractor

rasa/core/test.py
```diff
@@ -955,7 +1023,8 @@ async def _evaluate_core_model(model: Text, stories_file: Text) -> int:
     agent = Agent.load(model)
     generator = await _create_data_generator(stories_file, agent)
     completed_trackers = generator.generate_story_trackers()
-    story_eval_store, number_of_stories = await _collect_story_predictions(
+
+    story_eval_store, number_of_stories, _ = await _collect_story_predictions(
```
Ghostvv, line 1027: why don't we get entity results here?

rasa/core/test.py
joejuzl: I thought that these entities were the ones specified in the story step, e.g. `my name is [joe](name)`. What do you mean by nlu entities in this context?

rasa/core/test.py
joejuzl: So find all the unique extractor names in the collection of entities and evaluate them?

rasa/core/test.py
joejuzl: This method is just returning the number successful stories. Should we also incorporate the entities into this calculation?

rasa/core/test.py
Ghostvv: > I thought that these entities were the ones specified in the story step
yes, I'm calling `entities specified in the story step` as `e2e entities`. But there are also other entities that are extracted by nlu pipeline. I think we should test them separately

rasa/core/test.py
Ghostvv: I meant instead of hard coding `{"TEDPolicy"}`. How do we do it for testing nlu pipeline? We have several extractors there

rasa/core/test.py
Ghostvv: in this case, I'd say no. Could you please add a comment?

rasa/core/test.py
joejuzl: For NLU it looks at every component in the pipeline and checks if it is a subclass of `EntityExtractor`. For core we could check the config of every policy for `ENTITY_RECOGNITION=True`?

Ghostvv: I think you can just extract source from `EntitiesAdded` event?

joejuzl: but then we would only evaluate for policies that successfully extracted an entity.

rasa/core/test.py
joejuzl: so if the nlu pipeline extracts an entity it will also end up in `previous_event.entities`?

Ghostvv: it'll end up in `UserUttered` but not in `EntitiesAdded`. I think @wochinge knows this better

rasa/core/test.py
alwx: Why? I don't see any big reason of moving it to `shared`.

rasa/core/test.py
```diff
@@ -481,13 +489,38 @@ def emulate_loop_rejection(partial_tracker: DialogueStateTracker) -> None:
     partial_tracker.update(ActionExecutionRejected(rejected_action_name))
 
 
+def _get_processor_entity_evaluation_result(
+    processor: "MessageProcessor",
+    tracker: DialogueStateTracker,
+    prediction: PolicyPrediction,
+):
+    previous_event = tracker.events[-1]
+    if isinstance(previous_event, (UserUttered)):
```
alwx, line 498: probably the parentheses is no needed here

rasa/core/test.py
```diff
@@ -577,7 +616,12 @@ async def _predict_tracker_actions(
     agent: "Agent",
     fail_on_prediction_errors: bool = False,
     use_e2e: bool = False,
-) -> Tuple[EvaluationStore, DialogueStateTracker, List[Dict[Text, Any]]]:
+) -> Tuple[
+    EvaluationStore,
+    DialogueStateTracker,
+    List[Dict[Text, Any]],
+    List[EntityEvaluationResult],
+]:
```
alwx, line 624: Not really related to this PR but that looks a bit crazy — I would prefer to use some custom data structure to at least have some names for all the stuff that's part of the tuple now.

alwx (changes requested):

rasa/core/test.py
wochinge: Vova is right. [This](https://github.com/RasaHQ/rasa/blob/954e4734d479924df7211cd3bc5503285d54a66b/rasa/shared/core/events.py#L758) and [this](https://github.com/RasaHQ/rasa/blob/954e4734d479924df7211cd3bc5503285d54a66b/rasa/shared/core/events.py#L418) should hopefully explain things.

joejuzl: but I don't think this is the case when running `test`. If you look at `_collect_user_uttered_predictions` the predicted entities are not added to the event.

rasa/core/test.py
joejuzl: I think this area of the code base would need some major refactoring for us to remove these kinds of return types.

rasa/core/test.py
wochinge: I think that's because `_clean_entity_results` is creating a new list before the action could add the `EntitiesAdded` event. I guess we have to predict the next action _before_ collecting `_collect_user_uttered_predictions` as we're are otherwise missing a piece of information.

joejuzl: I don't think `_clean_entity_results` matters here. I'm not using the `user_uttered_eval_store` I'm just using what's in the `partial_tracker` which doesn't include the predicted entities, only the ones from the story.

rasa/core/test.py
```diff
@@ -481,13 +489,38 @@ def emulate_loop_rejection(partial_tracker: DialogueStateTracker) -> None:
...
 
 
+def _get_processor_entity_evaluation_result(
+    processor: "MessageProcessor",
+    tracker: DialogueStateTracker,
+    prediction: PolicyPrediction,
+):
+    previous_event = tracker.events[-1]
+    if isinstance(previous_event, UserUttered):
+        entity_predictions = [
```
wochinge, line 499: How about making this very explicit?
```suggestion
        entities_predicted_by_policies = [
```

rasa/core/test.py
```diff
@@ -481,13 +489,38 @@ def emulate_loop_rejection(partial_tracker: DialogueStateTracker) -> None:
...
+            entity
+            for prediction_event in prediction.events
+            if isinstance(prediction_event, EntitiesAdded)
+            for entity in prediction_event.entities
+        ]
+        entity_targets = previous_event.entities
```
wochinge, line 505: I think you're right @joejuzl . It doesn't contain the `policy_entities` yet as we haven't applied the events from the prediction to the tracker yet.

rasa/core/test.py
```diff
@@ -496,13 +529,19 @@ def _collect_action_executed_predictions(
     gold_action_text = event.action_text
     gold = gold_action_name or gold_action_text
 
+    entity_result = None
```
wochinge, line 532: these are only entities predicted by policies, right?

rasa/core/test.py
```diff
@@ -679,16 +732,20 @@ async def _collect_story_predictions(
     logger.info(f"Evaluating {number_of_stories} stories\nProgress:")
 
     action_list = []
+    entity_results = []
```
wochinge, line 735: It's only entities predicted by policies, right?
```suggestion
    policy_entity_results = []
```

rasa/core/test.py
```diff
@@ -802,11 +860,21 @@ async def test(
             report_filename = os.path.join(out_directory, REPORT_STORIES_FILE)
             rasa.shared.utils.io.dump_obj_as_json_to_file(report_filename, report)
             logger.info(f"Stories report saved to {report_filename}.")
+
         else:
             report, precision, f1, accuracy = get_evaluation_metrics(
                 targets, predictions, output_dict=True
             )
 
+        evaluate_entities(
```
wochinge, line 869: don't we want to show the failed entity predictions within the stories? This would dump them in a separate file, right? My intuition would have been to merge our policy predictions with the NLU predictions which were made in `_collect_user_uttered_predictions`

wochinge (commented):
Uff, what a complicated module. Sorry, if my comments are a bit over the place.

rasa/core/test.py
Ghostvv: for e2e we need to do exactly that

wochinge: I think what Joe means is that this wouldn't work in case the `TEDPolicy` isn't extracting any entity (and hence not adding an `EntitiesAdded` event)

changelog/7578.improvement.md
```diff
@@ -0,0 +1 @@
+When `TED` is configured to extract entities `rasa test` now evaluates them against the labels in the test stories.
```
wochinge, line 1: Should we also mention how / where the results are visible?

data/test_e2ebot/tests/test_stories.yml
```diff
@@ -0,0 +1,12 @@
+stories:
```
wochinge, line 1: to avoid warnings in the CI
```suggestion
version: "2.0"

stories:
```

rasa/core/test.py
```diff
@@ -481,13 +489,38 @@ def emulate_loop_rejection(partial_tracker: DialogueStateTracker) -> None:
     partial_tracker.update(ActionExecutionRejected(rejected_action_name))
 
 
+def _get_e2e_entity_evaluation_result(
+    processor: "MessageProcessor",
+    tracker: DialogueStateTracker,
+    prediction: PolicyPrediction,
+):
```
wochinge, line 496:
```suggestion
) -> Optional[EntityEvaluationResult]:
```

rasa/core/test.py
```diff
@@ -481,13 +489,38 @@ def emulate_loop_rejection(partial_tracker: DialogueStateTracker) -> None:
...
+        if entity_targets or entities_predicted_by_policies:
+            text = previous_event.text
+            parsed_message = processor.interpreter.featurize_message(
+                Message(data={TEXT: text})
+            )
+            tokens = parsed_message.get(TOKENS_NAMES[TEXT])
+            return EntityEvaluationResult(
+                entity_targets, entities_predicted_by_policies, tokens, text
+            )
+
```
wochinge, line 515: optional:
```suggestion

    return None
```

tests/core/test_evaluation.py
```diff
@@ -290,3 +290,34 @@ async def test_retrieval_intent_wrong_prediction(
 
     # check if the predicted entry contains full retrieval intent
     assert "# predicted: chitchat/ask_name" in failed_stories
+
+
+async def test_e2e_with_entity_evaluation(e2e_bot_agent: Agent, tmpdir: Path):
```
wochinge, line 295: `tmpdir` has a different (weird) type
```suggestion
async def test_e2e_with_entity_evaluation(e2e_bot_agent: Agent, tmp_path: Path):
```

data/test_e2ebot/config.yml
```diff
@@ -0,0 +1,16 @@
...
+  - name: CountVectorsFeaturizer
+  - name: CountVectorsFeaturizer
+    analyzer: char_wb
+    min_ngram: 1
+    max_ngram: 4
+  - name: DIETClassifier
+    epochs: 5
+policies:
+- name: TEDPolicy
+  epochs: 5
```
wochinge, line 16: we can't use less epochs for either of them, no?

data/test_e2ebot/domain.yml
```diff
@@ -0,0 +1,16 @@
+version: "2.0"
+
+actions:
+ - utter_greet
```
wochinge, line 4: that's duplicate (stuff from `responses:` will automatically be added to `actions`)
```suggestion
```

wochinge (commented):
Looks good to me 🚀 @Ghostvv is probably the best person to give the final thumbs up 👍🏻

rasa/core/test.py
joejuzl: yes exactly, and we would want to show that the recall is 0.

rasa/core/test.py
joejuzl: As discussed with @wochinge, we will keep the scope to evaluating entities in separate files like for the NLU components as there is a bug: https://github.com/RasaHQ/rasa/issues/8152 currently for displaying incorrect entity extracting in failed test stories.

rasa/core/test.py
Ghostvv: I see, yeah. Is there a way to know which algorithms are expected to extract these entities? I don't like the idea of hardcoding TED here

data/test_e2ebot/config.yml
joejuzl: Thought this was the lowest, but now seems like I can get down to `1` for DIET and `4` for `TED`.

rasa/core/test.py
joejuzl: Well we could look for `ENTITY_RECOGNITION` in the config of every policy in the ensemble. But that would only work if that exact config key was always used.

Ghostvv: maybe for simplicity, in order to unblock us, let's create a constant with a list of policies that can extract e2e entities for now?

rasa/shared/core/constants.py
```diff
@@ -80,3 +80,5 @@
 
 USE_TEXT_FOR_FEATURIZATION = "use_text_for_featurization"
 ENTITY_LABEL_SEPARATOR = "#"
+
+POLICIES_THAT_EXTRACT_ENTITIES = {"TEDPolicy"}
```
Ghostvv, line 84: should this be
```suggestion
POLICIES_THAT_EXTRACT_ENTITIES = {f"{TEDPolicy.__class__._name__}"}
```
?
also in TEDPolicy in line `728`?

Ghostvv: let's keep a string

Ghostvv (commented):
@joejuzl did you test it how does it work when you have e2e entities, but run an e2e test (core+nlu testing)? Does it differentiate that those entities are not e2e entities?

joejuzl:
> @joejuzl did you test it how does it work when you have e2e entities and nlu entities, but run an e2e test (core+nlu testing)? Does it differentiate between nlu entities and e2e entities?

@vova all the different entity extractor's results are stored in files prefixed with the name. So for example there will be:
```
DIETClassifier_confusion_matrix.png
DIETClassifier_report.json
TEDPolicy_confusion_matrix.png
TEDPolicy_report.json
```

Ghostvv:
> all the different entity extractor's results are stored in files prefixed with the name.

yes, but does it correctly read which entities are nlu and which are e2e? Sorry, I didn't test your PR, this is just the logical pitfall that came to my mind

joejuzl:
> but does it correctly read which entities are nlu and which are e2e?

Yes the NLU entities come from the nlu training data, and the e2e entities come from the test stories.

Ghostvv (approved):
looks good functionality wise

alwx (approved):

PR #8007, opened 2021-02-22
add kwargs to kafka event broker

melindaloubser1:
**Proposed changes**:
- Add a **kwargs argument to the Kafka event broker to allow using the config for a Kafka event consumer, which may contain parameters not supported by a Kafka producer, such as in the case of deploying Rasa X with the helm charts


rgstephens (approved):
LGTM

melindaloubser1:
@rgstephens @federicotdn pointed out this will allow potential errors in endpoint configs to go unchecked, so it would be better to just add the `group_id` param back, and give a warning when it is provided to indicate that it doesn't do anything. I'll create a new pr for the changes.

Issue #7999, opened 2021-02-19
Force DenseWithSparseWeights to have at least one entry per row


JEM-Mosig:
**Description of Problem**:

Whenever DenseWithSparseWeights has completely zero weight rows (which can happen if sparsity is high), it effectively reduces the layer size, which is pointless. So we should prevent this from happening.

**Overview of the Solution**:

**Examples** (if relevant):

**Blockers** (if relevant):

**Definition of Done**:

JEM-Mosig:
I'm also forcing every input to be connected to at least one output, because it doesn't make sense to ignore inputs randomly.

JEM-Mosig:
changes to the sparse layers definitely stabilize TED's performance at low densities: 

![image](https://user-images.githubusercontent.com/31377126/111444068-4a3f0980-870a-11eb-9d1f-ff5cf6f4eb0e.png)

Solid lines show performance at fixed density (see legend) with my modified sparse layers. Dash-dotted lines show the equivalent on the main branch.

The solid green curve is the fully dense layers one. The orange curves are the 20% density (or 80% sparsity) which are our default - for those my changes don't do anything.

The dash-dotted red curve is further to the left, because on main sparse layers are allowed to drop inputs or outputs and thus can have fewer trainable weights.

PR #8011, opened 2021-02-22
Force DenseWithSparseWeights to produce dense output and use all inputs

JEM-Mosig:
**Proposed changes**:
- Renames `DenseWithSparseWeights` into `RandomlyConnectedDense`
- Adds guarantees that even at density zero the output is dense and every input is connected to at least one output
- The former `weight_sparsity` parameter is now roughly equivalent to `1 - connection_density`, except at very low densities (high sparsities). 
- Fixes https://github.com/RasaHQ/rasa/issues/7999

- ~~updated the documentation~~


rasa/utils/tensorflow/constants.py
```diff
@@ -40,7 +40,7 @@
 DROP_RATE_LABEL = "drop_rate_label"
 CONSTRAIN_SIMILARITIES = "constrain_similarities"
 
-WEIGHT_SPARSITY = "weight_sparsity"
```
Ghostvv, line 43: you cannot simply remove it, you need to deprecate it where you recalculate CONNECTION_DENSITY

JEM-Mosig: Ahhh... right. Will do that.

JEM-Mosig:
/modeltest
```yml
include:
  - dataset: ["all"]
    config: ["all"]
```

JEM-Mosig:
/modeltest
```
dataset_branch: "variable_random_seed"
include:
  - dataset: ["Carbon Bot"]
    config: ["all"]
```

JEM-Mosig:
/modeltest
```yml
dataset_branch: "variable_random_seed"
include:
  - dataset: ["Carbon Bot"]
    config: ["all"]
```

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,149 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+
+        kernel_shape = tf.shape(self.kernel)
+        num_rows = kernel_shape[0].numpy()
+        num_cols = kernel_shape[1].numpy()
+
+        # Construct mask with given density and guarantee that every output is
+        # connected to at least one input
+        num_connected_per_output = max(
+            1, tf.cast(tf.math.ceil(self.density * num_rows), tf.int32)
         )
```
Ghostvv, line 240: this can be calculated inside `_random_mask_with_no_empty_rows`

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,149 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+        num_disconnected_per_output = num_rows - num_connected_per_output
+
+        # We create each column separately
+        # as [1 1 1 ... 1 0 0 0 ... 0] and then shuffle it randomly
+        kernel_mask_cols = [
+            tf.random.shuffle(
+                tf.concat(
+                    [
+                        tf.ones(
+                            shape=num_connected_per_output, dtype=self.kernel.dtype
```
Ghostvv, line 274: let, explicitly create columns 
```suggestion
                            shape=(num_connected_per_output, 1), dtype=self.kernel.dtype
```

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,149 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+                        ),
+                        tf.zeros(
+                            shape=num_disconnected_per_output, dtype=self.kernel.dtype
```
Ghostvv, line 277:
```suggestion
                            shape=(num_disconnected_per_output,1), dtype=self.kernel.dtype
```

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,149 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+                        ),
+                    ],
+                    axis=0,
+                )
+            )
+            for _ in range(num_cols)
+        ]
+        return tf.transpose(tf.stack(kernel_mask_cols, axis=0))
```
Ghostvv, line 285: then you should be able to concat over `axis=1`
```suggestion
        return tf.concat(kernel_mask_cols, axis=1)
```

rasa/utils/tensorflow/layers.py
Ghostvv: it means that every output sees exactly the same number of inputs, I'd say a bit of randomness would be nice

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,149 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+            num_cols: Number of columns in the mask
+            kernel_mask: The kernel mask matrix of size (num_rows, num_cols) that may
+                contain empty columns
+
+        Returns:
+            A mask with the missing connections.
+        """
+        empty_columns = tf.squeeze(
+            tf.where(tf.equal(tf.reduce_sum(kernel_mask, axis=1), 0)), -1
+        ).numpy()
```
Ghostvv, line 310: this is the equation for empty rows, no?

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,149 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+        random_rows = np.random.randint(num_cols, size=len(empty_columns))
+        indices = np.transpose([empty_columns, random_rows])
+        extra_entries = tf.sparse.to_dense(
+            tf.SparseTensor(
+                indices,
+                np.ones(len(empty_columns), dtype=np.float),
+                (num_rows, num_cols),
+            )
+        )
+        return tf.cast(extra_entries, dtype=kernel_mask.dtype)
```
Ghostvv, line 320: it seems to me that non-empty random mask got lost

changelog/7999.improvement.md
```diff
@@ -0,0 +1,3 @@
+Replace `weight_sparsity` with `connection_density` in all transformer-based models and add guarantees about internal layers.
+
+We rename `DenseWithSparseWeights` into `RandomlyConnectedDense`, and guarantee that even at density zero the output is dense and every input is connected to at least one output. The former `weight_sparsity` parameter is now roughly equivalent to `1 - connection_density`, except at very low densities (high sparsities).
```
Ghostvv, line 3: add that `sparsity` argument in the other layers changed as well

rasa/utils/tensorflow/layers.py
JEM-Mosig: It was added by `kernel_mask += self._missing_connections(num_rows, num_cols, kernel_mask)`

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,156 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+
+        # We might accidently have added a random connection on top of
+        # a fixed connection
+        kernel_mask = tf.clip_by_value(kernel_mask, 0, 1)
+
         self.kernel_mask = tf.Variable(
             initial_value=kernel_mask, trainable=False, name="kernel_mask"
         )
 
+    def _random_mask(
```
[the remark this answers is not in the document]
JEM-Mosig: The generation can be done in many ways, but this one does not involve type conversions, which I thought should be avoided. Using `tf.random.cathegorical` seems more complicated, and tensorflow doesn't have any other non-uniform `randint`.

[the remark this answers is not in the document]
Ghostvv: from my point of view, the way it created currently in `main` seems more intuitive to me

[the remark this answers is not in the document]
JEM-Mosig: Ok, I've changed it to that one.

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,150 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+
+        kernel_shape = tf.shape(self.kernel)
+        num_rows = kernel_shape[0].numpy()
+        num_cols = kernel_shape[1].numpy()
+
+        # Construct mask with given density and guarantee that every output is
+        # connected to at least one input
+
+        min_num_connected = max(num_rows, num_cols)
+        num_extra_connections = max(
```
Ghostvv, line 239: I don't think you need that anymore

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,150 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+            num_rows: Number of rows in the mask
+            num_cols: Number of columns in the mask
+            num_ones: Number of ones in the matrix
+            dtype: Type of the values
+
+        Returns:
+            A random mask matrix
+        """
+        mask = tf.random.uniform(tf.shape(self.kernel), 0, 1)
+        mask = tf.cast(tf.less_equal(mask, self.density), self.kernel.dtype)
```
Ghostvv, line 279: I think the convention is to use `<` for upper bounds
```suggestion
        mask = tf.cast(tf.math.less(mask, self.density), self.kernel.dtype)
```

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,150 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
+            A tiled and croped identity matrix.
+        """
+        short_dimension = min(num_rows, num_cols)
+        mask = tf.tile(
+            tf.eye(short_dimension, dtype=dtype),
+            [
+                tf.math.ceil(num_rows / short_dimension),
+                tf.math.ceil(num_cols / short_dimension),
+            ],
+        )[:num_rows, :num_cols]
```
Ghostvv, line 319: why do you need that? can you tile it precisely?

rasa/utils/tensorflow/layers.py
```diff
@@ -198,24 +188,134 @@ class DenseWithSparseWeights(tf.keras.layers.Dense):
...
-        kernel_mask = tf.cast(
-            tf.greater_equal(kernel_mask, self.sparsity), self.kernel.dtype
+
+        if self.density == 1.0:
+            self.kernel_mask = None
+            return
+
+        kernel_shape = tf.shape(self.kernel)
+        num_rows = kernel_shape[0].numpy()
+        num_cols = kernel_shape[1].numpy()
```
Ghostvv, line 233: please move it inside `_minimal_mask`

Ghostvv (commented):
looks good from my side

JEM-Mosig:
@alwx Can you review, please? Or assign someone else from Engineering?

JEM-Mosig:
@joejuzl (or anybody from Enable), when do you think you could have a look at this? The latest merges with main have broken something, but I want to hold off on fixing it until just before you've got some time scheduled.

rasa/utils/train_utils.py
```diff
@@ -104,6 +106,28 @@ def update_deprecated_loss_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
...
+def update_deprecated_sparsity_to_density(config: Dict[Text, Any]) -> Dict[Text, Any]:
+    """Updates `WEIGHT_SPARSITY` to `CONNECTION_DENSITY = 1 - WEIGHT_SPARSITY`.
+
+    Args:
+        config: model configuration
+
+    Returns:
+        Updated model configuration
+    """
+    # TODO: Completely deprecate this with 3.0
```
joejuzl, line 118: I don't think we need this TODO if we have the deprecation warning below.

rasa/utils/train_utils.py
```diff
@@ -104,6 +106,28 @@ def update_deprecated_loss_type(config: Dict[Text, Any]) -> Dict[Text, Any]:
...
+    if WEIGHT_SPARSITY in config:
+        rasa.shared.utils.io.raise_deprecation_warning(
+            f"`{WEIGHT_SPARSITY}` is deprecated."
+            f"Please update your configuration file to use"
+            f"`{CONNECTION_DENSITY}` instead.",
+            warn_until_version=NEXT_MAJOR_VERSION_FOR_DEPRECATIONS,
```
joejuzl, line 124: Can we point the user to any documentation explaining this change?

joejuzl (changes requested):
Can we update the docs and also add some info to the `migration-guide.mdx`?

rasa/utils/train_utils.py
JEM-Mosig: Good point! I'll add a section in the migration docs and link it here.

data/test_response_selector_bot/config.yml
```diff
@@ -5,10 +5,10 @@ pipeline:
   - name: "CountVectorsFeaturizer"
   - name: "DIETClassifier"
     entity_recognition: False
-    epochs: 5
+    epochs: 50
```
joejuzl, line 8: Any reason we need `50`? it will make the tests a lot slower...

docs/docs/migration-guide.mdx
```diff
@@ -14,6 +14,17 @@ how you can migrate from one version to another.
 
 ### Machine Learning Components
 
+### `DIET`, `TED`, and `ResponseSelector`
```
joejuzl, line 17: Is this under `### Machine Learning Components`? if so should be `####` I guess?

joejuzl (changes requested):

data/test_response_selector_bot/config.yml
JEM-Mosig: The less training data you have, the more epochs you have to train for. At 5 you can find random seeds where the tests fail - even on main (and so they may fail if we just change the architecture slightly). We might get away with fewer epochs and a higher learning rate, but tests like tests/core/test_evaluation.py::test_retrieval_intent can fail randomly.

docs/docs/migration-guide.mdx
JEM-Mosig: 🤦 Right.

data/test_response_selector_bot/config.yml
joejuzl: Then can we not set it to a random seed that works and keep the epochs low?

JEM-Mosig: Ok, I've reduced the number of epochs and increased the learning rate for now.

JEM-Mosig: >Then can we not set it to a random seed that works and keep the epochs low?
Sorry, GitHub didn't show this until I refreshed the page. We can do this, but I would call it bad practice. If you change anything in the model, this would be like changing the random seed, and your test may suddenly fail for no reason.

JEM-Mosig: I think with the increased learning rate we might be fine. But success is never 100% guaranteed.

joejuzl: Yeah, the point of unit tests is to test the code around the model rather than the effectiveness or accuracy of the model - so the most important thing is just having stability. And we need our unit tests to be fast so it doesn't hinder development and the CI .
However I do see your point about it being liable to change easily meaning having to update the test often which is not ideal. The other alternative is to mock the model, however I like this even less as it brings us further for reality. 
I would say the best (however not perfect) option is setting the random seed and having relatively low epochs.

JEM-Mosig: Ok, then let's keep it as it is now. From an ML perspective it makes sense to have a higher learning rate here, and the test succeeds.

joejuzl (approved):
Looks good!

Issue #7971, opened 2021-02-16
Research Team Slack Channel notification on Model Regression Test Performance Drops


kedz:
**Description of Problem**:

Our NLU models are tested weekly on some standard evaluation datasets ([and soon we will also test policy models](https://github.com/RasaHQ/rasa/issues/7835#)).
While the results of these tests are logged and [can be viewed in metabase](https://metabase.rasa.com/dashboard/166) it would be nice if the research team was automatically notified in slack when a model drops in performance. This would help us make sure we don't unintentionally roll out changes that break our models.

This was originally discussed during the implementation of the NLU model regression tests however it was not implemented because training on GPUs introduces some indeterminism in the order that floating point ops are done which can lead to small but not meaningful differences in performance, even when model configs and dataset were unchanged.  [As of Tensorflow 2.3, however, we can run the GPUs in a deterministic mode](https://github.com/NVIDIA/framework-determinism#stock-tensorflow-version-23), so provided we properly set random seeds, we should be able to implement this automatic warning feature.

**Overview of the Solution**:

Regression tests will need to export the following environment variable: `TF_DETERMINISTIC_OPS=1` before calling rasa train/test ([here for example](https://github.com/RasaHQ/rasa/blob/f4af88f0c165c8898b6924b0cbdb0984b8117e63/.github/workflows/ci-model-regression-on-schedule.yml#L174)) and ensure that all model configs ([specified here](https://github.com/RasaHQ/training-data/tree/main/configs)) have their random seeds set (it looks like this is already done). This should then enable consistent outputs when nothing touching the models or datasets has changed. Notification could be implemented similarly to [this PR](https://github.com/RasaHQ/rasa/issues/7834#).

sara-tagger:
Thanks for submitting this feature request 🚀 @TyDunn will get back to you about it soon! ✨

sara-tagger:
Thanks for submitting this feature request 🚀 @joejuzl will get back to you about it soon! ✨

HotThoughts:
Hi Chris, do you mean the tests performance for the *schduled* Model Regression Test? Just want to confirm because we have two Model Regression Test workflow.

kedz:
@HotThoughts yes just the scheduled tests should trigger an alert.

PR #8027, opened 2021-02-23
Slack and GH issue integration for scheduled model regression test workflow

HotThoughts:
**Proposed changes**:
Adding a new job in the scheduled Model Regression Test workflow. When model performance drops or test failure: 
- Notify slack
- Open issue with :bug: label
- Move the issue to research's board

Right now the issues for test failure and performance drop are separate. The test failure is notified per configuration, so one issue per configuration. Unlike test failure, performance is evaluated after all configurations is completed. 
In short, the research channel will receive many notifications if multiple configuration test fail, but only one notification for performance drop.

Related to https://github.com/RasaHQ/rasa/issues/8035 and https://github.com/RasaHQ/rasa/issues/7971

An improvement for this issue would be: https://github.com/RasaHQ/rasa/issues/8268

**Current status:**
- [X] Add TF2.3 deterministic mode
- [X] Detect performance drop by matching if the performance description (a table) has anything negative values inside parentheses. 
- [X] Create an issue for it
- [X] Randomly pick a team member to work on the issue
- [X] Add label to the issue
- [X] Put details into the issue body
- [X] Move the issue to the project inbox column as a card
        Example issue: https://github.com/RasaHQ/rasa/issues/8191

        

**TODOs**
- [x] Remove testing lines


HotThoughts:
@tczekajlo this task is almost there. We need a service account to move the issue to the project inbox column via GitHub REST API.

.github/workflows/ci-model-regression-on-schedule.yml
```diff
@@ -257,6 +261,148 @@ jobs:
...
+          curl \
+            -X POST \
+            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/$GITHUB_REPOSITORY/issues/${ISSUE_NUMBER}/labels \
+            -d '{"labels":["${{ env.LABEL }}"]}'
+
+      - name: Assign Issue to Project Column
+        run: |
+          ISSUE_ID=$(echo '${{ steps.open_issue.outputs.issue }}' | jq '.id')
```
tczekajlo, line 366:
```suggestion
          ISSUE_ID=$(echo '${{ steps.open_issue.outputs.issue }}' | jq -r '.id')
```

HotThoughts:
@dakshvar22 do you have time 👀? The workflow now added TF deterministic mode, and will send slack notification + create an issue whenever a configuration test failed. Performance drops will be evaluated when all test runs completed, so there will be only one slack notification and one issue for performance drops.

As you mentioned last week multiple test failures should ideally be captured in one issue, I create a new issue #8268 for it. To achieve this, we need a bit more time to investigate.

dakshvar22 (commented):
Setting the environment variable looks good to me. I'll hold off from approving before Tomasz can review everything else.
One question -> Do the slack notifications have the link to the github issue also included in them?

.github/workflows/ci-model-regression-on-schedule.yml
```diff
@@ -175,6 +175,9 @@ jobs:
         run: |-
           poetry run rasa --version
 
+          # Run GPUs in a deterministic mode
+          export TF_DETERMINISTIC_OPS=1
```
tczekajlo, line 179: Can we move to the `env` a few lines above?

tczekajlo (approved):
I've left one small comment

.github/workflows/ci-model-regression-on-schedule.yml
HotThoughts: Moved!

HotThoughts:
@tczekajlo @dakshvar22 This PR is ready for the second iteration of reviewing. Please ✅ when you have time. 🥰

HotThoughts:
I enabled the auto-merge feature and the branch is automatically merged 😳.

PR #8030, opened 2021-02-23
Fix logging of conversation level core metrics.

kedz:
Fix logging of conversation level core metrics to only compute and log conversation level accuracy. Formerly F1-Score and precision were being printed to console but upon investigation, these metrics were not very meaningful (precision was always 1.0 if any story was completely correct and 0 otherwise). Additionally, in-data fraction was being printed twice. The new output of rasa test looks like this:

```
2021-02-23 15:54:43 INFO     rasa.core.test  - Finished collecting predictions.
2021-02-23 15:54:43 INFO     rasa.core.test  - Evaluation Results on END-TO-END level:
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Correct:          6 / 7
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Accuracy:         0.857
2021-02-23 15:54:43 INFO     rasa.core.test  - Stories report saved to results/story_report.json.
2021-02-23 15:54:43 INFO     rasa.core.test  - Evaluation Results on ACTION level:
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Correct:          33 / 35
2021-02-23 15:54:43 INFO     rasa.core.test  - 	F1-Score:         0.960
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Precision:        0.970
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Accuracy:         0.961
2021-02-23 15:54:43 INFO     rasa.core.test  - 	In-data fraction: 0.886
```
Formerly the output looked like this:

```
2021-02-23 15:54:43 INFO     rasa.core.test  - Finished collecting predictions.
2021-02-23 15:54:43 INFO     rasa.core.test  - Evaluation Results on END-TO-END level:
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Correct:          6 / 7
2021-02-23 15:54:43 INFO     rasa.core.test  - 	F1-Score:         0.923   # <-- harmonic mean of 1 and recall 
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Precision:        1.0     # <-- this is always 1 if correct > 0
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Accuracy:         0.857
2021-02-23 15:54:43 INFO     rasa.core.test  - 	In-data fraction: 0.886   # <-- printed again below in ACTION level results
2021-02-23 15:54:43 INFO     rasa.core.test  - Stories report saved to results/story_report.json.
2021-02-23 15:54:43 INFO     rasa.core.test  - Evaluation Results on ACTION level:
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Correct:          33 / 35
2021-02-23 15:54:43 INFO     rasa.core.test  - 	F1-Score:         0.960
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Precision:        0.970
2021-02-23 15:54:43 INFO     rasa.core.test  - 	Accuracy:         0.961
2021-02-23 15:54:43 INFO     rasa.core.test  - 	In-data fraction: 0.886
```

Additionally, conversation level accuracy is logged to file (i.e., `results/story_report.json`) as an additional field in a JSON dictionary:

```
{
...
 "conversation_accuracy": {"accuracy": 0.857, "correct": 6, "total": 7}
...
}
```


rasa/core/test.py
```diff
@@ -799,6 +791,16 @@ async def test(
                 targets, predictions, output_dict=True
             )
 
+            # Add conversation level accuracy to story report.
+            num_failed = len(story_evaluation.failed_stories)
+            num_correct = len(story_evaluation.successful_stories)
+            num_convs = num_failed + num_correct
+            conv_acc = num_correct / num_correct if num_correct else 0.0
```
koernerfelicia, line 798: This seems like a typo? num_correct/num_correct == 1

koernerfelicia:
@akelad would we need to update the docs for this? I couldn't find any reference to the old CL output, or the old `story_report.json` in https://rasa.com/docs/rasa/testing-your-assistant

There wouldn't be anywhere else, right?

rasa/core/test.py
```diff
@@ -702,26 +702,18 @@ async def _collect_story_predictions(
...
-        "END-TO-END" if use_e2e else "CONVERSATION",
-        report,
-        precision,
-        f1,
-        accuracy,
-        in_training_data_fraction,
-        include_report=False,
+    num_convs = len(correct_dialogues)
+    num_correct = sum(correct_dialogues)
+    accuracy = num_correct / num_convs if num_convs else 0.0
```
joejuzl, line 710: Optional:
Could do `metrics.accuracy_score([1] * len(completed_trackers), correct_dialogues)`

rasa/core/test.py
```diff
@@ -702,26 +702,18 @@ async def _collect_story_predictions(
...
+
+    logger.info(
```
joejuzl, line 712: Could we make `_log_evaluation_table` handle `None` (and skip those) so we can reuse that method?

rasa/core/test.py
joejuzl, line 798: Does this differ from the `accuracy` above?

joejuzl (changes requested):

rasa/core/test.py
koernerfelicia: Also GH won't let me comment on the particular line because it's too far out of scope, but while you're at it would you remove the unused import on line 670? (`from rasa.test import get_evaluation_metrics`)

akelad:
no need to update docs, no

rasa/core/test.py
kedz: Oof that's embarrassing! Thank you for catching! Will fix!

kedz: Yeah the accuracy above it is the percentage of correctly predicted next actions. Should I modify it to be action_accuracy to make it more clear?

kedz: I went ahead and renamed the previous accuracy to action_accuracy to make the distinction clear.

rasa/core/test.py
kedz: Yeah that sounds good! Will do.

joejuzl (approved):

koernerfelicia:
@kedz looks good! Still needs a changelog entry, though

koernerfelicia (changes requested):
Looks good! It still needs a changelog entry, though

rasa/core/test.py
```diff
@@ -795,15 +788,26 @@ async def test(
...
                 targets, predictions, output_dict=True
             )
 
+            # Add conversation level accuracy to story report.
+            num_failed = len(story_evaluation.failed_stories)
+            num_correct = len(story_evaluation.successful_stories)
+            num_convs = num_failed + num_correct
+            if num_convs:
+                conv_accuracy = num_correct / num_convs
+                report["conversation_accuracy"] = {
```
joejuzl, line 801: We should test this new data is there and correct in a unit test.

joejuzl (changes requested):
Changes look good 🚀  - but I think we should definitely test the additions to the report file. We could also test the actual log output too as it's part of the UX.

rasa/core/test.py
```diff
@@ -702,25 +702,18 @@ async def _collect_story_predictions(
...
     in_training_data_fraction = _in_training_data_fraction(action_list)
 
+    if len(correct_dialogues):
+        accuracy = accuracy_score([1] * len(correct_dialogues), correct_dialogues)
+    else:
+        accuracy = 0
+
     _log_evaluation_table(
-        [1] * len(completed_trackers),
+        [1] * len(correct_dialogues),
```
koernerfelicia, line 714: I'm a bit lost on why this parameter changed. Was it wrong before? It doesn't look like the logic that uses it changed

kedz: This was unintentional but yes it does not affect the logic at all. I had deleted the use of the whole _log_evaluation_table method here previously. When I put it back in, I just needed a list of 1s with length equal to the number of test stories and I had just used correct_dialogues above so I used that list (I lexically entrained myself :)). Will change it back!

rasa/core/test.py
kedz: Totally! I'll create a unit test in tests/core/test_test.py

kedz:
@koernerfelicia I don't know what to do/where is/best practices for making a changelog entry. Is there somewhere I can find that?

kedz:
@joejuzl @koernerfelicia Working on a unit test but realizing I need an Agent with a core policy. I really just need one with a rule policy to test this. Can either point to an example of setting up a simple agent in a unit test?

rasa/core/test.py
koernerfelicia: Thank you!

koernerfelicia:
@kedz ah I didn't realise this was your first time doing a changelog entry, otherwise I would've said! It's [here](https://github.com/RasaHQ/rasa/tree/main/changelog), and linked in the `to do` item under `status` in the PR template if you ever need to refer to it again.

Screenshot 2021-02-26 at 08 23 12

koernerfelicia:
>I need an Agent with a core policy.

I think you should find this in one of the `conftest` files. Or `test_agent.py` might have some more variety?

kedz:
@koernerfelicia @joejuzl ok, I added a changelog file and I added some tests for rasa.core.test, testing both that the results of rasa.core.test.test get logged to file and that _log_evaluation_table logs output to console as expected. Let me know if there is anything else. Otherwise I think its good to merge?

changelog/8000.improvment.md
```diff
@@ -0,0 +1,5 @@
+Remove console logging of conversation level F1-score and precision since these calculations were not meaningful.
+
+Add conversation level accuracy to core policy results logged to file in `story_report.json` after running `rasa test core` or `rasa test`.
+
+Add test to check that `story_report.json` contents are correct.
```
koernerfelicia, line 5: I think you can scrap this line about adding a test - just to keep it more concise.

Also, typo in the filename improv**e**ment. I'm not sure, but I think this affects towncrier's behaviour, and it won't work properly if the file isn't named correctly

tests/core/test_test.py
```diff
@@ -0,0 +1,209 @@
...
+    ],
+)
+async def test_test(
+    tmpdir_factory: pathlib.Path,
+    out_directory: pathlib.Path,
+    stories_yaml: Text,
+    expected_results: Dict[Text, Dict[Text, Any]],
+) -> None:
+
+    stories_path = tmpdir_factory.mktemp("test_rasa_core_test").join("eval_stories.yml")
```
koernerfelicia, line 107: Correct me if I'm wrong, but I think you want to clean up this file afterwards? Or is that something that happens automatically?

kedz: tmpdir_factory and tmpdir test fixtures get deleted automatically, with the caveat that the three most recent tmpdirs are kept alive in case you need to inspect the contents due to a test failing. I did clean this up slightly and moved the creation of stories_path to a test fixture so that the args to test_test are clearer.

joejuzl (changes requested):
Thanks for adding tests! A couple of comments on the test code.

tests/core/test_test.py
```diff
@@ -0,0 +1,213 @@
...
+        evts=[
+            ActionExecuted(ACTION_LISTEN_NAME),
+            UserUttered(intent={"name": "intentB"}),
+            ActionExecuted("actionB"),
+            ActionExecuted(ACTION_LISTEN_NAME),
+        ],
+        is_rule_tracker=True,
+    )
+
+    policy.train([rt1, rt2], domain, RegexInterpreter())
```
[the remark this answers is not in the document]
kedz: Oh I did not know about those. Where can I find them?

tests/core/test_test.py
```diff
@@ -0,0 +1,213 @@
...
+                    "support": 8,
+                },
+                "conversation_accuracy": {
+                    "accuracy": 2.0 / 3.0,
+                    "total": 3,
+                    "correct": 2,
+                },
+            },
+        ],
+        ["", {}],
```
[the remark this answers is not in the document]
kedz: Will do!

tests/core/test_test.py
```diff
@@ -0,0 +1,213 @@
...
+import rasa.core.test
+from rasa.shared.core.constants import ACTION_LISTEN_NAME
+from rasa.shared.core.domain import Domain
+from rasa.shared.core.events import ActionExecuted, UserUttered
+from rasa.shared.core.generator import TrackerWithCachedStates
+from rasa.shared.nlu.interpreter import RegexInterpreter
+
+
+@pytest.fixture(scope="function")
+def out_directory(tmpdir: pathlib.Path):
```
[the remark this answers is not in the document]
kedz: I guess not. I just thought it looked cleaner to have the args to test_test be more informative, e.g., out_directory instead of tmpdir, but I can change it.

tests/core/test_test.py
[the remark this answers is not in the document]
koernerfelicia: Could you use the agent below? You'd have to change the expected values, but I think that could work
https://github.com/RasaHQ/rasa/blob/main/tests/conftest.py#L127

tests/core/test_test.py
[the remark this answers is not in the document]
joejuzl: make sense - and I'm generally all for more informative names. `tmpdir` is pretty widely used around the code base for this purpose though.

tests/core/test_test.py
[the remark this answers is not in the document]
joejuzl: there are lots in `rasa/tests/conftest.py` for example `default_agent` (which is used in `tests/core/test_evaluation.py` too)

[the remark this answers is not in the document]
kedz: Thank you both!

kedz:
OK, I made the following changes to the tests:
1. Used an agent test fixture.
2. Moved tests from tests/core/test_test.py to tests/core/test_evaluation.py
3. Split test_test into two tests, one for testing a stories.yml with stories, and one for the edge case for when the stories file is empty.
4. Removed creation of file test fixtures and just used tmpdir, and made the necessary files in the test.
4. test_test is now renamed to test_story_report and test_story_report_with_empty_stories, since the test_test name is confusing in test_evaluation, where test is mapped to evaluate_stories, and the tests are really about the creation and contents of story_report.json anyway.

koernerfelicia (commented):
This looks good to me! Tests looks so much neater with the `core_agent` (and faster ofc). I won't approve yet because I'd like to see what @joejuzl has to say about the tests

joejuzl (approved):
Great testing! 🚀

koernerfelicia (approved):

koernerfelicia:
Actually @joejuzl what's good with the failing test? Does Chris need to flag the test as being a "training" test or something like that? I'm not in the loop on the splitting of CI tests

joejuzl:
> Actually @joejuzl what's good with the failing test? Does Chris need to flag the test as being a "training" test or something like that? I'm not in the loop on the splitting of CI tests

haha yes good spot, this was my change too 🤦 
You need to use the decorator `@pytest.mark.trains_model` on any test that advertently or inadvertently trains a model.
(see: https://rasa-hq.slack.com/archives/C36SS4N8M/p1614682343228200)

kedz:
Ah I was also not aware of marking tests that train! Added the marks. I'm still seeing some errors on the CI but I don't know what they relate to. Any ideas?

joejuzl (approved):

koernerfelicia:
@kedz no idea, they don't look related! I just restarted the jobs, maybe it was just a fluke?

kedz:
Looks like it worked this time! Should I go ahead and merge?

koernerfelicia:
@kedz go for it!

kedz:
Oh wow, my first merge into Rasa OS, thanks @koernerfelicia and @joejuzl for reviewing/pointing me to all the things!

koernerfelicia:
Congratulations!

PR #8031, opened 2021-02-24
Support transports other than HTTP(S) (First approach)

Veigaribo:
This implements a solution for the issue #7789.

Basically I created an `on_agent_ready` method for InputChannels, that is registered to run once the server has started.

The ideia is that, by using this method, people will be able to stablish connections to things that work in some way other than HTTP(S), because the `on_new_message` will be available to run immediately (in contrast to `blueprint`, which runs at a point where one cannot call the `on_new_message` callback yet).

Two approaches were suggested in the aforementioned issue, and this is the implementation of the first one.

An example InputChannel that utilizes this feature is available at <https://github.com/Veigaribo/example-rasa-amqp-connector/blob/master/connectors/amqp.py>

**Questions**

- Are there any tests I need to change / add? I wasn't able to find any tests for the `channel` module in specific;

**Status**:

### DIFF
--- a/rasa/core/channels/channel.py
+++ b/rasa/core/channels/channel.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import uuid
+from asyncio import AbstractEventLoop
 from sanic import Sanic, Blueprint
 from sanic.request import Request
 from typing import (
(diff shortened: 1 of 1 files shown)


sara-tagger:
Thanks for opening a draft pull request 🚀If you have any questions, you can direct them to @rctatman ✨

PR #8078, opened 2021-03-01
Avoid allocating list if we need a set

Mazyod:
**Proposed changes**:
Instead of creating a list then converting it to a set, we generate a set from the start with simpler semantics.


sara-tagger:
Thanks for submitting a pull request 🚀 @samsucik will take a look at it as soon as possible ✨

samsucik:
Looks good to me (thanks @Mazyod!) but I think @degiz should check and approve.

tmbo (approved):
Looks great - thanks a lot for this style improvement 🚀

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -240,28 +240,27 @@ def number_of_examples_per_response(self) -> Dict[Text, int]:
...
+        entity_types = {
             e.get(ENTITY_ATTRIBUTE_GROUP)
             for e in self.sorted_entities()
             if ENTITY_ATTRIBUTE_GROUP in e
-        ]
-        return set(entity_types) - {NO_ENTITY_TAG}
+        }
+        return entity_types - {NO_ENTITY_TAG}
 
     def entity_roles_groups_used(self) -> bool:
```
tmbo, line 265:
```suggestion
    def entity_roles_groups_used(self) -> bool:
    """Returns if any entity roles or groups are used anywhere in the training data."""
```

PR #8085, opened 2021-03-02
Change policy test class fixtures to use class scope

joejuzl:
**Proposed changes**:
The memory usage of the policy tests seems leaky. This is causing CI failures when the total memory usage exceeds 4GB.
By changing the fixtures in the policy test class to be of scope "class" we can reduce this memory.
It is not totally clear if there is still any leak, but it's definitely an improvement and a quick win.

Memory usage charts when running `mprof run pytest -sxvv tests/core/policies/test_ted_policy.py`

**main branch**

Before:
Screenshot 2021-03-02 at 15 51 50

After:
Screenshot 2021-03-02 at 15 59 38

**train_step branch**

Before:
Screenshot 2021-03-02 at 16 22 51

After:
Screenshot 2021-03-02 at 16 32 07


wochinge (approved):
Nice catch!

Issue #7462, opened 2020-12-05
Interactive story counter doesn't update


aleronupe:
**Rasa version**: 2.1.1

**Python version**: Python 3.7.9

**Operating system** (windows, osx, ...): Linux

**Issue**: The number of each interactive story generated does not update between executions of the `rasa interactive ` command, because the counter used at ` _write_stories_to_file` function at `rasa/training/interactive.py` does not recognize previous interactive stories generated.

**Error (including full traceback)**:

[<span class="image-wrap" style="">![](https://user-images.githubusercontent.com/37146204/101261832-ac09d380-3718-11eb-9d4d-00526a205df5.png)</span>](https://user-images.githubusercontent.com/37146204/101261832-ac09d380-3718-11eb-9d4d-00526a205df5.png)

**Command or request that led to error**:

 rasa interactive

sara-tagger:
Thanks for raising this issue, @kalkbrennerei will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

kalkbrennerei:
Hi @aleronupe! Thanks for bringing this up and also for the details! ✨

rafaelteodosio:
Hi, I would like to know if someone is already working on this issue and if it still relevant

PR #8086, opened 2021-03-02
Fix Interactive story counter

rafaelteodosio:
**Proposed changes**:
- Fix bug Interactive story counter doesn't update
- Resolve [#7462](https://github.com/RasaHQ/rasa/issues/7462)

**Status**:


rasa/core/training/interactive.py
```diff
@@ -816,10 +816,15 @@ def _write_stories_to_file(
     else:
         append_write = "w"  # make a new file if not
 
+    with open(
+        export_story_path, "r", encoding=rasa.shared.utils.io.DEFAULT_ENCODING
+    ) as f:
+        i = sum(1 for s in f if 'interactive_story_' in s) + 1  # count the times string 'interactive story' occur in file
```
tmbo, line 822: let's make `interactive_story_ ` a constant to ensure it stays the same between this line and L831

tmbo, line 822: while we are here, let's give that `i` a more expressive name

rasa/core/training/interactive.py
```diff
@@ -816,10 +816,15 @@ def _write_stories_to_file(
     else:
         append_write = "w"  # make a new file if not
 
+    with open(
```
tmbo, line 819: do you mind adding a test to check this will work in the future?

tmbo (changes requested):
Thanks a lot for the fix on the interactive learning dumping. Do you mind taking a look at adding a test and a changelog entry? Let me know if you need pointers 🙂

rafaelteodosio:
Ok @tmbo , @joaoCeilandia and I will make the requested changes, thanks for your time.

sara-tagger:

rafaelteodosio:
@tmbo We made the requested changes, our only doubt is about the Tests, we don't know if it's right 😅

joaoCeilandia:
I don't know if we could do this, but can i request your review here @melindaloubser1? We need this for a subject of our college.

melindaloubser1:
I think since @tmbo  has been following this PR, it's best if he does a final review of it. It needs a test to be written as well.

joaoCeilandia:
Ok, thanks for your time and support @melindaloubser1 .

rasa/core/training/interactive.py
joaoCeilandia: @rafaelteodosio and i don't know how make the requested test. @tmbo could you give us some advices?

joejuzl: Look here `tests/core/training/test_interactive.py`. You can write a test that exports twice and assert that the correct count is used for both.

joejuzl (changes requested):
Hey @rafaelteodosio - any progress with this?

Issue #8092, opened 2021-03-03
python-socketio and python-engineio versions are incompatible


adrianvisovan:
**Rasa version**: 2.3.4

**Socket.IO Javacsript client version**: 3.1.2

**Operating system**: osx

**Issue**: Connecting to Rasa 2.3.4 using socket.io is not possible due to the versions incompatibility of python-socketio and python-engineio.

According to the official [python-socketio documentation](https://python-socketio.readthedocs.io/en/latest/intro.html#version-compatibility), if the version of the python-socketio is 5.X, then the version of python-engineio should be 4.X.

Solution could be to increment the version of python-engineio dependency to at least 4.0.0

**Error**:
```
message async handler error
Traceback (most recent call last):
  File "/opt/venv/lib/python3.8/site-packages/engineio/asyncio_server.py", line 423, in _trigger_event
    ret = await self.handlers[event](*args)
  File "/opt/venv/lib/python3.8/site-packages/socketio/asyncio_server.py", line 524, in _handle_eio_message
    await self._handle_connect(eio_sid, pkt.namespace)
  File "/opt/venv/lib/python3.8/site-packages/socketio/asyncio_server.py", line 408, in _handle_connect
    sid = self.manager.connect(eio_sid, namespace)
  File "/opt/venv/lib/python3.8/site-packages/socketio/base_manager.py", line 46, in connect
    sid = self.server.eio.generate_id()
AttributeError: 'AsyncServer' object has no attribute 'generate_id'
```

**Command or request that led to error**:
```
Connecting to Rasa 2.3.4 using socketio from the Javascript socket.io-client:

io('http://localhost:5005', {
      path: '/socket.io/',
      transports: ['websocket']
    })

```

**Content of configuration file (credentials.yml)** 
```yml
socketio:
  user_message_evt: user_uttered
  bot_message_evt: bot_uttered
  session_persistence: false

```

sara-tagger:
Thanks for raising this issue, @samsucik will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

kietdinh:
Is this fixed in 2.4.0?

nmhjoyal:
> Is this fixed in 2.4.0?

Also awaiting a fix for this. Our project uses pip-tools which will not allow us to upgrade Rasa to latest version until the dependencies are compatible

[...]

PR #8099, opened 2021-03-03
Bump python-engineio

pheel:
Fixes https://github.com/RasaHQ/rasa/issues/8092


sara-tagger:

pyproject.toml
```diff
@@ -101,8 +101,8 @@ multidict = "^4.6"
 aiohttp = "~3.6"
 questionary = "~1.5.1"
 prompt-toolkit = "^2.0"
-python-socketio = ">=4.4,<6"
-python-engineio = ">=3.11,<3.14"
+python-socketio = ">=5,<6"
```
joejuzl, line 104: is there a specific reason we need to change the `python-socketio` version range?

pyproject.toml
```diff
@@ -101,8 +101,8 @@ multidict = "^4.6"
...
+python-engineio = ">=4,<5"
```
joejuzl, line 105: is there a specific reason we have to increase the minimum bound?

joejuzl (changes requested):

pyproject.toml
pheel: Because of `>=4.4,<6`, `python-socketio` accidentally got bumped to `5`, which is incompatible with `python-engineio >=3.11,<3.14`. `python-socketio` and `python-engineio` need to be lockstep, as explained by rgstephens here: https://github.com/botfront/rasa-webchat/issues/344

joejuzl: Sure, but can't poetry handle finding compatible versions if there exists a valid overlap?

pheel: That would have been the hope, but there's no denying something went wrong [here](https://github.com/RasaHQ/rasa/pull/7479).

pheel: For some reason, `poetry` seems to have a leaning for 5.0.0 for `python-socketio`. [Correct deps are defined from 5.0.1.](https://github.com/miguelgrinberg/python-socketio/commit/d2bb2b12e536036e34d94a1ad87e0d48a1a504a8). So there's another potential solution.

joejuzl: ok I see 👍 
Would it be possible to just exclude `python-socketio 5.0.0`? 
We would like to be able to keep a wider range.

tmbo:
fixed as part of #8343

PR #8103, opened 2021-03-03
Added JWT Authentication to Socket.IO Channel

b-quachtran:
**Proposed changes**:
- Added JWT Authentication to Socket.IO Channel
- Exposed `jwt_key` and `jwt_method` parameter inputs in credentials.yml for the Socket channel


b-quachtran:
@wochinge Does this approach to JWT authentication in the Socket channel make sense? (I'm still in the process of creating test cases, but just want to get a sanity check before I keep going).

rasa/core/channels/socketio.py
```diff
@@ -175,8 +201,17 @@ async def health(_: Request) -> HTTPResponse:
             return response.json({"status": "ok"})
 
         @sio.on("connect", namespace=self.namespace)
-        async def connect(sid: Text, _) -> None:
-            logger.debug(f"User {sid} connected to socketIO endpoint.")
+        async def connect(sid: Text, environ: Dict) -> None:
```
wochinge, line 204: according the docs there is also an `auth` parameter (https://python-socketio.readthedocs.io/en/latest/server.html#defining-event-handlers) which seems to be the better fit:

wochinge (commented):
I think the approach is good! Would probably only try to deduplicate the code a bit in respects to the current `rasa` channel implementation

rasa/core/channels/rasa_chat.py
```diff
@@ -7,11 +7,9 @@
 import jwt
 import jwt.exceptions
 
-from rasa.core import constants
-from rasa.core.channels.channel import InputChannel
+from rasa.core.channels.channel import InputChannel, _decode_jwt
```
wochinge, line 10: can please reference functions by module? And in this case the function should also made public

rasa/core/channels/socketio.py
```diff
@@ -177,8 +191,20 @@ async def health(_: Request) -> HTTPResponse:
...
+                jwt_payload = None
+                if auth and auth.get("token"):
+                    jwt_payload = _decode_bearer_token(
+                        auth.get("token"), self.jwt_key, self.jwt_algorithm,
+                    )
+
+                if jwt_payload:
+                    logger.debug(f"User {sid} connected to socketIO endpoint.")
+                else:
+                    return False
```
wochinge, line 205: return `False` and the type annotation says `-> None`?

tests/core/test_channels.py
```diff
@@ -369,6 +370,75 @@ def test_socketio_channel():
     assert routes_list["handle_request"].startswith("/socket.io")
 
 
+async def test_socketio_channel_jwt_authentication():
+    import jwt
+    from rasa.core.channels.channel import _decode_bearer_token
+    from rasa.core.channels.socketio import SocketIOInput
```
wochinge, line 376: why do we need local imports? We should always use global imports except to avoid cyclic imports or performance reasons

rasa/core/channels/socketio.py
b-quachtran: Ah sorry, being sloppy

tests/core/test_channels.py
```diff
@@ -369,6 +372,71 @@ def test_socketio_channel():
...
+        bot_message_evt="bot_uttered",
+        # socket.io namespace to use for the messages
+        namespace=None,
+        # public key for JWT methods
+        jwt_key=public_key,
+        # method used for the signature of the JWT authentication payload
+        jwt_method=jwt_algorithm,
+    )
+
+    _ = rasa.core.run.configure_app([input_channel], port=5004)
```
wochinge, line 395: is this actually needed?

rasa/core/channels/channel.py
```diff
@@ -158,12 +162,30 @@ def get_metadata(self, request: Request) -> Optional[Dict[Text, Any]]:
...
+        self, bearer_token: Text, jwt_key: Text, jwt_algorithm: Text
+    ) -> Optional[Dict]:
+        # noinspection PyBroadException
+        try:
+            return self._decode_jwt(bearer_token, jwt_key, jwt_algorithm)
+        except jwt.exceptions.InvalidSignatureError:
+            logger.error("JWT public key invalid.")
+        except Exception:
+            logger.exception("Failed to decode bearer token.")
+
```
wochinge, line 181: why not making this public? both functions don't rely on any object state so I don't see why we should make the class methods. btw, Pycharm also gives a warning because we overrride `_decode_bearer_token` in `RasaChat` with a differrent signature

tests/core/test_channels.py
b-quachtran: Oh, nope it isn't

rasa/core/channels/channel.py
b-quachtran: Not sure what I was thinking, agreed making it public and yeah it doesn't rely on state at all

rasa/core/channels/channel.py
```diff
@@ -159,11 +163,50 @@ def get_metadata(self, request: Request) -> Optional[Dict[Text, Any]]:
...
+        bearer_token: Encoded Bearer token
+        jwt_key: Public JWT key for decoding the Bearer token
+        jwt_algorithm: JWT algorithm used for decoding the Bearer token
+
+    Returns:
+        `Dict` containing the decoded payload if successful or `None` if unsuccessful
+    """
+    # noinspection PyBroadException
+    try:
+        return _decode_jwt(bearer_token, jwt_key, jwt_algorithm)
```
wochinge, line 197: it's still not public 🙈

rasa/core/channels/rasa_chat.py
```diff
@@ -73,30 +72,20 @@ async def _fetch_public_key(self) -> None:
...
-            authorization_header_value, self.jwt_key, algorithms=self.jwt_algorithm
-        )
-
     async def _decode_bearer_token(self, bearer_token: Text) -> Optional[Dict]:
         if self.jwt_key is None:
             await self._fetch_public_key()
 
         try:
-            return self._decode_jwt(bearer_token)
+            return rasa.core.channels.channel._decode_jwt(
```
wochinge, line 80: it's still not public 🙈

rasa/core/channels/socketio.py
```diff
@@ -177,8 +191,22 @@ async def health(_: Request) -> HTTPResponse:
...
+                jwt_payload = None
+                if auth and auth.get("token"):
+                    jwt_payload = rasa.core.channels.channel._decode_bearer_token(
+                        auth.get("token"), self.jwt_key, self.jwt_algorithm,
+                    )
+
+                if jwt_payload:
+                    logger.debug(f"User {sid} connected to socketIO endpoint.")
+                else:
+                    return False
```
wochinge, line 207: @b-quachtran Shouldn't this call disconnect instead?

wochinge: ok, got it from the docs 👍🏻

wochinge (approved):

b-quachtran:
@wochinge Thanks a ton for taking care of this! 🙏 🙏

rasa/core/channels/rasa_chat.py
b-quachtran: Ugh yeah I was blanking, I need a re-fresher on the engineering guidelines🤦

PR #6372, opened 2020-08-10
yaml support for train endpoint

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/6354

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

Issue #6354, opened 2020-08-06
Fix `/train` endpoint (and potentially other server endpoints) for new data format


akelad:
With the switch to YAML for 2.0, we forgot to update the `/model/train` endpoint in the server to be able to handle that, and so the tests are failing/hanging in the PR #6352

It still writes the files to markdown files, that should be yaml now and also the concept of nlu vs stories files doesn't exist anymore: https://github.com/RasaHQ/rasa/blob/master/rasa/server.py#L746

I haven't checked if there's other endpoints that incorrectly handle that, so we may have to address that too

wochinge:
I'd add another key `rules` [here](https://github.com/RasaHQ/rasa/blob/5bc665140d7389a40f688446941b7c1b5e6ab93c/rasa/server.py#L776) and change everything to `yaml`. We can also check the the [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) and change to `md` if it says markdown. If we decide to drop `md` support it's easy too remove (and not a lot of engineering overhead from our side now). 

@akelad You mentioned
> and yes we can probably change to yaml but also the concept of stories vs nlu files doesn’t exist anymore

Is the additional `rules` key solution fine? I wouldn't merge the `rules` and `stories` keys into a `core` key because they are still somewhat separated in yaml (`stories: ` vs `rules: `)

akelad:
@wochinge i would avoid using the term "core" - why can't we just pass it all into one `training_data_files` key for example?

akelad:
passing in `rules:` and `stories` doesn't make sense to me since you can have those in one file

wochinge:
> passing in rules: and stories doesn't make sense to me since you can have those in one file

True. I assumed that when using the API (especially from Rasa X), we'd have everything separate anyway. Putting all in one ( core (stories+rules), nlu, domain) makes more sense when thinking from a disk perspective 👍 

On the other hand: If you think of YAML as JSON then it's already separated by keys now. As our training data is YAML now it also makes sense to support it's JSON representation, no? 

In addition we shouldn't simply change the API endpoint without at least support the old separated way as deprecated option. How about doing both:
- Having everything as giant string in a `training_data` key (I slightly prefer that to `training_data_files`)
- Having everything split by `rules`, `stories` and so on (maybe print a deprecation warning then)

We can add support for full JSON payloads later.

akelad:
[...]

PR #8104, opened 2021-03-04
Check domain file does not include NLU keys

chdorner:
**Proposed changes**:
In Rasa X we currently don't consider any files in the `data/` directory to be potential domain files, because we currently only check that at least one of the allowed domain keys is in the file. It happens to be that the `responses` key is allowed in NLU files, which results in X thinking that an NLU file with responses is a domain file.
The proposed changes will in addition to look for allowed keys look for explicitly disallowed keys (i.e. `nlu`).

Also tagging @wochinge for review to put this change on the radar of the Enable squad.

As discussed in Slack, this should be released as a micro release as soon as its merged.

- ~~[ ] updated the documentation~~

### DIFF
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -64,6 +64,12 @@ ALL_DOMAIN_KEYS = [
     KEY_E2E_ACTIONS,
 ]
 
+DISALLOWED_DOMAIN_KEYS = [
+    "nlu",
+    "stories",
+    "rules",
+]
+
 PREV_PREFIX = "prev_"
 
 # State is a dictionary with keys (USER, PREVIOUS_ACTION, SLOTS, ACTIVE_LOOP)
@@ -1588,6 +1594,9 @@ class Domain:
         try:
             content = rasa.shared.utils.io.read_yaml_file(filename)
         except (ValueError, YamlSyntaxException):
+            return False
+
+        if any(key in content for key in DISALLOWED_DOMAIN_KEYS):
             return False
 
         return any(key in content for key in ALL_DOMAIN_KEYS)
--- a/changelog/8104.improvement.md
+++ b/changelog/8104.improvement.md
@@ -1,0 +1,1 @@
+In addition to check for the existence of allowed keys when looking for valid domain files, check for absence of explicitly disallowed keys. 
--- a/tests/shared/core/test_domain.py
+++ b/tests/shared/core/test_domain.py
@@ -34,6 +34,7 @@ from rasa.shared.core.domain import (
     Domain,
     KEY_FORMS,
     KEY_E2E_ACTIONS,
+    DISALLOWED_DOMAIN_KEYS,
 )
 from rasa.shared.core.trackers import DialogueStateTracker
 from rasa.shared.core.events import ActionExecuted, SlotSet, UserUttered
@@ -1299,6 +1300,23 @@ def test_is_valid_domain_doesnt_raise_with_valid_domain(tmpdir: Path):
         domain_path,
     )
     assert Domain.is_domain_file(domain_path)
+
+
+@pytest.mark.parametrize("disallowed_key", DISALLOWED_DOMAIN_KEYS)
+def test_is_valid_domain_false_with_nlu_file(disallowed_key, tmpdir: Path):
+    domain_path = str(tmpdir / "domain.yml")
+    rasa.shared.utils.io.write_text_file(
+        f"""
+        responses:
+          utter_greet:
+            - text: hey there!
+        {disallowed_key}:
+          - intent: greet
+            examples: |
+              - hello""",
+        domain_path,
+    )
+    assert not Domain.is_domain_file(domain_path)
 
 
 def test_is_valid_domain_doesnt_raise_with_invalid_domain(tmpdir: Path):


rasa/shared/core/domain.py
```diff
@@ -1590,6 +1596,9 @@ def is_domain_file(filename: Text) -> bool:
         except (ValueError, YamlSyntaxException):
             return False
 
+        if any(key in content for key in DISALLOWED_DOMAIN_KEYS):
```
wochinge, line 1599: isn't this quite dangerous? What if want to put everything in one file, e.g. mix stories with the domain file? That should in my opinion still work.

chdorner: I was basing it off of the docs [here](https://rasa.com/docs/rasa/domain) where we have a full example of a domain file which doesn't include any of the ones that are not allowed in this PR.

chdorner: Assuming that the keys that are not listed on that page are generally not allowed in a domain file, is that assumption not correct?

wochinge: Afaik everything can be merged / split with the Rasa Open Source 2.0 YAML format as you want (cc @degiz ). For me the problem is how Rasa X handles these new yaml format capabilities and that we haven't resolved this properly. Not a huge fan of fixing things in Rasa Open Source for that.

degiz: > everything can be merged / split with the Rasa Open Source 2.0 YAML 

not really, domain still stays aside. The only thing from `domain` that's shared with `nlu` are the `responses`.

degiz: We clearly made this distinction, domain stays separate and **has** to be specified with a `-d` flag with all rasa cli related commands.

degiz: More generally, we have:
- training data files (nlu, stories, rules, responses)
- domain files
- configuration files (endpoints, config, credentials)

wochinge: > We clearly made this distinction

Thanks for clarifying 🤔 Did we write that down somewhere?

What about the server endpoints which accept YAML in Rasa? In that case all data is put in one big payload, no? Btw that was [my PR](https://github.com/RasaHQ/rasa/pull/6372/files) back then and some related discussion [here](https://github.com/RasaHQ/rasa/issues/6354#issuecomment-671328929) (Tom intervened to explicitly not split things 😄 ). 

If Rasa X can't deal with mixed files and Rasa Open Source currently can - shouldn't we rather add support for this to Rasa X then?

wochinge: Also in the [docs here](https://rasa.com/docs/rasa/chitchat-faqs#5-defining-the-responses) we explictly recommend putting responses in the domain file 😬

chdorner: > shouldn't we rather add support for this to Rasa X then?

That's the next step (see [this scope doc](https://www.notion.so/rasa/Rasa-X-dumping-mixed-training-files-d02e7973f88942ec912583f5bd1aa29a)).

But even if Rasa X is able to write mixed files, I don't think there's a way for Rasa X to distinguish between a domain file and an NLU file with responses at the moment, which means that such a file would show up on the domain screen.

wochinge: But shouldn't these responses show up in the domain screen? Just the other NLU stuff shouldn't show up, no?

chdorner: > But shouldn't these responses show up in the domain screen? Just the other NLU stuff shouldn't show up, no?

I guess that's another question. The responses from NLU files should show up in the Responses screen, not sure if I have an answer if they should show up in the domain screen.

chdorner: > Also in the docs here we explictly recommend putting responses in the domain file 😬

Another possible solution for this problem ultimately [this bug](https://github.com/RasaHQ/rasa-x/issues/4252) could also be to not allow responses in NLU files anymore.
I think what's missing at the moment is a clear direction of what kind of data is allowed in which files, or even a documented decision stating that a user can mix and match to their heart's content.

wochinge: > could also be to not allow responses in NLU files anymore.

This seems all a bit fragile to me to be honest. What is the current problem with files with `responses` being recognized as domain files? Is it because _1_ file is referenced by _two_ Rasa X entity objects which then override each other during the dump?

chdorner: I think we're honing in too fast on the "X overrides mixed data" problem, which in my opinion is a separate issue and wouldn't even solve [this bug](https://github.com/RasaHQ/rasa-x/issues/4252).

The more interesting issue to discuss about is rather that:
* X (currently) needs to differentiate between domain and NLU files (amongst others), and it needs to do that since we differentiate between a mix of files and object types in the UI
  *  i.e. Domain screen is for the domain file
  * Responses, "NLU data", Responses, etc. screens are around object types
* An NLU file that has `responses` should not be identified as a domain file

The overwriting of the file content is rather a symptom of the assumption that any file with at least one of the allowed domain file keys is in fact a domain file, which I think is an assumption that is worth challenging from a product perspective.

chdorner:
Closing this as we're going back to the drawing board when @TyDunn is back next week and look at it together with other issues around file handling.

Issue #8100, opened 2021-03-03
More Detailed Lock Store Log Messages


b-quachtran:
**Description of Problem**:

When Lock Store can acquire a lock, it shows a `Failed to acquire lock message` which could be indicative of a number of things (message still being processed or an actual error). 

**Overview of the Solution**:

It would be nice to add some more descriptions to this message, maybe something like "Failed to acquire lock because X message is currently being processed"

**Definition of Done**:
- [ ] Log messages updated for LockStore

PR #8105, opened 2021-03-04
improve Lockstore queuing log message

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/8100

### DIFF
--- a/rasa/core/lock_store.py
+++ b/rasa/core/lock_store.py
@@ -123,9 +123,14 @@ class LockStore:
                 logger.debug(f"Acquired lock for conversation '{conversation_id}'.")
                 return lock
 
+            items_before_this = ticket - (lock.now_serving or 0)
+            title = "items" if items_before_this > 1 else "item"
+
             logger.debug(
-                f"Failed to acquire lock for conversation ID '{conversation_id}'. "
-                f"Retrying..."
+                f"Failed to acquire lock for conversation ID '{conversation_id}' "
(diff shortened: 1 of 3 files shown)


rasa/core/lock_store.py
```diff
@@ -123,9 +123,14 @@ async def _acquire_lock(
                 logger.debug(f"Acquired lock for conversation '{conversation_id}'.")
                 return lock
 
+            items_before_this = ticket - (lock.now_serving or 0)
+            title = "items" if items_before_this > 1 else "item"
```
joejuzl, line 127: We could use https://pypi.org/project/inflect/ ? 
I think I suggested it once before and was told it was overkill for one time, but have seen this kind of thing a lot in the code base since.

wochinge: What's your experience regarding the speed of it? Loading the `engine` everytime we log this debug message seems quite expensive. I agree with you in general but for `debug` log messages I'm not sure if it's the right thing. I'm even inclined to do something like `because {items_before_this} other item(s) for ....`

b-quachtran:
@wochinge Yep, will review today!

rasa/core/lock_store.py
wochinge: @joejuzl What do you think?

b-quachtran (approved):
🚀

PR #8109, opened 2021-03-04
Add all missing dependencies to Rasa Playground `onLiveCodeChange` callback function 

alwx:
**Proposed changes**:
- fix https://github.com/RasaHQ/playground/issues/129


gausie (approved):
There is a slight change you can make here to reduce dependencies and thus reduce redefinitions of `onLiveCodeChange` and thus reduce rerenders. I would do it (I think! I haven't tested this) but it's up to you.

```ts
if (chatState === "ready") {
      clearPollingInterval();
      setChatState("needs_to_be_retrained");
      updateChatBlock();
    }
```

can be changed to something like

```ts
setChatState(cs => {
  if (cs !== "ready") return cs;
  clearPollingInterval();
  updateChatBlock();
  return "needs_to_be_retrained";
});
```

If `updateChatBlock` needs to change after `chatState` changes, you can do it in an effect.

The same as the above goes for `hasStarted`.

alwx:
To be fair, I think a big bunch of what's going on inside `prototyper.tsx` needs refactoring. I would prefer not to change much right now especially since it might break something but would definitely work on this later.

PR #8115, opened 2021-03-04
Added small Rasa-X example for Slack in Docs

koaning:
The [Receiving Messages section](https://rasa.com/docs/rasa/connectors/slack#receiving-messages) shows webhook examples for using `rasa run` as well as `ngrok` but not `rasa x`. It took me a while to figure out the correct setting there, so I figured it'd be good to add an extra sentence to the docs. 

## Before 

As a request URL enter the public url of your bot and append `/webhooks/slack/webhook`, e.g. `https://<host>/webhooks/slack/webhook` replacing `<host>` with your URL. If you are using ngrok, your url should look like `https://92832de0.ngrok.io/webhooks/slack/webhook`. 

## After 

As a request URL enter the public url of your bot and append `/webhooks/slack/webhook`, e.g. `https://<host>/webhooks/slack/webhook` replacing `<host>` with your URL. If you are using ngrok, your url should look like `https://92832de0.ngrok.io/webhooks/slack/webhook`. If you're using Rasa X, your url should look like `https://<host>/core/webhooks/slack/webhook`.


koaning:
@tmbo @TyDunn does it make sense if I add such a snippet to all connector docs? I can imagine it saves a bunch of folks some time.

TyDunn:
@koaning That sounds good to me

docs/docs/connectors/slack.mdx
```diff
@@ -115,7 +115,8 @@ your bot and tell you about new messages. If you are running locally, you can
 
    As a request URL enter the public url of your bot and append `/webhooks/slack/webhook`, e.g.
    `https://<host>/webhooks/slack/webhook` replacing `<host>` with your URL. If you
-   are using ngrok, your url should look like `https://92832de0.ngrok.io/webhooks/slack/webhook`.
+   are using ngrok, your url should look like `https://92832de0.ngrok.io/webhooks/slack/webhook`. If you're
+   using Rasa X, your url should look like `https://<host>/core/webhooks/slack/webhook`.
```
TyDunn, line 119: @koaning I believe we removed the need for `core/` in the Rasa X url cc @erohmensing

koaning: @TyDunn just to check: I did a quick-install using docker compose today and it was still required.

TyDunn: I think it will still work with `/core`, but I believe Rick changed it after Ella mentioned that we should just make it consistent, so that you don't have to use it

koaning: Ah! You're right! I just checked.  In that case, I should change the text.

koaning: It's now a *very* small change. But I suppose it's better that way. 😅

Issue #8128, opened 2021-03-05
Investigate `test_rasa_data`: All conversions take 10+ seconds


alwx:
**Rasa version**: 2.3.4 (used `main` branch)

**Rasa SDK version** (if used & relevant): 2.3.1

**Rasa X version** (if used & relevant):

**Python version**: 3.7 / 3.8

**Operating system** (windows, osx, ...): MacOS / Windows

**Issue**:
The conversions in `test_rasa_data.py` take too much time — one of the reasons is probably that we execute CLI commands in these tests, and CLI is slow in general.

There are the tests that need to be checked: 
- `test_rasa_data_convert_nlg_to_yaml`
- `test_rasa_data_convert_nlu_to_yaml`
- `test_rasa_data_convert_stories_to_yaml`

See more details about time measurements: https://docs.google.com/spreadsheets/d/1tDRD0vWaLW91W1wPj5h__dwXuXioRt9aBPYyF1n4MjI/edit#gid=57091041 

**Definition of done**:
- These tests run in less than 5 seconds

**Command or request that led to error**:
```
pytest tests/test_rasa_data.py
```

Imod7:
Also, a reason of these tests being slow might be that they are all writing to files in the disk. Maybe the use of [mock_open](https://docs.python.org/3/library/unittest.mock.html#mock-open) is relevant and can help in speed as described in an example also [here](https://stackoverflow.com/questions/20531072/writing-a-pytest-function-to-check-outputting-to-a-file-in-python).

wochinge:
This will be fixed by https://github.com/RasaHQ/rasa/pull/8141 + replacing the any policy imports in these modules with the usage of string constants (we currently only do imports to access the policy names) (see [here](https://github.com/RasaHQ/rasa/pull/8141#discussion_r591579752
) for more details.

This brings down the test duration for the module from 140 seconds to 34 seconds. No test takes longer than 2.2 seconds


<img width="972" alt="Screenshot 2021-03-10 at 16 05 19" src="https://user-images.githubusercontent.com/7667273/110650074-7f9fa080-81ba-11eb-9112-50130eb14c4a.png">
<img width="1063" alt="Screenshot 2021-03-10 at 16 05 25" src="https://user-images.githubusercontent.com/7667273/110650082-81696400-81ba-11eb-9079-8df5096c528c.png">

cc @RasaHQ/enable-squad

wochinge:
[...]

PR #8141, opened 2021-03-08
Improved startup time

### LINKED ISSUE #4280: Improve Rasa CLI start-up time
**Rasa version**: `master`

**Python version**: `3.5`, `3.6`, `3.7`

**Operating system** (windows, osx, ...): GNU/Linux, macOS, Windows

**Issue**:

After a fresh `rasa` install, the `rasa` CLI can take up to 1.5 seconds (on my system) to show any output, even when no action has been specified (e.g. `rasa --help`). For workflows that involve running the `rasa` command various times, this can be an inconvenience. We should aim to lower the start-up time for the CLI to under 1 second, or another set number that users will be comfortable with.

Related to: https://github.com/RasaHQ/rasa/pull/4248 and https://github.com/RasaHQ/rasa/issues/3010.

Tools that could be useful:
- https://docs.python.org/3.7/library/profile.html
- https://github.com/jrfonseca/gprof2dot
- https://github.com/joerick/pyinstrument

tmbo:
**Proposed changes**:
- avoid importing tensorflow where possible in CLI (tensorflow has an import time of ~4s)
- command to profile imports locally (requires https://github.com/nschloe/tuna to be installed and python >= 3.7): 
  ```
  python -X importtime -m rasa.__main__ --help  2> import.log
  tuna import.log
  ```
- fixes #4280 
- added a test to avoid a regression (checks if tensorflow gets imported on CLI start)


tmbo:
changes make the commands `--help` & `init` a lot quicker, but I don't think it makes sense to merge anything without a change as it is very easy to break this again.

rasa/cli/data.py
```diff
@@ -531,6 +536,9 @@ def _warn_about_manual_forms_migration() -> None:
 
 
 def _assert_nlu_pipeline_given(config: Dict, policy_names: List[Text]) -> None:
+    from rasa.core.policies.fallback import FallbackPolicy
```
wochinge, line 539: Good timing with the review 🚀 I'm currently working on [this issue](https://github.com/RasaHQ/rasa/issues/8128) and I would even go so far to drop the policy imports completely in favor of using constants in `rasa.shared.core` for them. For `test_rasa_data` this means:

* current `main`: ~140 seconds
* this branch: ~85 seconds
* this branch + dropped policy imports in `rasa.core.config` and `rasa.cli.data`: ~35 seconds

In my opinion up to you if you want to do the extra step right away or I do a followup PR

rasa/telemetry.py
```diff
@@ -482,7 +481,6 @@ def _default_context_fields() -> Dict[Text, Any]:
             "directory": _hash_directory_path(os.getcwd()),
             "python": sys.version.split(" ")[0],
             "rasa_open_source": rasa.__version__,
-            "gpu": len(tf.config.list_physical_devices("GPU")),
```
wochinge, line 485: this also requires updating the example in `telemetry.mdx`

rasa/run.py
```diff
@@ -12,50 +12,8 @@
 if typing.TYPE_CHECKING:
```
wochinge, line 12: there are a couple of unused imports left

rasa/__init__.py
```diff
@@ -1,12 +1,148 @@
...
 __version__ = version.__version__
 
-from rasa.run import run
-from rasa.train import train
-from rasa.test import test
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
+
+
+def run(
```
wochinge, line 19: Should we add a comment / module docstring which explains why we pulled out this functionality? This would avoid that somebody adds stuff without considering the import behavior to this module in the future.

On another note: I'd slighly prefer to have this behavior in a named module and just import in `__init__` from there. This would make the code more discoverable and align with best practices re the `__init__.py` usage.

wochinge (commented):
Thanks for tackling this! This is both amazing for the users and our CI speed 🚀 

Regarding the regression test: How about testing for the tensorflow import instead of a certain timing? E.g. you could write a script which 
* imports `rasa`
* sets `sys.argv` according to the commands you want to test
* runs the command (maybe with mocked behavior)
* assert that tensorflow isn't in the imported modules

rasa/cli/data.py
tmbo: yes that is a good point and a great idea :+1: added it to this PR

rasa/telemetry.py
tmbo: :+1:

rasa/__init__.py
tmbo: yes I was thinking about this as well and you are right better to move them to a separate file

tmbo: moved them to `api.py` and added a blurp about the imports

tmbo:
ready for another review 😉

rasa/__init__.py
wochinge: nice, thank you!

rasa/api.py
```diff
@@ -0,0 +1,155 @@
...
+    import asyncio
+
+
+def run(
+    model: "Text",
+    endpoints: "Text",
+    connector: "Text" = None,
+    credentials: "Text" = None,
+    **kwargs: "Dict[Text, Any]",
+):
```
wochinge, line 24:
```suggestion
) -> NoReturn:
```

rasa/core/config.py
```diff
@@ -13,6 +13,11 @@
 from rasa.shared.core.constants import (
     ACTION_DEFAULT_FALLBACK_NAME,
     ACTION_TWO_STAGE_FALLBACK_NAME,
+    POLICY_NAME_RULE,
+    POLICY_NAME_FALLBACK,
+    POLICY_NAME_MAPPING,
+    POLICY_NAME_TWO_STAGE_FALLBACK,
+    POLICY_NAME_FORM,
```
wochinge, line 20: unused
```suggestion
```

rasa/shared/core/constants.py
```diff
@@ -80,3 +80,11 @@
 
 USE_TEXT_FOR_FEATURIZATION = "use_text_for_featurization"
 ENTITY_LABEL_SEPARATOR = "#"
+
+# if you add more policy names, make sure to add a test as well to ensure
+# that the name and the policy stay in sync
```
wochinge, line 85: 🙌🏻

rasa/core/config.py
```diff
@@ -25,10 +30,6 @@
 
 import rasa.shared.utils.io
 import rasa.utils.io
-from rasa.core.policies.mapping_policy import MappingPolicy
-from rasa.core.policies.rule_policy import RulePolicy
-from rasa.core.policies.fallback import FallbackPolicy
-from rasa.core.policies.two_stage_fallback import TwoStageFallbackPolicy
 from rasa.nlu.classifiers.fallback_classifier import FallbackClassifier
```
wochinge, line 33: optional: one more request: Could you please also replace this one with a constant? This gives us the massive boost for `test_rasa_data` from 84 seconds to 38 seconds.

tests/cli/test_cli.py
```diff
@@ -1,24 +1,26 @@
...
+import shutil
 from typing import Callable
-from _pytest.pytester import RunResult
+from _pytest.pytester import RunResult, Testdir
 import pytest
 import sys
 
 
-def test_cli_start(run: Callable[..., RunResult]):
+def test_cli_start_is_fast(testdir: Testdir):
```
wochinge, line 9: very nice! Thanks! (cc @alwx Not sure what's the latest regarding the `test_import_time` test, but in any case we can drop it now)

Does it make sense to also test for numpy or is that neglible? 

And can we add some instructions how to fix this test in case it's failing? E.g. `If this is failing this means, that a simple "rasa --help" commands imports tensorflow which makes our CLI extremely slow. In case this test is failing you've very likely added a global import of "tensorflow" which should be avoided. Consider making this import (or the import of its parent module) a local import."`

rasa/__init__.py
tmbo: doesn't work though - windows tests are failing. looks like we have name clash now (`run` module and function at `rasa.run` - thinking about it I am not even sure why this worked before...)

tests/shared/core/test_constants.py
```diff
@@ -0,0 +1,28 @@
...
+    "name_in_constant, policy_class",
+    [
+        (POLICY_NAME_TWO_STAGE_FALLBACK, TwoStageFallbackPolicy),
+        (POLICY_NAME_FALLBACK, FallbackPolicy),
+        (POLICY_NAME_MAPPING, MappingPolicy),
+        (POLICY_NAME_FORM, FormPolicy),
+        (POLICY_NAME_RULE, RulePolicy),
+    ],
+)
+def test_policy_names(name_in_constant, policy_class):
```
wochinge, line 27:
```suggestion
def test_policy_names(name_in_constant: Text, policy_class: Type):
```

wochinge (approved):
Thanks for wrapping this up so quickly!  Let a few more minor comments but good to go from my side!

rasa/__init__.py
wochinge: Yeah it wasn't the best decision back then to name the modules the same as the most important function 🤦🏻

wochinge:
@tmbo How severe is the problem with clashing imports? Would be nice to get it in the release today 👀

tests/cli/test_cli.py
tmbo: will add the instructions, we can check for numpy in a separate issue - didn't look into it (and can't do it now)

tmbo:
@wochinge I made the other changes, but I can't look into the import issue today, my suspicion is that it worked beforehand because had 
```
from rasa.train import train
```
in `rasa.__init__` which populated the global state FIRST with `rasa.train` (module) and AFTERWARDS with `rasa.train()` (function) - due to the evaluation order.

Now, we are not importing `rasa.train` (module) in init anymore, which means that as soon as we do that for the first time, it apparently overwrites the defined function (and apparently windows is using a different resolution logic here?)

state on master:
Bildschirmfoto 2021-03-11 um 11 24 43

state on this branch:
Bildschirmfoto 2021-03-11 um 11 25 04

tmbo:
If you have time to investigate or try moving this around I am happy to help but I won't have enough time today to dig deeper myself 😞

wochinge:
Thanks for the heads up. I'll have a short look but it doesn't seem like something we should rush just to get it in the release. Especially as it might require change to our Python API.

tmbo:
I mean there is one quick solution we could try: move the functions back into the original modules and move all the imports in there into the functions (or at least all rasa ones). That will likely work and we could revisit this later.

wochinge:
> I mean there is one quick solution we could try: move the functions back into the original modules and move all the imports in there into the functions (or at least all rasa ones). That will likely work and we could revisit this later.

This seems way harder to maintain. I'm in favor of having these things explict. In my opinion it would be fine to rename the modules (e.g. `train` -> `model_training`, `test` -> `model_testing` and `create_agent` from `run` can go to `server.py`). They are not part of the official Python API as the official Python API is to use `rasa.train(...)` etc. so in my opinion this would be ok to do, what do you think?

tmbo:
well, if they don't appear in the docs I am good with that :+1:

wochinge:
I searched 
- `import rasa.core`
- `import rasa.nlu`
- `import rasa.test`
- `from rasa import` 
- `from rasa.core` (has matches but not related to `train`, `run`, `test`
- `from rasa.nlu`
- `from rasa.test`

and had 0 matches 🎉 So good to rename 🚀

wochinge:
I'd mention the rename in the changelog nevertheless to not leave any potential advanced Python API users in the dark.

tmbo:
yes that makes sense :+1:

tmbo:
@wochinge are you able to make these changes? Otherwise I'll put it in enable's inbox to make these additional changes.

wochinge:
Sure, will wrap it up later 👍🏻

wochinge:
What about deleted the `__init__` imports ([here](https://github.com/RasaHQ/rasa/blob/452a8cfd6338d378fb869fe999c3325d5cf7d889/rasa/core/__init__.py#L5) and [here](https://github.com/RasaHQ/rasa/blob/d39791eebf45360467cdd0d4d8b9ed6618cbba68/rasa/nlu/__init__.py#L6)) instead of renaming the `train` / `visualize` in `rasa.core` and `train` / `test` in `rasa.nlu`?

They aren't used except by us and in my opinion users should rather go through the `__init__` imports in `rasa` or `rasa.model_training`, what do you think? This also avoids needless imports this stuff.

rasa/core/agent.py
```diff
@@ -269,6 +270,33 @@ async def schedule_model_pulling(
     )
 
 
+def create_agent(model: Text, endpoints: Text = None) -> "Agent":
```
wochinge, line 273: moved from `rasa.run`

wochinge:
@tmbo Changes are ready. Do you have time for a short review? 🙌🏻

rasa/model_training.py
```diff
@@ -51,63 +51,8 @@ class TrainingResult(NamedTuple):
...
-            persist_nlu_training_data=persist_nlu_training_data,
-            core_additional_arguments=core_additional_arguments,
-            nlu_additional_arguments=nlu_additional_arguments,
-            model_to_finetune=model_to_finetune,
-            finetuning_epoch_fraction=finetuning_epoch_fraction,
-        ),
-        loop,
-    )
+# backwards compatibility
+train = rasa.train
```
tmbo, line 55:
```suggestion
```
same

rasa/model_testing.py
```diff
@@ -90,20 +90,8 @@ def test_core_models(models: List[Text], stories: Text, output: Text):
...
-):
-    if additional_arguments is None:
-        additional_arguments = {}
-
-    test_core(model, stories, output, additional_arguments)
-    rasa.utils.common.run_in_loop(
-        test_nlu(model, nlu_data, output, additional_arguments)
-    )
+# backwards compatibility
+test = rasa.test
```
tmbo, line 94:
```suggestion
```
If the module is renamed, I don't think we actually need this do we?

tmbo (commented):
can't approve since I opened the PR - but changes look good :+1:

rasa/model_training.py
wochinge: Definitely see your point. I decided to leave it in there as it seemed weird to not have the main `train` function in the `model_training` module. Coming to the module as developer and seeing all sorts of functions but not the most important one seemed not intuitive. What do you think?

tmbo: sounds good :+1: let's keep it then

Issue #7954, opened 2021-02-15
Investigate yaml schema validation


wochinge:
**Description of Problem**:
When we read YAML training data files, we validate them against a schema. [This is currently taking a lot of time](https://github.com/RasaHQ/rasa/issues/7892#issuecomment-778302866). Skipping this step e.g. brings down the training data reading times from 18 minutes to 8 minutes.

**Overview of the Solution**:
* is there a faster library? 
* Can we accelerate our current library?
* Should we add a flag to the CI to disable the schema validation?

**Definition of Done**:
- [ ] Propose next steps to speed up the YAML schema validation

wochinge:
related https://github.com/RasaHQ/rasa/issues/7977

We should also double check if the [schema extension here](https://github.com/RasaHQ/rasa/blob/b1fc73f3f238905ddd55f50e75bb494b8e9188d9/rasa/shared/utils/validation.py#L154) is needed or can be accelerated

joejuzl:
See: https://json-schema-everywhere.github.io/yaml#:~:text=While%20YAML%20has%20advanced%20features,supported%20choice%20for%20YAML%20validation.

gausie:
We should consider using jsonschema (and making the schemas public)

alwx:
We decided to drop this one because the performance is good already and because we have more important stuff to work on.

PR #8142, opened 2021-03-09
Allowing dictionary under `slot_was_set` in the stories schema

Imod7:
**Proposed changes**:
- Updating the schema file `stories.yml` to allow the presence of a dictionary under `slot_was_set` key in the `stories.yml`.

**Current investigation**:

- The validation of the `yaml schemas` happen in the function `validate_yaml_schema` at this [point](https://github.com/RasaHQ/rasa/blob/4cfc48a7ac0fa60f782076759dfe886c9f0fe12f/rasa/shared/utils/validation.py#L170) in the code where the `validate` function of the `Core class` (of `pykwalify`) is called.
- Currently, the change I did was to update the `stories schema` to allow a dictionary under `slow_was_set` by adding the type `map` with `allowempty : True` which means that it can 
[_have keys which are not present in the schema, and these can map to anything._](https://pykwalify.readthedocs.io/en/unstable/validation-rules.html#allowempty)
- I tried also `type: "any"` which also checks the stories.yml and returns it as a valid file but it feels like it undermines all the previous checks since based on the docs [_type any : will always be true no matter what the value is, even unimplemented types_](https://pykwalify.readthedocs.io/en/unstable/validation-rules.html#type). 
- I was also thinking that maybe the ideal would be to check the [stories schema file](https://github.com/RasaHQ/rasa/blob/main/rasa/shared/utils/schemas/stories.yml) in combination with the [domain schema file](https://github.com/RasaHQ/rasa/blob/main/rasa/shared/utils/schemas/domain.yml) in order to do something like _if the slot is of type any in the domain, it should pass the validation check in stories_ but I am not sure if this is possible and how in `pykwalify`. If it is possible then we could also take into account the ‘influence_conversation’ value in the domain file which can affect the check in the `stories schema`.

The solution in this PR is related and would be affected from the progress in this [ticket](https://github.com/RasaHQ/rasa/issues/7954). 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ examples/moodbot/errors.json
 examples/formbot/models*
 examples/concertbot/models*
 examples/moodbot/models*
+examples/e2ebot/results/*
 failed_stories.md
 errors.json
 pip-wheel-metadata/*
@@ -83,6 +84,6 @@ env
 .history
 rasa/segment_key
 rasa/keys
-
+/results/
 # Local Netlify folder
 .netlify
--- a/examples/e2ebot/tests/test_stories.yaml
+++ b/examples/e2ebot/tests/test_stories.yaml
@@ -1,0 +1,61 @@
+version: "2.0"
+stories:
+- story: Happy path accepts suggestion  
+  steps:
+  - user: |
+      hi
+    intent: greet
+  - action: utter_greet
+  - user: |
+      show me a place to eat  
+    intent: search_restaurant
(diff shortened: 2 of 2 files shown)


Imod7:
Thank you so much for the review @alwx ! 

Two more things : 
- If you could you also take a look at the changelog I added that would be awesome! 🙏🏽
- And also I wanted to ask you if I should add a test for this change or not? It looks like a small change to add a test but maybe I should since it affects the validation process.

PR #8151, opened 2021-03-09
Allow tests to be located anywhere, not just in `tests` directory

### LINKED ISSUE #7763: `rasa test` should search for test files recursively in all sub-directories, not just under `tests/`
**Rasa version**:
2.2.4
**Python version**:
3.7.2
OSX

**Issue**:
`rasa test` looks recursively for files prefixed with `test_`, but only in subdirectories of `tests/`. It should look recursively in all subdirectories.
`rasa test core` seems to look for these files in any subdirectory.

```
rasa test
```

alwx:
**Proposed changes**:
- fix #7763 


rasa/shared/constants.py
```diff
@@ -44,7 +44,7 @@
 
 DEFAULT_NLU_FALLBACK_INTENT_NAME = "nlu_fallback"
 
-DEFAULT_E2E_TESTS_PATH = "tests"
+DEFAULT_E2E_TESTS_PATH = "."
```
alwx, line 47: Not sure about this decision.
We don't really need this variable now since tests can be anywhere not only in `tests` directory but it might make sense to still keep this variable because it's part of `rasa.shared`.

tests/shared/test_data.py
```diff
@@ -82,7 +81,15 @@ def test_default_story_files_are_story_files():
...
 def test_default_conversation_tests_are_conversation_tests_yml(tmpdir: Path):
-    parent = tmpdir / DEFAULT_E2E_TESTS_PATH
+    e2e_path = tmpdir / "test_stories.yml"
+    e2e_story = """stories:"""
+    write_text_file(e2e_story, e2e_path)
+
+    assert rasa.shared.data.is_test_stories_file(str(e2e_path))
+
+
+def test_conversation_tests_in_a_directory(tmpdir: Path):
```
wochinge, line 91: I'd use `tmp_path`. `tmpdir` is a different type and not of type `Path`
```suggestion
def test_conversation_tests_in_a_directory(tmp_path: Path):
```

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -354,9 +368,8 @@ def is_test_stories_file(file_path: Union[Text, Path]) -> bool:
         if not rasa.shared.data.is_likely_markdown_file(file_path):
             return False
 
-        dirname = os.path.dirname(file_path)
         return (
-            DEFAULT_E2E_TESTS_PATH in dirname
+            MarkdownStoryReader._is_tests_file(file_path)
```
wochinge, line 372: optional: I'm a bit torn here. In my opinion this is rather an`improvement` than a `bug` and as markdown is deprecated I wouldn't change any of its behavior unless it's a bug. So in my opinion file also to just keep the old behavior here. This might also reduce the effort in case we have to do a follow-up PR for some reason in these parts.

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -341,6 +341,20 @@ def is_stories_file(file_path: Union[Text, Path]) -> bool:
             )
             return False
 
+    @staticmethod
+    def _is_tests_file(file_path: Text) -> bool:
+        """Check if the filename of a file at a path is a tests file.
```
wochinge, line 346:
```suggestion
        """Checks if the filename of a file at a path is a tests file.
```

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -341,6 +341,20 @@ def is_stories_file(file_path: Union[Text, Path]) -> bool:
...
+
+        Arguments:
+            file_path: path to the file
+
+        Returns:
+            `True` if the filename indicates that it's a test file, `False` otherwise.
+        """
+        return Path(file_path).name.startswith(TEST_STORIES_FILE_PREFIX) or Path(
+            file_path
+        ).name.endswith("_tests.md")
```
wochinge, line 356: mhm, I'd keep the old logic here (check if `tests` is in dirname`).

wochinge (commented):
💯

wochinge (approved):

Issue #8118, opened 2021-03-04
Split test suite in multipe GitHub jobs


wochinge:
**Description of Problem**:
The Rasa Open Source test suite is too slow. Even if we address single slow tests this doesn't mitigate the problem entirely. [As outlined in this document](https://www.notion.so/rasa/Proposal-for-tests-which-train-models-a091d698ef794b1fb8f0f89c752f7907) we should split up the suite in multiple jobs which can run in parallel.

**Overview of the Solution**:
Instead of having 2 jobs (Unit tests + Integration tests) we should have the following:

1. Unit tests: these should mainly contain the "engineering tests" *(based on [this](https://docs.google.com/spreadsheets/d/1Le0wXBbmpYs_QIziVvP5oEvDy7SP7rMsXdu-UQ-K6Q8/edit#gid=0&fvid=1829269968), roughly 11 minutes pure call time)*
2. Core featurizer test: Run everything in `tests/core/featurizers` *(based on [this](https://docs.google.com/spreadsheets/d/1Le0wXBbmpYs_QIziVvP5oEvDy7SP7rMsXdu-UQ-K6Q8/edit#gid=0&fvid=1829269968), roughly 2 minutes pure call time)*
3. Core policies: Run everything in `tests/core/policies` *(based on [this](https://docs.google.com/spreadsheets/d/1Le0wXBbmpYs_QIziVvP5oEvDy7SP7rMsXdu-UQ-K6Q8/edit#gid=0&fvid=1829269968), roughly 6 minutes pure call time)*
4. NLU featurizers: Run everything in `tests/nlu/featurizers` *(based on [this](https://docs.google.com/spreadsheets/d/1Le0wXBbmpYs_QIziVvP5oEvDy7SP7rMsXdu-UQ-K6Q8/edit#gid=0&fvid=1829269968), roughly 8.1 minutes pure call time)*
5. NLU classifiers: Run everything in `tests/nlu/classifiers`,  `tests/nlu/extractors` and `tests/nlu/selectors` *(based on [this,](https://docs.google.com/spreadsheets/d/1Le0wXBbmpYs_QIziVvP5oEvDy7SP7rMsXdu-UQ-K6Q8/edit#gid=0&fvid=1829269968) roughly 4.7 minutes pure call time)*
[...]

PR #8153, opened 2021-03-09
split test suite in multiple jobs

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/8118
- automatically add pytest `category_` markers based directory / python module
- split CI jobs based on `category_` markers
- removed deprecated `@pytest.yield_fixture` usage
- condensed `test_agent_train` and `test_moodbot_example`
- fixing coveralls behavior for parallalel tests

**Open Todos**:
- [ ] Update the required checks in GitHub as otherwise merging is blocked for everyone


.github/workflows/continous-integration.yml
```diff
@@ -285,6 +292,7 @@ jobs:
       env:
         COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         COVERALLS_SERVICE_NAME: github
+        COVERALLS_PARALLEL: true
```
wochinge, line 295: I think this is what caused the huge drops / spikes in the coverage reporting. I'm following the docs [here](https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html) which say:

> If you are running multiple jobs in parallel and want coveralls.io to merge those results, you should set COVERALLS_PARALLEL to true in your environment:
>
> `COVERALLS_PARALLEL=true coveralls`

tczekajlo: Nice catch

.github/workflows/continous-integration.yml
```diff
@@ -192,9 +192,16 @@ jobs:
     needs: [changes]
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8]
-        test: [test-non-training, test-training]
+        test:
```
wochinge, line 195: another note: I moved the `test` part up so that this shows up first when looking at the `Checks` as i think this more interesting than Python Version / OS

tests/conftest.py
```diff
@@ -60,11 +61,36 @@
 # these tests are run separately
 collect_ignore_glob = ["docs/*.py"]
 
+# Defines how tests are parallelized in the CI
+PATH_PYTEST_MARKER_MAPPINGS = {
```
joejuzl, line 65: Maybe it would be cleaner to have a conftest in each directory which specifies the marker for that path?

wochinge:
@tczekajlo Do we still need merge-pal with GitHub auto-merge? `merge-pal` takes 3.5 minutes although I didn't even attach the label to the pull request. Not sure if GitHub updates the branch automatically though 🤔

tests/conftest.py
wochinge: This is not only directories but also single files. It's possible to do markers for entire files (maybe also packages) but I decided against it because then
* everything is spread out and it's hard to see at a glance how things are mapped
* might be possible that some (new) test files / directories don't get a marker and then they aren't run

joejuzl: I was suggesting just handling the mapping via different conftests and still having the fixture that sets the marker (with a default) here. I.e. in the different conftests you set the constant `PYTEST_MARKER` if it's not already been set. You could also set it at the top of a test file.

_But_ on second thought, it probably is clearer having it all in one place. We just need to be judicious about updating it (e.g. when we get a new directory).

tests/conftest.py
```diff
@@ -478,5 +504,38 @@ def wrap_training_methods() -> None:
...
+    Returns:
+        A marker for this test based on which directory / python module the test is in.
+    """
+    test_path = Path(item.fspath).absolute()
+
+    for marker, paths_for_marker in PATH_PYTEST_MARKER_MAPPINGS.items():
+        if any(
+            path == test_path or path in test_path.parents for path in paths_for_marker
+        ):
+            return marker
```
joejuzl, line 526: Maybe it's worth checking and raising if it matches two items.

joejuzl (changes requested):
We are no longer running with `RAISE_ON_TRAIN=True` which means we are no longer enforcing that training tests are marked. I would say we either find another way to enforce it, or we remove the marker all together. WDYT?

tests/conftest.py
wochinge: done

tczekajlo (approved):
the CI part looks good to me

joejuzl (approved):
Nice! 
Maybe worth mentioning on engineering channel that people no longer have to mark tests that train 😁

PR #8161, opened 2021-03-10
Rules states should not include the initial values of slots.

### LINKED ISSUE #7450: Rule applicability depends on slot values when initial_value is used
**Rasa version**: 2.1.2

**Python version**: 3.8.5

**Operating system** (windows, osx, ...): Mac

**Issue**:
When using `initial_value` for some slot `s` whose value influences the conversation, rules that aren't written as depending on a specific value of `s` are still constructed such that they are conditioned on the exact value used for `initial_value`. Then, these rules can't be applied in situations where `s` has a different value.

Example:
```yml
intents:
- i1

slots:
  s_cat1:
    type: categorical
    values:
      - v1
      - v2
    initial_value: v1

actions:
- action_1

rules:
- rule: rule not conditioned on slot value
  steps:
  - intent: i1
  - action: action_1
  - slot_was_set:
    - s_cat1: v2

policies:
  - name: RulePolicy 
```

where `action_1` simply sets the slot to a different value:
```python
class Action1(Action):
    def name(self) -> Text:
        return "action_1"

    async def run(self, dispatcher, tracker, domain):
        return [SlotSet("s_cat1", "v2")]
```

after training and testing like this:
```bash
rasa train core
rasa test core -s tests/test_story.yml
```

leads to failed predictions in a test story because the rule couldn't be applied where the slot had a value different from the initial one:
```yml
  steps:
  - intent: i1
  - action: action_1
  - slot_was_set:
    - s_cat1: v2
  - intent: i1
  - action: action_1  # predicted: action_default_fallback
  - action: action_listen  # predicted: action_default_fallback
```

joejuzl:
**Proposed changes**:
Fixes #7450 

When calculating tracker states from rules we were including the initial value of slots.
This meant that the rule only matched when a slot had that initial value.
This change omits initial values of slots from the rule states.


tests/core/policies/test_rule_policy.py
```diff
@@ -922,6 +920,113 @@ async def test_predict_form_action_if_multiple_turns():
...
+    rule.is_rule_tracker = True
+
+    policy = RulePolicy()
+    policy.train([rule], domain, RegexInterpreter())
+
+    form_conversation = DialogueStateTracker.from_events(
+        "slot rule test",
+        evts=[
+            ActionExecuted(ACTION_LISTEN_NAME),
+            UserUttered(intent={"name": "i1"}),
```
samsucik, line 962: Are these two events needed? I think they can be removed...

tests/core/policies/test_rule_policy.py
```diff
@@ -922,6 +920,113 @@ async def test_predict_form_action_if_multiple_turns():
...
+"""
+    )
+
+    rule = DialogueStateTracker.from_events(
+        "slot rule",
+        evts=[
+            ActionExecuted(RULE_SNIPPET_ACTION_NAME),
+            SlotSet("s_cat1", "v1"),
+            UserUttered(intent={"name": "i1"}),
+            ActionExecuted(ACTION_LISTEN_NAME),
```
samsucik, line 1001: I think this event should be removed.

tests/core/policies/test_rule_policy.py
```diff
@@ -922,6 +920,113 @@ async def test_predict_form_action_if_multiple_turns():
...
+    )
+    rule.is_rule_tracker = True
+
+    policy = RulePolicy()
+    policy.train([rule], domain, RegexInterpreter())
+
+    form_conversation = DialogueStateTracker.from_events(
+        "slot rule test",
+        evts=[
+            # We are in an active form
```
samsucik, line 1016: What does this comment mean? 🤔

samsucik: Maybe it's just copy-pasted from elsewhere...

samsucik (changes requested):
Thanks @joejuzl, I know this bug was annoying to users and it's good to see it go.
 
I'm not familiar with all the code you've touched but the changes look good to me and I'm gonna trust the newly added test cases 🙂

tests/core/policies/test_rule_policy.py
joejuzl: why do you think it should be removed?

tests/core/policies/test_rule_policy.py
joejuzl: the `UserUttered` and the `action1`? They are not needed to test this fix, that is true. I was actually modelling it on your example in the issue, but maybe better to have it more succinct.

tests/core/policies/test_rule_policy.py
joejuzl: aha yes ooops - just copypasta

tests/core/policies/test_rule_policy.py
samsucik: Yep, I meant those two. I know, my example wasn't the best, but let's remove them from this test to not confuse anyone else with unnecessary stuff.

tests/core/policies/test_rule_policy.py
samsucik: Afaik `action_listen` only ever goes after some action, not after intents. It might actually be that somewhere down the line this `action_listen` gets removed/ignored as it has no real meaning/role here 🤔

tests/core/policies/test_rule_policy.py
joejuzl: ok I've reduced it to what I think is the smallest real example.

tests/core/policies/test_rule_policy.py
joejuzl: ah yes it's after the `UserUttered`! Good spot, I didn't see that.

rasa/shared/core/domain.py
```diff
@@ -1116,14 +1132,23 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
...
+        self, tracker: "DialogueStateTracker", omit_unset_slots: bool = False,
     ) -> List[State]:
-        """Array of states for each state of the trackers history."""
+        """List of states for each state of the trackers history.
+
+        Args:
+            tracker: dialog state tracker containing the dialog so far
+            omit_unset_slots: If `True` do not include the initial values of slots.
+
+        Returns A `State` for each prior tracker.
```
wochinge, line 1143:
```suggestion
        Returns: A `State` for each prior tracker.
```

rasa/shared/core/domain.py
```diff
@@ -1116,14 +1132,23 @@ def get_active_states(self, tracker: "DialogueStateTracker") -> State:
...
+        """
         return [
-            self.get_active_states(tr) for tr in tracker.generate_all_prior_trackers()
+            self.get_active_states(tr, omit_unset_slots=omit_unset_slots)
+            for tr in tracker.generate_all_prior_trackers()
         ]
 
     def slots_for_entities(self, entities: List[Dict[Text, Any]]) -> List[SlotSet]:
+        """Returns a list of `SlotSet` events from a list of entities."""
```
wochinge, line 1151: optional: I'd try to avoid duplicating specific types in the docstrings where possible 
```suggestion
        """Returns `SlotSet` events for extracted entities."""
```

wochinge (commented):
Concise change! 💯 

It feels a bit brute force which made me wonder how an ideal solution could look like: One idea could be to have a `SlotFeaturizer` interface with a specific implementation for the `RulePolicy`'s use case. The `RulePolicy` would then pass that to the `TrackerFeaturizer` which would than use that to featurize the slots. It's currently a bit unclear to me why we need `tracker.past_states` if it just calls the `Domain` 🤷🏻 Just to be clear: I'm fine with the current solution but in my opinion we should refactor as soon as we start passing a second flag down the line.

joejuzl:
> Concise change! 💯
It feels a bit brute force which made me wonder how an ideal solution could look like: One idea could be to have a SlotFeaturizer interface with a specific implementation for the RulePolicy's use case. The RulePolicy would then pass that to the TrackerFeaturizer which would than use that to featurize the slots. It's currently a bit unclear to me why we need tracker.past_states if it just calls the Domain 🤷🏻 Just to be clear: I'm fine with the current solution but in my opinion we should refactor as soon as we start passing a second flag down the line.

Yep very good points. Having ways to featurize differently based on context would nice, but yeah let's see how it develops.

wochinge:
👍🏻 

@samsucik Feel free to approve once you think Joe adressed all your changes. Engineering wise it's ready to go from my side.

samsucik (approved):
Looks good 👍

PR #8180, opened 2021-03-11
add twilio voice connector

mvielkind:
**Proposed changes**:
- Adds a new channel to interact with Twilio voice calls with some default behaviors.


wochinge:
@mvielkind Thanks for the contribution! I've a naive question: This only works for a single turn, right? So when the users responds to a question Twilio doesn't call the webhook again? The user would have to dial in for a second request, wouldn't they?

mvielkind:
@wochinge this will support a full conversation, not just single turn events. When the response is created in `build_twilio_voice_response` the response ends with Twilio's `Gather` verb. Adding the `Gather` verb instructs Twilio to listen for a user response and to send the response back to the Rasa webhook.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+        return "twilio_voice"
+
+    def blueprint(
+        self, on_new_message: Callable[[UserMessage], Awaitable[None]]
+    ) -> Blueprint:
+
+        twilio_voice_webhook = Blueprint(
+            "custom_webhook_{}".format(type(self).__name__),
+            inspect.getmodule(self).__name__,
+        )
```
wochinge, line 29:
```suggestion
        twilio_voice_webhook = Blueprint("Twilio_voice_webhook", __name__)

```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+        async def health(request: Request) -> HTTPResponse:
+            return response.json({"status": "ok"})
+
+        @twilio_voice_webhook.route("/webhook", methods=["POST"])
+        async def receive(request: Request):
+            sender_id = request.form.get("From")  # method to get sender_id
+            text = request.form.get("SpeechResult")  # method to fetch text
+            input_channel = self.name()  # method to fetch input channel
+            call_status = request.form.get("CallStatus")  # method to fetch call status
+            call_sid = request.form.get("CallSid")  # Get the call identifier
```
wochinge, line 41: To be honest I don't see the beneft of the comments as I think the code is already concise enough (maybe except for `call_sid`). I'd dump the comments or replace them with more useful ones.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+
+            collector = TwilioVoiceCollectingOutputChannel()
+
+            # Provide an initial greeting to answer the user's call.
+            if (text is None) & (call_status == "ringing"):
+                text = "hello"
```
wochinge, line 47: should this be customizable?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+                await on_new_message(
+                    UserMessage(
+                        text, collector, sender_id, input_channel=input_channel,
+                    )
+                )
+
+                # Parse the text responses and build the Twilio VoiceResponse.
+                respond_segments = []
+                for message in collector.messages:
+                    respond_segments.append(message["text"])
```
wochinge, line 60: How about overriding the `OutputChannel.send_***` methods for this? Seems preferrable to me over trying to detect the message type manually.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+                twilio_response = build_twilio_voice_response(respond_segments)
+                return response.text(str(twilio_response), content_type="text/xml")
+            # If the user doesn't respond to the previous message resend the last message.
+            elif text is None:
+                # Get last user utterance from tracker.
+                tracker = request.app.agent.tracker_store.retrieve(sender_id)
+                last_response = next(
+                    (e for e in reversed(tracker.events) if isinstance(e, BotUttered)),
+                    None,
+                )
```
wochinge, line 73: why do we need this?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+
+                # If no previous utterance found say something generic.
+                if last_response is None:
+                    last_response = "I didn't get that."
+                else:
+                    last_response = last_response.text
+
+                twilio_response = build_twilio_voice_response([last_response])
+                return response.text(str(twilio_response), content_type="text/xml")
```
wochinge, line 82: how about moving out this block instead of having it duplicate in both `if` and `elif` blocks?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+        action="/webhooks/twilio_voice/webhook",
+        actionOnEmptyResult=True,
+        speechTimeout="auto",
+    )
+
+    # Add pauses between messages.
+    # Add a listener to the last message to listen for user response.
+    for i, message in enumerate(messages):
+        if i + 1 == len(messages):
+            gather.say(message)
```
wochinge, line 110: should we make the `voice`, `language` configurable?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+            vr.append(gather)
+        else:
+            vr.say(message)
+            vr.pause(length=1)
```
wochinge, line 114: should the length of the pause be configurable?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+class TwilioVoiceCollectingOutputChannel(CollectingOutputChannel):
+    """Output channel that collects send messages in a list
+    (doesn't send them anywhere, just collects them)."""
+
+    @classmethod
+    def name(cls) -> Text:
+        return "twilio_voice"
+
+
+def build_twilio_voice_response(messages: List[Text]) -> VoiceResponse:
```
wochinge, line 96: would be nice to unit test this

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,116 @@
...
+            sender_id = request.form.get("From")  # method to get sender_id
+            text = request.form.get("SpeechResult")  # method to fetch text
+            input_channel = self.name()  # method to fetch input channel
+            call_status = request.form.get("CallStatus")  # method to fetch call status
+            call_sid = request.form.get("CallSid")  # Get the call identifier
+
+            collector = TwilioVoiceCollectingOutputChannel()
+
+            # Provide an initial greeting to answer the user's call.
+            if (text is None) & (call_status == "ringing"):
```
wochinge, line 46:
```suggestion
            if (text is None) and (call_status == "ringing"):
```

wochinge (commented):
Nice channel addition 💯 

Can you please
* add a changelog
* describe the channel in the documentation

rasa/core/channels/twilio_voice.py
mvielkind: This is a voice channel nuance. If there is a period of silence the bot will repeat the last utterance so the user knows they need to provide a response and remind them of what they need to respond to. Happy to implement this another way, but this seemed like the easiest way to get the last utterance to repeat.

rasa/core/channels/twilio_voice.py
mvielkind: Definitely. All of these Twilio verbs could be configurable. `voice` and `language` could be two places to start. Would these best be configured as a part of the `credentials.yml` file?

rasa/core/channels/twilio_voice.py
mvielkind: I created `TwilioVoiceCollectingOutputChannel` to override the button and image methods. In the voice channel the images will be removed.

rasa/core/channels/twilio_voice.py
mvielkind: Decided to allow the `voice` to be customizable to let the user inject some personality with the assistant. I still need to test some of the `language` features further to see what kind of disclaimers need to be put in place for how they interact with Rasa.

rasa/core/channels/twilio_voice.py
mvielkind: Having a pause of 1 seems like a sensible default. This could definitely be configured going forward if there is a need.

rasa/core/channels/twilio_voice.py
mvielkind: We could make this customizable if needed. This is a default behavior to start with. In the docs this behavior is described so users should be able to adjust their responses accordingly.

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,55 @@
...
+    assert output_channel.messages[0]["text"] == "Test:"
+
+    twiml = tv.build_twilio_voice_response(output_channel.messages)
+    assert (
+        str(twiml)
+        == '<?xml version="1.0" encoding="UTF-8"?><Response><Gather action="/webhooks/twilio_voice/webhook" actionOnEmptyResult="true" input="speech" speechTimeout="auto"><Say voice="woman">Test:</Say></Gather></Response>'
+    )
+
+
+@pytest.mark.asyncio
```
wochinge, line 30:
```suggestion
```

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,55 @@
...
+async def test_twilio_voice_twiml_response_buttons():
+
+    from rasa.core.channels.twilio_voice import TwilioVoiceInput
```
wochinge, line 33: can we please use global imports? In my opinion there is no advantage of using local ones in this case

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,55 @@
...
+async def test_twilio_voice_twiml_response_text():
+
+    from rasa.core.channels.twilio_voice import TwilioVoiceInput
+    from rasa.core.channels.twilio_voice import TwilioVoiceCollectingOutputChannel
+
+    tv = TwilioVoiceInput("woman")
+
+    output_channel = TwilioVoiceCollectingOutputChannel()
+
+    await output_channel.send_text_message(recipient_id="Chuck Norris", text="Test:")
```
wochinge, line 19: how about also adding a test for multiple responses and pauses?

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,55 @@
+import json
+import logging
+
+import pytest
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.asyncio
+async def test_twilio_voice_twiml_response_text():
```
wochinge, line 10: nice job on the tests! 🎉

wochinge (commented):
Sorry for the amount of comments 🙈 Really cool feature 💯

rasa/core/channels/twilio_voice.py
mvielkind: @wochinge is there a default action for repeating the last utterance or getting the last user utterance? I didn't want to require users to have to create a custom action to implement this channel. This was the best workaround I could find if there is another approach I'm open to that.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,228 @@
+import inspect
```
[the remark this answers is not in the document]
hsm207: removed!

rasa/core/channels/twilio_voice.py
mvielkind: Added a parameter where a user can provide a custom `initial_prompt`. If none is provided then "hello" will be used as a default.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,228 @@
...
+    def __init__(self, assistant_voice: Optional[Text],) -> None:
+        """Creates a connection to Twilio voice.
+
+        Args:
+            assistant_voice: name of the assistant voice to use.
+        """
+        self.assistant_voice = assistant_voice
+
+        if assistant_voice not in self.SUPPORTED_VOICES:
+            self.raise_invalid_voice_exception()
```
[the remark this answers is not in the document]
mvielkind: Added a test for unsupported voices in the latest commit.

docs/docs/connectors/twilio-voice.mdx
```diff
@@ -0,0 +1,45 @@
...
+---
+
+You can use the Twilio Voice connector to answer phone calls made to your Twilio phone number.
+
+## Running on Twilio
+
+To forward calls from Twilio to your Rasa bot the webhook for your phone number needs to be updated. Go to the [Phone
+Numbers](https://www.twilio.com/console/phone-numbers/incoming) section of your Twilio account and select the phone
+number you want to connect to Rasa. Find the `Voice & Fax` section on the screen. Under the `A CALL COMES IN` section
+add the webhook URL (e.g. `https://<host>:<port>/webhooks/twilio_voice/webhook`) for your Rasa bot replacing the host
```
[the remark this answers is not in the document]
mvielkind: Done. Added a screenshot of what this looks like in the console.

docs/docs/connectors/twilio-voice.mdx
```diff
@@ -0,0 +1,45 @@
+---
+id: twilio-voice
+sidebar_label: Twilio Voice
+title: Twilio Voice
+description: Deploy a Rasa Open Source IVR assistant via the Twilio Voice connector
+---
+
+You can use the Twilio Voice connector to answer phone calls made to your Twilio phone number.
```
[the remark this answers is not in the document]
mvielkind: Our new CSE member has been reviewing the documentation and is working to connect the voice channel to the financial demo.

tests/core/channels/test_twilio_voice.py
mvielkind: Hmmm seems like there would need to be a model for this to process the text inputs to the webhook. Do we have an example of a channel connector test that's connected to a model?

rasa/core/channels/twilio_voice.py
wochinge: we could do a new default intent + default action for that 🤔 

Let's keep it like it is for now for the sake of doing a quick iteration. In the end it's just a in a channel component and not in the main architecture. Consider double checking maybe if an agent is available.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,228 @@
...
+                # Get last user utterance from tracker.
+                tracker = request.app.agent.tracker_store.retrieve(sender_id)
+                last_response = next(
+                    (e for e in reversed(tracker.events) if isinstance(e, BotUttered)),
+                    None,
+                )
+
+                # If no previous utterance found say something generic.
+                if last_response is None:
+                    last_response = "I didn't get that."
```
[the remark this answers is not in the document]
wochinge: @mvielkind

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,235 @@
...
+
+    @classmethod
+    def from_credentials(cls, credentials: Optional[Dict[Text, Any]]) -> InputChannel:
+        """Load custom configurations."""
+        if not credentials:
+            return cls(initial_prompt="hello", assistant_voice="woman")
+
+        return cls(
+            credentials.get("initial_prompt", "hello"),
+            credentials.get("assistant_voice", "woman"),
```
wochinge, line 102: 💯

wochinge: should we mention that in the docs?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,235 @@
...
+    @classmethod
+    def name(cls) -> Text:
+        """Name of your custom channel."""
+        return "twilio_voice"
+
+    @classmethod
+    def from_credentials(cls, credentials: Optional[Dict[Text, Any]]) -> InputChannel:
+        """Load custom configurations."""
+        if not credentials:
+            return cls(initial_prompt="hello", assistant_voice="woman")
```
wochinge, line 98: that's covered by the lines after right? If you wanna handle the `is None` case you could also just do 
```suggestion
        credentials = credentials or {}
```

tests/core/channels/test_twilio_voice.py
wochinge: You could maybe use something like [here](https://github.com/RasaHQ/rasa/blob/13696886922f2d09ad47a9c1a5cb846573bf91bf/tests/core/channels/test_hangouts.py#L26). 

not sure why you need a model though. Couldn't you just run `output_channel.send_text_message` twice?

changelog/8180.feature.md
```diff
@@ -0,0 +1 @@
+Allows a Rasa bot to be connected to a Twilio Voice channel. More details in the [Twilio Voice docs](https://rasa.com/docs/rasa/connectors/twilio-voice)
```
wochinge, line 1: let's use relative paths 
```suggestion
Allows a Rasa bot to be connected to a Twilio Voice channel. More details in the [Twilio Voice docs](connectors/twilio-voice.mdx)
```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,235 @@
+from sanic import Blueprint, response
+from sanic.request import Request
+from sanic.response import HTTPResponse
+from twilio.twiml.voice_response import VoiceResponse, Gather
+from typing import Text, Callable, Awaitable, List, Any, Dict, Optional, NoReturn
```
wochinge, line 5: unused
```suggestion
from typing import Text, Callable, Awaitable, List, Any, Dict, Optional
```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,235 @@
...
+        await self._persist_message(self._message(recipient_id, text=text))
+
+        for b in buttons:
+            await self._persist_message(self._message(recipient_id, text=b["title"]))
+
+    async def send_image_url(
+        self, recipient_id: Text, image: Text, **kwargs: Any
+    ) -> None:
+        """For voice channel do not send images."""
+        raise_warning(
```
wochinge, line 233: how about a test for this?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,235 @@
...
+        """
+        self.initial_prompt = initial_prompt
+        self.assistant_voice = assistant_voice
+
+        if assistant_voice not in self.SUPPORTED_VOICES:
+            self.raise_invalid_voice_exception()
+
+    def raise_invalid_voice_exception(self) -> None:
+        """Raises an error if an invalid voice is provided."""
+        raise RasaException(
```
wochinge, line 121: Can we please use a subclass of `RasaException`? E.g. `InvalidConfigException`

docs/docs/connectors/twilio-voice.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+### Assistant Voice
+
+You can add personality to your assistant by specifying the type of voice your assistant should speak with. In the
+`credentials.yml` file you can add an option for `assistant_voice` to specify the type of voice of your assistant. For a
+list of supported voices you can check the [Twilio documentation](https://www.twilio.com/docs/voice/twiml/say#voice). An
+example implementation looks like this where a woman's voice is used. If no voice is provided then a woman's voice will
+be used by default.
+
+```
```
wochinge, line 33:
```suggestion
```title="credentials.yml"
```

docs/docs/connectors/twilio-voice.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+When Twilio receives a new call and forwards this to Rasa Open Source, Twilio does not provide a user message. In this
+ case Rasa Open Source will act as if the user sent the message "Hello". This behavior can be configured in your
+ `credentials.yml` file by setting the `initial_prompt` parameter to the desired input. The `initial_prompt` value will
+ be sent to Rasa Open Source and the response will be spoken to the user to start the voice conversation. How you greet
+ a user via a voice channel may differ from a text channel. You should review your responses and consider creating
+ [channel-specific variations](https://rasa.com/docs/rasa/responses/#channel-specific-response-variations) where
+ necessary.
+
+```
```
wochinge, line 48:
```suggestion
```title="endpoints.yml"
```

hsm207: done

docs/docs/connectors/twilio-voice.mdx
hsm207: done

mvielkind:
@wochinge the issue of handling emojis was raised. My initial thought was raising a warning similar to like what we're doing with the images. Another option would be to raise this error at training to detect where emojis/images are utilized without a voice alternative. Do you see value in adding those data validation checks in now or a later release?

docs/docs/connectors/twilio-voice.mdx
```diff
@@ -0,0 +1,121 @@
...
+Setting `enhanced` to `true` will allow you to use Twilio's premium speech model to improve the accuracy of
+transcription results. Note, setting this parameter to `true` will result in higher Twilio transcription costs. This
+parameter is only supported if you also have set the `speech_model` parameter to `phone_call`. By default `enhanced` is
+set to "false". You can find more about the `enhanced` speech model option in the
+[Twilio documentation](https://www.twilio.com/docs/voice/twiml/gather#enhanced).
+
+## Custom Voice Responses
+
+It is highly recommended that you provide voice
+[channel specific responses](https://rasa.com/docs/rasa/responses/#channel-specific-response-variations) for all
```
wochinge, line 111: this should use our relative links (the ones with `.mdx`)

wochinge: same in line 63

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,342 @@
...
+            if self.speech_timeout.lower() != "auto":
+                self.raise_invalid_speech_timeout_exception()
+
+        if speech_model not in self.SUPPORTED_SPEECH_MODELS:
+            self.raise_invalid_speech_model_exception()
+
+        if enhanced.lower() not in ["true", "false"]:
+            self.raise_invalid_enhanced_option_exception()
+
+        if (enhanced.lower() == "true") & (speech_model.lower() != "phone_call"):
```
wochinge, line 151:
```suggestion
        if (enhanced.lower() == "true") and (speech_model.lower() != "phone_call"):
```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,342 @@
...
+            self.raise_invalid_enhanced_speech_model_exception()
+
+        if (speech_model.lower() != "numbers_and_commands") & (speech_timeout.lower() == "auto"):
```
wochinge, line 154:
```suggestion
        if (speech_model.lower() != "numbers_and_commands") and (speech_timeout.lower() == "auto"):
```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,342 @@
...
+            assistant_voice: name of the assistant voice to use.
+        """
+        self.initial_prompt = initial_prompt
+        self.reprompt_fallback_phrase = reprompt_fallback_phrase
+        self.assistant_voice = assistant_voice
+        self.speech_timeout = speech_timeout
+        self.speech_model = speech_model
+        self.enhanced = enhanced
+
+        if assistant_voice not in self.SUPPORTED_VOICES:
```
wochinge, line 136: how about moving this code and the code after to something like `_validate_configuration`?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,342 @@
...
+            self.raise_invalid_voice_exception()
+
+        try:
+            int(self.speech_timeout)
+        except ValueError:
+            if self.speech_timeout.lower() != "auto":
+                self.raise_invalid_speech_timeout_exception()
```
wochinge, line 143: optional: no need to change but I don't see the value of extracting the exceptions in separate methods. I think the code would be as concise and DRY without doing this.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,342 @@
...
+        if enhanced.lower() not in ["true", "false"]:
+            self.raise_invalid_enhanced_option_exception()
+
+        if (enhanced.lower() == "true") & (speech_model.lower() != "phone_call"):
+            self.raise_invalid_enhanced_speech_model_exception()
+
+        if (speech_model.lower() != "numbers_and_commands") & (speech_timeout.lower() == "auto"):
+            self.raise_invalid_speech_model_timeout_exception()
+
+    def raise_invalid_speech_model_timeout_exception(self) -> None:
```
wochinge, line 157: they can in my opinion all be protected as no code outside of the class would use it
```suggestion
    def _raise_invalid_speech_model_timeout_exception(self) -> None:
```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,342 @@
...
+                "]+",
+                flags=re.UNICODE,
+            )
+
+    @classmethod
+    def name(cls) -> Text:
+        """Name of the output channel."""
+        return "twilio_voice"
+
+    async def emoji_warning(
```
wochinge, line 303: no need to make this `async`
```suggestion
    def emoji_warning(
```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,342 @@
...
+        return voice_response
+
+
+class TwilioVoiceCollectingOutputChannel(CollectingOutputChannel):
+    """Output channel that collects send messages in a list.
+
+    (doesn't send them anywhere, just collects them).
+    """
+
+    EMOJI = re.compile(
```
wochinge, line 284: we have sth similar [here](https://github.com/RasaHQ/rasa/blob/20f0d70d6cbad9e3a3a4db23a7b8df5d08bc173d/rasa/nlu/tokenizers/whitespace_tokenizer.py#L41). How about re-using it?

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,215 @@
...
+        TwilioVoiceInput(**inputs)
+
+
+async def test_twilio_voice_remove_image():
+
+    output_channel = TwilioVoiceCollectingOutputChannel()
+    await output_channel.send_response(
+        recipient_id="Chuck Norris",
+        message={
+            "image": "https://i.imgur.com/nGF1K8f.jpg",
```
wochinge, line 141: how about also testing for the warning in this case?

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,215 @@
...
+    assert message_str == "Buttons: Yes No"
+
+    twiml = tv.build_twilio_voice_response(output_channel.messages)
+    assert (
+        str(twiml)
+        == '<?xml version="1.0" encoding="UTF-8"?><Response><Say voice="woman">Buttons:</Say><Pause length="1" /><Say voice="woman">Yes</Say><Pause length="1" /><Gather action="/webhooks/twilio_voice/webhook" actionOnEmptyResult="true" enhanced="false" input="speech" speechModel="default" speechTimeout="5"><Say voice="woman">No</Say></Gather></Response>'
+    )
+
+
+async def test_twilio_invalid_assistant_voice():
```
wochinge, line 75: why are these tests `async` if we don't use `await` in them? 

I'd also consider using https://docs.pytest.org/en/stable/parametrize.html to group all invalid config tests! 

Nice and thorough testing btw 🎉

docs/docs/connectors/twilio-voice.mdx
```diff
@@ -0,0 +1,121 @@
...
+ case Rasa Open Source will act as if the user sent the message "Hello". This behavior can be configured in your
+ `credentials.yml` file by setting the `initial_prompt` parameter to the desired input. The `initial_prompt` value will
+ be sent to Rasa Open Source and the response will be spoken to the user to start the voice conversation. How you greet
+ a user via a voice channel may differ from a text channel. You should review your responses and consider creating
+ [channel-specific variations](https://rasa.com/docs/rasa/responses/#channel-specific-response-variations) where
+ necessary.
+
+### Assistant Voice
+
+You can add personality to your assistant by specifying the type of voice your assistant should speak with. In the
```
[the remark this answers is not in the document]
mvielkind: An example of `credentials.yml` is shown starting on line 44 then each of the options is discussed in detail after the example.

tests/core/channels/test_twilio_voice.py
mvielkind: Oh, that's much more extensible wrapping those config tests together instead of creating separate functions for each!

wochinge:
@mvielkind There is a merge conflict

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,335 @@
...
+                    last_response = last_response.text
+
+                twilio_response = self.build_twilio_voice_response(
+                    [{"text": last_response}]
+                )
+            return response.text(str(twilio_response), content_type="text/xml")
+
+        return twilio_voice_webhook
+
+    def build_twilio_voice_response(
```
wochinge, line 256:
```suggestion
    def _build_twilio_voice_response(
```

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,335 @@
...
+        """Name of the output channel."""
+        return "twilio_voice"
+
+    @staticmethod
+    def emoji_warning(
+        text: Text,
+    ) -> None:
+        """Raises a warning if text contains an emoji."""
+        tok = WhitespaceTokenizer()
+        emoji_regex = tok.get_emoji_regex()
```
wochinge, line 300: We don't really need the tokenizer - how about moving the `get_emoji` regex to something like `rasa.utils.io`?

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,335 @@
...
+    (doesn't send them anywhere, just collects them).
+    """
+
+    @classmethod
+    def name(cls) -> Text:
+        """Name of the output channel."""
+        return "twilio_voice"
+
+    @staticmethod
+    def emoji_warning(
```
wochinge, line 295:
```suggestion
    def _emoji_warning(
```

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,206 @@
...
+                "reprompt_fallback_phrase": "i didn't get that",
+                "speech_model": "default",
+                "speech_timeout": "5",
+                "assistant_voice": "woman",
+                "enhanced": "true"
+            }, InvalidConfigException
+        )
+    ]
+)
+def test_invalid_configs(configs, expected):
```
wochinge, line 120: can you please add type annotations here?

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,206 @@
...
+    with pytest.raises(expected):
+        TwilioVoiceInput(**configs)
+
+
+async def test_twilio_voice_remove_image():
+
+    with pytest.warns(UserWarning):
+        output_channel = TwilioVoiceCollectingOutputChannel()
```
wochinge, line 128: optional: for clarity I'd only include the call with raises the warning in the `with pytest.warns` block. Otherwise it's hard to see which part of the code is exactly doing this.

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,206 @@
...
+        await output_channel.send_response(
+            recipient_id="User",
+            message={
+                "text": "Howdy 😀"
+            }
+        )
+
+
+def test_twilio_incompatible_speech_timeout_and_model():
```
wochinge, line 152: can we please add these two to `test_invalid_configs`? Or use `pytest.mark.parametrize` here?

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,206 @@
+import json
```
wochinge, line 1: it would be really cool if we could also test the `receive` function of the webhook. Especially the parts which re-utter the last bot utterances seem something which would better be tested.

tests/core/channels/test_twilio_voice.py
mvielkind: Good idea. I'm separating this into two tests. One to make sure the warning gets raised and another to make sure the image is removed and the text is passed through.

tests/core/channels/test_twilio_voice.py
mvielkind: @wochinge I had tried doing this replicating this Google Hangouts test, but I wasn't able to get it to work. Is this the proper approach to building a test for this?

https://github.com/RasaHQ/rasa/blob/main/tests/core/channels/test_hangouts.py#L126

rasa/core/channels/twilio_voice.py
mvielkind: `get_emoji` was moved to `rasa.utils.io` and the reference was updated in both `twilio_voice.py` and in `WhitespaceTokenizer` to utilize the `get_emoji` utility.

tests/core/channels/test_twilio_voice.py
wochinge:
```python
async def test_twilio_receive(stack_agent: Agent):
    app = server.create_app(agent=stack_agent)

    inputs = {
        "initial_prompt": "hello",
        "reprompt_fallback_phrase": "i didn't get that",
        "speech_model": "default",
        "speech_timeout": "5",
        "assistant_voice": "woman",
        "enhanced": "false",
    }

    tv = TwilioVoiceInput(**inputs)
    channel.register([tv], app, "/webhooks/")

    client = app.asgi_client

    body = {"From": "Tobias", "SpeechResult": None, "CallStatus": "ringing"}
    _, response = await client.post(
        "/webhooks/twilio_voice/webhook",
        headers={"Content-type": "application/x-www-form-urlencoded"},
        data=body,
    )
    assert response.status == 200
    # Actual test xml content
    assert response.body
```

This should work. I think that would be super cool because then you can test the entire flow

mvielkind: Neat! This is good to know for the future.

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,253 @@
...
+
+    twiml = tv._build_twilio_voice_response(output_channel.messages)
+
+    assert (
+        str(twiml)
+        == '<?xml version="1.0" encoding="UTF-8"?><Response><Say voice="woman">message 1</Say><Pause length="1" /><Gather action="/webhooks/twilio_voice/webhook" actionOnEmptyResult="true" enhanced="false" input="speech" speechModel="default" speechTimeout="5"><Say voice="woman">message 2</Say></Gather></Response>'
+    )
+
+
+async def test_twilio_receive(stack_agent: Agent):
```
wochinge, line 217: Can we please add more tests for receive? There is still a whole bunch of uncovered cases (marked by the red lines):
Screenshot 2021-05-04 at 09 22 00

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,334 @@
...
+            speechTimeout=self.speech_timeout,
+            speechModel=self.speech_model,
+            enhanced=self.enhanced,
+        )
+
+        # Add pauses between messages.
+        # Add a listener to the last message to listen for user response.
+        for i, message in enumerate(messages):
+            if i + 1 == len(messages):
+                gather.say(message["text"], voice=self.assistant_voice)
```
wochinge, line 274: how about using a constant for the `text` field? You're using it a couple of times across the module.

rasa/core/channels/twilio_voice.py
```diff
@@ -0,0 +1,334 @@
+import re
+from sanic import Blueprint, response
+from sanic.request import Request
+from sanic.response import HTTPResponse
+from twilio.twiml.voice_response import VoiceResponse, Gather
+from typing import Text, Callable, Awaitable, List, Any, Dict, Optional
+
+from rasa.utils.io import get_emoji_regex
+from rasa.shared.utils.io import raise_warning
```
wochinge, line 9: please don't import functions directly but rather reference them by their module (see [here](https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#70d0247fdd1f48d2aec341fd91fe1dac))
```suggestion
import rasa.utils.io
import rasa.shared.utils.io
```

wochinge (commented):
Can you please also merge the latest `main` branch into this? As long as there are merge conflicts, no CI checks will run.

Do you think we can get this into the release on Thursday?

tests/core/channels/test_twilio_voice.py
```diff
@@ -0,0 +1,313 @@
...
+
+    client = app.asgi_client
+
+    body = {"From": "Tobias", "CallStatus": "ringing"}
+    _, response = await client.post(
+        "/webhooks/twilio_voice/webhook",
+        headers={"Content-type": "application/x-www-form-urlencoded"},
+        data=body,
+    )
+    assert response.status == 200
```
wochinge, line 237: optional (same for others)
```suggestion
    assert response.status == HTTPStatus.OK
```

wochinge (approved):
I left a few more small comments, but looks great otherwise! Sorry for nudging you with the tests, but I think the outstanding 95% test coverage were worth it 💯 Thanks for providing a super cool new feature + a ton of tests to enable easy maintenance of it 🥇

Issue #8164, opened 2021-03-10
[Can provide PR from working repo] Buttons click handling MS Botframework


thaume:
**Rasa version**: 2.3.0

**Rasa SDK version** (if used & relevant): Nope

**Rasa X version** (if used & relevant): Nope

**Python version**: I'm using the provided dockerfile to build and run Rasa

**Operating system** (windows, osx, ...): MacOS, but again rasa is running in Docker

**Issue**:

I'm having an issue with handling buttons clicks coming from Teams. The way Teams handles clicks and text input is different and right now the Botframework Rasa connector (that was built by my company 2 years ago) doesn't support that.

We have a working version that we keep up to date with the master repo from time to time but I think it would make more sense to take the time to merge it.

And... it's just 6 lines of code 😅

**Error (including full traceback)**: This is the stacktrace shown by Rasa after a click on a button
```
2021-03-10 09:47:44 DEBUG    rasa.core.processor  - Received user message '' with intent '{'name': None, 'confidence': 0.0}' and entities '[]'
```

**Command or request that led to error**:
After writing and sending a message ("Home") to Rasa from Teams, this JSON is sent to rasa endpoint :
```json
{
    "text": "Home",
    "textFormat": "plain",
    "type": "message",
    "timestamp": "<confidential>",
    "localTimestamp": "<confidential>",
    "id": "<confidential>",
    "channelId": "msteams",
    "serviceUrl": "https://smba.trafficmanager.net/emea/",
    "from": {
        "id": "<confidential>",
        "name": "<confidential>",
        "aadObjectId": "<confidential>"
    },
    "conversation": {
        "conversationType": "personal",
        "tenantId": "<confidential>",
        "id": "<confidential>"
    },
    "recipient": {
        "id": "<confidential>",
        "name": "<confidential>"
    },
    "entities": [
        {
            "locale": "en-GB",
            "country": "GB",
            "platform": "Web",
            "timezone": "Europe/Paris",
            "type": "clientInfo"
        }
    ],
    "channelData": {
        "tenant": {
            "id": "<confidential>"
        }
    },
    "locale": "en-GB",
    "localTimezone": "Europe/Paris"
}
```

After clicking on a button from a card (created by the chatbot) on Teams, this JSON is sent to rasa endpoint :
```json
{
    "type": "message",
    "timestamp": "<confidential>",
    "localTimestamp": "<confidential>",
    "id": "<confidential>",
    "channelId": "msteams",
[...]

PR #8190, opened 2021-03-12
Add better data input management for MS botframwork

thaume:
**Proposed changes**:
As discussed in https://github.com/RasaHQ/rasa/issues/8164

We have been using this feature in production for 2 years now, I felt it was time to share it with everyone else using buttons in Teams' adaptive cards.

It's not a groundbreaking feature but it is very important.

I'll add the tests once we validate that we need this feature here :)

### DIFF
--- a/rasa/core/channels/botframework.py
+++ b/rasa/core/channels/botframework.py
@@ -232,8 +232,18 @@ class BotFrameworkInput(InputChannel):
                         postdata["serviceUrl"],
                     )
 
+                    text = ""
+                    if postdata.get("text"):
+                        text = postdata.get("text")
+                    else:
+                        if postdata.get("value"):
+                            raw_value = postdata.get("value")
+                            if "value" in raw_value:
+                                raw_value = raw_value["value"]
+                            text = json.dumps(raw_value)
+
                     user_msg = UserMessage(
-                        text=postdata.get("text", ""),
+                        text=text,
                         output_channel=out_channel,
                         sender_id=postdata["from"]["id"],
                         input_channel=self.name(),


sara-tagger:

rasa/core/channels/botframework.py
```diff
@@ -232,8 +232,18 @@ async def webhook(request: Request) -> HTTPResponse:
                         postdata["serviceUrl"],
                     )
 
+                    text = ""
+                    if postdata.get("text"):
```
m-vdb, line 236: would be great to have a link to the documentation of MS botframework that describes the request format, for reference

m-vdb (commented):
looks good @thaume 💯 sorry for the late review

rasa/core/channels/botframework.py
thaume: Hey @m-vdb sure the doc is here : https://docs.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-api-reference?view=azure-bot-service-4.0#activity-object

It's not completely clear that `text` or `value` can be empty and will be empty depending on what you are sending. That wasn't clear to us in the first implementation of this connector :)

rasa/core/channels/botframework.py
```diff
@@ -232,8 +232,18 @@ async def webhook(request: Request) -> HTTPResponse:
...
+                        text = postdata.get("text")
+                    else:
+                        if postdata.get("value"):
```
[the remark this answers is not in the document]
thaume: it felt more readable like this, but I can totally change it to an `elif`, let me know if you'd prefer that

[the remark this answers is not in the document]
m-vdb: it's more pythonic this way :) and usually linters warn about this (though we do not have the rule set up yet)

rasa/core/channels/botframework.py
m-vdb: sounds good 👍🏻 can you add that to the docstring of the function? it'll be easier for other contributors to dive into the code next time we want to implement a change

thaume:
@m-vdb PR fixed following your comments

rasa/core/channels/botframework.py
```diff
@@ -215,6 +215,10 @@ async def health(request: Request) -> HTTPResponse:
 
         @botframework_webhook.route("/webhook", methods=["POST"])
         async def webhook(request: Request) -> HTTPResponse:
+            """ The content of the Activity object exchanged in this request is detailed in MS
+                docs here : https://docs.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-api-reference?view=azure-bot-service-4.0#activity-object
+
+            """
```
m-vdb, line 221: Lint is failing here + Added first sentence to describe the purpose of the function.

```suggestion
            """Webhook receiving requests from MS Botframework.

            The request contains data about end-user messages or actions.
            The content of the Activity object exchanged in this request
            is detailed in MS:
            https://docs.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-api-reference?view=azure-bot-service-4.0#activity-object
```

m-vdb (changes requested):
Thanks for updating the docstring. I've added a suggesting to fix the failing lint. Could you also take care of the remaining steps from the PR description?
> - [ ] added some tests for the functionality
> - [ ] updated the documentation
> - [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
> - [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

thaume:
Awesome I've updated the docstring, taking care of the other steps now

thaume:
@m-vdb I updated everything, I'm just wondering about the tests : since there are none it seems a bit overkill (and I won't have time right now) to write all the tests for this channel, what do you think ?

In terms of documentation, there is no impact, it will just return a correct value everytime.

Let me know if you need me to change anything else

rasa/core/channels/botframework.py
```diff
@@ -214,6 +214,14 @@ async def health(request: Request) -> HTTPResponse:
 
         @botframework_webhook.route("/webhook", methods=["POST"])
         async def webhook(request: Request) -> HTTPResponse:
+            """ Webhook receiving requests from MS Botframework.
+
```
m-vdb, line 218: oups, still some whitespaces left here:

```suggestion
            """Webhook receiving requests from MS Botframework.

```

m-vdb (commented):
Thanks for adding the changelog item. I agree with you that the documentation part can be skipped ✅ 

Regarding the test, asking you to write all the tests is a bit too much for sure 😅 but could you write 3 unit tests that check for the base use case and the ones you introduced?
1. postdata contains `text`
2. postdata doesn't contain `text` but contains `value`, which in turns _contains_ `value`
2. postdata doesn't contain `text` but contains `value`, which in turns _doesn't contain_ `value`

You could take some inspiration from [the tests written for the Slack channel](https://github.com/RasaHQ/rasa/blob/main/tests/core/channels/test_slack.py).

m-vdb:
@thaume following up on this one, do you need help to finalise the PR?

thaume:
@m-vdb hey ! I'll never find the time to wrap my head around the test strategy and write them myself. That's the main concern here 🙂

PR #8204, opened 2021-03-15
8116 replace moodbot example fixtures

### LINKED ISSUE #8116: Replace moodbot fixtures with quicker fixture
**Description of Problem**:
We are currently using the moodbot example from [here](https://github.com/RasaHQ/rasa/tree/main/examples/moodbot) in tests. This has multiple disadvantages:
* This model takes long to train
* This model takes long to load as the spacy language model has to be loaded into memory
* Our tests depend on examples for the users instead of a more elegant / custom tailored model for testing
 
**Overview of the Solution**:
Replace fixtures with a model which is quicker to train and load. Ideally this model uses a minimal amount of data and a very quick configuration (e.g. all `epochs` set to `1`)

**Definition of Done**:
- [ ] The fixtures `trained_moodbot_path`, `trained_nlu_moodbot_path` are replaced with quicker ones
- [ ] The new fixtures are based on training data form the `data` folder instead of the `examples/` folder
- [ ] Models using the new fixture load in less than 5 seconds (**this is somewhat arbitrary atm**)
- [ ] The fixture trains in less than 30 seconds (**this is somewhat arbitrary atm**)

joejuzl:
**Proposed changes**:
Fixes #8116 
We use the example bot moodbot in lots of our tests. It is slow to train and not part of our test data.
This PR creates similars but quicker bots in the data directory and uses it in the tests.


alwx (approved):
Looks nice! Do you have any sample time measurements for any of the tests before/after the change?

Issue #7640, opened 2020-12-29
During rasa test, YamlValidationException shows temp path


shubhvachher:
**Rasa version**: 2.0.0

**Rasa SDK version** (if used & relevant): Not Used

**Rasa X version** (if used & relevant): Not Used

**Python version**: 3.6.8 (anaconda)

**Operating system** (windows, osx, ...): windows

**Issue**:

During `rasa test` if there is a `YamlValidationException` then the path for the file shown points to the temp directory. 

**Error (including full traceback)**:
```
YamlValidationException: Failed to validate 'C:\Users\USER\AppData\Local\Temp\tmpjvcoar7v\3c3f65c658cd4a41b3fb1703892fa328_test_happypaths_basic.yml'. Please make sure 
the file is correct and all mandatory parameters are specified. Here are the errors found during validation:
...
```

**Command or request that led to error**:
```
rasa test
```

You could test using this `retrieval intent` example from my `test_happypaths_basic.yml`:
```
stories:
- story: faq/restaurant_available_testing
  steps:
  - user: "Does the hotel have a restaurant?"
  - intent: faq/restaurant_available
  - action: utter_faq/restaurant_available
```

**Definition of Done**
- [ ] evalute if copying to `/temp` is needed and drop it if not
- [ ] correctly report path to actual test file

shubhvachher:
I could submit a PR for this if reproducible! Seems straight forward on the face of it.

sara-tagger:
Thanks for raising this issue, @rgstephens will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

rgstephens:
@shubhvachher Thanks for reporting. Yes, a PR would be appreciated!

shubhvachher:
Sure, I'm out travelling till Jan 3rd. I'll get to this Jan 4.

wochinge:
@shubham3582 Any news on this?

Issue #5913, opened 2020-05-28
MultiProjectImporter should ignore tests directory in imported projects


ortsaCniveK:
**Description of Problem**:

I was in the process of testing out the `MultiProjectImporter` for use in my own projects (which I love), so I set up two nested directories using `rasa init`:

```log
├── actions.py
├── config.yml
├── data
│   ├── mood_1_nlu.md
│   └── mood_1_stories.md
├── domain.yml
├── models
│   └── 20200528-101144.tar.gz
├── projects
   └── Mood2Bot
       ├── config.yml
       ├── data
       │   ├── mood_2_nlu.md
       │   └── mood_2_stories.md
       ├── tests
       │   └── conversation_tests.md
       └── domain.yml
```
This lead to log messages like these:
```log
UserWarning: Found unknown intent 'bot_challenge: are you a bot?' on line 50. Please, make sure that all intents are listed in your domain yaml.
```
```log
2020-05-28 10:24:57 WARNING  rasa.core.featurizers  - Feature 'intent_bot_challenge: are you a bot?' could not be found in feature map.
```

I believe [this is the function](https://github.com/RasaHQ/rasa/blob/ff0848419d780b042bf2aa14f932fe1a2050f5ea/rasa/importers/multi_project.py#L96) causing this issue.

**Overview of the Solution**:

Ignore the `tests/` directory in an imported project, since this format is not for training, [as specified in the docs](https://rasa.com/docs/rasa/core/stories/#end-to-end-story-evaluation-format). Just as well, it promotes isolated, project-level conversation testing, [which seems to be the goal of this feature](https://rasa.com/docs/rasa/api/training-data-importers/#multiprojectimporter-experimental).

**Examples** (if relevant):

**Blockers** (if relevant):

**Definition of Done**:

- [x] Add test for `tests/` in imported project
- [x] Ignore `tests/` directory when walking the project directory
- [x] Document changes in change log

sara-tagger:
Thanks for submitting this feature request 🚀 @Ghostvv will get back to you about it soon! ✨

wochinge:
good catch, @ortsaCniveK 🎉 Do you wanna implement a fix for this? I'd be available for feedback and help. Otherwise we'd add it to our backlog.

ortsaCniveK:
I can work on this!

PR #8217, opened 2021-03-16
fix temporary files in 'rasa test'

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7640
- introduce a new method `TrainingDataImporter.get_conversation_tests` which only retrieves conversation test files
- change server to store test data in the expected format (`tests` directory for Markdown, `test_` prefix for YAML)
- use `TrainingDataImporter` to load data instead of moving eligble files to a separate temporary directory
- added docstrings for `importer` package
- renamed conversation test files to match expections (`test_` prefix for YAML)


rasa/shared/importers/rasa.py
```diff
@@ -54,10 +58,18 @@ async def get_stories(
             exclusion_percentage,
         )
 
+    async def get_conversation_tests(self) -> StoryGraph:
+        """Retrieves conversation test stories (see parent class for full docstring)."""
```
wochinge, line 62: the alternative is to filter in `get_stories` based on `use_e2e`. Considering [this](https://martinfowler.com/bliki/FlagArgument.html) and also to avoid anything breaking I decided to go for the explict extra method

tests/conftest.py
```diff
@@ -138,7 +138,7 @@ def incorrect_nlu_data_path() -> Text:
 
 @pytest.fixture(scope="session")
 def end_to_end_story_path() -> Text:
-    return "data/test_evaluations/end_to_end_story.yml"
+    return "data/test_evaluations/test_end_to_end_story.yml"
```
wochinge, line 141: test stories need to include the `test_` prefix. Our tests were violating this.

rasa/shared/importers/rasa.py
```diff
@@ -45,7 +49,7 @@ async def get_stories(
         use_e2e: bool = False,
         exclusion_percentage: Optional[int] = None,
     ) -> StoryGraph:
-
+        """Retrieves training stories / rules (see parent class for full docstring)."""
```
joejuzl, line 52: Why does this implementation not branch on `use_e2e` like the one in `MultiProjectImporter`?

joejuzl: don't we need to use `_conversation_test_files` path?

rasa/core/test.py
```diff
@@ -976,14 +980,16 @@ def _plot_story_evaluation(
...
+    model_dir: Text, stories_file: Text, output: Text, use_e2e: bool = False
 ) -> None:
-    """Evaluate multiple trained models in a directory on a test set.
+    """Evaluates multiple trained models in a directory on a test set.
 
     Args:
         model_dir: path to directory that contains the models to evaluate
         stories_file: path to the story file
         output: output directory to store results to
+        use_e2e: `True` if markdown story files should be parsed as end-to-end
```
joejuzl, line 991: This is not all it's doing right? It's also determining which files to load.

rasa/core/test.py
```diff
@@ -352,7 +352,11 @@ async def _create_data_generator(
     test_data_importer = TrainingDataImporter.load_from_dict(
         training_data_paths=[resource_name]
     )
-    story_graph = await test_data_importer.get_stories(use_e2e=use_e2e)
+    if use_e2e:
```
joejuzl, line 355: I think this branching is confusing, as we are also passing in a variable `resource_name` into the importer. It's hard to know how those two different things interplay. Would it not be clearer if we do this resource selection up higher when we start the test process, and then `use_e2e` could remain solely for determining how to read the file?

rasa/server.py
```diff
@@ -1438,12 +1440,20 @@ def _test_data_file_from_payload(
 ) -> Text:
     if request.headers.get("Content-type") == YAML_CONTENT_TYPE:
         return str(
-            _training_payload_from_yaml(request, temporary_directory)["training_files"]
+            _training_payload_from_yaml(
+                request,
+                temporary_directory,
+                # test stories have to prefixed with `test_`
+                file_name=f"{TEST_STORIES_FILE_PREFIX}data.yml",
```
joejuzl, line 1447: should all this new server behaviour be described in the changelog?

joejuzl (changes requested):
Could you pinpoint for me the line which used to move the files to the temporary directory?

rasa/server.py
wochinge: this doesn't change any server behavior. It's rather needed because our `test` functions now enforce our rules for test files (e.g. that they have to start with `test_`)

rasa/shared/importers/importer.py
```diff
@@ -48,16 +48,23 @@ async def get_stories(
...
+        # TODO: Drop `use_e2e` in Rasa Open Source 3.0.0 when removing Markdown support
         raise NotImplementedError()
 
+    async def get_conversation_tests(self) -> StoryGraph:
+        """Retrieves end-to-end conversation stories for testing.
+
+        Returns:
+            `StoryGraph` containing all loaded stories.
+        """
+        return await self.get_stories(use_e2e=True)
```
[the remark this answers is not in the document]
wochinge: This is implemented as fallback. Ideally they override this when they have a custom `TrainingDataImporter` class. We cannot let it raise in case it's not implemented as this might break existing implementations of custom `TrainingDataImporter`s

rasa/shared/importers/rasa.py
wochinge: We could do so. I found it a bit risky / fragile to select different training data based on a flag. That's why I preferred introducing a separate method on the class.

rasa/cli/test.py
```diff
@@ -154,8 +152,12 @@ async def run_nlu_test_async(
         test_nlu,
     )
 
-    nlu_data = rasa.cli.utils.get_validated_path(data_path, "nlu", DEFAULT_DATA_PATH)
-    nlu_data = rasa.shared.data.get_nlu_directory(nlu_data)
```
wochinge, line 158: @joejuzl This was moving NLU files to temporary directories

rasa/cli/test.py
```diff
@@ -78,10 +79,6 @@ def run_core_test(args: argparse.Namespace) -> None:
     stories = rasa.cli.utils.get_validated_path(
         args.stories, "stories", DEFAULT_DATA_PATH
     )
-    if args.e2e:
-        stories = rasa.shared.data.get_test_directory(stories)
-    else:
-        stories = rasa.shared.data.get_core_directory(stories)
```
wochinge, line 84: @joejuzl This was moving Core files to temporary directories

rasa/server.py
joejuzl: So it's now writing a different file name than before? That seems like a change in behaviour to me!

rasa/shared/importers/rasa.py
joejuzl: I guess my question then is why do we have:
```
if use_e2e:
            return await self.get_conversation_tests()
```
in `MultiProjectImporter.get_stories`?

rasa/core/test.py
wochinge: Very good point. The `TrainingDataImporter` is already categorizing files (e.g. into `nlu` and `stories`) and hence I think this should be part of the `TrainingDataImporter`. As this function `_create_data_generator` is called from different entrypoints I found it preferable to handle this in one place instead of multiple ones. Otherwise we might run into situations as we had it before where e.g. tests or also the server weren't respecting the conditions we have on these test files (e.g. the `test_` prefix). 

I agree that this is all confusing though 🙈 This still seemed the somewhat clearest approach to me though.

rasa/server.py
wochinge: Not for the user. They send a file via API and we save it to a temporary directory before loading it for testing. How we name it and how the temporary directory looks like doesn't affect the user and is purely internal in my opinion.

joejuzl: ah ok gotcha! 👍

rasa/shared/importers/rasa.py
wochinge: because the multiproject importer included test stories during training: https://github.com/RasaHQ/rasa/issues/5913. This couldn't really happen in the normal importer because users just provided different paths were one directory couldn't contain both. 

Ideally we'd explicitly exclude test files from `get_stories` as well in `RasaFileImporter` 🤔

rasa/core/test.py
joejuzl: Maybe we should use a different name than `use_e2e` then? As it normally dictates what reader to use.
e.g. `conversation_test_stories`?

rasa/shared/importers/importer.py
```diff
@@ -48,16 +48,23 @@ async def get_stories(
         Returns:
             `StoryGraph` containing all loaded stories.
         """
-
+        # TODO: Drop `use_e2e` in Rasa Open Source 3.0.0 when removing Markdown support
```
joejuzl, line 51: This is confusing now that `use_e2e` also determines which files are loaded right?

rasa/shared/importers/multi_project.py
```diff
@@ -183,18 +184,28 @@ async def get_stories(
...
+        return await utils.story_graph_from_paths(
+            self._e2e_story_paths, await self.get_domain(), use_e2e=True,
+        )
+
     async def get_config(self) -> Dict:
+        """Retrieves model config (see parent class for full docstring)."""
         return self.config
 
     async def get_nlu_data(self, language: Optional[Text] = "en") -> TrainingData:
+        """Retrieves NLU training data (see parent class for full docstring)."""
```
joejuzl, line 210: Do we have a convention for docstrings on subclasses?

rasa/shared/importers/rasa.py
joejuzl: 🤯 I'm not sure I understand. We have `get_stories` _and_ `get_conversation_tests` now. Can't we just call the correct method where needed? Why do we need to sometimes return conversation tests from the `get_stories` method in some importers?

joejuzl (changes requested):
Is there a reasonable way to test we are no longer creating the temp directories?

rasa/shared/importers/multi_project.py
wochinge: Not specifically. It's somewhat my personal preference to write a proper docstring in one place and then link to it (ideally I'd just omit but that doesn't work with `pydocstyle`)

rasa/core/test.py
wochinge: You mean renaming it within `_create_data_generator` or in general in the entire module? I'd be open to rename it for `_create_data_generator` but unfortunately `use_e2e` has multiple other implications on the testing

rasa/shared/importers/rasa.py
wochinge: true 👍🏻 Will remove the implementation from the multiprojectimporter

wochinge:
> Is there a reasonable way to test we are no longer creating the temp directories?

I gave it a manual testing by putting a story with an invalid schema in the `tests` directory. I don't really fancy testing specific log message so opted for not testing this.

rasa/core/test.py
joejuzl: > but unfortunately use_e2e has multiple other implications on the testing

This is why it might make sense to rename. You've added it to lots of methods in this PR with the sole aim of selecting different files in this method. Whereas it normally has the meaning of how a markdown file is read. When we remove markdown, and hopefully a lot of usages of `use_e2e`, we will still need this one to select conversation tests.

rasa/shared/importers/importer.py
wochinge: I hope the renamed parameters make this less confusing. The importer only uses `use_e2e` for configuring the markdown reading. Means we no longer need it when Markdown is removed.

rasa/model_testing.py
```diff
@@ -248,10 +293,19 @@ def plot_nlu_results(output_directory: Text, number_of_examples: List[int]) -> N
...
-):
+) -> None:
+    """Runs cross-validation on test data.
+
+    Args:
+        config: The model configuration.
+        data: The data which is used for the cross-validation.
+        output: Output directory for the cross-validation results.
+        additional_arguments: Additional arguments which are passed to the
+            cross-validation, like.
```
joejuzl, line 307: like what? :)

tests/test_memory.py
```diff
@@ -0,0 +1,223 @@
+import abc
```
joejuzl, line 1: ooops guess this is not meant to be here!

tests/test_memory_joe.py
```diff
@@ -0,0 +1,28 @@
...
+)
+
+featurizer = MaxHistoryTrackerFeaturizer(SingleStateFeaturizer(), max_history=None)
+policy = TEDPolicy(featurizer=featurizer, epochs=500)
+
+policy.train(training_trackers, default_domain, RegexInterpreter())
+
+"""
+Run:
+    mprof run python tests/test_memory_joe.py
```
joejuzl, line 26: I like the name! 😁

rasa/server.py
```diff
@@ -1437,12 +1439,20 @@ def _test_data_file_from_payload(
...
         )
     else:
-        return rasa.utils.io.create_temporary_file(
-            request.body, mode="w+b", suffix=suffix
-        )
+        # MD test stories have to be in the `tests` directory
+        test_dir = temporary_directory / DEFAULT_CONVERSATION_TEST_PATH
+        test_dir.mkdir()
+        test_file = test_dir / f"tests{suffix}"
+        test_file.write_bytes(request.body)
```
joejuzl, line 1454: Does this close the file after?

joejuzl (changes requested):

rasa/server.py
wochinge: yep

tests/test_memory_joe.py
wochinge: 🙈

joejuzl (approved):
LGTM 👍

PR #8222, opened 2021-03-17
Raising exception when having empty intent or entity mapping in the domain

Imod7:
**Proposed changes**:
- Added an InvalidDomain exception when having an empty intent/entity mapping in the `domain.yml` file.

- [x] added two tests for the functionality


rasa/shared/core/domain.py
```diff
@@ -327,7 +327,21 @@ def _transform_intent_properties_for_internal_use(
...
-        properties.setdefault(USE_ENTITIES_KEY, True)
+        if properties:
+            properties.setdefault(USE_ENTITIES_KEY, True)
+        else:
+            raise InvalidDomain(
+                f"In the `domain.yml` file, the intent `{name}` cannot have value of"
+                f" `{type(properties)}`. This can happen if you add an intent with the"
+                f" `:` (colon character) after it and do not add any"
+                f" further information to this specific intent."
+                f" If you remove the `:` character it should work. Please see "
```
alwx, line 338: I would add more certainty to this phrase. Something like `You need to remove the ":" character to make it work.`

rasa/shared/core/domain.py
```diff
@@ -327,7 +327,21 @@ def _transform_intent_properties_for_internal_use(
         """
         name, properties = list(intent.items())[0]
```
alwx, line 328: Could you please also update the docstring for this function? 

It currently says the following: `intent: The intents as provided by a domain file.`
In fact, `intent` is just a single intent, not many of them. The current docstring is pretty confusing.

alwx (approved):
Approved but please take a look at the two small comments I've left :)

Imod7:
Hi @alwx ! Thank you so much for your review and constructive feedback! Please let me know if any further rewording is needed in the docstring or the exception.

After making the current changes for the empty **intent** mapping, I also thought that a user might also insert an empty **entity** mapping so in the `domain.yml` file maybe we get something like this :
```
entities:
  - cuisine:
```
instead of the correct one : 
```
entities:
  - cuisine
```
and the error we get when having the empty entity mapping is the following : 
```
TypeError: argument of type 'NoneType' is not iterable
```

I checked the code and saw that if I add a check [here](https://github.com/RasaHQ/rasa/blob/8f8da24b486e8d9fc66d5b13ab176ad1f40b6a03/rasa/shared/core/domain.py#L412), then instead of the error I can raise another InvalidDomain exception message.  
Should I add this check?  If yes, should I do it in the current PR or open a new Issue ? 
Thank you so much for your time!!! 🙏🏽

alwx (approved):

alwx:
Since it's a small change and because it's related to what you've done here, then I think it's perfectly fine to do it in the same PR. Don't forget to update the changelog entry in this case! :)

wochinge:
@joejuzl Do you have time for another quick review so this PR can get wrapped up?

joejuzl (approved):
LGTM

PR #8226, opened 2021-03-17
`_utter_responses` should handle `template` and `response`

### LINKED ISSUE #8223: `dispatcher.utter_message(template=...)` no longer works in a custom action
**Rasa version**: 2.4.0

**Rasa SDK version** (if used & relevant): 2.4.0

**Python version**:

**Issue**: 
`dispatcher.utter_message(template=...)` no longer works in a custom action. Nothing is uttered by the bot. This is related to the changes introduced in https://github.com/RasaHQ/rasa/pull/6964.

The debug logs between Rasa 2.2 & 2.4 look the same, up to this point, after the custom action executed:
```
# rasa 2.2
2021-03-15 16:01:13 DEBUG    rasa.core.processor  - Action 'action_show_balance' ended with events '[BotUttered('Your bank account balance is $49691.97.', {"elements": null, "quick_replies": null, "buttons": null, "attachment": null, "image": null, "custom": null}, {"template_name": "utter_account_balance", "init_account_balance": "49691.97"}, 1615838473.6100328)]'.
# rasa 2.4
2021-03-15 15:54:31 DEBUG    rasa.core.processor  - Action 'action_show_balance' ended with events '[BotUttered('None', {"elements": null, "quick_replies": null, "buttons": null, "attachment": null, "image": null, "custom": null}, {"template": "utter_account_balance", "init_account_balance": "52946.23"}, 1615838071.5593858)]'.
```

```

```

```yml

```

```yml

```

yennycheung:
**Proposed changes**:
- `_utter_responses` should handle `template` and `response`
- Fixes #8223 

**Manual testing**
I tested this on formbot. Before, formbot wasn't able to respond with `utter_wrong_num_people`. After the change, it was able to.


rasa/core/actions/action.py
```diff
@@ -625,7 +625,14 @@ async def _utter_responses(
         """Use the responses generated by the action endpoint and utter them."""
         bot_messages = []
         for response in responses:
+            generated_response = None
```
wochinge, line 628: can we please add a test for this? Should have been caught by a test in the first place.

yennycheung:
@wochinge, I want to ask for your opinion here. I think we should still make the sdk change now, so that the deprecation message would make sense. Any dependencies I need to be aware of? If we push the sdk change first, it'll also make sense to update these https://github.com/RasaHQ/rasa/blob/main/examples/formbot/actions/actions.py#L51

wochinge (commented):
> I want to ask for your opinion here. I think we should still make the sdk change now, so that the deprecation message would make sense. Any dependencies I need to be aware of? If we push the sdk change first, it'll also make sense to update these

I agree somewhat. My reasoning behind doing Rasa Open Source first was that this bug blocks users right now while a weird deprecation message isn't blocking for users. 

I'd make the SDK send both (`template` + `response`) keys for now to make the new SDK version keeps on working with older Rasa Open Source versions. 

Depends all a bit on your time. I can do the SDK change tomorrow morning to support you?

rasa/core/actions/action.py
yennycheung: > > I want to ask for your opinion here. I think we should still make the sdk change now, so that the deprecation message would make sense. Any dependencies I need to be aware of? If we push the sdk change first, it'll also make sense to update these
> 
> I agree somewhat. My reasoning behind doing Rasa Open Source first was that this bug blocks users right now while a weird deprecation message isn't blocking for users.
> 
> I'd make the SDK send both (`template` + `response`) keys for now to make the new SDK version keeps on working with older Rasa Open Source versions.
> 
> Depends all a bit on your time. I can do the SDK change tomorrow morning to support you?

I can do it later tonight. :) I think I'm already getting a lot of help from you. 🙏🏼

wochinge: Ok, that would be awesome! 🙌🏻  As reviewer of the PR I should have caught this and we are in this together so always happy to help.

yennycheung: Re: the test, I agree with you to add test coverage. This function is not a pure function so it's a bit hard to test. But I'll do my best. I need some time looking through how we do tests here, probably need to mock out a bunch of stuff. Unfortunately, I'm stuck in a bunch of meetings till later in the evening. But I'll come back to this in the evening. 🙏🏼

wochinge: I think you can just copy [this](https://github.com/RasaHQ/rasa/blob/97ae5f1fbf966aece39273928398a3b5d891495a/tests/core/test_actions.py#L241) test and change `"response": "utter_ask_cuisine"` to `"template": "utter_ask_cuisine"`

wochinge: + a `pytest.warns(...)` block to check for your deprecation warning

yennycheung: I'll take a look. Also it *is* actually a pure function, if it doesn't work out I'll just mock stuff.

wochinge: True. It's not public though. Might make sense mock the endpoint and and run the action through the public interface to be closer to the real world and not as reliant on the specific implementation.

yennycheung: Yeah the test you sent me was perfect 💯

yennycheung:
@wochinge, I looked into the SDK, but I got some questions, I'll ping you tomorrow. :)

rasa/core/actions/action.py
```diff
@@ -625,7 +625,14 @@ async def _utter_responses(
...
             generated_response = response.pop("response", None)
+            if not generated_response:
+                generated_response = response.pop("template", None)
+                rasa.shared.utils.io.raise_deprecation_warning(
+                    "The terminology 'template' is deprecated and replaced by 'response', use the `response` parameter instead of `template` in the dispatcher.",
```
wochinge, line 633: How about making this warning more actionable? E.g. upgrading to Rasa SDK xy or adapting their custom SDK?

wochinge (approved):
Looks great! I'd update the deprecation message to make it more actionable but great change otherwise.

rasa/core/actions/action.py
```diff
@@ -625,7 +625,17 @@ async def _utter_responses(
...
         bot_messages = []
         for response in responses:
+            generated_response = None
             generated_response = response.pop("response", None)
+            if not generated_response:
+                generated_response = response.pop("template", None)
+                rasa.shared.utils.io.raise_deprecation_warning(
+                    "The terminology 'template' is deprecated and replaced by "
+                    "'response', use the `response` parameter instead of "
+                    "`template` in the dispatcher. You can do that by upgrading "
```
wochinge, line 635:
```suggestion
                    "`template` in 'dispatcher.utter_message'. You can do that by upgrading "
```

wochinge (approved):

wochinge (changes requested):
@yennycheung Can you please rebase this to be merged into 2.4.x?

yennycheung:
Thanks @wochinge. I created a new PR pointing to 2.4.x at https://github.com/RasaHQ/rasa/pull/8239

Issue #7827, opened 2021-01-27
Create CI check to detect training memory leaks


joejuzl:
**Description of Problem**:
We recently identified a memory leak in the training of Rasa models.
This leak was on the main branch, but only detected when it got so bad due to a new [change](https://github.com/RasaHQ/rasa/pull/7438) that it was crashing CI test workers.
After much investigation the leak was narrowed to down to a bit of tensorflow code and [fixed](https://github.com/RasaHQ/rasa/pull/7817).
Ideally this memory leak would have been caught sooner, as it would have affected users.
The leak was apparent when either training with a high number of epochs, or training multiple times (like in the test suite).
We would like to have an automated check to test that we don't introduce another memory leak.

**Overview of the Solution**:
Tensorflow memory leaks can be hard to identify and fix as they often occur when the graph is being executed which could be in c code for example. This means the "leaking" variables are often not visible when looking at all the python objects in the interpreter. However you can identify if a memory leak exists by looking at the total memory usage of the process.
We used https://pypi.org/project/memory-profiler/ to track the memory usage of a python process when training the TED model to find the leak mentioned above.
This tool tracks the total memory usage over time and writes it to a file which can be parsed or plotted.
To use this in a automated fashion we could:
  - Create a test which trains a a model with dummy data but a high number of epochs
  - Run this python process wrapped in the profiler
  - Analyse the output to see the trend of the total memory usage

We could have a threshold that if crossed fails the test, e.g. 1GB

**Definition of Done**:
- [ ] The check can identify the bug that is mentioned in the description.
- [ ] The check works in the CI.
- [ ] The check works locally.

PR #8249, opened 2021-03-18
add automated memory leak tests

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/7827
- add two checks to detect memory leaks during training:
  - for the training of the core policies
  - for the training of the NLU components
- add new marker  `category_performance` which runs the memory leak tests as separate job of the python test suite 

You can find different plots [here](https://www.notion.so/rasa/Create-CI-check-to-detect-training-memory-leaks-118d0e2f9cba4aa2aa0fd5e187a1ab78#ff09cf92a23f4b6e8b0cbac170983162)


wochinge:
It's not passing yet, but would be great to get some feedback @joejuzl @twerkmeister . To be honest I somewhat doubt the the sense of running this on infrastructure which we don't control.

joejuzl:
> To be honest I somewhat doubt the the sense of running this on infrastructure which we don't control.

This is why I would opt for a simple threshold. If memory usage is higher than expected on _any_ infrastructure it indicates a problem. Also if any of our test jobs go over 4gb they are going to fail anyway - so we should catch it earlier.

twerkmeister:
Will look into it tomorrow morning!

twerkmeister:
@wochinge I think this is a great additional testing tool 🥇 ! Will have to think about implications and possible edge cases a bit more, but I think this is a really good start for catching memory leaks. I assume we will add additional cases and tweaks over time. 

Only one limitation came to my mind so far. Afaik the memory profiler only concerns itself with the main memory of the machine, but not the graphics card memory. So if you run this on a machine with the necessary hardware and software for our neural network models, these might get pushed on the graphics card and you won't see that memory with the profiler. I think this is not a concern for the ci infrastructure, but if we want it to work in a similar fashion across all sorts of machines, this could be a pitfall. If we care about this, you could possibly set some config / env variable in the tests that would force models onto the CPU thus making them visible to the profiler.

twerkmeister:
Just ran `make test-performance` locally on my Laptop. It worked and took about 9 minutes which seems in line with the 11min it took on the ci infrastructure

twerkmeister:
Wondering whether we should have real data for the nlu case? Maybe a very small subset of multiwoz? I think this would cover some additional ground when components are susceptible to leaks with regards to number of different words or patterns in the data.

virtualroot (approved):

tests/test_memory_leak.py
```diff
@@ -0,0 +1,169 @@
...
+    ) -> None:
+        # Run as separate process to avoid other things affecting the memory usage.
+        # Unfortunately `memory-profiler` doesn't work properly with
+        # `multiprocessing.Process` as it can't handle the process exit
+        process = subprocess.Popen(
+            [
+                sys.executable,
+                "-c",
+                (
+                    f"from tests.test_memory_leak import {self.__class__.__name__}; "
```
joejuzl, line 69: Is it possible to get this bit `from tests.test_memory_leak` dynamically too?

tests/test_memory_leak.py
```diff
@@ -0,0 +1,169 @@
...
+        max_memory_usage = max(results, key=lambda memory_time: memory_time[0])[0]
+        assert max_memory_usage < self.max_memory_threshold_mb
+
+    @staticmethod
+    def _write_results(base_name: Text, results: List[Tuple[float]]) -> None:
+        mprof_plot = Path(f"{base_name}_plot.txt")
+        mprof_results = Path(f"{base_name}_raw.json")
+
+        # plot this via `mprof plot mprof_result.txt`
+        with open(mprof_plot, "w+") as f:
```
joejuzl, line 103: Why `"w+"`?

joejuzl (approved):
Nice job! Really like the way you've made the generic class 👍

PR #7589, opened 2020-12-17
Refactor feature combining layers in DIET and TED

samsucik:
**Proposed changes**:
- Move feature-processing logic used by TED, DIET and ResponseSelector into separate layer components (`rasa_layers`) as part of e2e leftovers (#6670); closes #7187. These layers, unlike those in `layers.py`, are meant to be used only with Rasa models (hence they also consume and utilise model configs). They fit between the model-specific code like `ted_policy.py` or `models.py` and the model-agnostic code like in `layers.py`. The newly introduced layers form a pipeline that goes like this (from bottom to top level):
  - `ConcatenateSparseDenseFeatures` combines multiple sparse and dense feature tensors into one.
  - `RasaFeatureCombiningLayer` additionally combines sequence-level and sentence-level features.
  - `RasaSequenceLayer` (only used for attributes with sequence-level features) additionally embeds the combined features with fully connected layers and a transformer, and facilitates masked language modeling.
- Add unit tests for the new layers, see [this comment](https://github.com/RasaHQ/rasa/pull/7589#issuecomment-801031998) for an overview.
- Fix a small bug in the masked language modelling implementation (which perhaps wasn't causing any serious degradation in model performance).

**To-do:**
- with ideas/help from engineering:
  - [x] improve naming re `_prepare_sequence_sentence_concat` (see [this comment](https://github.com/RasaHQ/rasa/pull/7589#discussion_r574446359) for context)
  - [x] get the new unit tests (testing the ML layer classes) to a state where we agree that they're good examples worth following
  - [x] agree on using constant names vs their values in docstrings (see [this Slack message](https://rasa-hq.slack.com/archives/C36SS4N8M/p1616171044244700?thread_ts=1614265007.143600&cid=C36SS4N8M)). Decision: either is acceptable, staying with constants' values.
- [x] add type annotations to added unit tests
- [ ] ~update `MINIMUM_COMPATIBLE_VERSION` (due to model-breaking changes)~ moved to #8498 

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] ~updated the documentation~
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/master/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #8254, opened 2021-03-19
Fix test_server.py:test_pushing_event occasionally failing

samsucik:
**Proposed changes**:
The test case sometimes failed due to the code executing faster than expected, see [this failure](https://github.com/RasaHQ/rasa/pull/7589/checks?check_run_id=2138461367#step:12:105) for instance. I fixed this.


wochinge (approved):
Thank you 🙌🏻

PR #8257, opened 2021-03-19
Simplify drop_intents_below_freq

Mazyod:
**Proposed changes**:
`drop_intents_below_freq` suffers from a bug where not all `TrainingData` content is being copied over to the newer object. The proposal fixes this issue and simplifies the code.


Mazyod:
I am currently cloning Rasa repo to run black and see if I can add some unit tests, however, the cloning process is extremely slow.

rasa/nlu/test.py
```diff
@@ -166,17 +166,12 @@ def drop_intents_below_freq(
...
-    return TrainingData(
-        keep_examples,
-        training_data.entity_synonyms,
-        training_data.regex_features,
-        responses=training_data.responses,
+    examples_per_intent = training_data.number_of_examples_per_intent.items()
+    intents_below_cutoff = {i for i, c in examples_per_intent if c < cutoff}
+
+    return training_data.filter_training_examples(
+        lambda ex: ex.get(INTENT) not in intents_below_cutoff
```
Mazyod, line 174: Alternatively...

```suggestion
    return training_data.filter_training_examples(
        lambda ex: examples_per_intent[ex.get(INTENT)] >= cutoff
```

sara-tagger:
Thanks for submitting a pull request 🚀 @wochinge will take a look at it as soon as possible ✨

Ghostvv (commented):
great change! the fix looks good. Could you please add tests where old code fails, but the new one works

Mazyod:
@Ghostvv I have update the existing test such that it fails on main, and succeeds on my patch. Not sure if I should've included the lookups in the json file or just inject them as I did.

Ghostvv (approved):
looks good to me

PR #8265, opened 2021-03-22
Fix custom facebook

wavymazy:
**Proposed changes**:
- Issues : https://github.com/RasaHQ/rasa/issues/5657
- This issue happens because of the custom response templates. If the custom template returns a list of dictionaries instead of a dictionary (json like object) the bot would not respond back to the facebook user.
- Therefore the change I made is to check if the json object is a Dict and if not we simply use the recipient_id that is given as a parameter of the function.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 19.10b0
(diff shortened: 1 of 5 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @koaning will take a look at it as soon as possible ✨

PR #8272, opened 2021-03-22
Fix sanic docs link

koaning:
There were some doc tests breaking because sanic moved their documentation. This PR addresses the bad links on our docs.


koaning:
@wochinge I'm checking because I know the CI changed a lot recently. Is the `Wait for docs tests` bug something I need to worry about here? It feels like this is a CI thing that's breaking, not my change, but I could be mistaken.

wochinge (approved):
💯

wochinge:
weird - let's rerun

Issue #8074, opened 2021-03-01
Warn users if their tests contain bot utterances not found in training data


koernerfelicia:
**Description of Problem**:
If test stories contain bot text not found in training e2e will not be able to correctly predict them. We should output a warning to users.

**Overview of the Solution**:
Warn users if their test stories contains a bot utterance that isn't found in the training data.

**Definition of Done**:
<!-- What needs to be there to consider this feature as done?
- [ ] Tests are added

PR #8280, opened 2021-03-23
Add warning

JEM-Mosig:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/8074

- ~~updated the documentation~~


chdorner:
Might be better to get a review from the enable squad on this one, I'm tagging @joejuzl but feel free to select somebody else

changelog/8074.improvement.md
```diff
@@ -0,0 +1 @@
+Rasa test now prints a warning if the test stories contain bot utterances that are not part of the domain.
```
joejuzl, line 1: From a user perspective these are from test stories, not the domain right?

rasa/shared/core/generator.py
```diff
@@ -651,6 +651,15 @@ def _process_step(
 
         end_trackers = []
         for event in events:
+            if (
```
joejuzl, line 654: Is this method only called from `rasa test`?

joejuzl (changes requested):

changelog/8074.improvement.md
koernerfelicia: *train stories

koernerfelicia: Although I guess from a user perspective that might be confusing as well, since this problem is unique to e2e - we always warned users if their bot utterances were not defined, like this:

`Action 'utter_1' is listed as a response action in the domain file, but there is no matching response defined`

rasa/shared/core/generator.py
JEM-Mosig: No, it is also called during training and validation. But in those cases the action_texts would have been read from the training stories, this this `if` doesn't get triggered.

changelog/8074.improvement.md
JEM-Mosig: Wait, no this really is about the test stories, right? The action_text comes from the training stories, and the present PR is about warning the user when the test stories contain an action_text that did not occur in the training set.

koernerfelicia: Eek, yes, I'm wrong. Sorry about that, I got mixed up.

koernerfelicia: Does this mean we aren't checking whether the `action_names` in test stories are defined?
So if I have a test story like:
```
stories:
- story: ...
  steps:
    ...
    - bot: utter_1
    ...
 ```
 and `utter_1` isn't defined in the domain there's no warning?

Comment (JEM-Mosig) [reply changelog/8074.improvement.md:1]: In your example, you would now get a warning, because `utter_1` is an `action_text` (not an `action_name`). The `action_text`s don't appear in the domain anyway.

**Review by joejuzl — APPROVED** [review]

**Comment by JEM-Mosig** [comment]
@koernerfelicia Would you be able to review this week?

**Comment by koernerfelicia** [comment]
@JEM-Mosig, yes, sorry, I thought I had already reviewed it. Taking another look now!

File: rasa/core/test.py [hunk 3]
```diff
@@ -12,6 +12,7 @@
 from rasa.shared.exceptions import RasaException
 from rasa.shared.nlu.training_data.message import Message
 import rasa.shared.utils.io
+import rasa.shared.utils.cli
```
koernerfelicia, line 15: Is this used somewhere?

koernerfelicia (commented):
LGTM, only thing I'm not sure about is the import in `core/test.py`, is this needed?

rasa/core/test.py
JEM-Mosig: Good catch! I don't think it is!

Issue #7685, opened 2021-01-06
Improve handling of multiple entity extractors in config


samsucik:
**Description of Problem**:
Currently, having multiple entity extractors in the NLU pipeline in the config file can lead to surprising behaviour: an entity being extracted multiple times, e.g. user message `I'll travel to Edinburgh` can appear in interactive learning as `I'll travel to [Edinburgh](city)[Edinburgh](city)` (see also #7533 for an example).

**Overview of the Solution**:
Using multiple extractors can lead to this kind of a surprise, but it doesn't have to. For instance, if `DucklingHTTPExtractor` is used to extract `time` and `date` entities, and `CRFEntityExtractor` is trained on annotated entities `city` and `cuisine`, then these extractors should never extract the same thing.

Therefore, we should allow multiple extractors, but we should also warn the user appropriately, in particular when there are multiple extractors being trained on user data (because then these extractors can "clash" at prediction time).

Importantly, we should check that each entity is displayed correctly in interactive learning (and exported into data files) when it's extracted by multiple extractors -- i.e. in the above example, we want to show only `I'll travel to [Edinburgh](city)`.

*A peculiar pathological case which we might want to discuss elsewhere is when 2 extractors extract 2 different things from the same word, but I haven't verified this can actually happen in reality...*

**Examples** (if relevant):
See #7533.

**Definition of Done**:
- [ ] The config gets checked for multiple potentially clashing extractors and appropriate warning is issued
- [ ] Entities extracted multiple times are displayed correctly
- [ ] Tests are added
- [ ] Docs mention what happens when multiple extractors are used
- [ ] Feature mentioned in the changelog

twerkmeister:
related to #7490

twerkmeister:
I made some comments on this problem in the previously linked issue.

The biggest issue is probably two entity extractors looking for the same type of entities as you outlined. Here we can warn people if they use multiple extractors that just relate to the training data, like you using `DIETClassifier`and `CRFEntityExtractor` together. Also we can add a warning if someone uses regexes + `RegExEntityExtractor` for the same types that they use DIET or CRF for.

Thinking about it a bit more, however, even entities like `date` and `meal` could overlap as in `I'd like to order the monday special` where the meal here might be `monday special` and some date or time entity `monday`. Or you use duckling for numbers and also have another extractor for addresses. Probably really difficult to know with certainty which entities might clash. Here we can add a runtime warning whenever there are overlapping entities.

[...]

Issue #7490, opened 2020-12-08
Double entity extraction


desmarchris:
**Rasa version**: 2.1.2

**Rasa SDK version** (if used & relevant):

**Rasa X version** (if used & relevant):

**Python version**: 3.8.6

**Operating system** (windows, osx, ...): ubuntu

**Issue**: After adding `RegexEntityExtractor` to my pipeline and using a regex to collect email, a detected email is returned twice (once from DIET, once from regex?). In this case, I think the DIET entity detection should be ignored if it's matching a return from regex, or if a regex is returned at all.

**Error (including full traceback)**:
```
I can help you find the status of your order. What's your email address so I can find the order?
Your input ->  it's example@gmail.com
Based on the latest order from ['example@gmail.com', 'example@gmail.com'], it looks like your order should arrive in a few days.
```

Response: 
```
utter_order_status:
  - text: Based on the latest order from {email}, it looks like your order should arrive in a few days.
```

Rules: 
```
- rule: Activate email form for order status
  steps:
  - intent: order_status
  - action: order_status_form
  - active_loop: order_status_form
- rule: Submit order status form
  condition:
  - active_loop: order_status_form
  steps:
  - action: order_status_form
  - active_loop: null
  - slot_was_set:
    - email
  - action: utter_order_status
```

**Content of configuration file (config.yml)** (if relevant):
```yml
pipeline:
# # No configuration for the NLU pipeline was provided. The following default pipeline was used to train your model.
# # If you'd like to customize it, uncomment and adjust the pipeline.
# # See https://rasa.com/docs/rasa/tuning-your-model for more information.
  - name: WhitespaceTokenizer
  - name: RegexFeaturizer
  - name: LexicalSyntacticFeaturizer
  - name: CountVectorsFeaturizer
  - name: CountVectorsFeaturizer
    analyzer: char_wb
    min_ngram: 1
    max_ngram: 4
  - name: DIETClassifier
    epochs: 100
  - name: EntitySynonymMapper
  - name: RegexEntityExtractor
  - name: ResponseSelector
    epochs: 100
  - name: FallbackClassifier
    threshold: 0.3
    ambiguity_threshold: 0.1
```

**Content of domain file (domain.yml)** (if relevant):
```yml

```

sara-tagger:
Thanks for the issue, @m-vdb will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

[...]

PR #8289, opened 2021-03-24
7685 warn of competing entity extractors

twerkmeister:
**Proposed changes**:
- Added warnings when pipelines contain multiple statistical entity extractors
- Added a warning when statistical and regex entity extractors might go for the same entities
- Added info boxes in documentation to raise awareness for multiple extraction issue and solutions.
- Warning for overlapping entities during inference time. I imagined it would be fairly annoying to constantly get this warning if you have a reasonable use case for this. So I made it such the warning is only issued once 
- Addresses #7685 
- Addresses #7490

**Open questions**
- I think it is a bit odd that the `RegexEntityExtractor` only uses those entities that are also annotated in the example sentences. [This behavior is hinted at in the docs](https://rasa.com/docs/rasa/components#regexentityextractor). As a result, if you use a statistical extractor like the CRF, Mitie, or DIET, you always have to add the Regex Entities also to these entity extractors training domain, which invites competition / conflict. 


samsucik:
> I think it is a bit odd that the RegexEntityExtractor only uses those entities that are also annotated in the example sentences. This behavior is hinted at in the docs. As a result, if you use a statistical extractor like the CRF, Mitie, or DIET, you always have to add the Regex Entities also to these entity extractors training domain, which invites competition / conflict.

This would be resolved by being able to specify for each extractor which entities it should extract, right? I.e. adding this feature to DIET/CRF/MITIE. I'm sure there's an issue open for this already; sounds like a reasonable feature request to me...

docs/docs/components.mdx
```diff
@@ -1743,7 +1743,28 @@ See [starspace paper](https://arxiv.org/abs/1709.03856) for details.
...
+
+  Another, less obvious case of duplicate/overlapping extraction can happen even if extractors focus on different
+  entity types. Imagine a food delivery bot and a user message like `I would like to order the Monday special`.
+  Hypothetically, if your time extractor's performance isn't very good, it might extract `Monday` here as a time for the order,
+  and your other extractor might extract `Monday special` as the meal.
+  If you struggle with overlapping entities of this sort, it might make sense to add additional training data
+  to improve your extractor. If that does not suffice, you can add a
+  [custom component](components.mdx#custom-components) that resolves conflicts in entity
+  extraction according to your own logic.
+:::
```
samsucik, line 1767: To be honest, I don't know how to think about this long note (and partly about the other one below too). If the note is only under the general Entity extractors heading, then people might miss it (if they click through to component-specific sections). The other thing is that the note is very long. I guess I find the second note below much more helpful, and perhaps this general note could just very briefly:
- say what the issue is with using multiple extractors, and then
- point folks to the more involved reading in component-specific sections

But then it's probably worth having a short note in each component-specific section that'd mention the issue (again, very briefly) & what the current component's role in the issue is, and then perhaps point to one big note (listed e.g. under `RegexEntityExtractor`) which would actually explain the workarounds...

samsucik (changes requested):
Thanks for the changes @twerkmeister! I mostly pointed out minor issues, but the real challenge to me remains: How to communicate this well to users. As with so many other edge cases, I'm struggling to see a clear compromise between having easy-to-read, concise docs vs providing the complete info, some of which most (but not all!) users will never need...

rasa/shared/nlu/training_data/message.py
```diff
@@ -398,3 +398,18 @@ def is_e2e_message(self) -> bool:
...
+
+    def find_overlapping_entities(
+        self,
+    ) -> List[Tuple[Dict[Text, Any], Dict[Text, Any]]]:
+        """Finds any overlapping entity annotations."""
+        entities = self.get("entities", [])[:]
+        entities.sort(key=lambda e: e["start"])
+        overlapping_pairs: List[Tuple[Dict[Text, Any], Dict[Text, Any]]] = []
+        for i, entity in enumerate(entities):
+            for other_entity in entities[i + 1 :]:
```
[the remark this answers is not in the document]
twerkmeister:
```
>>> l = [1, 2, 3]
>>> l[1:]
[2, 3]
>>> l[2:]
[3]
>>> l[3:]
[]
>>> l[4:]
[]
>>> l[5:]
[]
>>>
```
The slicing operator simply returns an empty list if you request a non-existent slice. With an empty list, the second loop won't be started. So the code looks good to me here

data/test/overlapping_regex_entities.yml
```diff
@@ -0,0 +1,29 @@
...
+    - search for restaurants
+    - anywhere in the [west](location)
+    - anywhere near [18328](zipcode)
+    - I am looking a restaurant in [29432](zipcode)
+    - I am looking for [mexican indian fusion](cuisine)
+    - [central](location) [indian](cuisine) restaurant
+    - What [pizza](meal) places have been [buzzing](zz-words) lately?
+- regex: zipcode
+  examples: |
+    - [0-9]{5}
```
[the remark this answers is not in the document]
twerkmeister: not needed indeed, simplified it

rasa/nlu/components.py
```diff
@@ -322,6 +340,69 @@ def validate_required_components_from_data(
...
+        rasa.shared.utils.io.raise_warning(
+            f"You have defined multiple entity extractors that do the same job"
+            f"in your pipeline: "
+            f"{', '.join(extractors_in_pipeline)}. This can lead to multiple "
+            f"extraction of entities. Please read the documentation section"
+            f"on entity extractors to make sure you understand the implications."
+        )
+
+
+def warn_of_competing_regex_and_statistical_extractor(
```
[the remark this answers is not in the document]
twerkmeister: changed to `warn_of_competition_with_regex_extractor` as this method is explicitly checking for conflict with the regex extractor

rasa/shared/nlu/training_data/message.py
```diff
@@ -398,3 +398,18 @@ def is_e2e_message(self) -> bool:
...
+                if other_entity["start"] < entity["end"]:
+                    overlapping_pairs.append((entity, other_entity))
+                else:
+                    break
```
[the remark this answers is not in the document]
twerkmeister: It actually collects all overlapping pairs. The break is here since the list is ordered: once you find an `other_entity` which starts after your `entity` ends, there can't be any `other_entity` which overlaps with `entity` anymore. So you move on to the next  `entity`.

twerkmeister:
Hey @samsucik thanks for the review and comments. I very much agree that we are walking a fine line here. Most people will never run into these issues with multiple entity extraction. I think the way the docs are written it is now, it isn't perfect, but I think we are at least giving some more detail to people who have trouble with the entity extraction. I have added links to the warning that will take people directly to the respective documentation sections. So those who run into the warnings will get some detail. And the others might never notice

docs/docs/components.mdx
twerkmeister: For now I would keep it as it is. I can't really think of a way which is obviously far superior to this. I have added specific links to the warnings so people won't have to search for the info themselves but will land at these more detailed explanations when they are actually at risk of this issue of duplicate extraction

rasa/shared/nlu/training_data/message.py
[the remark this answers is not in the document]
samsucik: Oh, okay, my bad!

rasa/shared/nlu/training_data/message.py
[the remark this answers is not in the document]
samsucik: Makes sense, thanks 🙂

docs/docs/components.mdx
samsucik: Alright. @twerkmeister do you know if the CI checks check for out-dated links in the code? Strikes me as an important thing if we include a link in a warning (particularly because I'm not aware of cases where we'd include links like this, evn though I'm a fan of the idea).

samsucik (approved):
Looks good to me, though I'm not an expert in all the best practices and patterns followed in the non-research codebase parts...

data/test/overlapping_regex_entities.yml
```diff
@@ -0,0 +1,12 @@
...
+- intent: restaurant_search
+  examples: |
+    - What [pizza](meal) places have been [buzzing](zz-words) lately?
+- regex: zz-words
+  examples: |
+    - \w*zz\w*
+- lookup: meal
+  examples: |
+    - pizza
+    - pasta
```
wochinge, line 12:
```suggestion
    - pasta

```

docs/docs/components.mdx
```diff
@@ -1743,7 +1743,28 @@ See [starspace paper](https://arxiv.org/abs/1709.03856) for details.
...
+:::note
+  If you use multiple entity extractors, we advise that each extractor targets an exclusive
+  set of entity types. For example, use [Duckling](components.mdx#ducklinghttpextractor) to extract dates and times, and
+  [DIETClassifier](components.mdx#dietclassifier-1) to extract person names. Otherwise, if multiple extractors
+  target the same entity types, it is very likely that entities will be extracted multiple times.
+
+  For example, if you use two or more general purpose extractors like [MitieEntityExtractor](components.mdx#mitieentityextractor),
+  [DIETClassifier](components.mdx#dietclassifier-1), or [CRFEntityExtractor](components.mdx#crfentityextractor),
+  the entity types in your training data will be found and
+  extracted by all of them. If the [slots](domain.mdx#slots) you are filling with your entity types are of type text,
```
wochinge, line 1755:
```suggestion
  extracted by all of them. If the [slots](domain.mdx#slots) you are filling with your entity types are of type `text`,
```

docs/docs/components.mdx
```diff
@@ -1743,7 +1743,28 @@ See [starspace paper](https://arxiv.org/abs/1709.03856) for details.
...
+  then the last extractor in your pipeline will win. If the slot is of type list, then all results
```
wochinge, line 1756:
```suggestion
  then the last extractor in your pipeline will win. If the slot is of type `list`, then all results
```

docs/docs/components.mdx
wochinge: It doesn't check in the code.

rasa/nlu/components.py
```diff
@@ -208,18 +209,36 @@ def validate_pipeline(pipeline: List["Component"]) -> None:
...
 
-def any_components_in_pipeline(components: Iterable[Text], pipeline: List["Component"]):
+def any_components_in_pipeline(
+    components: Iterable[Text], pipeline: List["Component"]
+) -> bool:
     """Check if any of the provided components are listed in the pipeline.
 
     Args:
-        components: A list of :class:`rasa.nlu.components.Component`s to check.
+        components: A list of str of component class names to check
```
wochinge, line 218: optional: I prefer to not duplicate type in type annotation and docstrings.
```suggestion
        components: Component class names to check.
```

rasa/nlu/components.py
```diff
@@ -208,18 +209,36 @@ def validate_pipeline(pipeline: List["Component"]) -> None:
...
+
+    Args:
+        components: A list of str of component class names to check.
+        pipeline: A list of :class:`rasa.nlu.components.Component`s.
+
+    Returns:
+        A list of str of component class names that are present in the pipeline.
+    """
+    pipeline_component_names = set([c.name for c in pipeline])
+    return list(pipeline_component_names.intersection(components))
```
wochinge, line 241: do we actually need the string cast?

rasa/nlu/components.py
```diff
@@ -322,6 +341,70 @@ def validate_required_components_from_data(
...
+        "DIETClassifier",
+        "CRFEntityExtractor",
+        "MitieEntityExtractor",
+    ]
+    extractors_in_pipeline = find_components_in_pipeline(
+        general_purpose_entity_extractors, pipeline
+    )
+    if len(extractors_in_pipeline) > 1:
+        rasa.shared.utils.io.raise_warning(
+            f"You have defined multiple entity extractors that do the same job"
```
wochinge, line 361:
```suggestion
            f"You have defined multiple entity extractors that do the same job "
```

rasa/nlu/components.py
```diff
@@ -322,6 +341,70 @@ def validate_required_components_from_data(
...
+            f"in your pipeline: "
+            f"{', '.join(extractors_in_pipeline)}. This can lead to the same entity getting"
```
wochinge, line 363:
```suggestion
            f"{', '.join(extractors_in_pipeline)}. This can lead to the same entity getting "
```

rasa/nlu/components.py
```diff
@@ -322,6 +341,70 @@ def validate_required_components_from_data(
...
+            f"extracted multiple times. Please read the documentation section"
+            f"on entity extractors to make sure you understand the implications:"
```
wochinge, line 365:
```suggestion
            f"extracted multiple times. Please read the documentation section "
            f"on entity extractors to make sure you understand the implications: "
```

rasa/nlu/components.py
```diff
@@ -322,6 +341,70 @@ def validate_required_components_from_data(
...
+        general_entity_extractors, pipeline
+    )
+    has_general_extractors = len(present_general_extractors) > 0
+    has_regex_extractor = any_components_in_pipeline(["RegexEntityExtractor"], pipeline)
+
+    regex_entity_types = {rf["name"] for rf in data.regex_features}
+    overlap_between_types = data.entities.intersection(regex_entity_types)
+    has_overlap = len(overlap_between_types) > 0
+
+    if has_general_extractors and has_regex_extractor and has_overlap:
```
wochinge, line 395: isn't this check enough?
```suggestion
    if has_overlap:
```

rasa/nlu/components.py
```diff
@@ -322,6 +341,70 @@ def validate_required_components_from_data(
...
+        rasa.shared.utils.io.raise_warning(
+            f"You have an overlap between the RegexEntityExtractor and the "
+            f"statistical entity extractors {', '.join(present_general_extractors)}"
```
wochinge, line 398:
```suggestion
            f"statistical entity extractors {', '.join(present_general_extractors)} "
```

tests/nlu/test_components.py
```diff
@@ -220,3 +222,144 @@ async def test_validate_component_keys_raises_warning_on_invalid_key(
...
+            [
+                {"name": "WhitespaceTokenizer"},
+                {"name": "LexicalSyntacticFeaturizer"},
+                {"name": "DIETClassifier"},
+            ],
+            False,
+        ),
+    ],
+)
+def test_warn_of_competing_extractors(pipeline_template, should_warn):
```
wochinge, line 249: can you please add type annotations to the tests?

tests/nlu/test_components.py
```diff
@@ -220,3 +222,144 @@ async def test_validate_component_keys_raises_warning_on_invalid_key(
...
+    trainer = Trainer(config)
+
+    if should_warn:
+        with pytest.warns(UserWarning):
+            rasa.nlu.components.warn_of_competing_extractors(trainer.pipeline)
+    else:
+        with pytest.warns(None) as records:
+            rasa.nlu.components.warn_of_competing_extractors(trainer.pipeline)
+
+        assert len(records) == 0
```
wochinge, line 260: optional: I'd favor to have to separate test cases of this instead of `if ... else` for the asserts.

rasa/nlu/model.py
```diff
@@ -472,3 +483,26 @@ def featurize_message(self, message: Message) -> Message:
...
+        This warning is only issued once per Interpreter life time.
+        """
+        overlapping_entity_pairs = message.find_overlapping_entities()
+        if len(overlapping_entity_pairs) > 0:
+            message_text = message.get("text")
+            first_pair = overlapping_entity_pairs[0]
+            entity_1 = first_pair[0]
+            entity_2 = first_pair[1]
+            rasa.shared.utils.io.raise_warning(
+                f"Parsing of message: '{message_text}' lead to overlapping"
```
wochinge, line 499:
```suggestion
                f"Parsing of message: '{message_text}' lead to overlapping "
```

rasa/shared/nlu/training_data/message.py
```diff
@@ -398,3 +398,18 @@ def is_e2e_message(self) -> bool:
             (self.get(ACTION_TEXT) and not self.get(ACTION_NAME))
             or (self.get(TEXT) and not self.get(INTENT))
         )
+
+    def find_overlapping_entities(
+        self,
+    ) -> List[Tuple[Dict[Text, Any], Dict[Text, Any]]]:
+        """Finds any overlapping entity annotations."""
+        entities = self.get("entities", [])[:]
+        entities.sort(key=lambda e: e["start"])
```
wochinge, line 407: I think some entities might not have a `start` key

changelog/7685.improvement.md
```diff
@@ -0,0 +1,2 @@
+* Added warnings for competing entity extractors at training time and for overlapping entities at inference time
```
wochinge, line 1: Mhm, I think this could breeak the generation of the changelog file. I think this would at least need a title before starting the bullet pointed list.

wochinge (commented):
thanks for the change! 

Do we really need the runtime warning? 👀 Wondering how many people are actually looking at them - especially if this warning only appears once.

docs/docs/components.mdx
twerkmeister: there are some warnings that use the links. That's why there are all these doc link constants in `rasa/shared/constants.py`. So the main doc links should work whenever the docs url is changed and the code is changed. But the anchor might be more prone to change and it won't be so easily noticable. That's a fair point. hmm

rasa/shared/nlu/training_data/message.py
twerkmeister: oh really, damn, do you know in which instances this might happen?

rasa/nlu/components.py
twerkmeister: Then it could for example trigger if you have no regex extractor but regex features. Which would be a false positive. Or say you have the regex extractor but no general extractor.

tests/nlu/test_components.py
twerkmeister: Would make the code a bit simpler, but you would add duplicate some more lines. Don't have a strong preference here

rasa/nlu/components.py
twerkmeister: You meant set cast right? I simplified that method a bit

twerkmeister:
> thanks for the change!
> 
> Do we really need the runtime warning? 👀 Wondering how many people are actually looking at them - especially if this warning only appears once.

The runtime warning would help if there's some combination that we are not catching up front or if people don't pay attention to the warnings at training time. Ah well.

rasa/nlu/components.py
wochinge: Sorry, meant the `list(...)` cast 🙈

rasa/nlu/components.py
```diff
@@ -322,6 +341,70 @@ def validate_required_components_from_data(
...
+) -> None:
+    """Warns when regex entity extractor is competing with a general one.
+
+    This might be the case when the following conditions are all met:
+    * You are using a general entity extractor and the `RegexEntityExtractor`
+    * AND you have regex patterns for entity type A
+    * AND you have annotated text examples for entity type A
+    """
+    general_entity_extractors = [
+        "DIETClassifier",
```
wochinge, line 381: can we use constants for these? E.g. like [here](https://github.com/RasaHQ/rasa/blob/8258c5500108110525aef16fa20ee0f47f435b1e/rasa/shared/core/constants.py#L89)?

rasa/shared/nlu/training_data/message.py
wochinge: I think mostly custom components: https://github.com/RasaHQ/rasa/issues/8114

rasa/shared/nlu/training_data/message.py
```diff
@@ -398,3 +398,18 @@ def is_e2e_message(self) -> bool:
...
+        overlapping_pairs: List[Tuple[Dict[Text, Any], Dict[Text, Any]]] = []
+        for i, entity in enumerate(entities):
+            for other_entity in entities[i + 1 :]:
+                if other_entity["start"] < entity["end"]:
```
wochinge, line 411: can we use the constants which we have for `start` and `end`?

tests/nlu/test_components.py
wochinge: In my opinion fine to have duplicate code in tests and rather make them self contained and concise. However, fine for me.

tests/nlu/test_components.py
```diff
@@ -220,3 +222,146 @@ async def test_validate_component_keys_raises_warning_on_invalid_key(
...
+    else:
+        with pytest.warns(None) as records:
+            rasa.nlu.components.warn_of_competition_with_regex_extractor(
+                trainer.pipeline, training_data
+            )
+
+        assert len(records) == 0
+
+
+async def test_warn_for_overlapping_entities(tmp_path: Path):
```
wochinge, line 330: I'd be in favor of splitting this test following the paradigm of "Unit Tests Should Verify a Single-Use Case" (https://dzone.com/articles/unit-testing-best-practices-how-to-get-the-most-ou) into
* `test_warn_for_overlapping_entities`
* `test_no_duplicate_warning_for_overlapping_entities`

This makes it easier to find the issue when the test fails instead of having to understand the test first.

rasa/nlu/model.py
```diff
@@ -472,3 +483,26 @@ def featurize_message(self, message: Message) -> Message:
             if not isinstance(component, (EntityExtractor, IntentClassifier)):
                 component.process(message, **self.context)
         return message
+
+    def warn_of_overlapping_entities(self, message: Message) -> None:
+        """Issues a warning when there are overlapping entity annotations.
+
+        This warning is only issued once per Interpreter life time.
```
wochinge, line 490: can we please also describe the `message` input param?

rasa/nlu/components.py
```diff
@@ -322,6 +341,70 @@ def validate_required_components_from_data(
...
+            f"{DOCS_URL_COMPONENTS}#entity-extractors"
+        )
+
+
+def warn_of_competition_with_regex_extractor(
+    pipeline: List["Component"], data: TrainingData
+) -> None:
+    """Warns when regex entity extractor is competing with a general one.
+
+    This might be the case when the following conditions are all met:
```
wochinge, line 375: can we please also describe the input params to follow the google docstyle standards?

rasa/nlu/components.py
twerkmeister: Would you be okay with creating an extra issue for this as a potential improvement? The names as strings are used throughout the components file. I think this is because the components themselves derive their name property directly from their class name. If you create constants you also need to have another mechanism that makes sure the constants stay in sync with the class names. So it's a relevant change, but a bit more involved than just a side note inside this PR

rasa/shared/nlu/training_data/message.py
twerkmeister: Okay, only using entities here for overlap search that have a start and end

tests/nlu/test_components.py
twerkmeister: done

rasa/nlu/components.py
twerkmeister: Also the component names are used in many many more places. Like when creating pipelines for test code

wochinge: Sure - could you please create an issue for this then?

wochinge (approved):
💯

rasa/nlu/components.py
twerkmeister: yup: https://github.com/RasaHQ/rasa/issues/8455

athulvingt:
I used `RegexEntityextractor` so it would look for entity in look up, I had to make `entityrecognition : false` in DIETClassifier so as to avoid conflict between extractors, Also synonyms isn't working. Can you suggest a extractor or combination of some extractors and mappers such that entity extraction, lookup and synonyms work well

twerkmeister:
Best would be to use your lookup with [RegexFeaturizer](https://rasa.com/docs/rasa/components/#regexfeaturizer) which will provide information about the lookup in a way that DIET can use. Then you only have one extractor.

PR #8303, opened 2021-03-25
Delete rule source predicting action_listen from rule starting with action-intent pair 

### LINKED ISSUE #8176: Rule starting with action-intent pair wrongly contradicts stories
**Rasa version**: 2.3.4 (`main`)

**Python version**: 3.8.5

**Operating system** (windows, osx, ...): OSX

**Issue**:
I want to write a rule that basically conditions on an intent `i3` IF that intent follows after a particular action `utter_1` (see the training rules below). Unfortunately, such rule (very non-intuitively) also predicts `action_listen` after `utter_1`. Hence, if I have a training story that contains a different action after `utter_1`, it creates a contradiction (see the training stories below). Note that having two stories in which `utter_1` is followed once by an intent and the other time by an action works fine -- because the conversation history is different for these two cases. But as soon as I create the mentioned rule, things break.

**Suggested solution:**
See [this Slack thread](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1615466129437400) for more context and a discussion that led to this suggestion. Basically, if there is a rule that starts with an action followed by an intent and another action, we'll assume that it is only the last action that the user really meant to predict with the rule. Therefore, IF the predicted `action_listen` (which the user most likely doesn't know about) contradicts a story, we remove this particular rule (only the one that predicts the `action_listen`) from `RulePolicy`'s lookup and proceed (possibly also raising a warning).

```
InvalidRule:
Contradicting rules or stories found 🚨

- the prediction of the action 'utter_2' in story 'action after action' is contradicting with rule(s) 'conditioned on action' which predicted action 'action_listen'.
Please update your stories and rules so that they don't contradict each other.
You can find more information about the usage of rules at https://rasa.com/docs/rasa/rules.
```

```
rasa train core --augmentation 0
```

```yml
language: en
policies:
  - name: RulePolicy
  - name: TEDPolicy
    epochs: 1
```

```yml
intents:
- i1
- i2
- i3
- i4
actions:
- utter_1
- utter_2
- utter_3
```

**Training data**:
```yml
stories:
- story: action after action
  steps:
  - intent: i1
  - action: utter_1
  - action: utter_2 # action goes after utter_1
- story: intent after action
  steps:
  - intent: i2
  - action: utter_1
  - intent: i3
  - action: utter_2 # intent goes after utter_1 because of different history
  - intent: i4
  - action: utter_3
rules:
- rule: conditioned on action
  steps:
  - action: utter_1
  - intent: i3 # this intent is preceded by an implicit action_listen, which break things
  - action: utter_2

### PR DESCRIPTION
**Pull request by koernerfelicia**
closes #8176

### CODE REVIEW

File: rasa/core/policies/rule_policy.py [hunk 1]
```diff
@@ -519,6 +520,25 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
             )
         return loop_sources
 
+    def _remove_action_listen_contradiction(self, prediction_source: Text) -> None:
+        prev_action = json.loads(prediction_source)[-1][PREVIOUS_ACTION][ACTION_NAME]
```
Ghostvv, line 524: I don't think you need to do that `prediction_source` should be identical to one of the keys in rules lookup

rasa/core/policies/rule_policy.py
```diff
@@ -530,12 +550,17 @@ def _check_prediction(
             return []
 
         tracker_type = "rule" if tracker.is_rule_tracker else "story"
+        if tracker_type == "story" and predicted_action_name == ACTION_LISTEN_NAME:
```
Ghostvv, line 553:
```suggestion
        if not tracker.is_rule_tracker and predicted_action_name == ACTION_LISTEN_NAME:
```

Ghostvv: we need to check that previous action before this action_listen was unpredictable

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +520,25 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
...
+                    and states[-2][PREVIOUS_ACTION][ACTION_NAME] == prev_action
+                    and PREVIOUS_ACTION in states[-1]
+                    and states[-1][PREVIOUS_ACTION][ACTION_NAME] == ACTION_LISTEN_NAME
+                    and USER in states[-1]
+                ):
+                    # check that the rule pattern is: action, intent, prediction
+                    # delete the rule that predicts action_listen
+                    rules_to_delete.append(self._create_feature_key(states[:-1]))
+        for rule_to_delete in rules_to_delete:
+            self._rules_sources.pop(rule_to_delete)
```
Ghostvv, line 540: `self.lookup[RULES]`

rasa/core/policies/rule_policy.py
```diff
@@ -530,12 +550,17 @@ def _check_prediction(
             return []
 
         tracker_type = "rule" if tracker.is_rule_tracker else "story"
+        if tracker_type == "story" and predicted_action_name == ACTION_LISTEN_NAME:
+            self._remove_action_listen_contradiction(prediction_source)
```
Ghostvv, line 554:
```suggestion
            self.lookup[RULES].pop(prediction_source)
```

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +520,26 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
             )
         return loop_sources
 
+    def _should_delete(
+        self, prediction_source: Text, tracker: TrackerWithCachedStates
+    ) -> bool:
+        """
+        Checks whether this contradiction is due to action, intent pair.
```
Ghostvv, line 527:
```suggestion
        """Checks whether this contradiction is due to action, intent pair.
```

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +520,26 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
...
+
+        Args:
+            prediction_source: the states that result in the prediction
+            tracker: the tracker that raises the contradiction
+
+        Returns:
+            true if the contradiction is a result of an action, intent pair in the rule.
+
```
Ghostvv, line 535:
```suggestion
```

rasa/core/policies/rule_policy.py
```diff
@@ -44,6 +44,7 @@
     ACTIVE_LOOP,
     RULE_ONLY_SLOTS,
     RULE_ONLY_LOOPS,
+    USER,
```
Ghostvv, line 47:
```suggestion
```

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +520,26 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
...
+        Returns:
+            true if the contradiction is a result of an action, intent pair in the rule.
+
+        """
+        if tracker.is_rule_tracker or prediction_source.count(PREVIOUS_ACTION) > 1:
+            return False
+        for source in self.lookup[RULES]:
+            if prediction_source[:-2] in source and not prediction_source == source:
+                return True
+
```
Ghostvv, line 542:
```suggestion
        return False
```

Ghostvv: you need to `return False` in the end as well

rasa/core/policies/rule_policy.py
```diff
@@ -535,6 +556,9 @@ def _check_prediction(
             for rule_name, action_name in self._rules_sources[prediction_source]
             if action_name != gold_action_name
         }
+        if self._should_delete(prediction_source, tracker):
+            self.lookup[RULES].pop(prediction_source)
+            return []
```
Ghostvv, line 561: I'd move it to line `552`, to create a clear visual separation when there is no contradiction

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +519,24 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
...
+            tracker: the tracker that raises the contradiction
+
+        Returns:
+            true if the contradiction is a result of an action, intent pair in the rule.
+        """
+        if tracker.is_rule_tracker or prediction_source.count(PREVIOUS_ACTION) > 1:
+            return False
+        for source in self.lookup[RULES]:
+            if prediction_source[:-2] in source and not prediction_source == source:
+                return True
```
Ghostvv: could you please add comments, what these `if`s mean

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +519,24 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
...
+            prediction_source: the states that result in the prediction
+            tracker: the tracker that raises the contradiction
+
+        Returns:
+            true if the contradiction is a result of an action, intent pair in the rule.
+        """
+        if tracker.is_rule_tracker or prediction_source.count(PREVIOUS_ACTION) > 1:
+            return False
+        for source in self.lookup[RULES]:
+            if prediction_source[:-2] in source and not prediction_source == source:
```
Ghostvv, line 544: I think it's better to do
```suggestion
            if source.startswith(prediction_source[:-2]) and not prediction_source == source:
```
this way, we ensure, that it is the first dialogue step

rasa/core/policies/rule_policy.py
Ghostvv: I don't think you moved it

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +519,37 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
...
+        Returns:
+            true if the contradiction is a result of an action, intent pair in the rule.
+        """
+        if (
+            tracker.is_rule_tracker  # only apply to contradicting story, not rule
+            or prediction_source.count(PREVIOUS_ACTION)
+            > 1  # only apply for prediction after unpredictable action
+            or gold_action_name
+            != ACTION_LISTEN_NAME  # only apply for prediction of action_listen
+        ):
```
Ghostvv, line 543: I think this way it is a bit prettier
```suggestion
        if (
            # only apply to contradicting story, not rule
            tracker.is_rule_tracker
            # only apply for the first prediction in a rule
            or prediction_source.count(PREVIOUS_ACTION) > 1
            # only apply for prediction of action_listen
            or gold_action_name != ACTION_LISTEN_NAME
        ):
```

rasa/core/policies/rule_policy.py
```diff
@@ -519,6 +519,37 @@ def _handling_loop_sources(domain: Domain) -> Set[Text]:
...
+            return False
+        for source in self.lookup[RULES]:
+            if (
+                source.startswith(prediction_source[:-2])
+                and not prediction_source == source
```
Ghostvv, line 548: could you please also add comments, explaining this

koernerfelicia: Yep! I also added a test to test this, one to test `prediction_source.count(PREVIOUS_ACTION) > 1`, and one to test only applying this for stories. I think there is no point in testing not deleting for not action listen since if this broke, contradicting rules would break?

koernerfelicia: Not sure if I should combine the tests with mark parametrize or if that just makes them less readable because each have a different setup

tests/core/policies/test_rule_policy.py
```diff
@@ -2624,3 +2624,185 @@ def test_hide_rule_turn_with_loops_as_followup_action():
...
+        evts=[
+            UserUttered(intent={"name": intent_1}),
+            ActionExecuted(utter_1),
+            ActionExecuted(utter_2),
+        ],
+    )
+    policy = RulePolicy()
+    policy.train(
+        [rule, story], domain, RegexInterpreter(),
+    )
```
Ghostvv, line 2669: could you please also check that `policy.lookup` doesn't contain the rule

rasa/core/policies/rule_policy.py
Ghostvv: I think separate tests are better

Ghostvv: > I think there is no point in testing not deleting for not action listen since if this broke, contradicting rules would break?

what do you mean? I think given how the code is organized, it'd be good to test that second non listen action raises contradiction

koernerfelicia: Not sure what you mean? I was saying that I wrote tests for all the conditions in the if blocks of `_should_delete` (is a story, isn't after a predictable action, isn't the last action of the rule...) except for "is action_listen":
```
rule:
- action: A
- action: B #(NOT action_listen)
```
```
story:
- action: A
- action: C
```
(in this case, the `should_delete` method returns false because the contradicting action is not `action_listen`)
Reason I say this it's not necessary to test is because this is already covered by the test case `test_contradicting_rules`.

koernerfelicia: But now that I think about it, myabe it's good to have it anyways for symmetry. Was this what you meant by
> it'd be good to test that second non listen action raises contradiction
?

Ghostvv: yes, but if it is a duplicate, then no need to add it

koernerfelicia: I think it's not quite a duplicate, because `test_contradicting_rules_and_stories` has:
```
rule:
- intent
- action
- action_listen
```

and the test I added specifically has no action_listen

Ghostvv (approved):
🚀

Issue #6836, opened 2020-09-29
Partial dataset loading


tabergma:
**Description of Problem**:
If you want to train your assistant on a large dataset, it might happen that you run into memory issues as the complete dataset is loaded into memory and featurized. 

related to https://github.com/RasaHQ/rasa/issues/1621

**Overview of the Solution**:
No concrete solution yet. But ideally we would not load all the data at once but load them partly, so that not all data is loaded into memory and featurized at the same time.

Things to keep in mind:
- `CountVectorsFeaturizer` needs to build up a vocabulary. We could either use predefined vocabulary or we need to pass all the data once to build full vocabulary from all the data
- if we want to support continue training, than it makes sense to create placeholders for unseen words in `CountVectorsFeaturizer` vocabulary (https://rasa-hq.slack.com/archives/C8LEHMTAM/p1601390370106100?thread_ts=1601389639.104600&cid=C8LEHMTAM)

tabergma:
Let's collect ideas in the following [document](https://docs.google.com/document/d/1dwKfdju5ZqhOQtURLF3-eU3PdxvdlyXafPHE_0sLPzI/edit?usp=sharing).

Ghostvv:
@tabergma shouldn't it be rather in research projects? I don't think it is feasible in one sprint

tabergma:
True, it is not. But it is also not a research project. We could also leave it in the column "smaller projects" and just add the smaller issues related to this issues to the "in progress" column etc.

Ghostvv:
I'd move it to research projects board for simplicity of boards. I'd say it has the spirit of a project

Ghostvv:
Actually, I'd move all `Small Projects` to research projects board

tabergma:
According to my test runs in https://github.com/RasaHQ/rasa/issues/7120, we should be at least able to load the Advising and the MultiWOZ training dataset into memory and perform tokenization. As the Ubuntu corpus is not yet converted we cannot make any statement on this kind of dataset.

Things we would need to do:
- [ ] Create a new interface for the NLU components, e.g. `prepare_training`, `train_chunk`, `persist_chunk`, that is needed to handle the partial dataset training. The interface can be left empty for now.
- [ ] Creating NLU components should take the domain as an optional parameter.
- [ ] We should make sure to not load the training data multiple times before actually reading and processing it. This takes up a lot of time if the dataset is very large.
[...]

PR #8306, opened 2021-03-25
Partial dataset loading for NLU

Ghostvv:
**Proposed changes**:
- described here: https://github.com/RasaHQ/rasa/issues/6836


twerkmeister:
Hey @Ghostvv good stuff and wow this is one big boi of a PR. I know you linked the issue where initial discussions around this issue took place. However there are many lists of checkboxes that seem to not have been used consistently, and it's difficult to know what is really relevant here.

If you could provide a high level overview / conceptual description of the things you actually ended up implementing, it would really help determining where to start with reviewing this big PR

Ghostvv:
@twerkmeister did you take a look at the linked doc: https://docs.google.com/document/d/1dwKfdju5ZqhOQtURLF3-eU3PdxvdlyXafPHE_0sLPzI/edit?usp=sharing?

rasa/cli/interactive.py
```diff
@@ -93,6 +93,7 @@ def interactive(args: argparse.Namespace) -> None:
 def _set_not_required_args(args: argparse.Namespace) -> None:
     args.fixed_model_name = None
     args.store_uncompressed = False
+    args.number_of_chunks = 1
```
twerkmeister, line 96: looks like this will overwrite whatever the user has provided for interactive training. Not obvious to me that this is desirable

rasa/nlu/model.py
```diff
@@ -211,6 +226,138 @@ def train(self, data: TrainingData, **kwargs: Any) -> "Interpreter":
...
+            train_path: The training path to use.
+            fixed_model_name: The model name to use.
+            number_of_chunks: The number of chunks to use.
+
+        Returns:
+            The trained interpreter and the path to the persisted model.
+        """
+        import tempfile
+        from rasa.nlu.utils.hugging_face.hf_transformers import HFTransformersNLP
+        from rasa.nlu.utils.mitie_utils import MitieNLP
```
twerkmeister, line 249: Does this mean any pipeline would need Mitie dependencies?

Ghostvv: what do you mean?

rasa/cli/interactive.py
Ghostvv: @dakshvar22 do you know why it is done?

dakshvar22: Not sure, looks like something Tanja had added when implementing the CLI support.

rasa/nlu/model.py
twerkmeister: I was worried that this would require the mitie dependency even if you don't use mitie in your pipeline. But seems it doesn't. False Alarm :)

twerkmeister:
@Ghostvv really appreciate your efforts here. It looks like it's been a lot of work and it's been challenging. Yet, that should also make you realize that reviewing the PR is also challenging.

> @twerkmeister did you take a look at the linked doc: https://docs.google.com/document/d/1dwKfdju5ZqhOQtURLF3-eU3PdxvdlyXafPHE_0sLPzI/edit?usp=sharing?

I had a look now and I learned something about the ideas that were floating around. However, it's an 'ideas' document and it's not obvious what parts of the things you originally discussed you ended up implementing. I wish you would provide some more context in form of a summary of the current status. Otherwise it's difficult to think things through and judge whether we are moving in a good direction, not having high level context to guide one's attention as a starting point.

For example, I wonder:
* This pull request seems to be only part of the full solution. What parts and key ideas are included in this PR and which have been left out? And why?
* What is the pathway forward to the full solution?
* What is the rational for merging this into main at this stage?
* How come you decided for managing two separate data set representations and training styles? `TrainingDataFull` could be just a special case of chunked datasets, for number of chunks = 1.
* Why setting `number of chunks` instead of the somewhat more dataset-independent concept of `examples per chunk`?
* What will happen with the components that raise the `RasaTrainChunkException` with message `"This method should neither be called nor implemented in our code."` There are 5 of these (intent classifiers and entity extractors). Does this mean that we don't intend to support the chunk training methods for these? 
* A core part of this PR seems to be the caching of featurized data to hard disc to save memory. Potentially this a special case in the the general Rasa 3.0 pipeline / computation graph discussions that @wochinge, @joejuzl are having - has there been any exchange on that front?

More detailed (secondary in terms of importance) code Questions / notes:
* Is the randomization procedure of `RasaDataChunkFileGenerator` random enough? Looks to me that if you have 50k examples and split into 4 chunks, you'd still have buckets of 12.5k items each that will be generated in the same order in every epoch. Either we need to give some hints to setting a good chunk size or we could shuffle big chunks internally again after loading. 
* There needs to be more documentation / better error messages for when components not support chunked training. Imagine someone having memory issues with some pipeline, discovering this option, switching to it and then just getting a `RasaTrainChunkException('This method should neither be called nor implemented in our code')`.
* naming of `_max_batch_number_per_chunk` related things in the `RasaDataChunkFileGenerator` are somewhat confusing. It's not a list of number of batches per chunk but a cumulative sum of batches across the chunks

Ghostvv:
> This pull request seems to be only part of the full solution. What parts and key ideas are included in this PR and which have been left out? And why?

This is the full solution for nlu. The next part will be core. We decided to split them to make the sizes of the PRs smaller

> What is the pathway forward to the full solution?

it is described in last 2 comments in the linked issue: https://github.com/RasaHQ/rasa/issues/6836

> What is the rational for merging this into main at this stage?

this PR incorporates partial dataset loading for nlu, since core is build upon NLU, it makes these changes independent

> How come you decided for managing two separate data set representations and training styles? `TrainingDataFull` could be just a special case of chunked datasets, for number of chunks = 1.

I don't think, it can. `TrainingDataFull` contain more data and helpers like responses, lookup tables etc

> Why setting `number of chunks` instead of the somewhat more dataset-independent concept of `examples per chunk`?

I don't know, it was just a decision to make. We can change it to `examples per chunk` if you think it is better, but I expect it'll be hard for users to estimate number of examples

> What will happen with the components that raise the `RasaTrainChunkException` with message `"This method should neither be called nor implemented in our code."` There are 5 of these (intent classifiers and entity extractors). Does this mean that we don't intend to support the chunk training methods for these?

they shouldn't be present in the config if users want/need to train in chunks

> A core part of this PR seems to be the caching of featurized data to hard disc to save memory. Potentially this a special case in the the general Rasa 3.0 pipeline / computation graph discussions that @wochinge, @joejuzl are having - has there been any exchange on that front?

it started before Rasa 3.0 pipeline / computation graph discussions. But yes it is linked. I'd say the core part is make tf model load data from files on the fly

> Is the randomization procedure of `RasaDataChunkFileGenerator` random enough? Looks to me that if you have 50k examples and split into 4 chunks, you'd still have buckets of 12.5k items each that will be generated in the same order in every epoch. Either we need to give some hints to setting a good chunk size or we could shuffle big chunks internally again after loading.

yes, but shuffling data between chunks would require additional persist of chunks from inside tf training. I suggest to pursue it separately in follow up PR, if you think it is necessary

> There needs to be more documentation / better error messages for when components do not support chunked training. Imagine someone having memory issues with some pipeline, discovering this option, switching to it and then just getting a `RasaTrainChunkException('This method should neither be called nor implemented in our code')`.

could you please make a suggestion in code review?

> naming of `_max_batch_number_per_chunk` related things in the `RasaDataChunkFileGenerator` are somewhat confusing. It's not a list of number of batches per chunk but a cumulative sum of batches across the chunks

could you suggest a better name?

Ghostvv:
> The new training procedure is a specialized solution suited to selected components. Other components that might also suffer from large memory requirements have been excluded from this training procedure for unspecified reasons.

the reason is that it is not obvious how some ML algorithms like `CRFEntityExtractor` can support training on the files

> From a user's point of view, it isn't obvious what components support the new training procedure and which don't.

that is the documentation issue, that is not merged yet

> Maintenance of two training procedures increases complexity of the components.

yes it is the nature of this ML task

> Each training procedure has slightly different preparatory steps that are done with the data

yes, it happens because majority of our components are drastically different from each other

> For example, surprisingly, the original training procedure of the `DietClassifier`, that operates on the full dataset in memory, now calls `prepare_partial_training`, although this naming points directly to the opposite training procedure.

that's not the case. `prepare_partial_training` in `DietClassifier` doesn't contain the training.

> To me, this pull request has the quality of an experiment/prototype.

that's not for me to decide

> I am not convinced that a more elegant and generic design does not exist

it exists, but I don't think it is possible to achieve within current rasa oss architecture

> Why do components need to know whether their training data is stored in memory or on disc in separate files?

because the training loop significantly differs in these cases

> Whether that stream comes from memory, from hard disc, or from the cloud shouldn't matter.

it does matter, because it changes how batching is implemented and which classes are used to perform batching

> If a component needs to collect some dataset-wide metadata first, it can consume the stream twice

that is what we're doing

> Such a solution wouldn't suffer any of the drawbacks of the current proposal mentioned before.

I don't see the difference in your proposal, except that I don't think you have taken into account changes needed to internal training loop of ML components

Ghostvv:
@twerkmeister I agree that this PR has some qualities of being a prototype. The problem is that rasa architecture doesn't suit this feature. Maybe, it makes sense not to merge it, but use it as a template for 3.0 major refactor? cc @wochinge @joejuzl

wochinge:
>  Maybe, it makes sense not to merge it, but use it as a template for 3.0 major refactor? cc @wochinge @joejuzl

@joejuzl and I looked into how partial training could fit into the new architecture yesterday. This PR was already very insightful for us, e.g. the fact that `DIET` needs access to all chunks to shuffle them after each epoch. We've also figured that we can most probably do without the `prepare` step. Our plan yesterday was very close to the current implementation (separate `train_on_chunks` method on `Component`s. Note that we rather focused upon general feasability of the new architecture - definitely worth giving it another thought when creating the actual plan. 

Thanks for your comments, Vova and Thomas! They are super high quality, concise and helpful!

Ghostvv:
> most probably do without the prepare step

how would you achieve that? do you plan move model and label data creation out of diet?

twerkmeister:
> the reason is that it is not obvious how some ML algorithms like `CRFEntityExtractor` can support training on the files

It's true, that `CRFEntityExtractor` will build some representation of the entire dataset internally. Yet, I think we can save roughly 50% memory if we didn't hold the entire featurized dataset in memory on top of what the crf extractor is doing. When I was fixing the memory leak in that component recently, I was using a dataset with about 60k sentences, which consumed about 4GB by itself when featurized with 300 dimensional vectors from spacy. When the CRF training started, the crf - internal representation took another 4 GB or so. Makes 8GB in total. If we have a way where we don't keep the whole data set in memory we could save around 4GB of that. Obviously not perfect, and there will be limits at some point, but 50% saving isn't bad.

> yes it is the nature of this ML task

I think we were talking past each other here. Yes, ML is complicated, but what I am saying is that we shouldn't just add additional complexity on top through our engineering/design choices.

> that's not the case. prepare_partial_training in DietClassifier doesn't contain the training.

I was saying that the "training on the full dataset" calls prepare partial training, which is confusing. I wanted to underline that managing two separate training procedures will become very tiresome very fast.

https://github.com/RasaHQ/rasa/blob/9e731d54e883fee5521a677b4aef4ec60f2af6f0/rasa/nlu/classifiers/diet_classifier.py#L895-L907

> because the training loop significantly differs in these cases

> it does matter, because it changes how batching is implemented and which classes are used to perform batching

I don't think this is the case. From my understanding, what most components are doing is iterating over a stream of messages, preprocessing them, and training on them. It doesn't really matter what happens behind this interface of a list or stream of messages. They could be kept in memory, they could be kept on disc, they could be downloaded from the cloud, they could just be generated on the fly. A component that just consumes a stream of messages, doesn't need to care. If a specific batching procedure is needed by a component, that component can take care of that. If it wants batches of 32, it can just take 32 Messages out of the stream and do its thing. Shuffling and the like would have to happen where the stream is managed.

> I don't see the difference in your proposal, except that I don't think you have taken into account changes needed to internal training loop of ML components

I don't really see why you couldn't just keep the normal `TrainingData` class and instead of it managing the examples as a list in memory it manages some data on the hard drive, maybe with some shuffling when needed etc. From the outside you are just asking for the nlu_examples as before, and you wouldn't notice that anything has changed from the outside. I think this way is much more generic, works for all components, results in a single training procedure per component, and probably doesn't require as much code changes.

What are some of the things that I am overlooking here with respect to the training loop?

Ghostvv:
> I think we can save roughly 50% memory `CRFEntityExtractor`

that's a completely different problem that we don't try to address in this PR

> I was saying that the "training on the full dataset" calls prepare partial training, which is confusing.

then I'd say it's just the matter of naming and comments. Logic-wise, it's very clear what's going on there

Ghostvv:
> From my understanding, what most components are doing is iterating over a stream of messages, preprocessing them, and training on them.

I'm sorry, but that's not accurate, there are a couple of methods that require to see all the messages

Ghostvv:
> It doesn't really matter what happens behind this interface of a list or stream of messages.

ye if you persist per message, but then you'd have awful lot of read operations

Ghostvv:
> I don't really see why you couldn't just keep the normal TrainingData class and instead of it managing the examples as a list in memory it manages some data on the hard drive, maybe with some shuffling when needed etc. 

we want to avoid unnecessary read/write operations, therefore we cannot persist per message

twerkmeister:
> that's a completely different problem that we don't try to address in this PR

I am saying that with a different design you would get this for free. And it is what you are adressing: only loading the dataset partially at any given time.  

> then I'd say it's just the matter of naming and comments. Logic-wise, it's very clear what's going on there

What I am saying is that with managing two very similar methods with a lot of code duplication but subtle differences, it will always be very prone to error.

> I'm sorry, but that's not accurate, there are a couple of methods that require to see all the messages

That's what we talked about earlier. If you need to see all messages to collect dataset-wide metadata you run over the stream twice, no?

> ye if you persist per message, but then you'd have awful lot of read operations
obviously I am not proposing to persist every single message one by one, but in batches/chunks just like you did in principle

Ghostvv:
> But I'd be doing it inside the the TrainingDataset which makes that whole procedure invisible to whoever consumes the messages

but implementation of which tf data generator to use depends on whether you load from files or keep data in the memory

wochinge:
> how would you achieve that? do you plan to move model and label data creation out of diet classifier?

With the graph we can pass in the original training data set (without features) in addition to the chunks

PR #8328, opened 2021-03-29
Bug in is_entity_extractor_present() in rasa/nlu/test.py

### LINKED ISSUE #8325: Error in is_entity_extractor_present() in rasa/nlu/test.py
**Rasa version**: Rasa 1.10.23, but code also seems unchanged in master 

**Python version**: 3.7.9

**Operating system** (windows, osx, ...): Linux

**Issue**:
_is_entity_extractor_present_() checks if the return of _get_entity_extractors_() is not an empty list.
But _get_entity_extractors_() returns a set, so the the comparison is always True.

```
rasa test nlu 
```

koernerfelicia:
closes #8325

Problem: 
we compare set to empty list to check whether set is empty.

Solution:
Check for set len.

- ~updated the documentation~
- ~updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)~

### DIFF
--- a/data/test_config/stack_config.yml
+++ b/data/test_config/stack_config.yml
@@ -3,6 +3,7 @@
 language: en
 pipeline:
   - name: KeywordIntentClassifier
+  - name: RegexEntityExtractor
 
 # Configuration for Rasa Core.
 # https://rasa.com/docs/rasa/policies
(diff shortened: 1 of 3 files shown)


data/test_config/stack_config.yml
```diff
@@ -3,6 +3,7 @@
 language: en
 pipeline:
   - name: KeywordIntentClassifier
+  - name: RegexEntityExtractor
```
federicotdn, line 6: Why do we need to modify this file?

federicotdn (approved):
Code looks good! Only have one question, see below.

data/test_config/stack_config.yml
federicotdn: Makes sense! Thanks :+1:

PR #8334, opened 2021-03-29
Create a `skip_validation` parameter to yaml story reader

### LINKED ISSUE #8333: Introduce a `skip_validation` parameter in yaml reader
**Description of Problem**:
Currently `yaml_story_reader::read_from_string` always validates it's input.
This makes reading strings from the database (that were already validated) very slow, which e.g. happens in requests to `GET /domainWarnings` in rasa-x.

**Overview of the Solution**:
To speed up reading strings from the database we introduce a `skip_validation` parameter in `yaml_story_reader::read_from_string` and `yaml_story_reader::read_from_file`

**Examples** (if relevant):

**Blockers** (if relevant):

**Definition of Done**:

kalkbrennerei:
closes #8333

**Proposed changes**:
- add a `skip_validation` flag to `yaml_story_reader::read_from_file` and `yaml_story_reader::read_from_string`


rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -53,7 +53,9 @@ def __init__(
                 docs=DOCS_URL_MIGRATION_GUIDE_MD_DEPRECATION,
             )
 
-    def read_from_file(self, filename: Union[Text, Path]) -> List[StoryStep]:
+    def read_from_file(
+        self, filename: Union[Text, Path], skip_validation: bool = False
```
kalkbrennerei, line 57: I added the `skip_validation` parameter here because I added it to the base class. I'm not using it here in `markdown_story_reader` so it might be a bit misleading. Happy to hear ideas on how to solve this!! 🤯

rasa/shared/core/training_data/story_reader/story_reader.py
```diff
@@ -55,7 +55,19 @@ def __init__(
...
 
-    def read_from_file(self, filename: Text) -> List[StoryStep]:
+    def read_from_file(
+        self, filename: Text, skip_validation: bool = False
+    ) -> List[StoryStep]:
+        """Reads stories or rules from file.
+
+        Args:
+            filename: Path to the story/rule file.
+            skip_validation: `True` if the file was already validated
```
degiz, line 65: I'd change the description and just mention that this flag will skip the validation step. I don't think that users of this API will benefit from mentioning the database in this context 🤔

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
degiz: cc @chdorner in case that's not aligned with the changes that you've planned 😬

changelog/8333.bugfix.md
```diff
@@ -0,0 +1,3 @@
+Introduce a `skip_validation` flag in order to speed up reading yaml files that were already validated.
+
+Previously all files were validated, also the one that were already stored in the database, which slowed down requests to `GET /domainWarnings` in rasa-x.
```
degiz, line 3: that's an OS change, I don't think we need to mention the Rasa X endpoints

degiz (changes requested):
Thanks 👍

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
chdorner: Thanks for the heads-up! We haven't been on such a detailed level yet, so all good! Looks like a worthwhile flag to add anyways for when we're sure that it's valid.

degiz (changes requested):
Looks good! Can we add a unit tests though? 🤔

changelog/8333.bugfix.md
```diff
@@ -0,0 +1 @@
+Introduce a `skip_validation` flag in order to speed up reading yaml files that were already validated.
```
federicotdn, line 1:
```suggestion
Introduce a `skip_validation` flag in order to speed up reading YAML files that were already validated.
```

federicotdn (approved):
Looks good! Do we also need to update the Markdown implementation of the reader? We should add the parameter there as well, even if we do not use it.

kalkbrennerei:
> Looks good! Do we also need to update the Markdown implementation of the reader? We should add the parameter there as well, even if we do not use it.

@federicotdn Since we'll drop markdown support, can we actually leave it out? Also I think that validation works differently in the markdown reader, so it might be tricky to add the `skip_validation` there afaik

federicotdn:
I actually was thinking of only adding the parameter (either named `_` or `_skip_validation`), just so the method's signature is identical to it's parent class' (without any implementation).

kalkbrennerei:
ohhh sorry, I didn't read properly 🙈
I added the `skip_validation` parameter to the `read_from_file` method of the markdown reader without using it, so I think what you're asking is already there? Not sure if we're talking about the same thing though 😅

federicotdn:
Aaah I missed it! Yes exactly, that was what I was referring to 👍🏻 👍🏻

degiz (commented):
let's goo 🚀

degiz (approved):

Issue #8124, opened 2021-03-05
Remove unnecessary model trainings from tests


wochinge:
**Description of Problem**:

We found out that[ a lot of unit tests use model fixtures](https://www.notion.so/rasa/Proposal-for-tests-which-train-models-a091d698ef794b1fb8f0f89c752f7907?d=fb0dc5b6-482c-4e94-9c97-199bc46bc429#963b3348c938429b9a204b012deae40b) which require training although they don't require a trained model. In combination with the `@pytest.mark.trains_model` annotation this causes unnecessary noise and makes the tests slower to run locally.

**Overview of the Solution**:
* introduce new fixtures which don't require a trained model

**Definition of Done**:
- [ ] All tests in [this document](https://www.notion.so/rasa/Proposal-for-tests-which-train-models-a091d698ef794b1fb8f0f89c752f7907?d=fb0dc5b6-482c-4e94-9c97-199bc46bc429#963b3348c938429b9a204b012deae40b) which are in the column `no training needed` are no longer using a trained model and hence no longer have the `@pytest.mark.trains` annotation

PR #8346, opened 2021-03-31
Amend fixtures in tests that don't need model training

ancalita:
**Proposed changes**:
- Complete #8124 

### DIFF
--- a/changelog/8124.misc.md
+++ b/changelog/8124.misc.md
@@ -1,0 +1,1 @@
+Amend fixtures in tests that do not require model training.
--- a/changelog/8124.misc.md
+++ b/changelog/8124.misc.md
@@ -1,1 +1,0 @@
-Amend fixtures in tests that do not require model training.
--- a/tests/cli/test_rasa_test.py
+++ b/tests/cli/test_rasa_test.py
@@ -1,4 +1,6 @@
 import os
+import shutil
+from pathlib import Path
 from shutil import copyfile
 
 import pytest
@@ -6,7 +8,7 @@ import pytest
 from rasa.core.test import CONFUSION_MATRIX_STORIES_FILE
 from rasa.constants import RESULTS_FILE
 from rasa.shared.constants import DEFAULT_RESULTS_PATH
-from rasa.shared.utils.io import list_files, write_yaml
+from rasa.shared.utils.io import list_files, write_yaml, write_text_file
 from typing import Callable
 from _pytest.pytester import RunResult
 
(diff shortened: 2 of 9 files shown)


tests/test_model.py
```diff
@@ -55,13 +55,17 @@
 from tests.core.conftest import DEFAULT_DOMAIN_PATH_WITH_MAPPING, DEFAULT_STACK_CONFIG
 
 
-def test_get_latest_model(trained_rasa_model: str):
-    path_of_latest = os.path.join(os.path.dirname(trained_rasa_model), "latest.tar.gz")
-    shutil.copy(trained_rasa_model, path_of_latest)
+def test_get_latest_model(tmpdir: Path):
```
wochinge, line 58: `tmpdir` is not of type `Path` but of type `LocalPath` which is slightly different. I suggest to use always `tmp_path`
```suggestion
def test_get_latest_model(tmp_path: Path):
```

changelog/8124.misc.md
```diff
@@ -0,0 +1 @@
+Amend fixtures in tests that do not require model training.
```
wochinge, line 1: in my opinion we don't need a changelog as this isn't user facing

wochinge (approved):
Nicely done! 💯

wochinge:
Re `test_test_core_comparison_after_train`: Would it be possible to use a pre-trained model (e.g. from a fixture) and copy it to `comparison_models/run_1` and so on? It would still train but would use the fixture

changelog/8124.misc.md
alwx: It's a misc entry, so it shouldn't be included to the backlog anyway :)

wochinge: it actually does https://github.com/RasaHQ/rasa/blob/f52c12a069923ac7cd06fb65bb5b1d8f670fc2e1/CHANGELOG.mdx#L122-L121 . But you're right: doens't really matter 👍🏻

alwx: Interesting that the `changelog/README.md` states that it won't be while in fact it's getting included :D need to fix that

tests/test_model.py
ancalita: That's been amended & also the `test_test_core_comparison_after_train` was changed too 🥳  Let me know your thoughts.

tests/cli/test_rasa_test.py
```diff
@@ -148,42 +150,32 @@ def test_test_core_comparison(
...
-    write_yaml(
-        {"language": "en", "policies": [{"name": "MemoizationPolicy"}]}, "config_1.yml"
-    )
 
-    write_yaml(
-        {"language": "en", "policies": [{"name": "MemoizationPolicy"}]}, "config_2.yml"
-    )
+    path = Path(tmp_path / "comparison_models")
+    path.mkdir()
+    assert path.exists()
```
wochinge, line 160: I'd remove this line as this only checks if the Python code `path.mkdir` works as expected
```suggestion
```

tests/cli/test_rasa_test.py
```diff
@@ -148,42 +150,32 @@ def test_test_core_comparison(
...
-    )
+    run_one = Path(path / "run_1")
+    run_one.mkdir()
+    assert run_one.exists()
+    shutil.copy(trained_rasa_model, run_one)
+
+    run_two = Path(path / "run_2")
+    run_two.mkdir()
+    assert run_two.exists()
+    shutil.copy(trained_rasa_model, run_two)
```
wochinge, line 170: optional: some tiny nit picks
```suggestion
    run_one = path / "run_1"
    run_one.mkdir()
    shutil.copy(trained_rasa_model, run_one)

    run_two = path / "run_2"
    run_two.mkdir()
    shutil.copy(trained_rasa_model, run_two)
```

wochinge (approved):
💯

tests/cli/test_rasa_test.py
ancalita: all done now 😃

PR #8349, opened 2021-03-31
Count only nlu training examples

### LINKED ISSUE #7379: Make error message about too few training examples more informative
**Description of Problem**:

The message about too few **core** training examples doesn't inform of what kind of training example is needed (i.e. more intent phrasings or more stories containing the intent). It also doesn't help that the message appears right after `Training NLU model...` giving the impression that this is about nlu data.

```
UserWarning: Intent 'out_of_scope' has only 1 training examples! Minimum is 2, training may fail.
```
**Overview of the Solution**:

One way to make this better would be to explicit what kind of training example is meant:
```
UserWarning: Intent '{intent}' is present in only one story! Minimum is 2, training may fail.
```

Alternatively, raise the warning only after `Training Core model...` which would also help put the warning in context.

ancalita:
**Proposed changes**:
- Fix #7379 

### DIFF
--- a/tests/shared/nlu/training_data/test_training_data.py
+++ b/tests/shared/nlu/training_data/test_training_data.py
@@ -8,7 +8,6 @@ import pytest
 
 import rasa.shared.utils.io
 from rasa.shared.core.constants import USER_INTENT_OUT_OF_SCOPE
-from rasa.shared.importers.utils import training_data_from_paths
 from rasa.shared.nlu.constants import (
     TEXT,
     INTENT_RESPONSE_KEY,
@@ -23,7 +22,6 @@ from rasa.shared.nlu.constants import (
 from rasa.nlu.convert import convert_training_data
 from rasa.nlu.extractors.mitie_entity_extractor import MitieEntityExtractor
 from rasa.nlu.tokenizers.whitespace_tokenizer import WhitespaceTokenizer
-from rasa.shared.nlu.training_data.formats import RasaYAMLReader
 from rasa.shared.nlu.training_data.message import Message
 from rasa.shared.nlu.training_data.training_data import TrainingData
 from rasa.shared.nlu.training_data.loading import guess_format, UNK, load_data
--- a/tests/shared/nlu/training_data/test_training_data.py
+++ b/tests/shared/nlu/training_data/test_training_data.py
@@ -8,6 +8,7 @@ import pytest
 
 import rasa.shared.utils.io
 from rasa.shared.core.constants import USER_INTENT_OUT_OF_SCOPE
+from rasa.shared.importers.utils import training_data_from_paths
 from rasa.shared.nlu.constants import (
     TEXT,
     INTENT_RESPONSE_KEY,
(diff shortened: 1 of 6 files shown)


tests/shared/nlu/training_data/test_training_data.py
```diff
@@ -295,6 +295,34 @@ def test_train_test_split(filepaths: List[Text]):
         )
 
 
+def test_number_of_examples_per_intent():
```
joejuzl, line 298: would this test have failed before your change?

joejuzl (changes requested):

tests/shared/nlu/training_data/test_training_data.py
ancalita: yup, exactly! this test fails before the amendment to `nlu_examples`

joejuzl: because `message_intent_two` would be counted? Maybe a comment or different names would help to show which should and shouldn't be counted? 

Also would it be possible to test this on some actual yaml training data files? e.g. from in the `data/` directory? Would be good to show how it fits into the bigger picture.

ancalita: Ok, yes, have amended names to reflect which is `nlu `and which `non_nlu `- haven't yet pushed changes.
I couldn't find an example in the data/ directory that has matching intents in `nlu.yml `and `stories.yml `or `rules.yml `files? Should I create my own subdirectory with some yml files - let's say test_number_of_nlu_examples?

ancalita: all done now 😄

tests/shared/nlu/training_data/test_training_data.py
```diff
@@ -295,6 +297,87 @@ def test_train_test_split(filepaths: List[Text]):
...
+    non_nlu_td = TrainingData(training_examples=non_nlu_data)
+    assert non_nlu_td.number_of_examples_per_intent.get("greet") is None
+    assert non_nlu_td.intents == {"greet", "ask_weather"}
+
+    from rasa.shared.importers.utils import training_data_from_paths
+
+    td = training_data_from_paths(
+        ["data/test_number_nlu_examples/nlu.yml"], language="en"
+    )
+    td.merge(non_nlu_td)
```
joejuzl, line 353: A easier and more realistic way to get a combined training data from the nlu, stories and rules could instead be something like:
```
    importer = TrainingDataImporter.load_from_dict(
        {}, str(config_path), str(domain_path), ["data/test_number_nlu_examples/nlu.yml", "data/test_number_nlu_examples/stories.yml", "data/test_number_nlu_examples/rules.yml"]
    )
    training_data = await importer.get_nlu_data()
```

(See `test_without_additional_e2e_examples` for how to create empty domain config files.)

joejuzl (changes requested):

tests/shared/nlu/training_data/test_training_data.py
ancalita: Neat! Actioned 😃

tests/shared/nlu/training_data/test_training_data.py
```diff
@@ -8,6 +8,7 @@
 
 import rasa.shared.utils.io
 from rasa.shared.core.constants import USER_INTENT_OUT_OF_SCOPE
+from rasa.shared.importers.utils import training_data_from_paths
```
joejuzl, line 11: this can go now I think.

joejuzl (approved):
Nice succinct change, and good tests! 
(just a couple of unused imports to remove)

Issue #8316, opened 2021-03-26
implement train on chunk for policies


Ghostvv:
part of https://github.com/RasaHQ/rasa/issues/6836

similar as in diet, we need to implement `train_on_chunk` for policies

Note: This feature should branch off from `partial_dataset_loading-core` and should be squashed and merged back into that branch.

Issue #8315, opened 2021-03-26
implement train chunk for core featurizers


Ghostvv:
part of https://github.com/RasaHQ/rasa/issues/6836

core featurizers should create smth like `TrainingDataChunk` files in nlu

PR #8353, opened 2021-04-01
make ted train on chunks

Ghostvv:
**Proposed changes**:
- closes https://github.com/RasaHQ/rasa/issues/8315
- closes https://github.com/RasaHQ/rasa/issues/8316 

### DIFF
--- a/rasa/core/featurizers/tracker_featurizers.py
+++ b/rasa/core/featurizers/tracker_featurizers.py
@@ -299,6 +299,16 @@ class TrackerFeaturizer:
         trackers_as_states = np.array(trackers_as_states)
         trackers_as_actions = np.array(trackers_as_actions)
         trackers_as_entities = np.array(trackers_as_entities)
+        # we need to make sure that each chunk contains entities
+        example_with_entities = []
+        for i, turns_as_entities in enumerate(trackers_as_entities):
+            if any(turn_entities for turn_entities in turns_as_entities):
+                example_with_entities = [
+                    trackers_as_states[i],
+                    trackers_as_actions[i],
+                    trackers_as_entities[i],
+                ]
+                break
 
         data = []
         for label, num_label_examples in Counter(labels).items():
@@ -346,6 +356,16 @@ class TrackerFeaturizer:
                     trackers_as_entities_for_label[start:end]
                 )
 
+            # we need to make sure that each chunk contains entities
+            entities_present = any(
+                any(turn_entities for turn_entities in turns_as_entities)
+                for turns_as_entities in trackers_as_entities_chunk
+            )
+            if not entities_present and example_with_entities:
+                trackers_as_states_chunk.append(example_with_entities[0])
+                trackers_as_actions_chunk.append(example_with_entities[1])
+                trackers_as_entities_chunk.append(example_with_entities[2])
+
             tracker_state_features_chunk = self._featurize_states(
                 trackers_as_states_chunk, interpreter
             )
@@ -383,8 +403,12 @@ class TrackerFeaturizer:
         tf_features[LABEL] = chunk_utils.int_feature(turn_label_ids)
 
         for i, entity_tag in enumerate(turn_entity_tags):
-            for features in entity_tag.values():
-                tf_features.update(chunk_utils.encode_features(features, f"{i}"))
+            for attribute, features in entity_tag.items():
+                tf_features.update(
+                    chunk_utils.encode_features(
+                        features, f"{i}{chunk_utils.TF_RECORD_KEY_SEPARATOR}{attribute}"
+                    )
+                )
 
         return tf_features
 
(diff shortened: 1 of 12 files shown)


koernerfelicia:
@Ghostvv heads up I'm out starting tomorrow until next Friday

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -240,6 +247,299 @@ def featurize_trackers(
...
+            for turn_state_features, turn_label_ids, turn_entity_tags in zip(
+                tracker_state_features_chunk, label_ids_chunk, entity_tags_chunk
+            ):
+                example = tf.train.Example(
+                    features=tf.train.Features(
+                        feature=self._to_tf_features(
+                            turn_state_features, turn_label_ids, turn_entity_tags
+                        )
+                    )
+                )
```
dakshvar22, line 445: Hmm, we should check if we can merge this function and `persist_chunk` inside rasa.shared.nlu.training_data.training_data.py` and create a generic TFRecord writer function that can be reused.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -240,6 +247,299 @@ def featurize_trackers(
...
+                # Append each example into tfrecord
+                tfwriter.write(example.SerializeToString())
+
+        return file_path
+
+    def load_chunk(
```
dakshvar22, line 451: Could you add some comments for future purposes in this function?

rasa/model_training.py
```diff
@@ -590,7 +590,9 @@ async def _train_core_with_validated_data(
                 await rasa.core.train.train_in_chunks(
                     domain_file=domain,
                     training_resource=file_importer,
-                    output_path=Path(_train_path, DEFAULT_CORE_SUBDIRECTORY_NAME),
+                    output_path=os.path.join(
```
dakshvar22, line 593: Can you leave it to `Path` to be consistent with the code? Any particular reason why it should change?

rasa/shared/utils/chunk_utils.py
```diff
@@ -0,0 +1,152 @@
+import logging
```
dakshvar22, line 1: We can add the generic persist and load methods for chunk here.

rasa/model_training.py
```diff
@@ -590,7 +590,9 @@ async def _train_core_with_validated_data(
                 await rasa.core.train.train_in_chunks(
```
dakshvar22, line 590: Do we want to support finetuning?
It's okay to leave it out in case this is going to be a prototype, but i think you will just need to add a `model_to_finetune` parameter to this method and set the agent's ensemble to `model_to_finetune`'s policy ensemble and the rest should be fine.

dakshvar22 (commented):
The logic seems largely fine to me at first glance.
I take that the code quality isn't expected to be merge ready and just a prototype?

rasa/model_training.py
Ghostvv: this is the change from `main`

Ghostvv:
> I take that the code quality isn't expected to be merge ready and just a prototype?

yes

rasa/model_training.py
Ghostvv: my thinking was: if you do fine-tuning, you don't fine-tune on a lot of data, so normal train should be good

koernerfelicia:
NOTE: there is a known bug (maybe two) with this implementation (see discussion [here](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1619533190265700?thread_ts=1618303780.462500&cid=C8LEHMTAM)).

The first, known bug occurs with datasets that have a long tail distribution of action labels. For example, end-to-end datasets with lots of unique bot utterances (as the label for these is `action_text`).

This means that the chunks are not (even roughly) equal in size. The first chunk will contain examples for each label, and subsequent chunks will be smaller, as there are no more examples for this label. This occurs [here](https://github.com/RasaHQ/rasa/blob/97c8d8d28552cdbab452e4ff61caac216f32a213/rasa/core/featurizers/tracker_featurizers.py#L351-L357).

The second, potential bug, which is possibly related is that the subsequent smaller chunks may not have entities. We hypothesized that this may be because (in this scenario with the long tail of unique bot utterances) the large chunk contains all the unique utterances and the subsequent chunks contain examples only for `action_listen`. It's not clear if this is related, since Vova did provision for missing entities [here](https://github.com/RasaHQ/rasa/blob/97c8d8d28552cdbab452e4ff61caac216f32a213/rasa/core/featurizers/tracker_featurizers.py#L359-L363).

The bug can be reproduced with MultiWOZ for 400 dialogues (312 train, these dialogues are produced with the script [here](https://github.com/RasaHQ/training-data/blob/main/public/MultiWOZ/scripts/dataset_to_rasa.py) with number of dialogues flag set to 400) and chunk=5.

Issue #8025, opened 2021-02-23
Load label data in batches to avoid OOM for many `actions_or_text` in e2e


koernerfelicia:
related to #6836

With e2e training and testing, the number of items in `actions_or_text` has grown considerably. This leads to memory issues due to large tensors.

To prevent this we need to:
- [ ] Load label data in batches for **train**, using some heuristic to create the batches.
- [ ] Load label data in batches for **test**

dakshvar22:
The code for this exists on `batch_labels` branch and also merged with the larger branch of training core in chunks (`policy-train-in-chunk-experiments`). This issue should be closed for now as it is done for the prototype used to run experiments.

PR #8357, opened 2021-04-01
Batch labels for training and inference through TED to reduce memory consumption

dakshvar22:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/8025

### DIFF
--- a/rasa/core/policies/ted_policy.py
+++ b/rasa/core/policies/ted_policy.py
@@ -399,21 +399,41 @@ class TEDPolicy(Policy):
         label_data.add_features(
             LABEL_KEY,
             LABEL_SUB_KEY,
-            [FeatureArray(np.expand_dims(label_ids, -1), number_of_dimensions=2),],
+            [FeatureArray(np.expand_dims(label_ids, -1), number_of_dimensions=2)],
         )
 
-        # If any action texts are present, explicitly add the first label id
-        # belonging to action text. This is necessary to distinguish between
-        # action name label ids and action text label ids inside TF graph code.
-        num_action_texts = len(domain.action_texts)
-        action_text_ids = label_ids[domain.num_actions - num_action_texts :]
+        label_batch_size = self.config[LABEL_BATCH_SIZE]
 
-        if np.size(action_text_ids):
-            label_data.add_features(
-                LABEL_KEY,
-                SAMPLE_ACTION_TEXT_ID,
-                [FeatureArray(action_text_ids[:1], number_of_dimensions=1)],
-            )
+        # Get label ids for each key in attribute data by using the pre-computed mask.
+        for key, data in attribute_data.items():
+            if MASK in data:
+                mask = np.reshape(data[MASK][0], (-1,)).astype(dtype=np.bool)
+                key_ids = label_ids[mask].astype(np.int32)
+                label_data.add_features(
+                    f"{LABEL_KEY}_{key}",
+                    LABEL_SUB_KEY,
+                    [FeatureArray(np.expand_dims(key_ids, -1), number_of_dimensions=2)],
+                )
+                key_label_batch_size = (
+                    key_ids.shape[0]
+                    if label_batch_size == -1
+                    else max(
+                        1,
+                        int(
+                            (np.shape(key_ids)[0] / np.shape(label_ids)[0])
+                            * label_batch_size
+                        ),
+                    )
+                )
+                label_data.add_features(
+                    f"{LABEL_KEY}_{key}",
+                    LABEL_BATCH_SIZE,
+                    [
+                        FeatureArray(
+                            np.array([key_label_batch_size]), number_of_dimensions=1
+                        )
+                    ],
+                )
 
         return label_data, encoded_all_labels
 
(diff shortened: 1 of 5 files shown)


rasa/core/policies/ted_policy.py
```diff
@@ -1092,36 +1095,84 @@ def _compute_dialogue_indices(
...
+
+    def _slice_tensor(
+        self, tensor: Union[tf.Tensor, tf.SparseTensor], indices: tf.Tensor
+    ) -> Union[tf.Tensor, tf.SparseTensor]:
+        if isinstance(tensor, tf.Tensor):
+            return tf.gather_nd(tensor, tf.cast(indices, dtype=tf.int32))
+        elif isinstance(tensor, tf.SparseTensor):
+            dense_tensor = tf.sparse.to_dense(tensor)
+            sliced_tensor = tf.gather_nd(dense_tensor, tf.cast(indices, dtype=tf.int32))
+            return tf.sparse.from_dense(sliced_tensor)
```
dakshvar22, line 1136: This will blow up the memory a bit but not sure what else can be done because `tf.gather_nd` does not support sparse tensors.

Ghostvv:
could you please rebase it to `policy-train-in-chunk`, since I don't think we're going to properly finalize core chunking in the near future

rasa/core/policies/ted_policy.py
```diff
@@ -1092,36 +1095,84 @@ def _compute_dialogue_indices(
...
+    ) -> Tuple[tf.Tensor, tf.Tensor]:
+
         all_label_ids = self.tf_label_data[LABEL_KEY][LABEL_SUB_KEY][0]
-        # labels cannot have all features "fake"
-        all_labels_encoded = {}
+
+        if number_of_random_labels != -1:
+
+            # First, sample a fixed number of random labels ids
+            total_num_label_ids = all_label_ids.shape[0]
```
Ghostvv, line 1107: I think, it should be
```suggestion
            total_num_label_ids = tf.shape(all_label_ids)[0]
```
not sure if `all_label_ids` has static shape, but maybe fine if it is

rasa/core/policies/ted_policy.py
```diff
@@ -1697,10 +1748,21 @@ def batch_loss(
...
-        label_ids = tf_batch_data[LABEL_KEY][LABEL_SUB_KEY][0]
-        labels_embed = self._get_labels_embed(label_ids, all_labels_embed)
+        (
+            sampled_label_ids,
+            sampled_labels_embed,
+        ) = self._create_all_labels_embed_for_training(
+            number_of_random_labels=self.config[LABEL_BATCH_SIZE]
+        )
+
+        batch_labels_embed = self._compute_embedding_for_label_ids(
```
Ghostvv, line 1760: by doing separate computation, the number of computations are doubled compared to previous implementation

rasa/core/policies/ted_policy.py
dakshvar22: `all_label_ids` has static shape, also using `tf.shape` will return a dynamic shape which cannot be used as an argument to `candidate_sampler`.
In summary, this should work as expected.

dakshvar22: `all_label_ids` has static shape, also using `tf.shape` will return a dynamic shape which cannot be used as an argument to `candidate_sampler`.
In summary, this should work as expected.

rasa/core/policies/ted_policy.py
```diff
@@ -1179,36 +1183,179 @@ def _compute_dialogue_indices(
...
+                filtered_label_data[key] = {}
+                for sub_key in self.tf_label_data[key]:
+                    filtered_attribute_data = [
+                        self._slice_tensor(data, label_ids)
+                        for data in self.tf_label_data[key][sub_key]
+                    ]
+                    filtered_label_data[key][sub_key] = filtered_attribute_data
+
+        labels_encoded = {}
+        for key in filtered_label_data:
```
Ghostvv, line 1337:
```suggestion
        for key in filtered_label_data.keys():
```

rasa/core/policies/ted_policy.py
```diff
@@ -1179,36 +1183,179 @@ def _compute_dialogue_indices(
...
+            input_tensor.indices, sel, axis=0
+        )  # [[2,2], [2,3], [3,1]]
+        indices_new = tf.concat(
+            [
+                indices_new[:, :axis],
+                tf.expand_dims(selection_axis_indices_new, 1),
+                indices_new[:, axis + 1 :],
+            ],
+            axis=1,
+        )  # [[0,2], [0,3], [0,1]]
```
Ghostvv, line 1301:
```suggestion
        )  # [[0,2], [0,3], [1,1]]
```

Ghostvv (commented):
I didn't do the code review, in principle couple of variables need to be renamed and more small helper created, also the whole thing should be probably abstracted into a rasa keras layer. But idea-wise looks good to me for testing

rasa/utils/tensorflow/model_data.py
```diff
@@ -398,12 +398,13 @@ def number_of_examples(self, data: Optional[Data] = None) -> int:
         if not example_lengths:
             return 0
 
+        # TODO: See, why this needs to be commented
```
twerkmeister, line 401: @dakshvar22 seems this needs to be commented as the default actions have an `action_name` such as `action_listen` while the e2e bot utterances use `action_text`. If you have 1000 actions, you end up with 11 or so default actions and 989 e2e actions. And their ids get sorted into different buckets through the masks. That's why then lengths don't work out.

twerkmeister:
Closing for now, I think this is not relevant anymore

PR #8358, opened 2021-04-01
7005 rasa interactive cannot write non ascii character

joaoCeilandia:
**Proposed changes**:
- fix #7005 

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #8372, opened 2021-04-06
fix wait for docs test step in CI: 2nd try

HotThoughts:
**The problem**:
The previous fix  https://github.com/RasaHQ/rasa/pull/8371 only works when the changes in the branch are only from the owner of the forked repo. This may imply that the workflow run exists in the forked repo when all the changes are from the forked repo. If we update the PR with the base branch via the 'update branch' bottom, the workflow run will be hosted in the base repo, thus we cannot find the workflow run in the forked repo. 

Example: https://github.com/RasaHQ/rasa/pull/8358
Although the `wait for docs test` is green, it didn't find the checks.
```
No completed checks named Test Documentation, waiting for 60 seconds...
166
No completed checks after 3000 seconds, exiting with conclusion 'timed_out'
```

**Proposed changes**:
- Add back the checkout step and remove the owner settings. 
- Add a new step to fail the job if the conclusion is 'timed_out'.

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -69,7 +69,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           checkName: Test Documentation
-          ref: ${{ github.head_ref || github.sha }}
+          ref: ${{ github.event.pull_request.head.ref || github.sha }}
           owner: ${{ github.event.pull_request.head.repo.owner.login || github.repository_owner }}
           timeoutSeconds: ${{ env.WAIT_TIMEOUT_SECS }}
           intervalSeconds: ${{ env.WAIT_INTERVAL_SECS }}
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -61,7 +61,7 @@ jobs:
     name: Wait for docs tests
     runs-on: ubuntu-latest
     needs: [ changes ]
-    if: needs.changes.outputs.docs == 'true'
+    # if: needs.changes.outputs.docs == 'true'
 
     steps:
       - name: Wait for doc tests
@@ -74,7 +74,7 @@ jobs:
           owner: ${{ github.event.pull_request.head.repo.owner.login || github.repository_owner }}
           timeoutSeconds: ${{ env.WAIT_TIMEOUT_SECS }}
           intervalSeconds: ${{ env.WAIT_INTERVAL_SECS }}
-
+  
 
   quality:
     name: Code Quality
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -71,7 +71,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           checkName: Test Documentation
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-          owner: ${{ github.event.pull_request.head.repo.owner.login || github.repository_owner }}
+          # owner: ${{ github.event.pull_request.head.repo.owner.login || github.repository_owner }}
           timeoutSeconds: ${{ env.WAIT_TIMEOUT_SECS }}
           intervalSeconds: ${{ env.WAIT_INTERVAL_SECS }}
   
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -74,7 +74,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           checkName: Test Documentation
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-          owner: ${{ github.event.pull_request.head.repo.owner.login || github.repository_owner }}
+          # owner: ${{ github.event.pull_request.head.repo.owner.login || github.repository_owner }}
           timeoutSeconds: ${{ env.WAIT_TIMEOUT_SECS }}
           intervalSeconds: ${{ env.WAIT_INTERVAL_SECS }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/continous-integration.yml
```diff
@@ -70,10 +71,9 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           checkName: Test Documentation
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-          owner: ${{ github.event.pull_request.head.repo.owner.login || github.actor }}
           timeoutSeconds: ${{ env.WAIT_TIMEOUT_SECS }}
```
erohmensing, line 74: in the run you reference (https://github.com/RasaHQ/rasa/runs/2279118028#step:2:166) it exits after timing out without failing, do we need to change that as well (great job catching it anyway btw!)

erohmensing (commented):
With regards to 

> Remove the previous fix since we don't need to check docs in most of the PRs. Restrict the condition to run the wait for docs tests step as it is only relevant to docs changes. The step runs only on doc changes, or on tag pushes and pushes to the main branch exclusively, as it is a dependency of release jobs.

Does that mean it will fail if someone tries to update documentation from a fork? This is one of the most common community contributions

.github/workflows/continous-integration.yml
HotThoughts: > With regards to
> 
> > Remove the previous fix since we don't need to check docs in most of the PRs. Restrict the condition to run the wait for docs tests step as it is only relevant to docs changes. The step runs only on doc changes, or on tag pushes and pushes to the main branch exclusively, as it is a dependency of release jobs.
> 
> Does that mean it will fail if someone tries to update documentation from a fork? This is one of the most common community contributions

Yes, it will fail if someone updated the documentation and opened a pull request directly from their fork. But if the pull request got updated with the base branch, it will pass. 🤷🏻‍♀️

HotThoughts: > in the run you reference (https://github.com/RasaHQ/rasa/runs/2279118028#step:2:166) it exits after timing out without failing, do we need to change that as well (great job catching it anyway btw!)

We can catch the conclusion of this step and fail the step if the conclusion is not `success`.

.github/workflows/continous-integration.yml
```diff
@@ -61,18 +61,27 @@ jobs:
     name: Wait for docs tests
     runs-on: ubuntu-latest
     needs: [ changes ]
+    # if: needs.changes.outputs.docs == 'true' || github.event_name == 'push' && (startsWith(github.event.ref, 'refs/tags') || github.ref == 'refs/heads/main')
```
erohmensing, line 64:
```suggestion
```

.github/workflows/continous-integration.yml
```diff
@@ -61,18 +61,27 @@ jobs:
...
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           checkName: Test Documentation
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
-          owner: ${{ github.event.pull_request.head.repo.owner.login || github.actor }}
           timeoutSeconds: ${{ env.WAIT_TIMEOUT_SECS }}
           intervalSeconds: ${{ env.WAIT_INTERVAL_SECS }}
+          
+      - name: fail the step if the doc tests run could not be found
+        if: ${{ steps.wait-for-doc-tests.conclusion == 'cancelled' || steps.wait-for-doc-tests.conclusion == 'failure' }}
```
erohmensing, line 81:
```suggestion
      - name: Fail the step if the doc tests run could not be found
        if: ${{ steps.wait-for-doc-tests.conclusion == 'timed_out' }}
```

I guess if the "docs tests" fails, that will stop the CI from passing? not sure if that means this one should fail, as it still did its job correctly (same if the conclusion is cancelled)

Looks like there is a specific outcome for if the wait timed out, in which case "waiting for the tests" failed. What do you think?

HotThoughts: 😆You are right. The wait for docs step will fail if the docs test conclusion is 'failure'. The target conclusion here is 'timed_out'. Totally missed the point yesterday.

HotThoughts:
@erohmensing Thanks for your review. I update the code based on your suggestion. Now the job uses the head sha to find the docs test workflow in the base repo. The CI looks good

```
Retrieving check runs named Test Documentation on RasaHQ/rasa@d6f4eb2db31a967c97a00267002e156630e6198a...
Retrieved 1 check runs named Test Documentation
Found a completed check with id 2297832862 and conclusion skipped
```
In addition, I create a new PR and see if it can find the workflow run. It can! #8401 🤩 

This PR is quite dirty containing many testing commits. For the simplicity of the commit history, shall we close this PR and merge #8401 instead? What do you think?

---

Note:
In case you wonder why `steps.wait-for-doc-tests.conclusion` is changed to `steps.wait-for-doc-tests.outputs.conclusion.`:
The conclusion `timed out` is the output under the variable name `conclusion` from the step (given by the [action-wait-for-check ](https://github.com/fountainhead/action-wait-for-check) we use here), not the conclusion of the step (given by GitHub Action).

erohmensing:
> The conclusion timed out is the output under the variable name conclusion from the step (given by the action-wait-for-check we use here), not the conclusion of the step (given by GitHub Action).

Ah nice catch! I wouldn't have even noticed 👀 

Yes, sounds good, let's do that since it's already open, but in the future we can also "squash and merge" if we want to combine it into one commit before merging!

erohmensing:
I want to run one more test, which is to push a docs change to this branch to make sure the orginal docs test workflow runs where the other one can find it 👀 then will approve the other one :)

erohmensing:
It worked! ☃️

HotThoughts:
Close this because of #8401

PR #8376, opened 2021-04-07
Adding comment for wrongly predicted entities in failed_test_stories

Imod7:
**Issue**:
When running `rasa test` and the stories are tested from `tests/test_stories.yml`, we can see the results of the wrong predictions in the file `results/failed_test_stories.yml`. The comments in the `failed_test_stories.yml` denote the errors in the predictions. In the level of the entities, if there is a wrongly predicted entity until now there was no comment to show what the wrong prediction.

**Proposed changes**:
Added a comment in `failed_test_stories` at the level of each entity value that depicts the wrongly predicted entity value in the following format : 
```
entities:
- entity_name: entity_value  # predicted: wrong_predicted_entity: entity_value
- other_entity_name: other_entity_value  # predicted: other_wrong_predicted_entity_name: other_entity_value
```

**Example**:
Example that shows the result of the current implementation by forcing wrong predictions.

If I have in `test_stories.yml` a wrong annotation on purpose : 
```
  - user: |
      i am looking for [indian](feedback) food for [4](seating) persons
    intent: request_restaurant
```

The I will get a result of course in `failed_test_stories.yml` with the wrong predictions so I will have the following result : 
```
  - intent: request_restaurant  # predicted: request_restaurant: i am looking for [indian](cuisine) food for [4](number) persons
    entities:
    - feedback: indian  # predicted: cuisine: indian
    - seating: '4'  # predicted: number: 4
```


Imod7:
My intention was to try to add the comment with the wrongly predicted entity right next to the each specific entity like shown below : 
```
entities: 
- feedback: indian  # predicted: `number`
- seating: '4'      # predicted: `cuisine`
```
@joejuzl is this something we would like to have ? To implement that, I am still not quite sure where the specific entity is actually written in `failed_test_stories.yml` so to add the wrong prediction comment right next to it.

joejuzl:
> My intention was to try to add the comment with the wrongly predicted entity right next to the each specific entity like shown below

I think this would probably be the nicest solution, if it's not too difficult.
I think the entities are added inside `TrainingDataWriter.generate_message`. So you may instead need to do it in the `test` module so that you can add the comments.

Imod7:
@joejuzl Sad news! I cannot make it work to add a comment in the specific `entity` line. I also saw the `TrainingDataWriter.generate_message` that you mentioned where the entities are added and tried something there but did not work. From my understanding, the comments inserted use the `CommentedMap` and to put a comment it has to be attached to a specific key. 

For example, to add a comment in the `intent` line (in the failed stories) like this : 
`- intent: request_restaurant`

we already used this code : 
```
result.yaml_add_eol_comment(
                user_utterance.inline_comment(), KEY_USER_INTENT
            )
```

and to add a comment in the `entity` line like I do now : 
`entities: # predicted: number ....`

I used the same method as above so writing the following : 
```
result.yaml_add_eol_comment(
                user_utterance.inline_comment(KEY_ENTITIES), KEY_ENTITIES
            )
```
So what I tried to find is how to attach a key in each entity value but this is kind of hard because it is not a standard text like `intent` or `entities`, every time it changes. Then I also tried to add it when the entities are added but I did not have the CommentedMap object to use. Then I also checked if I could add the comments somehow directly in the dictionary object right next to each entity but then I did not know how to print it without the CommentedMap. Tbh, I tried a bunch of other stuff but still no luck 😭
Do you have any other ideas maybe ?

joejuzl:
In `as_story_string` it seems to be adding a comment directly by using `<!--` right? One option would be to re-implement the functionality `format_message` here to just add the entities and their respective comments?
However if this proves too difficult it's probably fine to have it all on the top entities line like you initially proposed.

Imod7:
In `results/failed_test_stories.yml` the specific entities (cuisine, number) under the key `entities`, e.g. :
```
    entities: 
    - cuisine: indian
    - number: '4'
```
are added [here](https://github.com/RasaHQ/rasa/blob/4411ddf0c8c5169c4d29d0fa71a9875fc5226a80/rasa/shared/core/training_data/story_writer/yaml_story_writer.py#L234)

so for example if I change this [line](https://github.com/RasaHQ/rasa/blob/4411ddf0c8c5169c4d29d0fa71a9875fc5226a80/rasa/shared/core/training_data/story_writer/yaml_story_writer.py#L231) : 

`entities.append(OrderedDict([(entity["entity"], entity["value"])]))`

to this
```
entity_value_with_comment = entity["value"] + " # sample comment"
entities.append(OrderedDict([(entity["entity"], entity_value_with_comment)]))
```

in `results/failed_test_stories.yml` I will see : 

```
   entities:
    - feedback: 'indian # sample comment'
    - seating: '4 # sample comment'
```
which is closer to what I had but still not quite correct since the comment is inside the entity value (`'indian # sample comment'`) and not separate (`'indian' # sample comment`). That is because it is an `OrderedDict` so I cannot print it separately if I want to print there.

joejuzl:
@Imod7 I managed to get comments working on each entity line by changing each entity `OrderedDict` to a `CommentedMap`. See this code snippet (modified from in `process_user_utterance`):

```
            entities = []
            for entity in user_utterance.entities:
                if "value" in entity:
                    entity_map = CommentedMap([(entity["entity"], entity["value"])])
                    entity_map.yaml_add_eol_comment('hello!', entity["entity"])
                    entities.append(entity_map)
                else:
                    entities.append(entity["entity"])
            result[KEY_ENTITIES] = entities
```

Does that help?
You will still need a way to get the right comment for each entity. Let me know if you need any help with that :)

Imod7:
@joejuzl wowowowowowoowow! Joe you did it!!! Amaaaazing!!! 🏆 🥇 Convert it to a `CommentedMap`, wow, that is smart and that way you still access first level keys!!! Craaaazy! THANK YOUUUUU !!! 🙏🏽 Ok, I will work with that now and let you know!!!

Imod7:
Now I am trying to write a test to see if comments are correctly added next to the wrongly predicted entity values. So I am looking at this [test](https://github.com/RasaHQ/rasa/blob/78fce3bf922f4ea99c9cb5f3355ba583502305ee/tests/core/test_evaluation.py#L286) to make something similar but for entities. I am not sure exactly how it works so I am looking into it to adapt it correctly.

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
```diff
@@ -228,7 +228,26 @@ def process_user_utterance(
             entities = []
             for entity in user_utterance.entities:
                 if "value" in entity:
-                    entities.append(OrderedDict([(entity["entity"], entity["value"])]))
+                    if user_utterance.__dict__.get("predicted_entities"):
```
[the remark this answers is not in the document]
joejuzl: How about having a new method on `WronglyClassifiedUserUtterance` like `inline_comment_for_entitiy(entity)` - then you could just check if the method exists like this line `hasattr(user_utterance, "inline_comment")`? This would also mean keeping the functionality encapsulated in `WronglyClassifiedUserUtterance`.

The issue with moving `WronglyClassifiedUserUtterance` to shared is that it can then be used by Rasa X and we would have to be careful about ever updating it.

data/test_response_selector_bot/config.yml
```diff
@@ -4,8 +4,8 @@ pipeline:
   - name: "WhitespaceTokenizer"
   - name: "CountVectorsFeaturizer"
   - name: "DIETClassifier"
-    entity_recognition: False
-    epochs: 8
+    entity_recognition: True
+    epochs: 17
```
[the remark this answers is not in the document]
joejuzl: Yeah this is an issue we've run into before. Especially with testing `rasa test` it requires the model to provide deterministic (and correct?) predictions. I would try creating a new bot as you suggested, with more focused data. Also you can try setting `random_seed` to make sure it's deterministic. Then see how low you can get the epochs.

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
```diff
@@ -228,7 +229,30 @@ def process_user_utterance(
             entities = []
             for entity in user_utterance.entities:
                 if "value" in entity:
-                    entities.append(OrderedDict([(entity["entity"], entity["value"])]))
+                    if hasattr(user_utterance, "inline_comment_for_entity"):
+                        for predicted in user_utterance.predicted_entities:
+                            if predicted["start"] == entity["start"]:
+                                if predicted["entity"] != entity["entity"]:
```
Imod7, line 235: I added this check because otherwise if I have more than one entities and one is correctly predicted and the other is wrongly predicted then it does not print the correct one 🤦🏽‍♀️ 

This happens because with the check `if hasattr(user_utterance, "inline_comment_for_entity"):` I get always `True` if : 
- I have at least one wrongly predicted entity but I don't know which one it is beforehand. 
- I have all entities correct and the `intent` is wrong because this just checks if the method is there and it is there if the object is of the class `WronglyClassifiedUserUtterance`.

These two cases give me also `True` for the `if hasattr(user_utterance, "inline_comment"):` check and that is why I added also an extra check on intent on line 209.

tests/core/test_evaluation.py
```diff
@@ -495,3 +495,78 @@ def test_log_evaluation_table(caplog, skip_field, skip_value):
         assert f"Classification report: \n{kwargs['report']}" in caplog.text
     else:
         assert "Classification report:" not in caplog.text
+
```
Imod7, line 498: I added three tests because my change also affects if a comment is printed in the `intent` line. Before these changes, we always printed a comment in the `intent` line regardless if the prediction error was in the `intent` or the `entities`. Now we print only where the wrong prediction so : 
- If intent is correct and there is a wrongly predicted entity -> we print a comment only on entity line
- if intent is wrongly predicted and entities correct -> we print only on intent line
- if both are wrongly predicted --> we print on intent and entity line

That is why I added the 3 separate tests that corresponds to each case.

data/test_yaml_stories/test_prediction_with_correct_intent_wrong_entity.yml
```diff
@@ -0,0 +1,6 @@
+stories:
+- story: test story with correct intent and wrong entity
```
Imod7, line 2: Something else that I tested locally with the normal formbot but did not add to the tests because I did not want to make it slower (I would have to add another entity ?), is to have more than one entities. For example, have one wrongly predicted entity and one correctly predicted entity. That way I can check if it also prints all the entities regardless if there is a wrong prediction or not.

Imod7:
Hey @joejuzl I pushed the last changes and added some comments! Looking forward for your feedback because I think there is a better way to do some of the things done 🧐 or even more correct. Thank you!

data/test_yaml_stories/test_prediction_with_correct_intent_wrong_entity.yml
joejuzl: I think it would make sense to have this in the test.

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
```diff
@@ -206,9 +206,10 @@ def process_user_utterance(
             result[KEY_USER_INTENT] = user_utterance.intent_name
 
         if hasattr(user_utterance, "inline_comment"):
-            result.yaml_add_eol_comment(
-                user_utterance.inline_comment(), KEY_USER_INTENT
-            )
+            if user_utterance.predicted_intent != user_utterance.intent["name"]:
```
joejuzl, line 209: why do we need this check now?

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
joejuzl: I think this logic should not be here. The `YAMLStoryWriter` should not have to know how to determine in an entity prediction is correct.
What if we changed `inline_comment_for_entity` to return `None` if there is no comment (i.e. it was correct), and then we just only add it if it returns something?

tests/conftest.py
```diff
@@ -501,6 +501,24 @@ async def trained_response_selector_bot(trained_async: Callable) -> Path:
     return Path(zipped_model)
 
 
+@pytest.fixture(scope="session")
+async def trained_formbot(trained_async: Callable) -> Path:
```
joejuzl, line 505: Why `formbot`?

tests/core/test_evaluation.py
```diff
@@ -495,3 +495,78 @@ def test_log_evaluation_table(caplog, skip_field, skip_value):
...
+
+@pytest.mark.parametrize(
+    "test_file",
+    ["data/test_yaml_stories/test_prediction_with_correct_intent_wrong_entity.yml",],
```
joejuzl, line 502: why paramatrize if there is only one option?

joejuzl: maybe these three tests can be combined?

joejuzl (changes requested):

tests/conftest.py
Imod7: Why I named it formbot? Because I took the least data possible from `examples/formbot`.

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
Imod7: True! I did not think about which module should know what. Good point! This is what they call [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) ?

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
Imod7: Because as mentioned [here](https://github.com/RasaHQ/rasa/pull/8376#discussion_r633310981) for the entities, this line
`if hasattr(user_utterance, "inline_comment"):`
will always be `True` even if the wrong prediction is in entities or intent. So it will not distinguish. It will always return `True` because in any wrong prediction we will end up with an object of class `WronglyClassifiedUserUtterance` which always `has attribute ` the `inline_comment`. 

So with this check, if the wrong prediction is in entities but the intent is correct, I do not print any comment in the `intent` line.

tests/core/test_evaluation.py
Imod7: > maybe these three tests can be combined?

True! I forgot to change it because I had everything in one test before.

> maybe these three tests can be combined?

I had it like this before with an extra variable parametrized and then inside the test function use this variable in `if` statements. But then I remembered a comment from Tobias [here](https://github.com/RasaHQ/rasa/pull/7931#discussion_r615957054) where he alternatively suggests to split in different tests because the if statements makes it hard to read. But maybe it is because in that case I had a whole domain added in every if statement. 

I will put them all in one and maybe in this case it is not that hard to read.

data/test_yaml_stories/test_prediction_with_correct_intent_wrong_entity.yml
Imod7: I added it.

tests/core/test_evaluation.py
Imod7: I put the 3 tests into the same function and it actually goes faster because before each test was taking ~30seconds to finish so in total 90seconds all 3 and now that I have them in the same function it takes ~31seconds to complete. So very good call Joe!!! 👌🏾This happens because it does the formbot training once right?

tests/conftest.py
joejuzl: I was referring to the name. There is no form in the bot now right? So maybe a different name would be clearer

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
```diff
@@ -228,7 +229,29 @@ def process_user_utterance(
             entities = []
             for entity in user_utterance.entities:
                 if "value" in entity:
-                    entities.append(OrderedDict([(entity["entity"], entity["value"])]))
+                    if hasattr(user_utterance, "inline_comment_for_entity"):
+                        for predicted in user_utterance.predicted_entities:
+                            if user_utterance.inline_comment_for_entity(
```
joejuzl, line 234: If you put this into a variable you won't have to call the same method again on line 241

tests/conftest.py
Imod7: Great catch! 💯 I didn't even think about it. I just took it from the formbot so I kept the same name and when I removed the basic logic that makes this bot a formbot, I did not think of renaming it. It is so simple but so easy to miss it.

Imod7:
I checked the docs to see if it mentions how the comments are printed in the `failed_test_stories.yml` file but I did not find something. 

The only thing I found is that the file `failed_test_stories.yml` is mentioned [here](https://rasa.com/docs/rasa/2.0.x/testing-your-assistant#evaluating-a-dialogue-model) where is says the following : 

_This will print any failed stories to results/failed_test_stories.yml. A story fails if at least one of the actions was predicted incorrectly._

I am not sure if this needs an update that the story fails if at least an action or a user utterance (entity or intent) was predicted incorrectly. What do you think?

tests/core/test_evaluation.py
```diff
@@ -495,3 +495,86 @@ def test_log_evaluation_table(caplog, skip_field, skip_value):
...
+        assert "- intent: request_restaurant  # predicted:" not in failed_stories
+        # check if there is a comment with the predicted entity on the entity line
+        assert "# predicted: cuisine: greek" in failed_stories
+        # check that the correctly predicted entity is printed as well
+        assert "- seating: outside\n" in failed_stories
+        # check that it does not double print entities
+        assert failed_stories.count("\n") == 8
+
+    elif test_case == "wrong_intent_correct_entity":
+        await evaluate_stories(
```
joejuzl, line 543: This is duplicate code and can be extracted to the top right?

tests/core/test_evaluation.py
```diff
@@ -495,3 +495,86 @@ def test_log_evaluation_table(caplog, skip_field, skip_value):
...
+            "data/test_yaml_stories/test_prediction_with_correct_intent_wrong_entity.yml",
+            "correct_intent_wrong_entity",
+        ],
+        [
+            "data/test_yaml_stories/test_prediction_with_wrong_intent_correct_entity.yml",
+            "wrong_intent_correct_entity",
+        ],
+        [
+            "data/test_yaml_stories/test_prediction_with_wrong_intent_wrong_entity.yml",
+            "wrong_intent_wrong_entity",
```
joejuzl, line 513: maybe rather than a string it would be easier to have 2 booleans e.g. `correct_intent` and `connect_entity`?

joejuzl (changes requested):
just a couple of comments about the test.

joejuzl:
> I checked the docs to see if it mentions how the comments are printed in the failed_test_stories.yml file but I did not find something that needs an update related to that.

However the file failed_test_stories.yml is mentioned here where is says the following :

This will print any failed stories to results/failed_test_stories.yml. A story fails if at least one of the actions was predicted incorrectly.

I am not sure if this needs an update that the story fails if at least an action or a user utterance (entity or intent) was predicted incorrectly. What do you think?

Seems fine to me I think.

joejuzl (approved):
Nice job 👍

PR #8385, opened 2021-04-07
Stop pytest timeouts from including fixture duration and reduce global timeout.

### LINKED ISSUE #8122: Update `pytest-timeout` usage
**Description of Problem**:
Our `pytest.mark.timeouts` are
* too big by default (120 seconds) to identify slow tests
* include the fixture setup time which means that if a fixture is very slow and one test fails due to a timeout, the next one is likely to fail as well 

**Overview of the Solution**:
* Have a low default timeout so we can identify slow tests outside of the machine learning / integration tests quickly.
* only measure the pure test runtime and don't include the fixture setup time

**Blockers** (if relevant):
* https://github.com/RasaHQ/rasa/issues/8118 needs to be done first

**Definition of Done**:
- [ ] Set pytest.timeout to `10` for all tests except 
  * Core featurizers / policies
  * NLU featurizers / components
  * integration tests
- [ ] Make sure that every increased timeout in above exceptions (which are annotated using `@pytest.mark.timeout`) use `func_only=True`. This excludes the fixture time to be included in the timeout
- [ ] optional: Ensure that no tests outside ML + integration can have a customized `pytest.mark.timeout`

joejuzl:
Fixes #8122.

**Proposed changes**:
- Use global and decorator level `fun_only=True` to stop including fixture duration in timeout. (See: https://github.com/pytest-dev/pytest-timeout/issues/36 for why it has to be on every decorator)
- Decrease the global timeout from 120s to 60s. (lower numbers resulted in more failures)
- At timeout decorator to two tests that fail with 60s global limit.


alwx (approved):
I have to say that even finding any information about `func_only` n `python-timeout` repo is quite hard — there is basically nothing in the `README` about this and there is no documentation or whatever. So yeah, definitely a good catch!

ancalita (approved):
Looks complete 🚀

Issue #8135, opened 2021-03-07
Rasa test/rasa test core can't find any intent in domain


subercui:
**Rasa version**:
2.3.4
**Rasa SDK version** (if used & relevant):
2.3.1
**Rasa X version** (if used & relevant):

**Python version**:
3.8.8
**Operating system** (windows, osx, ...):
Linux-4.4.0-186-generic-x86_64-with-glibc2.10
**Issue**:
This is annoying. There is no error in training and testing manually, but `rasa test/rasa test core` can't find any intent in domain.

**Error (including full traceback)**:
Found intent 'xxxx' in stories which is not part of the domain.
```
/h/haotian/.conda/envs/rasa/lib/python3.8/site-packages/rasa/shared/utils/io.py:93: UserWarning: Issue found in '/tmp/tmp3z0di4sm/0441e511624c4c069bf1b3cf0526e1b4_test_stories.yml': 
Found intent 'how_can_i_help_you' in stories which is not part of the domain.
  More info at https://rasa.com/docs/rasa/stories
/h/haotian/.conda/envs/rasa/lib/python3.8/site-packages/rasa/shared/utils/io.py:93: UserWarning: Issue found in '/tmp/tmp3z0di4sm/0441e511624c4c069bf1b3cf0526e1b4_test_stories.yml': 
Found intent 'tell_what_we_have' in stories which is not part of the domain.
  More info at https://rasa.com/docs/rasa/stories
/h/haotian/.conda/envs/rasa/lib/python3.8/site-packages/rasa/shared/utils/io.py:93: UserWarning: Issue found in '/tmp/tmp3z0di4sm/0441e511624c4c069bf1b3cf0526e1b4_test_stories.yml': 
Found intent 'ask_color' in stories which is not part of the domain.
  More info at https://rasa.com/docs/rasa/stories
/h/haotian/.conda/envs/rasa/lib/python3.8/site-packages/rasa/shared/utils/io.py:93: UserWarning: Issue found in '/tmp/tmp3z0di4sm/0441e511624c4c069bf1b3cf0526e1b4_test_stories.yml': 
Found intent 'ask_quantity' in stories which is not part of the domain.
  More info at https://rasa.com/docs/rasa/stories
[...]

PR #8388, opened 2021-04-07
Fix for missing intent warnings when running rasa test

b-quachtran:
**Proposed changes**:
- Fix for missing intent warnings when running rasa test
- Closes issue #8135 


m-vdb:
@b-quachtran thanks for pushing fix -- should we ask someone from Enable to review this PR?

b-quachtran:
Hey @m-vdb, yes definitely. I still need to wrap up a few things in this PR related to test cases, but will tag Enable as soon as it's ready.

rasa/core/test.py
```diff
@@ -372,9 +372,20 @@ async def _create_data_generator(
...
+        core_model, _ = get_model_subdirectories(agent.model_directory)
+
+    if core_model and os.path.exists(os.path.join(core_model, DEFAULT_DOMAIN_PATH)):
+        domain_path = os.path.join(core_model, DEFAULT_DOMAIN_PATH)
+    else:
+        domain_path = None
 
     test_data_importer = TrainingDataImporter.load_from_dict(
-        training_data_paths=[resource_name]
+        training_data_paths=[resource_name], domain_path=domain_path
```
joejuzl, line 388: Why can't we just use `DEFAULT_DOMAIN_PATH` like in NLU?

joejuzl (changes requested):
Nice one 👍 
Is there an easy to test this is fixed with a unit test?

rasa/core/test.py
b-quachtran: Hmm, so this was something I wasn't sure about. In the Core case, the domain file is also stored inside the model file unlike with NLU, which is why NLU has to use the `DEFAULT_DOMAIN_PATH` but with Core it can be either one (model or default path). Does it make sense from a best practices perspective to keep it consistent here? I'm also thinking about the case when the domain file isn't in the default location, since the above will work regardless of where it's stored.

b-quachtran:
@joejuzl Yep, I'll add a unit test as well 👍

rasa/core/test.py
joejuzl: hmm yeah it's a weird situation. I guess ideally we would always package the domain with the model. This seems fine for now though.

joejuzl (approved):

nearlyhuman-jwheat:
I've just updated to the recent release of 2.7.0 and the first thing I did was check to see if this issue was fixed.  

I ran 

`rasa test --stories tests/test_stories.yml --fail-on-prediction-errors`

And it can't find my domain.yml files.

```yaml
/home/jwheat/anaconda2/envs/rasa263/lib/python3.7/site-packages/rasa/shared/utils/io.py:97: UserWarning: Loading domain from 'domain.yml' failed. Using empty domain. Error: 'Failed to load domain specification from '/mnt/d/Code/NearlyHuman/RASA2/rasa-demo-server/domain.yml'. File not found!'
```

I admit I have a more robust directory structure, but when I train I add `--domain data` and it finds them all, however `rasa test` doesn't accept that flag, so I get screens of 

```
/home/jwheat/anaconda2/envs/rasa263/lib/python3.7/site-packages/rasa/shared/utils/io.py:97: UserWarning: Issue found in 'data/general-support/stories.yml':    
Found intent 'affirm' in stories which is not part of the domain.
```

More importantly, all my test stories fail, I'm guessing because it can't find my domain.yml files.

FWIW my structure looks like this where I have individual "skills" broken out into their own directory structure.

```yml
data/
    /skill1/
           domain.yml
           rules.yml
           nlu.yml
           stories.yml
    /skill2/
           domain.yml
           rules.yml
           nlu.yml
           stories.yml
etc.
```

I have no files in the root of my `data` directory and it looks like that's where this version is looking based on the first UserWarning.

Is there any way to get `rasa test` to see my structure and act like `rasa train` does with a `--domain data` flag?  

Also curious what it does with the training data itself.  Shouldn't `rasa test` only use the pre-trained model?  Why does it need to look at all the domain files?

PR #8393, opened 2021-04-08
Add missing type annotations

### LINKED ISSUE #8248: mypy: Enforce remaining type annotations
[Parent issue](https://github.com/RasaHQ/rasa/issues/7513)

To enforce type annotations which should start running `mypy` with the `--disallow-untyped-defs` and `--disallow-untyped-decorators` flag.

Currently blocked by:
* https://github.com/RasaHQ/rasa/issues/8246
* https://github.com/RasaHQ/rasa/issues/8247

This currently causes: 52+x issues

**Definition of Done**:
- [ ] the Rasa Open Source CI runs `make types` with the `--disallow-untyped-defs` and `--disallow-untyped-decorators`
- [ ] all detected mypy issues are fixed

ancalita:
**Proposed changes**:
- Fix #8248 


rasa/core/channels/channel.py
```diff
@@ -13,6 +13,7 @@
     Iterable,
     Awaitable,
     NoReturn,
+    Coroutine,
```
wochinge, line 16: can this be removed?
```suggestion
```

rasa/core/run.py
```diff
@@ -121,7 +121,7 @@ def configure_app(
         rasa.core.utils.list_routes(app)
 
     # configure async loop logging
```
wochinge, line 123: not your change, but I think that comment is useless (the method name says exactly the same)
```suggestion
```

rasa/core/tracker_store.py
```diff
@@ -991,7 +993,7 @@ def _create_database(engine: "Engine", database_name: Text) -> None:
         conn.close()
 
     @contextlib.contextmanager
-    def session_scope(self):
+    def session_scope(self) -> Generator["sessionmaker", None, None]:
```
wochinge, line 996: this yields a `Session`, doesn't it? `sessiomaker()` returns a `Session` as far as I know

rasa/core/training/interactive.py
```diff
@@ -1603,17 +1615,19 @@ def start_visualization(image_path: Text, port: int) -> None:
 
     # noinspection PyUnusedLocal
     @app.exception(NotFound)
-    async def ignore_404s(request, exception):
+    async def ignore_404s(request: Any, exception: Any) -> HTTPResponse:
```
wochinge, line 1618: ?
```suggestion
    async def ignore_404s(request: Request, exception: Exception) -> HTTPResponse:
```

rasa/core/training/interactive.py
```diff
@@ -1603,17 +1615,19 @@ def start_visualization(image_path: Text, port: int) -> None:
...
         return response.text("Not found", status=404)
 
     # noinspection PyUnusedLocal
     @app.route(VISUALIZATION_TEMPLATE_PATH, methods=["GET"])
-    def visualisation_html(request):
+    def visualisation_html(request: Any) -> Coroutine[Any, Any, HTTPResponse]:
```
wochinge, line 1623: why `Coroutine`?
```suggestion
    def visualisation_html(request: Request) -> HTTPResponse:
```

rasa/core/training/interactive.py
```diff
@@ -1603,17 +1615,19 @@ def start_visualization(image_path: Text, port: int) -> None:
...
         return response.file(visualization.visualization_html_path())
 
     # noinspection PyUnusedLocal
     @app.route("/visualization.dot", methods=["GET"])
-    def visualisation_png(request):
+    def visualisation_png(
+        request: Any,
+    ) -> Union[Coroutine[Any, Any, HTTPResponse], HTTPResponse]:
```
wochinge, line 1630:
```suggestion
        request: Request,
    ) -> HTTPResponse:
```

rasa/core/training/interactive.py
wochinge, line 1623: the type annotations should in my opinion be similar to the ones in `server.py`

rasa/nlu/model.py
```diff
@@ -93,14 +93,14 @@ def get(self, property_name: Text, default: Any = None) -> Any:
         return self.metadata.get(property_name, default)
 
     @property
-    def component_classes(self):
+    def component_classes(self) -> List[Optional[Any]]:
```
wochinge, line 96: we get a list of component class names, no?
```suggestion
    def component_classes(self) -> List[Optional[Text]]:
```

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -40,7 +40,7 @@ def set(self, prop: Text, info: Any) -> None:
     def get(self, prop: Text, default: Optional[Any] = None) -> Any:
         return self.data.get(prop, default)
 
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> Any:
```
wochinge, line 43:
```suggestion
    def __eq__(self, other: Any) -> bool:
```

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -50,7 +50,7 @@ def __eq__(self, other):
             other.lemma,
         )
 
-    def __lt__(self, other):
+    def __lt__(self, other: Any) -> Any:
```
wochinge, line 53:
```suggestion
    def __lt__(self, other: Any) -> bool:
```

rasa/shared/core/domain.py
```diff
@@ -1537,7 +1537,7 @@ def intent_config(self, intent_name: Text) -> Dict[Text, Any]:
         return self.intent_properties.get(intent_name, {})
 
     @rasa.shared.utils.common.lazy_property
-    def intents(self):
+    def intents(self) -> Text:
```
wochinge, line 1540: as it's sorted I'd expect it to be a collection?
```suggestion
    def intents(self) -> List[Text]:
```

setup.cfg
```diff
@@ -39,6 +39,7 @@ warn_redundant_casts = True
 warn_unused_ignores = True
 disallow_untyped_calls = True
 disallow_incomplete_defs = True
+disallow_untyped_defs = True
```
wochinge, line 42: what about the other `disallow_xy` from the issue? I think it was about decorators?

rasa/shared/utils/io.py
```diff
@@ -290,7 +290,7 @@ def json_to_string(obj: Any, **kwargs: Any) -> Text:
 def fix_yaml_loader() -> None:
     """Ensure that any string read by yaml is represented as unicode."""
 
-    def construct_yaml_str(self, node):
+    def construct_yaml_str(self: Any, node: Any) -> Any:
```
wochinge, line 293: no need to add typing for `self` 
```suggestion
    def construct_yaml_str(self, node: ScalarNode) -> Any:
```

rasa/shared/utils/io.py
```diff
@@ -306,7 +306,7 @@ def replace_environment_variables() -> None:
     env_var_pattern = re.compile(r"^(.*)\$\{(.*)\}(.*)$")
     yaml.Resolver.add_implicit_resolver("!env_var", env_var_pattern, None)
 
-    def env_var_constructor(loader, node):
+    def env_var_constructor(loader: BaseConstructor, node: Any) -> AnyStr:
```
wochinge, line 309:
```suggestion
    def env_var_constructor(loader: BaseConstructor, node: ScalarNode) -> AnyStr:
```

wochinge, line 309: what's a `AnyStr` and why is it not a pure `Text`?

wochinge (commented):
rapid PR during OKR week :100:

rasa/core/tracker_store.py
ancalita: Yes, you're right, it was initially Session before I knew I had to put it in `" "` 😄 , I'll change it now

rasa/core/training/interactive.py
ancalita: I have nothing to say in my defense 😄 amended

rasa/core/training/interactive.py
ancalita: I was guiding myself by PyCharm suggestions looking at what this function returns (IDE warns `Expected type 'HTTPResponse' , got 'Coroutine' instead` )

rasa/shared/utils/io.py
ancalita: PyCharm suggestion - seems to think it can return either `str` or `bytes`: https://docs.python.org/3/library/typing.html#typing.AnyStr

rasa/shared/utils/io.py
ancalita: It actually requires one - otherwise when I run `make types` it fails on this last issue.

rasa/shared/utils/io.py
wochinge: Thanks for the link! In our case it in our case `expandvars` will always return `str`, won't it?

rasa/shared/utils/io.py
wochinge: interesting!

rasa/core/channels/channel.py
```diff
@@ -81,7 +81,9 @@ def __init__(
 def register(
     input_channels: List["InputChannel"], app: Sanic, route: Optional[Text]
 ) -> None:
-    async def handler(*args, **kwargs):
+    """Registers input channel blueprints with the application."""
```
wochinge, line 84: optional: `application` sounds a bit vague to me. `app` is Sanic server instance so I'd be tiny bit more explicit here
```suggestion
    """Registers input channel blueprints with Sanic."""
```

rasa/core/channels/channel.py
```diff
@@ -81,7 +81,9 @@ def __init__(
...
+
+    async def handler(*args: Any, **kwargs: Any) -> None:
         await app.agent.handle_message(*args, **kwargs)
```
wochinge, line 87: this probably doesn't cause a mypy error because we have suppressed this kind of error. However, according to this [line](https://github.com/RasaHQ/rasa/blob/82e98163d3701f490716760213b92619283a4a00/rasa/core/channels/channel.py#L111) the handler should be of type:

```suggestion
    async def handler(message: UserMessage) -> None:
        await app.agent.handle_message(message)
```

rasa/core/policies/sklearn_policy.py
```diff
@@ -130,10 +130,11 @@ def _default_model() -> Any:
         return LogisticRegression(solver="liblinear", multi_class="auto")
 
     @property
-    def _state(self):
+    def _state(self) -> Dict[Text, Any]:
         return {attr: getattr(self, attr) for attr in self._pickle_params}
 
     def model_architecture(self, **kwargs: Any) -> Any:
+        """Sets valid parameters to the model."""
```
wochinge, line 137:
```suggestion
        """Sets model parameters for training."""
```

rasa/core/tracker_store.py
```diff
@@ -364,8 +366,8 @@ def _set_key_prefix(self, key_prefix: Text) -> None:
     def _get_key_prefix(self) -> Text:
         return self.key_prefix
 
-    def save(self, tracker, timeout=None):
-        """Saves the current conversation state"""
+    def save(self, tracker: DialogueStateTracker, timeout: Any = None) -> None:
```
wochinge, line 369: the timeout has to be some sort of number, hasn't it? Maybe you can also look if we actually use the `timeout` parameter anywhere and potentially just remove it 🤷🏻 
```suggestion
    def save(self, tracker: DialogueStateTracker, timeout: Optional[float] = None) -> None:
```

rasa/shared/core/training_data/story_reader/markdown_story_reader.py
```diff
@@ -3,6 +3,12 @@
 import os
 import re
 from pathlib import Path
+import sys
+
+if sys.version_info >= (3, 7):
```
wochinge, line 8: can we please add a comment why we need that? 🙌🏻

rasa/shared/utils/io.py
```diff
@@ -290,7 +290,7 @@ def json_to_string(obj: Any, **kwargs: Any) -> Text:
 def fix_yaml_loader() -> None:
     """Ensure that any string read by yaml is represented as unicode."""
 
-    def construct_yaml_str(self, node):
+    def construct_yaml_str(self: Any, node: ScalarNode) -> Any:
```
wochinge, line 293: ?
```suggestion
    def construct_yaml_str(self: BaseConstructor, node: ScalarNode) -> Any:
```

rasa/core/training/interactive.py
wochinge: seems Pycharm can't really deal with coroutines and type annoations yet 🤔

wochinge (commented):
Thanks for the docstrings 💯

rasa/core/channels/channel.py
ancalita: Hmm, doesn't that line 111 actually show the output of `handler` function is a UserMessage that becomes the input for `channel.blueprint()'? When I try to find the declaration of `agent.handle_message` it appears with `*args` and `**kwargs` [here](https://github.com/RasaHQ/rasa/blob/7e9db8cd6c97b8109a2cd138b16ab635815d7cc6/rasa/core/agent.py#L563) ... unless I'm looking in the wrong place? 🤔

wochinge: The output of `handler` is `None`. The `handler` itself (a function) becomes input to `channel.blueprint`.

You're absolutely right about the `*args`, `**kwargs` thing 💯 The interface definition [here](https://github.com/RasaHQ/rasa/blob/82e98163d3701f490716760213b92619283a4a00/rasa/core/channels/channel.py#L111) specifies the handler as a function which **only**  receives a `UserMessage` object and nothing else. In my opinion these two (interface definition and `handler` need to be consistent`. I prefer the interface definition as this is also the way how the handler (aka `on_new_message`) is used by the channels

wochinge (approved):

rasa/core/channels/channel.py
ancalita: I see now, thanks for the explanation 👍

PR #8400, opened 2021-04-08
[Testing] A dummy branch for testing

HotThoughts:
Testing for https://github.com/RasaHQ/rasa/pull/8372


HotThoughts:
Close it because it may not a good example for testing.

PR #8401, opened 2021-04-08
fix wait for docs test step in CI

HotThoughts:
**Proposed changes**:
Change ref for `action-wait-for-check` action to the sha of the head branch such that the step can use it to find the target workflow run in the base repo.

**Status (please check what you already did):**
- [X] Add back the checkout step and remove the owner settings. Now the ref used to look for workflow run is the head sha.
- [X] Add a new step to fail the job if the conclusion is `timed_out`.

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -63,6 +63,9 @@ jobs:
     needs: [ changes ]
 
     steps:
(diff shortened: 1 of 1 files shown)


erohmensing (approved):
Our testing on https://github.com/RasaHQ/rasa/pull/8372 seems successful! 🎉

PR #8402, opened 2021-04-08
Update the ConvertRT Featurizer template URL

### LINKED ISSUE #8229: Persist ConveRT model in a local persistent cache
**Description of Problem**: Currently, the `ConveRTFeaturizer` downloads the ConveRT model from a remote repository and stores it in the temporary cache directory of tensorflow hub which is located at `$HOME/.cache`. Since, this is a temporary cache directory it gets wiped out on every system restart. This is even more problematic for docker setups as you can imagine much more frequent docker container restarts during experimentation because of which the ConveRT model needs to be downloaded everytime from scratch. There has also been an issue in the past where the ConveRT model was taken offline from the central remote repository and it caused some assistants in production to break.

**Overview of the Solution**: There are a couple of solutions possible:

1.  We add a configuration option to `ConveRTFeaturizer` called `cache_dir` (or something similar) which when specified, stores the ConveRT model to a persistent directory. This directory path should be set as the value for `TFHUB_CACHE_DIR` environment variable so that during training the model gets persisted at this location automatically and during inference the model is loaded from this path. Once this option is added, docs should suggest users to use this option necessarily in production.
2.  We could use a default persistent cache directory if Rasa uses one and persist / load the model from there just like we do in option 1.

**Definition of Done**:

*   [ ] Tests are added
*   [ ] Feature described the docs
*   [ ] Feature mentioned in the changelog

guilherme-mendes:
**Proposed changes**:
- Update the valid link pass as a value for the ConveRTFeaturizer model_url option.
- Resolves #8229 

### DIFF
--- a/changelog/8229.enhancement.md
+++ b/changelog/8229.enhancement.md
@@ -1,0 +1,1 @@
+Persist ConveRT model in a local persistent cache
--- a/rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
+++ b/rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
@@ -39,9 +39,7 @@ logger = logging.getLogger(__name__)
 
 # URL to the new remote location of the model which
 # users might use. The model is no longer hosted here.
-ORIGINAL_TF_HUB_MODULE_URL = (
-    "https://github.com/davidalami/ConveRT/releases/download/1.0/nocontext_tf_model.tar.gz"
-)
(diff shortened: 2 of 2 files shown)


sara-tagger:

ancalita:
@guilherme-mendes I see that the local branch was deleted 10 months ago, so I'm assuming you will not continue working on this?
I'm closing this PR in the meantime. Happy to receive a new PR if you want to follow-up on this improvement 💯

PR #8428, opened 2021-04-09
fix flake8 config and drop unused code

wochinge:
**Proposed changes**:
- drop unused code
- re-configure flake8 to flag unused code / imports


rasa/core/channels/twilio.py
```diff
@@ -10,7 +10,7 @@
 from rasa.core.channels.channel import UserMessage, OutputChannel
 
 if TYPE_CHECKING:
-    from twilio.rest.api.v2010.account.message import MessageInstance
+    from twilio.rest.api.v2010.account.message import MessageInstance  # noqa: F401
```
m-vdb, line 13: sounds a bit weird to require this comment, can't vulture respect `TYPE_CHECKING`?

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -378,7 +378,7 @@ def __init__(self) -> None:
         super().__init__()
         rasa.shared.utils.io.raise_warning(
```
m-vdb, line 379: can you use `raise_deprecation_warning` instead?

rasa/nlu/utils/__init__.py
```diff
@@ -6,13 +6,6 @@
 import rasa.shared.utils.io
```
m-vdb, line 7: FWIW none of the remaining utils in the file seem NLU specific

rasa/shared/core/events.py
```diff
@@ -111,18 +111,6 @@ def format_message(
     )
 
 
-def first_key(d: Dict[Text, Any], default_key: Any) -> Any:
```
m-vdb, line 114: I guess you've checked that everything that is in `rasa.shared.*` and removed in this PR isn't used by Rasa X?

rasa/shared/data.py
```diff
@@ -69,6 +70,10 @@ def get_test_directory(paths: Optional[Union[Text, List[Text]]]) -> Text:
     Returns:
         Path to temporary directory containing all found Core training files.
     """
+    rasa.shared.utils.io.raise_warning(
```
m-vdb, line 73: can you use `raise_deprecation_warning`?

rasa/shared/importers/autoconfig.py
```diff
@@ -263,7 +263,7 @@ def _get_commented_out_autoconfig_lines(
     autoconfig_lines = {}
 
     for key in auto_configured_keys:
-        stream = yaml.compat.StringIO()
+        stream = ruamel.yaml.compat.StringIO()
```
m-vdb, line 266: but `ruamel.yaml` is imported as `yaml` above 🤷🏻

rasa/shared/nlu/training_data/message.py
```diff
@@ -179,11 +179,10 @@ def get_full_intent(self) -> Text:
         )
 
     def get_combined_intent_response_key(self) -> Text:
-        """Get intent as it appears in training data"""
-
+        """Get intent as it appears in training data."""
         rasa.shared.utils.io.raise_warning(
```
m-vdb, line 183: can you use `raise_deprecation_warning`?

rasa/shared/nlu/training_data/training_data.py
```diff
@@ -464,8 +464,12 @@ def sorted_entities(self) -> List[Any]:
         return sorted(entity_examples, key=lambda e: e["entity"])
 
     def sorted_intent_examples(self) -> List[Message]:
-        """Sorts the intent examples by the name of the intent and then response"""
-
+        """Sorts the intent examples by the name of the intent and then response."""
+        rasa.shared.utils.io.raise_warning(
```
m-vdb, line 468: +1

tests/nlu/test_persistor.py
```diff
@@ -7,51 +9,13 @@
...
 import rasa.nlu.train
 from rasa.nlu.config import RasaNLUModelConfig
+from rasa.nlu.persistor import Persistor
 
 
 class Object:
     pass
 
 
-async def test_list_method_method_in_AWS_persistor(  # noinspection PyPep8Naming
```
m-vdb, line 16: why removing all of these tests?

m-vdb (commented):
Do you have an example of how it runs in the CI? I can't find an example of workflow with a failing check. Do we need some sort of documentation to explain what to do if the check fails?

rasa/core/featurizers/single_state_featurizer.py
wochinge: sure, also changing it to `FutureWarning` . `DeprecationWarning` doesn't make sense here.

rasa/shared/core/events.py
wochinge: yes!

rasa/shared/importers/autoconfig.py
wochinge: yeah 🤷🏻 I somewhat agree with vulture that it is weird to import a module / package by alias and then access a submodule in this package through the alias.

tests/nlu/test_persistor.py
wochinge: I removed the list methods of all persistors as it's not used anywhere. Hence, the tests are going with it.

rasa/core/channels/twilio.py
wochinge: yeah - it sometimes works and sometimes doesn't (e.g. it doesn't raise for any other imports in `TYPE_CHECKING`)

rasa/nlu/utils/__init__.py
wochinge: yep, I'd say it's out of the scope of this PR though.

wochinge:
pushed a commit which triggers a vulture error. Also happy to skip adding vulture to the CI for now. It's not perfect but in my opinion a good start to have a least some check for it.

rasa/shared/importers/autoconfig.py
m-vdb: can we either remove the renaming of the module or tweak vulture to disregard these? sounds like a bug of vulture

rasa/core/featurizers/single_state_featurizer.py
m-vdb: why? I think that it has implication on how it is displayed to users. I wouldn't change this without looking at the impact on UX, and I think it's beyond the scope of the PR 🤔

rasa/core/channels/twilio.py
m-vdb: sounds a potential for flakiness. What can we do about it? Is this an issue Vulture people know about?

m-vdb:
@wochinge it didn't fail because the lint for docstrings failed first :D

I don't think it's a good idea to not add it to the CI, because it will become dead code too (a bit ironic ahah)

rasa/core/featurizers/single_state_featurizer.py
wochinge: That's exactly why. Users should see the warning and hence it should be `FutureWarning`. `DeprecationWarnings` don't print to the shell and make only sense if this is used from the Python API of rasa which isn't the case for these featurizers.

rasa/core/channels/twilio.py
wochinge: Funny enough: They use `rasa_nlu` to discuss the handling of `if TYPE_CHECKING` 😆

wochinge: omg, our flake8 config is evil! Will probably dump vulture and use flake8 instead

m-vdb: wdym?

rasa/shared/importers/autoconfig.py
wochinge: done

rasa/core/channels/twilio.py
wochinge: we configured flake8 to not flag unused imports 🙈

setup.cfg
```diff
@@ -22,7 +22,10 @@ max-doc-length = 88
 # See these resources for descriptions of the error codes:
 # E***, F***, W*** : https://flake8.pycqa.org/en/latest/user/error-codes.html
 # D***: http://www.pydocstyle.org/en/stable/error_codes.html
-ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, E231,
+ignore = W503, # line break before binary operator
+    E203,  # whitespace before ‘:’
+    E231,  # missing whitespace after ‘,’, ‘;’, or ‘:’
+    E501,  # too long lines - FIX THIS IN SEPARATE PR
```
wochinge, line 28: will tackle this separately

wochinge:
I dropped vulture and now use flake8 instead @m-vdb Removed more unused imports - flake8 is a lot better than vulture for this. Dropped vulture since it has to many false positives.

m-vdb:
but can flake8 detect dead code? unused import are a subset of dead code

wochinge:
not as good as vulture - but vulture has to many false positives so we can't really run it as part of the CI

rasa/shared/data.py
wochinge: changed these back as I need type `DeprecationWarning` in this case and I can't specify a category with `raise_deprecation_warning`

wochinge:
@m-vdb Would be great if we could get this in before my vacation 👀 The review should be quick despite its size as it's just removing dead code.

rasa/nlu/components.py
```diff
@@ -13,7 +13,6 @@
 from rasa.shared.exceptions import InvalidConfigException
 from rasa.shared.nlu.training_data.training_data import TrainingData
 from rasa.shared.nlu.training_data.message import Message
-from rasa.nlu.constants import COMPONENT_INDEX, TRAINABLE_EXTRACTORS
```
m-vdb, line 16: `COMPONENT_INDEX` seems to be needed at line 488

m-vdb (changes requested):

rasa/shared/data.py
m-vdb: why do you specifically need `DeprecationWarning`? What's the requirement from the user perspective?

wochinge: I need a deprecation warning because I want to warn other developers or users using the Python API. This isn't called as part of anything when using the CLI (this would make a `FutureWarning` necessary). **If** anybody decides to use that in their Python code then they'll see that it's deprecated (e.g. Pycharm automatically strikes it through)

setup.cfg
```diff
@@ -22,7 +22,10 @@ max-doc-length = 88
 # See these resources for descriptions of the error codes:
 # E***, F***, W*** : https://flake8.pycqa.org/en/latest/user/error-codes.html
 # D***: http://www.pydocstyle.org/en/stable/error_codes.html
-ignore = W503, E121, E126, E211, E225, E501, E203, E402, F401, F811, E231,
+ignore = W503, # line break before binary operator
+    E203,  # whitespace before ‘:’
+    E231,  # missing whitespace after ‘,’, ‘;’, or ‘:’
```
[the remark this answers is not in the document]
wochinge: it's due to black - adding a comment for it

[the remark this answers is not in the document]
wochinge: it's due to black - adding a comment for it

rasa/cli/export.py
```diff
@@ -12,10 +12,10 @@
 from rasa.shared.constants import DOCS_URL_EVENT_BROKERS, DOCS_URL_TRACKER_STORES
 from rasa.exceptions import PublishingError
 from rasa.shared.exceptions import RasaException
+from rasa.core.brokers.pika import PikaEventBroker
```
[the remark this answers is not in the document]
wochinge: nope double checked. In my opinion we should aim for top level import where possible and just make exceptions for `tensorflow` (which is covered by a test)

rasa/cli/run.py
```diff
@@ -6,7 +6,7 @@
 from rasa.cli import SubParsersAction
 from rasa.cli.arguments import run as arguments
 import rasa.cli.utils
-import rasa.shared.utils.cli
+import rasa.shared.utils.cli  # noqa: F401
```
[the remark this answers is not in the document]
wochinge: I know - seems to be a false positive. Same as in `visualize - haven't found a way around it unfortunately.

m-vdb (approved):
⭐ thanks for the quick turnaround!

setup.cfg
wochinge: https://github.com/RasaHQ/rasa/issues/8490

PR #8444, opened 2021-04-12
Clarified the behavior of the Two-Stage Fallback

hsm207:
**Proposed changes**:
- Clarified behavior of Two-Stage Fallback when the user denies the rephrased intent


docs/docs/fallback-handoff.mdx
```diff
@@ -283,6 +285,44 @@ rules:
   - active_loop: action_two_stage_fallback
 ```
 
+### 4. Defining an ultimate fallback action (optional)
```
melindaloubser1, line 288:
```suggestion
### 4. Defining an ultimate fallback action
```

I would not mark this optional because a silent ending to fallback isn't the intended default behaviour

docs/docs/fallback-handoff.mdx
```diff
@@ -283,6 +285,44 @@ rules:
...
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: Dict[Text, Any],
+    ) -> List[Dict[Text, Any]]:
+
+        dispatcher.utter_message(template="my_custom_fallback_template")
+
+        # Revert user message which led to fallback.
+        return [UserUtteranceReverted()]
+```
```
melindaloubser1, line 324: Definitely good to give an example of overwriting.
To indicate why you'd want to overwrite this, the example could be e.g. handoff to human - not implemented, just a mock call to show that's the intention, since that's a very common final fallback

docs/docs/fallback-handoff.mdx
```diff
@@ -283,6 +285,44 @@ rules:
...
+        self,
+        dispatcher: CollectingDispatcher,
+        tracker: Tracker,
+        domain: Dict[Text, Any],
+    ) -> List[Dict[Text, Any]]:
+
+        dispatcher.utter_message(template="my_custom_fallback_template")
+
+        # Revert user message which led to fallback.
+        return [UserUtteranceReverted()]
```
melindaloubser1, line 323: As I rediscovered today, keeping this `UserUtteranceReverted()` in the return events is crucial to not ending up with repeated fallback actions leading to the bot not responding - I forget what it's called, but when it endlessly predicts the same action.

If you can verify this behaviour, we should document it clearly so people don't overwrite the action and get unexpected behaviour.

melindaloubser1 (changes requested):
Thanks for clarifying! Two suggestions to follow up on.

sara-tagger:
Thanks for opening a draft pull request 🚀If you have any questions, you can direct them to @Ghostvv ✨

docs/docs/fallback-handoff.mdx
```diff
@@ -283,6 +285,49 @@ rules:
...
+        
+        # assume there's a function to call customer service
+        # pass the tracker so that the agent has a record of the converstation between the user
+        # and the bot for context
+        call_customer_service(tracker)
+     
+        # pause the tracker so that the bot stops responding to user input
+        return [ConversationPaused()]
+```
+
```
hsm207, line 330: @melindaloubser1 what do you think of this mock handoff?

melindatestuser: That's great!

docs/docs/fallback-handoff.mdx
melindatestuser: Sorry should have clarified: If you **do not** return `UserUtteranceReverted()` ,  you will see this "catastrophic backtracking"

melindaloubser1: Sorry should have clarified: If you **do not** return `UserUtteranceReverted()` ,  you will see this "catastrophic backtracking"

melindaloubser1: Maybe it was something specific to the way I tried it 🤷‍♀️

docs/docs/fallback-handoff.mdx
```diff
@@ -283,6 +285,57 @@ rules:
...
+        return [ConversationPaused(), UserUtteranceReverted()]
+```
+
+  :::caution Events Returned By A Custom Ultimate Fallback Action
+    It is a good practice to include `UserUtteranceReverted()` as one of the events returned by your custom
+    `action_default_fallback`. Not including this event will cause the tracker to include all events that happened
+    during the Two-Stage Fallback process which could intefere with subsequent action predictions from the bot's policy
+    pipeline. It is better to treat events that occured during the Two-Stage Fallback process as if they did not happen
+    so that your bot can apply its rules or memorised stories to correctly predict the next action. 
+  :::
```
hsm207, line 337: @melindaloubser1 regarding your comment on on `UserUtteranceReverted()`, what do you think of this update?

docs/docs/fallback-handoff.mdx
hsm207: yes, removing the TED Policy was the root cause.

melindaloubser1: Right, we do still need overwriting behaviour to work in the absence of TEDPolicy though, so root-root cause is something more.

docs/docs/fallback-handoff.mdx
melindaloubser1: That sounds good to me
```suggestion
  :::caution Events Returned By A Custom Ultimate Fallback Action
    You should include `UserUtteranceReverted()` as one of the events returned by your custom
    `action_default_fallback`. Not including this event will cause the tracker to include all events that happened
    during the Two-Stage Fallback process which could intefere with subsequent action predictions from the bot's policy
    pipeline. It is better to treat events that occured during the Two-Stage Fallback process as if they did not happen
    so that your bot can apply its rules or memorised stories to correctly predict the next action. 
  :::
```

Sounds good! Just one language update since we have a "kill list" for writing docs  =)

docs/docs/fallback-handoff.mdx
melindaloubser1: I think with your explanation below we can resolve this at any rate

melindaloubser1 (approved):
🚀

Issue #8422, opened 2021-04-09
IntentTEDPolicy: Adapt `rasa test`


wochinge:
**Description of Problem**:
`IntentTEDPolicy` predicts `action_unlikely_intent`  whenever there is an unlikely intent. This can also happen as part of test stories. This is just an indicator that the intent isn't expected but it doesn't necessarily lead to failing predictions. E.g. the Rasa could still predict the correct action after the `action_unlikely_intent` prediction. In this case we should warn the user but don't fail.

**Overview of the Solution**:
There are two paths:
1. Test story contains `action_unlikely_intent` : If Rasa Open Source does not predict this, this leads to a test error (wrongly predicted action) -> Same behavior as the current default for `rasa test`

2. Test story does not contain `action_unlikely_intent` and Rasa Open Source does predict it:
    - the prediction of  `action_unlikely_intent`  will be ignored for the evaluation (and hence not lead to a wrong prediction).
    - prediction of `action_unlikely_intent` will be logged in a file called `stories_with_warnings.yml`
    - `story_report.json` contains the number of stories with warnings

[See more information in the implementation proposal.](https://www.notion.so/rasa/Implementation-Proposal-IntentTedPolicy-f16c3eb785964b1ca1249d508f6dc234#b491aa7f877b4db7a8629af1029d288c)

**Blockers** (if relevant):
- [x] `IntentTED` is implemented and fully functional

**Definition of Done**:
- [x] Missing `action_unlikely_intent` leads to failing `rasa test` if the test story contains `action_unlikely_intent` 
  - [x] in this case the printed `failed_stories.yml` contains `action_unlikely_intent` 
- [x] a prediction of `action_unlikely_intent` leads to the creation of a file `stories_with_warnings.yml` in case the test story didn't contain `action_unlikely_intent`
- [x] `story_report.json` contains the number of stories with warnings

PR #8457, opened 2021-04-13
IntentTEDPolicy: `action_unlikely_intent`

### LINKED ISSUE #8417: IntentTEDPolicy: create `action_unlikely_intent`
**Description of Problem**:
`IntentTEDPolicy` predicts an action `action_unlikely_intent` to indicate that the intent predicted by NLU is unexpected. This action needs to be a default action which means that the user doesn't have to add it their domain by default. In addition the action needs to be able to except the information about the prediction and store it in its metadata.

**Overview of the Solution**:

**Blockers** (if relevant):
None

**Definition of Done**:
- [x] `action_unlikely_intent` is added to the domain by default 
- [x] `IntentTEDPolicy` can pass its prediction metadata via `PolicyPrediction` to the action
- [x] the minimum model compatibility was changed due to the new default action

### LINKED ISSUE #8418: `IntentTEDPolicy`: Ignore `action_unlikely_intent` from story dumping
**Description of Problem**:
`action_unlikely_intent` should not be dumped in the `YAMLStoryWriter` + `MarkdownStoryWriter` when writing stories in Rasa X or `rasa interactive`

**Overview of the Solution**:

**Blockers** (if relevant):
- [ ] https://github.com/RasaHQ/rasa/issues/8417

**Definition of Done**:
- [x] If a `DialogueStateTracker` does contain a `action_unlikely_intent` action, `YAMLStoryWriter` + `MarkdownStoryWriter` does not add it to its output

alwx:
**Proposed changes**:
- [x] `action_unlikely_intent` is added to the domain by default
- [x] add new `action_metadata` field to `PolicyPrediction`.
- [x] `IntentTEDPolicy` will be able to pass its prediction metadata via `PolicyPrediction` to the action
- [x] Change minimum model compatibility as the new default action `action_unlikely_intent` leads to a breaking model change

this PR fixes #8417 and #8418


rasa/shared/core/events.py
```diff
@@ -1475,8 +1475,8 @@ def __init__(
 
     def __repr__(self) -> Text:
         """Returns event as string for debugging."""
-        return "ActionExecuted(action: {}, policy: {}, confidence: {})".format(
-            self.action_name, self.policy, self.confidence
+        return "ActionExecuted(action: {}, policy: {}, confidence: {}, metadata: {})".format(
```
dakshvar22, line 1478: This metadata can be quite large. Is it good to always include it in the representation string?

dakshvar22:
@alwx @wochinge As I understand, the current changes to `YAMLStoryWriter` will never write `action_unlikely_intent`, but we do need to output it for one of the cases [here](https://github.com/RasaHQ/rasa/issues/8422)?

wochinge:
>  Should that be handled separately as part of the respective issue?

In my opinion fine to keep it like that and then adapt it in the other issue 👍🏻 I'd maybe not mention this in the changelog then as we're gonna change that.

Will you have an integration branch or how do you plan to collaborate @dakshvar22 and @alwx ?

rasa/shared/core/training_data/structures.py
```diff
@@ -212,8 +217,17 @@ def is_action_listen(event: Event) -> bool:
         # we don't want to allow subclasses here
         return type(event) == ActionExecuted and event.action_name == ACTION_LISTEN_NAME
 
+    @staticmethod
+    def is_action_unlikely_intent(event: Event) -> bool:
+        """Check if the executed action is a `action_unlikely_intent`."""
```
wochinge, line 222:
```suggestion
        """Checks if the executed action is a `action_unlikely_intent`."""
```

rasa/shared/core/training_data/structures.py
```diff
@@ -212,8 +217,17 @@ def is_action_listen(event: Event) -> bool:
...
+        return (
+            type(event) == ActionExecuted
+            and event.action_name == ACTION_UNLIKELY_INTENT_NAME
+        )
+
     @staticmethod
     def is_action_session_start(event: Event) -> bool:
+        """Check if the executed action is a `action_session_start`."""
```
wochinge, line 230:
```suggestion
        """Checks if the executed action is a `action_session_start`."""
```

tests/core/test_processor.py
```diff
@@ -901,6 +902,33 @@ def predict_action_probabilities(
     )
 
 
+async def test_action_unlikely_intent_metadata(default_processor: MessageProcessor):
```
wochinge, line 905: how about also testing that the metadata from `PolicyPrediction` becomes part of the tracker metadata?

rasa/core/actions/action.py
```diff
@@ -336,6 +339,7 @@ def event_for_successful_execution(
             confidence=prediction.max_confidence,
             action_text=self.action_text,
             hide_rule_turn=prediction.hide_rule_turn,
+            metadata=prediction.action_metadata,
```
ancalita, line 342: How did you know that this `ActionEndtoEndResponse` method needs to be updated? By searching for all usages when an `ActionExecuted` gets returned?

rasa/core/actions/action.py
```diff
@@ -780,7 +784,31 @@ async def run(
...
+
+    async def run(
+        self,
+        output_channel: "OutputChannel",
+        nlg: "NaturalLanguageGenerator",
+        tracker: "DialogueStateTracker",
+        domain: "Domain",
+    ) -> List[Event]:
+        """Runs action. Please see parent class for the full docstring."""
+        evts = await super().run(output_channel, nlg, tracker, domain)
```
ancalita, line 805: I'm confused by the use of `super()` here: doesn't this call the parent call's run method which raises `NotImplementedError`?

tests/core/test_actions.py
```diff
@@ -115,11 +116,12 @@ def test_domain_action_instantiation():
     assert instantiated_actions[6].name() == ACTION_DEFAULT_ASK_AFFIRMATION_NAME
     assert instantiated_actions[7].name() == ACTION_DEFAULT_ASK_REPHRASE_NAME
     assert instantiated_actions[8].name() == ACTION_TWO_STAGE_FALLBACK_NAME
-    assert instantiated_actions[9].name() == ACTION_BACK_NAME
-    assert instantiated_actions[10].name() == RULE_SNIPPET_ACTION_NAME
-    assert instantiated_actions[11].name() == "my_module.ActionTest"
-    assert instantiated_actions[12].name() == "utter_test"
-    assert instantiated_actions[13].name() == "utter_chitchat"
+    assert instantiated_actions[9].name() == ACTION_UNLIKELY_INTENT_NAME
```
ancalita, line 119: How did you know this action would be at index 9? 🤔

tests/core/test_processor.py
alwx: I've just updated this test and added one more explicit check if metadata is correct. Let me know what you think!

ancalita (commented):
Not sure if I missed anything, but I couldn't find this step from the implementation proposal in the PR: `add the IntentTEDPolicy to the default model configuration which is auto suggested to users`  + as well as the `rasa test` [notes](https://www.notion.so/f16c3eb785964b1ca1249d508f6dc234#6e766b58727b4d5c95050024b7fdac42)?

tests/core/test_actions.py
alwx: `ACTION_UNLIKELY_INTENT_NAME` is a default action which means it's added to `DEFAULT_ACTION_NAMES` explicitly. We know its position (the index is 9) which means this test simply checks if all the default actions were added to the list of instantiated actions.

See `DEFAULT_ACTION_NAMES` in `rasa/shared/core/constants.py` and `_combine_user_with_default_actions` in `rasa/shared/core/domain.py` for more details.

alwx:
@ancalita `IntentTEDPolicy` itself is not implemented yet — it's in progress though. There is also a separate issue for `rasa test`: https://github.com/RasaHQ/rasa/issues/8422

rasa/core/actions/action.py
alwx: Yes, basically. We need to make sure that `ActionExecuted`'s metadata is updated with `PolicyPrediction`'s metadata — which means we need to check where the constructor for `ActionExecuted` is being called and update it accordingly.

rasa/core/actions/action.py
alwx: In fact, you're correct here — I've just updated it to return `[]`

ancalita (approved):
All good with me 😃

rasa/core/policies/policy.py
```diff
@@ -503,6 +504,8 @@ def __init__(
                 fine-tuning purposes.
             hide_rule_turn: `True` if the prediction was made by the rules which
                 do not appear in the stories
+            action_metadata: Specifies an additional metadata that can be passed
```
wochinge, line 507: optional
```suggestion
            action_metadata: Specifies additional metadata that can be passed
```

rasa/shared/core/training_data/structures.py
```diff
@@ -212,8 +217,17 @@ def is_action_listen(event: Event) -> bool:
         # we don't want to allow subclasses here
         return type(event) == ActionExecuted and event.action_name == ACTION_LISTEN_NAME
 
+    @staticmethod
+    def is_action_unlikely_intent(event: Event) -> bool:
+        """Checks if the executed action is a `action_unlikely_intent`."""
+        return (
+            type(event) == ActionExecuted
```
wochinge, line 224: optional
```suggestion
            isinstance(event, ActionExecuted)
```

tests/core/test_processor.py
wochinge: I think the test is great! It would be slightly nicer to test via the public processor interface (e.g. `execute` or `handle_message`) instead of via protectecd functions.

wochinge (approved):
💯

rasa/shared/core/events.py
```diff
@@ -1473,6 +1473,14 @@ def __init__(
 
         super().__init__(timestamp, metadata)
 
+    def __members(self) -> Tuple[Optional[Text], Optional[Text], Text]:
```
dakshvar22, line 1476: Just curious: Did you mean to do `__members__` instead of `__members`?
Ref: https://docs.python.org/3/library/enum.html#iteration

dakshvar22 (approved):
Looks good to me 💯

rasa/shared/core/events.py
alwx: 👍

PR #8458, opened 2021-04-13
Don't pin version for dm-tree

tczekajlo:
**Proposed changes**:
- Don't pin version for dm-tree

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -18,8 +18,8 @@ optional = false
 python-versions = ">3.5.*, <4"
 
 [package.dependencies]
-aiormq = ">=3.2.3,<4"
 yarl = "*"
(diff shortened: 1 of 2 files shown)


wochinge:
did we fix the missing bazel dep in the docker builds?

tczekajlo:
> did we fix the missing bazel dep in the docker builds?

No, and I can't reproduce the error that we ran into yesterday. In the logs from yesterday, we can see that it was trying to install dm-tree from sources, so it looks like wheels for the package weren't available yet, now everything looks good and it installs wheels instead of building from the source.

https://pypi.org/project/dm-tree/#files

wochinge (approved):

PR #8466, opened 2021-04-14
Turn Docs example into valid json

koaning:
I couldn't copy/paste the example because it didn't list valid json. There's a comma too many. So I made a quick fix. 

The example can be found [here](https://rasa.com/docs/rasa/connectors/your-own-website#request-and-response-format). 

This isn't valid json. 

```json
{
  "sender": "test_user",
  "message": "Hi there!",
}
```

This is. 

```json
{
  "sender": "test_user",
  "message": "Hi there!"
}
```


erohmensing (approved):
Thanks for the catch!

PR #8467, opened 2021-04-14
Best model checkpoint allows metrics to be equal to previous best if at least one metric improves

### LINKED ISSUE #7286: checkpoint_model ignores best model if just one metric stays the same
This issue concerns the checkpoint_model parameter in every trainable component (DietClassifier, ResponseSelector, RulePolicy, etc...)
In order to save a model that has improved in every metric except one that has staled (can even stale at 1.00 accuracy !), the _does_model_improve function could compare with a superior or equal sign rather than superior strict.

https://github.com/RasaHQ/rasa/blob/0ccfa20ef72f7f6d4e4e1f3e0bf88d8568bfcb7e/rasa/utils/tensorflow/models.py#L483
```
all_improved = all(
            [
                float(current_results[key]) >= self.best_metrics_so_far[key]
                for key in self.best_metrics_so_far.keys()
            ]
        )
```

koernerfelicia:
closes #7286 

**Proposed changes**:
- Allow for equal metrics if at least one metric improves, do not allow for any metric to decrease.
- Also add a test to warn users if they opt for best model checkpoint but do not allow for any examples to be used for validation and/or specify that validation should not occur, or only occur at the very end of training.


rasa/utils/tensorflow/callback.py
```diff
@@ -86,15 +86,20 @@ def _does_model_improve(self, current_results: Dict[Text, Any]) -> bool:
...
-                for key in self.best_metrics_so_far.keys()
-            ]
-        )
-
-        if all_improved:
+        at_least_one_improved = False
+        for key in self.best_metrics_so_far.keys():
+            if key not in current_results:
+                # current_results.keys is not a subset of best_metric_so_far.keys
+                return False
```
samsucik, line 93: Can this happen?

rasa/utils/tensorflow/callback.py
```diff
@@ -86,15 +86,20 @@ def _does_model_improve(self, current_results: Dict[Text, Any]) -> bool:
...
 
-        all_improved = all(
-            [
-                float(current_results[key]) > self.best_metrics_so_far[key]
-                for key in self.best_metrics_so_far.keys()
-            ]
-        )
-
-        if all_improved:
+        at_least_one_improved = False
```
samsucik, line 89: Since you've touched this method, could you add a docstring to it as well? (Sorry, I can't attach a comment to a line that you didn't change, hence commenting here.)

rasa/utils/tensorflow/callback.py
```diff
@@ -86,15 +86,20 @@ def _does_model_improve(self, current_results: Dict[Text, Any]) -> bool:
...
+        for key in self.best_metrics_so_far.keys():
```
samsucik, line 90: Nitpicking, but renaming `key` to `metric_name` would help me (and perhaps others) read the code quicker 🙂

rasa/utils/train_utils.py
```diff
@@ -503,6 +504,18 @@ def validate_configuration_settings(component_config: Dict[Text, Any]) -> None:
...
+
+
+def _check_checkpoint_setting(component_config: Dict[Text, Any]) -> None:
+    if component_config[CHECKPOINT_MODEL] and (
+        component_config[EVAL_NUM_EXAMPLES] <= 0
+        or component_config[EVAL_NUM_EPOCHS] <= 0
+    ):
+        rasa.shared.utils.io.raise_warning(
+            f"You have opted to save the best model, but one or both of"
+            f" {EVAL_NUM_EXAMPLES} or {EVAL_NUM_EPOCHS} is not greater than 0."
```
samsucik, line 517: Since this is not an exception but a warning, we somehow deal with the weird config (presumably by not saving checkpoints?) and move on. Could you clarify this in the warning as well?

tests/utils/tensorflow/test_callback.py
```diff
@@ -0,0 +1,71 @@
...
+            [{"val_i_acc": 0.64, "val_f1": 0.5}, {"val_i_acc": 0.41, "val_f1": 0.7}],
+            False,
+        ),  # one improved, one worse
+        (
+            [{"val_i_acc": 0.71, "val_f1": 0.35}, {"val_i_acc": 0.52, "val_f1": 0.35}],
+            False,
+        ),  # one worse, one equal
+    ],
+)
+def test_does_model_improve(logs: List[Dict[Text, float]], improved, tmpdir: Path):
```
samsucik, line 36:
```suggestion
def test_does_model_improve(logs: List[Dict[Text, float]], improved: bool, tmpdir: Path) -> None:
```

tests/utils/tensorflow/test_callback.py
```diff
@@ -0,0 +1,71 @@
...
+
+import pytest
+
+from rasa.core.agent import Agent
+from rasa.core.policies.ted_policy import TEDPolicy
+from rasa.utils.tensorflow.callback import RasaModelCheckpoint
+
+
+@pytest.mark.parametrize(
+    "logs, improved",
```
samsucik, line 12: To improve clarity, you could split `logs` into `previous_values` and `new_values`.

tests/utils/tensorflow/test_callback.py
```diff
@@ -0,0 +1,71 @@
...
+    [
+        ([{"val_i_acc": 0.5, "val_f1": 0.5}, {"val_i_acc": 0.5, "val_f1": 0.7}], True),
+        ([{"val_i_acc": 0.5, "val_f1": 0.5}, {"val_i_acc": 0.4, "val_f1": 0.5}], False),
+    ],
+)
+def test_on_epoch_end_saves_checkpoints_file(
+    logs: List[Dict[Text, float]], improved: bool, trained_ted: TEDPolicy, tmpdir: Path
+):
```
samsucik, line 60:
```suggestion
) -> None:
```

tests/utils/tensorflow/test_callback.py
```diff
@@ -0,0 +1,71 @@
...
+    model_name = "checkpoint"
+    best_model_file = Path(str(tmpdir), model_name)
```
samsucik, line 62: I don't think the `str` is needed. I keep seeing it elsewhere which might mean that no-one has questioned it or that it's actually needed 😅

tests/utils/tensorflow/test_callback.py
```diff
@@ -0,0 +1,71 @@
...
+@pytest.fixture(scope="function")
+def trained_ted(mood_agent: Agent) -> TEDPolicy:
+    # use the moodbot agent to get a trained TEDPolicy
+    for policy in mood_agent.policy_ensemble.policies:
+        if isinstance(policy, TEDPolicy):
+            return policy
+
+
+@pytest.mark.parametrize(
+    "logs, improved",
```
samsucik, line 52: Here as well, could you split `logs`?

tests/utils/test_train_utils.py
```diff
@@ -151,3 +154,34 @@ def test_update_confidence_type(
...
+
+@pytest.mark.parametrize(
+    "component_config",
+    [
+        ({CHECKPOINT_MODEL: True, EVAL_NUM_EPOCHS: -1, EVAL_NUM_EXAMPLES: 0}),
+        ({CHECKPOINT_MODEL: True, EVAL_NUM_EPOCHS: -1, EVAL_NUM_EXAMPLES: 10}),
+        ({CHECKPOINT_MODEL: True, EVAL_NUM_EPOCHS: 10, EVAL_NUM_EXAMPLES: 0}),
+    ],
+)
+def test_warning_incorrect_checkpoint_setting(component_config: Dict[Text, Text]):
```
samsucik, line 167:
```suggestion
def test_warning_incorrect_checkpoint_setting(component_config: Dict[Text, Text]) -> None:
```

samsucik, line 167: Given that we check the warning, does it make sense to also check that we don't actually save checkpoints in this case?

tests/utils/test_train_utils.py
```diff
@@ -151,3 +154,34 @@ def test_update_confidence_type(
...
+
+
+@pytest.mark.parametrize(
+    "component_config",
+    [
+        ({CHECKPOINT_MODEL: False, EVAL_NUM_EPOCHS: -1, EVAL_NUM_EXAMPLES: 0}),
+        ({CHECKPOINT_MODEL: True, EVAL_NUM_EPOCHS: 10, EVAL_NUM_EXAMPLES: 10}),
+    ],
+)
+def test_no_warning_correct_checkpoint_setting(component_config: Dict[Text, Text]):
```
samsucik, line 184:
```suggestion
def test_no_warning_correct_checkpoint_setting(component_config: Dict[Text, Text]) -> None:
```

samsucik (changes requested):
Thanks, Felicia! The core looks good to me; I've added a bunch of small comments about some details.

rasa/utils/tensorflow/callback.py
koernerfelicia: Yeah, it happens all the time actually, but due to a weird design decision (not made by me), so I wonder if I should refactor it to be cleaner/clearer. Here's what's happening:
This method is called at the end of **every** epoch, but metrics are only prefixed with `val` if it's an epoch where we've specified validation (so if its `epoch_number%evaluate_every_number_of_epoch==0`). So basically we fail this condition and nothing happens.

koernerfelicia: I'll refactor it, you asking this question is a good push :)

samsucik: Oh, that was a rather creative design decision, I must say 😆  Thanks a lot for taking care of it!

rasa/utils/tensorflow/callback.py
koernerfelicia: Yep!

rasa/utils/tensorflow/callback.py
```diff
@@ -72,29 +72,49 @@ def on_epoch_end(self, epoch: int, logs: Optional[Dict[Text, Any]] = None) -> No
...
+        Args:
+            current_results: The training metrics for this epoch.
+        """
+        curr_metric_names = [
+            k
+            for k in current_results.keys()
+            if k.startswith("val") and (k.endswith("_acc") or k.endswith("_f1"))
+        ]
+        # the "val" prefix is prepended to metrics in fit if _should_eval returns true
+        # for this particular epoch
```
koernerfelicia, line 89: @samsucik let me know if this is clear enough? Not sure...

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -613,6 +613,80 @@ async def test_train_model_checkpointing(
...
+    _config = RasaNLUModelConfig(
+        {
+            "pipeline": [
+                {"name": "WhitespaceTokenizer"},
+                {"name": "CountVectorsFeaturizer"},
+                {
+                    "name": "DIETClassifier",
+                    EPOCHS: 2,
+                    CONSTRAIN_SIMILARITIES: True,
+                    MODEL_CONFIDENCE: "linear_norm",
```
samsucik, line 669: Same here.

samsucik (approved):
Looks good to me, thanks @koernerfelicia!

rasa/utils/train_utils.py
```diff
@@ -531,6 +533,21 @@ def validate_configuration_settings(component_config: Dict[Text, Any]) -> None:
...
+        if component_config[EVAL_NUM_EPOCHS] <= 0:
+            rasa.shared.utils.io.raise_warning(
+                f"You have opted to save the best model, {EVAL_NUM_EPOCHS} is not greater "
+                f"than 0, training will fail."
+            )
+        if component_config[EVAL_NUM_EXAMPLES] <= 0:
+            rasa.shared.utils.io.raise_warning(
+                f"You have opted to save the best model, {EVAL_NUM_EXAMPLES} is not greater "
+                f"than 0. No model will be saved."
+            )
```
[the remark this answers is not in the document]
koernerfelicia: Eek, actually just realised that `EVAL_NUM_EPOCHS` is allowed to be `-1`... Sorry for this confusion, will address that.

Not sure what you mean by enable/disable validation. Really the correct way to disable is to set `checkpoint_model` to false. Should I add information about the consequences of incorrect configurations?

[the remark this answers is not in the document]
samsucik: > Not sure what you mean by enable/disable validation. Really the correct way to disable is to set checkpoint_model to false.

What I mean is enabling/disabling model validation. Let's consider this comment that accompanies `EVAL_NUM_EPOCHS` in TED: `How often calculate validation accuracy. Small values may hurt performance, e.g. model accuracy.` For now ignoring the part about hurting model accuracy (which I kinda don't get), I might want to disable validation altogether if I want to train faster. And I might be completely unaware of model checkpointing. In such case, it'd help me to know how to control validation as such.

With `EVAL_NUM_EPOCHS` being possible to set to -1, this becomes easier indeed, but I still have to figure out this magical special value of `-1`. I think that a simple comment 1) in all the models where `EVAL_NUM_EPOCHS` is listed as a parameter and 2) in the docs (in a visible way) would help here. I'm unsure if it's still within the scope of this issue though.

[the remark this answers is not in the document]
koernerfelicia: Ohh I see, you mean enable/disable validation separately from enable/disable saving the model. Sorry, I've been (incorrectly) conflating the two.

I actually don't think this can be done with the `EVAL_NUM_EPOCHS` param, because even `EVAL_NUM_EPOCHS=-1` will still validate at the very end of training. You'd have to set `EVAL_NUM_EXAMPLES=0` for no validation. I've adjusted the comments in the code and tried to clarify in the docs.

tests/utils/test_train_utils.py
koernerfelicia: I *think* this should now be covered by the DIET and TED tests, both of which check that nothing is saved when there are incorrect settings, and also check that the warnings are produced. This may be overkill.. or more dependency on the train_utils method than desired. What do you think?

samsucik: Agree with you.

rasa/utils/train_utils.py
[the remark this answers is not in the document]
samsucik: Makes sense. Thanks!

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -564,26 +579,20 @@ async def test_train_model_checkpointing(
...
-                    "analyzer": "char_wb",
-                    "min_ngram": 3,
-                    "max_ngram": 17,
-                    "max_features": 10,
-                    "min_df": 5,
-                },
+                {"name": "CountVectorsFeaturizer"},
                 {
                     "name": "DIETClassifier",
                     EPOCHS: 5,
```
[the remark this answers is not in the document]
koernerfelicia: Yes, that's a good point. I made it 1 for ones where we expect failure, and 2 for all the others now for consistency.

koernerfelicia:
@samsucik sorry, this is the PR that never ends. Just occurred to me I tagged this as a bug in the changelog bc it was tagged as such in the issue, but now that I think about it, not sure if this is *actually* a bug or really an improvement. If the former, I'd retarget 2.7.x, if the latter I'd change the changelog. What do you think?

samsucik (approved):
I think this is good to go. I get that it's somewhat of an improvement, but I'm gonna argue for the bugfix label as it's addressing an issue in the logic which, I believe, wasn't intended (we didn't want to restrict best model saving only to the cases when all metrics improve).

data/test_config/config_ted_policy_model_checkpointing.yml
```diff
@@ -1,13 +1,7 @@
...
   - name: TEDPolicy
-    max_history: 5
-    epochs: 5
+    epochs: 2
     evaluate_on_number_of_examples: 14
     evaluate_every_number_of_epochs: 1
-    checkpoint_model: True
-  - name: RulePolicy
\ No newline at end of file
+    checkpoint_model: True
```
wochinge, line 7: can we please add empty lines at the end of all these test files due to [this](https://stackoverflow.com/a/2287990/3429596)?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -609,10 +618,122 @@ async def test_train_model_checkpointing(
...
+        data=nlu_data_path,
+        component_builder=component_builder,
+        fixed_model_name=model_name,
+    )
+
+    assert not checkpoint_dir.is_dir()
+
+
+async def test_train_fails_with_zero_eval_num_epochs(
+    component_builder: ComponentBuilder, tmpdir: Path, nlu_data_path: Text,
```
wochinge, line 655: can we please use the `tmp_path` fixture instead of `tmpdir`? `tmpdir` doesn't truly have type `Path`

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -609,10 +618,122 @@ async def test_train_model_checkpointing(
...
+):
+    model_name = "nlu-fail"
+    checkpoint_dir = get_checkpoint_dir_path(tmpdir, Path(tmpdir, model_name))
```
wochinge, line 658: optional: slightly slicker
```suggestion
    checkpoint_dir = get_checkpoint_dir_path(tmpdir, tmpdir / model_name)
```

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -609,10 +618,122 @@ async def test_train_model_checkpointing(
...
+                component_builder=component_builder,
+                fixed_model_name=model_name,
+            )
+    assert not checkpoint_dir.is_dir()
+    warn_text = (
+        f"You have opted to save the best model, but the value of '{EVAL_NUM_EPOCHS}' "
+        f"is not -1 or greater than 0. Training will fail."
+    )
+    print(f"\n{warn_text}")
+    print(warning[2].message)
```
wochinge, line 692:
```suggestion
```

koernerfelicia:
@wochinge I actually had to make a new PR because this one had targeted the wrong branch. I'm sorry for the confusion, I should have closed this one. The new PR (targeting `2.7.x`) was merged in yesterday. I would like to apply your comments, can I just make a new PR from `2.7.x` with just these comments and tag you as a reviewer?

Also, for future reference, enable was automatically tagged as a reviewer when I changed the base branch. Should I have added enable as a reviewer myself when I made the new PR? I thought these files were mostly our area (ted, diet, callback, train_utils..)

wochinge:
> can I just make a new PR from 2.7.x with just these comments and tag you as a reviewer?

Yes please

Mhm, interesting. Apparently `CODEOWNERS` only protects `main` at the moment 🤔

tests/nlu/classifiers/test_diet_classifier.py
koernerfelicia: 🤦‍♀️

koernerfelicia:
@wochinge oh that would probably be the wonkiness arising from switching the base branch... Sorry for this mess again, I promise I have learned for the future.

tests/nlu/classifiers/test_diet_classifier.py
samsucik: Wow, I can't believe I overlooked this 🤦

wochinge:
no worries 😄 Having to switch the base branch happens all the time to me (GitHub is sometimes quite weird about it, too) 😆

PR #8473, opened 2021-04-15
add log time rotating handle

XiaofeiQian:
**Proposed changes**:
- ...

The log of `rasa run` will become larger and larger with the increase of time, and it is necessary to divide it in actual production. The custom component part can be implemented by your own code, but the log of the core part is always written in a file by default. This change only adds the option of converting by time.

### DIFF
--- a/rasa/cli/arguments/run.py
+++ b/rasa/cli/arguments/run.py
@@ -42,6 +42,24 @@ def add_server_arguments(parser: argparse.ArgumentParser) -> None:
         default=None,
         help="Store logs in specified file.",
     )
+    parser.add_argument(
+        "--log-rotating",
+        type=str,
+        # Rasa should not rotaing log file by default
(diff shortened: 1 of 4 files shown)


sara-tagger:

rasa/cli/arguments/run.py
```diff
@@ -42,6 +42,24 @@ def add_server_arguments(parser: argparse.ArgumentParser) -> None:
         default=None,
         help="Store logs in specified file.",
     )
+    parser.add_argument(
+        "--log-rotating",
+        type=str,
+        # Rasa should not rotaing log file by default
```
joejuzl, line 48:
```suggestion
        # Rasa should not rotate log files by default
```

joejuzl (changes requested):
Thanks for this!

PR #8511, opened 2021-04-20
Improve vulnerability scan workflow

### LINKED ISSUE #8474: Vulnerability GitHub issues created by trivy are created in duplicate
See issues here: https://github.com/RasaHQ/rasa/issues?q=is%3Aissue+author%3Aapp%2Fgithub-actions+label%3Atool%3Atrivy+is%3Aclosed

Additionally, it'd be nice to add them automatically to Enable's project board

m-vdb:
**Proposed changes**:
- Fixes #8474 
- Tested with this repository: https://github.com/m-vdb/test-repo-project
- I forked the action to support 2 cases:
   - If a user creates an issue and assigns it to a project during the issue creation, the workflow was failing because the issue is "awaiting triage" (more of an edge case)
   - Issues already in a project should be left untouched (more of safety on issue creation to prevent race conditions, but it's important if we want to support the use case of adding/removing labels)
- After merge of this PR, remove the `RASABOT_ASSIGN_ISSUE_RESEARCH_BOARD_TOKEN` value from this repository's secrets (it has been renamed to a more generic name)

### DIFF
--- a/.github/workflows/assign-default-project.yml
+++ b/.github/workflows/assign-default-project.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - name: Assign vulnerability scans or Sentry issues to Enable
       if: contains(github.event.issue.labels.*.name, 'tool:trivy') || contains(github.event.issue.labels.*.name, 'tool:sentry')
-      uses: m-vdb/create-or-update-project-card@9c632af4932aabf8115780d762841f3ca07f7b6b
+      uses: RasaHQ/create-or-update-project-card@9c632af4932aabf8115780d762841f3ca07f7b6b
 
       with:
         token: ${{ secrets.RASABOT_ASSIGN_ISSUES_BOARD_TOKEN }}
@@ -24,7 +24,7 @@ jobs:
 
     - name: Assign model regression issues to Research
       if: contains(github.event.issue.labels.*.name, 'tool:model-regression-tests')
-      uses: m-vdb/create-or-update-project-card@9c632af4932aabf8115780d762841f3ca07f7b6b
+      uses: RasaHQ/create-or-update-project-card@9c632af4932aabf8115780d762841f3ca07f7b6b
 
       with:
         token: ${{ secrets.RASABOT_ASSIGN_ISSUES_BOARD_TOKEN }}
--- a/.github/workflows/assign-default-project.yml
+++ b/.github/workflows/assign-default-project.yml
@@ -22,6 +22,17 @@ jobs:
         project-number: 40
         column-name: Inbox
 
+    - name: Assign model regression issues to Research
+      if: contains(github.event.issue.labels.*.name, 'tool:model-regression-tests')
+      uses: m-vdb/create-or-update-project-card@9c632af4932aabf8115780d762841f3ca07f7b6b
+
+      with:
+        token: ${{ secrets.RASABOT_ASSIGN_ISSUES_BOARD_TOKEN }}
+        # https://github.com/orgs/RasaHQ/projects/21
+        project-location: RasaHQ
+        project-number: 21
+        column-name: Inbox
+
     - name: Notify slack on failure
       if: failure()
       env:
--- a/.github/workflows/assign-default-project.yml
+++ b/.github/workflows/assign-default-project.yml
@@ -16,7 +16,7 @@ jobs:
       uses: peter-evans/create-or-update-project-card@3d5056faa1b952f322b49b5d5f37adefd702e3e1
 
       with:
-        token: ${{ secrets.RASABOT_ASSIGN_ISSUE_RESEARCH_BOARD_TOKEN }}
+        token: ${{ secrets.RASABOT_ASSIGN_ISSUES_BOARD_TOKEN }}
         # https://github.com/orgs/RasaHQ/projects/40
         project-location: RasaHQ
         project-number: 40
--- a/.github/workflows/assign-default-project.yml
+++ b/.github/workflows/assign-default-project.yml
@@ -6,12 +6,12 @@ on:
     - opened
 
 jobs:
-  vulnerability_scan:
-    name: Assign relevant issues to Enable project board
+  assign_default_project:
+    name: Assign default project to issue
     runs-on: ubuntu-latest
 
     steps:
-    - name: Add issue to project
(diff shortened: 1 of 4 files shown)


.github/workflows/assign-default-project.yml
```diff
@@ -0,0 +1,23 @@
...
+    - name: Add issue to project
+      if: contains(github.event.issue.labels.*.name, 'tool:trivy') || contains(github.event.issue.labels.*.name, 'tool:sentry')
+      uses: peter-evans/create-or-update-project-card@3d5056faa1b952f322b49b5d5f37adefd702e3e1
+
+      with:
+        token: ${{ secrets.RASABOT_ASSIGN_ISSUE_RESEARCH_BOARD_TOKEN }}
+        # https://github.com/orgs/RasaHQ/projects/40
+        project-location: RasaHQ
+        project-number: 40
+        column-name: Inbox
```
m-vdb, line 23: TODO: add message to Slack channel if failing

.github/workflows/assign-default-project.yml
```diff
@@ -0,0 +1,23 @@
+name: Assign default project to issues
+
+on:
+  issues:
+    types:
+    - opened
```
m-vdb, line 6: we could also support more use cases when a label is added to an existing issue, not sure it's needed at this stage. If we want to support it, `peter-evans/create-or-update-project-card` might not be the best choice for a GH action because there is no way to control the "update" behaviour (might fork it to make it controllable?)

.github/workflows/assign-default-project.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: Assign relevant issues to Enable project board
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Add issue to project
+      if: contains(github.event.issue.labels.*.name, 'tool:trivy') || contains(github.event.issue.labels.*.name, 'tool:sentry')
+      uses: peter-evans/create-or-update-project-card@3d5056faa1b952f322b49b5d5f37adefd702e3e1
+
+      with:
+        token: ${{ secrets.RASABOT_ASSIGN_ISSUE_RESEARCH_BOARD_TOKEN }}
```
m-vdb, line 19: pending renaming

.github/workflows/ci-model-regression-on-schedule.yml
```diff
@@ -419,7 +419,7 @@ jobs:
           # Add this issue to a project column
           curl \
             -X POST \
-            -s -H "Authorization: token ${{ secrets.RASABOT_ASSIGN_ISSUE_RESEARCH_BOARD_TOKEN }}" \
+            -s -H "Authorization: token ${{ secrets.RASABOT_ASSIGN_ISSUES_BOARD_TOKEN }}" \
```
m-vdb, line 422: We could possibly move the assignment of the issue to the new `assign-default-project` workflow, it may make things more modular

.github/workflows/assign-default-project.yml
```diff
@@ -0,0 +1,33 @@
...
+        # https://github.com/orgs/RasaHQ/projects/40
+        project-location: RasaHQ
+        project-number: 40
+        column-name: Inbox
+
+    - name: Notify slack on failure
+      if: failure()
+      env:
+        SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+      uses: voxmedia/github-action-slack-notify-build@v1
```
wochinge, line 29: optional: there is also a v2 (but also a newer 1.5 release 🙈 )

.github/workflows/ci-model-regression-on-schedule.yml
wochinge: up to you in my opinion. I think both has it's advantages. Your approach is more DRY while the current approach keeps things together which makes it easier to understand a workflow without understand all workflows). Maybe it's worth adding a comment where the issue to project mapping is done in this workflow in case you decide to move it out?

wochinge (approved):
Thank you 💯

.github/workflows/ci-model-regression-on-schedule.yml
m-vdb: yeah I agree with you, both ways of doing it are acceptable. I wanted get @tczekajlo opinion on this before moving it (or not).

I was thinking that we'll likely have other rules in the future to automate project assignment, and keeping all of these rules in one place might make things more maintainable 🤔

.github/workflows/assign-default-project.yml
m-vdb: I pined the latest commit c227368e592

.github/workflows/ci-model-regression-on-schedule.yml
tczekajlo: The new `assign-default-project` workflow looks good to me 😄  The only thing that has to be changed before moving it is the label that is used for the issues created for the research team, now it uses the "bug" label that is quite common, I guess we could add an additional label like 'model-regression-tests:performance-drop` (?) in order to distinguish the issues

m-vdb: or something like `tool:model-regression-tests`? That'd make it consistent with other tools we use elsewhere (even if this one is in-house ^^)

m-vdb: @tczekajlo I've made the change in this commit 1e114bad245 and also added the label to existing issues. Can I get a review? 🤩

tczekajlo (approved):
Beautiful 🎖️

.github/workflows/assign-default-project.yml
```diff
@@ -0,0 +1,44 @@
...
+      with:
+        token: ${{ secrets.RASABOT_ASSIGN_ISSUES_BOARD_TOKEN }}
+        # https://github.com/orgs/RasaHQ/projects/40
+        project-location: RasaHQ
+        project-number: 40
+        column-name: Inbox
+
+    - name: Assign model regression issues to Research
+      if: contains(github.event.issue.labels.*.name, 'tool:model-regression-tests')
+      uses: m-vdb/create-or-update-project-card@9c632af4932aabf8115780d762841f3ca07f7b6b
```
tczekajlo, line 27: Can we fork it to the Rasa org?

PR #8512, opened 2021-04-20
Added import to the example code

bastbnl:
**Proposed changes**:
- Added import of missing constant to the examples concerning fallback handoff 


bastbnl:
Bit in doubt here, since you might want to run the actions in an environment that doesn't contain the full installation, in which case returning the string that I removed in the second commit makes more sense

sara-tagger:
Thanks for submitting a pull request 🚀 @rctatman will take a look at it as soon as possible ✨

ancalita (commented):
Thanks for your contribution! 💯 
We also need a doc changelog entry where you specify the change you've made.

Issue #8407, opened 2021-04-09
Architecture prototype: Explore partial dataset loading for NLU graph


wochinge:
**Description of Problem**:
Research already did a [PR](https://github.com/RasaHQ/rasa/pull/8306) for training the NLU model in chunks and any new model architecture in the future will have to be able to support this use case to enable training large scale models on huge datasets. **This issue is limited to the NLU part of partial training data set loading.** Partial training data set loading / processing doesn't need to be explored.

**Overview of the Solution**:
There is no need for implementation here. However, we need to have clear picture of how an implementation with the proposed graph architecture could look like.

**Blockers** (if relevant):
None

**Definition of Done**:
- [x] Specify in a document how much control the `Component`s need over the chunked training data (e.g. whether the components need intermedite write operations)
- [x] Specify the requirements which `Component`s have to be able to train on chunks (e.g. what kind of summary data they need before - these things are currently done in `prepare_partial_training`)
- [x] Have an answer whether it's possible to abstract the current training (`chunks==1`) as a subset of training with multiple chunks
- [x] propose an interface which abstracts the passed training data (it shouldn't matter to the `Component` if the data is on disk or in memory) and which integrates into the graph structure

twerkmeister:
#### What happens currently in prepare partial training
* general case: nothing
* `RegexEntityExtractor`
    * collects patterns from `TrainingData` object
        * requires: lookup tables and defined regexes, (also knowledge about which regex types are also entity annotations - although this might be solved also in other ways)
* `DIETClassifier`
    * sorts intents by name, and creates mapping {intent: id} and {id: intent} if intent classification is enabled
        * requires: all intents
    * creates mappings {entity-type: id}, {entity-role: id}, {entity-group: id}. If bilou is enabled, expand mappings for each bilou prefix
        * requires: all entity types, -roles, and -groups
    * checks all entity annotations of all messages and logs warnings
        * requires: all messages tokenized and entity annotations
* `EntitySynonymMapper`
    * collects synonyms from `TrainingData` object
        * requires: synonyms
* `CountVectorsFeaturizer`
    * trains the component and creates a vocabulary plus some buffer for fine-tuning
        * requires: message text, response text, action text (+ intent, action name, intent response key if trained on word level)
    * Actually the situation seems to be the following
        * `prepare_partial_training` is the actual training/creation of the vectorizers
        * `_train_on_examples` is the processing of the messages and adding the vectorizer output as additional info to the messages
[...]

PR #8518, opened 2021-04-21
proof of concept for streaming messages in nlu training

twerkmeister:
**DRAFT FOR DISCUSSION PURPOSES **:
- proof of concept for #8407 
- The implementation is pretty rough and just to theoretically test some of the bigger risks (such as, could diet be trained from a stream of messages)
- implemented for CountVectorsFeaturizer, Spacy Featurizer, CRF, and DietClassifier
- main training data object (with unfeaturized messages) stays in memory right now, as so many parts of the codebase use it that way
- features can be successively built and stored on disc (right now just pickling a list of features per message, ultimately a safer and more sophisticated structure should be chosen)
- messages can be joined with features on the fly and consumed as a Generator. For the matching I am using a fingerprint of the message that is just based on the following attributes of the message: text, intent, response, action name, action text, and intent response key. I think this way the fingerprint would be invariant across processing steps and across runs.
- keras.Sequence can be used to hide the generator, but there are probably better abstractions available. Possibly also some multiprocessing here to fill a batch queue in a second thread or so.
- Chunking directories are just based on the component name right now, and if some component requests a message stream, the latest chunking dir will be used
- tested with memory profiling on multiwoz, for details see below
- shuffling is implemented by: using smallish chunks and randomizing chunk reading order + reading multiple chunks at once and then shuffling the acumulated again

**notes**
- I think the methods of many components could really benefit from taking a more message centric perspective. So many of components' methods operate on collections of messages. That makes the functions more complex, harder to understand, and harder to reuse. Was able to throw out a lot of code in the `CountVectorFeaturizer` by focusing more on messages
- `RasaModelData` seems a bit like a parallel world to the `TrainingData` class. I think there are many parallel implementations of the same logic on these classes, such as splitting training data according to some labels. I am not sure this intermediate `RasaModelData` class is really necessary. It seems all that's needed is some logic to arrange messages in a certain way and to turn a bunch of them into tensors.

### DIFF
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -100,6 +100,8 @@ from rasa.utils.tensorflow.constants import (
     MODEL_CONFIDENCE,
     SOFTMAX,
 )
+from functools import partial
+from rasa.utils.tensorflow.data_generator import MessageStreamDataGenerator
 
 logger = logging.getLogger(__name__)
 
@@ -601,7 +603,7 @@ class DIETClassifier(IntentClassifier, EntityExtractor):
         labels_idx_examples = []
         for label_name, idx in label_id_dict.items():
             label_example = self._find_example_for_label(
-                label_name, training_data.intent_examples, attribute
+                label_name, training_data.stream_featurized_messages("test2"), attribute
             )
             labels_idx_examples.append((idx, label_example))
 
@@ -757,7 +759,7 @@ class DIETClassifier(IntentClassifier, EntityExtractor):
         model_data.add_lengths(LABEL, SEQUENCE_LENGTH, LABEL, SEQUENCE)
 
     # train helpers
-    def preprocess_train_data(self, training_data: TrainingData) -> RasaModelData:
+    def preprocess_train_data(self, training_data: TrainingData) -> None:
         """Prepares data for training.
 
         Performs sanity checks on training data, extracts encodings for labels.
@@ -768,10 +770,10 @@ class DIETClassifier(IntentClassifier, EntityExtractor):
         label_id_index_mapping = self._label_id_index_mapping(
             training_data, attribute=INTENT
         )
-
-        if not label_id_index_mapping:
-            # no labels are present to train
-            return RasaModelData()
+        #
+        # if not label_id_index_mapping:
+        #     # no labels are present to train
+        #     return RasaModelData()
 
         self.index_label_id_mapping = self._invert_mapping(label_id_index_mapping)
 
@@ -780,20 +782,20 @@ class DIETClassifier(IntentClassifier, EntityExtractor):
         )
 
         self._entity_tag_specs = self._create_entity_tag_specs(training_data)
-
-        label_attribute = (
-            INTENT if self.component_config[INTENT_CLASSIFICATION] else None
-        )
-
-        model_data = self._create_model_data(
-            training_data.nlu_examples,
-            label_id_index_mapping,
-            label_attribute=label_attribute,
-        )
-
-        self._check_input_dimension_consistency(model_data)
-
-        return model_data
+        #
+        # label_attribute = (
+        #     INTENT if self.component_config[INTENT_CLASSIFICATION] else None
+        # )
+        #
+        # model_data = self._create_model_data(
+        #     training_data.nlu_examples,
+        #     label_id_index_mapping,
+        #     label_attribute=label_attribute,
+        # )
+        #
+        # self._check_input_dimension_consistency(model_data)
+        #
+        # return model_data
 
     @staticmethod
     def _check_enough_labels(model_data: RasaModelData) -> bool:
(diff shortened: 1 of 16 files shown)


rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -835,14 +844,13 @@ def train(
...
 
-        data_generator, validation_data_generator = train_utils.create_data_generators(
-            model_data,
-            self.component_config[BATCH_SIZES],
-            self.component_config[EPOCHS],
-            self.component_config[BATCH_STRATEGY],
-            self.component_config[EVAL_NUM_EXAMPLES],
-            self.component_config[RANDOM_SEED],
+        data_generator, validation_data_generator = (
+            MessageStreamDataGenerator(
```
wochinge, line 848: do you know if the `Caution:` part from [here](https://www.tensorflow.org/guide/data#consuming_python_generators)  applies?

rasa/utils/tensorflow/data_generator.py
```diff
@@ -423,3 +426,66 @@ def _linearly_increasing_batch_size(self) -> int:
...
+            try:
+                messages.append(next(self.current_message_stream))
+            except StopIteration:
+                break
+        data = self.message_to_data_function(messages)
+        return RasaDataGenerator.prepare_batch(data), None
+
+    def on_epoch_end(self) -> None:
+        """Update the data after every epoch."""
+        self.current_message_stream = self.training_data.stream_featurized_messages(
```
wochinge, line 489: how would we e.g. shuffle chunks after epoch end?

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -734,19 +698,24 @@ def train(
...
-                    training_data.training_examples,
-                )
+        self.initialize(training_data, config, **kwargs)
+
+        msg_pipe = training_data.add_features(None, "test2")
+        try:
+            msg = next(msg_pipe)
+            while msg:
+                features = self.msg_to_features(msg)
+                msg = msg_pipe.send(features)
```
wochinge, line 716: using `send` is neat!

rasa/nlu/classifiers/diet_classifier.py
twerkmeister: Yeah maybe something like Dataset.from_generator would be the right way to go

rasa/utils/tensorflow/data_generator.py
twerkmeister: The TrainingData object could tell the featurereader to read the chunks in a random/different order. And then you still need to assure that the right message gets the right features. Some ID of the message that is also stored in the chunk or so.

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
twerkmeister: Yeah first time for me! It's cool, but easy to get wrong. At first I only stored every second feature set because I both consumed the generator and used send too. Didn't know that send would also make the generator continue and return the next yield immediately.

wochinge: It makes sense once you think about it but it's not intuitive 🙈

rasa/nlu/classifiers/diet_classifier.py
wochinge: so it seems `tf.data` is preferred for non-python preprocessing (== all pre-processing is expressed via TensorFlow functions) while the Keras sequence is preferred if you do a lot of Python preprocessing. So in our case the Keras sequence seems to be preferable.

rasa/utils/tensorflow/data_generator.py
wochinge: sweet, thanks for implementing this as well 💯

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -572,14 +582,14 @@ def _train_model(self, df_train: List[List[CRFToken]]) -> None:
...
             X_train = (
-                self._crf_tokens_to_features(sentence, include_tag_features)
-                for sentence in df_train
+                self._message_to_features(msg, tag_name)
+                for msg in training_data.stream_featurized_messages()
             )
             y_train = (
-                self._crf_tokens_to_tags(sentence, tag_name) for sentence in df_train
+                self._message_to_tags(msg, tag_name)
+                for msg in training_data.stream_featurized_messages()
```
wochinge, line 592: that's potentially dangerous if shuffle would be enabled as X and Y would be shuffled differently, right?

wochinge (commented):
How about showing this to Daksh / Vova? I'd be interested in their feedback. I think your prototype seems very promising! 💯

twerkmeister:
I ran some mem profilings on multiwoz dataset with roughly 55k nlu sentences. I compared the current main branch with my branch here. It saves memory all around, but the effect is very much dependent on the specific component. It's the least pronounced for the crf in this case as that model is building up the entire dataset internally itself.

### config 
I think most of it pretty standard, added the crf features here to get `text_dense_features` from the spacy vectors in
```
language: en

pipeline:
  - name: "SpacyNLP"
  - name: "SpacyTokenizer"
  - name: "SpacyFeaturizer"
  - name: "CRFEntityExtractor"
    max_iterations: 1
    features: [
    ["pos", "pos2"],
    [
      "bias",
      "prefix5",
      "prefix2",
      "suffix5",
      "suffix3",
      "suffix2",
      "pos",
      "pos2",
      "digit",
      "text_dense_features"
    ],
    ["pos", "pos2"],
    ]
    L1_c: 0.01,
    L2_c: 0.05
  - name: "DIETClassifier"
    entity_recognition: False
    epochs: 1
    batch_size: 64
```

### Results
##### main
standard_with_child

##### this branch
streamed_with_child

Memory:
* For each model jumps to about 2.5 GB when loading spacy model and tokenizing all the messages
* During crf data accumulation:
   * main: jumps and hovers at around 5.5 to 6GB
   * this branch: slow increase from 2.5 to about 4.5GB
* During crf training:
   * main: peak memory usage at about 7.5 GB
   * this branch: peak memory usage at about 6 GB
* During Diet training
   * main: constant memory consumption of around 4.3 GB
   * this branch: memory consumption of around 2-2.2 GB

Time:
* main: ~1600s
* this branch: ~1850s   
* Maybe a minute of that goes to CRF training or before that, both of them end at similar times at around 1000s
* The rest seems to be attributable to DIET training
* I think this is mostly expected as 1) the current implementation for the keras sequence is very rough, and 2) caching on disc is fundamentally a trade off between memory and time

rasa/nlu/extractors/crf_entity_extractor.py
twerkmeister: Yes, indeed that would break. The reason this is here is that you can only consume a generator once. However, you could split the generator, if you can guarantee that no copy is advancing more than `n` elements ahead of the others. It would be possible to do this here as the crf internally zips x and y and then runs through them together.

twerkmeister:
> How about showing this to Daksh / Vova? I'd be interested in their feedback. I think your prototype seems very promising! 💯

I'd be interested in their thoughts aswell

twerkmeister:
With splitting (reusing) the generators inside the crf the consumption is again 500mb lower. Timing isn't really comparable as I was doing a lot of other stuff while profiling.

streamed_with_child_combined_crf_stream

rasa/nlu/extractors/crf_entity_extractor.py
twerkmeister: updated that code again a bit

joejuzl:
How does changing the `MSG_PER_CHUNK` affect the memory usage?

rasa/nlu/classifiers/diet_classifier.py
dakshvar22: > Yeah maybe something like Dataset.from_generator would be the right way to go

Constructing a `tf.data.Dataset` from `Dataset.generator` becomes cumbersome when you need control on what should happen after the end of every epoch. If you need that control you need to write a custom `train_step` function instead of replying on `keras.model.fit()` to handle that for you. 

[tf.keras.utils.Sequence](https://www.tensorflow.org/api_docs/python/tf/keras/utils/Sequence#on_epoch_end) explicitly has a function for it. That's why we shifted from `tf.data.Dataset.from_generator` to `tf.keras.utils.Sequence` during the transition from [pure TF to reusing Keras methods](https://github.com/RasaHQ/rasa/pull/7438).

However, we recently recognized that tensorflow has [breaking support for `tf.RaggedTensors` inside `tf.keras.utils.Sequence`.](https://github.com/tensorflow/tensorflow/issues/48188#issue-844491927). 
Using `tf.RaggedTensors` will simplify a LOT of code inside our TF models. So, I am all up for exploring how we can use `tf.data.Dataset.from_generator` without having to write our own `model.fit()` function.

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
dakshvar22: I still haven't been able to wrap my head around how it's working, but I okay considering it as some sort of magic 😅  for now which sets the features computed by the specific featurizer for that message, right?

wochinge: >  for now which sets the features computed by the specific featurizer for that message, right?

Yes. With the advantage that we control how features are added outside of the components (which in turn means the `Component`s don't have to know about these details). Whether we write the features to disks for chunks or keep them in memory is completely abstracted and exchangable.

wochinge: And `Components` are also not aware where data is coming from. It could be in-memory, read from disk or some stuff is coming from in-memory and some is coming from disk.

dakshvar22: > With the advantage that we control how features are added outside of the components (which in turn means the Components don't have to know about these details).

But we do make a distinction between different kind of features? for e.g. - sparse and dense features.

twerkmeister:
> How does changing the `MSG_PER_CHUNK` affect the memory usage?

The larger the chunk the more memory is used. At the extreme you could have a chunk that is the size of the dataset even for a big dataset.

The features in this particular setup for 55k messages are about 2GB in memory. So 512 messages are about 20MB.

For the shuffling I am currently using 3 random chunks. So that would make 60mb when using shuffling.

Both the size and number of chunks when shuffling is adjustable.

twerkmeister:
@wochinge @dakshvar22 Brief update on the benchmarking over multiple epochs:

#### Balanced Batching and Preprocessing
Balanced batching can make datasets quite a bit. If you look at my old two graphs of main vs streaming again there's something interesting. Turns out the noisy pattern at the end is actually the DIET training. So it was much faster on the streaming branch, but only because main used balacned batching by default. Given that the nlu data of multiwoz follows a power law for examples per intent (few intents have thousands of examples, and a lot of intents only have a few examples) balanced batching effectively really more than doubled the size of the dataset during training. 

The first long flat stretch in the streaming training on the other hand was a preprocessing step that was very costly on streaming form disc. I would load the data more than a 100 times from disc during this step in the first naive implementation.

With the batching strategy set to `sequence` (no rebalancing) and adjusting the preproccessing slightly, the two systems are now more comparable

#### Timing comparison
I ran the experiments multiple times and consistently observed a 10-15% time penalty for the streaming solution. So for example, on the latest run streaming would take 39s per epoch on an 80% multiwoz train set, and in memory would take 34s. That was consistent over three runs, doing 15-25 epochs each on a google cloud k80 instance.

Issue #8532, opened 2021-04-23
Conditional Response Variations: amend domain schema and TemplatedNaturalLanguageGenerator


ancalita:
**Description of Problem**:
Adjusting response variations based on conditions is not currently handled in a uniform way, most often this involves writing custom actions code. This can prove to be a blocker for non-technical users which would prefer to maintain or build new response texts without having to rely heavily on custom actions code.

**Overview of the Solution**:
A detailed view of the scoping and implementation proposal can be found [here](https://www.notion.so/rasa/Conditional-Utterances-Selection-d9b4f96c4c1b47a983e78a651f965bab). The first iteration aims to implement unnamed conditions format, where `channel` field is not a constraint part of `condition`.

**Blockers** (if relevant):
None

**Definition of Done**:
- [x] amend `rasa/shared/nlu/training_data/schemas/responses.yml` to reflect new YAML format.
- [x] `TemplatedNaturalLanguageGenerator` handles new response format and filters by slot-value(s) based condition and channel if any.
- [x] mark feature as experimental in the command line using `rasa.shared.utils.common.mark_as_experimental_feature`
- [x] Tests are added
- [x] Feature is mentioned in the changelog
- [x] Telemetry is implemented to track uptake of feature.

PR #8551, opened 2021-04-26
Conditional Response Variations Implementation: v1 iteration

ancalita:
**Proposed changes**:
- Implement #8532 


rasa/core/nlg/response.py
```diff
@@ -24,29 +24,56 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
                 default_responses.append(response)
+            else:
+                matched_response = self._matches_filled_slots(
+                    filled_slots=filled_slots, response=response
+                )
+                if matched_response:
+                    conditional_responses.append(response)
+
+        if conditional_responses:
+            potential_responses = conditional_responses[:]
```
wochinge, line 60: why do we need the `[:]` part?

rasa/core/nlg/response.py
```diff
@@ -24,29 +24,56 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
+
+    def _matches_filled_slots(
+        self, filled_slots: Dict[Text, Any], response: Dict[Text, Any],
+    ) -> bool:
+        """Checks if the conditional response variation matches the filled slots."""
+        condition_list = response.get("condition")
+        for condition in condition_list:
+            name = condition["name"]
+            value = condition["value"]
+            if filled_slots.get(name) == value:
```
wochinge, line 37: this returns `True` as soon as one condition matches - what if there are multiple conditions?

rasa/core/nlg/response.py
ancalita: True, not required - removed.

rasa/core/nlg/response.py
ancalita: Very good point - changed the method a bit to support multiple constraints.

rasa/core/nlg/response.py
```diff
@@ -25,28 +25,58 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
+        for constraint in constraints:
+            name = constraint["name"]
+            value = constraint["value"]
+            if filled_slots.get(name) == value:
+                satisfied_constraints.append(constraint)
+
+        if len(satisfied_constraints) == len(constraints):
+            return True
+
+        return False
```
wochinge, line 43: optional: slightly simpler in my opinion 
```suggestion
        constraints = response.get("condition")
        for constraint in constraints:
            name = constraint["name"]
            value = constraint["value"]
            if filled_slots.get(name) != value:
                return False

        return True
```

rasa/core/nlg/response.py
```diff
@@ -25,28 +25,58 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
+
     def _responses_for_utter_action(
-        self, utter_action: Text, output_channel: Text
+        self, utter_action: Text, output_channel: Text, filled_slots: Dict[Text, Any]
     ) -> List[Dict[Text, Any]]:
-        """Return array of responses that fit the channel and action."""
-        channel_responses = []
+        """Return array of responses that fit the channel, action and condition."""
```
wochinge, line 48: according to our [code conventions](https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#30a6eae49f8f4b4f97c5cf39b577a516) we use "descriptive" docstrings (third person instead of imperative)
```suggestion
        """Returns array of responses that fit the channel, action and condition."""
```

tests/core/test_nlg.py
```diff
@@ -328,3 +328,126 @@ def test_nlg_fill_response_quick_replies(
...
+            {
+                "text": "Conditional Response B",
+                "condition": [{"type": "slot", "name": "slot test", "value": "B"}],
+            },
+            {"text": "Default response"},
+        ]
+    }
+    t = TemplatedNaturalLanguageGenerator(responses=responses)
+
+    condition_a = t._matches_filled_slots(
```
wochinge, line 355: I'd vote for testing through the public interface which is `TemplatedNaturalLanguageGenerator.generate` instead of private / protetected methods which are too closely tied to the implementation

tests/core/test_nlg.py
```diff
@@ -328,3 +328,126 @@ def test_nlg_fill_response_quick_replies(
...
+        filled_slots={"slot test": "A"}, response=responses["utter_test"][0]
+    )
+    assert condition_a is True
+
+    condition_b = t._matches_filled_slots(
```
wochinge, line 360: this seems like multiple separate tests to me. I'd split this either in different tests or use `pytest.parametrize`  (e.g. see `Unit Tests Should Verify a Single-Use Case` [here](https://dzone.com/articles/unit-testing-best-practices-how-to-get-the-most-ou)

wochinge: e.g. named tests would make it easier to me to grasp which case which assertion is testing

wochinge (commented):
Blazing fast implementation 🚀

tests/core/test_nlg.py
wochinge: (same for the other tests btw)

tests/core/test_nlg.py
ancalita: Yup, all testing-related feedback taken onboard - I wrote these phase 1 tests to roughly check the implementation. I'll refactor them: I'll have a think how best to create a tracker with the slot mappings I want in order to use the `generate` method.

rasa/core/nlg/response.py
ancalita: Very nice! I will follow this suggestion 👍

rasa/core/nlg/response.py
```diff
@@ -25,28 +25,54 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
-        channel_responses = []
+        """Returns array of responses that fit the channel, action and condition."""
         default_responses = []
+        conditional_responses = []
 
         for response in self.responses[utter_action]:
-            if response.get("channel") == output_channel:
-                channel_responses.append(response)
-            elif not response.get("channel"):
+            if response.get("condition") is None:
```
wochinge, line 49: optional: might be worth making `condition` a constant. 2 usages is kind of at the edge but I think it's likely that Rasa X might need it as well.

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
...
+    ("slot_name", "slot_value", "output_channel", "response_variation"),
+    (
+        ("slot test", "A", "os", "Conditional Response A"),
+        ("slot test", "A", "app", "Conditional app response A"),
+        ("slot test", "B", "", "Conditional Response B"),
+        ("slot test", "C", "", "Default response"),
+    ),
+)
+async def test_nlg_conditional_response_variations(
+    slot_name, slot_value, output_channel, response_variation
```
wochinge, line 346: can you please add types here?

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
...
+    r = await t.generate(
+        utter_action="utter_test", tracker=tracker, output_channel=output_channel
+    )
+    assert r.get("text") == response_variation
+
+    no_slots_tracker = DialogueStateTracker(sender_id="nlg_test_default", slots=None)
+    default_response = await t.generate(
+        utter_action="utter_test", tracker=no_slots_tracker, output_channel=""
+    )
+    assert default_response.get("text") == "Default response"
```
wochinge, line 381: optional: slightly nitpicking but in my opinion that would be a different test. Not a huge deal though (hence `optional`)

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
...
+    ),
+)
+async def test_nlg_conditional_response_variations(
+    slot_name, slot_value, output_channel, response_variation
+):
+    responses = {
+        "utter_test": [
+            {
+                "text": "Conditional Response A",
+                "condition": [{"type": "slot", "name": "slot test", "value": "A"}],
```
wochinge, line 352: ?
```suggestion
                "condition": [{"type": "slot", "name": slot_name, "value": "A"}],
```

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
...
+    no_slots_tracker = DialogueStateTracker(sender_id="nlg_test_default", slots=None)
+    default_response = await t.generate(
+        utter_action="utter_test", tracker=no_slots_tracker, output_channel=""
+    )
+    assert default_response.get("text") == "Default response"
+
+
+async def test_nlg_when_multiple_conditions_satisfied():
+    responses = {
+        "utter_action": [
```
wochinge, line 386: optional: Does it make sense to have a third response which doesn't match?

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
...
+        "second_slot_value",
+        "response_variation",
+    ),
+    (
+        ("test", "A", "test_other", "B", "example AB"),
+        ("test", "C", "test_other", "D", "example CD"),
+    ),
+)
+async def test_nlg_with_multiple_constraints(
+    first_slot, first_slot_value, second_slot, second_slot_value, response_variation
```
wochinge, line 422: types please

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
...
+@pytest.mark.parametrize(
+    (
+        "first_slot",
+        "first_slot_value",
+        "second_slot",
+        "second_slot_value",
+        "response_variation",
+    ),
+    (
+        ("test", "A", "test_other", "B", "example AB"),
```
wochinge, line 417: how about a test where a slot is `null`?

wochinge: or where there is a slot which isn't referenced by any constraint?

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
         filled_slots={quick_replies_slot_name: quick_replies_slot_value},
     )
     assert result == {"quick_replies": str(quick_replies_slot_value)}
+
```
wochinge, line 334: can we at least move your tests to a module `tests.core.nlg.test_response.py` to mirror that you're testing the `response.py` module? It's legacy that we don't have this mirroring between production code and test code and it's not your job to fix it but we can at least do it right for new tests. This way it's way easier to see which tests are testing what - otherwise one might have to search for tests.

tests/core/test_nlg.py
```diff
@@ -328,3 +331,271 @@ def test_nlg_fill_response_quick_replies(
...
+            "You are not allowed to withdraw any amounts. Please check permission.",
+        ),
+        (
+            "account_type",
+            "secondary",
+            "Withdrawal was sent for approval to primary account holder.",
+        ),
+    ),
+)
+async def test_nlg_conditional_response_variations_with_yaml_single_condition(
```
wochinge, line 504: the yaml tests feel a bit duplicate to me. What do they test what your previous tests don't test? You could also write the responses as yaml string within this module and use it in your tests by doing something like `Domain.from_yaml` -> `Domain.responses`

tests/core/test_nlg.py
wochinge: and a test for a condition which requires a slot to be `None` would be interesting!

wochinge (commented):
Some small comments around the tests - the change itself looks great and is in every way ready to go 💯 my comments in the tests are a bit nitpicking (test coverage etc. is great 💯 ) so please feel free to oppose any comment if it feels like an overkill.

One question: We have the policy that we don't merge new features without documentation. Are you adding the documentation in this branch or will you do a separate PR which targets this branch or what's the plan? 

In my opinion we can merge to `main` as soon as the documentation is ready. From what I remember the rest can be merged separately then.

rasa/core/nlg/response.py
wochinge, line 49: @ancalita I think this line is causing your bug. You might be adding responses with a specified `channel` to `default_responses`. This in turn might end with you returning channel specific responses in case a not matching `output_channel` is set in line 71.  


```
def test_anca():
    domain = Domain.from_yaml("""
    responses:
        utter_hi:
        - text: "hello"
        - text: "Hello Slack"
          channel: "slack"
    """)

    t = TemplatedNaturalLanguageGenerator(domain.responses)
    responses = t._responses_for_utter_action("utter_hi", "signal", {})

    assert len(responses) == 1
```

ancalita: Thanks, great insight! I have now fixed line 71 to filter out any non-matching channel responses 🥳

ancalita: Yup, can definitely action this - is there a certain file I should turn this into a constant? Or at the top of `response.py` is fine as well?

tests/core/test_nlg.py
ancalita: Yup, good spot, that will make the test complete. I've added two more responses to make it bulletproof 😄

tests/core/test_nlg.py
ancalita: The first two suggestions are very good and prove that we need to enforce a default for these cases. But I'll see what errors are raised when no response variation is specified as default and slot is null.
I'm not entirely sure about the third suggestion - is that in the case when a user specifies None value for a slot in `domain.yml` ?

ancalita: When the slot is null or not referenced by any constraint, and no response variation as default, then the response generated is None - this could cause AttributeError when later the assistant tries to get value for `text` key. I'll list this as something we should recommend in the docs - to cover all bases with a default.

tests/core/test_nlg.py
ancalita: Yup sure, two questions:
- Should I move both the new `test_response.py` and the legacy `test_nlg.py` to `tests.core.nlg` directory?
- I won't have to rename all tests in `test_response.py` from `test_nlg` to `test_response_etc`, would I? 😬

rasa/core/nlg/response.py
wochinge: I'd go for `rasa.shared.constants` or `rasa.shared.nlu.constants`. From the point of this PR it's fine to define it at the top of this file, but considering Rasa X and also that you'll need that for sanity checks in the domain sooner or later, I'd go for a `constants` file

wochinge: I think you didn't push yet, right?

tests/core/test_nlg.py
ancalita: I have received feedback before that I should always test importing from file to make sure everything works as expected in an almost 'real-life' scenario - I guess in the first tests I used a `dict `directly because I wanted to focus on the new implementation rather to bypass any yaml validation errors. 
I could remove anything that appears as a duplicate and just keep the yaml ones.

tests/core/test_nlg.py
wochinge: > I'm not entirely sure about the third suggestion - is that in the case when a user specifies None value for a slot in domain.yml ?

Let's say there is a slot `name` and for some reason the user doesn't provide it which means it becomes `None` (it could have had a different value before).

Is this

```yaml
responses:
  utter_hi:
  - text: "Hello"
```

the same as this?

```yaml
responses:
  utter_hi:
  - text: "Hello"
     condition:
     - type: slot
        name: name
        value: null
```

would in my opinion still be nice to have a test to ensure this is the same / works, what do you think?

rasa/core/nlg/response.py
ancalita: Cool, I'll define it in `rasa.shared.constants` 👍

ancalita: Nope, not yet 😄 I was going through all the feedback first.

tests/core/test_nlg.py
wochinge: > I have received feedback before that I should always test importing from file to make sure everything works as expected in an almost 'real-life' scenario

That's good feedback 👍🏻 Especially if you think of things like schema validation.

> I guess in the first tests I used a dict directly because I wanted to focus on the new implementation rather to bypass any yaml validation errors.

fair 👍🏻 

> I could remove anything that appears as a duplicate and just keep the yaml ones.

I'd slightly prefer this as I also think the YAML syntax is more intuitive to read / understand when reading the code as well. Not worth to put too much work into this though - you could also just keep the existing tests and maybe reduce the YAML ones a bit and rather treat them as some sort of integration test.

tests/core/test_nlg.py
wochinge: > Should I move both the new test_response.py and the legacy test_nlg.py to tests.core.nlg directory?

Up to you. In my opinion it's also fine to focus upon your changes and to keep the diff small.

tests/core/nlg/test_response.py
```diff
@@ -0,0 +1,291 @@
...
+    r = await t.generate(
+        utter_action="utter_action", tracker=tracker, output_channel=""
+    )
+    assert r.get("text") == message
+
+
+async def test_nlg_non_matching_channel():
+    domain = Domain.from_yaml(
+        """
+    responses:
```
wochinge, line 220: we need to add the `version: "2.0"` part - otherwise this test will cause a warning in the CI due to the missing version part 😬

ancalita: I see, I'll add this in the test yaml file I created too 👍

docs/docs/telemetry/events.json
```diff
@@ -57,6 +57,14 @@
           "type": "integer",
           "description": "Number of templates or responses defined in the domain."
         },
+        "num_conditional_response_variations": {
```
wochinge, line 60: it just occurred to me that we also have to update `telemetry.mdx`  (I think yours is not the first one to be missing there 😬 )

rasa/core/nlg/response.py
```diff
@@ -25,28 +26,55 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
         """
         self.responses = responses
 
+    def _matches_filled_slots(
+        self, filled_slots: Dict[Text, Any], response: Dict[Text, Any],
+    ) -> bool:
+        """Checks if the conditional response variation matches the filled slots."""
+        constraints = response.get(CONDITION)
```
wochinge, line 33: optional: maybe the name `RESPONSE_CONDITION` would make the usage of the constant a bit clearer

rasa/shared/core/domain.py
```diff
@@ -591,6 +592,15 @@ def __init__(
         action_names += overridden_form_actions
 
         self.responses = responses
+        # checks if responses contain conditions
```
wochinge, line 595: optional: how about extracting that to a separate method / function? It in my opinion has a different level of abstraction than the other code in the constructor (http://principles-wiki.net/principles:single_level_of_abstraction)

tests/core/nlg/test_response.py
```diff
@@ -0,0 +1,295 @@
...
+                "condition": [{"type": "slot", "name": "test_int", "value": 12}],
+            },
+            {
+                "text": "example list",
+                "condition": [{"type": "slot", "name": "test_list", "value": []}],
+            },
+        ]
+    }
+    t = TemplatedNaturalLanguageGenerator(responses=responses)
+    slot = Slot(name=slot_name, initial_value=slot_value, influence_conversation=False)
```
wochinge, line 208: this uncovers a faulty piece in our code. `Slot` is actually an abstract class and shouldn't be instantiated (the methods raising `NotImplementedError` should be marked with `abc.abstractmethod`). Out of scope of your PR but if you truly want to test different slot types you should use things like `TextSlot` or `ListSlot`

wochinge (approved):
Blazing fast and high quality implementation with top notch test coverage 💯

docs/docs/telemetry/events.json
ancalita: Np, I'll add the properties after [L96 ](https://github.com/RasaHQ/rasa/blob/main/docs/docs/telemetry/telemetry.mdx)- is that the only update I have to make?

rasa/shared/core/domain.py
ancalita: Yup, I'm not very happy about this check in the constructor - the question is where would I make this call to the separate method? Or should I just extract to an external function that I then call in the constructor?

ancalita: I've opted finally for a function that gets called in the constructor which I'll push in the next few minutes.

docs/docs/telemetry/events.json
wochinge: yep 👍🏻 Maybe have a quick look if something else is missing

rasa/telemetry.py
```diff
@@ -717,6 +722,8 @@ async def track_model_training(
             # Old nomenclature from when 'responses' were still called
             # 'templates' in the domain
             "num_templates": len(domain.responses),
+            "num_conditional_response_variations": count_conditional_responses,
+            "has_conditional_response_variations": has_conditional_responses,
```
kedz, line 726: If we are also recording counts of CRVs do we need to still record this boolean field?

rasa/core/nlg/response.py
```diff
@@ -25,28 +26,55 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
+                )
+                if matched_response:
+                    conditional_responses.append(response)
+
+        if conditional_responses:
+            potential_responses = conditional_responses
+        else:
+            potential_responses = default_responses
+
+        channel_responses = list(
```
kedz, line 64: I think channel filtering should happen first and then apply constraint filtering. The following scenario fails currently

```python    
    domain = Domain.from_yaml(
        """
        version: "2.0"
        responses:
           utter_action:
             - text: "example with channel"
               condition:
                - type: slot
                  name: test
                  value: A
               channel: os
             - text: "app default"
               channel: app
        """
    )
    t = TemplatedNaturalLanguageGenerator(domain.responses)
    slot = TextSlot("test", "A", influence_conversation=False)
    tracker = DialogueStateTracker(sender_id="test", slots=[slot])
    r = await t.generate("utter_action", tracker, "app")
    assert r.get("text") == "app default"
```
because the constraint matches so defaults are discarded, but then no channel + constraint option works and so None is returned.

kedz (changes requested):
Overall it looks really great! There's one edge case that I made an inline comment for, but once this is dealt with, it looks good to me!

rasa/telemetry.py
ancalita: Very good point, @wochinge asked the same - I can remove the boolean field, I just thought both metrics are wanted based on what was agreed in the scoping doc.

rasa/core/nlg/response.py
ancalita: Reversing the order of constraints is not enough to solve for this edge case - I've just tried this different implementation, and it fails another edge case present in the unit tests (when channel is matched, but condition is not). Instead of spending too much time trying to figure this one out, I've kept the original implementation and added a new check to it and all tests now pass 🥳

wochinge: I think @kedz meant something like this?

```python
        responses_matching_channel = [
            r for r in self.responses[utter_action] if CHANNEL not in r
        ]
        if output_channel:
            responses_matching_channel = [
                response
                for response in self.responses[utter_action]
                if response.get(CHANNEL) == output_channel
            ]

        matching_both = [
            response
            for response in responses_matching_channel
            if RESPONSE_CONDITION in response
            and self._matches_filled_slots(filled_slots, response)
        ]

        if not matching_both:
            matching_both = [
                response
                for response in responses_matching_channel
                if RESPONSE_CONDITION not in response
            ]

        if matching_both:
            return matching_both

        default_answers = [
            response
            for response in self.responses[utter_action]
            if CHANNEL not in response and RESPONSE_CONDITION not in response
        ]

        return default_answers
```

rasa/core/nlg/response.py
```diff
@@ -25,28 +26,67 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
+        )
 
         # always prefer channel specific responses over default ones
         if channel_responses:
             return channel_responses
-        else:
-            return default_responses
+
+        # if no channel match in conditional responses, search in default responses
+        if len(output_channel) > 0:
```
wochinge, line 75: an empty string evaluates to `False`
```suggestion
        if output_channel:
```

rasa/core/nlg/response.py
ancalita: I see, I tried doing something similar but didn't succeed. 
I'm happy to keep which version is more readable 😄 No strong preference either way, although your version could end up looping twice through `self.responses[utter_action]` in the worst-case - probably not a huge concern?

kedz (changes requested):
The latest change fixes the last edge case but I've found a few more issues. 
The following edge cases are still breaking and I think they should be handled by this feature:
```python
@pytest.mark.parametrize(
    "slots,channel,expected_response",
    [
        (
            [
                TextSlot("test", "B", influence_conversation=False),
            ],
            "app",
            "condition example B no channel"
        ),
        (
            [
                TextSlot("test", "C", influence_conversation=False),
            ],
            "",
            "default"
        ),
        (
            [
                TextSlot("test", "D", influence_conversation=False),
            ],
            "app",
            "default"
        ),

    ]
)
async def test_edgecases(
    slots,
    channel,
    expected_response
):
    domain = Domain.from_yaml(
        """
        version: "2.0"
        responses:
           utter_action:
             - text: "condition example A with channel"
               condition:
                - type: slot
                  name: test
                  value: A
               channel: app

             - text: "condition example C with channel"
               condition:
                - type: slot
                  name: test
                  value: C
               channel: app

             - text: "condition example A no channel"
               condition:
                - type: slot
                  name: test
                  value: A

             - text: "condition example B no channel"
               condition:
                - type: slot
                  name: test
                  value: B

             - text: "default"
        """
    )
    t = TemplatedNaturalLanguageGenerator(domain.responses)
    tracker = DialogueStateTracker(sender_id="test", slots=slots)
    r = await t.generate("utter_action", tracker, channel)
    assert r.get("text") == expected_response
```
In the first case since there is no channel default so None is returned. I think the thing to do here is check the None channel for crvs that match and if nothing is found then try for the None channel defaults. In the second case, the None channel defaults are not being tried. The third case above currently works correctly. 

I apologize, I didn't fully specify this in the prototype but it would probably simplify things if we decide on an explicit order of precedence. There are four classes of responses:
1. CRVs with channels
2. defaults with channels
3. CRVs with no channel
4. defaults with no channel.

If the current session has a no channel set:
- look in 3 for a response
- if no matches are found look in 4
- otherwise return None

If the current session does have a channel:
- look in 1 for a CRV with matching constraints and channel
- if none look in 2 for a default with matching channel
- if none look in 3 for matching constraints
- if none look in 4 for defautl
- return None otherwise

Does that seem reasonable? Or does that break other things that have been decided?

wochinge:
Thanks for uncovering this case @kedz 👍🏻 Ideally we can abstract that logic a bit as we will likely have more condition types in the future.

ancalita:
@kedz thanks for uncovering more edge cases and for the flow you shared 🙌 That was really helpful for me in redesigning the implementation and all unit tests now pass and I hope the change is readable and maintainable 🤞

rasa/core/nlg/response.py
```diff
@@ -25,28 +26,71 @@ def __init__(self, responses: Dict[Text, List[Dict[Text, Any]]]) -> None:
...
-                channel_responses.append(response)
-            elif not response.get("channel"):
+            if response.get(RESPONSE_CONDITION) is None:
                 default_responses.append(response)
+            else:
+                matched_response = self._matches_filled_slots(
+                    filled_slots=filled_slots, response=response
+                )
+                if matched_response:
+                    conditional_responses.append(response)
```
wochinge, line 57: how adapting this and the prior code to your `filter` stuff below so that it's more uniform?

ancalita: Ok will do, I was also thinking of simplifying the if statements by removing the top first 2 if statements and adapting according to the priority order outlined by Chris.

kedz (approved):
Looks great and super maintainable/understandable to me!

wochinge (approved):
💯 Will you add the documentation here so that CSE can have a final look?

docs/docs/glossary.mdx
```diff
@@ -45,6 +45,10 @@ description: Glossary for all Rasa-related terms
 
   Incoming messages are processed by a sequence of components called a pipeline. A component can perform tasks ranging from entity extraction to intent classification to pre-processing.
 
+## [Conditional Response Variation](./responses.mdx#conditional-response-variations-experimental)
+
+   Response variation that can only be used when the current dialogue state satisfies some constraints as defined in the domain or responses files. If there's a match between the constraints and user filled slots, Rasa can use this variation.
```
melindaloubser1, line 50:
```suggestion
   Response variation that can only be used when the current dialogue state satisfies some constraints as defined in the domain or responses files. If there's a match between the constraints and the dialogue state, Rasa can use this variation.
```
Since the definition doesn't limit constraints to slots

docs/docs/responses.mdx
```diff
@@ -140,6 +140,135 @@ in the shell and in interactive learning.
...
+
+When a response is triggered during a dialogue, the constraints of each conditional response variation
+are checked against the current dialogue state. If all constraint slot values are equal to the corresponding
+slot values of the current dialogue state, the response variation is eligible to be used by your conversational
+assistant.
+
+:::note
+The comparison of dialogue state slot values and constraint slot values is performed by the
+equality "==" operator which requires the type of slot values to match too.
+:::
```
melindaloubser1, line 169:
```suggestion
The comparison of dialogue state slot values and constraint slot values is performed by the
equality "==" operator which requires the type of slot values to match too.  For example, if the constraint is specified as `value: true`, then the slot needs to be filled with a boolean `true`, not the string `"true"`.
:::
```

melindaloubser1 (approved):

PR #8552, opened 2021-04-26
Architecture prototype/rasa run

wochinge:
**Proposed changes**:
- use graph model if zipped version of the graph model is in the `models` directory
- refactory Agent / Processor slighly to move model logic into `Processor` so that we can swap it out with a `GraphProcessor`


wochinge:
cc @joejuzl

Issue #8439, opened 2021-04-12
IntentTEDPolicy: Implement Policy


dakshvar22:
**Description of Problem**:
`IntentTEDPolicy` uses a different tracker featurizer and a different loss function than `TEDPolicy`.

**Overview of the Solution**:
Implement a new policy called `IntentTEDPolicy` which uses the [tracker featurizer](https://github.com/RasaHQ/rasa/issues/8437) and [new loss function](https://github.com/RasaHQ/rasa/issues/8438).

A draft version of this is already implemented in `itedp_for_merge` branch.

[Scope Document](https://docs.google.com/document/d/1kjZD59NtnvAkxSfu6sjLzgDucTQT8LJhdfZnILCDVcE/edit?usp=sharing) explains the working of it during training and during inference.

[Implementation proposal](https://www.notion.so/rasa/Implementation-Proposal-IntentTedPolicy-f16c3eb785964b1ca1249d508f6dc234#6ceb8d8b71024e27b94c1809dcdc22d7) explains how it should predict [`action_unlikely_intent`](https://github.com/RasaHQ/rasa/issues/8417).

Ensure that any metadata added during inference is serialisable.

Have a `ranking_length` parameter which limits the length of the ranking for predicted intent confidence scores.

**Blockers**:
Should be tackled after https://github.com/RasaHQ/rasa/issues/8438 and https://github.com/RasaHQ/rasa/issues/8437 . Potentially blocked by https://github.com/RasaHQ/rasa/issues/8417 also for testing the inference time logic.

**Definition of Done**:
- [x] The policy uses the new tracker featurizer
- [x] The policy uses the new loss function
- [x] The priority of the policy is correctly set
- [ ] `ranking_length` parameter is added
- [x] Prediction time behaviour is correctly handled (according to the implementation proposal).
- [x] Tests are added
- [ ] Docs for the policy are added
- [ ] Feature mentioned in the changelog

Issue #7412, opened 2020-11-30
IntentTEDPolicy: Add a parameter `tolerance` to configuration


dakshvar22:
**Description of Problem**:

If we look at the distribution of similarity values predicted by `IntentTED` for a single intent, it would look something like this -

![distribution_plot](https://user-images.githubusercontent.com/8708249/100598446-bc8b0a00-3324-11eb-8546-ae8b49faa8a5.png)

The blue bars show similarity values predicted for all input trackers where the intent `install_rasa` was the positive label.
The orange bars show similarity values predicted for all input trackers where the intent `install_rasa` was the negative label.
 
Currently, the threshold is taken as the minimum value from all blue bars denoted by the red vertical line. We want this threshold value to be slightly more "flexible" which can account for cases where the threshold value and the actual predicted value at inference are extremely close. For e.g., the threshold can be - `0.11325` and the predicted value can be - `0.11322`. Such close cases did come up while evaluating the assistant on 3 different assistants and hence resulted in extra user turns flagged by `IntentTEDPolicy`.

**Overview of the Solution**:
- [ ] Add a parameter `tolerance` which can be either a floating point in `[0,1]` or one of `low`, `medium` and `high`. The idea is that depending on the value of `tolerance`, we slide the red vertical line further to the left to reduce the number of user events that the `IntentTEDPolicy` flags. Specifically, the floating point `tolerance` value would be the percentage of negative training examples that would still not be flagged. E.g. if `tolerance=0.30`, then 30% of the area of the negative-example bars (orange) would be to the right of the red line. What floating point numbers should be assigned to the `low`, `medium`, and `high` categories should be determined in a separate issue

**Blockers**:
Should be tackled after https://github.com/RasaHQ/rasa/issues/8439

Ghostvv:
@dakshvar22 this is blocked, since it looks like it doesn't make sense to work on it until `IntentTED` is merged?

dakshvar22:
It's one of the issues of the larger `IntentTEDPolicy` [issue](https://github.com/RasaHQ/research/issues/122). So, it needs to be picked up again once the work on it resumes. I'll un-assign myself from this for now as I don't see that happening at least in the next sprint.

PR #8555, opened 2021-04-26
IntentTEDPolicy: Implement Policy

dakshvar22:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/8439
- Does everything that the definition of done mentions on the issue except - `ranking_length parameter is added`. This will be done as part of https://github.com/RasaHQ/rasa/issues/7412.


rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,492 @@
...
+    def __init__(
+        self,
+        featurizer: Optional[TrackerFeaturizer] = None,
+        priority: int = UNLIKELY_INTENT_POLICY_PRIORITY,
+        max_history: Optional[int] = None,
+        model: Optional[RasaModel] = None,
+        fake_features: Optional[Dict[Text, List["Features"]]] = None,
+        entity_tag_specs: Optional[List[EntityTagSpec]] = None,
+        should_finetune: bool = False,
+        intent_thresholds: Dict[int, float] = None,
```
JEM-Mosig, line 247: Why not `List[float]` or `np.array`?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,492 @@
...
+            confidences,
+            action_metadata=self._collect_action_metadata(domain, similarities),
+        )
+
+    def _should_check_for_intent(self, intent: Text, domain: Domain) -> bool:
+
+        if domain.intents.index(intent) not in self.intent_thresholds:
+            # This means the intent was never present in a story
+            return False
+        if intent in self.config[IGNORE_INTENTS_LIST]:
```
JEM-Mosig, line 373: Can make this an `elif`

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,492 @@
...
+            query_label = last_user_event.intent_name
+            query_label_id = domain.intents.index(query_label)
+            query_label_score = similarities[0][query_label_id]
+
+            logger.debug(f"Querying for intent {query_label}")
+
+            if self._should_check_for_intent(query_label, domain):
+
+                logger.debug(
+                    f"Score for user intent {query_label} likely to occur here is "
```
JEM-Mosig, line 415: Can we say "Probability for intent {query_label} to occur here is, ..."?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,492 @@
...
+                    f"{query_label_score}, while threshold is {self.intent_thresholds[query_label_id]}"
+                )
+                logger.debug(
+                    f"Top 5 intents(in ascending order) that are likely here are: {sorted_intent_scores[-5:]}"
+                )
+
+                # If score for query intent is below threshold and
+                # the query intent is not the top likely intent
```
JEM-Mosig, line 423: Ah, good point. I would have forgotten that.

JEM-Mosig (changes requested):
Looks good! Only minor comments, and we need to implement the `loss_layer` function for `TED` and assign it `SingleLabelDotProductLoss`.

rasa/utils/tensorflow/models.py
```diff
@@ -575,9 +575,13 @@ def _prepare_ffnn_layer(
...
     def _prepare_dot_product_loss(
-        self, name: Text, scale_loss: bool, prefix: Text = "loss"
+        self,
+        name: Text,
+        scale_loss: bool,
+        prefix: Text = "loss",
+        loss_layer: tf.keras.layers.Layer = layers.DotProductLoss,
     ) -> None:
-        self._tf_layers[f"{prefix}.{name}"] = layers.DotProductLoss(
+        self._tf_layers[f"{prefix}.{name}"] = loss_layer(
```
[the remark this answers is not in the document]
dakshvar22: Not sure if that's the correct way because that would make it mandatory for any class inheriting from `TransformerModel` to have a function named `_prepare_dot_product_loss`. Right now if we keep it this way, only those models which do not use `SingleLabelDotProductLoss` but a variant of it need to call the base class's function with the correct argument.

rasa/core/policies/intent_ted_policy.py
dakshvar22: It's not a probability. It's dot product similarity which is unbounded.

rasa/core/constants.py
```diff
@@ -20,21 +20,29 @@
 
 # the lowest priority intended to be used by machine learning policies
 DEFAULT_POLICY_PRIORITY = 1
+# the priority intended to be used by intent prediction policies.
+# This should be below all rule based policies but higher than ML
+# based policies. This enables a loop inside ensemble where if none
```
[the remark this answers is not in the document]
dakshvar22: Reworded it to explain the above as well. Reason I didn't use `user-prediction policies` was that we don't predict entities, only intent.

rasa/core/policies/intent_ted_policy.py
dakshvar22: Dictionary was easier because then the code does not rely on a particular order of intents in the threshold array. We still do rely on the order of intents from the domain to construct the intent label id though.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -224,8 +224,6 @@ def featurize_trackers(
                 f"to get numerical features for trackers."
             )
 
-        self.state_featurizer.prepare_for_training(domain, interpreter, bilou_tagging)
```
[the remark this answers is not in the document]
dakshvar22: It's needed but I have moved it to `_prepare_for_training` inside `TEDPolicy`. Reason being that we need the state featurizer to be prepared before label data is constructed inside `_create_label_data`.  But in the old code, the preparation of state featurizer was done inside `featurize_trackers` as a result of which there is an inherent assumption made in the code that state featurizer will be prepared by the time `_create_label_data` is called.

rasa/core/policies/ted_policy.py
```diff
@@ -323,9 +327,13 @@ def __init__(
         """Declare instance variables with default values."""
         self.split_entities_config = rasa.utils.train_utils.init_split_entities(
             kwargs.get(SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE),
-            self.defaults[SPLIT_ENTITIES_BY_COMMA],
+            self.defaults.get(
+                SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE
+            ),
```
alwx, line 332: Isn't it always `SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE`?

rasa/core/policies/ted_policy.py
```diff
@@ -347,22 +355,32 @@ def __init__(
...
-        self.data_example: Optional[Dict[Text, List[np.ndarray]]] = None
+        self.data_example: Optional[Dict[Text, Dict[Text, List[FeatureArray]]]] = None
 
         self.tmp_checkpoint_dir = None
         if self.config[CHECKPOINT_MODEL]:
             self.tmp_checkpoint_dir = Path(rasa.utils.io.create_temporary_directory())
 
+    @staticmethod
+    def model_class() -> Type[RasaModel]:
+        return TED
```
alwx, line 366: Curious why this method is needed. Is it gonna be used dynamically later? Now it simply returns a constant and this constant doesn't seem to be used somewhere else except by this particular class.

rasa/core/policies/ted_policy.py
```diff
@@ -590,6 +633,56 @@ def train(
...
+        self, model_data: RasaModelData, label_ids: np.ndarray
+    ) -> None:
+        """Run any tasks to be completed after the model has been trained.
+
+        Args:
+            model_data: Featurized data on which the model has been trained.
+            label_ids: Label ids corresponding to the featurized data.
+        """
+        # No post training procedure for `TEDPolicy`
+        return
```
alwx, line 646: But why do we define this method and even call it from `train` method?
`run_post_training_procedures` is not a part of `Policy` class specification (thus it doesn't need to be defined) so it seems like it can just be removed from here.

rasa/core/policies/ted_policy.py
```diff
@@ -590,6 +633,56 @@ def train(
...
+        if model_data.is_empty():
+            logger.error(
+                f"Can not train '{self.__class__.__name__}'. No data was provided. "
+                f"Skipping training of the policy."
+            )
+            return
+
+        self.run_training(model_data)
+
+        self.run_post_training_procedures(model_data, label_ids)
```
alwx, line 684: Again, looks like this call is not needed.

rasa/core/policies/ted_policy.py
```diff
@@ -347,22 +355,32 @@ def __init__(
...
         if self.config[CHECKPOINT_MODEL]:
             self.tmp_checkpoint_dir = Path(rasa.utils.io.create_temporary_directory())
 
+    @staticmethod
+    def model_class() -> Type[RasaModel]:
+        return TED
+
+    @classmethod
+    def _metadata_filename(cls) -> Optional[Text]:
+        return SAVE_MODEL_FILE_NAME
```
alwx, line 370: It might be similar to one of my previous comments but what's the reason of returning a constant from a class method? Especially since this class method is private.

alwx (changes requested):

alwx:
It's mostly questions here. I also haven't checked the `intent_ted_policy.py` itself yet, will do after lunch.

rasa/core/constants.py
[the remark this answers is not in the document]
JEM-Mosig: I thought so, but the same logic would also apply if we ever predicted entities in the future, and then we'd not need to rewrite the comment. Not really important.

rasa/core/policies/intent_ted_policy.py
JEM-Mosig: Right. But "user intent" is redundant. Just "intent" would do, and we don't need "likely".

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,688 @@
...
+            # be the first label id
+            first_pos_label_id = all_pos_labels[0]
+
+            if first_pos_label_id not in label_id_scores:
+                label_id_scores[first_pos_label_id] = []
+
+            label_id_scores[first_pos_label_id].append(
+                outputs["similarities"][index, 0, first_pos_label_id]
+            )
+        return label_id_scores
```
alwx, line 688: Good job! This one is actually clear and straightforward, especially after going through the `TEDPolicy` :)

rasa/core/policies/ted_policy.py
dakshvar22: Yes, inside `IntentTEDPolicy`

rasa/core/policies/ted_policy.py
dakshvar22: It's needed for `IntentTEDPolicy`. Any policy inheriting from `TEDPolicy` can have a post training procedure to run. `IntentTEDPolicy` does and that's why the call is needed.

rasa/core/constants.py
```diff
@@ -20,21 +20,35 @@
...
+# of the rule based policies predict an action and intent prediction
+# policy predicts one, it's prediction is chosen by the ensemble and
+# then the ML based policies are again run to get the prediction for
+# an actual action. To prevent an infinite loop, intent prediction
+# policies only predict an action if the last event in
+# the tracker is of type `UserUttered`. Hence, they make at most
+# one action prediction in each conversation turn. This allows other
+# policies to predict a winning action prediction.
+UNLIKELY_INTENT_POLICY_PRIORITY = 2
```
koernerfelicia, line 35: How come these aren't defined relative to the other constants?
```
DEFAULT_POLICY_PRIORITY = 1
UNLIKELY_INTENT_POLICY_PRIORITY = DEFAULT_POLICY_PRIORITY + 1
```

koernerfelicia: (and so on) I think it might help make things clearer, though I realise this is just my opinion.

dakshvar22: That's a very good point and I don't know why is it so but it makes sense to me too. 😅

rasa/core/policies/ted_policy.py
dakshvar22: It's an invalid parameter for `IntentTEDPolicy` and hence it wouldn't exist in the defaults of `IntentTEDPolicy` which is why this line will error out when instantiating `IntentTEDPolicy`.

rasa/core/policies/ted_policy.py
dakshvar22: Again, `_load_model_utilities` relies on it and `SAVE_MODEL_FILE_NAME` needs to be of the correct class.

rasa/core/constants.py
dakshvar22: Changed it.

dakshvar22:
@alwx I've answered your questions. Could you take another look and complete the review?

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -354,10 +354,33 @@ def _encode_action(
 
         return self._extract_state_features(action_as_sub_state, interpreter)
 
+    def encode_all_labels(
```
koernerfelicia, line 357: Just to check -- this is really just a re-naming of the old method (encode_all_actions)? Was this to be more accurate/descriptive?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,691 @@
...
+        # Max position for relative embeddings
+        MAX_RELATIVE_POSITION: None,
+        # Use a unidirectional or bidirectional encoder
+        # for `text`, `action_text`, and `label_action_text`.
+        UNIDIRECTIONAL_ENCODER: False,
+        # ## Training parameters
+        # Initial and final batch sizes:
+        # Batch size will be linearly increased for each epoch.
+        BATCH_SIZES: [64, 256],
+        # Strategy used whenc creating batches.
```
koernerfelicia, line 153:
```suggestion
        # Strategy used when creating batches.
```

rasa/core/featurizers/single_state_featurizer.py
dakshvar22: Yes, it's a renaming for the function inside `SingleStateFeaturizer`, so that a common function (encode_all_labels) can be invoked from the policy and not an implementation specific function (encode_all_intents / encode_all_actions) is invoked

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,691 @@
...
+        Args:
+            domain: Domain of the assistant.
+            similarities: Predicted similarities for each intent.
+
+        Returns:
+            Metadata to be attached.
+        """
+        metadata = {}
+        for intent in domain.intents:
+            intent_index = domain.intents.index(intent)
```
koernerfelicia, line 356:
```suggestion
        for intent_index, intent in enumerate(domain.intents):
```
Unless we have some reason to believe the order might change?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,691 @@
...
+        """
+        if domain.intents.index(intent) not in self.label_thresholds:
+            # This means the intent was never present in a story
+            logger.debug(
+                f"Query intent index {domain.intents.index(intent)} not "
+                f"found in label thresholds - {self.label_thresholds}"
+            )
+            return False
+        if intent in self.config[IGNORE_INTENTS_LIST]:
+            return False
```
koernerfelicia, line 445: Shouldn't this check occur first, to avoid showing the above debug message for an ignored intent?

rasa/core/policies/ted_policy.py
JEM-Mosig: `IntentTEDPolicy` is derived from `TEDPolicy` and re-uses the `train` method, but only overwrites this `run_post_training_procedures`. Makes sense to me, but it could also be done by doing a `super` call followed by `run_post_training_procedures` inside `IntentTEDPolicy`.

rasa/core/policies/ted_policy.py
```diff
@@ -773,7 +866,8 @@ def persist(self, path: Union[Text, Path]) -> None:
             return
 
         model_path = Path(path)
-        tf_model_file = model_path / f"{SAVE_MODEL_FILE_NAME}.tf_model"
+        metadata_filename = self._metadata_filename()
```
JEM-Mosig, line 869: Why do we call this "metadata"? To me that sounds like "date of training" or something, not its weights or unlikely-intent-thresholds.

tests/core/policies/test_intent_ted_policy.py
```diff
@@ -0,0 +1,382 @@
...
+            training_trackers, default_domain, interpreter
+        )
+
+        data_for_prediction = trained_policy._prepare_data_for_prediction(
+            training_model_data
+        )
+
+        assert all(
+            [key in PREDICTION_FEATURES for key in data_for_prediction.data.keys()]
+        )
```
JEM-Mosig, line 113: I think you can write this as `data_for_prediction.data.keys() <= PREDICTION_FEATURES` or `data_for_prediction.data.keys().issubset(PREDICTION_FEATURES)`.

rasa/core/policies/intent_ted_policy.py
dakshvar22: Hmm, I think the earlier debug message is necessary just to communicate which intent is being queried. However, I think it would be better if I add another debug message inside the above `if` construct, something along the lines of - 
`Query intent {intent} found in {IGNORE_INTENTS_LIST}. Check for `action_unlikely_intent` prediction will be skipped.` What do you think?

rasa/core/policies/ted_policy.py
```diff
@@ -347,22 +355,32 @@ def __init__(
...
         self.config = rasa.utils.train_utils.override_defaults(
             self.defaults, new_config
         )
 
-        self.config = rasa.utils.train_utils.update_confidence_type(self.config)
+        self._auto_update_configuration()
 
+    def _auto_update_configuration(self) -> None:
+        """Take care of deprecations and compatibility of parameters."""
+        self.config = rasa.utils.train_utils.update_confidence_type(self.config)
```
koernerfelicia, line 382: Why does this happen before `rasa.utils.train_utils.validate_configuration_settings(self.config)`?

rasa/core/policies/ted_policy.py
```diff
@@ -375,12 +393,33 @@ def _create_label_data(
...
+        Args:
+            attribute_data: Feature data for all labels.
+            domain: Domain of the assistant.
+
+        Returns:
+            Features of labels ready to be fed to the model.
+        """
         label_data = RasaModelData()
         label_data.add_data(attribute_data, key_prefix=f"{LABEL_KEY}_")
         label_data.add_lengths(
```
koernerfelicia, line 425: I'm a bit confused about what's going on here, the docstring says it should contain `label_action_name` but I only see `label_action_text` being added?

rasa/core/policies/ted_policy.py
koernerfelicia: +1

tests/core/policies/test_ted_policy.py
```diff
@@ -130,6 +136,34 @@ def test_normalization(
...
+
+        attribute_data, _ = model_data_utils.convert_to_data_format(encoded_all_labels)
+        assembled_label_data = trained_policy._assemble_label_data(
+            attribute_data, default_domain
+        )
+        assembled_label_data_signature = assembled_label_data.get_signature()
+
+        assert list(assembled_label_data_signature.keys()) == [
+            f"{LABEL}_{ACTION_NAME}",
+            f"{LABEL}",
```
koernerfelicia, line 155: Again, I thought based on the docstring we expect `label_action_text` as well?

rasa/core/policies/intent_ted_policy.py
koernerfelicia: Wouldn't you never get this second message? Because it would return `False` beforehand?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,691 @@
...
+        Returns:
+            Similarities grouped by intent label id.
+        """
+        label_id_scores = {}
+        for index, all_pos_labels in enumerate(label_ids):
+
+            # Take the contribution of only first label id
+            # because `outputs` will contain another
+            # data point where remaining label ids will
+            # be the first label id
```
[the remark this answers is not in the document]
koernerfelicia: Yes please, I think I am not sure what's going on here, either.

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,690 @@
...
+        """
+        self._training = False
+
+        batch_size = (
+            self.config[BATCH_SIZES]
+            if isinstance(self.config[BATCH_SIZES], int)
+            else self.config[BATCH_SIZES][0]
+        )
+        outputs = self.run_inference(
+            model_data, batch_size=batch_size, output_keys_expected=["similarities"]
```
koernerfelicia, line 632: Should this be a constant?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,690 @@
...
+                for intent_label, score in predicted_intent_scores.items()
+            ],
+            key=lambda x: x[1],
+        )
+        query_intent_id = domain.intents.index(query_intent)
+        query_intent_similarity = similarities[0][query_intent_id]
+
+        logger.debug(f"Querying for intent {query_intent}")
+
+        if self._should_check_for_intent(query_intent, domain):
```
koernerfelicia, line 481: How come this doesn't happen earlier?

rasa/core/policies/ted_policy.py
```diff
@@ -534,24 +577,26 @@ def train(
         model_data = self._create_model_data(
             tracker_state_features, label_ids, entity_tags, encoded_all_labels
         )
-        if model_data.is_empty():
-            logger.error(
-                f"Can not train '{self.__class__.__name__}'. No data was provided. "
-                f"Skipping training of the policy."
-            )
-            return
```
koernerfelicia, line 542: Why do we no longer need this?

koernerfelicia: Oh actually, it was moved into `train`, yes?

rasa/core/policies/ted_policy.py
dakshvar22: I can make a `super()` call as well if that's preferred by both of you.

rasa/core/policies/intent_ted_policy.py
dakshvar22: I meant adding the above debug message before `return False` but inside `if intent in self.config[IGNORE_INTENTS_LIST]:`

rasa/core/policies/intent_ted_policy.py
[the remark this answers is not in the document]
dakshvar22: Rephrased it.

rasa/core/policies/ted_policy.py
```diff
@@ -323,9 +327,13 @@ def __init__(
...
         self.split_entities_config = rasa.utils.train_utils.init_split_entities(
             kwargs.get(SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE),
-            self.defaults[SPLIT_ENTITIES_BY_COMMA],
+            self.defaults.get(
+                SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE
+            ),
         )
 
+        # TODO: check if the else statement can be removed.
+        #  More context here - https://github.com/RasaHQ/rasa/issues/5786#issuecomment-840762751
```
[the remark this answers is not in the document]
dakshvar22: It's a Todo from other part of the codebase and not necessarily for this feature. Once enable has closed this issue, we should refactor the code here.
I think it makes sense to request enable to let us know once the issue is closed.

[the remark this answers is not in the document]
dakshvar22: I think we should create a separate issue for the TODO, but place it in the blocked column. It will be revisited as part of the refinement meeting then.

rasa/core/policies/ted_policy.py
dakshvar22: Is `model_file_name` okay?

rasa/core/policies/ted_policy.py
dakshvar22: It's because of deprecations we need to do it this way. For the scope of this PR, i just refactored it into a separate method and didn't change any order.
Cleaning this up is something for 3.0

rasa/core/policies/ted_policy.py
dakshvar22: `attribute_data` will contain `label_action_name` and `label_action_text`. Hence, when `add_data` is called, they should be present in `label_data`.

tests/core/policies/test_ted_policy.py
dakshvar22: My reply on the docstring comment should help clarify this further.

rasa/core/policies/intent_ted_policy.py
dakshvar22: It wasn't before in `TEDPolicy` as well, so I've left it for now. Should be done outside of the PR IMO.

rasa/core/policies/intent_ted_policy.py
dakshvar22: Where would you like to place it?

rasa/core/policies/ted_policy.py
dakshvar22: Yes.

tests/core/policies/test_intent_ted_policy.py
```diff
@@ -0,0 +1,380 @@
...
+import tensorflow as tf
+
+import numpy as np
+import pytest
+from _pytest.monkeypatch import MonkeyPatch
+from rasa.core.featurizers.single_state_featurizer import (
+    IntentTokenizerSingleStateFeaturizer,
+)
+from rasa.core.featurizers.tracker_featurizers import (
+    IntentMaxHistoryTrackerFeaturizer,
```
alwx, line 13: I noticed that this `IntentMaxHistoryTrackerFeaturizer` doesn't exist. Is it something that's done separately?

dakshvar22: Yes, that's being implemented [here](https://github.com/RasaHQ/rasa/pull/8570)

rasa/core/policies/intent_ted_policy.py
koernerfelicia: Yes, I meant even higher up than that. If the intent is in `IGNORE_INTENTS_LIST` wouldn't this:
`if domain.intents.index(intent) not in self.label_thresholds:`
be true?

rasa/core/policies/intent_ted_policy.py
koernerfelicia: I would put the below at the very beginning of the method. Otherwise you're creating  variables like `sorted_intent_scores` even if they may not be needed.
```
if self._should_check_for_intent(query_intent, domain):
    return False
```

rasa/core/policies/intent_ted_policy.py
koernerfelicia: fair.

rasa/core/policies/ted_policy.py
koernerfelicia: Fine by me

rasa/core/policies/ted_policy.py
koernerfelicia: Thank you!

alwx (approved):
Approving it but this one https://github.com/RasaHQ/rasa/pull/8570 needs to be merged first in order for it to work.

rasa/core/policies/intent_ted_policy.py
JEM-Mosig: You mean `if not ...`, right?

koernerfelicia: Yep, oops

rasa/core/policies/ted_policy.py
JEM-Mosig: I think `super` would be nicer.

rasa/core/policies/ted_policy.py
```diff
@@ -783,78 +878,62 @@ def persist(self, path: Union[Text, Path]) -> None:
...
+        """
+        metadata_filename = self._metadata_filename()
         io_utils.json_pickle(
-            model_path / f"{SAVE_MODEL_FILE_NAME}.priority.pkl", self.priority
+            model_path / f"{metadata_filename}.priority.pkl", self.priority
         )
+        io_utils.pickle_dump(model_path / f"{metadata_filename}.meta.pkl", self.config)
         io_utils.pickle_dump(
-            model_path / f"{SAVE_MODEL_FILE_NAME}.meta.pkl", self.config
+            model_path / f"{metadata_filename}.data_example.pkl", self.data_example,
```
JEM-Mosig, line 895: Similar to my comment on IntentTED, we could make these private properties of the class, so it's ensured that they are exactly the same as in the load methods.

JEM-Mosig (approved):
A few more minor things, but looks good!

rasa/core/policies/ted_policy.py
dakshvar22: The problem with call to `super` is that `train` method of `IntentTEDPolicy` won't have access to `model_data` and `label_ids` which are needed for running the post training procedures. It will have to construct the two on its own resulting in duplicate code.

dakshvar22: Or else, I can make it part of the `run_training` method. I think this would be better. Will do it now.

rasa/core/policies/intent_ted_policy.py
dakshvar22: Nope. `label_thresholds` are learnt during training for all intents that are seen across training stories.
`IGNORE_INTENTS_LIST` is primarily meant to be used at inference time where users can set it to a subset of intents for which they don't want to trigger `action_unlikely_intent`.

A follow up question from your side could be - "Can we even change the component's configuration parameters without retraining the pipeline?" Answer: Not currently, but something to be prepared for in 3.0 . Once that's possible, users can change `IGNORE_INTENTS_LIST` and play with the policy to get the best set of intents for which they want to trigger `action_unlikely_intent` without having to retrain the pipeline.
Does that help clarify?

rasa/core/policies/intent_ted_policy.py
dakshvar22: Good point! Refactored it to have the check at the top.

rasa/core/policies/intent_ted_policy.py
```diff
@@ -0,0 +1,693 @@
...
+        self._prepare_dot_product_loss(
+            LABEL, self.config[SCALE_LOSS], loss_layer=layers.MultiLabelDotProductLoss
+        )
+
+    def _prepare_dot_product_loss(
+        self,
+        name: Text,
+        scale_loss: bool,
+        prefix: Text = "loss",
+        loss_layer: tf.keras.layers.Layer = layers.SingleLabelDotProductLoss,
```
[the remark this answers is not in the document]
dakshvar22: `MultiLabelDotProductLoss` is being passed as the argument value in `_prepare_label_classification_layers`. Needed the default value to match the function signature with its counterpart in `TED`.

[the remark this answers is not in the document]
JEM-Mosig: But you could use a class variable `self.DEFAULT_LOSS` or something and use that as default, no? Would be cleaner since IntentTED really should use the MultiLabel loss.

JEM-Mosig (approved):
One last thing. 😬

rasa/core/policies/intent_ted_policy.py
koernerfelicia: Yes, thank you!

koernerfelicia (approved):
Great work! Thanks for taking the time to answer all my questions (even the tangential ones)

PR #8558, opened 2021-04-27
Fix matplotlib backend in decorator

m-vdb:
**Proposed changes**:
Fixes https://github.com/RasaHQ/rasa/issues/8545
Or rather mitigates it. People with falsy config encounter this issue when running virtually any rasa command (some stacktraces occurred for `rasa init` for instance). This moves matplotlib backend manipulation in a decorator, instead of at the module level today.

### DIFF
--- a/rasa/utils/plotting.py
+++ b/rasa/utils/plotting.py
@@ -16,9 +16,19 @@ logger = logging.getLogger(__name__)
 
 def _fix_matplotlib_backend() -> None:
     """Tries to fix a broken matplotlib backend."""
+    try:
+        backend = matplotlib.get_backend()
+    except Exception:  # skipcq:PYL-W0703
+        logger.error(
+            "Cannot retrieve Matplotlib backend, likely due to a compatibility "
+            "issue with system dependencies. Please refer to the documentation: "
+            "https://matplotlib.org/stable/tutorials/introductory/usage.html#backends"
+        )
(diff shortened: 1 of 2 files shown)


rasa/utils/plotting.py
```diff
@@ -14,10 +15,20 @@
...
+    """Tries to fix a broken matplotlib backend."""
+    try:
+        backend = matplotlib.get_backend()
+    except Exception:  # skipcq:PYL-W0703
+        logger.error(
+            "Cannot retrieve Matplotlib backend, likely due to a compatibility "
+            "issue with system dependencies. Please refer to the documentation: "
+            "https://matplotlib.org/stable/tutorials/introductory/usage.html#backends"
+        )
+        raise
```
m-vdb, line 27: should we even raise a RasaException here?

wochinge: Good question 🤔 I'd prefer to keep visibility about dependency issues to be honest - even this one. I'd rather ignore this one within Sentry.

wochinge (approved):
Thanks for taking this on! 

* Can you please also add a changelog? 
* can you please target 2.5.x as it's a bug fix?

rasa/utils/plotting.py
m-vdb: I agree!

PR #8560, opened 2021-04-27
Implement interface for bulk inferencing in TF models

dakshvar22:
**Proposed changes**:
- All models now use `run_inference` method to generate predictions through the model. `run_inference` is meant to perform batch inferencing as well which means that it implements the batching and combining of output for different batches. The specific TF models like `TED`, `DIET` do not need to know the implementation details of this batch inferencing.
- Needed for future features like `IntentTEDPolicy`.

### DIFF
--- a/rasa/utils/tensorflow/models.py
+++ b/rasa/utils/tensorflow/models.py
@@ -304,7 +304,7 @@ class RasaModel(TmpKerasModel):
         all_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
         batch_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
     ) -> Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]:
-        """Merge a batch's output into the output for all batches.
+        """Merges a batch's output into the output for all batches.
 
         Function assumes that the schema of batch output remains the same,
         i.e. keys and their value types do not change from one batch's
--- a/rasa/utils/tensorflow/models.py
+++ b/rasa/utils/tensorflow/models.py
@@ -256,7 +256,7 @@ class RasaModel(TmpKerasModel):
         """Custom prediction method that builds tf graph on the first call.
 
         Args:
-            batch_in: Prepared batch ready for input to predict_step method of model.
+            batch_in: Prepared batch ready for input to `predict_step` method of model.
 
         Return:
             Prediction output, including diagnostic data.
--- a/rasa/utils/tensorflow/models.py
+++ b/rasa/utils/tensorflow/models.py
@@ -62,7 +62,8 @@ def _merge_batch_outputs(
         batch_output: Output for a batch.
 
     Returns:
-        Merged output with the output for current batch stacked below the output for all previous batches.
+        Merged output with the output for current batch stacked
+        below the output for all previous batches.
     """
     if not all_outputs:
         return batch_output
@@ -318,7 +319,8 @@ class RasaModel(TmpKerasModel):
         while True:
             try:
                 # data_generator is a tuple of 2 elements - input and output.
-                # We only need input, since output is always None and not consumed by our TF graphs.
+                # We only need input, since output is always None and not
+                # consumed by our TF graphs.
                 batch_in = next(data_iterator)[0]
                 batch_out = self.rasa_predict(batch_in)
                 outputs = _merge_batch_outputs(outputs, batch_out)
--- a/rasa/utils/tensorflow/models.py
+++ b/rasa/utils/tensorflow/models.py
@@ -307,7 +307,8 @@ class RasaModel(TmpKerasModel):
         """Merge a batch's output into the output for all batches.
 
         Function assumes that the schema of batch output remains the same,
-        i.e. keys and their value types do not change from one batch's output to another.
+        i.e. keys and their value types do not change from one batch's
+        output to another.
 
         Args:
             all_outputs: Existing output for all previous batches.
--- a/rasa/utils/tensorflow/models.py
+++ b/rasa/utils/tensorflow/models.py
@@ -52,6 +52,18 @@ def _merge_batch_outputs(
     all_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
     batch_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
 ) -> Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]:
+    """Merge a batch's output into the output for all batches.
+
+    Function assumes that the schema of batch output remains the same,
+    i.e. keys and their value types do not change from one batch's output to another.
+
+    Args:
+        all_outputs: Existing output for all previous batches.
+        batch_output: Output for a batch.
+
+    Returns:
+        Merged output with the output for current batch stacked below the output for all previous batches.
+    """
     if not all_outputs:
         return batch_output
     for key, val in batch_output.items():
@@ -305,7 +317,8 @@ class RasaModel(TmpKerasModel):
         data_iterator = iter(data_generator)
         while True:
             try:
-                # Only want x, since y is always None out of our data generators
+                # data_generator is a tuple of 2 elements - input and output.
+                # We only need input, since output is always None and not consumed by our TF graphs.
                 batch_in = next(data_iterator)[0]
                 batch_out = self.rasa_predict(batch_in)
                 outputs = _merge_batch_outputs(outputs, batch_out)
--- a/rasa/utils/tensorflow/models.py
+++ b/rasa/utils/tensorflow/models.py
@@ -35,6 +35,7 @@ from rasa.utils.tensorflow.constants import (
     CONSTRAIN_SIMILARITIES,
     MODEL_CONFIDENCE,
 )
+import rasa.utils.train_utils
 from rasa.utils.tensorflow import layers
 from rasa.utils.tensorflow import rasa_layers
 from rasa.utils.tensorflow.temp_keras_modules import TmpKerasModel
@@ -231,12 +232,12 @@ class RasaModel(TmpKerasModel):
         return [element_spec]
 
     def rasa_predict(
-        self, model_data: RasaModelData
+        self, batch_in: Tuple[np.ndarray]
     ) -> Dict[Text, Union[np.ndarray, Dict[Text, Any]]]:
         """Custom prediction method that builds tf graph on the first call.
 
         Args:
-            model_data: The model data to use for prediction.
+            batch_in: Prepared batch ready for input to predict_step method of model.
 
         Return:
             Prediction output, including diagnostic data.
@@ -247,8 +248,6 @@ class RasaModel(TmpKerasModel):
             # after training, we need to prepare the model for prediction once
             self.prepare_for_predict()
             self.prepared_for_prediction = True
-
-        batch_in = RasaBatchDataGenerator.prepare_batch(model_data.data)
 
         if self._run_eagerly:
             outputs = tf_utils.to_numpy_or_python_type(self.predict_step(batch_in))
@@ -267,6 +266,48 @@ class RasaModel(TmpKerasModel):
             outputs[DIAGNOSTIC_DATA]
         )
         return outputs
+
+    def run_inference(
+        self, data_generator: RasaBatchDataGenerator
+    ) -> Dict[Text, Union[np.ndarray, Dict[Text, Any]]]:
+        """
+
+        Args:
+            data_generator:
+
+        Returns:
+
+        """
+        outputs = {}
+        data_iterator = iter(data_generator)
+        while True:
+            try:
+                # Only want x, since y is always None out of our data generators
+                batch_in = next(data_iterator)[0]
+                batch_out = self.rasa_predict(batch_in)
+                outputs = self._merge_batch_outputs(outputs, batch_out)
+            except StopIteration:
+                # Generator ran out of batches, time to finish inferencing
+                break
+        return outputs
+
+    def _merge_batch_outputs(
+        self,
+        all_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+        batch_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+    ) -> Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]:
+        if not all_outputs:
+            return batch_output
+        for key, val in batch_output.items():
+            if isinstance(val, np.ndarray):
+                all_outputs[key] = np.concatenate(
+                    [all_outputs[key], batch_output[key]], axis=0
+                )
+            elif isinstance(val, dict):
+                # recurse and merge the inner dict first
+                all_outputs[key] = self._merge_batch_outputs(all_outputs[key], val)
+
+        return batch_output
 
     @staticmethod
     def _empty_lists_to_none_in_dict(input_dict: Dict[Text, Any]) -> Dict[Text, Any]:
@@ -339,7 +380,10 @@ class RasaModel(TmpKerasModel):
         # predict on one data example to speed up prediction during inference
         # the first prediction always takes a bit longer to trace tf function
         if not finetune_mode and predict_data_example:
-            model.rasa_predict(predict_data_example)
+            (data_generator, _,) = rasa.utils.train_utils.create_data_generators(
+                model_data=predict_data_example, batch_sizes=1, epochs=1, shuffle=False
+            )
+            model.run_inference(data_generator)
 
         logger.debug("Finished loading the model.")
         return model
--- a/rasa/utils/tensorflow/models.py
+++ b/rasa/utils/tensorflow/models.py
@@ -48,6 +48,25 @@ from tensorflow.python.keras.utils import tf_utils
 logger = logging.getLogger(__name__)
 
 
+def _merge_batch_outputs(
+    all_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+    batch_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+) -> Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]:
+    if not all_outputs:
+        return batch_output
+    for key, val in batch_output.items():
+        if isinstance(val, np.ndarray):
+            all_outputs[key] = np.concatenate(
+                [all_outputs[key], batch_output[key]], axis=0
+            )
+
+        elif isinstance(val, dict):
+            # recurse and merge the inner dict first
+            all_outputs[key] = _merge_batch_outputs(all_outputs[key], val)
+
+    return all_outputs
+
+
 # noinspection PyMethodOverriding
 class RasaModel(TmpKerasModel):
     """Abstract custom Keras model.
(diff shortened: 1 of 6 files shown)


dakshvar22:
/modeltest

```yml
include:
 - dataset: ["all"]
   config: ["all"]

```

samsucik:
@dakshvar22 I can see [failed training on Sara](https://github.com/RasaHQ/rasa/pull/8560/checks?check_run_id=2452501303) with
```
    ...
    ...
    class RasaTrainingLogger(tf.keras.callbacks.Callback):
AttributeError: module 'tensorflow' has no attribute 'keras'
```
Is this something to be fixed in this PR, or is it unrelated?

dakshvar22:
I don't think that's related but weird that it fails specifically on that pair of dataset and config.
Let me rerun that configuration once more

dakshvar22:
@samsucik Okay it ran successfully this time.

rasa/utils/tensorflow/models.py
```diff
@@ -47,6 +48,25 @@
 logger = logging.getLogger(__name__)
 
 
+def _merge_batch_outputs(
+    all_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+    batch_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+) -> Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]:
+    if not all_outputs:
```
samsucik, line 55: Could you add a simple docstring? I'm hesitant to agree that the function is ["very short" and also "obvious"](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods).

rasa/utils/tensorflow/models.py
```diff
@@ -231,12 +251,12 @@ def _dynamic_signature(
...
 
     def rasa_predict(
-        self, model_data: RasaModelData
+        self, batch_in: Tuple[np.ndarray]
     ) -> Dict[Text, Union[np.ndarray, Dict[Text, Any]]]:
         """Custom prediction method that builds tf graph on the first call.
 
         Args:
-            model_data: The model data to use for prediction.
+            batch_in: Prepared batch ready for input to predict_step method of model.
```
samsucik, line 259:
```suggestion
            batch_in: Prepared batch ready for input to `predict_step` method of model.
```

rasa/utils/tensorflow/models.py
```diff
@@ -231,12 +251,12 @@ def _dynamic_signature(
         return [element_spec]
 
     def rasa_predict(
-        self, model_data: RasaModelData
+        self, batch_in: Tuple[np.ndarray]
```
samsucik, line 254: `Tuple[np.ndarray]` stands for a tuple containing [exactly one `np.ndarray`](https://stackoverflow.com/a/39458225/5157159), but I just noticed that we have this mistake at many places around the model code (in the context of batches), so I'm not sure what's the best thing to do -- to change it everywhere now or to create an issue...

Here, I assume that `batch_in` can contain any number of `np.ndarray`s, requiring this kind of annotation:
```suggestion
        self, batch_in: Tuple[np.ndarray, ...]
```

rasa/utils/tensorflow/models.py
```diff
@@ -268,6 +286,34 @@ def rasa_predict(
...
+            Model outputs corresponding to the inputs fed.
+        """
+        outputs = {}
+        (data_generator, _,) = rasa.utils.train_utils.create_data_generators(
+            model_data=model_data, batch_sizes=batch_size, epochs=1, shuffle=False,
+        )
+        data_iterator = iter(data_generator)
+        while True:
+            try:
+                # Only want x, since y is always None out of our data generators
```
samsucik, line 308: Could you make this comment a bit clearer? The thing I didn't understand when I read it was the x vs y thing. A generator always gives an (x, y) pair, where x are inputs and y are labels?

rasa/utils/train_utils.py
```diff
@@ -392,6 +393,7 @@ def create_data_generators(
         batch_strategy: The batch strategy to use.
         eval_num_examples: Number of examples to use for validation data.
         random_seed: The random seed.
+        shuffle: Whether to shuffle data inside the data generator
```
samsucik, line 396:
```suggestion
        shuffle: Whether to shuffle data inside the data generator.
```

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+            {"a": np.array([5, 6]), "b": {"c": np.array([2, 4])}},
+            {"a": np.array([1, 2, 5, 6]), "b": {"c": np.array([3, 1, 2, 4])}},
+        ),
+    ],
+)
+def test_merging_batch_outputs(
+    existing_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+    new_batch_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+    expected_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+):
```
samsucik, line 36:
```suggestion
): -> None
```

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+
+    predicted_output = _merge_batch_outputs(existing_outputs, new_batch_outputs)
+
+    def test_equal_dicts(
+        dict1: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+        dict2: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+    ):
```
samsucik, line 43:
```suggestion
    ): -> None
```

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+
+@pytest.mark.parametrize(
+    "batch_size, number_of_data_points, expected_number_of_batch_iterations",
+    [(2, 3, 2), (1, 3, 3), (5, 3, 1),],
+)
+def test_batch_inference(
+    batch_size: int,
+    number_of_data_points: int,
+    expected_number_of_batch_iterations: int,
+):
```
samsucik, line 67:
```suggestion
): -> None
```

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+    model = RasaModel()
+
+    def batch_predict(batch_in: np.ndarray):
```
samsucik, line 70:
```suggestion
    def _batch_predict(batch_in: np.ndarray) -> Dict[Text, Union[tf.Tensor, Dict[Text, tf.Tensor]]]::
```

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+
+        dummy_output = batch_in[0]
+        output = {
+            "dummy_output": dummy_output,
+            DIAGNOSTIC_DATA: tf.constant(np.array([[1, 2]]), dtype=tf.int32),
+        }
+        return output
+
+    # Monkeypatch batch predict so that run_inference interface can be tested
+    model.batch_predict = batch_predict
```
samsucik, line 80:
```suggestion
    model.batch_predict = _batch_predict
```

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Create dummy model data to pass to model
+    model_data = RasaModelData(
+        label_key="label",
+        label_sub_key="ids",
+        data={
+            "text": {
+                "sentence": [
```
samsucik, line 88: Could you, please, replace the strings with the constants (e.g. `LABEL_IDS`) used in the model code?

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Firstly, the number of data points in dummy_output should be equal
+    # to the number of data points sent as input.
+    assert output["dummy_output"].shape[0] == number_of_data_points
+
+    # Secondly, the number of data points inside diagnostic_data should be
+    # equal to the number of batches passed to the model because for every
+    # batch passed as input, it would have created a
+    # corresponding diagnostic data entry.
+    assert output[DIAGNOSTIC_DATA].shape == (expected_number_of_batch_iterations, 2)
```
samsucik, line 107: Maybe this is nitpicking, but in the real model code `output[DIAGNOSTIC_DATA]` is always a dict (usually containing numerical arrays somewhere down the line), not a numerical object itself. Perhaps there's value in changing the diagnostic data in the monkey-patched `_batch_predict` to be closer to reality in this sense. What do you think?

samsucik (changes requested):
Thanks, Daksh, especially for adding the tests! A few tiny things may require changes, but nothing serious as far as I can see 🙂

rasa/utils/tensorflow/models.py
```diff
@@ -47,6 +48,25 @@
...
+        if isinstance(val, np.ndarray):
+            all_outputs[key] = np.concatenate(
+                [all_outputs[key], batch_output[key]], axis=0
+            )
+
+        elif isinstance(val, dict):
+            # recurse and merge the inner dict first
+            all_outputs[key] = _merge_batch_outputs(all_outputs[key], val)
+
+    return all_outputs
```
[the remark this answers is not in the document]
dakshvar22: You can't really create it as a static method as well because it is recursive and hence needs `self` argument.

rasa/utils/tensorflow/models.py
dakshvar22: Hmm, good point. I didn't know about this. I would lean towards creating a separate issue so that we can do it across the codebase.

tests/utils/tensorflow/test_models.py
dakshvar22: Pytests are not expected to have a return type because they always return `None`

tests/utils/tensorflow/test_models.py
dakshvar22: Same as above.

tests/utils/tensorflow/test_models.py
dakshvar22: I think that's a bit of an overkill because `test_merging_batch_outputs` already tests for the nested dict structure you mentioned. I actually wanted to not use `DIAGNOSTIC_DATA` as one of the keys of the output of `batch_predict` but then `rasa_predict` does something specific for `DIAGNOSTIC_DATA` and hence had to include it. Lmk what you think.

rasa/utils/tensorflow/models.py
```diff
@@ -47,6 +48,25 @@
...
+        return batch_output
+    for key, val in batch_output.items():
+        if isinstance(val, np.ndarray):
+            all_outputs[key] = np.concatenate(
+                [all_outputs[key], batch_output[key]], axis=0
+            )
+
+        elif isinstance(val, dict):
+            # recurse and merge the inner dict first
+            all_outputs[key] = _merge_batch_outputs(all_outputs[key], val)
```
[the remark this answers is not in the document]
dakshvar22: > what about turning this into a recursive inner helper function and leaving the rest as non-recursive?

Not sure what you mean here 🙈 

> from the "outside world", I don't have to know that it's recursive

Why do you have to know about it now as well?

>its type annotation also simplifies to Dict[Text, Dict[Text, np.ndarray]], right?

I think this is linked to your first observation which I didn't understand. Could you elaborate?

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
dakshvar22: Also, keeping it inside `RasaModel` felt unusual because it's more of a utility method which can be tested alone as well. Keeping it inside `RasaModel` means that we always invoke it as `rasa_model._merge_batch_outputs`.
What do you think?

rasa/utils/tensorflow/models.py
samsucik: Fair. I'll create an issue.

tests/utils/tensorflow/test_models.py
samsucik: Ah, I see! You made me realise that my PR actually [added code in `rasa_predict` which now effectively requires `DIAGNOSTIC_DATA` to be present](https://github.com/RasaHQ/rasa/commit/ca66e3494524d2ce3a30c3b815da27c4d0a558d8#diff-1b0d53e90a7f1dcca1113cda22c58c98685004f8dfc7ba3d4e5bf9cee78e6a19R255). That's a mistake I guess, and should be fixed. Then you won't have to use `DIAGNOSTIC_DATA` in your test 🙂  Sounds good?

dakshvar22: Sounds good. Do you have a fix for it in mind already?

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
samsucik: So, what I mean is: As someone merely calling the function, I don't care whether it's internally implemented as recursive or not. But I think the current type annotation is complicated as a result of reflecting the particulars of the implementation. I have in mind something like this, which hides the recursiveness:
```python
def _merge_batch_outputs(
    all_outputs: Dict[Text, Dict[Text, np.ndarray]],
    batch_output: Dict[Text, Dict[Text, np.ndarray]],
) -> Dict[Text, Dict[Text, np.ndarray]]:
    if not all_outputs:
        return batch_output

    def _recurse(
        _all_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]], 
        _batch_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]
    ) -> Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]:
        for key, val in _batch_output.items():
            if isinstance(val, np.ndarray):
                _all_outputs[key] = np.concatenate(
                    [_all_outputs[key], _batch_output[key]], axis=0
                )

            elif isinstance(val, dict):
                # recurse and merge the inner dict first
                _all_outputs[key] = _recurse(all_outputs[key], val)

        return _all_outputs

    return _recurse(all_outputs, batch_output)
```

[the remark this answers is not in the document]
dakshvar22: But the type annotation of `all_outputs` in your method will be wrong then. The passed input will be `Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]`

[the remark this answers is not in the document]
dakshvar22: Also, the recursion of `_recurse` method will not terminate I think.

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
samsucik: I think `RasaModel` currently contains a lot of these utility methods -- because they're specifically used by the class. After all, the method is tailored specifically for that class and can't be re-used in some other implementation of a model. So, including this method in `RasaModel` wouldn't break a pattern. And I think there's nothing wrong about calling it as `self._merge_batch_outputs` (from within `RasaModel` even when it's static) and as `RasaModel._merge_batch_outputs` from outside (when actually needed, currently only in tests).

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
samsucik: Ah, that's what I wasn't sure about. I thought the `Dict[Text, np.ndarray]` case was added only because of the recursive calls.

tests/utils/tensorflow/test_models.py
samsucik: Yeah, the lines I added over there have to be conditioned on `DIAGNOSTIC_DATA` being present in the outputs dict. Simple 🙂  If you think it can go in your PR, feel free to make a fix; otherwise, I'll create a mini-PR.

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
dakshvar22: The complete annotation comes because we expect the batch outputs to contain nested dicts for 'some' keys only. This structure will get reflected in the type annotation of the arguments anyhow.

> As someone merely calling the function, I don't care whether it's internally implemented as recursive or not

I still don't get how the caller needs to know about the recursion :) Making a call like `rasa_model._merge_batch_outputs(all_outputs, batch_output) does not reveal anything about the implementation, no?

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
dakshvar22: > After all, the method is tailored specifically for that class and can't be re-used in some other implementation of a model.

👍  I'll add it in.

> And I think there's nothing wrong about calling it as self._merge_batch_outputs (from within RasaModel even when it's static)

There is no `self` variable if the method is static.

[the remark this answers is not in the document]
dakshvar22: Okay, I changed the recursive call to `RasaModel._merge_batch_outputs`.

rasa/utils/tensorflow/models.py
samsucik: The issue is [here](https://github.com/RasaHQ/research/issues/150).

tests/utils/tensorflow/test_models.py
dakshvar22: Done :+1

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
samsucik: > There is no self variable if the method is static.

Ah, I get you now. I meant when calling the method from within the model, not from within itself. You're right.

dakshvar22:
@samsucik I also made `rasa_predict` private now because there should be only one public method for running inference ideally and that can be `run_inference` now.

rasa/utils/tensorflow/models.py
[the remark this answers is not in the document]
samsucik: Yeah, it was all about me assuming that the annotation could be simplified if we didn't call the function from within itself. My bad. Basically, I thought that the way the function was implemented was affecting its annotation (which would've been undesirable).

tests/utils/tensorflow/test_models.py
samsucik: Awesome, thanks!

rasa/utils/tensorflow/models.py
```diff
@@ -263,11 +263,74 @@ def rasa_predict(
...
+                # Generator ran out of batches, time to finish inferencing
+                break
         return outputs
 
+    @staticmethod
+    def _merge_batch_outputs(
+        all_outputs: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+        batch_output: Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]],
+    ) -> Dict[Text, Union[np.ndarray, Dict[Text, np.ndarray]]]:
+        """Merge a batch's output into the output for all batches.
```
samsucik, line 307:
```suggestion
        """Merges a batch's output into the output for all batches.
```

tests/utils/tensorflow/test_models.py
```diff
@@ -0,0 +1,118 @@
...
+)
+def test_batch_inference(
+    batch_size: int,
+    number_of_data_points: int,
+    expected_number_of_batch_iterations: int,
+):
+    model = RasaModel()
+
+    def _batch_predict(
+        batch_in: np.ndarray,
```
samsucik, line 74: Now I noticed that `batch_predict` in `models.py` has `batch_in: Union[Tuple[tf.Tensor], Tuple[np.ndarray]]`; I guess here it could be made compatible by changing:
```suggestion
        batch_in: Tuple[np.ndarray, ...],
```
(or perhaps `Tuple[np.ndarray]` for consistency, even though it's incorrect -- as discussed elsewhere).

samsucik (approved):
Looks good 🚀

tests/utils/tensorflow/test_models.py
dakshvar22: Good point. Changed it to `Tuple[np.ndarray]`.

dakshvar22:
Thanks for the good discussion @samsucik 🙌

Issue #8437, opened 2021-04-12
Implement tracker featurizer for IntentTEDPolicy


dakshvar22:
**Description of Problem**:
`IntentTEDPolicy` requires incoming trackers to be featurized in a way such that intents in trackers become labels during training and the conversation context becomes the input. During inference, the conversation context before the last intent becomes the input.

**Overview of the Solution**:
Implement a tracker featurizer inherited from `MaxHistoryTrackerFeaturizer` to do the above.
A draft version of this is already implemented in `itedp-for-merge` branch.

[Scope Document](https://docs.google.com/document/d/1kjZD59NtnvAkxSfu6sjLzgDucTQT8LJhdfZnILCDVcE/edit?usp=sharing)

[Implementation proposal](https://www.notion.so/rasa/Implementation-Proposal-IntentTedPolicy-f16c3eb785964b1ca1249d508f6dc234#6ceb8d8b71024e27b94c1809dcdc22d7)

The featurization should ignore `action_unlikely_intent` action in the tracker.

**Note**: This is not particularly blocking but the implementation should take into account how [`action_unlikely_intent`](https://github.com/RasaHQ/rasa/issues/8417) is implemented.

**Definition of Done**:
- [ ] Tracker featurizer implemented as per the proposal
- [ ] Tests are added
- [ ] Docs added
- [ ] Changelog entry

PR #8570, opened 2021-04-29
Intent max history tracker featurizer

kedz:
**Proposed changes**:
- Deprecate `training_states_actions_and_entities` and `training_states_and_actions` for `training_states_labels_and_entities` and `training_states_and_labels`.
- Refactor `MaxHistoryTrackerFeaturizer` to ease subclassing for `IntentMaxHistoryTrackerFeaturizer`. 
- Implement `IntentMaxHistoryTrackerFeaturizer`
- Extend tests in `tests/core/featurizers/test_tracker_featurizer.py`.
- Closes https://github.com/RasaHQ/rasa/issues/8437

Pending items:
- [x] To make rule hiding work for `IntentTED`, we needed to change how trackers are yielded in [`generate_all_prior_trackers`](https://github.com/RasaHQ/rasa/blob/28e1df057c21e5bfdabe58834f5933a81e701d3a/rasa/shared/core/trackers.py#L460). Earlier, a tracker was always yielded at the end of the for-loop which meant that there could be a duplicate tracker yielded in case the last event in the tracker was not of type `ActionExecuted`. This was problematic to the logic of hiding rules in the tracker for ML policies. The logic was written for `TEDPolicy` in mind originally and it works well there but not for `IntentTEDPolicy`. We changed the tracker yielding logic to not yield the duplicate tracker always because that seemed inefficient anyways. However, with that change `RulePolicy` is currently untrainable (you can see the tests failing). There is a discussion on slack where we are currently trying to figure out why is it so. Subject to that discussions changes will be made.
- [x] Docstrings


tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +55,1060 @@ def test_convert_labels_to_ids(domain: Domain):
...
+
+    tracker_featurizer = IntentMaxHistoryTrackerFeaturizer()
+
+    actual_labels = tracker_featurizer._convert_labels_to_ids(
+        trackers_as_intents, domain
+    )
+
+    expected_labels = np.array(
+        [
+            [7, 8, 9, 10],
```
JEM-Mosig, line 72: These numbers depend on `domain.intents.index(...)`. Can we avoid this?

JEM-Mosig: Not so important, though.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +55,1060 @@ def test_convert_labels_to_ids(domain: Domain):
...
+
+@pytest.mark.parametrize(
+    "tracker_featurizer_class", 
+    [
+        FullDialogueTrackerFeaturizer, 
+        MaxHistoryTrackerFeaturizer,
+        IntentMaxHistoryTrackerFeaturizer,
+    ]
+)
+def test_training_states_actions_and_entities_deprecated(
```
JEM-Mosig, line 1079: Not sure we need tests for deprecation warnings... 🤔

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +55,1060 @@ def test_convert_labels_to_ids(domain: Domain):
...
+    for actual, expected in zip(
+        actual_features,
+        expected_features 
+    ):
+        assert compare_featurized_states(
+            actual, 
+            expected
+        )
+
+
```
JEM-Mosig, line 950: That's a lot of tests! 💯

JEM-Mosig (commented):
Looks good!

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -526,102 +616,193 @@ def slice_state_history(
...
+    def _check_example_cache(
+        self,
+        tracker: DialogueStateTracker,
+        states: List[State],
+        label: List[Text],
+    ) -> bool:
+        """Returns True if training example is a duplicate."""
+        if not self.remove_duplicates:
+            return False
+        else:
```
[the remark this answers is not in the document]
kedz: Will do!

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -662,3 +843,171 @@ def prediction_states(
         self._choose_last_user_input(trackers_as_states, use_text_for_last_user_input)
 
         return trackers_as_states
+
+
+class IntentMaxHistoryTrackerFeaturizer(MaxHistoryTrackerFeaturizer):
+
+    LABEL_NAME = "intent"
+
+    def _convert_labels_to_ids(
```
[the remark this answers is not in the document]
kedz: I actually tried something like this first and timed it against the draft implementation that is more or less used now and it was substantially slower than the draft implementation. I can do an implementation like this though if you prefer.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -662,3 +843,171 @@ def prediction_states(
...
+        The number of rows is equal to `len(trackers_as_intents)`.
+        The number of columns is equal to the maximum number of labels
+        that any Labels item has. Rows are padded with -1 if not all Labels 
+        items have the same number of labels.
+
+        Returns:
+            A 2d np.ndarray of label ids.
+        """
+
+        # store labels in numpy arrays so that it corresponds to np arrays of input features
```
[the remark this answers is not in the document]
kedz: Changing to `# Convert label text lists to label id lists.`

tests/core/featurizers/test_tracker_featurizer.py
kedz: Do you mean it would be better to make `expected_labels` using `domain.intents.index` directly instead of having hard coded it?

tests/core/featurizers/test_tracker_featurizer.py
kedz: Yeah I can get rid of them, mostly just for myself to make sure that I had refactored everything correctly.

twerkmeister:
Hey @kedz, do I understand correctly, that the PRs most important part is the implementation of the new featurizer `IntentMaxHistoryTrackerFeaturizer`? If so, would you mind adding a few lines in the PR text describing what problem this new featurizer solves? If I understand the why, I can usually understand and review the changes much better

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -178,14 +183,60 @@ def training_states_and_actions(
...
+
+        Args:
+            trackers: The trackers to transform
+            domain: The domain
+            omit_unset_slots: If `True` do not include the initial values of slots.
+
+        Returns:
+            A tuple of list of states, list of labels and list of entity data.
+        """
+        return self.training_states_actions_and_entities(
```
twerkmeister, line 237: If you swap the method logic you can throw out a lot of repetitive in the subclasses:
1. let this method here, `training_states_labels_and_entities`, raise the not implemented error
2. let `training_states_actions_and_entities` print the deprecation warning and call `training_states_labels_and_entities`
3. remove all the implementations for `training_states_actions_and_entities` from subclasses

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -178,14 +183,60 @@ def training_states_and_actions(
...
+        rasa.shared.utils.io.raise_deprecation_warning(
+            "'training_states_and_actions' is being deprecated in favor of "
+            "'training_states_and_labels'."
+        )
+
+        return self.training_states_and_labels(
+            trackers, domain, omit_unset_slots=omit_unset_slots
+        )
+
+    def training_states_and_labels(
```
twerkmeister, line 196: there's still 12 calls to `training_states_and_actions` in the code base although it is deprecated now. I assume these should be changed to calling `training_states_and_labels` now otherwise we'll get a lot of warnings.

twerkmeister: In general, it might be worth asking: why keep `training_states_and_actions` at all?

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +606,172 @@ def slice_state_history(
...
+                pbar.set_postfix({f"# {self.LABEL_NAME}": f"{len(example_labels):d}"})
+
+        self._cleanup_example_iterator()
+        self._remove_user_text_if_intent(example_states)
+
+        logger.debug(f"Created {len(example_states)} {self.LABEL_NAME} examples.")
+
+        return example_states, example_labels, example_entities
+
+    def _setup_example_iterator(self) -> None:
```
twerkmeister, line 709: I am not a fan of these iterator setup and cleanup procedures. This reliance on side effects makes the code very path dependent and difficult to understand locally, debug, and reuse. It seems to me, it should be possible to put the whole filtering of duplicates into the `_example_iterator` method so that it doesn't need to rely on instance variables that have to be setup independently

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +606,172 @@ def slice_state_history(
...
+        """Create set for filtering out duplicated training examples."""
+        if self.remove_duplicates:
+            self.hashed_examples = set()
+
+    def _example_iterator(
```
twerkmeister, line 714: `_example_iterator` is not a good method name imo. They usually start with a verb. I could imagine `extract_examples` as in extracting the training examples from a tracker or `generate_examples_from_tracker` or something similar

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +606,172 @@ def slice_state_history(
...
+
+        label_index = 0
+        entity_data = {}
+        for event in tracker.applied_events():
+            if isinstance(event, UserUttered):
+                entity_data = self._entity_data(event)
+
+            elif isinstance(event, ActionExecuted):
+
+                label_index += 1
```
twerkmeister, line 739: I guess this doesn't go directly to you, but I gotta say that it's not trivial to understand / verify that this relationship between `tracker_states` and the events works out in that way that you only advance the quasi pointer to the position in `tracker_states` for every `ActionExecuted` event and not for other types

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +606,172 @@ def slice_state_history(
...
+                yield sliced_states, label, entities
 
                 # reset entity_data for the the next turn
                 entity_data = {}
-                pbar.set_postfix({"# actions": "{:d}".format(len(trackers_as_actions))})
-
-        self._remove_user_text_if_intent(trackers_as_states)
 
-        logger.debug("Created {} action examples.".format(len(trackers_as_actions)))
+    def _check_example_cache(
```
twerkmeister, line 756: Not a good method name imo. `check` is too unspecific to say whether the method would return true or false if it is a duplicate. And example cache doesn't point to the method being about duplicates. All things you put into the doc string instead ;)

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +606,172 @@ def slice_state_history(
...
+        self, tracker: DialogueStateTracker, states: List[State], label: List[Text],
+    ) -> bool:
+        """Returns True if training example is a duplicate."""
+        if not self.remove_duplicates:
+            return False
+        else:
+            hashed = self._hash_example(tracker, states, label)
+            if hashed not in self.hashed_examples:
+                self.hashed_examples.add(hashed)
```
twerkmeister, line 765: I don't like that this method relies on side effects. I think it's not necessary. Side effects are evil 👿 😛   and apart from the disadvantages I have mentioned before, they also make it really difficult or impossible to parallelize code.

For example you can probably manage the duplicate tracking either in `training_states_labels_and_entities` or in `_example_iterator`. And if you really need an extra method where you check whether it is a duplicate you can pass the existing set as a parameter. Then the result of the method only depends on its parameters, and not on some outside variables -> which guarantees that it's free of side effects.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +606,172 @@ def slice_state_history(
...
+                return False
+            else:
+                return True
 
-        return trackers_as_states, trackers_as_actions, trackers_as_entities
+    def _cleanup_example_iterator(self) -> None:
```
twerkmeister, line 770: as before probably not necessary

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -662,3 +812,178 @@ def prediction_states(
...
+        self, tracker: DialogueStateTracker, states: List[State], label: List[Text],
+    ) -> bool:
+        if not super()._check_example_cache(tracker, states, label):
+            state_hash = self._hash_example(tracker, states)
+            self._state_hash_to_labels[state_hash].append(label)
+            return False
+        else:
+            return True
+
+    def _cleanup_example_iterator(self) -> None:
```
twerkmeister, line 908: this is really really difficult to wrap ones head around - even just thinking about the types of the variables and what they might represent is challenging.

I keep repeating myself, but I don't like the side effects. It's also not self evident that a cleanup method actually changes the extracted labels. I would challenge you to get this seemingly important logic out of the cleanup into the example extraction procedures.

twerkmeister (changes requested):
Hey @kedz really looking forward to seeing this tracker featurizer in action! Cool stuff. I made quite some change suggestions and I hope they make sense to you. Happy to discuss all of them. Just let me know!

rasa/core/featurizers/tracker_featurizers.py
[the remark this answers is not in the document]
dakshvar22: We should go with what's faster IMO. I suspect the nested `for` loops in the suggestion above is inefficient.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -662,3 +812,181 @@ def prediction_states(
...
+        """
+
+        # store labels in numpy arrays so that it corresponds to np arrays
+        # of input features
+        label_ids = [
+            [domain.intents.index(intent) for intent in tracker_intents]
+            for tracker_intents in trackers_as_intents
+        ]
+
+        pad_val = -1
```
dakshvar22, line 841: Could you make this value a constant declared in `rasa/utils/tensorflow/constants.py`? It can then be reused by the policy and that would ensure that there is no mismatch between what's used as padding value in the tracker featurizer and what padding value is used in the policy.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +606,172 @@ def slice_state_history(
...
+        """
+        rasa.shared.utils.io.raise_deprecation_warning(
+            "'training_states_actions_and_entities' is being deprecated in "
+            "favor of 'training_states_labels_and_entities'."
+        )
+        return self.training_states_labels_and_entities(
+            trackers, domain, omit_unset_slots=omit_unset_slots
+        )
+
+    def training_states_labels_and_entities(
```
dakshvar22, line 650: I don't think this featurizer or intent max history featurizer unfeaturizes `action_unlikely_intent`. I'd suggest doing a similar cleanup to final states to achieve this as `_cleanup_last_user_state_with_action_listen` does.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +55,1010 @@ def test_convert_labels_to_ids(domain: Domain):
...
+                    return False
+                if feature1.attribute != feature2.attribute:
+                    return False
+                if feature1.type != feature2.type:
+                    return False
+
+    return True
+
+
+@pytest.fixture
```
dakshvar22, line 140: I think all fixtures should go to `tests/core/conftest.py` to 1. make them available across tests and 2. Have them in one place . @twerkmeister Could you please advise?

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +55,1010 @@ def test_convert_labels_to_ids(domain: Domain):
...
+    assert actual_labels.shape == expected_labels.shape
+    assert np.all(actual_labels == expected_labels)
+
     # moodbot doesn't contain e2e entities
     assert not any([any(turn_tags) for turn_tags in entity_tags])
 
 
-def test_featurize_trackers_with_max_history_tracker_featurizer(moodbot_domain: Domain):
+@pytest.mark.parametrize("remove_duplicates", [True, False])
+@pytest.mark.parametrize("max_history", [None, 2])
```
dakshvar22, line 399: Can you please put the 4 combinations in a single `pytest.mark.parametrize`? I think that's how all other tests do and would be good to keep it universal.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +55,1010 @@ def test_convert_labels_to_ids(domain: Domain):
...
+        [moodbot_tracker, moodbot_tracker], moodbot_domain, RegexInterpreter()
+    )
+
+    expected_features = [
+        [{},],
+        [
+            {},
+            {
+                "intent": [moodbot_features["intents"]["greet"]],
+                "action_name": [moodbot_features["actions"]["action_listen"]],
```
dakshvar22, line 423: Please use the constants `INTENT`, `ACTION_NAME`, `ACTION_LISTEN` wherever possible.

dakshvar22 (changes requested):
Looking good 💪  ! Functionality for unfeaturizing `action_unlikely_intent` is missing. I've pointed it out in the comments.

rasa/core/featurizers/tracker_featurizers.py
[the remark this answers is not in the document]
JEM-Mosig: Fine with me.

tests/core/featurizers/test_tracker_featurizer.py
JEM-Mosig: Yes. Because we don't want to test if the domain has changed, and this test would fail in this case.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -662,3 +843,171 @@ def prediction_states(
...
+        # each example has equal number of labels
+        multiple_labels_count = [len(a) for a in label_ids]
+        max_labels_count = max(multiple_labels_count)
+        num_padding_needed = [max_labels_count - len(a) for a in label_ids]
+
+        new_label_ids = []
+        for ids, num_pads in zip(label_ids, num_padding_needed):
+            if num_pads:
+                ids.extend([pad_val] * num_pads)
+            new_label_ids.append(ids)
```
JEM-Mosig, line 853: Can you package this into a helper method? It would make the code a bit easier to read.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -526,102 +616,193 @@ def slice_state_history(
...
-                if isinstance(event, UserUttered):
-                    entity_data = self._entity_data(event)
+            for states, label, entities in self._example_iterator(
+                tracker,
+                domain,
+                omit_unset_slots=omit_unset_slots
+            ):
 
-                if not isinstance(event, ActionExecuted):
+                if self._check_example_cache(tracker, states, label):
```
JEM-Mosig, line 693: Can we give this a more meaningful name, like `_check_duplicate_example`?

rasa/core/featurizers/tracker_featurizers.py
kedz: Will do!

rasa/core/featurizers/tracker_featurizers.py
kedz: Agree with both of these comments. I found this part pretty confusing. @dakshvar22 if this isn't already in the doc, would this be a good candidate for the rasa 3.0 todos?

rasa/core/featurizers/tracker_featurizers.py
kedz: Removing these setup and cleanup functions.

rasa/core/featurizers/tracker_featurizers.py
kedz: Sounds good! How does `LABEL_PAD_ID` sound to you?

rasa/core/featurizers/tracker_featurizers.py
kedz: Calling it `_is_duplicate_example`

rasa/core/featurizers/tracker_featurizers.py
kedz: Will remove these functions with side effects and bring them inside `training_states_labels_and_entities`.

rasa/core/featurizers/tracker_featurizers.py
kedz: Will rename to `extract_examples`.

rasa/core/featurizers/tracker_featurizers.py
kedz: I'm renaming it `_is_duplicate_example` -- does that sound good?

rasa/core/featurizers/tracker_featurizers.py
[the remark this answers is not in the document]
kedz: I'll keep as is but roll the padding into a helper function as @JEM-Mosig has suggested below.

rasa/core/featurizers/tracker_featurizers.py
kedz: Will do!

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +55,1010 @@ def test_convert_labels_to_ids(domain: Domain):
...
+@pytest.mark.parametrize("remove_duplicates", [True, False])
+@pytest.mark.parametrize("max_history", [None, 2])
+def test_deduplicate_featurize_trackers_with_intent_max_history_tracker_featurizer(
+    moodbot_tracker: DialogueStateTracker,
+    moodbot_domain: Domain,
+    moodbot_features: Dict[Text, Dict[Text, Features]],
+    remove_duplicates: bool,
+    max_history: Optional[int],
+):
+    state_featurizer = SingleStateFeaturizer()
```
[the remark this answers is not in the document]
kedz: Will do!

kedz:
@twerkmeister @dakshvar22 I believe I have addressed your comments and I think its ready for another look. The biggest change is that I have removed setup/cleanup/duplication checking functions with side effects and moved their logic into the `training_states_labels_and_entities` function.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -73,18 +76,48 @@ def _create_states(
...
+    @staticmethod
+    def _remove_action_unlikely_intent(states: List[State]) -> List[State]:
+        """Remove `action_unlikely_intent` from tracker state history.
+
+        Args:
+            states: A list of states produced by a `DialogueStateTracker`
+                instance.
+
+        Returns:
+            Filtered states with last `action_listen` removed.
```
dakshvar22, line 97:
```suggestion
            Filtered states with all states containing `action_unlikely_intent` removed.
```

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -73,18 +76,48 @@ def _create_states(
         Returns:
             a list of states
         """
-        return tracker.past_states(
-            domain,
-            omit_unset_slots=omit_unset_slots,
-            ignore_rule_only_turns=ignore_rule_only_turns,
-            rule_only_data=rule_only_data,
+        return cls._remove_action_unlikely_intent(
```
dakshvar22, line 79: Don't think this is correct. As mentioned in the [issue description](https://github.com/RasaHQ/rasa/issues/8437), the decision to remove `action_unlikely_intent` from the states depends on which policy is requesting the features for the tracker. Hence, it should only be removed for all policies which use story data for training (`TEDPolicy` and `IntentTEDPolicy`), but shouldn't be removed for others like `RulePolicy`. It's similar to how rule turns are hidden only for the former policies and not latter. See [line](https://github.com/RasaHQ/rasa/blob/3d68c685c92bb423aebaf521987990c2d89a1a35/rasa/core/policies/policy.py#L263). Since you have added the call to `_remove_action_unlikely_intent` inside `_create_states` function of the parent class, it will always be removed.

dakshvar22 (commented):
Haven't looked at all changes, but the logic for removing `action_unlikely_intent` should be rectified. Added a comment for it.

rasa/core/featurizers/tracker_featurizers.py
kedz: In the latest commit, I have removed `action_unlikely_intent` filtering from `_create_states` and put it in the `FullDialogue/MaxHistory/IntentMaxHisory-TrackerFeaturizer` implementations of `prediction_states` and `training_states_labels_and_entities`.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -205,8 +327,12 @@ def featurize_trackers(
             domain: the domain
             interpreter: the interpreter
             bilou_tagging: indicates whether BILOU tagging should be used or not
+            ignore_action_unlikely_intent: Remove action_unlikely_intent from
+                training state features.
```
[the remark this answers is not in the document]
dakshvar22: I'd suggest the change in all docstrings where `ignore_action_unlikely_intent` is mentioned. (If you agree with the change)

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -410,6 +567,9 @@ def training_states_actions_and_entities(
                 tracker, domain, omit_unset_slots=omit_unset_slots
             )
 
+            if ignore_action_unlikely_intent:
+                states = self._remove_action_unlikely_intent(states)
+
```
dakshvar22, line 572: Can't we move this `if` construct just before `trackers_as_states.append(states[:-1])`. That way you don't need the `if` construct on line 584 as well. Right now, the logic will be fragile if someone messes with any of these `if` constructs, hence cleaning up the actions once the states have been created is a better option IMO.

Alternatively, you can do it before `self._remove_user_text_if_intent(trackers_as_states)`. Either way, you won't need the `if` construct inside the `for` loop over tracker events.
What do you think?

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +710,439 @@ def slice_state_history(
...
+        Args:
+            trackers: The tracker from which to extract training examples.
+            domain: The domain of the training data.
+            omit_unset_slots: If `True` do not include the initial values of slots.
+            ignore_action_unlikely_intent: Remove action_unlikely_intent from
+                training states.
+
+
+
```
dakshvar22, line 826: Too many whitespaces, only need 1 line of whitespace

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +64,2238 @@ def test_convert_labels_to_ids(domain: Domain):
...
+            {
+                USER: {INTENT: "mood_unhappy"},
+                PREVIOUS_ACTION: {ACTION_NAME: "utter_cheer_up"},
+            },
+            {
+                USER: {INTENT: "mood_unhappy"},
+                PREVIOUS_ACTION: {ACTION_NAME: "utter_did_that_help"},
+            },
+        ]
+    ]
```
aeshky, line 2039: I don't understand why this is identical to `expected_state` in `test_prediction_states_with_max_history_tracker_featurizer` but missing the last two entries.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +710,439 @@ def slice_state_history(
...
+        for event in tracker.applied_events():
+            if isinstance(event, UserUttered):
+                entity_data = self._entity_data(event)
+
+            elif isinstance(event, ActionExecuted):
+
+                if (
+                    ignore_action_unlikely_intent
+                    and event.action_name == ACTION_UNLIKELY_INTENT_NAME
+                ):
```
dakshvar22, line 849: Okay, I can see why you need the `if` construct here. You want to do the cleanup before the states are sliced. In that case I would recommend moving the call to `slice_state_history` outside of this function and move it inside the `for` loop (line 781) of `training_states_labels_and_entities`. So, `_extract_examples` can operate as if it knows nothing about slicing but then `training_states_label_and_entities` applies the slicing as needed.
 What do you think?

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +710,439 @@ def slice_state_history(
...
+
+        label_index = 0
+        for event in tracker.applied_events():
+
+            if isinstance(event, ActionExecuted):
+                if (
+                    ignore_action_unlikely_intent
+                    and event.action_name == ACTION_UNLIKELY_INTENT_NAME
+                ):
+                    continue
```
dakshvar22, line 1114: Same feedback here about `ignore_Action_unlikely_intent` and moving the slicing outside of the function to reduce the dependency between the two.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +64,2238 @@ def test_convert_labels_to_ids(domain: Domain):
...
+            {
+                USER: {INTENT: "mood_great"},
+                PREVIOUS_ACTION: {ACTION_NAME: ACTION_UNLIKELY_INTENT_NAME},
+            },
+            {
+                USER: {INTENT: "mood_great"},
+                PREVIOUS_ACTION: {ACTION_NAME: "utter_happy"},
+            },
+        ]
+    ]
```
aeshky, line 2228: I'm confused why we don't flip intents and actions for intent ted. 
In this example you have: intent greet, followed by three possible actions. 
To train intent ted, dont you need the opposite? each action and the possible intents that can follow?

aeshky (commented):
Hi @kedz.
I went through your code and ran all of your tests. I also stepped through some of your code/tests. Apologies in advance for my poor understanding. Maybe we can clarify things in a call? I have two main comments: 
1) In the tests, I found it difficult to understand why the expected features or states are what they are. This is probably due to my lack of familiarity with the rasa representations. I would ask you to add comments, but if someone with a better understanding knows exactly what's going on then that's not necessary. 
2) The main thing is that I expected the following: ted predicts the next action given the previous context, while intent ted predicts the next intent given the previous context. However, looking at the expected features and expected states, it looks like we are not slicing the dialogs differently in the case of ted vs. intent ted, we're just labelling things differently. Is my understanding correct? In that case, would we learn the correct thing? In many places we say PREVIOUS ACTION when it is the next action. It makes sense to predict the next action for ted, but why are we doing the same for intent ted? Don't we need the opposite (action followed by possible intents)?

rasa/core/featurizers/tracker_featurizers.py
kedz: I also need the if statement in the event for loop to not include `action_unlikely_intent` events as labels (although maybe I am not fully understanding the implications of your suggestion). What if I do something like this:
- Create the event list with tracker_states
- Filter the event list in the same if statement that tracker_states is being filtered
This removes the need for the second if statement in the event for loop.

This would look like this:
```
        tracker_states = self._create_states(
            tracker, domain, omit_unset_slots=omit_unset_slots
        )
        events = tracker.applied_events()   ### <-- Create event list here

        if ignore_action_unlikely_intent:
            tracker_states = self._remove_action_unlikely_intent(tracker_states)
            events = [          ### <-- Filter events here when I filter the states
                event for event in events
                if not isinstance(event, ActionExecuted) or event.action_name != ACTION_UNLIKELY_INTENT_NAME
            ]

        label_index = 0
        entity_data = {}
        for event in events:
            if isinstance(event, UserUttered):
                entity_data = self._entity_data(event)

            elif isinstance(event, ActionExecuted):
             
                label_index += 1      ### <-- No if statement here any more, can assume we have a valid state label
                ### Do slicing and yielding here ...

```

Something similar can be done for the `IntentMaxHistoryTrackerFeaturizer`.
Additionally, per your comment below, I can move the state slicing out of this method and into `training_states_labels_and_entities` if you still think that is needed after this change.

kedz:
@aeshky 
> I would ask you to add comments, but if someone with a better understanding knows exactly what's going on then that's not necessary.

I can definitely add comments although it sounds like I may have broken the line counter on test length , so lets see how the broader discussion shakes out and we can make the case for comments there. I'm not sure the logic of featurization code is written down anywhere and maybe if should be? I like the idea of having lots of tests that both check that things are working and show you what the expected inputs and outputs of things are, but maybe that should be a separate document?

kedz:
@aeshky 

> The main thing is that I expected the following: ted predicts the next action given the previous context, while intent ted predicts the next intent given the previous context. However, looking at the expected features and expected states, it looks like we are not slicing the dialogs differently in the case of ted vs. intent ted, we're just labelling things differently. Is my understanding correct?

Not entirely. Let me take a stab at explaining the process and @dakshvar22 correct me where I am also mistaken. Let's start with prediction time since I think that is easier to understand. First even though TED and IntentTED are predicting actions and intents respectively they make predictions at the same time during a dialogue: when the next action needs to be predicted, which is pretty much any time the last event wasn't `action_listen` (technically IntentTED's prediction are only needed when the last event was a UserUttered but that's outside the scope of the tracker featurizer).

Let's say we have a the following tracker events:

```yaml
- ActionExecuted("action_listen")
- UserUttered("greet")
```

States are created for every action prediction site in the event sequence historically as well as the final position where TED now needs to make a prediction. So that would be two locations in the tracker above:

```yaml
#<-- Prediction Site 1: TED needed a state here to predict action_listen
- ActionExecuted("action_listen")
- UserUttered("greet")
#<-- Prediction Site 2: TED needs a state now to make its next prediction
```

The process for creating a state is to start from a prediction site and look backwards to see what the previous actions and user utterances were (this is where `prev_action` comes from). So for the second site above, the last user intent is `greet` and the previous action was `action_listen` so the corresponding state would be:

```json
{
    "user": {"intent": "greet"},
    "prev_action": {"action_name": "action_listen"}
}
``` 
Repeating the process, for the first site, its somewhat of a special case as its the beginning of the dialogue and moving backward, there is neither a previous action nor a user utterance so the state is just the empty dictionary:

```json
{}
```

These two states form the entire state sequence that TED would use at this point if `max_history=None` to make a prediction. That is, it would featurize the following state sequence: 

```json
[
    {}, 
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "action_listen"}
    }
]
```

If `max_history = 1`, it would use the last 1 state, so just:

```json
[
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "action_listen"}
    }
]
```

And now IntentTED is going to be run but its actually going to be run a step behind where TED is at because it is going to be used to flag whether the `greet` intent is likely at this juncture and so it should not see the corresponding state. The general rule is that if the last event is a `UserUttered` event then TED will get to see the last `action_listen` and `UserUttered` events while IntentTED will not: 

```yaml
#<-- IntentTED can use all information prior to this point
- ActionExecuted("action_listen")
- UserUttered("greet")
#<-- TED Can use all information prior to this point
```
So `MaxHistoryTrackerFeaturizer` and `IntentMaxHistoryTrackerFeaturizer` will be given the same tracker as input, but in the case where the last event is a `UserUttered` event, `IntentMaxHistoryTrackerFeaturizer` will filter out that last state. So in the above case, the only state before that point is same state for the start of conversation, i.e. the empty dictionary:

```json
[
    {},
]
```
which is what `IntentMaxHistoryTrackerFeaturizer` would produce from the tracker events above.

kedz:
@aeshky 

Now the process for converting a tracker to training data. Before we were just creating one state sequence for prediction but now we need to create a state sequence/label pair for every prediction site in a tracker. Let's say we had the following training tracker:

```yaml
- ActionExecuted("action_listen")
- UserUttered("greet")
- ActionExecuted("utter_greet")
- ActionExecuted("utter_how_are_you")
- ActionExecuted("action_listen")
- UserUttered("mood_happy")
- ActionExecuted("utter_goodbye")
- ActionExecuted("action_listen")
```

This story has the following prediction sites for TED that can be turned into training state sequence/action label pairs:

```yaml
# <-- Prediction Site 1   (label is action_listen)
- ActionExecuted("action_listen")
- UserUttered("greet")
# <-- Prediction Site 2   (label is utter_greet)
- ActionExecuted("utter_greet")
# <-- Prediction Site 3   (label is utter_how_are_you)
- ActionExecuted("utter_how_are_you")
# <-- Prediction Site 4   (label is action_listen)
- ActionExecuted("action_listen")
- UserUttered("mood_happy")
# <-- Prediction Site 5   (label is utter_goodbye)
- ActionExecuted("utter_goodbye)
# <-- Prediction Site 6   (label is action_listen)
- ActionExecuted("action_listen")
```

Assuming that `max_history = None`, the state sequence at each prediction site is all previous states up to and including the current prediction site.

So the first state sequence is

```json
[
    {},
]
```
and its label is `action_listen`.

The second state sequence is 

```json
[
    {}, 
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "action_listen"}
    }
]
```
and its label is `utter_greet`.

The third state is 

```json
[
    {}, 
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "action_listen"}
    },
        {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "utter_greet"}
    }
]
```
and its label is `utter_how_are_you`.

And so on.


IntentTED on the other hand only has two prediction sites where it can create training data, corresponding to the states before the two UserUttered events:

```yaml
# <-- IntentTED prediction site 1
- ActionExecuted("action_listen")
- UserUttered("greet")
- ActionExecuted("utter_greet")
- ActionExecuted("utter_how_are_you")
# <-- IntentTED prediction site 2
- ActionExecuted("action_listen")
- UserUttered("mood_happy")
- ActionExecuted("utter_goodbye")
- ActionExecuted("action_listen")
```

The state sequence for the first IntentTED training example would be all states created up to and including the first prediction site:

```json
[
    {},
]
```
and the label is `greet`.

The state sequence for the second example would be:

```json
[
    {}, 
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "action_listen"}
    },
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "utter_greet"}
    },
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "utter_how_are_you"}
    }
]
```
and the label is `mood_happy`.

kedz:
A final note, when a state sequence gets featurized (`featurize_trackers` or `create_state_features`), it produces a sequence of dictionaries corresponding to the individual state dictionaries, but the keys point to lists of `Features` instances (which are essentially a wrapper around a one-hot sparse vector representation of the intent or action names. Additionally, while the intent is often repeated in states when there are multiple actions, only the first instance of an intent is actually featurized. So if we featurize: 

```json
[
    {}, 
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "action_listen"}
    },
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "utter_greet"}
    },
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "utter_how_are_you"}
    }
]
```

we should get:

```json

[
    {},
    {
        "intent": [<rasa.shared.nlu.training_data.features.Features object at 0x7fdf83171f10>],
        "action_name": [<rasa.shared.nlu.training_data.features.Features object at 0x7fdf83171e90>]
    },
    {
        "action_name": [<rasa.shared.nlu.training_data.features.Features object at 0x7fdf8318c910>]
    },
    {
        "action_name": [<rasa.shared.nlu.training_data.features.Features object at 0x7fdf8318c6d0>]
    }
]

```

twerkmeister:
Thanks for the additional explanation, makes a lot of sense @kedz!

Just one thing caught my eye. I think
```
# <-- IntentTED prediction site 1
- ActionExecuted("action_listen")
- UserUttered("greet")
- ActionExecuted("utter_greet")
- ActionExecuted("utter_how_are_you")
# <-- IntentTED prediction site 2
- ActionExecuted("action_listen")
- UserUttered("mood_happy")
- ActionExecuted("utter_goodbye")
- ActionExecuted("action_listen")
```
should probably be 
```
- ActionExecuted("action_listen")
- UserUttered("greet")
# <-- IntentTED prediction site 1
- ActionExecuted("utter_greet")
- ActionExecuted("utter_how_are_you")
- ActionExecuted("action_listen")
- UserUttered("mood_happy")
# <-- IntentTED prediction site 2
- ActionExecuted("utter_goodbye")
- ActionExecuted("action_listen")
```
right?

kedz:
@twerkmeister 
> But maybe it's a confusion about "what data does it see" and "when does it make the prediction"

Yeah this part is confusing. At prediction time, IntentTED is making predictions alongside TED, and where you have pointed to is the state of the tracker that would be passed into both featurizers. The `IntentMaxHistoryTrackerFeaturizer` internally will roll back the last `action_listen` and `UserUttered` events before passing the features on to IntentTED, because we want it to make a prediction of the counterfactual (what would a user likely have uttered at this point), and so it is as if IntentTED is running from the prediction sites where I have indicated.

For training you don't need to worry about this because IntentTED and TED are not coupled in this way, the IntentTED prediction sites are where I have pointed out, just before the `action_listen` that preceeds a `UserUttered` event.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +64,2238 @@ def test_convert_labels_to_ids(domain: Domain):
...
+            },
+            {ACTION_NAME: [moodbot_features["actions"]["utter_cheer_up"]]},
+            {ACTION_NAME: [moodbot_features["actions"]["utter_did_that_help"]]},
+            {
+                INTENT: [moodbot_features["intents"]["deny"]],
+                ACTION_NAME: [moodbot_features["actions"][ACTION_LISTEN_NAME]],
+            },
+        ]
+    ]
+
```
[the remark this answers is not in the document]
kedz: I just ordered them arbitrarily but consistently because the state is a dictionary. It won't always be the case that the intent follows the action. For instance,

```yaml
- ActionExecuted("action_listen")
- UserUttered("greet")
- ActionExecuted("utter_greet")
- ActionExecuted("utter_how_are_you")
```

would produce the following states (putting the state keys in chronological order) :

```json
[
    {}, 
    {
        "prev_action": {"action_name": "action_listen"},
        "user": {"intent": "greet"}       
    },
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "utter_greet"}
    },
    {
        "user": {"intent": "greet"},
        "prev_action": {"action_name": "utter_how_are_you"}
    },
]
```

I could reorder them this way if you think that would help. I don't think there is any Rasa convention around this.

rasa/core/featurizers/tracker_featurizers.py
dakshvar22: Hmm, that's a good point about pruning them for labels. I agree with your suggestion above 👍 
Just a minor (maybe it was just a typo from you) - it will be `if not isinstance(event, ActionExecuted) and event.action_name != ACTION_UNLIKELY_INTENT_NAME` and not `if not isinstance(event, ActionExecuted) or event.action_name != ACTION_UNLIKELY_INTENT_NAME`

dakshvar22: State slicing can stay there only then.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +64,2238 @@ def test_convert_labels_to_ids(domain: Domain):
...
+            {USER: {INTENT: "deny"}, PREVIOUS_ACTION: {ACTION_NAME: "utter_goodbye"},},
+        ]
+    ]
+
+    assert actual_states is not None
+    assert len(actual_states) == len(expected_states)
+
+    for actual, expected in zip(actual_states, expected_states):
+        assert actual == expected
+
```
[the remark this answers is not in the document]
kedz: [See this comment.](https://github.com/RasaHQ/rasa/pull/8570#discussion_r640551284) Does that clear things up?

tests/core/featurizers/test_tracker_featurizer.py
kedz: [This comment](https://github.com/RasaHQ/rasa/pull/8570#issuecomment-849965353) also address this question.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -354,10 +354,30 @@ def _encode_action(
...
+        ]
+
     def encode_all_actions(
         self, domain: Domain, interpreter: NaturalLanguageInterpreter
     ) -> List[Dict[Text, List["Features"]]]:
-        """Encode all action from the domain using the given interpreter.
+        """Encode all actions from the domain using the given interpreter.
+
+        This method is deprecated and will be removed in Rasa Open Source 3.0.0 .
+        It is recommended to use `encode_all_labels` instead.
```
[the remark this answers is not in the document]
kedz: Great point! I changed the existing test from `test_single_state_featurizer_creates_encoded_all_actions` to `test_single_state_featurizer_creates_encoded_all_labels` and changed the call inside to `encode_all_labels`.

I also added an analogous test `test_intent_tokenized_single_state_featurizer_creates_encoded_all_labels` for the `IntentTokenizedSingleStateFeaturizer`.

And the answer to your question is yes. "Action" was being used synonymously with "label" throughout the core code but now labels can be either actions or intents depending on the model (TED for actions, IntentTED for intents).

tests/core/featurizers/test_tracker_featurizer.py
kedz: Ok I made that change in test_convert_action_labels_to_ids and test_convert_intent_labels_to_ids. To be consistent I should probably do that through but I think it could be done as part of the larger tests refactor.

tests/core/featurizers/test_tracker_featurizer.py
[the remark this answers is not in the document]
aeshky: Ah, this is much more readable, Chris. As someone who is trying to understand what the code does, I think this order will help me a lot.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -49,48 +64,2238 @@ def test_convert_labels_to_ids(domain: Domain):
...
+    actual_labels = tracker_featurizer._convert_labels_to_ids(
+        trackers_as_intents, domain
+    )
+
+    expected_labels = np.array([[7, 8, 9, 10], [5, 6, 0, LABEL_PAD_ID]])
+    assert expected_labels.size == actual_labels.size
+    assert expected_labels.shape == actual_labels.shape
+    assert np.all(expected_labels == actual_labels)
+
```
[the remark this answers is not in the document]
aeshky: You answered part of my question, but now I remember that we encode the labels as well so I think I've answered the second part of my question. This was only a point of curiosity. I'll mark this as resolved. Thanks!

tests/core/featurizers/test_tracker_featurizer.py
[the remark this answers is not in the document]
aeshky: Yes it does. It's a confusing label, but I think it's inherited from older code. So I'm happy to mark it as resolved.

rasa/core/featurizers/single_state_featurizer.py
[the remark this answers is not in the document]
aeshky: Thanks!

aeshky:
Thanks for the explanation @kedz. I'm concenred about the following: 

> And now IntentTED is going to be run but its actually going to be run a step behind where TED is at because it is going to be used to flag whether the greet intent is likely at this juncture and so it should not see the corresponding state.

So intent ted doesn't have access to the last action because it's part of the state. In your example the action was action_listen, but it can be something else, potentially a critical action for predicting the next user intent. No?

tests/core/featurizers/test_tracker_featurizer.py
aeshky: > IntentMaxHistoryTrackerFeaturizer will filter out the last UserUttered and action_listen events for reasons that are explained here.

Your linked comment explains this. Thanks. 

> This tracker is different than the one that you are mentioning for the equivalent MaxHistoryTrackerFeaturizer test. 

Thanks. I can see that you pop the last two events.

aeshky:
On second thought, I believe we only accept a user utterance after an action_listen, so in practice when we hide the last state, the only possible action it can contain is action_listen, is that right? In that case I see no issue with the implementation.

tests/core/featurizers/test_tracker_featurizer.py
aeshky: Thanks.

tests/core/featurizers/test_tracker_featurizer.py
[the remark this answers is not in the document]
aeshky: But it's also a massive hassle to go back and change this, so I'll let you decide if it's worth it.

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -42,55 +57,2276 @@ def test_convert_labels_to_ids(domain: Domain):
...
+    assert actual_states is not None
+    assert len(actual_states) == len(expected_states)
+
+    for actual, expected in zip(actual_states, expected_states):
+        assert actual == expected
+
+
+@pytest.mark.parametrize("max_history", [None, 3])
+def test_prediction_states_keeps_action_intent_unlikely_with_max_history_tracker_featurizer(
```
dakshvar22, line 1490: Over here too, you can use parameterization to get rid of redundant code. (Make `ignore_action_unlikely_intent` and expected states parameters for the tests.)

dakshvar22 (commented):
Just minor comment about docstrings and logic of filtering `action_unlikely_intent` in `FullDialogueTrackerFeaturizer`. Comments on tests can be addressed as part of a follow up PR.

rasa/utils/tensorflow/constants.py
```diff
@@ -2,6 +2,7 @@
 
 LABEL = "label"
 IDS = "ids"
+LABEL_PAD_ID = -1
```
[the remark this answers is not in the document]
kedz: How does 
```python
# LABEL_PAD_ID is used to pad multi-label training examples.
# It should be <0 to avoid index out of bounds errors by tf.one_hot. 
LABEL_PAD_ID = -1
```

sound?

tests/core/featurizers/test_tracker_featurizer.py
```diff
@@ -42,55 +57,2276 @@ def test_convert_labels_to_ids(domain: Domain):
...
+        assert compare_featurized_states(actual, expected)
+
+
+@pytest.mark.parametrize(
+    "max_history,moodbot_features",
+    [
+        [None, "IntentTokenizerSingleStateFeaturizer"],
+        [2, "IntentTokenizerSingleStateFeaturizer"],
+    ],
+    indirect=["moodbot_features"],
```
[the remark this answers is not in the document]
kedz: The actual state features produced by a state featurizer have a different source string depending on whether its using the `IntentTokenizerSingleStateFeaturizer` or `SingleStateFeaturizer` and so I pass the name of the state featurizer to the `moodbot_features` fixture to indicate what the source string should be in the expected state features.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -528,100 +702,437 @@ def slice_state_history(
...
+            tracker_states = self._remove_action_unlikely_intent(tracker_states)
+            # Remove all action_unlikely_intent events.
+            events = [
+                event
+                for event in events
+                if (
+                    not isinstance(event, ActionExecuted)
+                    or event.action_name != ACTION_UNLIKELY_INTENT_NAME
+                )
+            ]
```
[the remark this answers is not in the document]
kedz: Made into a class and used in all three tracker featurizer instantiations.

kedz:
@aeshky 

> On second thought, I believe we only accept a user utterance after an action_listen, so in practice when we hide the last state, the only possible action it can contain is action_listen, is that right? 

That's correct, an `action_listen` always precedes a `UserUttered` event.

kedz:
@dakshvar22 
> Any other blockers for it to be added to the code?

I just addressed the comments on `LABEL_PAD_ID`, fixed that last docstring, and added the same event filtering behavior to `FullDialogueTrackerFeaturizer`. I believe we should be good to go on the merge of implementation.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -83,7 +86,18 @@ def _featurize_states(
...
         trackers_as_states: List[List[State]],
         interpreter: NaturalLanguageInterpreter,
-    ) -> List[List[Dict[Text, List["Features"]]]]:
+    ) -> List[List[Dict[Text, List[Features]]]]:
+        """Featurizes state histories with `state_featurizer`.
+
+        Args:
+            trackers_as_states: Lists of states produced by a `DialogueStateTracker`
+                instance.
+            interpreter: A `NaturalLanguageInterpreter` for the `state_featurizer`
```
dakshvar22, line 95: @kedz I see data types mentioned in the arg descriptions and return descriptions of some functions of this file. Can you please double check across the whole PR and remove those?

kedz: On it!

kedz: Ok any mentions to class types in docstrings should be in reference to things not already defined in the type annotations but let me know if you see anything else that should be changed.

tests/core/featurizers/test_tracker_featurizer.py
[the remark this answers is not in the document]
kedz: I was able to make this change pretty quickly.

dakshvar22 (approved):
LGTM for merge 💯  . Can you please create a follow up issue regarding test refactoring and label it with the intent_ted feature label so that it doesn't get lost?

kedz:
@dakshvar22 Created the issue [here](https://github.com/RasaHQ/rasa/issues/8798) and put it into the research issues inbox. I went through and collected all the comments relevant to the tests from this PR and linked to them in issue as well so if someone else takes this up they should be able to see what has been suggested already.

aeshky (approved):
@kedz I think you've addressed all of my comments / questions. I'm happy to approve.

PR #8601, opened 2021-05-04
Architecture Prototype v1

wochinge:
**Proposed changes**:
- ...


rasa/architecture_prototype/graph_components.py
```diff
@@ -0,0 +1,233 @@
...
+        if isinstance(messages, TrainingData):
+            messages = messages.training_examples
+        additional_features = {}
+        for message in messages:
+            key = next(
+                k
+                for k in message.data.keys()
+                if k in {ACTION_NAME, ACTION_TEXT, INTENT, TEXT}
+            )
+            additional_features[key] = message
```
twerkmeister, line 163: Was just cherry picking some parts to get the new featurization working without the graph and noticed that this code is wrong. Should probably be something like:
```
        for message in messages:
            key = next(
                v
                for k, v in message.data.items()
                if k in {ACTION_NAME, ACTION_TEXT, INTENT, TEXT}
            )
            additional_features[key] = message
```
otherwise you are literally saving `additional_features["action_name"] = message` for almost every message

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -227,26 +217,20 @@ def _get_name_attribute(attributes: Set[Text]) -> Optional[Text]:
...
-        # that prediction and train time features are the same
-        if self._use_regex_interpreter and not isinstance(
-            interpreter, RegexInterpreter
-        ):
-            interpreter = RegexInterpreter()
+        key = next(
+            k for k in sub_state.keys() if k in {ACTION_NAME, ACTION_TEXT, INTENT, TEXT}
+        )
+        parsed_message = e2e_features[key]
+        assert parsed_message
```
twerkmeister, line 227: Same Problem here:

Was just cherry picking some parts to get the new featurization working without the graph and noticed that this code is wrong. Should probably be something like:
```
            key = next(
                v
                for k, v in message.data.items()
                if k in {ACTION_NAME, ACTION_TEXT, INTENT, TEXT}
            )
            additional_features[key] = message
```
otherwise you are literally saving `additional_features["action_name"] = message` for almost every message

Also I got the case where an substate with `{action_name: action_listen}` would get in there, and that wasn't in `e2e_features` dict at all. For that case I have done this so far: combining the old and the new one. Not sure it's working though. Maybe sth else is wrong

`parsed_message = e2e_features.get(key, Message(data=sub_state))`

rasa/architecture_prototype/graph_components.py
wochinge: oh dear - yes, it was very late when I did this 🤦🏻

rasa/core/featurizers/single_state_featurizer.py
wochinge: I think there can be training trackers without `action_listen` as that's added by the `generator`. I'd just add the action_listen part as default then.

rasa/architecture_prototype/graph_components.py
wochinge: thanks for the catch!

Issue #8576, opened 2021-04-29
Explore the heavily changed Rasa NLU training progress


samsucik:
**Rasa version**: 2.5.1 (current `main` branch)

**Python version**: 3.7.9

**Operating system** (windows, osx, ...): MacOS

**Issue**: 
After the recent DIET & TED refactoring PR (#7589) got merged, DIET's training characteristic (how the training loss and accuracy evolves) changed drastically -- at least for one small dataset we've got (see [this Slack thread](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1619685661333200)).

This matters because convergence properties shouldn't change (degrade) this much, and the changes introduced in the PR were not meant to change this at all.

What needs to be done:
- [x] check if a similar effect can be observed on other NLU datasets
- [x] check what the impact is on the test-set results (because training accuracy/loss and generalisation performance are different things)
- [x] explore the different versions (commits) of Rasa OSS to pinpoint the changes that impacted this
- [ ] identify what (if anything) needs to change in the code if we want to "bring things back to normal"

**Error**:
![plot](https://user-images.githubusercontent.com/7470182/116589631-be2d1e00-a914-11eb-8400-7f4bbe36e004.png)

**Content of configuration file (config.yml)** (if relevant):
```yml
language: en
pipeline:
  - name: WhitespaceTokenizer
  - name: RegexFeaturizer
  - name: LexicalSyntacticFeaturizer
  - name: CountVectorsFeaturizer
  - name: CountVectorsFeaturizer
    analyzer: "char_wb"
    min_ngram: 1
    max_ngram: 4
  - name: DIETClassifier
    epochs: 100
    random_seed: 47
```

samsucik:
Something weird happens to training accuracy very early in the history of changes on the PR (#7589) and its predecessor (#7378). I went through all the commits on these PRs and pinned down the consecutive commit pairs where the numbers change. Below are results that should cover the key commits and some more.

For now ignoring the test-set accuracy, here are stats for the final training accuracy (what one sees in the training progress bar); averaged over the same 4 random seeds, with std. deviation shown as well. Ordered by commit time, from oldest to newest. All the commits come from the 2 mentioned PRs.

The "car bot":
```
timestamp_commit-hash | datetime | final train acc. ± stdev | test acc. ± stdev
1605195798_f87c134f7aed137944dbf3be49cb9da30bbeceee | 2020-11-12 15:43:18 | 0.95±0.02 | 0.00±0.00
[...]

PR #8607, opened 2021-05-05
Fix dropout applied to labels in ML models

samsucik:
**Proposed changes**:
- Fixes the recently discovered bug (#8576) where dropout is wrongly applied to both sparse and dense input features (introduced as part of TED & DIET refactoring in #7589)

- [ ] ~added some tests for the functionality~
- [ ] ~updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)~


samsucik:
I've made a fix in all the 3 ML models. What's left to do:
- model regression tests (just to be on the safe side, though we know they're not bulletproof)
- some overall tests especially to check Core (I'm currently training both NLU and Core on rasa-demo and on the 2.0 version of Carbon bot, aiming to compare performance from the #7589 merge commit and from this PR)

@dakshvar22 what do you think about adding a test that checks that no dropout is applied to label features within the models? I'm kind of hesitant especially because I think that these things should be caught with better regression tests and also because I can't see where exactly such test cases would fit into our current DIET/TED tests, but happy to hear your opinion (indeed, the default should be to have tests for everything important).

dakshvar22:
> model regression tests (just to be on the safe side, though we know they're not bulletproof)

@samsucik Can you create a new branch in the dataset repo where you add the car dataset as well and then run the MR tests on that dataset branch? At least for this bug we know that the car dataset catches it.

dakshvar22:
> what do you think about adding a test that checks that no dropout is applied to label features

I think it's a good idea but how would you write it? As far as I can see we'll have to rely on the `attribute_name` (or a similar argument) to check which attribute is the layer invoked / instantiated for and explicitly add an exception inside the layers constructor based on this info. I don't link how much we rely on the value of `attribute_name` in this method. Do you have any other thought?

samsucik:
/modeltest
```
dataset_branch: "car-dataset"
include:
 - dataset: ["all"]
   config: ["all"]
```

samsucik:
Ah, no, @dakshvar22 I think you misunderstood me. I didn't mean adding a check in the layer code or similar, but adding a _test case_. Though I think that would also require some relatively dirty tricks and assuming a lot about the private methods of the ML models, so I'm unsure if I wanna write such test cases. Improving our model regression tests sounds much better to me.

Btw I triggered the regression tests but apparently they take the list of datasets from the `main` branch of `training-data` (at least when using `all` datasets). So they ignore the newly added dataset. I'll let these regression tests finish and then will trigger again just for the `Car` dataset (I think the regression test script doesn't actually check that the dataset is in the list from `main`, so specifying the dataset by name should work even if it isn't in `main`).

dakshvar22:
> I didn't mean adding a check in the layer code or similar, but adding a test case. Though I think that would also require some relatively dirty tricks and assuming a lot about the private methods of the ML models, so I'm unsure if I wanna write such test cases. Improving our model regression tests sounds much better to me.

I also meant that you will have to do all of that to add a test case. I agree that we should improve the model regression tests by adding this dataset to the mix.

samsucik:
Ah, right, okay, so we are on the same page after all 😅  Yeah, adding this dataset is a cheap thing to do first, later on we could also create tests for checking the training curves.

samsucik:
I think the training curves and test-test scores show that the fix indeed works. Shown are results for 3 commits: [just before the buggy PR merge](https://github.com/RasaHQ/rasa/commit/00aad3213f34a8f3fad74b0b054fa9cb7cb7230b), [just after](https://github.com/RasaHQ/rasa/commit/ca66e3494524d2ce3a30c3b815da27c4d0a558d8), and [the last commit of this fixing PR](https://github.com/RasaHQ/rasa/pull/8607/commits/ae3757049226ca86619d48341160b477191ffc2f). All results are computed using only one (fixed) random seed.

Carbon bot 2.0 (with only Core test data):
plot-after-fix

Sara. The loops in ResponseSelector curves are due to multiple selectors being trained and me smoothing the curves with sliding average. Core not trained for the last commit due to a bug in rule-story validation which doesn't let the model finish training (if needed, I can still obtain the training curve by excluding other policies from the config):
plot-after-fix

The Car dataset (DIET only):
plot-after-fix

samsucik:
@dakshvar22 I'm not adding a changelog entry as the buggy PR hasn't been released at all so far (it was only planned for 2.6.0), hope that's fine.

Let me know if you need something more to finish the review (the model regression test results are still to come).

samsucik:
Looks like I can't use the Car dataset at all if it isn't listed on the `main` branch of `training-data`. I guess we'll have to go with the model regression test results from above + my locally obtained results. Later, we can merge the Car dataset into `main` on the `training-data` repo.

dakshvar22:
> Looks like I can't use the Car dataset at all if it isn't listed on the main branch of training-data

Why did you think so ;) Adding it to `dataset_config_mapping.json` should be enough. :)

Looking at the results - `Sparse + DIET(seq)` seems low right?

samsucik:
@dakshvar22 here are the 3 training curves for TED on Sara:
plot-after-fix

I'm trying to find a quick hack that would allow me to test the trained ensemble on the new test set (need to work around the bug mentioned [here](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1619180509183600?thread_ts=1615407488.423600&cid=C8LEHMTAM)) so that we can also get the test-set scores.

Btw as you can see, the Car dataset did get through in the end, but I'm unsure whether the numbers are of any value now that they can't really be compared to some previously obtained ones.

samsucik:
> Why did you think so 😉 

Because the regression tests literally errored out in the config-reading phase. Only after I re-triggered everything it (apparently) went through.

dakshvar22:
> Because the regression tests literally errored out in the config-reading phase

Fair point 🙂

samsucik:
Alright, @dakshvar22 has [verified](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1620291881043600?thread_ts=1620290398.039100&cid=C8LEHMTAM) that the `Sparse + DIET(seq)` config produces meaningful training (and test) numbers, so we shouldn't worry about it.

I'll add my bit: The test-set numbers on Sara. The ensemble uses only TED and RulePolicy and is tested in the `bad` (more challenging) portion of the new Sara test set.
plot-after-fix

samsucik:
Here are the training curves and test-set numbers for Sara where I added 2 more runs (with different random seeds) to the first and last of the 3 commits:
plot-after-fix

dakshvar22 (approved):
💯  Thanks for all the testing Sam!

PR #8608, opened 2021-05-05
Update `rasa data validate` to warn if `action_unlikely_intent` occurs in the training data.

### LINKED ISSUE #8563: `rasa data validate` should warn if `action_unlikely_intent` occurs in the training data
**Description of Problem**:
`rasa data validate` should warn if `action_unlikely_intent` occurs in the training stories (it's ok in rules), as the policies that use story data (`TEDPolicy`, `MemoizationPolicy`, `IntentTEDPolicy`) should not learn to predict it.

**Overview of the Solution**:

Just implement it in the data validation code.

alwx:
**Proposed changes**:
- fix #8563 


rasa/core/training/story_conflict.py
```diff
@@ -170,10 +175,18 @@ def find_story_conflicts(
...
+    )
+
     # Iterate once more over all states and note the (unhashed) state,
     # for which a conflict occurs
     conflicts = _build_conflicts_from_states(
-        trackers, domain, max_history, conflicting_state_action_mapping, tokenizer,
+        trackers,
+        domain,
+        max_history,
+        {**conflicting_state_action_mapping, **unpredictable_state_action_mapping},
```
ancalita, line 188: I like how you merged the two dictionaries here, neat 👍

rasa/core/training/story_conflict.py
```diff
@@ -242,6 +255,47 @@ def _find_conflicting_states(
...
+    """Identifies all states that contain actions that cannot be predicted.
+
+    Args:
+        trackers: Trackers that contain the states.
+        domain: The domain object.
+        max_history: Number of turns to take into account for the state descriptions.
+        tokenizer: A tokenizer to tokenize the user messages.
+
+    Returns:
+        A dictionary mapping state-hashes to a list of actions that cannot be predicted.
```
ancalita, line 285: This might be nitpicking: doesn't this return a mapping of state-hases to a list of action-hashes, rather than actions?

rasa/core/training/story_conflict.py
```diff
@@ -242,6 +255,47 @@ def _find_conflicting_states(
...
+    """
+    # Create a 'state -> list of actions' dict, where the state is
+    # represented by its hash
+    state_action_mapping = defaultdict(list)
+    for element in _sliced_states_iterator(trackers, domain, max_history, tokenizer):
+        hashed_state = element.sliced_states_hash
+        current_hash = hash(element.event)
+        if _unpredictable_action(element.event):
+            state_action_mapping[hashed_state] += [current_hash]
```
ancalita, line 294: Why does it have to store a list of action-hashes instead of a list of the actual action events? 🤔

ancalita (commented):
Have you tried testing this with CLI? I'm wondering if a custom warning should be printed if `action_unlikely_intent` gets used in stories.yml to tell the user why the validation failed. I think right now a generic message would be printed: "Project validation completed with errors." Might be wrong.

rasa/core/training/story_conflict.py
alwx: That's a good catch. Will update that!

rasa/core/training/story_conflict.py
alwx: Just to make it compatible with the existing function called `_build_conflicts_from_states` and to make it possible to create `StoryConflict` instances.

alwx:
Sure I checked, here is how the warning looks like:
```
2021-05-05 12:09:24 WARNING  rasa.validator  - Story structure conflict after intent 'greet':
  action_unlikely_intent predicted in 'happy path'
```

There is also a test to check if `action_unlikely_intent` is being printed in the warning message.

ancalita (approved):
Thanks for the explanations 😄 👍

rasa/core/training/story_conflict.py
wochinge: @alwx Sorry my review comment from this morning didn't make it through for some reason. 

What about doing as Anca said and iterating over `tracker.events` directly but then return a `StoryConflict` object right away instead of the hashed states? We could add this story conflict to the list of conflicts in line 184. We could also subclass `StoryConflict` to override its string representation as the current warning for `action_unlikely_intent` doesn't seem super helpeful. This would give us simpler code + better warnings.

PR #8610, opened 2021-05-05
Fix assign project workflow

m-vdb:
**Proposed changes**:
- In the `ci-model-regression-on-schedule` workflow, use `RASABOT_GITHUB_TOKEN` secret to create and label issues, otherwise workflows are not triggered ([docs here](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#using-the-github_token-in-a-workflow))
- In the `assign-default-project` listen to `labeled` events (fired when an issue is labeled) and:
  - enable `skip-update` to not update issues already added to a project board
  - bump the GH action (see [changes here](https://github.com/RasaHQ/create-or-update-project-card/pull/1)) to check for the existence of a project card in all columns of all projects of the org (instead of just all the columns of one project)


HotThoughts (approved):
LGTM 🚀 Busy Roberto 🤣

PR #8618, opened 2021-05-06
Add error message for rasa validate when text is null in a response

### LINKED ISSUE #7908: Rasa validator shows empty error message when text: null for a response
```
Rasa Version     : 2.2.2
Rasa SDK Version : 2.2.0
Rasa X Version   : None
Python Version   : 3.8.6
Operating System : macOS-10.16-x86_64-i386-64bit
```
**Issue**:
Originally reported in [our forum](https://forum.rasa.com/t/yamlvalidationexception-failed-to-validate-c-users-domain-yml-please-make-sure-the-file-is-correct-and-all-mandatory-parameters-are-specified-here-are-the-errors-found-during-validation/39970/14), training/validating threw an error but didn't show anything in the file to fix. After looking at the file, one of the responses had `text: null`. Changing this to a random string fixed the problem. We should look into why this error wasn't showing up in the traceback 

```
(e2e) chris@ChristophersMBP e2e % rasa data validate -d original-domain.yml
The configuration for pipeline was chosen automatically. It was written into the config file at 'config.yml'.
YamlValidationException: Failed to validate '/Users/chris/rasa/e2e/original-domain.yml'. Please make sure the file is correct and all mandatory parameters are specified. Here are the errors found during validation
```

```
rasa train
```
```
rasa validate
```

[File is here](https://forum.rasa.com/t/yamlvalidationexception-failed-to-validate-c-users-domain-yml-please-make-sure-the-file-is-correct-and-all-mandatory-parameters-are-specified-here-are-the-errors-found-during-validation/39970/15?u=desmarchris)

**Definition of Done** 
- [ ] `YamlValidationException` should show what made the schema validation fail

ancalita:
**Proposed changes**:
- Fix #7908 

### DIFF
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -168,19 +168,12 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
 
     try:
         c.validate(raise_exception=True)
-    except SchemaError as e:
-        for error in c.errors:
-            if error.msg in e.msg:
-                validation_errors = c.errors
-                break
-        else:
-            validation_errors = c.errors + [e]
-
+    except SchemaError:
         raise YamlValidationException(
             "Please make sure the file is correct and all "
             "mandatory parameters are specified. Here are the errors "
             "found during validation",
-            validation_errors,
+            c.errors,
             content=source_data,
         )
 
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -169,10 +169,12 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
     try:
         c.validate(raise_exception=True)
     except SchemaError as e:
-        if e not in c.errors:
-            validation_errors = c.errors + [e]
+        for error in c.errors:
+            if e.msg in error.msg:
+                validation_errors = c.errors
+                break
         else:
-            validation_errors = c.errors
+            validation_errors = c.errors + [e]
 
         raise YamlValidationException(
             "Please make sure the file is correct and all "
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -169,11 +169,16 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
     try:
         c.validate(raise_exception=True)
     except SchemaError as e:
+        if e not in c.errors:
+            validation_errors = c.errors + [e]
+        else:
+            validation_errors = c.errors
+
         raise YamlValidationException(
             "Please make sure the file is correct and all "
             "mandatory parameters are specified. Here are the errors "
             "found during validation",
-            c.errors + [e],
+            validation_errors,
             content=source_data,
         )
 
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -168,12 +168,12 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
 
     try:
         c.validate(raise_exception=True)
-    except SchemaError:
+    except SchemaError as e:
         raise YamlValidationException(
             "Please make sure the file is correct and all "
             "mandatory parameters are specified. Here are the errors "
             "found during validation",
-            c.errors,
+            c.errors + [e],
             content=source_data,
         )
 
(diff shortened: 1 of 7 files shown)


rasa/shared/utils/validation.py
```diff
@@ -168,12 +168,12 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
...
     try:
         c.validate(raise_exception=True)
-    except SchemaError:
+    except SchemaError as e:
         raise YamlValidationException(
             "Please make sure the file is correct and all "
             "mandatory parameters are specified. Here are the errors "
             "found during validation",
-            c.errors,
+            c.errors + [e],
```
alwx, line 176: I am not sure, but can't it lead to potential duplications?

ancalita: I could add `e` if e not in `c.errors`?

rasa/shared/utils/validation.py
```diff
@@ -168,12 +168,17 @@ def validate_yaml_schema(yaml_file_content: Text, schema_path: Text) -> None:
 
     try:
         c.validate(raise_exception=True)
-    except SchemaError:
+    except SchemaError as e:
+        if e not in c.errors:
```
joejuzl, line 172: Are these the same kind of errors? Are they even comparable?

ancalita: I couldn't find much about SchemaError in `pykwalify` documentation, but in their library they're the same kind (`c.errors` is a list of SchemaErrors)

ancalita: I double-checked and you're right @joejuzl , there is a nuance, `c.errors` is a list of `SchemaError.SchemaErrorEntry` which is an inner class. 

I can check `e.msg` against each message in `c.errors` to prevent duplication but it doesn't seem to cause any issues in our code if I add the `SchemaError` in the list of validation_errors alongside `c.errors` 🤔

I might have to change the type annotation to Union of SchemaError and SchemaErrorEntry in the `YamlValidationException`. WDYT?

joejuzl: I think it was fine how it was to begin with tbh. As we know what they are not duplicate now.

ancalita: I've actually kept the check at `msg` level because in some cases there might be duplicates based on what I could understand from `pykwalify` codebase - for example when the `SchemaError` that gets raised during validation includes the `c.errors` messages rather than it being a custom SchemaError that Rasa introduces through pykwalify extensions. I'm going to have to edit the check (reverse it) based on this insight.

joejuzl:
Ok so I looked a bit deeper into the pykwalify code... and I think the problem is with our extension. Because we `raise` it kicks out of the validation and doesn't add it to the errors. However if we were to return the error message instead, it would be collected as an error and end up in the final message.

ancalita:
> Ok so I looked a bit deeper into the pykwalify code... and I think the problem is with our extension. Because we `raise` it kicks out of the validation and doesn't add it to the errors. However if we were to return the error message instead, it would be collected as an error and end up in the final message.

@joejuzl Yup, that's what I understood as well, but didn't know you can return a Error 😄  so try changing L20 in `pykwalify_extensions.py` to a `return` and remove the check I have for `msg` ?

joejuzl:
> changing L20 in pykwalify_extensions.py to a return

I actually just returned a string value, but does returning the SchemaError also work?

ancalita:
@joejuzl yup it works with both string and SchemaError - the printed error message will differ just slightly:

- with string only:
```
Failed to validate YAML. Please make sure the file is correct and all mandatory parameters are specified. Here are the errors found during validation:
  in Line 7:
      Missing 'text' or 'custom' key in response or null 'text' value in response.. Path: /responses/utter_ask_name
```
- with SchemaError:
```
Failed to validate YAML. Please make sure the file is correct and all mandatory parameters are specified. Here are the errors found during validation:
  in Line 7:
      <SchemaError: error code 2: Missing 'text' or 'custom' key in response or null 'text' value in response.: Path: '/'>. Path: /responses/utter_ask_name
```
Which one do you think it's more informative for the user? I'd vote for 2nd since it makes it clear it's a schema related error, however it's not nicely formatted and that might confuse users?

joejuzl (approved):
💯

rasa/validator.py
```diff
@@ -247,6 +247,18 @@ def verify_nlu(self, ignore_warnings: bool = True) -> bool:
...
+            return False
+
+        for intent_key, intent_dict in self.domain.intent_properties.items():
+            if "triggers" in intent_dict:
+                rasa.shared.utils.io.raise_warning(
+                    f"The intent {intent_key} in the domain file is using the MappingPolicy format "
+                    f"which has now been deprecated. "
+                    f"Please migrate to RulePolicy."
+                )
+                return False
```
alwx, line 262: Nice 👍

alwx (approved):
🚀🚀🚀

Issue #8110, opened 2021-03-04
Improve errors in the Rasa transformer


samsucik:
**Description of Problem**: Currently, I'm able to get multiple errors that originate in the transformer code. Two errors I've encountered so far:

1.  for `ResponseSelector` (but probably also for DIET or TED) when I specify `number_of_transformer_layers > 0` but don't specify `transformer_size`, the code breaks because the default `transformer_size` is `None` which can't be used as a number in the code.
2.  If I then specify `transformer_size` but it isn't divisible by the number of transformer heads (which, by default, is 4), then, I get: `ValueError: number of units 50 should be proportional to number of attention heads 4.` I don't think "proportional" is the most understandable term here. Additionally, the docs never mention that `transformer_size` has to be a multiple of the number of heads.

**Overview of the Solution**: For both of the abovementioned errors, I can imagine checking the model config upon model instantiation in order to throw much more meaningful errors as soon as possible. Additionally, the docs should be updated.

**Definition of Done**:

*   [ ] Meaningful exceptions are being thrown early in the pipeline
*   [ ] Tests are added
*   [ ] The docs mention the constraints which, if not met, lead to the exceptions

PR #8652, opened 2021-05-11
Improve response selector defaults when transformer is enabled

samsucik:
**Proposed changes**:
- Raise warnings when `number_of_transformer_layers > 0` is used with a ResponseSelector but:
  -  `transformer_size` is not specified or is set to a non-positive value
  - `hidden_layers_sizes` is not specified or is set such that the hidden layers are not disabled (we recommend disabling hidden layers when using a transfomer)
- Set sensible default `transformer_size` in the 1st case from above. Previously, the default remained `None` which was breaking transformer training down the line.

**To-do**:
- [x] check that `transformer_size=256` and `num_transformer_layers=2` are sensible default values -- they were previously used until [this commit](https://github.com/RasaHQ/rasa/commit/792a6a9a5722669e197c2e1041fade48204d6331#diff-a09a171b147b1e582bc05975958773927f94a31316c29e77b8290ffff0d792adR101) but back then no `HIDDEN_LAYERS_SIZES` were specified (today we've got `{TEXT: [256, 128], LABEL: [256, 128]}`)
    - Turned out we don't need a default `num_transformer_layers` because the user should always set this by themself to activate or de-activate the transformer
- [x] decide if we want to set `transformer_size` to a value compatible with `num_heads` (if provided) or always to a fixed default value
    - Decision: go with a fixed size; in a separate issue improve the config checking so that incompatible `num_heads` and `transformer_size` are warned about early (see #8110 )


samsucik:
After brief [discussion with Daksh](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1621250497288200), I'm now checking the ResponseSelector config regardless of the `use_text_as_label` value.

tttthomasssss:
@samsucik a question regarding your note
> Turned out we don't need a default num_transformer_layers because the user should always set this by themself to activate or de-activate the transformer

Do you think we need a less opaque way of enabling/disabling the transformer (maybe just a note in the docs)? I don't have any strong opinions either way but relying on `num_transformer_layers=0` to mean `transformer=False` seems a bit implicit to me.

changelog/8046.bugfix.md
```diff
@@ -0,0 +1,3 @@
+`ResponseSelector` can now be trained with the transformer enabled (i.e. when a positive
+number of transformer layers is provided) even if one doesn't specify the transformer's
```
tttthomasssss, line 2: Can you name the attribute explicitly (e.g. `num_transformer_layers>0`)?

samsucik:
> Do you think we need a less opaque way of enabling/disabling the transformer (maybe just a note in the docs)?

We use this kind of pattern in multiple places and I'm never sure if it's accessible enough... I guess this could be part of 3.0 discussions (config-related) because it's also about what kind of users do we imagine tweaking all the config options of our models. For now, I'll add an explicit note in the docs.

rasa/nlu/selectors/response_selector.py
```diff
@@ -308,9 +311,88 @@ def _load_selector_params(self, config: Dict[Text, Any]) -> None:
...
         self.use_text_as_label = config[USE_TEXT_AS_LABEL]
 
+    def _warn_about_transformer_and_hidden_layers_enabled(
+        self, selector_name: Text
+    ) -> None:
+        """Warns user if they enabled the transformer but didn't disable hidden layers.
+
+        ResponseSelector defaults specify considerable hidden layer sizes, but
+        this is for cases where no transformer is used. If a transformer exists,
+        then the best results are chieved with no hidden layers used between the
```
tttthomasssss, line 321: `[...] are achieved [...]`

tttthomasssss:
Another related question - have we got similar consistency checks for `DIETClassifier` and `TEDPolicy` regarding a discrepancy between `transformer_size` and `number_of_transformer_layers`? I understand in the case of the `ResponseSelector` its easier to get an inconsistency because the transformer is switched off by default, but the same kind of behaviour can happen for the other two, right?

I don't think this should be tackled within the scope of this PR, but we might want to add an extra issue for this if you think this is a problem.

samsucik:
No, we don't have these checks for TED or DIET. Both have a reasonable default values for `NUM_TRANSFORMER_LAYERS` and `TRANSFORMER_SIZE`, so the user would have to explicitly provide bad values in order for things to break. I'd say it's worth adding a check but it's a very, very low priority...

tttthomasssss (approved):
Awesome work 🎸  - there are only the minor comments I've added, so I'll approve it right away.

Issue #8364, opened 2021-04-05
form replaces list slot with length 1 with the item in the list


amn41:
User raised this on the [forum](https://forum.rasa.com/t/slot-type-list-and-text/41187/6) and I was able to reproduce using their example code. 

You can see this happen if you run with `rasa shell --debug`

```
2021-04-05 16:15:28 DEBUG    rasa.core.processor  - Predicted next action 'order_pizza_form' with confidence 1.00.
2021-04-05 16:15:28 DEBUG    rasa.core.actions.forms  - Activated the form 'order_pizza_form'.
2021-04-05 16:15:28 DEBUG    rasa.core.actions.forms  - Validating pre-filled required slots: {'toppings': ['pepperoni']}
2021-04-05 16:15:28 DEBUG    rasa.core.actions.forms  - Validating extracted slots: {'toppings': ['pepperoni']}
2021-04-05 16:15:28 DEBUG    rasa.core.actions.forms  - Validating user input 'UserUttered(text: I want a pepperoni pizza, intent: pizza, entities: pepperoni (Type: toppings, Role: None, Group: None), use_text_for_featurization: False)'.
2021-04-05 16:15:28 DEBUG    rasa.core.actions.forms  - Extracted 'pepperoni' for extra slot 'toppings'.
2021-04-05 16:15:28 DEBUG    rasa.core.actions.forms  - Validating extracted slots: {'toppings': 'pepperoni'}
2021-04-05 16:15:28 DEBUG    rasa.core.actions.forms  - Deactivating the form 'order_pizza_form'
```

```
Rasa Version     : 2.3.1
Rasa SDK Version : 2.3.1
Rasa X Version   : 0.37.0
Python Version   : 3.7.6
Operating System : Darwin-19.6.0-x86_64-i386-64bit
```

PR #8675, opened 2021-05-12
fix slot filling for `ListSlot`s in case only a single matching entity was extracted

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/8364
- The issue was that we decided whether something should be filled with a list based on how many matching entities were extracted. This means if we e.g. asked for pizza toppings and a user only wants "mushrooms" we would have set the slot with the string "mushroom" instead of `[mushroom]`. This behavior is fine unless it's for `ListSlot`s as these should (as the name suggests) always be `List`s
- we are now checking the slot type before deciding whether to convert the entities to single items or not
- I also modified the setter for `ListSlot` so that items which aren't `List`s are automatically put into a list to ensure `ListSlot` values are always `List`s

**Things I didn't do**
- I considered adding a warning if there are multiple entities for a slot type which isn't a `ListSlot`. I didn't do so as this might lead to false warnings for users who are using [custom slots](https://rasa.com/docs/rasa/domain#custom-slot-types)

### DIFF
--- a/rasa/shared/core/slots.py
+++ b/rasa/shared/core/slots.py
@@ -276,6 +276,8 @@ class ListSlot(Slot):
 
 
 class UnfeaturizedSlot(Slot):
+    """Deprecated slot type to represent slots which don't influence conversations."""
+
     type_name = "unfeaturized"
 
     def __init__(
--- a/rasa/shared/core/slots.py
+++ b/rasa/shared/core/slots.py
@@ -265,6 +265,15 @@ class ListSlot(Slot):
             # we couldn't convert the value to a list - using default value
             return [0.0]
 
+    @Slot.value.setter
+    def value(self, value: Any) -> None:
+        """Sets the slot's value."""
+        if not isinstance(value, list):
+            # Make sure we always store list items
+            value = [value]
+
+        Slot.value.fset(self, value)
+
 
 class UnfeaturizedSlot(Slot):
     type_name = "unfeaturized"
--- a/rasa/shared/core/slots.py
+++ b/rasa/shared/core/slots.py
@@ -288,6 +288,18 @@ class UnfeaturizedSlot(Slot):
         auto_fill: bool = True,
         influence_conversation: bool = False,
     ) -> None:
+        """Creates unfeaturized slot.
+
+        Args:
+            name: The name of the slot.
+            initial_value: Its initial value.
+            value_reset_delay: After how many turns the slot should be reset to the
+                initial_value. This is behavior is currently not implemented.
+            auto_fill: `True` if it should be auto-filled by entities with the same
+                name.
+            influence_conversation: `True` if it should be featurized. Only `False`
+                is allowed. Any other value will lead to a `InvalidSlotConfigError`.
+        """
         if influence_conversation:
             raise InvalidSlotConfigError(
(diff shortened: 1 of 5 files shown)


tests/core/actions/test_forms.py
```diff
@@ -808,6 +809,66 @@ def test_extract_requested_slot_when_mapping_applies(
...
+        # Two entities were extracted for `ListSlot``
+        (
+            [
+                {"entity": "topping", "value": "mushrooms"},
+                {"entity": "topping", "value": "kebab"},
+            ],
+            ["mushrooms", "kebab"],
+        ),
+        # One entities was extracted for `ListSlot``
+        ([{"entity": "topping", "value": "kebab"},], ["kebab"],),
```
wochinge, line 824: I like my pizza with kebab topping

tests/core/actions/test_forms.py
```diff
@@ -1298,6 +1359,64 @@ def test_extract_other_slots_with_entity(
...
+            SlotSet(REQUESTED_SLOT, "some slot"),
+            UserUttered(
+                "bla", intent={"name": "greet", "confidence": 1.0}, entities=entities
+            ),
+            ActionExecuted(ACTION_LISTEN_NAME),
+        ],
+        slots=domain.slots,
+    )
+
+    slots = form.extract_other_slots(tracker, domain)
```
ancalita, line 1416: This test appears almost identical to the one above `test_extract_requested_slot_with_list_slot` except for the different form extract methods - why is this 2nd one required?

rasa/shared/core/slots.py
```diff
@@ -265,8 +265,19 @@ def _as_feature(self) -> List[float]:
...
             return [0.0]
 
+    @Slot.value.setter
+    def value(self, value: Any) -> None:
+        """Sets the slot's value."""
+        if not isinstance(value, list):
+            # Make sure we always store list items
+            value = [value]
+
+        Slot.value.fset(self, value)
```
ancalita, line 275: PyCharm compains of unexpected argument `value` in this line, I'm also not familiar with `fset` to understand why, but wouldn't it be easier just to assign `self._value = value`?
Also since this overrides the value.setter in Slot interface, shouldn't it also include `self._has_been_set = True`?

tests/core/actions/test_forms.py
wochinge: The first one tests the `ListSlot` extraction for the slot which was request the form - this test tests the extraction of slots which weren't requested by the form but accidentally filled by the user

rasa/shared/core/slots.py
wochinge: Thanks for the hint! Fixed it and clarified the `fset` part. I'm calling the superclass property setter here to avoid duplicating `Slot`s `value.setter` logic. You can find more on it [here](https://docs.python.org/3/library/functions.html#property) or in [this Stackoverflow thread](https://stackoverflow.com/questions/44164816/overload-a-property-setter-in-python?noredirect=1&lq=1)

ancalita: Oh I see, so that's why there's no need to set `self._has_been_true` to `True`.

ancalita (approved):
👍

PR #8684, opened 2021-05-13
Handle empty slot mappings in domain forms

### LINKED ISSUE #8590: Rasa shell crashes if form with only custom slot mappings is defined
**Rasa version**:
2.5.1
**Python version**:
3.8.7
OSX

**Issue**:

If a user wants to define a form with only custom slot mappings, then `required_slots` for that form in the domain is empty. This results in a crash when we try to access `REQUIRED_SLOTS` for that form. We need to either force at least one slot with a pre-defined mapping for a form defined in a domain, or deal with this case more gracefully in `domain.slot_mapping_for_form`:
```
    def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
        """...
        """
        form = self.forms.get(form_name, {})
        if REQUIRED_SLOTS_KEY in form:
            return form[REQUIRED_SLOTS_KEY]
        else:
            return {}
```
```
21-04-30 19:13:55 ERROR    rasa.core.processor  - Encountered an exception while running action 'contract_form'.Bot will continue, but the actions events are lost. Please check the logs of your action server for more information.
  File "/home/jsv/Documents/CARRERA/tfg4/resources/rasa-chatbot/.env/lib/python3.6/site-packages/rasa/core/processor.py", line 773, in _run_action
    output_channel, nlg, temporary_tracker, self.domain
  File "/home/jsv/Documents/CARRERA/tfg4/resources/rasa-chatbot/.env/lib/python3.6/site-packages/rasa/core/actions/loops.py", line 26, in run
    events += await self.activate(output_channel, nlg, tracker, domain)
  File "/home/jsv/Documents/CARRERA/tfg4/resources/rasa-chatbot/.env/lib/python3.6/site-packages/rasa/core/actions/forms.py", line 706, in activate
    for slot_name in self.required_slots(domain):
  File "/home/jsv/Documents/CARRERA/tfg4/resources/rasa-chatbot/.env/lib/python3.6/site-packages/rasa/core/actions/forms.py", line 58, in required_slots
    return list(domain.slot_mapping_for_form(self.name()).keys())
AttributeError: 'NoneType' object has no attribute 'key
```

```
rasa shell
```

```yml

```

```yml
forms:
    contract_form:
```

```
forms:
  tobi_form:
    required_slots:
```

ancalita:
**Proposed changes**:
- Fix #8590 

### DIFF
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -1860,7 +1860,11 @@ class Domain:
         Returns:
             The slot mapping or an empty dictionary in case no mapping was found.
         """
-        return self.forms.get(form_name, {})[REQUIRED_SLOTS_KEY]
+        form = self.forms.get(form_name)
+        if form:
+            return form.get(REQUIRED_SLOTS_KEY)
+
+        return {}
 
 
 class SlotMapping(Enum):
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -1862,7 +1862,7 @@ class Domain:
         """
         form = self.forms.get(form_name)
         if form:
-            return form.get(REQUIRED_SLOTS_KEY)
+            return form[REQUIRED_SLOTS_KEY]
 
         return {}
 
--- a/changelog/8590.bugfix.md
+++ b/changelog/8590.bugfix.md
@@ -1,0 +1,1 @@
+Handle `AttributeError ` thrown by empty slot mappings in domain form through refactoring.
--- a/tests/shared/core/test_domain.py
+++ b/tests/shared/core/test_domain.py
@@ -1424,3 +1424,26 @@ def test_domain_count_conditional_response_variations():
     )
     count_conditional_responses = domain.count_conditional_response_variations()
     assert count_conditional_responses == 5
+
+
+def test_domain_with_no_form_slots():
+    domain = Domain.from_yaml(
+        """
+        version: "2.0"
+        forms:
+          contract_form:
+        """
+    )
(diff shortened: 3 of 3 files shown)


rasa/shared/core/domain.py
```diff
@@ -1860,7 +1860,11 @@ def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
         Returns:
             The slot mapping or an empty dictionary in case no mapping was found.
         """
-        return self.forms.get(form_name, {})[REQUIRED_SLOTS_KEY]
+        form = self.forms.get(form_name)
+        if form:
+            return form.get(REQUIRED_SLOTS_KEY)
```
joejuzl, line 1865: This will return `None` if there is no `REQUIRED_SLOTS_KEY` - should it be an empty dict?

joejuzl (changes requested):

rasa/shared/core/domain.py
ancalita: That's the interesting bit:  when the form has custom slot mappings, but without the `REQUIRED_SLOTS_KEY` explicitly mentioned in `domain.yml`, the form will contain this key because of the transformation that happens in [here](https://github.com/RasaHQ/rasa/blob/f1e19f5e2853880873bbfcec828fb54b7ccb12bf/rasa/shared/core/domain.py#L707). Do you still think anything should be changed in these conditions?

joejuzl: ah ok yes I see. Then maybe `form[REQUIRED_SLOTS_KEY]` is more explicit?

joejuzl (approved):

PR #8697, opened 2021-05-14
Handling missing dependency exception for tensorflow

Archish27:
**Proposed changes**:
- Handling missing dependency exception for tensorflow
- fixes: #8682 


sara-tagger:

rasa/utils/train_utils.py
```diff
@@ -471,7 +473,13 @@ def create_common_callbacks(
     Returns:
         A list of callbacks.
     """
-    import tensorflow as tf
+    try:
+        import tensorflow as tf
+    except ModuleNotFoundError:
+        raise MissingDependencyException(
+            f"Tensorflow does not seem to be installed, but it is needed for this "
+            f"{DOCS_BASE_URL}/installation"
```
tttthomasssss, line 481: Could you make the error message more descriptive (example usage of `MissingDependencyException can be found in [cli/x.py](https://github.com/RasaHQ/rasa/blob/1aa1b6c93ba2603064737814f9fcdce764af40d9/rasa/cli/x.py#L470) and [nlu/components.py](https://github.com/RasaHQ/rasa/blob/1aa1b6c93ba2603064737814f9fcdce764af40d9/rasa/nlu/components.py#L87))? Also, the general documentation page linked in the message doesn't mention tensorflow, could you therefore add something along the lines of `For an overview of how to install Rasa, please visit URL`.

tttthomasssss (changes requested):
Thanks very much for your PR 🚀 !

Left a comment regarding the raised error message.

rasa/utils/train_utils.py
Archish27: Kept it simple :)

rasa/utils/train_utils.py
```diff
@@ -471,7 +473,14 @@ def create_common_callbacks(
...
     """
-    import tensorflow as tf
+    try:
+        import tensorflow as tf
+    except ModuleNotFoundError:
+        raise MissingDependencyException(
+            f"Tensorflow does not seem to be installed, but it is needed for this."
+            f"For an overview of how to install Rasa, please visit URL:"
+            f"{DOCS_BASE_URL}/installation"
+        )
```
wochinge, line 483:
```suggestion
    except ModuleNotFoundError as e:
        raise MissingDependencyException(
            f"Tensorflow does not seem to be installed, but it is needed for this."
            f"For an overview of how to install Rasa, please visit URL:"
            f"{DOCS_BASE_URL}/installation"
        ) from e
```

rasa/utils/train_utils.py
```diff
@@ -471,7 +473,14 @@ def create_common_callbacks(
     Returns:
         A list of callbacks.
     """
-    import tensorflow as tf
+    try:
+        import tensorflow as tf
+    except ModuleNotFoundError as e:
+        raise MissingDependencyException(
+            f"Tensorflow does not seem to be installed, but it is needed for this."
```
wochinge, line 480: Should we also mention [this page](https://www.tensorflow.org/install/errors) ? TensorFlow will always be installed as part of Rasa Open Source - so if the import fails it's usually a TensorFlow problem with the underlying OS than a Rasa Open Source problem.

wochinge:
One thing: This solution will break as soon as another module with a tensorflow import gets imported before the `train_utils.py` one. Maybe we can think of a more robust solution here 🤔

Archish27:
So what do you suggest should we create generic import statements & then accordingly handle imports?

rasa/utils/train_utils.py
ancalita: @wochinge I don't see the specific error caught by Sentry in this list - but we could redirect to this GH Tensorflow [thread ](https://github.com/tensorflow/tensorflow/issues/44110) for resolution? In this case do you think it would be a good idea to create `RasaException` subclass called maybe `ImportException` instead of using `MissingDependencyException` since it would indicate the wrong thing as you just stated?

ancalita (changes requested):
@Archish27 thanks for your contribution 🚀  
This code should be lifted to a different file: `rasa.utils.tensorflow.callback` [L5.](https://github.com/RasaHQ/rasa/blob/d4d606b12370d440d9f9ed3f9e4d8e2a23b959b8/rasa/utils/tensorflow/callback.py#L5)

Additionally, the actual exception is `ImportError` not `ModuleNotFoundError`, please amend. @wochinge would this be a more robust solution?

ancalita:
@Archish27 one more thing - could you also please add a [changelog entry](https://github.com/RasaHQ/rasa/tree/main/changelog), this would be classified as a `misc` entry, the details are in the README 🙏

wochinge:
> @wochinge would this be a more robust solution?

Mhm, not really as we still can't be sure this is the first import. How about filtering for `ImportError`s in our `telemetry` module? Catching all `ImportError`s seems a bit too brute force - we could e.g. filter as described [here](https://docs.sentry.io/platforms/python/configuration/filtering/) . What do you think? That would make it independent from where TensorFlow is imported.

ancalita:
> > @wochinge would this be a more robust solution?
> 
> Mhm, not really as we still can't be sure this is the first import. How about filtering for `ImportError`s in our `telemetry` module? Catching all `ImportError`s seems a bit too brute force - we could e.g. filter as described [here](https://docs.sentry.io/platforms/python/configuration/filtering/) . What do you think? That would make it independent from where TensorFlow is imported.

@wochinge Sure, this does sound like a complete solution 👍 
To clarify on two points:

- the user will still be able to see the error, but this solution will not send it to Sentry - did I understand that right?
- based on the Sentry docs you shared, I gather the change that needs to happen is add the below check before [L603](https://github.com/RasaHQ/rasa/blob/db3cb7d52ab30e39633a0b0fdfd85f64b66178df/rasa/telemetry.py#L603).
```
    # drop ImportError events as these constitute an usage error
    if 'exc_info' in _unused_hint:
        exc_type, exc_value, tb = hint['exc_info']
        if isinstance(exc_value, Import Error):
            return None
```
The only concern I have that I don't know how to solve for is that the existing function name obscures the new filtering that is added. Would updating the docstrings be enough for clarification?

wochinge:
> the user will still be able to see the error, but this solution will not send it to Sentry - did I understand that right?

Exactly (same as with any (Rasa)exception) 👍🏻 

> based on the Sentry docs you shared, I gather the change that needs to happen is add the below check before

Tbh I've never implemented one before but looks about right 👍🏻 

> The only concern I have that I don't know how to solve for is that the existing function name obscures the new filtering that is added. Would updating the docstrings be enough for clarification?

Very good point 💯 I'd introduce something like a new function `before_send` which merges both functionalities, e.g.:

```python
def before_send(event: Dict[Text, Any], _unused_hint: Optional[Dict[Text, Any]] = None) -> Optional[Dict[Text, Any]]:
  event = filter_tf_errors(event)
  event = strip_sensitive_data_from_sentry_event(event)
  return event
```

ancalita:
>  I'd introduce something like a new function `before_send` which merges both functionalities, e.g.:

Sounds good! How would we test we don't introduce unintended bugs - for example if after TF error filtering, the event is None, there might be an exception raised in `strip_sensitive_data_from_sentry_event` in [L582](https://github.com/RasaHQ/rasa/blob/db3cb7d52ab30e39633a0b0fdfd85f64b66178df/rasa/telemetry.py#L582) 🤔 . Maybe it's better to reverse the order of the function calls in `before_send`.

@Archish27 Please let us know if the above discussion makes sense for you and if you're happy to implement these changes?
The only thing not mentioned is that the new `before_send` function would have to replace `strip_sensitive_data_from_sentry_event` [here](https://github.com/RasaHQ/rasa/blob/db3cb7d52ab30e39633a0b0fdfd85f64b66178df/rasa/telemetry.py#L635).

wochinge:
> Maybe it's better to reverse the order of the function calls in before_send.

Excellent idea!

changelog/8682.misc.md
```diff
@@ -0,0 +1 @@
+Filtering import errors before sending to sentry
```
ancalita, line 1: I would format this in active voice, e.g. 'Filters out `ImportError` from Sentry tracking'.

rasa/telemetry.py
```diff
@@ -563,6 +563,27 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
 
     rasa_utils.write_global_config_value(CONFIG_FILE_TELEMETRY_KEY, configuration)
```
ancalita, line 565: In order to resolve both E302 errors, please leave 2 lines empty between two functions.

rasa/telemetry.py
```diff
@@ -563,6 +563,27 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
 
+def filter_errors(event: Dict[Text, Any], _unused_hint: Optional[Dict[Text, Any]] = None) -> None:
```
ancalita, line 566: Two things here need amending:

- please rename function to either `filter_import_errors` or `filter_tf_errors`
- return type is actually `Optional[Dict[Text, Any]]`

rasa/telemetry.py
```diff
@@ -563,6 +563,27 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+    """Filtering errors.
+
+    Args:
+        event: event to be logged to sentry
+        _unused_hint: some hinting information sent alongside of the event
+
+    Returns:
+        the event without any sensitive / PII data or `None` if the event should
```
ancalita, line 574: I would add add here "or`None` if the event constitutes an `ImportError` which should be discarded".

rasa/telemetry.py
```diff
@@ -563,6 +563,27 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+    """
+    if 'exc_info' in _unused_hint:
+        exc_type, exc_value, tb = _unused_hint['exc_info']
+        if isinstance(exc_value, ImportError):
+            return None
+    return event
+
+def before_send(event: Dict[Text, Any], _unused_hint: Optional[Dict[Text, Any]] = None) -> Optional[Dict[Text, Any]]:
+  event = filter_errors(event, _unused_hint)
+  event = strip_sensitive_data_from_sentry_event(event, _unused_hint)
```
ancalita, line 585: As discussed in the thread, it's best to re-order these function calls, so that `strip_sensitive_data_from_sentry_event` runs first, then `filter_errors`.

rasa/telemetry.py
Archish27: Kept it as filter_errors going forward we can utilize this function if any other errors that we need to handle. What are your thoughts on it?

rasa/telemetry.py
Archish27: How did I miss this? Valid point though. Nice catch 🚀

rasa/telemetry.py
```diff
@@ -574,8 +602,8 @@ def strip_sensitive_data_from_sentry_event(
         _unused_hint: some hinting information sent alongside of the event
 
     Returns:
-        the event without any sensitive / PII data or `None` if the event should
-        be discarded.
+        the event without any sensitive / PII data or `None` if the event constitutes
+        an `ImportError` which should be discarded.
```
ancalita, line 606: This amendment in the docstring was added in the wrong function - you should move it to `filter_errors`.

rasa/telemetry.py
```diff
@@ -564,6 +564,34 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+    if "exc_info" in hint:
+        exc_type, exc_value, tb = hint["exc_info"]
+        if isinstance(exc_value, ImportError):
+            return None
+    return event
+
+
+def before_send(
+    event: Dict[Text, Any], _unused_hint: Optional[Dict[Text, Any]] = None
+) -> Optional[Dict[Text, Any]]:
```
ancalita, line 589: Please add a docstring here as per CI requirements.

rasa/telemetry.py
ancalita: Yes, true, better not to pin the name to a specific error. However, since the type of error that would be handled here would always be usage errors, I'd still rename this to `filter_usage_errors`.

ancalita:
Hi @Archish27 hope all is well - checking in if you're ok with making the last final tweaks?

Archish27:
@ancalita Updated, ready to re-review

rasa/telemetry.py
ancalita: Almost ready 💯 - there are still a few CI Code Quality checks that need to be fixed first for this docstring in L593, I think in this line you need to provide a short summary of what the function does, e.g. `Removes PII data and usage errors.`

ancalita:
Hi @Archish27 I have a big favour to ask 🙏  Could you also please rebase your branch and PR to target `2.8.x`? 2.8 was the last minor we release before the 3.0 release later this year. We'd love to include your change in one of the micro releases, rather than wait for 3.0 release. Let us know if you have any q's.

ancalita:
Hi @Archish27 I haven't heard back from you in a month - are you still interested in completing this PR?
If I don't hear back from you by Fri 20th August, I will close this PR, you are still welcome to open it later if you want to continue making progress on it.

Archish27:
@ancalita Hey, I missed this. Can you let me know anything pending? Do you want me to point to rasa `2.8`?

rasa/telemetry.py
Archish27: updated

changelog/8682.misc.md
```diff
@@ -0,0 +1 @@
+Filters out ImportError from Sentry tracking
```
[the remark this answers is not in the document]
Archish27: fixed

changelog/8682.misc.md
Archish27: updated

rasa/telemetry.py
```diff
@@ -567,6 +567,43 @@ def toggle_telemetry_reporting(is_enabled: bool) -> None:
...
+    """Strips the sensitive data and filters errors before sending to sentry.
```
ancalita, line 593: Great stuff, thank you for the quick action and rebasing! In order for the Code Quality check to pass, the below flag needs to be resolved - a blank line needs to follow L593.

Archish27: Aaahh :( How am I missing this stuff my bad fixing it.

Archish27: Sorry for delays & back and forth will take care of this for next time :)

Archish27:
@ancalita All green ✅ Ready to merge 🚀

ancalita (approved):
🎉

ancalita:
@tttthomasssss we also need your approval on this PR in order to merge it - could you please have a look?

tttthomasssss (approved):
LGTM

ancalita:
@Archish27 All approvals are in, all it's required now is to update your branch with the `2.8.x` latest changes and then we can merge it 👍🏼

Issue #8625, opened 2021-05-06
rasa train emits a domain error without additional detail when using an incorrect field name in an utter


bastbnl:
**Rasa version**: 
```
Rasa Version      :         2.5.1
Minimum Compatible Version: 2.2.0a1
Rasa SDK Version  :         2.5.0
Rasa X Version    :         None
Python Version    :         3.7.10
Operating System  :         Linux-4.19.0-16-amd64-x86_64-with-debian-10.9
Python Path       :         /home/bastb/bot/venv/bin/python3
```

**Rasa SDK version** (if used & relevant): 
```
(venv) bastb@i7-nuc:~/bot/bot_nld$ pip freeze | grep rasa-sdk
rasa-sdk==2.5.0
```

**Rasa X version** (if used & relevant):

**Python version**:
```
(venv) bastb@i7-nuc:~/bot/bot_nld$ python --version
Python 3.7.10
```

**Operating system** (windows, osx, ...):
`Linux i7-nuc 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux`

**Issue**:
I was working on `domain.yml` and used an incorrect key below an `utter_`, like this:
```
utter_request_cancelled:
- text: yadieyadie ya
- utter: different words, same message 

```

`rasa train` then reports an expected error, but with very little detail:
```
(venv) bastb@i7-nuc:~/bot/bot_nld$ rasa train
YamlValidationException: Failed to validate '/home/bastb/bot/bot_nld/domain.yml'. Please make sure the file is correct and all mandatory parameters are specified. Here are the errors found during validation

```

**Error (including full traceback)**:
```
YamlValidationException: Failed to validate '/home/bastb/bot/bot_nld/domain.yml'. Please make sure the file is correct and all mandatory parameters are specified. Here are the errors found during validation

```

**Command or request that led to error**:
```
rasa train
```

**Content of configuration file (config.yml)** (if relevant):
```yml

```

**Content of domain file (domain.yml)** (if relevant):
```yml

```

The source is easily found when running `rasa train -v`, but I think it would be nice if the error message itself contained some additional detail

sara-tagger:
Thanks for raising this issue, @jupyterjazz will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

Imod7:
[...]

PR #8698, opened 2021-05-14
Raises an error when an undefined key is used in domain.yml

bastbnl:
I validated this using a working domain.yml using `rasa train --dry-run`

### DIFF
--- a/rasa/shared/nlu/training_data/schemas/responses.yml
+++ b/rasa/shared/nlu/training_data/schemas/responses.yml
@@ -11,6 +11,7 @@ schema;responses:
       - type: "map"
         required: True
         allowempty: False
+        matching: any
         mapping:
           text:
             type: "str"


rasa/shared/nlu/training_data/schemas/responses.yml
```diff
@@ -11,6 +11,7 @@ schema;responses:
       - type: "map"
         required: True
         allowempty: False
+        matching: any
```
Imod7, line 14: Hey @bastbnl thank you for submitting a PR for this issue! I tried this change locally but still the error message is the same as before and does not provide further/more detailed information about why the `domain.yml` file is not valid. Did it work for you? I use a non valid `domain.yml` that just has the same invalid key that you used in the description of the issue [here](https://github.com/RasaHQ/rasa/issues/8625#issue-877880747).

bastbnl:
No need to apply, has been resolved in rasa==2.6.1

PR #8704, opened 2021-05-17
Pickling exceptions

Mazyod:
**Proposed changes**:
- Fix error thrown when attempting to pickle `UnsupportedModelError` and `InvalidModelError`

I needed to pass exceptions from background processes to be handled in the main process, but it was these exceptions were causing issues during pickling. The fix is easy, and hopefully a better practice especially for a project like rasa which relies on multiprocessing.

Previous error:

```python
def test_exception_pickling():
        exception = UnsupportedModelError("test run")
>       cycled_exception = pickle.loads(pickle.dumps(exception))
E       TypeError: __init__() missing 1 required positional argument: 'message'
```


Mazyod:
Another option is to make `message` an optional argument:
https://stackoverflow.com/a/16245182/456434

Mazyod:
would appreciate any feedback on this PR

tttthomasssss (changes requested):
Hey, apologies this took so long. Thanks very much for submitting this PR.

All looks good codewise - great work 🎸 !  Could you just expand on the purpose of being able to pickle the exceptions in the `changelog`? I.e. just a line or two on why this is needed and what use-case it covers.

Mazyod:
Appreciate it @tttthomasssss. Certainly, I have added the purpose for making exceptions pickleable. I hope the explanation would suffice, otherwise I could try adding more information.

tttthomasssss (approved):
Awesome, thanks very much for the lightning fast resolution 🚀

PR #8711, opened 2021-05-18
avoid unnecessary retrain caused by "false" miscomparison

cmanovit:
model.should_retrain was being called twice and on the second time there could be a collision on /nlu created from the first run, creating a false failure in comparison step.

**Proposed changes**:
- ...


sara-tagger:
Thanks for submitting a pull request 🚀 @joejuzl will take a look at it as soon as possible ✨

joejuzl:
Hi @cmanovit thanks for this PR!

Could you give a little detail on what you mean by a "collision", and how to replicate this bug? 
Also a unit test that demonstrates the bug, and shows it is now fixed, would be amazing!

cmanovit:
Hi @joejuzl . Thanks for getting back to me.

By collision, I mean:

2021-06-21 08:00:58 DEBUG    root  - Could not merge model: Destination path '/tmp/tmpuut69u_8/core/core' already exists
2021-06-21 08:00:58 DEBUG    root  - /tmp/tmpnhn1craa/nlu
2021-06-21 08:00:58 DEBUG    root  - ['metadata.json', 'component_3_DIETClassifier.entity_tag_specs.json', 'component_3_DIETClassifier.index_label_id_mapping.json', 'nlu', 'component_3_DIETClassifier.tf_model.index', 'checkpoint', 'component_3_DIETClassifier.tf_model.data-00000-of-00001', 'component_3_DIETClassifier.label_data.pkl', 'component_3_DIETClassifier.data_example.pkl']
2021-06-21 08:00:58 DEBUG    root  - /tmp/tmpuut69u_8/nlu
2021-06-21 08:00:58 DEBUG    root  - ['nlu', 'core']
**2021-06-21 08:00:58 DEBUG    root  - Could not merge model: Destination path '/tmp/tmpuut69u_8/nlu/nlu' already exists**
Training NLU model...

Note that there are two levels of "nlu".

To replicate this bug, we can use "moodbot" under rasa/examples/moodbot.

1. rasa train
2. Modify domain.yml, say, by adding following line under "utter_goodbye":
  - text: "goodbye" 
3. rasa train
We wouldn't see unnecessary retrain of NLU yet but if we look in the model.tar.gz we will start seeing the issue where core and nlu are now included twice in the package. (We can see that the model size is roughly doubled.)
4. Undo the change in step 2.
5. rasa train -vv
Now NLU would be unnecessarily retrained.

With the suggested change, there would be no duplication of nlu and core in the model.tar.gz and NLU would not be retrained.

joejuzl:
@cmanovit thanks for the detailed explanation of how to reproduce - that's really useful.
You are right it's caused by `model.should_retrain` being run twice as it has the side-affect of moving parts of the old model.

rasa/model_training.py
```diff
@@ -218,11 +218,10 @@ async def _train_async_internal(
     new_fingerprint = await model.model_fingerprint(file_importer)
     old_model = model.get_latest_model(output_path)
 
-    fingerprint_comparison = model.should_retrain(
-        new_fingerprint, old_model, train_path, force_training=force_training
-    )
-
     if dry_run:
+        fingerprint_comparison = model.should_retrain(
```
joejuzl, line 222: Would it be possible to leave this above `if dry_run` but use the result later in the method instead of re-running?

joejuzl (changes requested):

rasa/model_training.py
cmanovit: The two calls to model.should_retrain() are passing different numbers of arguments and I have not investigated further into the subroutine's code to understand how to combine them into a single call.

joejuzl: Ok, this should be fine for now then, as we will be refactoring this area of code soon anyway.

joejuzl:
As this is a bug fix could you please target `2.7.x` instead of `main`?

joejuzl:
Sorry forgot to ask before, but could you also please add a changelog entry? (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)

joejuzl (approved):
Thanks for this!

PR #8718, opened 2021-05-19
remake: Tensorlfow cpp warn log supression for pytest

### LINKED ISSUE #7762: Suppress noisy tensorflow logs on linux and Windows
**Rasa version**:
2.2.4
**Python version**:
3.7.2
Windows, Linux

**Issue**:

It seems that the hack to suppress noisy tensorflow logs is no longer working for Windows or Linux. This confuses users unnecessarily.

example:
```
2021-01-15 12:12:22.059483: W tensorflow/stream_executor/platform/default/dso_loader.cc:59] Could not load dynamic library 'libcudart.so.10.1'; dlerror: libcudart.so.10.1: cannot open shared object file: No such file or directory
```

**Definition of Done**:

- [ ] determine the cause of Tensorflow warnings when using `rasa` in non-debug mode & document
- [ ] fix the current issue (that warnings get shown)

twerkmeister:
remake of #8702 / seems there was some problem with my poetry version that messed up dependencies

**Proposed changes**:
- fixes #7762 
- added pytest-env dependency
- added TF_CPP_MIN_LOG_LEVEL=2 as default to pytest.ini to suppress tensorflow C++ code warnings, info, and debug by default


twerkmeister:
To see functionality compare log of this branchs ubuntu run to main branch ubuntu run. On main branch, you'll see tf logs of the type

```
2021-05-18 13:15:45.501852: W tensorflow/stream_executor/platform/default/dso_loader.cc:59] Could not load dynamic library 'libcudart.so.10.1'; dlerror: libcudart.so.10.1: cannot open shared object file: No such file or directory
2021-05-18 13:15:45.501895: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
```

but not on this one

pyproject.toml
```diff
@@ -152,6 +152,7 @@ bandit = "^1.6.3"
 typing-extensions = "^3.7.4"
 memory-profiler = "^0.58.0"
 psutil = "^5.8.0"
+pytest-env = "^0.6.2"
```
wochinge, line 155: I'm a bit hesitant to add a dependency which wasn't updated since 2017 to our codebase even if it's just a dev dependency. How about using `pytest` [hooks](https://docs.pytest.org/en/latest/how-to/writing_hook_functions.html) for it or editing our `Makefile`?

wochinge (commented):
Just to be clear: This warnings currently don't show in the CLI - only during testing?

twerkmeister:
> Just to be clear: This warnings currently don't show in the CLI - only during testing?

Yes. In the CLI the logs are successfully surpressed as you had shown me. They do show up however when using the `--debug` flag which I think is fine.

pyproject.toml
twerkmeister: Fair point about the updates. At the same time its a 50 code line plugin. And I think it does what it is supposed to do in a straightforward and discoverable manner, setting the env var for all pytest runs.

I think using hooks to set env variables obfuscates things. Make file could be adjusted, there we just need to put it at all invocations, which is 4 so far. There's also another variable - OMP_NUM_THREADS=1 that is set there already. So either we can add the log var there or pull the OMP into the pytest env.

wochinge: Let's do the Makefile then + adding a comment why we need it.

twerkmeister:
@wochinge alrighty, just did the make file change

wochinge (approved):

PR #8724, opened 2021-05-19
Added SASL mechanism as Optional Kafka Producer Parameter

b-quachtran:
**Proposed changes**:
- Added SASL mechanism as optional Kafka Producer parameter that can be specified in `endpoints.yml`


changelog/8724.feature.md
```diff
@@ -0,0 +1 @@
+Added `sasl_mechanism` as an optional configurable parameters for the Kafka Producer.
```
wochinge, line 1: can we please add a links to the Kafka producer docs?

rasa/core/brokers/kafka.py
```diff
@@ -20,6 +20,7 @@ def __init__(
         partition_by_sender: bool = False,
         sasl_username: Optional[Text] = None,
         sasl_password: Optional[Text] = None,
+        sasl_mechanism: Optional[Text] = "PLAIN",
```
wochinge, line 23: do we have a test which asserts this is correctly set on the class?

rasa/core/brokers/kafka.py
```diff
@@ -45,6 +46,10 @@ def __init__(
                 sender_id or not
             sasl_username: Username for plain authentication.
             sasl_password: Password for plain authentication.
+            sasl_mechanism: Authentication mechanism when security_protocol is
+                configured for SASL_PLAINTEXT or SASL_SSL.
+                Valid values are: PLAIN, GSSAPI, OAUTHBEARER, SCRAM-SHA-256,
```
wochinge, line 51: How about validating that the values are in this range? Or is this done by the library? If it's done by the library and it raises an exception this should wrapped in an exception type which inherits from `RasaException` so that we don't see this event in sentry

rasa/core/brokers/kafka.py
b-quachtran: Ah, not yet. Just added a test that checks it's been correctly set.

changelog/8724.feature.md
```diff
@@ -0,0 +1 @@
+Added `sasl_mechanism` as an optional configurable parameters for the [Kafka Producer](https://kafka-python.readthedocs.io/en/master/apidoc/KafkaProducer.html).
```
wochinge, line 1: Shouldn't the link go to the Rasa docs? 🤔

b-quachtran: Yep 🤦

data/test_endpoints/event_brokers/kafka_invalid_sasl_mechanism.yml
```diff
@@ -0,0 +1,12 @@
...
+  security_protocol: SASL_SSL
+  topic: topic
+  url: localhost
+  sasl_username: username
+  sasl_password: password
+  sasl_mechanism: SOMETHING
+  ssl_cafile: CARoot.pem
+  ssl_certfile: certificate.pem
+  ssl_keyfile: key.pem
+  ssl_check_hostname: True
```
wochinge, line 12:
```suggestion
  ssl_check_hostname: True

```

rasa/core/brokers/kafka.py
```diff
@@ -7,6 +7,7 @@
 from rasa.core.brokers.broker import EventBroker
 from rasa.shared.utils.io import DEFAULT_ENCODING
 from rasa.utils.endpoints import EndpointConfig
+from rasa.shared.exceptions import KafkaProducerInitializationError
```
wochinge, line 10: I'd keep this exception in the `Kafka` module. I don't think there is a reason to use this exception outside of the `Kafka` module.

rasa/core/brokers/kafka.py
```diff
@@ -137,15 +144,21 @@ def _create_producer(self) -> None:
...
-            self.producer = kafka.KafkaProducer(
-                client_id=self.client_id,
-                bootstrap_servers=self.url,
-                value_serializer=lambda v: json.dumps(v).encode(DEFAULT_ENCODING),
-                sasl_plain_username=self.sasl_username,
-                sasl_plain_password=self.sasl_password,
-                sasl_mechanism="PLAIN",
-                security_protocol=self.security_protocol,
-            )
+            try:
```
wochinge, line 147: how about moving the producer creation to a method so that we only need one `try ... except` block?

rasa/core/brokers/kafka.py
b-quachtran: Ah yeah it's handled by the library and raises an AssertionError

rasa/core/brokers/kafka.py
b-quachtran: Ah yeah that makes sense, 👍

rasa/core/brokers/kafka.py
b-quachtran: Since the constructor params change depending on the security protocol, I did a slight re-factor so that it's passed in as a dict of arguments into one try-except block where the KafkaProducer is instantiated. Does this approach make sense or is there a better way to handle it?

rasa/core/brokers/kafka.py
```diff
@@ -7,10 +7,15 @@
...
+from rasa.shared.exceptions import RasaException
 
 logger = logging.getLogger(__name__)
 
 
+class KafkaProducerInitializationError(RasaException):
+    """Raised if the Kafka Producer cannot be properly initialized."""
+
+
 class KafkaEventBroker(EventBroker):
```
b-quachtran, line 19: @wochinge Unrelated to this PR, but a CI check is failing because of the docs-string thing. Should all the arg comments after __init__ be moved up to line 20?

alwx: No, we usually just write a short docstring for the class itself, and the args should still be where they are right now.

rasa/core/brokers/kafka.py
```diff
@@ -20,6 +25,7 @@ def __init__(
         partition_by_sender: bool = False,
         sasl_username: Optional[Text] = None,
         sasl_password: Optional[Text] = None,
+        sasl_mechanism: Optional[Text] = "PLAIN",
```
alwx, line 28: Optional: I would actually make it `enum.Enum` instead of `Text`, since here you actually provide the set of allowed SASL mechanisms. 

Something like that:

```
class SASLMechanism(enum.Enum):
    PLAIN = 0
    GSSAPI = 1
    ...
    
    def __str__(self) -> Text:
        """Returns a string representation of the object."""
        return self.name.lower()

PR #8730, opened 2021-05-21
Improve validation of action events

### LINKED ISSUE #8127: Better validation of action events
**Rasa version**: Tested on 2.3.4, bug should still exist on currently latest `main@eb408bb2d216de3a1cd1f72799d7b80eceeb1fd2`

**Rasa SDK version** (if used & relevant): 2.3.1

**Rasa X version** (if used & relevant): 0.36.0 and later (not relevant however)

**Python version**: 3.8.6

**Operating system** (windows, osx, ...): Mac OS 11.2.1

**Issue**:
This came out of [this Rasa X issue](https://github.com/RasaHQ/rasa-x/issues/4521) (see @tmbo's comment [here](https://github.com/RasaHQ/rasa-x/issues/4521#issuecomment-790057056) for a summary).

The [validation of action responses are quite optimistic](https://github.com/RasaHQ/rasa/blob/eb408bb2d216de3a1cd1f72799d7b80eceeb1fd2/rasa/core/actions/action.py#L579), allowing actions to create invalid events (i.e. `"entities": {}` instead of the valid `"entities": []`), resulting on a bunch of different issues downstream in Rasa X, since so far Rasa X seems to trust that events sent over the broker are valid.

Here's a handy example repo made by @rgstephens to illustrate the issue: [RasaHQ/action-execute-convo-issue](https://github.com/RasaHQ/action-execute-convo-issue). Note that this is actually [a bug](https://github.com/RasaHQ/action-execute-convo-issue/blob/8bc08a75d7ad0dfd2917cd3c29fb5f09c85169e0/actions/actions.py#L38) in the action code that should, however, be caught by Rasa OSS and have the error (invalid event) logged.

ancalita:
**Proposed changes**:
- Fix #8127 


ancalita:
Hey @wochinge I wanted to check a few things with you:
- Because I enforced type "number" (float) for `timestamp`, I had to amend a few tests where event timestamp was set to None, such as `tests.core.test_examples.py::test_formbot_example`, `tests.core.test_actions.py::test_remote_action_utterances_with_none_values` and `tests.core.test_actions.py::test_remote_action_with_template_param`. Hope that's ok!
- The value for `event` label gets changed for `LegacyFormValidation` [here](https://github.com/RasaHQ/rasa/blob/41e3b227101e6ace3f85c2d99a7f48f4528a8b93/rasa/shared/core/events.py#L1799) and for `LegacyForm` [here](https://github.com/RasaHQ/rasa/blob/41e3b227101e6ace3f85c2d99a7f48f4528a8b93/rasa/shared/core/events.py#L1699). Is it ok to remove the subschemas for these events since they're covered by the subschemas of `LoopInterrupted` and respectively `ActiveLoop`?
- I noticed that `EntitiesAdded`, `ReminderScheduled` and `ReminderCancelled` all have `entities` key in the payload, but for the latter 2, because `entities` is an optional argument in their constructor, I've decided not to add detailed subschema here.

ancalita:
> @ancalita Just before I review this: Did you do some manual testing with the action server for this? e.g. using the `formbot` example and modifying the custom actions a bit to ensure things still work as expected?

Ah no I didn't have the idea - which custom actions do I need to modify?

wochinge:
Nothing specific, but I'd try out the following scenarios:
* do forms still work as expected
* try dispatching different event types from the SDK with different parameters and see if things still work (SlotSet, ActionExecuted, etc.)

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                                },
+                                            },
+                                        },
+                                    },
+                                }
+                            },
+                            {
+                                "properties": {
+                                    "event": {"const": "action"},
+                                    "policy": {},
```
wochinge, line 636: that's optional text, isn't it"

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                    "confidence": {},
+                                    "name": {"type": "string"},
```
wochinge, line 638: that can also be `None` no?

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                                    },
+                                                },
+                                            },
+                                        },
+                                    },
+                                }
+                            },
+                            {
+                                "properties": {
+                                    "event": {"const": "action"},
```
wochinge, line 635: `metadata`, `timestamp`, `hide_rule_turn` and `action_text` are missing

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                    "policy": {},
+                                    "confidence": {},
```
wochinge, line 637: optional number?

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                                "end": {"type": "integer"},
+                                                "name": {"type": "string"},
+                                                "confidence": {"type": "number"},
+                                                "extractor": {"type": "string"},
+                                                "value": {"type": "object"},
+                                            },
+                                        },
+                                    },
+                                }
+                            },
```
wochinge, line 666: what about deduplicating this as you use the same for `UserUttered`?

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
                     "type": "array",
                     "items": {
                         "type": "object",
-                        "properties": {"event": {"type": "string"}},
+                        "properties": {"event": {"type": "string"}, "timestamp": {}},
```
wochinge, line 594: timestamp is optional number, isn't it? What about also adding metadata here?

rasa/server.py
```diff
@@ -377,6 +377,128 @@ def validate_request_body(request: Request, error_message: Text) -> None:
         raise ErrorResponse(HTTPStatus.BAD_REQUEST, "BadRequest", error_message)
 
 
+def events_request_format_spec() -> Dict[Text, Any]:
+    """Expected request body schema used for validation."""
+    return {
```
wochinge, line 382: that's the same schema as for the actions module, isn't it? How about deduplicating that?

rasa/server.py
```diff
@@ -744,6 +866,8 @@ async def append_events(request: Request, conversation_id: Text) -> HTTPResponse
             "to the state of a conversation.",
         )
 
+        validate_events_in_request_body(request)
```
wochinge, line 869: can't this replace `validate_request_body`?

rasa/server.py
```diff
@@ -377,6 +377,128 @@ def validate_request_body(request: Request, error_message: Text) -> None:
...
+
+    try:
+        validate(events, events_request_format_spec())
+    except ValidationError as e:
+        e.message += (
+            " Failed to validate the events format in the action server request body."
+            "For more information about the format visit: "
+            "https://rasa.com/docs/action-server/pages/action-server-api"
+        )
+        raise e
```
wochinge, line 499: we should raise an `ErrorResponse` with status code 400. This will currently result in a 500 error code

tests/core/test_actions.py
```diff
@@ -255,10 +270,14 @@ async def test_remote_action_utterances_with_none_values(
 
     response = {
         "events": [
-            {"event": "form", "name": "restaurant_form", "timestamp": None},
+            {
```
wochinge, line 273: why did we need to adapt these?

tests/core/test_actions.py
```diff
@@ -344,6 +367,164 @@ async def test_remote_action_with_template_param(
     ]
```
wochinge, line 369: can we please add a test that makes sure that every event type is covered by the jsonschema? [rasa.shared.utils.common.all_subclasses](https://github.com/RasaHQ/rasa/blob/74ce55924e820be3cd81fcc105ed4ba6b7225df4/tests/shared/core/test_events.py#L292) should helpful here

tests/core/test_actions.py
```diff
@@ -344,6 +367,164 @@ async def test_remote_action_with_template_param(
...
 
 
+@pytest.mark.parametrize(
+    ("event", "expected_event"),
+    (
+        ({"event": "entities", "timestamp": 1621590172.3872123}, EntitiesAdded),
```
wochinge, line 373: How about changing this test to use event objects and then call `as_dict()` on them instead of putting the output payload directly? This would make the test more robust in case somebody decides to change how an event is serialized.

e.g. something along these lines:

```python
events = [ActionExecuted(action_name="Hi Anca", timestamp=1234)]
events = [e.as_dict() for e in events]
validate(events)
```

tests/test_server.py
```diff
@@ -1232,6 +1229,28 @@ async def test_predict(rasa_app: SanicASGITestClient):
...
+            },
+        },
+    ]
+
+    _, response = await rasa_app.post(
+        "/model/predict",
+        json=data,
+        headers={"Content-Type": rasa.server.JSON_CONTENT_TYPE},
+    )
+    assert response.status == HTTPStatus.INTERNAL_SERVER_ERROR
```
wochinge, line 1251: it's not an internal server error but a bad request which leads to that. Hence, I'd use 400

wochinge (changes requested):
Thanks for jumping on this 💯 And sorry for the late review! Great detective work 🕵🏻‍♀️

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                        "oneOf": [
+                            {
+                                "properties": {
+                                    "event": {"const": "user"},
+                                    "text": {"type": "string"},
+                                    "input_channel": {},
```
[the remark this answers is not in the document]
ancalita: For `input_channel` key? That is actually equivalent to type `Any`.

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                    "message_id": {"type": "string"},
+                                    "parse_data": {
+                                        "type": "object",
+                                        "properties": {
+                                            "text": {"type": "string"},
+                                            "intent": {
```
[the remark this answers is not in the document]
ancalita: Are you referring to event `metadata`? Because `metadata` is Optional[Dict], I will add it as `"metadata": { }`.
Will also add `intent_ranking`.

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
                         "type": "object",
-                        "properties": {"event": {"type": "string"}},
+                        "properties": {"event": {"type": "string"}, "timestamp": {}},
+                        "oneOf": [
+                            {
+                                "properties": {
+                                    "event": {"const": "user"},
+                                    "text": {"type": "string"},
+                                    "input_channel": {},
+                                    "message_id": {"type": "string"},
```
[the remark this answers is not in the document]
ancalita: Yes, good spot. I'll make `message_id` optional.

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                                "type": "object",
+                                                "properties": {
+                                                    "name": {"type": "string"},
+                                                    "confidence": {"type": "number"},
+                                                },
+                                            },
+                                            "entities": {
+                                                "type": "array",
+                                                "items": {
+                                                    "type": "object",
```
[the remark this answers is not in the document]
ancalita: would these be additional properties to the dictionary item in the `entities` array? 🤔

rasa/core/actions/action.py
```diff
@@ -591,7 +591,103 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                                    "properties": {
+                                                        "start": {"type": "integer"},
+                                                        "end": {"type": "integer"},
+                                                        "name": {"type": "string"},
+                                                        "confidence": {
+                                                            "type": "number"
+                                                        },
+                                                        "extractor": {"type": "string"},
```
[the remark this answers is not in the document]
ancalita: I wasn't sure since I've never seen them in the 'wild'. I'll make `extractor` optional to be on the safe side.

rasa/core/actions/action.py
ancalita: For `policy`? Yes, it's actually type `Any`, I couldn't find the equivalent of Optional in json schema, not sure if it's supported.

rasa/core/actions/action.py
ancalita: I'll add `action_text` and `hide_rule_turn`, however I thought that if `metadata` and `timestamp` are mentioned in L594 `properties`, I don't need to add them in 'oneOf', was that incorrect?

rasa/core/actions/action.py
ancalita: Actually type `Any`, if there's an equivalent for Optional type in JSON schema, I wasn't able to find it.

rasa/core/actions/action.py
ancalita: By deduplicating you mean extracting this dictionary in a variable and assigning the variable instead?

rasa/server.py
ancalita: Yes, good idea! I can extract the events subschema - however where would I place it, as a global variable in `action.py`?

rasa/server.py
ancalita: I've tried and most of the tests in `server.py` fail because the `request.body` didn't contain events, unless I do a check ```
if "events" in request.body:
    validate_events_in_request_body(request)
```
Would that be a good idea? 🤔

File: tests/core/test_actions.py (hunk 10 above)
Comment (ancalita) [reply tests/core/test_actions.py:273]: Because I had enforced type `number` for `timestamp` - but now that it's type Any, I can revert this change.

File: tests/core/test_actions.py (hunk 11 above)
Comment (ancalita) [reply tests/core/test_actions.py:369]: This was super helpful - I found two subclasses that are not covered: `rasa.core.test.WronglyPredictedAction` and `rasa.core.test.WronglyClassifiedUserUtterance` - I can add these, although they reside in a different module. Please let me know if I should ignore them?

File: rasa/core/actions/action.py (hunk 14 above)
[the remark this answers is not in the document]
Comment (wochinge) [reply rasa/core/actions/action.py:618]: Sorry, what I mean is: Shouldn't `input_channel` be `Optional[Text]`?

File: rasa/core/actions/action.py [hunk 19]
```diff
@@ -584,19 +584,122 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                            "timestamp": {},
+                            "metadata": {},
+                        },
+                        "oneOf": [
+                            {
+                                "properties": {
+                                    "event": {"const": "user"},
+                                    "text": {"type": "string"},
+                                    "input_channel": {},
+                                    "message_id": {},
```
wochinge, line 619: don't we know the type for `message_id`?

rasa/core/actions/action.py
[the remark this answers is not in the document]
wochinge: yes 👍🏻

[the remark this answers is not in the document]
wochinge: Just found this by coincidence when doing a global search for "group": https://github.com/RasaHQ/rasa/blob/94eb8b78bffa817b8917c48b348cfdf66f96a5f1/rasa/shared/nlu/training_data/schemas/data_schema.py#L13

rasa/core/actions/action.py
```diff
@@ -584,19 +584,122 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
             "type": "object",
             "properties": {
                 "events": {
                     "type": "array",
                     "items": {
                         "type": "object",
-                        "properties": {"event": {"type": "string"}},
+                        "properties": {
+                            "event": {"type": "string"},
+                            "timestamp": {},
```
wochinge, line 610: don't we know the timestamp is `Optional[float]`?

rasa/core/actions/action.py
```diff
@@ -584,19 +584,122 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                            "metadata": {},
```
wochinge, line 611: does this exclude `list`?

rasa/core/actions/action.py
```diff
@@ -584,19 +584,122 @@ def action_response_format_spec() -> Dict[Text, Any]:
...
+                                }
+                            },
+                            {
+                                "properties": {
+                                    "event": {"const": "action"},
+                                    "policy": {},
+                                    "confidence": {},
+                                    "name": {},
+                                    "hide_rule_turn": {"type": "boolean"},
+                                    "action_text": {},
```
wochinge, line 655: that's `Optional[Text]`, isn't it?

rasa/core/actions/action.py
wochinge: Pretty sure it can only be a string or null. Where did you find `Any`? 

Apparently you can do `Optional` as described [here](https://stackoverflow.com/a/47592496/3429596)

rasa/server.py
wochinge: How about creating a new module in `rasa.shared.utils.schemas`?

rasa/server.py
wochinge: mhm, can you point me to an example of a failing check?

tests/core/test_actions.py
wochinge: Should be `Optional[float]` - let's revert though 👍🏻

tests/core/test_actions.py
wochinge: mhm, these can actually be ignored 👍🏻 They are only used for testing and will never enter from the outside / leave Rasa Open Source

rasa/core/actions/action.py
ancalita: Ah nice find! The `Any` type solution I sourced from first answer [here.](https://stackoverflow.com/questions/16826128/why-is-this-json-schema-invalid-using-any-type/16826176) ... I should have scrolled below and I could have found the multiple types solution you shared 🤦‍♀️

rasa/server.py
ancalita: Ah I tried locally and I didn't make a note, but I vaguely remember these 🤔 :

- https://github.com/RasaHQ/rasa/blob/fa0fc6119b214c2f9be39486d53bcacf487892f0/tests/test_server.py#L445
- https://github.com/RasaHQ/rasa/blob/fa0fc6119b214c2f9be39486d53bcacf487892f0/tests/test_server.py#L474
- https://github.com/RasaHQ/rasa/blob/fa0fc6119b214c2f9be39486d53bcacf487892f0/tests/test_server.py#L510
- https://github.com/RasaHQ/rasa/blob/fa0fc6119b214c2f9be39486d53bcacf487892f0/tests/test_server.py#L606
- https://github.com/RasaHQ/rasa/blob/fa0fc6119b214c2f9be39486d53bcacf487892f0/tests/test_server.py#L772

rasa/core/actions/action.py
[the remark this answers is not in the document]
ancalita: Got it, will change.

rasa/core/actions/action.py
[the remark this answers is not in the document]
ancalita: Thanks 👍 I'm conflicted about `value` now - should it be `type: object` as [here](https://github.com/wochinge/go-rasa-sdk/blob/f3adc0dcaf02dd2aa15df80ffa63e1f39933ffb3/v2/rasa/events/events.go#L237) or `type: [string, number]`?

rasa/core/actions/action.py
ancalita: Yup, amended to Optional string.

rasa/core/actions/action.py
ancalita: yup, amended now.

rasa/core/actions/action.py
ancalita: I've made `metadata` as type `Optional[Dict]`.

rasa/server.py
ancalita: Done 😃

changelog/8127.improvement.md
```diff
@@ -0,0 +1,2 @@
+Added detailed json schema validation for `UserUttered`, `SlotSet`, `ActionExecuted` and `EntitiesAdded` events both sent and received from the action server, as well as covered at high-level the validation of the rest of the 20 events.
+In case the events are invalid, a json schema `ValidationError` will be raised.
```
wochinge, line 2:
```suggestion
In case the events are invalid, a `ValidationError` will be raised.
```

rasa/server.py
wochinge: Yeah, that makes sense. They receive completely different payloads. So they should continue to use the old validation function `validate_request_body`. Only endpoints which receive events should use `validate_events_in_request_body`

tests/core/test_actions.py
```diff
@@ -344,6 +362,156 @@ async def test_remote_action_with_template_param(
     ]
 
 
+@pytest.mark.parametrize("event_class", rasa.shared.utils.common.all_subclasses(Event))
+async def test_remote_action_validate_all_event_subclasses(event_class: Type[Event]):
+    response = {"events": [{"event": event_class.type_name}], "responses": []}
+
+    from jsonschema import validate
```
wochinge, line 369: why are we using a local import here?

tests/core/test_actions.py
```diff
@@ -344,6 +362,156 @@ async def test_remote_action_with_template_param(
     ]
 
 
+@pytest.mark.parametrize("event_class", rasa.shared.utils.common.all_subclasses(Event))
+async def test_remote_action_validate_all_event_subclasses(event_class: Type[Event]):
```
wochinge, line 366: should we move this test to a module `tests.shared.utils.schemas.events`? The test doesn't seem particularly related to the action server response

tests/core/test_actions.py
```diff
@@ -344,6 +362,156 @@ async def test_remote_action_with_template_param(
...
+    response = {"events": [{"event": event_class.type_name}], "responses": []}
```
wochinge, line 367: would it be possible to create objects out of this classes and then call `as_dict()`? This test doesn't seem particularly realistic (nobody would send an event which only has the `event` key)

tests/core/test_actions.py
```diff
@@ -344,6 +362,156 @@ async def test_remote_action_with_template_param(
...
+        SlotSet(key="my_slot", value={}),
+        ConversationResumed(),
+        ConversationPaused(),
+        FollowupAction(name="test"),
+        StoryExported(),
+        Restarted(),
+        ActionReverted(),
+        UserUtteranceReverted(),
+        BotUttered(text="Test bot utterance"),
+        UserUttered(text="hello", parse_data={}),
```
wochinge, line 404: can we please test a full fledged user uttered event which includes multiple entities and ranking etc.?

tests/core/test_actions.py
```diff
@@ -344,6 +362,156 @@ async def test_remote_action_with_template_param(
...
+        ReminderScheduled(
+            timestamp=None, trigger_date_time=datetime.now(), intent="greet"
+        ),
+        ActionExecutionRejected(action_name="my_action"),
+        LegacyFormValidation(validate=True, timestamp=None),
+        LoopInterrupted(timestamp=None, is_interrupted=False),
+        ActiveLoop(name="loop"),
+        LegacyForm(name="my_form"),
+        AllSlotsReset(),
+        SlotSet(key="my_slot", value={}),
```
wochinge, line 395: Can we test that with a couple of different values? E.g. lists, dicts, simple text?

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,105 @@
...
+                "properties": {
+                    "event": {"const": "user"},
+                    "text": {"type": ["string", "null"]},
+                    "input_channel": {"type": ["string", "null"]},
+                    "message_id": {"type": ["string", "null"]},
+                    "parse_data": {
+                        "type": "object",
+                        "properties": {
+                            "text": {"type": "string"},
+                            "intent_ranking": {
```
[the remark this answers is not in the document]
wochinge:
```
{
      "event": "user",
      "timestamp": 1619445071.2267451,
      "text": "hi",
      "parse_data": {
        "intent": {
          "id": -4389344335148575888,
          "name": "greet",
          "confidence": 0.9604260921478271
        },
        "entities": [],
        "text": "hi",
        "message_id": "3f4c04602a4947098c574b107d3ccc50",
        "metadata": {},
        "intent_ranking": [
          {
            "id": -4389344335148575888,
            "name": "greet",
            "confidence": 0.9604260921478271
          },
          {
            "id": 7180145986630405383,
            "name": "goodbye",
            "confidence": 0.01835782080888748
          },
          {
            "id": 4246019067232216572,
            "name": "deny",
            "confidence": 0.011255578137934208
          },
          {
            "id": -4048707801696782560,
            "name": "bot_challenge",
            "confidence": 0.004019865766167641
          },
          {
            "id": -5942619264156239037,
            "name": "affirm",
            "confidence": 0.002524246694520116
          },
          {
            "id": 677880322645240870,
            "name": "mood_great",
            "confidence": 0.002214624546468258
          },
          {
            "id": -5973454296286367554,
            "name": "chitchat",
            "confidence": 0.0009614597074687481
          },
          {
            "id": -4598562678335233249,
            "name": "mood_unhappy",
            "confidence": 0.00024030178610701114
          }
        ],
        "response_selector": {
          "all_retrieval_intents": [],
          "default": {
            "response": {
              "id": -226546773594344189,
              "responses": [
                {
                  "text": "chitchat\/ask_name"
                }
              ],
              "response_templates": [
                {
                  "text": "chitchat\/ask_name"
                }
              ],
              "confidence": 0.9618658423423767,
              "intent_response_key": "chitchat\/ask_name",
              "utter_action": "utter_chitchat\/ask_name",
              "template_name": "utter_chitchat\/ask_name"
            },
            "ranking": [
              {
                "id": -226546773594344189,
                "confidence": 0.9618658423423767,
                "intent_response_key": "chitchat\/ask_name"
              },
              {
                "id": 8392727822750416828,
                "confidence": 0.03813415765762329,
                "intent_response_key": "chitchat\/ask_weather"
              }
            ]
          }
        }
      }
```

wochinge (commented):
Looks pretty good 💯 A few more things.

rasa/server.py
```diff
@@ -377,6 +378,34 @@ def validate_request_body(request: Request, error_message: Text) -> None:
         raise ErrorResponse(HTTPStatus.BAD_REQUEST, "BadRequest", error_message)
 
 
+def events_request_format_spec() -> Dict[Text, Any]:
+    """Expected request body schema used for validation."""
+    return EVENTS_SCHEMA
```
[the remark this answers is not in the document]
ancalita: I followed the example from [action server response validation ](https://github.com/RasaHQ/rasa/blob/7d4164f67f594a441dc951092ad15b8a096d3bca/rasa/core/actions/action.py#L606)and that was using a callable as schema argument, but I've just tried this `validate(events, EVENTS_SCHEMA)` and the tests pass. I can remove this function since `jsonschema` doesn't enforce a type for schema argument.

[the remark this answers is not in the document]
wochinge: in the action server response validation it also passes **the result** from calling the callable. 
Let's omit this 👍🏻

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,105 @@
...
+                "properties": {
+                    "event": {"const": "slot"},
+                    "name": {"type": "string"},
+                    "value": {},
+                }
+            },
+            {
+                "properties": {
+                    "event": {"const": "entities"},
+                    "entities": ENTITIES_SCHEMA,
```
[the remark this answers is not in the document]
ancalita: True, I've added it.

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,105 @@
...
+                    "confidence": {"type": ["number", "null"]},
+                    "name": {"type": ["string", "null"]},
+                    "hide_rule_turn": {"type": "boolean"},
+                    "action_text": {"type": ["string", "null"]},
+                }
+            },
+            {
+                "properties": {
+                    "event": {"const": "slot"},
+                    "name": {"type": "string"},
```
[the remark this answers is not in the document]
ancalita: True, added.

tests/core/test_actions.py
ancalita: It depends what the purpose is: I thought this test only checks that any new potential event is added to the schema as well, otherwise it fails if forgotten. 
The test below this one `test_remote_action_valid_payload_all_events` verifies that the event objects are valid - I can tweak those to add more arguments for a more realistic scenario?

tests/core/test_actions.py
ancalita: Removed.

tests/core/test_actions.py
ancalita: Ok, done 👍

tests/core/test_actions.py
ancalita: I recycled the example you provided 😁

tests/core/test_actions.py
wochinge: The advantage of `as_dict` would be that we'd have relatively realistic payloads which automatically change when somebody does changes to the `as_dict` implementation. In my opinion some of these events shouldn't be valid if just the `event` key is provided.

ancalita: Yes, some of them failed after enforcing required properties - I've hacked the test for now 😅  but my problem was how to instantiate an event from class type? Do I use static method `from_parameters`? Although the if/else will get very long since parameters will differ event by event. If that works, I can use `as_dict` then in the response.

wochinge: mhm 😬 Note sure, but do all Event classes have defaults for all constructor params? If not I'd simply add a test which makes sure that each `Event.type_name` is present in the schema and then hope that you're test below this captures the rest 👍🏻

rasa/server.py
```diff
@@ -377,6 +378,28 @@ def validate_request_body(request: Request, error_message: Text) -> None:
...
+
+    try:
+        validate(events, EVENTS_SCHEMA)
+    except ValidationError as error:
+        raise ErrorResponse(
+            HTTPStatus.BAD_REQUEST,
+            "BadRequest",
+            f"Failed to validate the events format. "
+            f"For more information about the format visit: "
+            f"https://rasa.com/docs/action-server/pages/action-server-api. Error: {error}",
```
wochinge, line 399: can we use our docs constants for this?

ancalita: I just found the base URL constant applicable and used this - let me know if that doesn't look ok!

rasa/server.py
```diff
@@ -377,6 +379,28 @@ def validate_request_body(request: Request, error_message: Text) -> None:
...
+
+    try:
+        validate(events, EVENTS_SCHEMA)
+    except ValidationError as error:
+        raise ErrorResponse(
+            HTTPStatus.BAD_REQUEST,
+            "BadRequest",
+            f"Failed to validate the events format. "
+            f"For more information about the format visit the docs. Error: {error}",
+            help_url=DOCS_BASE_URL_ACTION_SERVER + "/pages/action-server-api",
```
wochinge, line 400: * let's use a f-string here please
* shouldn't we rather redirect them to the API spec for the Rasa Open Source server endpoints? That's [here](https://rasa.com/docs/rasa/pages/http-api)

rasa/server.py
wochinge: Can we then drop `validate_request_body` for .e.g. `append_events`, `replace_events` and `validate_events_in_request_body`?

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,155 @@
...
+                        "type": "object",
+                        "properties": {
+                            "text": {"type": "string"},
+                            "intent_ranking": {"type": "array", "items": INTENT},
+                            "intent": INTENT,
+                            "entities": ENTITIES_SCHEMA,
+                            "response_selector": {
+                                "type": "object",
+                                "properties": {
+                                    "all_retrieval_intents": {"type": "array"},
```
wochinge, line 50: array of what?

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,155 @@
...
+                                    "default": {
```
wochinge, line 51: mhm, it might not always be `default` according to [this part](https://github.com/RasaHQ/rasa/blob/fe9f14919df4dcb454ba1052de344fae1466a6d1/rasa/nlu/selectors/response_selector.py#L451). Is there a simple solution in jsonschema for dynamic keys? Otherwise I'd skip that for now

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,155 @@
...
+                                                    "utter_action": {"type": "string"},
+                                                    "template_name": {"type": "string"},
+                                                },
+                                            },
+                                            "ranking": {
+                                                "type": "array",
+                                                "items": {
+                                                    "type": "object",
+                                                    "properties": {
+                                                        "confidence": {
```
wochinge, line 92: there is also `id` field which seems to be `number`

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,155 @@
+ENTITIES_SCHEMA = {
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
```
wochinge, line 5: how making `entity` and `value` required properties?

tests/core/test_actions.py
```diff
@@ -344,6 +362,227 @@ async def test_remote_action_with_template_param(
...
+        BotUttered(text="Test bot utterance"),
+        UserUttered(
+            text="hello",
+            parse_data={
+                "intent": {
+                    "id": -4389344335148575888,
+                    "name": "greet",
+                    "confidence": 0.9604260921478271,
+                },
+                "entities": [],
```
wochinge, line 402: can we please also have something with entities? They are super important in Rasa Open Source and should in my opinion be covered

tests/shared/utils/schemas/test_events.py
```diff
@@ -0,0 +1,28 @@
...
+            "responses": [],
+        }
+    elif event_class.type_name == "entities":
+        response = {"events": [{"event": "entities", "entities": []}], "responses": []}
+    else:
+        response = {"events": [{"event": event_class.type_name}], "responses": []}
+
+    if (
+        event_class.type_name != "wrong_utterance"
+        and event_class.type_name != "wrong_action"
```
wochinge, line 26: can we please add a comment to the test why we are excluding these?

tests/shared/utils/schemas/test_events.py
```diff
@@ -0,0 +1,28 @@
...
+import pytest
+import jsonschema
+
+from rasa.core.actions.action import RemoteAction
+from rasa.shared.core.events import Event
+import rasa.shared.utils.common
+
+
+@pytest.mark.parametrize("event_class", rasa.shared.utils.common.all_subclasses(Event))
+async def test_remote_action_validate_all_event_subclasses(event_class: Type[Event]):
```
wochinge, line 12: not needed, right?
```suggestion
def test_remote_action_validate_all_event_subclasses(event_class: Type[Event]):
```

tests/test_server.py
```diff
@@ -1204,22 +1204,19 @@ async def raiseUnexpectedError() -> NoReturn:
...
-                        "entities": [],
-                        "intent": {"confidence": 0.57, INTENT_NAME_KEY: "greet"},
-                        "text": "hello",
-                    },
-                },
-            ]
-        }
-    }
+    data = [
+        {"event": "action", "name": "action_listen"},
```
wochinge, line 1208: omg - how did this even work before? 🤯

rasa/server.py
ancalita: Yup, will change to f-string. I didn't know of this page, thanks for pointing this out - would the specific section be [this](https://rasa.com/docs/rasa/pages/http-api#tag/Tracker)? particularly the POST `append events to a tracker` and PUT `replace a tracker's events`?

rasa/server.py
ancalita: By that you mean adding the check if `request.body` is not empty in `validate_events_in_request_body`? Then removing this from the 3 routes (`append_events`, `replace_events`, `tracker_predict`)?

rasa/shared/utils/schemas/events.py
ancalita: I wasn't sure myself where to look that info up since the example you provided above was an empty array:
```
"response_selector": {
          "all_retrieval_intents": [],
```

rasa/shared/utils/schemas/events.py
ancalita: I was only able to find resource on defining [regular expressions](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties) for keys - would this be suitable?

rasa/shared/utils/schemas/events.py
ancalita: I wouldn't make them required at this schema level because it wouldn't be applicable for the `UserUttered` event because of [this.](https://github.com/RasaHQ/rasa/blob/d4d606b12370d440d9f9ed3f9e4d8e2a23b959b8/rasa/shared/core/events.py#L396). However I can make them required in the `EntitiesAdded` subschema, as that shouldn't be an empty list.

rasa/server.py
wochinge: Exactly 👍🏻 Probably good enough to just link to the overall page instead of the individual sections

rasa/server.py
wochinge: > adding the check if request.body is not empty in validate_events_in_request_body?

Isn't that covered by jsonschema validation on `request.json`? So you might not even need to validate the presence of a body separately.

rasa/shared/utils/schemas/events.py
wochinge: Sorry - how is the linked code blocking making `entity` and `value` required properties? Any empty list would still satisfy the schema, no?

rasa/shared/utils/schemas/events.py
wochinge: You can train a model (e.g. one from the example dir) and then run with `rasa run --enable-api`. You can then
* speak to the model via the REST webhook
* retrieve the tracker for the conversation so that you can see all events

I'm using [Insomnia](https://insomnia.rest) for my REST requests. Happy to give you an introduction and / or share my request collection with you

rasa/shared/utils/schemas/events.py
wochinge: yeah, that would certainly work 👍🏻

rasa/shared/utils/schemas/events.py
ancalita: You're absolutely right - I was under the false impression that required properties will make an empty list fail but just added a few test cases for empty list and they passed 👍

rasa/server.py
ancalita: Hmm, according to this [post](https://stackoverflow.com/questions/65999411/json-schema-validate-non-empty-body), it seems json-schema doesn't cover validation over an empty instance...unless i'm reading this completely wrong 🙃 Would an empty `request.json` be an empty `{ }` in which case according to this post, validation will fail and I can remove the `validate_request_body` as requested?

wochinge: Either that or just `None`. I don't know - you'd probably need to try out (or write a test) for it.

rasa/shared/utils/schemas/events.py
wochinge: > So it seems it's a dead end and I should not implement patternProperties in the schema? WDYT?

Sounds good for me for now. We can still revisit that in the future 👍🏻

ancalita: I've found a way to make this work actually 💡 I used `oneOf` and the tests pass - the caveat is that if `all_retrieval_intents` is a dict instead of list, then the schema won't fail because the `patternProperties` subschema will be valid for this key name 😞

rasa/server.py
ancalita: Added some tests => Empty `request.json` evaluates to `None` and the schema validation fails, so all good to remove `validate_request_body`; the only drawback is that the error message won't be as informative for an empty request body as it was with `validate_request_body`.

tests/test_server.py
```diff
@@ -1232,6 +1229,51 @@ async def test_predict(rasa_app: SanicASGITestClient):
...
+async def test_append_and_replace_events_empty_request_body(
+    rasa_app: SanicASGITestClient,
+):
+    _, response = await rasa_app.post(
+        "/conversations/testid/tracker/events",
+        headers={"Content-Type": rasa.server.JSON_CONTENT_TYPE},
+    )
+    assert response.status == HTTPStatus.BAD_REQUEST
+
+    _, response_2 = await rasa_app.put(
```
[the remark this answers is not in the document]
ancalita: Sure thing! I just thought I'd need to POST before using PUT? But since request body is empty anyway, it probably doesn't matter much?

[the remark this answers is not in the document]
wochinge: yeah shouldn't matter in this case 👍🏻 Just had a quick look on the code and seems that `append_events` doesn't validate the conversation ID (but maybe double check that as I just skimmed that)

rasa/shared/utils/schemas/events.py
```diff
@@ -0,0 +1,173 @@
...
+                            },
+                        },
+                    },
+                },
+            },
+            {
+                "properties": {
+                    "event": {"const": "action"},
+                    "policy": {"type": ["string", "null"]},
+                    "confidence": {"type": ["number", "null"]},
```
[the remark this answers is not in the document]
ancalita: according to [this](https://github.com/RasaHQ/rasa/blob/690f8b891305e5bc36c22febd6e56da26a66be6e/rasa/shared/core/events.py#L1448), seems so 🤔

rasa/server.py
```diff
@@ -377,6 +378,28 @@ def validate_request_body(request: Request, error_message: Text) -> None:
         raise ErrorResponse(HTTPStatus.BAD_REQUEST, "BadRequest", error_message)
 
 
+def validate_events_in_request_body(request: Request) -> None:
+    """Validates events format in request body."""
+    from jsonschema import validate
+    from jsonschema import ValidationError
```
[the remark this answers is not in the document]
ancalita: Got it, will fix in the next commit 👍

rasa/shared/utils/schemas/events.py
ancalita: I've just had an attempt at this using Postman, managed to do step 2 (get the tracker with initial events which are`ActionExecuted` (action_session_start) and  `SessionStarted`), but i don't know how to do step 1 - speak to the bot so I can send an UserUttered event through the `responseselectorbot` 😬

rasa/shared/utils/schemas/events.py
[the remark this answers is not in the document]
ancalita: Having checked `rasa run --enable-api` just now, i can see an event with confidence `null`:
```
{
            "event": "action",
            "timestamp": 1623251973.613313,
            "name": "action_listen",
            "policy": null,
            "confidence": null,
            "action_text": null,
            "hide_rule_turn": false
        }
```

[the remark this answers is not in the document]
wochinge: Mhm, that makes sense 👍🏻 Thanks for checking!

PR #8471, opened 2021-04-14
Improve plot_paired_histogram

JEM-Mosig:
**Proposed changes**:
- Rename `plot_histogram` to `plot_paired_histogram`, add tests, and fix an issue with a missing bar in the chart
- Closes https://github.com/RasaHQ/rasa/issues/7912

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- ~~updated the documentation~~
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #8747, opened 2021-05-26
Adding a rasa exception warning for sentry plot_histogram error

Imod7:
**Proposed changes**:
Adding a rasa exception for plot_histogram error message.

**Steps to replicate the error**:
I tried to replicate the error message shown in this [issue](https://github.com/RasaHQ/rasa/issues/8582) by 
- force setting `thresh = None` in line 127 of this [file](https://github.com/RasaHQ/rasa/blob/main/rasa/utils/plotting.py) 
- and then running the command 
`rasa test nlu --cross-validation`
from an example bot (like formbot)

These settings/run, after quite a while, will produce the following error message :
`TypeError: '>' not supported between instances of 'int' and 'NoneType'`

This error message looks similar to the original one in the corresponding ticket but it makes me doubt because : 
- it is actually half of the original sentry error message 
- it is not in the line mentioned in the sentry 
`[max(hist_data[0], default=0), max(hist_data[1], default=0)], default=0`

### DIFF
--- a/rasa/utils/plotting.py
+++ b/rasa/utils/plotting.py
@@ -10,7 +10,6 @@ from matplotlib.ticker import FormatStrFormatter
 
 import rasa.shared.utils.io
 from rasa.constants import RESULTS_FILE
-from rasa.shared.exceptions import RasaException
 
 logger = logging.getLogger(__name__)
 
--- a/rasa/utils/plotting.py
+++ b/rasa/utils/plotting.py
@@ -125,8 +125,6 @@ def plot_confusion_matrix(
         logger.info(f"Confusion matrix, without normalization: \n{confusion_matrix}")
 
     thresh = zmax / 2.0
-    if not thresh:
-        raise RasaException(f"The variable `thresh` cannot be `{type(thresh)}`.")
     for i, j in itertools.product(
         range(confusion_matrix.shape[0]), range(confusion_matrix.shape[1])
     ):
--- a/rasa/utils/plotting.py
+++ b/rasa/utils/plotting.py
@@ -10,6 +10,7 @@ from matplotlib.ticker import FormatStrFormatter
 
 import rasa.shared.utils.io
 from rasa.constants import RESULTS_FILE
+from rasa.shared.exceptions import RasaException
 
 logger = logging.getLogger(__name__)
 
@@ -124,6 +125,8 @@ def plot_confusion_matrix(
         logger.info(f"Confusion matrix, without normalization: \n{confusion_matrix}")
 
     thresh = zmax / 2.0
+    if not thresh:
+        raise RasaException(f"The variable `thresh` cannot be `{type(thresh)}`.")
     for i, j in itertools.product(
         range(confusion_matrix.shape[0]), range(confusion_matrix.shape[1])
     ):
--- a/tests/nlu/test_evaluation.py
+++ b/tests/nlu/test_evaluation.py
@@ -1228,14 +1228,18 @@ def test_plot_attribute_confidences_with_invalid_values(tmp_path: Path):
 
     rasa.shared.utils.io.create_directory(report_folder)
 
-    intent_results = [
-        IntentEvaluationResult("", "restaurant_search", 76, -21),
-        IntentEvaluationResult("greet", "greet", -32, 99),
+    response_results = [
+        ResponseSelectionEvaluationResult(
+            "chitchat/ask_weather", "chitchat/ask_weather", "What's the weather", 98,
+        ),
+        ResponseSelectionEvaluationResult(
+            "chitchat/ask_name", "chitchat/ask_name", "What's your name?", 12
+        ),
     ]
 
     with pytest.raises(InvalidTypeException):
-        evaluate_intents(
-            intent_results,
+        evaluate_response_selections(
+            response_results,
             report_folder,
             successes=True,
             errors=True,
--- a/tests/nlu/test_evaluation.py
+++ b/tests/nlu/test_evaluation.py
@@ -78,7 +78,7 @@ from rasa.utils.tensorflow.constants import EPOCHS, ENTITY_RECOGNITION
 # this event_loop is used by pytest-asyncio, and redefining it
 # is currently the only way of changing the scope of this fixture
 from tests.nlu.utilities import write_file_config
-
+from rasa.exceptions import InvalidTypeException
 
 # Chinese Example
 # "对面食过敏" -> To be allergic to wheat-based food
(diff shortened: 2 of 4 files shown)


ancalita (commented):
I actually think this has to do with the fact that the list elements in the `hist_data` list( which should be of type `List[List[float]]` are actually of type None so the `max` comparison cannot be done between NoneType. For example, if you try this in python REPL: `print(max([None, None], default=0))`, you'll get the error message `TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'`.

I would actually place a check that the types of elements in `pos_hist` and `neg_hist` are all `float` in [here](https://github.com/RasaHQ/rasa/blob/1aa1b6c93ba2603064737814f9fcdce764af40d9/rasa/nlu/test.py#L311), otherwise raise `RasaException` 🤔

Imod7:
Awesome!!! 💯🥇 Yes, I could not figure how to reproduce it! I will try that then!!! Thanks a bunch !!! 🙏🏽

rasa/utils/plotting.py
```diff
@@ -124,6 +125,8 @@ def plot_confusion_matrix(
         logger.info(f"Confusion matrix, without normalization: \n{confusion_matrix}")
 
     thresh = zmax / 2.0
+    if not thresh:
+        raise RasaException(f"The variable `thresh` cannot be `{type(thresh)}`.")
```
wochinge, line 129: we should in my opinion also not raise `RasaException` directly but some more specific sub type of it

Imod7: True! Maybe I can raise a `TypeError` like [here](https://github.com/RasaHQ/rasa/blob/1aa1b6c93ba2603064737814f9fcdce764af40d9/rasa/core/actions/action.py#L799)

ancalita: I don't think `TypeError` is ignored by Sentry 🤔 what about adding a new RasaException subclass for this [here](https://github.com/RasaHQ/rasa/blob/1aa1b6c93ba2603064737814f9fcdce764af40d9/rasa/exceptions.py)?

tests/nlu/test_evaluation.py
```diff
@@ -1219,3 +1219,25 @@ def test_replacing_fallback_intent():
...
+def test_plot_attribute_confidences_with_invalid_values(tmp_path: Path):
+    path = tmp_path / "evaluation"
+    path.mkdir()
+    report_folder = str(path / "reports")
+
+    rasa.shared.utils.io.create_directory(report_folder)
+
+    intent_results = [
+        IntentEvaluationResult("", "restaurant_search", 76, -21),
+        IntentEvaluationResult("greet", "greet", -32, 99),
```
Imod7, line 1233: @wochinge I put some random integer values here and this gives me a `pos_hist = [99]` [here](https://github.com/RasaHQ/rasa/blob/1aa1b6c93ba2603064737814f9fcdce764af40d9/rasa/nlu/test.py#L329) so the exception is raised correctly and the test pass. Maybe it is a stupid question but I was just wondering if I will always get the same result? It looks deterministic but I still wanted to make sure and thus ask.

tests/nlu/test_evaluation.py
```diff
@@ -1219,3 +1219,25 @@ def test_replacing_fallback_intent():
...
+    ]
+
+    with pytest.raises(InvalidTypeException):
+        evaluate_intents(
```
ancalita, line 1237: I noticed in the Sentry report that the error was triggered by a call to `evaluate_response_selections` function, what about creating a separate test for this too?

Imod7: Very good catch on seeing that it started from there ! 💡 It looks though that if I call `evaluate_intents` or `evaluate_response_selections`, it does the same thing in a test related level. It arrives at the function `plot_attribute_confidences`, checks that it has an integer in `pos_hist` or `neg_hist` and raises an exception. I would not add a separate test for this but I can replace the current test by using `evaluate_response_selections`.

ancalita: Sounds good 👍

wochinge (changes requested):
I've got a couple of questions around the issues (and hence somewhat about the solution):

* In which cases can the `confidence` property of `ResponseSelectionEvaluationResult` become `None`? Can we reproduce what the user actually ran to produce the error? Might make sense to reach out to Daksh if he knows more.
* Is `None` actually a valid value for `ResponseSelectionEvaluationResult` or is something going wrong beforehand and the error we're seeing is rather a symptom?
* If `None` is valid could we simply ignore it / treat it as `0` instead of raising an error?

Imod7:
Hey @wochinge Thank you so much for your very insightful [comment](https://github.com/RasaHQ/rasa/pull/8747#pullrequestreview-670099681)! 💯💜 These are very valid points to take into account! 

Hey @dakshvar22 ! Any feedback on any of those 3 points mentioned by Tobias would be more than appreciated! 💜In case I can check / test something that enables me to answer them, please let me know! 🙏🏽 Thank you so much in advance!

dakshvar22:
I am not sure how / when `confidence` property of `ResponseSelectionEvaluationResult` can be `None`. I tried the following:

1. Go into `responseselectorbot` under `examples` folder.
2. Train with `rasa train nlu`
3. Eval with `rasa test nlu`

I don't see an error. Next, I remove all retrieval intent examples from the NLU data but keep `ResponseSelector` in my config. Eval still doesn't fail after retraining. Other than that I can't see a situation where confidence would be `None`. Have we been able to reproduce the error?

dakshvar22:
I see this now - 

> I tried to replicate the error message shown in this issue by
> 
> force setting thresh = None in line 127 of this file
> and then running the command
> rasa test nlu --cross-validation
> from an example bot (like formbot)

But looks like `thresh` variable is no longer used in the histogram plotting method of that file. Do we know which exact line was Dominique referring to?

ancalita:
> Have we been able to reproduce the error?

@dakshvar22 Not really from the user perspective - but rather took a shortcut to get the specific error message caught by Sentry. So we're right to think that this could be a usage error and that the suggested fix is appropriate?

wochinge:
>  but rather took a shortcut to get the specific error message caught by Sentry.

I don't think that the shortcut proves anything. We need to actually try to reproduce this (Daksh might just help us to be able to find a fitting scenario quicker)

ancalita:
Hey @dakshvar22 could you please help us with establishing the way forward if we cannot reproduce the user issue?

dakshvar22:
Since the last time this error was generated, the exact implementation of plotting histograms has changed significantly ([PR](https://github.com/RasaHQ/rasa/pull/8471/files)). Also [this](https://github.com/RasaHQ/rasa/pull/8813) now ensures that an exception is raised if the input data to histogram plotting function is empty. So, I am not sure how to reproduce the error anymore. Maybe we can wait out and see if these errors pop up again in the near future?

ancalita:
> Since the last time this error was generated, the exact implementation of plotting histograms has changed significantly ([PR](https://github.com/RasaHQ/rasa/pull/8471/files)). Also [this](https://github.com/RasaHQ/rasa/pull/8813) now ensures that an exception is raised if the input data to histogram plotting function is empty. So, I am not sure how to reproduce the error anymore. Maybe we can wait out and see if these errors pop up again in the near future?

Thanks for the quick update - I guess we can close this PR considering the new changes @wochinge? The exception that gets raised now is `ValueError` so Sentry will pick this up in the future if the problem re-appears. Not sure what we should do about the past issue, if to close it as well?

wochinge:
👍🏻 Let's close the PR and mark the sentry issue  as resolved in release xy (not sure which release contains the updates to the plotting) (both in GitHub and Sentry itself)

ancalita:
This PR is being closed as a result of the issue being handled in a different [PR](https://github.com/RasaHQ/rasa/pull/8813/).

PR #8757, opened 2021-05-27
Fix Sentry PyMongo error

### LINKED ISSUE #8726: OperationFailure: not authorized on logs to execute command { createIndexes: "conversations", indexes: [ { key: { s...
Sentry Issue: [RASA-OPEN-SOURCE-786](https://sentry.io/organizations/rasahq/issues/2260221914/?referrer=github_integration)

```
OperationFailure: not authorized on logs to execute command { createIndexes: "conversations", indexes: [ { key: { sender_id: 1 }, name: "sender_id_1" } ], lsid: { id: UUID("13ff5029-6135-4385-8e55-fe5b616c34c9") }, $db: "logs", $readPreference: { mode: "primary" } }
(16 additional frame(s) were not displayed)
...
  File "rasa/core/run.py", line 255, in load_agent_on_start
    _tracker_store = TrackerStore.create(endpoints.tracker_store, event_broker=_broker)
  File "rasa/core/tracker_store.py", line 123, in create
    return _create_from_endpoint_config(obj, domain, event_broker)
  File "rasa/core/tracker_store.py", line 1235, in _create_from_endpoint_config
    tracker_store = MongoTrackerStore(
  File "rasa/core/tracker_store.py", line 577, in __init__
    self._ensure_indices()
  File "rasa/core/tracker_store.py", line 586, in _ensure_indices
    self.conversations.create_index("sender_id")
```

**Definition of Done**
* Catch `OperationalError` during tracker store creation and convert to `RasaException`

ancalita:
**Proposed changes**:
- Fix #8726 

### DIFF
--- a/rasa/core/tracker_store.py
+++ b/rasa/core/tracker_store.py
@@ -43,7 +43,7 @@ from rasa.shared.core.trackers import (
     DialogueStateTracker,
     EventVerbosity,
 )
-from rasa.shared.exceptions import ConnectionException, RasaException
+from rasa.shared.exceptions import ConnectionException
 from rasa.shared.nlu.constants import INTENT_NAME_KEY
 from rasa.utils.endpoints import EndpointConfig
 import sqlalchemy as sa
--- a/rasa/core/tracker_store.py
+++ b/rasa/core/tracker_store.py
@@ -43,7 +43,7 @@ from rasa.shared.core.trackers import (
     DialogueStateTracker,
     EventVerbosity,
 )
-from rasa.shared.exceptions import ConnectionException
+from rasa.shared.exceptions import ConnectionException, RasaException
 from rasa.shared.nlu.constants import INTENT_NAME_KEY
 from rasa.utils.endpoints import EndpointConfig
 import sqlalchemy as sa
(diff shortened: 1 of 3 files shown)


rasa/core/tracker_store.py
```diff
@@ -129,6 +129,8 @@ def create(
             ConnectionError,
         ) as error:
             raise ConnectionException("Cannot connect to tracker store.") from error
+        except pymongo.errors.OperationFailure as error:
```
joejuzl, line 132: why not just include it in the list above?

joejuzl (changes requested):

rasa/core/tracker_store.py
ancalita: I had a discussion with Tobias in the issue [thread](https://github.com/RasaHQ/rasa/issues/8726#issuecomment-847659020) which concluded that it would be confusing to label it as `ConnectionException`.

I also tried putting the code in the `MongoTrackerStore` constructor in a try-except block, but the unit test i added still failed, because the exception that got raised was still `OperationFailure` instead of `RasaException`. Not sure what I was doing wrong so I ended up putting exception handling in the factory method.

joejuzl: I agree with Tobias that really these framework specific exceptions should be caught in their dedicated implementation classes. However I guess since we are already doing a few here it should be fine.

joejuzl (approved):

PR #8782, opened 2021-05-31
Fix choco install error

HotThoughts:
**Proposed changes**:
- Resolves https://github.com/RasaHQ/rasa/issues/8656

The errors are most likely caused by the downtime of the external server. The package manager we are using for Windows is very unreliable. We can check the status here: https://status.chocolatey.org

There seems [no out-of-box solution for it](https://docs.chocolatey.org/en-us/choco/commands/install). The good news is that the GitHub Action team implemented a choco installation helper in their Windows VM, we can use it directly in the CI.


tczekajlo (approved):
LGTM 🎖️

PR #8799, opened 2021-06-03
Failfast when publishing docs

m-vdb:
**Proposed changes**:
- Failfast in the bash script that pushes docs
- Notify Slack channel when docs workflow fails (currently we rely on "people" telling us that the docs do not appear on the docs website)
- unifies the behaviour with Rasa X (see https://github.com/RasaHQ/rasa-x/pull/5137)

**This PR needs...**:
- [x] Code review


tmbo (approved):
looks good to me ⭐

PR #8802, opened 2021-06-03
[Dockerfile] Create rasa user and group

tczekajlo:
**Proposed changes**:
- Create rasa user and group
- Set the `HOME` environment variable to `/app`


erohmensing (approved):
Looks good 👀 can we also add a changelog entry so that anyone who manually sets the `HOME` env var knows it has changed?

PR #8809, opened 2021-06-03
MultiLabelDotProductLoss for IntentTED (c)

JEM-Mosig:
**Proposed changes**:
- like MultiLabelDotProductLoss for IntentTED (new) #8688 but with kwargs
just for testing


JEM-Mosig:
Merged into intent-ted-johannes-8438b

Issue #8676, opened 2021-05-12
Add Sara and insurance and retail starter packs to regression test configs.


kedz:
**Description of Problem**:
Add Sara, insurance-demo, retail-demo datasets to regression test configs after the [following issue](https://github.com/RasaHQ/rasa/issues/8386) has been resolved.

**Overview of the Solution**:
Edit the `dataset_config_mapping.json` in the `RasaHQ/training-data` repo.
The modified Sara dataset should be:
```json
    "Sara": {
      "path": "public/Sara",
      "language": "en",
      "domain": "domain.yml"
    },
```
the following should be added:
```json
    "retail-demo": {
      "repository": "RasaHQ/retail-demo",
      "language": "en",
      "domain": "domain.yml",
      "train": "data",
      "test": "tests"
    },
    "insurance-demo": {
      "repository": "RasaHQ/insurance-demo",
      "language": "en",
      "domain": "domain.yml",
      "train": "data",
      "test": "tests"
    },
```

**Examples** (if relevant):

**Blockers** (if relevant):
~~the [following issue](https://github.com/RasaHQ/rasa/issues/8386)~~
 - waiting for updated Sara data [PR](https://github.com/RasaHQ/training-data/pull/9) to be merged which includes a `domain.yml`

**Definition of Done**:
- [x] `dataset_config_mapping.json` is modified as specified above.
- [x] Verify that the model regression tests continue to work with these two new test data sets

kedz:
[The blocking issue has been resolved.](event.action_name) This can be implemented now.

samsucik:
Fyi I've added Sara as part of [this `training-data` PR](https://github.com/RasaHQ/training-data/pull/9) (waiting to be merged).

kedz:
@samsucik I think this issue can be closed now as [core regression tests are working](https://github.com/RasaHQ/rasa/pull/9171#issuecomment-883720135) for all three datasets?

alopez:
assignee: @kedz 
reviewer: @aeshky

kedz:
This requires updating following repos:
- [ ] `retail-demo` ([PR](https://github.com/RasaHQ/retail-demo/pull/33)) -- update domain to 2.8
- [ ] `insurance-demo` ([PR](https://github.com/RasaHQ/insurance-demo/pull/18)) -- update domain to 2.8
[...]

PR #8817, opened 2021-06-04
Check the new Sara test data for regression tests

samsucik:
**Proposed changes**:
- Check how the new Sara test set does in our model regression tests before we merge the dataset into `training-data/main` in [this PR](https://github.com/RasaHQ/training-data/pull/9).
- Addresses part of #8676


samsucik:
/modeltest
```yml
dataset_branch: "sara-update"
include:
 - dataset: ["Sara"]
   config: ["all-core"]
```

PR #8843, opened 2021-06-07
`IntentTEDPolicy`: `rasa test` updates

### LINKED ISSUE #8422: IntentTEDPolicy: Adapt `rasa test`

alwx:
**Proposed changes**:
1. Adapt `rasa test` to `IntentTEDPolicy`:
- if test story contains `action_unlikely_intent` and Rasa does not predict this, this leads to 
a test error (wrongly predicted action) 
- if the story does not contain `action_unlikely_intent` and Rasa Open Source does predict it then 
the prediction of `action_unlikely_intent` will be ignored for the evaluation (and hence not lead 
to a wrong prediction) but the story will be logged in a file called `stories_with_warnings.yml`

2. Add `with_warnings` key to `story_report.json` — it contains the number of stories that have warnings.

3. Fix #8422


data/test_moodbot/intent_ted_policy_config.yml
```diff
@@ -0,0 +1,23 @@
...
+    analyzer: char_wb
+    min_ngram: 1
+    max_ngram: 4
+  - name: "DIETClassifier"
+    entity_recognition: False
+    epochs: 1
+
+policies:
+  - name: MemoizationPolicy
+  - name: IntentTEDPolicy
```
joejuzl, line 19: Should we have `TEDPolicy` too so that it can predict actions. Or is `MemoizationPolicy` enough for that?

examples/intent_ted/config.yml
```diff
@@ -0,0 +1,39 @@
...
+#     ambiguity_threshold: 0.1
+
+# Configuration for Rasa Core.
+# https://rasa.com/docs/rasa/core/policies/
+policies:
+  # # No configuration for policies was provided. The following default policies were used to train your model.
+  # # If you'd like to customize them, uncomment and adjust the policies.
+  # # See https://rasa.com/docs/rasa/policies for more information.
+  - name: MemoizationPolicy
+  - name: IntentTEDPolicy
```
joejuzl, line 35: Ditto about `TEDPolicy`.

rasa/core/test.py
```diff
@@ -619,11 +627,20 @@ def _collect_action_executed_predictions(
...
+        action_executed_eval_store.add_to_store(
+            action_predictions=[predicted], action_targets=[gold]
+        )
 
-    if action_executed_eval_store.has_prediction_target_mismatch():
+    has_prediction_target_mismatch = (
+        action_executed_eval_store.has_prediction_target_mismatch()
+    )
+    if has_prediction_target_mismatch or (
+        should_ignore_prediction and predicted != gold
```
joejuzl, line 642: I don't totally understand this line, why do we need this?

tests/core/test_test.py
```diff
@@ -40,3 +40,57 @@ async def test_testing_valid_with_non_e2e_core_model(core_agent: Agent):
...
+    tmpdir: Path, intent_ted_policy_moodbot_agent: Agent
+):
+    result = await rasa.core.test.test(
+        "data/test_yaml_stories/test_action_unlikely_intent_2.yml",
+        intent_ted_policy_moodbot_agent,
+        out_directory=str(tmpdir),
+    )
+    assert "report" in result.keys()
+    assert result["report"]["conversation_accuracy"]["correct"] == 1
+    assert result["report"]["conversation_accuracy"]["with_warnings"] == 1
```
joejuzl, line 68: Should we have a warning if `action_unlikely_intent` was in the test story?

tests/core/test_test.py
```diff
@@ -40,3 +40,57 @@ async def test_testing_valid_with_non_e2e_core_model(core_agent: Agent):
         "data/test_yaml_stories/test_stories_entity_annotations.yml", core_agent
     )
     assert "report" in result.keys()
+
+
+async def test_action_unlikely_intent_warnings_1(
+    tmpdir: Path, intent_ted_policy_moodbot_agent: Agent
+):
+    result = await rasa.core.test.test(
+        "data/test_yaml_stories/test_action_unlikely_intent_1.yml",
```
joejuzl, line 49: I wonder if it would be more readable if the stories were inline in this file?

joejuzl (changes requested):

data/test_yaml_stories/test_action_unlikely_intent_1.yml
```diff
@@ -0,0 +1,7 @@
+stories:
+  - story: unlikely path
```
ancalita, line 2: Could you rename either the file and/or the story name to explicitly refer to the case being tested (some catchy reference to distinguish between `test_action_unlikely_intent_1` and `test_action_unlikely_intent_2`) ?

rasa/core/test.py
```diff
@@ -156,6 +160,10 @@ def has_prediction_target_mismatch(self) -> bool:
             or self.action_predictions != self.action_targets
         )
 
+    @staticmethod
+    def should_ignore_prediction(predicted: Text) -> bool:
```
ancalita, line 164: Shouldn't this have a short docstring?

rasa/core/test.py
```diff
@@ -619,11 +627,20 @@ def _collect_action_executed_predictions(
...
             predicted = action.name()
 
-    action_executed_eval_store.add_to_store(
-        action_predictions=[predicted], action_targets=[gold]
+    should_ignore_prediction = action_executed_eval_store.should_ignore_prediction(
+        predicted
     )
+    if not should_ignore_prediction:
+        action_executed_eval_store.add_to_store(
+            action_predictions=[predicted], action_targets=[gold]
```
ancalita, line 635: This might be a silly q - what does `gold` mean in this context - I can see it's defined earlier, but I don't grasp what it means?

rasa/core/test.py
```diff
@@ -719,19 +736,20 @@ async def _predict_tracker_actions(
...
-                    "predicted": action_executed_result.action_predictions[0],
-                    "policy": prediction.policy_name,
-                    "confidence": prediction.max_confidence,
-                }
-            )
-            should_predict_another_action = processor.should_predict_another_action(
-                action_executed_result.action_predictions[0]
-            )
-            num_predicted_actions += 1
+            if action_executed_result.action_targets:
```
ancalita, line 739: Why is there a need for a `if` condition here?

rasa/core/test.py
```diff
@@ -960,11 +993,19 @@ async def test(
         _log_stories(
             story_evaluation.failed_stories,
             os.path.join(out_directory, FAILED_STORIES_FILE),
+            "None of the test stories failed - all good!",
         )
     if successes and out_directory:
         _log_stories(
             story_evaluation.successful_stories,
             os.path.join(out_directory, SUCCESSFUL_STORIES_FILE),
+            "None of the test stories succeeded :(",
```
ancalita, line 1002: I might be confused now - but shouldn't this comment be in the `if errors` branch earlier - and then the `None of the test stories failed - all good!" comment should be in the `if successes` branch? Basically, should they be swapped?

data/test_moodbot/intent_ted_policy_config.yml
alwx: It's enough to have `IntentTEDPolicy` and `MemoizationPolicy`

tests/core/test_test.py
alwx: Well, at least that's what's written in the task description but we can try changing that.

tests/core/test_test.py
alwx: that would mean mocking `_create_data_generator` and `TrainingDataImporter.load_from_dict`, and that's something I would like to avoid in order to make these tests simpler

rasa/core/test.py
alwx: Added it!

rasa/core/test.py
alwx: It's the action that's specified in the test story. I guess it's `gold` because we treat as the expected action, something like the "golden standard" or whatever.
Even though it was defined earlier, I've actually just replaced it with `expected_action` just because it makes a bit more sense in this context.

rasa/core/test.py
alwx: No, this message is what gets printed when there are no successful stories — `successes` is basically a CLI flag that specifies if we need to create a `SUCCESSFUL_STORIES_FILE`

rasa/core/test.py
alwx: Because now the `action_unlikely_intent` will not be added to `action_executed_result.action_targets` and it means getting `action_executed_result.action_targets[0]` might be problematic

tests/core/test_test.py
alwx: Done!

rasa/core/test.py
alwx: I updated it a bit so I guess it's more clear now.
The idea is simple — we add `WronglyPredictedAction` not only if there is a prediction mismatch but also when `action_unlikely_intent` got predicted.

ancalita (approved):
Thanks for the explanations 🚀

tests/core/test_test.py
```diff
@@ -40,3 +40,57 @@ async def test_testing_valid_with_non_e2e_core_model(core_agent: Agent):
         "data/test_yaml_stories/test_stories_entity_annotations.yml", core_agent
     )
     assert "report" in result.keys()
+
+
+async def test_action_unlikely_intent_1(
+    tmpdir: Path, intent_ted_policy_moodbot_agent: Agent
```
wochinge, line 46: `tmpdir` doesn't have type `Path` -> I'd suggest to use `tmp_path` instead

rasa/core/test.py
```diff
@@ -617,24 +635,33 @@ def _collect_action_executed_predictions(
             # Even if the prediction is also wrong, we don't have to undo the emulation
             # of the action rejection as we know that the user explicitly specified
             # that something else than the form was supposed to run.
-            predicted = action.name()
+            predicted_action = action.name()
 
-    action_executed_eval_store.add_to_store(
-        action_predictions=[predicted], action_targets=[gold]
+    should_ignore_prediction = action_executed_eval_store.should_ignore_prediction(
```
joejuzl, line 640: I think the name `should_ignore_prediction` was confusing me before, as we do still compare the prediction. Maybe a name like `predicted_warning` would be clearer?

rasa/core/test.py
```diff
@@ -845,12 +882,14 @@ async def _collect_story_predictions(
     )
 
 
-def _log_stories(trackers: List[DialogueStateTracker], file_path: Text) -> None:
+def _log_stories(
+    trackers: List[DialogueStateTracker], file_path: Text, comment: Text
```
joejuzl, line 886: `comment` -> `message_if_no_trackers`?

tests/core/test_test.py
joejuzl: you could also just write it to a temporary yaml file.

joejuzl (changes requested):

rasa/core/test.py
```diff
@@ -811,11 +840,18 @@ async def _collect_story_predictions(
...
+            failed_stories.append(predicted_tracker)
             correct_dialogues.append(0)
         else:
+            successful_stories.append(predicted_tracker)
             correct_dialogues.append(1)
-            success.append(predicted_tracker)
+
+            if any(
+                type(event) == WronglyPredictedAction
+                and event.action_name_prediction == ACTION_UNLIKELY_INTENT_NAME
```
joejuzl, line 851: Could we use the `predicted_warning` function here too?

joejuzl (approved):
Looks good! 🚀 
Added one more small comment.

dakshvar22:
@alwx Why are we adding a new example bot as part of this PR?

changelog/8422.improvement.md
```diff
@@ -0,0 +1,8 @@
+Adapt `rasa train` to `IntentTEDPolicy`:
```
dakshvar22, line 1:
```suggestion
Adapt `rasa test` to `IntentTEDPolicy`:
```

PR #8849, opened 2021-06-08
second empty branch for pull request regression test

aeshky:
An empty pull request to test the fix for the following regression test failure: https://github.com/RasaHQ/rasa/issues/8362


aeshky:
/modeltest

```yml
include:
 - dataset: ["Private 2"]
```

Issue #8774, opened 2021-05-28
Async trackerstore support


leonidas-tolias:
**Description of Problem**: Currently, the trackerstore interface uses synchronous calls to save and retrieve conversations. This leads to performance issues as any non-memory trackerstore will use network calls and block the event loop.

**Overview of the Solution**: To resolve, there needs to be an update of all uses of the trackerstore override endpoints (save, retrieve, keys, etc) to conditionally await the results of these function in the case that they have been overridden with async functions.

**Definition of Done**: Feature will be considered complete when we can register a trackerstore with either sync or async override functions.

*   [ ] Tests are added
*   [ ] Feature described the docs
*   [ ] Feature mentioned in the changelog

leonidas-tolias:
I'm working on submitting a pr to implement this functionality

sara-tagger:
Thanks for submitting this feature request 🚀 @kalkbrennerei will get back to you about it soon! ✨

wochinge:
@leonidas-tolias That definitely seems worth doing. One thing we have to consider is that users might be using custom tracker stores which aren't `async` yet. These should continue to work (but raise a deprecation message) until the next major (3.0.0). I think before you get started implementing it would be nice if you could outline the details of the change as comment in this issue as this is quite a sensitive and big change. Thank you 🙌🏻

leonidas-tolias:
Hey, for backwards compatibility I mirrored the approach used with FailSafeTrackerStore and created a new wrapper class which checks if its child trackerstore has the save/retrieve/etc functions implemented as async or normal functions and calls them with await (or not) accordingly.  
The remaining changes were just to add an await everywhere the save/retrieve calls are made (mainly in processor.py, server.py) and to have agent.py create an AwaitableTrackerStore(FailSafeTrackerStore(tracker_store)) at startup.  In my testing so far, this seems to work fine with both async and sync trackerstore implementations. 
Implementation is mostly done, just fighting my firms internal dependency systems to get all the tests to run : ), after which I can submit a PR.  Let me know if you'd prefer some other approach.

wochinge:
>  I mirrored the approach used with FailSafeTrackerStore and created a new wrapper class which checks if its child trackerstore has the save/retrieve/etc functions implemented as async or normal functions and calls them with await (or not) accordingly.

Slick approach! Wondering if that makes the class structure a bit too nested. Anyway, we can iterate on that in the PR comments then 🚀 

Excited for your change! 💯

[...]

PR #8854, opened 2021-06-09
Async trackerstore support Issue #8774

leonidas-tolias:
**Description of Problem**:
Currently, the trackerstore interface uses synchronous calls to save and retrieve conversations.  This leads to performance issues as any non-memory trackerstore will use network calls and block the event loop.

**Overview of the Solution**:
To resolve, there needs to be an update of all uses of the trackerstore override endpoints (save, retrieve, keys, etc) to conditionally await the results of these function in the case that they have been overridden with async functions.

**Definition of Done**:
Feature will be considered complete when we can register a trackerstore with either sync or async override functions.
- [x] Tests are added
- [ ] Feature described the docs
- [ ] Feature mentioned in the changelog


sara-tagger:

akelad:
@kalkbrennerei @wochinge putting this in Enable's inbox

data/test_config/config_defaults.yml
```diff
@@ -23,3 +23,13 @@ pipeline: []
 #     ambiguity_threshold: 0.1
 
 data:
+policies:
```
wochinge, line 26: I think that's an unwanted change, isn't it?

Same as the `TEDPolicy_confusion_matrix.png` and `other`

rasa/cli/export.py
```diff
@@ -66,9 +66,9 @@ def _get_tracker_store(endpoints: "AvailableEndpoints") -> "TrackerStore":
             f"Exiting. "
         )
 
-    from rasa.core.tracker_store import TrackerStore
+    from rasa.core.tracker_store import TrackerStore, AwaitableTrackerStore
 
-    return TrackerStore.create(endpoints.tracker_store)
+    return AwaitableTrackerStore.create(TrackerStore.create(endpoints.tracker_store))
```
wochinge, line 71: This will hide the use tracker store in the telemetry tracking [here](https://github.com/RasaHQ/rasa/blob/e09e1310754e6eb69441e7b4154c8d6d37cc7b16/rasa/telemetry.py#L806)

rasa/core/agent.py
```diff
@@ -35,6 +35,7 @@
     FailSafeTrackerStore,
```
wochinge, line 35: can you please fix the flask8 warnings?

rasa/core/tracker_store.py
```diff
@@ -65,7 +67,7 @@
 DEFAULT_REDIS_TRACKER_STORE_KEY_PREFIX = "tracker:"
 
 
-class TrackerStore:
+class TrackerStore: #retrieve keys get_or_create_tracker create_tracker exists stream_events  number_of_existing_events retrieve_full_tracker save
```
wochinge, line 70: can we please remove this comment?

rasa/core/tracker_store.py
```diff
@@ -1238,7 +1421,7 @@ def _create_from_endpoint_config(
 
     if endpoint_config is None or endpoint_config.type is None:
         # default tracker store if no type is set
-        tracker_store = InMemoryTrackerStore(domain, event_broker)
+        tracker_store = AwaitableTrackerStore.create(InMemoryTrackerStore(domain, event_broker))
```
wochinge, line 1424: how about we swap your approach and only wrap the `tracker_store` in an awaitable `TrackerStore` when `_load_from_module_name_in_endpoint_config` is called? We know from the inbox tracker stores that they will already support `async`

wochinge (commented):
Thanks for the PR! Nice change! 💯 

I'd vote to decrease the surface area of the change by only wrapping `TrackerStore` implementations which don't support `async` yet in an `AwaitableTrackerStore`. What do you think?

wochinge:
@leonidas-tolias Are you still working on this? Would be great to get this one in

wochinge:
@leonidas-tolias Anything I can help you with?

wochinge:
I'll close the PR for now @leonidas-tolias Feel free to re-open if you're picking this up again. Still very much excited for this change 🎉 🎉

PR #8855, opened 2021-06-09
IntentTEDPolicy: Add `tolerance` parameter

dakshvar22:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/7412
- Add `tolerance` parameter to `IntentTEDPolicy` and use it for computing thresholds.


rasa/core/policies/intent_ted_policy.py
```diff
@@ -536,6 +552,117 @@ def _check_unlikely_intent(
...
+                )
+                if negative_scores
+                else min(positive_scores)
+                for tolerance_value in range(0, 100, 5)
+            ]
+
+        return label_tolerance_thresholds
+
+    @staticmethod
+    def _compute_exact_label_thresholds_from_tolerance(
```
JEM-Mosig, line 646: Isn't this more like `_pick_threshold_from_quantiles`?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -239,7 +243,14 @@ def __init__(
             **kwargs,
         )
 
-        self.label_thresholds = label_thresholds
+        self.label_tolerance_thresholds = label_tolerance_thresholds
```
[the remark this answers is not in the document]
JEM-Mosig: Also, I'd rather call this something like `negative_score_quantiles`

rasa/core/policies/intent_ted_policy.py
```diff
@@ -239,7 +243,14 @@ def __init__(
...
+        self.label_thresholds = (
```
JEM-Mosig, line 247: I'd put this into a separate public function. Here we kind of assume that you have to re-instantiate the model for prediction. If you wanted to train the model and then use the same instance for inference, you'd have to set this `label_thresholds` yourself, using the private method. Not that we need this, but it'd be nicer to have a separate function I think.

rasa/core/policies/intent_ted_policy.py
JEM-Mosig: or just `_pick_threshold`?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -536,6 +552,117 @@ def _check_unlikely_intent(
...
+
+        """
+
+        def _get_threshold_for_tolerance(
+            negative_values, tolerance_percentile
+        ) -> float:
+            index = max(
+                0,
+                min(
+                    np.size(negative_values) - 1,
```
JEM-Mosig, line 622: So we don't allow tolerances > 1? Just checking.

JEM-Mosig (changes requested):
I only reviewed `intent_ted_policy.py` for now. The naming is a bit confusing (though it might have been clearer with the comments that you are still adding) and I think you can use `np.quantile`.

rasa/core/policies/intent_ted_policy.py
```diff
@@ -536,6 +552,117 @@ def _check_unlikely_intent(
...
+        """Construct
+
+        Args:
+            label_id_scores:
+
+        Returns:
+
+        """
+
+        def _get_threshold_for_tolerance(
```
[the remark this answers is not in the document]
dakshvar22: We could but we want the exact opposite logic of how quantiles are calculated. In other words, we can use it but then we will have to change the sign of each score, pass it to `np.quantile`, change the sign of the result again and use it.
I thought that brings in hidden complexity to understand the code. I don't have a strong preference though. Wdyt?

rasa/core/policies/intent_ted_policy.py
dakshvar22: Nope, I've added a check in `train_utils.py` for it.

rasa/core/policies/intent_ted_policy.py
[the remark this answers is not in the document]
dakshvar22: Ohh wait, what am i even saying 🙈  I can just pass `1-quantile` to `np.quantile`. Will do that!

rasa/core/policies/intent_ted_policy.py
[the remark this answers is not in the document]
dakshvar22: How does `label_quantiles` sound?

rasa/core/policies/intent_ted_policy.py
dakshvar22: We are trying to set an instance variable of the class which needs to be initialized to something in the constructor. I think it would be an overkill if we first declare it with a default value and then set it in a public function called after.

dakshvar22: Do you have a workaround in mind?

dakshvar22:
@kedz Requested your review as @JEM-Mosig is offline until next wednesday.

kedz:
@dakshvar22 will do this monday.

rasa/core/policies/intent_ted_policy.py
kedz: @dakshvar22 how do you envision modifying tolerance at prediction time? Would that happen at instantiation or would this be done after instantiation. If the latter then I agree with @JEM-Mosig this should be a public method.

dakshvar22: The policy is loaded from the `load` method (which calls its own constructor to return an object) with the config used during training. So, as of now tolerance cannot be modified at prediction time because the trained config cannot be changed. Once that's possible, the `load` method should really be called with the new config (I am speculating here as it depends on how we really make it possible) and hence the constructor will take that into account. Based on this do you still recommend we refactor the logic into a separate public method?

rasa/core/policies/intent_ted_policy.py
```diff
@@ -211,8 +213,21 @@ class IntentTEDPolicy(TEDPolicy):
...
+        # List of intents to ignore for `action_unlikely_intent` prediction.
         IGNORE_INTENTS_LIST: [],
+        # Tolerance for prediction of action_unlikely_intent.
+        # This is specified as a ratio of negative trackers for
+        # which the predicted similarity score of the
+        # corresponding correct label is above a threshold.
+        # Hence, any tracker at inference time
+        # should result in a predicted similarity score equal
+        # to or above the threshold in order to avoid triggering
+        # `action_unlikely_intent`. Higher values of `tolerance`
```
kedz, line 225: I'm having trouble making sense of this description. Is the following technically correct and if so would it suffice as a description:
For each intent, the tolerance is the percentage of negative training instances that would be ignored by IntentTED. This is converted into a similarity threshold by identifying the similarity score for the (1 - tolerance) percentile of negative examples. Any tracker with a similarity score below this threshold will trigger an `action_unlikely_intent`.

rasa/core/policies/intent_ted_policy.py
```diff
@@ -536,6 +566,115 @@ def _check_unlikely_intent(
...
+                if negative_scores
+                else min(positive_scores)
+                for tolerance_value in range(0, 100, 5)
+            ]
+
+        return label_quantiles
+
+    @staticmethod
+    def _pick_thresholds(
+        label_quantiles: Dict[int, List[float]], tolerance: float
```
kedz, line 655: Should we raise and exception if tolerance is not in [0,1]?

rasa/utils/train_utils.py
```diff
@@ -532,6 +533,16 @@ def validate_configuration_settings(component_config: Dict[Text, Any]) -> None:
...
     _check_similarity_loss_setting(component_config)
+    _check_tolerance_setting(component_config)
+
+
+def _check_tolerance_setting(component_config: Dict[Text, Any]) -> None:
+    if not (0.0 <= component_config.get(TOLERANCE, 0.0) <= 1.0):
+        raise InvalidConfigException(
+            f"{TOLERANCE} was set to {component_config.get(TOLERANCE)} "
+            f"which is an invalid setting. Please set it to a value "
+            f"between 0.0 and 1.0 ."
```
kedz, line 544: "between 0.0 and 1.0 **inclusive**."

rasa/core/policies/intent_ted_policy.py
```diff
@@ -536,6 +566,115 @@ def _check_unlikely_intent(
...
+        Returns:
+            Both buckets of similarity scores grouped by each unique label id.
+        """
+        label_id_scores: Dict[int, List[List[float]]] = defaultdict(list)
+        unique_label_ids = np.unique(label_ids).tolist()
+        if LABEL_PAD_ID in unique_label_ids:
+            unique_label_ids.remove(LABEL_PAD_ID)
+
+        for label_id in unique_label_ids:
+            label_id_scores[label_id] = [[], []]
```
kedz, line 594: I think it would be better to make this a dict with keys like "positive" and "negative" to avoid having to remember which index is which.

rasa/core/policies/intent_ted_policy.py
```diff
@@ -536,6 +566,115 @@ def _check_unlikely_intent(
...
+                    min(positive_scores),
+                    np.quantile(
+                        negative_scores,
+                        1 - tolerance_value / 100.0,
+                        interpolation="lower",
+                    ),
+                )
+                if negative_scores
+                else min(positive_scores)
+                for tolerance_value in range(0, 100, 5)
```
kedz, line 648: Perhaps the granularity of this calculation should be a variable/a policy parameter? We may want to have increments of 1 for instance.

kedz: The if statement for a condition that never changes in the loop is a little hard to read. Why not:
```python
if negative_scores:
  # .. make quantile thresholds ..
else:
  # .. make default thresholds ..
```

Additionally, this for loop could be made into a vector op since np.quantile takes an array of quantiles to compute.

rasa/core/policies/intent_ted_policy.py
kedz, line 655: I know we check the config to confirm tolerance is in [0, 1] but maybe we should raise an exception here if it is not, in case any one every takes intentTED for a spin outside the standard codebase?

tests/core/policies/test_intent_ted_policy.py
```diff
@@ -197,29 +267,99 @@ async def test_post_training_threshold_computation(
...
+                -0.1,
+                -0.1,
+                -0.1,
+                -0.1,
+                -0.1,
+                -0.1,
+                -0.1,
+                -0.1,
+            ],
+        }
```
kedz, line 341: This test case feels somewhat unnecessary since the value never changes. Can you make a sequence of monotonically decreasing floats distinct from the other one?

tests/core/policies/test_intent_ted_policy.py
```diff
@@ -229,58 +369,58 @@ def test_unlikely_intent_check(
...
-        unlikely_intent_prediction = trained_policy._check_unlikely_intent(
+        unlikely_intent_prediction = loaded_policy._check_unlikely_intent(
             default_domain, similarities, query_intent
         )
 
         assert is_unlikely == unlikely_intent_prediction
 
         # Revert back the label thresholds to original values.
-        trained_policy.label_thresholds = original_label_thresholds
+        loaded_policy.label_thresholds = original_label_thresholds
```
kedz, line 384: Is modifying a fixture safe to do? Tests can be run with multiple threads or tests could fail before resetting? Perhaps we should clone the fixture at the start of the test instead?

kedz (changes requested):
I had mostly minor comments or suggestions for improving clarity but the overall structure of the code is in good shape. I had one concern about whether it is good practice to modify test fixtures but am not an experienced enough software engineer to know what the right thing to do here is. Perhaps we could get someone from enable to look at that?

rasa/core/policies/intent_ted_policy.py
kedz: In that case (constructor is called with a new config) I think it is not needed to have a separate public method.

tests/core/policies/test_intent_ted_policy.py
dakshvar22: > This test case feels somewhat unnecessary since the value never changes.

That's the purpose of the test to check if the threshold is picked correctly even if there is just one unique value for all tolerances for a particular label id.

> Can you make a sequence of monotonically decreasing floats distinct from the other one?

What would we be testing there? Do you want to see if the threshold is picked correctly when all thresholds corresponding to every possible value of tolerance is unique?

rasa/core/policies/intent_ted_policy.py
dakshvar22: Thanks for the suggestion. I incorporated it in.

rasa/core/policies/intent_ted_policy.py
dakshvar22: Refactored 👍

rasa/core/policies/intent_ted_policy.py
dakshvar22: > Perhaps the granularity of this calculation should be a variable/a policy parameter? 

I don't think we have enough evidence to show that people would like to customize this. In general, we try to reduce the number of configurable parameters as much as possible so that users don't need to think a lot about what parameters to tune. Moreover, adding new configuration parameters is always easier than removing one so I would prefer to take a more conservative approach to it until we have had more feedback from users. However, I think we should describe in the docs that `tolerance` can only be tuned in increments of `0.05` for now. Will take care of that in the docs PR. Wdyt?

Took care of the other comment on the restructuring the function. Thanks for the push 👍

rasa/core/policies/intent_ted_policy.py
dakshvar22: Even then, they would be expected to initialize `IntentTEDPolicy` with its constructor or the `load` method (since they are the public methods) which means the check will be applied and will trigger an exception if tolerance is not in accepted ranges.

tests/core/policies/test_intent_ted_policy.py
dakshvar22: Added a test where all quantiles are unique and monotonically decreasing.

dakshvar22: On second thought, I think what you meant was checking if the correct thresholds are picked for all possible values of tolerance, that is `range(0.0, 1.1, 0.05)`. Is that correct?

rasa/core/policies/intent_ted_policy.py
kedz: Looks great!

rasa/core/policies/intent_ted_policy.py
kedz: great points! 

> However, I think we should describe in the docs that tolerance can only be tuned in increments of 0.05 for now. 

I think that is more my concern, that we had this parameter advertised as continuous but under the hood it is discrete. I think documenting this primarily solves this issue.

tests/core/policies/test_intent_ted_policy.py
kedz: No you understood me right the first time but I see your point now with this statement:

> the purpose of the test to check if the threshold is picked correctly even if there is just one unique value

so withdrawing my comment.

rasa/core/policies/intent_ted_policy.py
kedz: Thanks! Can you not also make this line: 

```python
label_id_scores[label_id] = [[], []]
```

into

```python
label_id_scores[label_id] = {POSITIVE_SCORES_KEY: [], NEGATIVE_SCORES_KEY: []}
```

and the corresponding changes to the lines where you append the scores?

dakshvar22: My bad, I should have done this in the first place.

tests/core/policies/test_intent_ted_policy.py
kedz: But also appreciate the extra test :) !

kedz (changes requested):
All of my implementation comments have been addressed. The final thing that needs changing is that we clone any test fixtures that we modify -- after that is done I'll approve the PR.

tests/core/policies/test_intent_ted_policy.py
dakshvar22: Removed the fixture. Instead using a common function which persists and loads the `trained_policy`(which is a fixture). All changes are made to the loaded policy and hence the fixture remains untouched.

kedz (approved):
All comments were addressed. New approach to test fixtures looks good.

PR #8875, opened 2021-06-14
Best model checkpoint allows metrics to be equal to previous best if at least one metric improves

koernerfelicia:
duplicate of https://github.com/RasaHQ/rasa/pull/8467, redone to target correct branch


samsucik (approved):
Let's do this! 🚀

PR #8885, opened 2021-06-15
Apply suggestions from code review post merge of #8875

koernerfelicia:
See [here](https://github.com/RasaHQ/rasa/pull/8467#issuecomment-861456671)


tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -576,11 +576,11 @@ async def test_train_tensorboard_logging(
 
 
 async def test_train_model_checkpointing(
-    component_builder: ComponentBuilder, tmpdir: Path, nlu_data_path: Text,
+    component_builder: ComponentBuilder, tmp_path: Path, nlu_data_path: Text,
 ):
     model_name = "nlu-checkpointed-model"
-    model_dir = Path(tmpdir / model_name)
-    checkpoint_dir = get_checkpoint_dir_path(tmpdir, model_dir)
+    model_dir = Path(tmp_path / model_name)
```
wochinge, line 582: you don't need the extra `Path` cast then. With `/` only the first param needs to be of type `Path` everything else will become a `Path` then 
```suggestion
    model_dir = tmp_path / model_name
```

wochinge (approved):
Thanks for following up on these 💯 💜

tests/nlu/classifiers/test_diet_classifier.py
koernerfelicia: Cool, thank you! Removed the casts in the helper methods as well :)

Issue #8525, opened 2021-04-22
Implement code boundaries in CODEOWNERs file


wochinge:
**Description of Problem**:
We recently split the codebase into Research / Engineering owned code: https://www.notion.so/rasa/Proposal-for-Engineering-Research-Code-boundaries-d292e938324b416db03b382503690839?d=a6365544-6866-4b9c-bffe-a8a40a3e7bec

**Overview of the Solution**:
Adapt the `CODEOWNERS` file to reflext the ownerships specified in the document.

**Definition of Done**:
- [x] CODEOWNERs file is updated.
- [x] Share message in channel that this is in place

joejuzl:
I've moved this above the prio line as we should follow up on this to make use of all the work that went into the determining the boundaries.

PR #8891, opened 2021-06-15
Implement code boundaries

ancalita:
**Proposed changes**:
- Implement suggested changes to CODEOWNERS file as described in #8525 


HotThoughts:
@virtualroot could you add `@RasaHQ/infrastructure` and `@RasaHQ/research` to this repo? Looks like the same situation we have for rasa-sdk earlier. We need to grant write access to them so they can get assigned. 

> The people you choose as code owners must have write permissions for the repository. When the code owner is a team, that team must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.
> 
> https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners

virtualroot:
> @virtualroot could you add `@RasaHQ/infrastructure` and `@RasaHQ/research` to this repo? Looks like the same situation we have for rasa-sdk earlier. We need to grant write access to them so they can get assigned.

✅ Done, `@RasaHQ/infrastructure` and `@RasaHQ/research` added with write access.

CODEOWNERS
```diff
@@ -1,5 +1,112 @@
 # Code in the `rasa.shared` package is potentially re-used by downstream dependencies
 # such as Rasa X. Hence, changes within this package require double checking.
-/rasa/shared/ @RasaHQ/enable-squad
 /docs/docs/prototype-an-assistant.mdx @ricwo @alwx
 /.github/workflows/ @RasaHQ/infrastructure
+/rasa/cli/ @RasaHQ/enable-squad
+/rasa/core/actions/ @RasaHQ/enable-squad
+/rasa/core/brokers/ @RasaHQ/enable-squad
+/rasa/core/channels/ @RasaHQ/enable-squad
+/rasa/core/featurizers/
```
wochinge, line 9:
```suggestion
/rasa/core/featurizers/ @RasaHQ/research
```

CODEOWNERS
```diff
@@ -1,5 +1,112 @@
...
+/rasa/core/policies/registry.py @RasaHQ/enable-squad
+/rasa/core/policies/__init__.py @RasaHQ/enable-squad
+/rasa/core/policies/fallback.py @RasaHQ/research
+/rasa/core/policies/form_policy.py @RasaHQ/research
+/rasa/core/policies/mapping_policy.py @RasaHQ/research
+/rasa/core/policies/memoization.py @RasaHQ/research
+/rasa/core/policies/rule_policy.py @RasaHQ/research
+/rasa/core/policies/sklearn_policy.py @RasaHQ/research
+/rasa/core/policies/ted_policy.py @RasaHQ/research
+/rasa/core/policies/two_stage_fallback.py @RasaHQ/research
```
wochinge, line 22: How about simplifying this and doing something like the following (according to the GitHub docs the **last match** takes precedence)
```suggestion
/rasa/core/policies/*.py @RasaHQ/research
/rasa/core/policies/ensemble.py @RasaHQ/enable-squad
/rasa/core/policies/registry.py @RasaHQ/enable-squad
```

CODEOWNERS
```diff
@@ -1,5 +1,112 @@
...
+/rasa/core/lock.py @RasaHQ/enable-squad
+/rasa/core/lock_store.py @RasaHQ/enable-squad
+/rasa/core/processor.py @RasaHQ/enable-squad
+/rasa/core/registry.py @RasaHQ/enable-squad
+/rasa/core/run.py @RasaHQ/enable-squad
+/rasa/core/test.py @RasaHQ/enable-squad
+/rasa/core/tracker_store.py @RasaHQ/enable-squad
+/rasa/core/train.py @RasaHQ/enable-squad
+/rasa/core/utils.py @RasaHQ/enable-squad
+/rasa/core/visualize.py @RasaHQ/enable-squad
```
wochinge, line 40: this seems to be mostly enable-squad - what about making enable-squad default for `rasa/core/*` and then just adding the exceptions for research?

wochinge (commented):
Thanks for taking this one 💯 What do you think of adding a default rule which states that every python file should be reviewed by `@enable-squad` and then only adding the exception for research? This would in my opinion have the following advantages:
* if somebody creates a new file, it would automatically have at least some ownership. If think this doesn't make sense for us to review we can still change the code owners file but nothing falls through the cracks
* hopefully shorter and easier to read `CODEOWNERS` file?

CODEOWNERS
ancalita: Sure, I'll action that - how do I make sure that a certain file has no codeowners, would this work?
`! /rasa/core/constants.py @RasaHQ/enable-squad`

ancalita:
> What do you think of adding a default rule which states that every python file should be reviewed by `@enable-squad` and then only adding the exception for research

Sure - would this be something like this `*.py @RasaHQ/enable-squad` added right at the top after L4?

CODEOWNERS
wochinge: oh, very good catch! Should we just try out? 🙈

CODEOWNERS
```diff
@@ -1,5 +1,40 @@
 # Code in the `rasa.shared` package is potentially re-used by downstream dependencies
 # such as Rasa X. Hence, changes within this package require double checking.
-/rasa/shared/ @RasaHQ/enable-squad
 /docs/docs/prototype-an-assistant.mdx @ricwo @alwx
 /.github/workflows/ @RasaHQ/infrastructure
+*.py @RasaHQ/enable-squad
+/rasa/cli/ @RasaHQ/enable-squad
+/rasa/core/ @RasaHQ/enable-squad
```
wochinge, line 7: do we even need these two then?

CODEOWNERS
```diff
@@ -1,5 +1,40 @@
 # Code in the `rasa.shared` package is potentially re-used by downstream dependencies
 # such as Rasa X. Hence, changes within this package require double checking.
-/rasa/shared/ @RasaHQ/enable-squad
 /docs/docs/prototype-an-assistant.mdx @ricwo @alwx
 /.github/workflows/ @RasaHQ/infrastructure
+*.py @RasaHQ/enable-squad
```
wochinge, line 5: I think we've to limit this to the `rasa/` dir (otherwise we also get assigned for tests). According to the docs it would be something like this:
```suggestion
rasa/**.py @RasaHQ/enable-squad
```

CODEOWNERS
wochinge: (also some of them below seem to become superfluous I think 🤔

wochinge (commented):
I think we should be able to test it by just creating PRs against this branch, no?

CODEOWNERS
ancalita: Alright, I'm thinking I should create branches for 4 distinct changes in: 
-  Enable owned file
-  file without owner
-  Research owned file
- dual Enable-Research ownership
I'll make some minimal change (add a comment this is for testing CODEOWNERS so it acts as a heads-up 😄 )

ancalita: Hmm, first test didn't work out :( I created [this test PR](https://github.com/RasaHQ/rasa/pull/8932) but without the expected result. Not sure where things went wrong, should I target `main` instead?

ancalita: I'm actually wondering if this syntax is valid: `/rasa/**.py @RasaHQ/enable-squad` - where did you find this in the docs? All I could find is the below - which actually states the opposite (that nested files won't be targeted):
```
# The `docs/*` pattern will match files like
# `docs/getting-started.md` but not further nested files like
# `docs/build-app/troubleshooting.md`.
docs/*  docs@example.com
```
Maybe the rule I should use is `rasa/ @RasaHQ/enable-squad` which will target **all** files in this directory as instructed here:
```
# In this example, @octocat owns any file in an apps directory
# anywhere in your repository.
apps/ @octocat
```

ancalita:
> I think we should be able to test it by just creating PRs against this branch, no?

I think the reason why my test failed yesterday was also because this branch is not a [protected branch](https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-pull-request-reviews-before-merging). Probably the same will go for any `2.x` branches.
Who should I ask for this branch to be protected so I can test PRs against it?

ancalita:
Hey @wochinge,

I've discovered another source of invalid syntax which was the `!` line, I've removed this and I've double-checked the shield icon on top of a few files and it shows the correct owner as described in the rules. The only disadvantage is that the `constants.py` file that shouldn't have any owners now has Enable as owner, but negating patterns is not yet possible in CODEOWNERS, as described in the docs (I was just reading it backwards at first 😅 ).

In addition, I also created a [test PR](https://github.com/RasaHQ/rasa/pull/8940) which tagged the correct code-owner.
To avoid spamming Research, do you think this is sufficient and ready to be merged? Or should I proceed opening 2 new PRs to test this?

CODEOWNERS
```diff
@@ -1,5 +1,23 @@
-# Code in the `rasa.shared` package is potentially re-used by downstream dependencies
-# such as Rasa X. Hence, changes within this package require double checking.
-/rasa/shared/ @RasaHQ/enable-squad
 /docs/docs/prototype-an-assistant.mdx @ricwo @alwx
```
HotThoughts, line 1: `/docs/docs/prototype-an-assistant.mdx @alwx`

Should we remove Rick as he left?

ancalita: maybe I can tag the squad instead of individuals to avoid this situation when they leave?

CODEOWNERS
```diff
@@ -1,5 +1,23 @@
-# Code in the `rasa.shared` package is potentially re-used by downstream dependencies
-# such as Rasa X. Hence, changes within this package require double checking.
-/rasa/shared/ @RasaHQ/enable-squad
-/docs/docs/prototype-an-assistant.mdx @ricwo @alwx
+/docs/docs/prototype-an-assistant.mdx @RasaHQ/enable-squad
 /.github/workflows/ @RasaHQ/infrastructure
+/rasa/ @RasaHQ/enable-squad
+/rasa/core/featurizers/ @RasaHQ/research
+/rasa/core/policies/*.py @RasaHQ/research
```
wochinge, line 5: Should we do a double star to match also subdirectories in case they're created?
```suggestion
/rasa/core/policies/**.py @RasaHQ/research
```

wochinge: or we just do 
```suggestion
/rasa/core/policies/ @RasaHQ/research
```

wochinge (approved):
> but negating patterns is not yet possible in CODEOWNERS, as described in the docs (I was just reading it backwards at first 😅 ).

The same happened to me 🙈 

> The only disadvantage is that the constants.py file that shouldn't have any owners now has Enable as owner, 

I think that's fine. I think the CODEOWNERs file is way more transparent the way it is now 💯 

Thanks for doing a test PR 💯

CODEOWNERS
ancalita: Yup, the latter approach makes more sense 👍

PR #8897, opened 2021-06-16
CLI command to chat with an assistant using rasa webchat

tmbo:
**Proposed changes**:
- add a new CLI command `rasa chat` to run the assistant using SocketIO and open webchat
- added HTML page to host the webchat, keeping everything on localhost

**Example interaction**:

https://user-images.githubusercontent.com/1098412/122185610-83178780-ce8d-11eb-9755-8e99cb922705.mov

If we want to add this, we'd need to update the documentation and I'd suggest recommending using the webchat rather than `rasa shell`.

The HTML surrounding the webchat currently is very basic. We'd at least need some background/design for it. (random idea in the future @TyDunn we could use the background to feature conferences or community events - just a random thought though). We could also use the space to show more state information about the conversation, e.g. grabbing the sidepanel we use in Rasa X:
Bildschirmfoto 2021-06-16 um 10 35 51

### DIFF
--- a/rasa/cli/chat.py
+++ b/rasa/cli/chat.py
@@ -1,0 +1,69 @@
+import argparse
+from asyncio import AbstractEventLoop
+import logging
+from typing import List, Text
+import uuid
+import webbrowser
+
+from sanic import Sanic
+
+from rasa.cli import SubParsersAction
+from rasa.cli.arguments import shell as arguments
+from rasa.core import constants
+
+logger = logging.getLogger(__name__)
+
+
+def add_subparser(
+    subparsers: SubParsersAction, parents: List[argparse.ArgumentParser]
+) -> None:
+    """Add all chat parsers.
+
+    Args:
+        subparsers: subparser we are going to attach to
+        parents: Parent parsers, needed to ensure tree structure in argparse
+    """
+    chat_parser = subparsers.add_parser(
+        "chat",
+        parents=parents,
+        conflict_handler="resolve",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        help=(
+            "Loads your trained model and lets you talk to your "
+            "assistant in the browser."
+        ),
+    )
+    chat_parser.set_defaults(func=chat)
+
+    chat_parser.add_argument(
+        "--conversation-id",
+        default=uuid.uuid4().hex,
+        required=False,
+        help="Set the conversation ID.",
+    )
+
+    arguments.set_shell_arguments(chat_parser)
+    chat_parser.set_defaults(enable_api=True)
+
+
+async def open_chat_in_browser(server_url: Text) -> None:
+    """Opens the rasa chat in the default browser."""
+    webbrowser.open(f"{server_url}/webhooks/socketio/chat.html")
+
+
+def chat(args: argparse.Namespace) -> None:
+    """Chat to the bot using the most recent model."""
+    import rasa.cli.run
+
+    args.connector = "socketio"
+
+    logging.getLogger("rasa.core.tracker_store").setLevel(logging.INFO)
+
+    async def after_start_hook_open_chat(_: Sanic, __: AbstractEventLoop) -> None:
+        """Hook to open the browser on server start."""
+        server_url = constants.DEFAULT_SERVER_FORMAT.format("http", args.port)
+        await open_chat_in_browser(server_url)
+
+    args.server_listeners = [(after_start_hook_open_chat, "after_server_start")]
+
+    rasa.cli.run.run(args)
--- a/rasa/core/channels/chat.html
+++ b/rasa/core/channels/chat.html
@@ -1,0 +1,150 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <style>
+    .content {
+      max-width: 30em;
+    }
+
+    .container {
+      display: grid;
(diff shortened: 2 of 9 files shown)


tmbo:
I can add necessary tests / documentation, but we should discuss first if this is the path we want to take.

TyDunn:
@tmbo This super cool! Sorry for the delay on this one. It slipped through the cracks. Do you think we should continue to support `rasa shell` as well? If so, why?

tmbo:
I don't think we should; do you think there is a use case that is specific to a cli-only interaction?

TyDunn:
Hmm. Not really. What about users running Rasa Open Source on headless servers? For example, we are using a GCP VMs for user tests today and won't have access to a browser, so we'll use `rasa shell` (and would not be able to use `rasa chat` if it were available)

amn41:
`rasa shell --debug` is pretty helpful

tmbo:
you'll get the same information though talking through the webchat. the downside would be that it isn't "interleaved" with your interaction in the same way. (something we can fix though with better log messages)

tmbo:
> For example, we are using a GCP VMs for user tests today and won't have access to a browser, so we'll use rasa shell (and would not be able to use rasa chat if it were available)

@TyDunn that is a good point and something you can't get around without some tooling to expose the interface (e.g. ngrok OR `rasa --share` 😉 ). 

I think this is a discussion we should also have with the community as there might be other hidden use cases for `rasa shell`.

TyDunn:
> I think this is a discussion we should also have with the community as there might be other hidden use cases for rasa shell.

Yeah, I was talking to a hero last week, and he said that he heavily uses the `debug` flag with `rasa shell`. It's how he resolves issues. For example, he will use it if he wants to check if the event tracker storing things matches their stories and things like that.  He uses `rasa shell `often, although he is moving more towards using automated tests. He mentioned it would be nice if could get the debug information out of rasa tests because then he wouldn't necessarily need to use `rasa shell` so much

tmbo:
@RasaHQ/enable-squad any plans to include this? otherwise I'd go ahead and close as it is going stale

TyDunn:
I think we need to figure out a plan with `rasa shell` if we were to merge this. We don't have too many more urgent priorities right now, so I will close this

rasa/core/run.py
```diff
@@ -94,6 +94,7 @@ def configure_app(
     syslog_port: Optional[int] = None,
     syslog_protocol: Optional[Text] = None,
     request_timeout: Optional[int] = None,
+    server_listeners: Optional[List[Tuple[Callable, Text]]] = None,
```
ancalita, line 97: What are these server listeners for? Asking out of ignorance really 😄

rasa/core/channels/chat.html
```diff
@@ -0,0 +1,29 @@
...
+      data-new-gr-c-s-loaded="14.1014.0">
+<div class="logged-in-box auth0-box logged-in">
+    <h2>Happy chatting!</h2>
+    <div>You can include the chat on your own website using</div>
+    <div><code><pre>
+    &lt;div id="rasa-chat-widget" data-websocket-url="https://example.com/"&gt;&lt;/div&gt;
+&nbsp;&nbsp;&nbsp;&nbsp;&lt;script src="https://unpkg.com/@rasahq/rasa-chat" type="application/javascript"&gt;&lt;/script&gt;
+    </pre></code></div>
+</div>
+make sure to replace <code>https://example.com/</code> with the URL of your server
```
ancalita, line 13: Does this refer to the `data-websocket-url` in L9?

rasa/cli/chat.py
```diff
@@ -0,0 +1,66 @@
...
+
+def chat(args: argparse.Namespace) -> None:
+    """Chat to the bot using the most recent model."""
+    import rasa.cli.run
+
+    args.connector = "socketio"
+
+    async def after_start_hook_open_chat(_: Sanic, __: AbstractEventLoop) -> None:
+        """Hook to open the browser on server start."""
+        server_url = constants.DEFAULT_SERVER_FORMAT.format("http", args.port)
```
ancalita, line 61: How does the server_url work with the `data-websocket-url` from `chat.html` file?

ancalita (commented):
I was just checking any PRs I got assigned to and noticed this via the squad assignment 😄 
Looks good 💯 I only asked a few questions out of curiosity.

A few things to follow-up on though:
- How best do we go about testing this works fine?
- We also need to update docs here with instructions on how to set this up, do you plan adding these to this PR?
- A cli test is failing because it needs an assertion updated to contain the new command.
- There are some code quality checks that need fixing + changelog entry.

rasa/core/run.py
tmbo: they are a more idiomatic way to register hooks to be triggered after the server starts - e.g. for the chat command we want to open the browser after the server is ready.

rasa/core/channels/chat.html
tmbo: yes it does

rasa/cli/chat.py
tmbo: I am not sure what you mean?

tmbo:
@chandrikas not sure what to do with this one, I personally am using it but I am not sure what the status of the webchat UI is. I am checking in with the rest of the product team first

PR #8898, opened 2021-06-16
Connect local assistant to a static url

tmbo:
**!WARNING: THIS IS A TECHNICAL PROTOTYPE; even if implemented you shouldn't rely on this current implementation nor any generated URLs, they will be removed!**

**Proposed changes**:
- extends #8897 (`rasa chat` command) and adds `--share` which will create a public URL that can be used for others to chat with your local assistant (e.g. https://bridge.rasa.com/chat/59f40eb98c8e48a1a9)
- the website uses our new webchat to talk to the assistant
- shared URL is stable across restarts - this is the difference to `rasa chat` which uses `localhost`

**Example workflow**:

https://user-images.githubusercontent.com/1098412/122189868-962c5680-ce91-11eb-8dc7-58de7d924527.mov

I am not sure how useful the sharable webchat is: I think the lacking piece is the ability to view conversations others have with your assistant. I think the static URL functionality is a lot more useful to connect your local assistant to e.g. slack / messenger / ... any external connector requiring a secured (https) and public URL (non `localhost`).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -200,7 +200,7 @@ wrapt = "*"
 
 [[package]]
 name = "azure-core"
-version = "1.14.0"
+version = "1.15.0"
 description = "Microsoft Azure Core Library for Python"
 category = "dev"
 optional = false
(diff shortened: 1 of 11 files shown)


tmbo:
@TyDunn any plans to include an ngrok replacement? I think it enables us to connect channels a lot easier (as we could potentially do that in a UI at rasa.com). but given that the team has a lot on its plate happy to close this one.

TyDunn:
@tmbo Not at this point. I really want to get to this, but it's too far from priorities atm

Issue #8700, opened 2021-05-17
Verify whether bug fix by tensorflow unblocks #7619


koernerfelicia:
**Description of Problem**:
Our upgrade to TF 2.4 was blocked by the bug described [here](https://github.com/tensorflow/tensorflow/issues/46511). According to TF, this bug has been fixed in 2.5 (see their comment [here](https://github.com/tensorflow/tensorflow/issues/46511#issuecomment-841408480)).

Unblocking the above should bring us closer to being able to support users with M1 chips.

**Overview of the Solution**:
We should check whether this bug has actually been fixed for the purposes of our upgrade. If the bug hasn't been fixed, we need to notify TF. If it has, we need to move [this issue](https://github.com/RasaHQ/rasa/issues/7619) out of the "Blocked" column.

**Definition of Done**:
- [x] Verify if the bug has been fixed or not
- [x] If it has been fixed, then move [this issue](https://github.com/RasaHQ/rasa/issues/7619) out of "Blocked". If it has not been fixed, then update the bug report [here](https://github.com/tensorflow/tensorflow/issues/46511) with evidence about how the bug *isn't fixed*.

mpomarole:
2.5 is now out!

https://developer.apple.com/metal/tensorflow-plugin/

https://github.com/tensorflow/tensorflow/releases/tag/v2.5.0

Seems like this can move forward?

koernerfelicia:
Update:
Not sure yet whether the bug is truly fixed for our purposes. The gist does indeed run without problems. The bug would affect any component or policy that uses our `CRF` layer, namely `DIET` with entity extraction and TED with e2e and entity extraction. `formbot` (which has DIET and entity extraction) did not train with `TF=2.4` (see branch [here](https://github.com/RasaHQ/rasa/tree/tf-2.4-iss-8700)), but *does* train with `TF=2.5` (see draft PR [here](https://github.com/RasaHQ/rasa/pull/8899)).

_However_ tests that involve exactly this scenario (DIET + EE and TED with e2e + EE) are flakey due to timeouts. I've observed this locally and on CI.

Will keep investigating.

koernerfelicia:
I can say it is fixed in the sense that the code no longer hangs, but I'm not sure if it is a complete fix, since there seem to be some performance issues which look like they're related to the original bug, simply because they only occur in the same scenario in which the original bug occurred (anytime we are using our custom CRF layer).

[...]

PR #8899, opened 2021-06-16
Update the code for tf 2.5

koernerfelicia:
DRAFT PR FOR TESTING


koernerfelicia:
Note this is a draft for https://github.com/RasaHQ/rasa/issues/8700. Known remaining tasks in order to actually update code for tf 2.5 are:
- [ ] see TODOs marked with 2.5
- [ ] investigate failed memory leak tests
- [ ] remove keras-nightly from lockfile and figure out appropriate substitute

TyDunn:
@koernerfelicia It sounds like we were able to verify the bug was fixed?

koernerfelicia:
@TyDunn not much has changed from my update here: https://github.com/RasaHQ/rasa/issues/8700#issuecomment-863405940

I guess I can say it is fixed in the sense that the code no longer hangs, but I'm not sure if it is a complete fix, since there seem to be some performance issues which *look* like they're related to the original bug, simply because they only occur in the same scenario in which the original bug occurred (anytime we are using our custom CRF layer).

How would you suggest I handle this? I'm thinking I'll update the original issue with the TODO's I wrote above, and add a note about this flakiness?

TyDunn:
> How would you suggest I handle this? I'm thinking I'll update the original issue with the TODO's I wrote above, and add a note about this flakiness?

@koernerfelicia  That seems reasonable to me. And then throw it in the Research inbox?

koernerfelicia:
@TyDunn, done! I updated the issue description to include the list of todos and added a note about the somewhat fishy behaviour around our custom CRF layer.

samsucik:
/modeltest

```yml
include:
 - dataset: ["all"]
   config: ["all"]

PR #8903, opened 2021-06-16
Switch back to upstream lazy-actions/gitrivy

virtualroot:
**Proposed changes**:

lazy-actions/gitrivy#62 was merged. We can switch back to upstream.


tczekajlo (approved):
LGTM 🎖️

PR #8906, opened 2021-06-17
`IntentTEDPolicy`: `rasa interactive` updates

### LINKED ISSUE #8421: IntentTEDPolicy: Adapt `rasa interactive`
**Description of Problem**:
When there is an unlikely intent, `IntentTEDPolicy` predicts an `action_unlikely_intent` to indicate the unlikely intent. This could be confusing to the user as they don't know where this action is coming from.

**Overview of the Solution**:
When the model predicted this action as part of `rasa interactive` we want to give them additional explanation why this action was predicted. Users can then choose to confirm / deny the action prediction as usual but it doesn't really matter as running this action doesn't have any implications.

**Blockers** (if relevant):
- [x] Implemented `IntentTEDPolicy`
- [x] https://github.com/RasaHQ/rasa/issues/8417

**Definition of Done**:
- [ ] Interactive displays the explanation for the prediction of `action_unlikely_intent` to the user.
- [ ] somebody from CSE double checked the UX of the explanation

alwx:
**Proposed changes**:
- Adapts `rasa interactive` to `IntentTEDPolicy` (`action_unlikely_intent`)
- fixes #8421 

**It requires `rasa test` updates to be merged first! https://github.com/RasaHQ/rasa/pull/8843**

### DIFF
--- a/rasa/core/training/interactive.py
+++ b/rasa/core/training/interactive.py
@@ -1132,16 +1132,20 @@ async def _validate_action(
     Returns `True` if the prediction is correct, `False` otherwise."""
 
     if action_name == ACTION_UNLIKELY_INTENT_NAME:
-        print(f"The bot wants to run '{action_name}' "
-              f"(https://rasa.com/docs/rasa/default-actions#action_unlikely_intent) "
-              f"to indicate that this user message was unexpected "
-              f"given the context of the conversation. "
-              f"Check out IntentTEDPolicy to learn more. "
-              f"Press any key to continue…")
+        print(
+            f"The bot wants to run '{action_name}' "
+            f"(https://rasa.com/docs/rasa/default-actions#action_unlikely_intent) "
+            f"to indicate that this user message was unexpected "
+            f"given the context of the conversation. "
+            f"Check out IntentTEDPolicy to learn more. "
+            f"Press any key to continue…"
+        )
         input()
         is_correct = True
     else:
-        question = questionary.confirm(f"The bot wants to run '{action_name}', correct?")
+        question = questionary.confirm(
+            f"The bot wants to run '{action_name}', correct?"
+        )
         is_correct = await _ask_questions(question, conversation_id, endpoint)
 
     if not is_correct:
--- a/rasa/core/training/interactive.py
+++ b/rasa/core/training/interactive.py
@@ -1131,7 +1131,12 @@ async def _validate_action(
     Returns `True` if the prediction is correct, `False` otherwise."""
 
     if action_name == "action_unlikely_intent":
-        print(f"The bot wants to run '{action_name}'. Press any key to continue...")
+        print(f"The bot wants to run '{action_name}' "
+              f"(https://rasa.com/docs/rasa/default-actions#action_unlikely_intent) "
+              f"to indicate that this user message was unexpected "
+              f"given the context of the conversation. "
+              f"Check out IntentTEDPolicy to learn more. "
+              f"Press any key to continue…")
         input()
         is_correct = True
     else:
--- a/rasa/core/training/interactive.py
+++ b/rasa/core/training/interactive.py
@@ -65,7 +65,7 @@ from rasa.shared.constants import (
     INTENT_MESSAGE_PREFIX,
     DEFAULT_SENDER_ID,
     UTTER_PREFIX,
-    DOCS_URL_POLICIES
+    DOCS_URL_POLICIES,
 )
 from rasa.shared.core.trackers import EventVerbosity, DialogueStateTracker
 from rasa.shared.core.training_data import visualization
@@ -1137,20 +1137,19 @@ async def _validate_action(
     Returns `True` if the prediction is correct, `False` otherwise."""
 
     if action_name == ACTION_UNLIKELY_INTENT_NAME:
-        print(
+        question = questionary.confirm(
             f"The bot wants to run '{action_name}' "
             f"to indicate that the last user message was unexpected "
             f"at this point in the conversation. "
             f"Check out IntentTEDPolicy ({DOCS_URL_POLICIES}/#intent-ted-policy) "
             f"to learn more. Press any key to continue…"
         )
-        input()
-        is_correct = True
     else:
         question = questionary.confirm(
             f"The bot wants to run '{action_name}', correct?"
         )
-        is_correct = await _ask_questions(question, conversation_id, endpoint)
+
+    is_correct = await _ask_questions(question, conversation_id, endpoint)
 
     if not is_correct:
         action_name, is_new_action = await _request_action_from_user(
--- a/rasa/core/training/interactive.py
+++ b/rasa/core/training/interactive.py
@@ -65,7 +65,6 @@ from rasa.shared.constants import (
     INTENT_MESSAGE_PREFIX,
     DEFAULT_SENDER_ID,
     UTTER_PREFIX,
-    DOCS_URL_DEFAULT_ACTIONS,
     DOCS_URL_POLICIES
 )
 from rasa.shared.core.trackers import EventVerbosity, DialogueStateTracker
@@ -1140,9 +1139,8 @@ async def _validate_action(
     if action_name == ACTION_UNLIKELY_INTENT_NAME:
         print(
             f"The bot wants to run '{action_name}' "
-            f"({DOCS_URL_DEFAULT_ACTIONS}/#action_unlikely_intent) "
-            f"to indicate that this user message was unexpected "
-            f"given the context of the conversation. "
+            f"to indicate that the last user message was unexpected "
+            f"at this point in the conversation. "
             f"Check out IntentTEDPolicy ({DOCS_URL_POLICIES}/#intent-ted-policy) "
             f"to learn more. Press any key to continue…"
         )
--- a/rasa/core/training/interactive.py
+++ b/rasa/core/training/interactive.py
@@ -1142,14 +1142,17 @@ async def _validate_action(
             f"to indicate that the last user message was unexpected "
             f"at this point in the conversation. "
             f"Check out IntentTEDPolicy ({DOCS_URL_POLICIES}/#intent-ted-policy) "
-            f"to learn more. Press any key to continue…"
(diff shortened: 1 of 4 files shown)


alwx:
There are literally no tests for `rasa interactive` and `_validate_action` function that's why nothing was added here as well. Probably the right action would be to add an issue to add tests for `rasa interactive`.

rasa/core/training/interactive.py
```diff
@@ -1130,9 +1131,22 @@ async def _validate_action(
...
     Returns `True` if the prediction is correct, `False` otherwise."""
 
-    question = questionary.confirm(f"The bot wants to run '{action_name}', correct?")
-
-    is_correct = await _ask_questions(question, conversation_id, endpoint)
+    if action_name == ACTION_UNLIKELY_INTENT_NAME:
+        print(
+            f"The bot wants to run '{action_name}' "
+            f"(https://rasa.com/docs/rasa/default-actions#action_unlikely_intent) "
+            f"to indicate that this user message was unexpected "
```
ancalita, line 1138: Should this be rephrased to `the last user message was unexpected`?

rasa/core/training/interactive.py
```diff
@@ -1130,9 +1131,22 @@ async def _validate_action(
...
+            f"given the context of the conversation. "
+            f"Check out IntentTEDPolicy to learn more. "
```
ancalita, line 1140: Should we link to the policy docs here?

rasa/core/training/interactive.py
```diff
@@ -1130,9 +1131,22 @@ async def _validate_action(
...
+            f"Press any key to continue…"
+        )
+        input()
+        is_correct = True
```
ancalita, line 1144: I think this part of the GH solution is not implemented:`Users can then choose to confirm / deny the action prediction`, is that something that was decided after the issue submission?

rasa/core/training/interactive.py
```diff
@@ -1130,9 +1131,22 @@ async def _validate_action(
 
     Returns `True` if the prediction is correct, `False` otherwise."""
 
-    question = questionary.confirm(f"The bot wants to run '{action_name}', correct?")
-
-    is_correct = await _ask_questions(question, conversation_id, endpoint)
+    if action_name == ACTION_UNLIKELY_INTENT_NAME:
+        print(
+            f"The bot wants to run '{action_name}' "
+            f"(https://rasa.com/docs/rasa/default-actions#action_unlikely_intent) "
```
wochinge, line 1137:
```suggestion
            f"{DOCS_URL_DEFAULT_ACTIONS}#action_unlikely_intent) "
```

rasa/core/training/interactive.py
wochinge: I'd also explain what it means that a message is unexpected, e.g. `the last user message was unexpected at this point in the conversation`. How about asking somebody from CSE for review?

rasa/core/training/interactive.py
```diff
@@ -1130,9 +1131,22 @@ async def _validate_action(
...
+            f"to indicate that this user message was unexpected "
+            f"given the context of the conversation. "
+            f"Check out IntentTEDPolicy to learn more. "
+            f"Press any key to continue…"
+        )
+        input()
```
wochinge, line 1143: how about using `questionary` here?

wochinge (commented):
I know it's a hard to test module but could we please try to add a test for the new edge case?

rasa/core/training/interactive.py
alwx: From the description: "Users can then choose to confirm / deny the action prediction as usual but it doesn't really matter as running this action doesn't have any implications."
@wochinge  So my question is: if the answer doesn't matter, why do we need to ask a question?

rasa/core/training/interactive.py
alwx: The original message was written by @TyDunn so it's more like a question to you.

rasa/core/training/interactive.py
alwx: Added them.

rasa/core/training/interactive.py
alwx: It's related to my question above.

rasa/core/training/interactive.py
TyDunn: `the last user message was unexpected at this point in the conversation` is an improvement. We discussed with CSE here: https://rasa-hq.slack.com/archives/C36SS4N8M/p1623402780458000

rasa/core/training/interactive.py
wochinge: I see your point. I'd asking the question makes sense as
* aligns with behavior for other predicted actions
* makes uses explicitly think about what happened. If it's just a print statement they might just skim over it without actually considering the implications.

wochinge (commented):
Can we please add a test for it, Alex?

tests/core/training/test_interactive.py
```diff
@@ -733,6 +737,51 @@ def test_retry_on_error_success(monkeypatch: MonkeyPatch):
...
+    )
+    monkeypatch.setattr(
+        interactive, "_ask_questions", asyncio.coroutine(Mock(return_value=True))
+    )
+    monkeypatch.setattr(interactive, "_form_is_rejected", Mock(return_value=False))
+    monkeypatch.setattr(interactive, "_form_is_restored", Mock(return_value=False))
+    monkeypatch.setattr(interactive, "send_action", asyncio.coroutine(Mock()))
+
+    mocked_confirm = Mock(return_value=None)
+    monkeypatch.setattr(interactive.questionary, "confirm", mocked_confirm)
```
wochinge, line 776: how about testing this with a `True` and `False` value to cover both paths?

tests/core/training/test_interactive.py
```diff
@@ -733,6 +737,51 @@ def test_retry_on_error_success(monkeypatch: MonkeyPatch):
...
+    monkeypatch: MonkeyPatch,
+    mock_endpoint: EndpointConfig,
+    action_name: Text,
+    question: Text,
+):
+    # mock those functions that shouldn't be called
+    monkeypatch.setattr(
+        interactive,
+        "retrieve_tracker",
+        asyncio.coroutine(
```
wochinge, line 764: asyncio.coroutine is deprecated as of Python 3.8 - can we please us `async def` instead?

tests/core/training/test_interactive.py
```diff
@@ -733,6 +737,51 @@ def test_retry_on_error_success(monkeypatch: MonkeyPatch):
...
+            Mock(return_value=DialogueStateTracker.from_events("one", []))
+        ),
+    )
+    monkeypatch.setattr(
+        interactive, "_ask_questions", asyncio.coroutine(Mock(return_value=True))
+    )
+    monkeypatch.setattr(interactive, "_form_is_rejected", Mock(return_value=False))
+    monkeypatch.setattr(interactive, "_form_is_restored", Mock(return_value=False))
```
wochinge, line 772: why do we need to mock these two? If we provide a good mock tracker then this should just give us the desired results, no?

tests/core/training/test_interactive.py
```diff
@@ -733,6 +737,51 @@ def test_retry_on_error_success(monkeypatch: MonkeyPatch):
...
+    monkeypatch.setattr(interactive, "send_action", asyncio.coroutine(Mock()))
```
wochinge, line 773: how about making sure this gets called with the expected values?

alwx: Done

rasa/core/training/interactive.py
```diff
@@ -1130,9 +1136,23 @@ async def _validate_action(
...
     Returns `True` if the prediction is correct, `False` otherwise."""
 
-    question = questionary.confirm(f"The bot wants to run '{action_name}', correct?")
+    if action_name == ACTION_UNLIKELY_INTENT_NAME:
+        question = questionary.confirm(
+            f"The bot wants to run '{action_name}' "
+            f"to indicate that the last user message was unexpected "
+            f"at this point in the conversation. "
+            f"Check out IntentTEDPolicy ({DOCS_URL_POLICIES}/#intent-ted-policy) "
+            f"to learn more."
```
ancalita, line 1145: Do we need to ask the user here to press Y/N to continue?

alwx: That's what we've decided to do, yes (to keep it consistent with the rest)

rasa/core/training/interactive.py
```diff
@@ -1130,9 +1136,23 @@ async def _validate_action(
...
+        )
+    else:
+        question = questionary.confirm(
+            f"The bot wants to run '{action_name}', correct?"
+        )
 
-    is_correct = await _ask_questions(question, conversation_id, endpoint)
+    is_correct = (
+        await _ask_questions(question, conversation_id, endpoint)
+        or action_name == ACTION_UNLIKELY_INTENT_NAME
```
ancalita, line 1154: Does this line imply that regardless of user answer to the query, the action is always correct for `action_unlikely_intent`?

alwx: It means that the choice (Y/N) doesn't affect what's gonna happen next.

rasa/core/training/interactive.py
ancalita: Ah ok, then maybe it should be explicitly stated? Unless `questionary` takes care of this 😄

wochinge (approved):
Thanks for adding the test! 💯

rasa/core/training/interactive.py
alwx: IMO this should instead be written in docs with the explanation what `action_unlikely_intent` is and how it works. We can also write it here but I am afraid the message will be too long in this case.

ancalita: Ah I meant maybe it's not obvious what keys to press from the question and that should be explicitly stated, thats all 👍

alwx: Oh, got it! Actually, `questionary` handles that and shows the available options automatically :)

PR #8908, opened 2021-06-17
UnexpecTEDIntentPolicy: Docs

dakshvar22:
**Proposed changes**:
- Docs for `UnexpecTEDIntentPolicy`
- **Note:** The introduction of a new default action will make the overall release breaking backwards compatibility of models.
This has been communicated in the changelog and migration guide.


docs/docs/policies.mdx
```diff
@@ -396,6 +398,268 @@ for details.
...
+To reduce false warnings, `IntentTEDPolicy` has two mechanisms in place at inference time:
+
+1. `IntentTEDPolicy`'s [priority](./policies.mdx#policy-priority) is intentionally kept lower than all
+[rule based policies](./policies.mdx#rule-based-policies). This ensures that if an `action_unlikely_intent`
+is predicted for the last intent, but there is a rule that can be triggered to predict
+the next action, the ensemble of policies is not affected by it and the winning prediction is
+still the one made by `RulePolicy`.
+
+2. `IntentTEDPolicy` does not predict an `action_unlikely_intent` if the last predicted intent
+isn't present in any of the training stories.
```
kedz, line 463: " which might happen if an intent is only used in rules."

docs/docs/policies.mdx
```diff
@@ -396,6 +398,268 @@ for details.
...
+whereas there can be multiple intents that are likely to occur in the same conversation context. Hence,
+`IntentTEDPolicy` uses a multi-label cross entropy loss function during training.
+
+#### Prediction of `action_unlikely_intent`
+
+After training the internal ML model, `IntentTEDPolicy` does the following for each intent:
+
+1. Collect conversation contexts from training stories for which the corresponding intent is not the correct intent.
+2. Run these conversation contexts through the trained ML model and capture the predicted scores for the corresponding intent.
+3. Compute multiple quantiles
```
kedz, line 476: What about this:


#### Prediction of `action_unlikely_intent`

`IntentTEDPolicy` is invoked immediately after a user utterance and can either predict `action_unlikely_intent` or abstain (in which case other policies will predict actions). To determine if `action_unlikely_intent` should be predicted, `IntentTEDPolicy` computes the confidence score of the intent of the actual user utterance using its internal intent classifier. If the confidence is below a threshold, `IntentTEDPolicy` predicts `action_unlikely_intent`, otherwise it abstains.

Each intent has a unique threshold which is determined at training time after `IntentTEDPolicy`'s internal intent classifier is trained. A synthetic dataset of dialogue contexts paired with unlikely intents is created from your assistant's original training stories. To create the threshold for a specific intent, scores for all examples in the synthetic dataset that have that intent as its label are computed using `IntentTEDPolicy`'s internal intent classifier 
and the threshold is set to be the maximum of these confidence scores. In its default setting, `IntentTEDPolicy` would predict an `action_unlikely_intent` for any of the examples in the synthetic dataset.

If the `IntentTEDPolicy` is too sensitive (predicting too many `action_unlikely_intent`s) its tolerance parameter can be raised to decrease the chances of predicting `action_unlikely_intent`. The tolerance parameter defaults to `0.0` and ranges between `0.0` and `1.0` inclusive. At a tolerance of `0.25`, `IntentTEDPolicy` would not predict an `action_unlikely_intent` on 25% of the examples from the synthetic dataset. When the tolerance is set to `1.0`, 100% of the examples would not trigger an `action_unlikely_intent`.

docs/docs/conversation-driven-development.mdx
```diff
@@ -59,6 +59,20 @@ Your test users had at least some instruction about what the bot was intended to
...
 *   Look at conversations where an “out_of_scope” intent or fallback behavior occurred. These could indicate a potential new skill, or just a misclassified user utterance.
 *   Look for user frustration, such as requests for transfer to a human.
+*   Look for conversations flagged as "surprising" conversations with the
+    help of [`IntentTEDPolicy`](./policies.mdx#intenttedpolicy)'s output: This can be done by running a [standalone
+    script]. The script does the following:
+    - Fetch real conversations from a tracker store.
+    - Run `rasa test` on the fetched conversations and write conversations for which `action_unlikely_intent`
+        is triggered in a separate warnings file.
+
+    Reviewing these subset of conversations can help understand if real users have taken a
```
kedz, line 69: "Reviewing this subset of conversations can help you understand"

docs/docs/conversation-driven-development.mdx
```diff
@@ -59,6 +59,20 @@ Your test users had at least some instruction about what the bot was intended to
...
+    conversation path which is not present in the training data and hence "surprising"
+    for machine learning policies like `TEDPolicy`. Adding these conversation paths back to
```
kedz, line 71: "Adding these conversation paths (with potential corrections if TEDPolicy subsequently failed)..."

docs/docs/policies.mdx
kedz: Yeah I really like this! My only other suggestion would be to change the middle paragraph slightly:

This threshold score is learned at training time but can be adjusted at test time via the `tolerance` parameter.  The higher the `tolerance`, the lower the intent's score (the more unlikely the intent) needs to be before IntentTED triggers the `action_unlikely_intent`.

docs/docs/policies.mdx
```diff
@@ -396,6 +398,268 @@ for details.
...
+  ```
+
+* `max_history`:
+  This parameter controls how much dialogue history the model looks at to decide which
+  action to take next. Default `max_history` for this policy is `None`,
+  which means that the complete dialogue history since session restart is taken into
+  account. If you want to limit the model to only see a certain number of previous
+  dialogue turns, you can set `max_history` to a finite value.
+  Please note that you should pick `max_history` carefully, so that the model has enough
+  previous dialogue turns to create a correct prediction.
```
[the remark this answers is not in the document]
kedz: Should we also give the recommendation to set it to the same value as TED since we haven't really explored using them together with different settings?

docs/docs/policies.mdx
```diff
@@ -396,6 +398,268 @@ for details.
...
+  - user: |
+        I'm looking for a restaurant.
+    intent: request_restaurant
+  - action: restaurant_form
+  - active_loop: restaurant_form
+  - slot_was_set:
+    - requested_slot: cuisine
+  - user: |
+        Does it matter? I want to be quick.
+    intent: deny
```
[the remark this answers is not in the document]
dakshvar22: I am quite sure that's not the case. I actually took an example very similar to our [docs for handling contextual interjections]((./default-actions.mdx#action_unlikely_intent))

docs/docs/policies.mdx
dakshvar22: Thank you both! These look really good 🚀

docs/docs/policies.mdx
```diff
@@ -396,6 +398,268 @@ for details.
...
+
+#### Tuning the tolerance parameter
+
+Users are also encouraged to tune the `tolerance` parameter in `IntentTEDPolicy`'s configuration to reduce the number
+of false warnings. As you increase the value of `tolerance` from `0` to `1` in steps of `0.05`,
+the number of false warnings should decrease. However, increasing the `tolerance` will
+also result in fewer triggers of `action_unlikely_intent` and hence more number of conversation
+paths not present in training stories missing from the set of flagged conversations. This is particularly helpful
+when you are [reviewing real conversations](./conversation-driven-development.mdx#review) that
+your assistant has processed.
```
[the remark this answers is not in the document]
dakshvar22: I am not sure what parts of it we can add here. Are those scripts shareable?
Wdyt @kedz

docs/docs/policies.mdx
dakshvar22: I used the first two paragraphs for this section and moved the 3rd paragraph to the description of `tolerance` parameter in the `Configuration` section connecting the two.

docs/docs/policies.mdx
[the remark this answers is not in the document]
kedz: I don't think it would be trivial to share those scripts at this point. 
Beyond cleaning up/better documenting the code, there are some pretty significant hacks I had to do:
1. Make an environment variable that adjusted the tolerance parameter at test time
2. Disable the check to not return an action_unlikely_intent if the user intent was the most likely intent predicted by IntentTED (even though its score was below the threshold).
3. The scripts assume that `utter_unlikely_intent` warnings are in the `failed_test_stories.yml` and would need to be changed to look in the warnings file.

The third issue is trivial, but 1 and 2 seem more complicated to change, but perhaps we can discuss today?

docs/docs/policies.mdx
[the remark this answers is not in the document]
dakshvar22: @kedz Not sure about it. What were the values that you used in your experiments to validated intentTED as an error detector of TED?

docs/docs/default-actions.mdx
```diff
@@ -187,3 +187,39 @@ Overriding the default action for forms should **only** be used during the proce
...
+    def name(self) -> Text:
+        return "action_unlikely_intent"
+
+    async def run(
+        self, dispatcher, tracker: Tracker, domain: Dict[Text, Any],
+    ) -> List[Dict[Text, Any]]:
+
+        # Implement custom logic here
+        return []
+```
```
dakshvar22, line 225: @JEM-Mosig You had tested this right?

docs/docs/policies.mdx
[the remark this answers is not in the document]
dakshvar22: Incorporated @JEM-Mosig 's suggestion.

docs/docs/conversation-driven-development.mdx
```diff
@@ -59,6 +59,23 @@ Your test users had at least some instruction about what the bot was intended to
...
+    - Run `rasa test` on the fetched conversations and filter conversations containing `action_unlikely_intent`
+      in a separate warnings file. Read more on [how to interpret these warnings](./testing-your-assistant.mdx#interpreting-the-generated-warnings).
+
+    Reviewing this subset of conversations can help you understand if real users have taken a
+    conversation path which is not present in the training data and hence "surprising"
+    for machine learning policies like `TEDPolicy`. Adding these conversation paths (with potential
+    corrections if `TEDPolicy` subsequently failed) as training stories will help in more robust action prediction
+    by policies such as `TEDPolicy`. Users are encouraged to [adjust the `tolerance` parameter of
+    `IntentTEDPolicy`](./policies.mdx#tuning-the-tolerance-parameter) to control how
+    severely "surprising" a conversation should be to be included in the warnings file.
```
kedz, line 78: I would drop severely in 'severely "surprising"'

docs/docs/testing-your-assistant.mdx
```diff
@@ -377,6 +377,26 @@ The test script will also save a confusion matrix to a file called
...
+a failed story.
+
+The stories are sorted by the severity of `action_unlikely_intent`'s prediction.
+This severity is calculated by `IntentTEDPolicy` itself at prediction time.
+Higher the severity, the more unlikely is the intent and hence reviewing that particular
+conversation path becomes more critical.
+
+It should be noted that `action_unlikely_intent` is predicted by
+[`IntentTEDPolicy`](./policies.mdx#intenttedpolicy) which employs a machine learning based model
+under the hood and hence can result in multiple false warnings as well. Users are encouraged to go through the
```
kedz, line 396: remove "multiple"

docs/docs/testing-your-assistant.mdx
```diff
@@ -377,6 +377,26 @@ The test script will also save a confusion matrix to a file called
...
+stories in the warnings file and verify if the conversation path is already present in one of the
+training stories and add it as a training story if it isn't.
+
```
kedz, line 399: This is somewhat confusing to me, because it sounds like you are saying to copy your test stories into training stories.

docs/docs/testing-your-assistant.mdx
```diff
@@ -377,6 +377,26 @@ The test script will also save a confusion matrix to a file called
...
 incorrect action was predicted instead.
 
+### Interpreting the generated warnings
+
+The test script will also generate a warnings file called `results/stories_with_warnings.yml`.
+This file contains all test stories for which [`action_unlikely_intent`](./default-actions.mdx#action_unlikely_intent)
+was predicted at any conversation turn but all actions from the original story were predicted correctly.
+However, if a test story originally consisted of `action_unlikely_intent` but wasn't predicted by
+the ensemble of policies, then the corresponding story will end up in `results/failed_test_stories.yml` as
+a failed story.
```
kedz, line 387: This is a little confusingly worded. What about:
However, if a test story originally included an `action_unlikely_intent` ([as part of a rule](link to customizing unlikely intents) for example) but wasn't predicted by
the `IntentTEDPolicy`, then the corresponding story will end up in `results/failed_test_stories.yml` as
a failed story.

Also, I changed ensemble of policies to IntentTEDPolicy above since only this one can predict action unlikely intent right?

docs/docs/policies.mdx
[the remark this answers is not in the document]
kedz: @dakshvar22 both TED and IntentTED max history was set to 6 for both SuperSTARBot and Sara.

docs/docs/policies.mdx
```diff
@@ -396,6 +398,307 @@ for details.
...
+  - action: restaurant_form
+  - active_loop: restaurant_form
+  - slot_was_set:
+    - requested_slot: cuisine
+  - user: |
+        Does it matter? I want to be quick.
+    intent: deny
+```
+
+As soon as `deny` intent gets triggered, the policy handling the form will keep requesting for `cuisine` slot
```
kedz, line 452: "As soon as the `deny` ...

docs/docs/policies.mdx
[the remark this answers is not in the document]
kedz: I see the logic behind @JEM-Mosig 's resasoning but also think the exact behavior in practice is likely to be a complicated interaction between the stories you are running intentTED on, the overall contextual dependence of your training data, the number of unique intents and their coverage in the training stories, etc, so I would be cautious about describing the behavior without doing some empirical exploration. 

An argument I see for matching histories of TED and IntentTED is that they will be making their decisions on the same history and so there wont be any cases where IntentTED fired spuriously because there was something extra it saw or something it didn't see that TED saw, but TED is operating fine. (also if we really want to commit to this, then maybe IntentTED max istory should be TED max history - 1)

docs/docs/policies.mdx
```diff
@@ -396,6 +398,307 @@ for details.
...
++---------------------------------------+------------------------+--------------------------------------------------------------+
+```
+
+</details>
+
+#### Tuning the tolerance parameter
+
+When [reviewing real conversations](./conversation-driven-development.mdx#review), users are
+ encouraged to tune the `tolerance` parameter in `IntentTEDPolicy`'s configuration to reduce the number
+of false warnings. As you increase the value of `tolerance` from `0` to `1` in steps of `0.05`,
```
kedz, line 692: Should probably define what a false warning is. How about:
false warnings (intents that actually are likely given the context) ...

docs/docs/policies.mdx
```diff
@@ -396,6 +398,307 @@ for details.
...
+the number of false warnings should decrease. However, increasing the `tolerance` will
+also result in fewer triggers of `action_unlikely_intent` and hence more number of conversation
```
kedz, line 694: and hence more conversation paths ...

docs/docs/policies.mdx
```diff
@@ -689,36 +992,59 @@ default restart, label `2` a greeting and `3` indicates goodbye.
...
-represented as index in a list of all possible actions.
+A policy can be trained to learn two kinds of labels -
+
+1. The next most appropriate action to be triggered by the assistant. For example, [`TEDPolicy`](#ted-policy) is trained to do this.
+2. The next most likely intent that a user can express. For example, [`IntentTEDPolicy`](#intenttedpolicy) is trained to do this.
+
+Hence, a tracker can be featurized to learn one of the above kind of labels.
+Depending on the policy, the target labels correspond to bot actions or bot utterances
+represented as index in a list of all possible actions or
+set of intents represented as index in a list of all possible intents.
```
kedz, line 1003: as an index

kedz (changes requested):
Added some grammar/wording changes as well as some areas where I was confused by the wording.

docs/docs/default-actions.mdx
JEM-Mosig: Yes, it did work with the super airline bot. But I had to add the action name to the domain.

docs/docs/testing-your-assistant.mdx
dakshvar22: Hmm, that's a good point. What do you think about this - 

"It should be noted that `action_unlikely_intent` is predicted by
[`IntentTEDPolicy`](./policies.mdx#intenttedpolicy) which employs a machine learning based model
under the hood and hence can result in false warnings as well. You can choose to ignore such warnings if the conversation paths in these stories are already present in the training stories."

docs/docs/testing-your-assistant.mdx
dakshvar22: Not sure I completely agree here - 

> However, if a test story originally included an action_unlikely_intent ([as part of a rule]

Whether it's part of a rule or not does not matter because this is a story and rule snippets are not part of a story.

> Also, I changed ensemble of policies to IntentTEDPolicy above since only this one can predict action unlikely intent right?

I think it should be ensemble of policies (or dialogue manager) because it does not matter whether a particular policy predicted it or not. The test story is evaluated against what the ensemble predicts. For example, it can happen that the ensemble didn't choose `action_unlikely_intent` because a rule was also matching for some reason and `RulePolicy` predicted an alternate action.

dakshvar22:
@kedz I incorporated all the grammatical errors you pointed out.

docs/docs/testing-your-assistant.mdx
kedz: I see both points. Re: ensemble vs IntentTEDPolicy, your explanation makes sense and I think keeping the ensemble of policies is fine. 

Regarding (as part of a rule..), this can be deleted but I think what I am worried about is it may not be clear at this point why you would include an `action_unlikely_intent` in a test story, and there could be some brief motivation or pointer to motivation about why you would do that. What I was hoping to draw reference to with the mention of the rule was the scenario where you included a rule within a larger test story that tested both that the `action_unlikely_intent` got triggered where you expected and then the corresponding rule got triggered.

docs/docs/testing-your-assistant.mdx
kedz: Looks good!

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+ `IntentTEDPolicy` can trigger an `action_unlikely_intent` action right after `deny` intent.
+ Subsequently, you can improve your assistant by adding a new training story that handles this particular case.
+
+To reduce false warnings, `IntentTEDPolicy` has two mechanisms in place at inference time:
+
+1. `IntentTEDPolicy`'s [priority](./policies.mdx#policy-priority) is intentionally kept lower than all
+[rule based policies](./policies.mdx#rule-based-policies). This ensures that if an `action_unlikely_intent`
+is predicted for the last intent, but there is a rule that can be triggered to predict
+the next action, the ensemble of policies is not affected by it and the winning prediction is
+still the one made by `RulePolicy`.
```
JEM-Mosig, line 464: I'd just remove this sentence, or say "since rules may exist for situations that are novel for TED or IntentTED".

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+
+2. `IntentTEDPolicy` does not predict an `action_unlikely_intent` if the last predicted intent
+isn't present in any of the training stories, which might happen if an intent is only used in rules.
+
+As mentioned earlier, while `TEDPolicy` learns the next best action, `IntentTEDPolicy` learns the
+next likely intent that a user can express. Given a conversation context, there can only be a single correct action,
+whereas there can be multiple intents that are likely to occur in the same conversation context. Hence,
+`IntentTEDPolicy` uses a multi-label cross entropy loss function during training. Moreover, it uses
+the [`IntentMaxHistoryTrackerFeaturizer`](#3-intent-max-history) to treat intents as target labels instead of assistant actions or utterances.
```
JEM-Mosig, line 473: Do we need to say this in the docs?

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+via the `tolerance` parameter.  The higher the `tolerance`, the lower the intent's
+score (the more unlikely the intent) needs to be before `IntentTEDPolicy` triggers
+the `action_unlikely_intent` action.
+
+**Configuration:**
+
+You can pass configuration parameters to the `IntentTEDPolicy` using the `config.yml` file.
+If you want to fine-tune model's performance, start by modifying the following parameters:
+
+* `epochs`:
```
JEM-Mosig, line 493: I wonder if one can finetune the model with zero epochs and thereby adjust the tolerance parameter :D

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+  Here is how the config would look like:
+
+  ```yaml-rasa title="config.yml"
+  policies:
+  - name: IntentTEDPolicy
+    max_history: 8
+  ```
+
+* `ignore_intents_list`:
+    This parameter lets you configure `IntentTEDPolicy` to ignore the prediction of `action_unlikely_intent` for
```
JEM-Mosig, line 529: "Ignoring the prediction" sounds a bit weird. How about "to not predict"?

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+    that `IntentTEDPolicy` had determined at training time.
+    A tolerance of `1.0` means that the threshold score is so low that `IntentTED` would not
+    trigger `action_unlikely_intent` for any of the "negative examples" that it has encountered
+    during training. These negative examples are combinations of dialogue contexts and user
+    intents that are _incorrect_. `IntentTEDPolicy` generates these negative examples from your
+    training data by picking a random story part and pairing it with a random intent that doesn't
+     occur at this point.
+
+The above configuration parameters are the ones you should try tweaking according to your use case and training data.
+However, additional parameters exist that can be adapted.
```
JEM-Mosig, line 546: I'd avoid passive voice. So "that you could adapt".

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+|                                       |                        | doesn't occur at this point.                                 |
++---------------------------------------+------------------------+--------------------------------------------------------------+
+```
+
+</details>
+
+#### Tuning the tolerance parameter
+
+When [reviewing real conversations](./conversation-driven-development.mdx#review), users are
+ encouraged to tune the `tolerance` parameter in `IntentTEDPolicy`'s configuration to reduce the number
```
JEM-Mosig, line 691: Maybe "we encourage you"

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+of false warnings (intents that actually are likely given the conversation context).
+As you increase the value of `tolerance` from `0` to `1` in steps of `0.05`,
```
JEM-Mosig, line 693: Why `0.05`?

docs/docs/policies.mdx
```diff
@@ -689,36 +993,59 @@ default restart, label `2` a greeting and `3` indicates goodbye.
...
+1. The next most appropriate action to be triggered by the assistant. For example, [`TEDPolicy`](#ted-policy) is trained to do this.
+2. The next most likely intent that a user can express. For example, [`IntentTEDPolicy`](#intenttedpolicy) is trained to do this.
+
+Hence, a tracker can be featurized to learn one of the above kind of labels.
+Depending on the policy, the target labels correspond to bot actions or bot utterances
+represented as an index in a list of all possible actions or
+set of intents represented as an index in a list of all possible intents.
 
-There are two different tracker featurizers:
+There are three different tracker featurizers:
```
JEM-Mosig, line 1006: "Tracker featurizers come in three different flavours:" (this avoids the "There are" at the beginning of the sentence, which carries no information)

docs/docs/policies.mdx
```diff
@@ -689,36 +993,59 @@ default restart, label `2` a greeting and `3` indicates goodbye.
...
 Deduplication is performed to filter out duplicated turns (bot actions
 or bot utterances) in terms of their previous states.
 
 For some algorithms a flat feature vector is needed, so input features
 should be reshaped to `(num_unique_turns, max_history * num_input_features)`.
 
+##### 3. Intent Max History
+
+`IntentMaxHistoryTrackerFeaturizer` inherits from `MaxHistoryTrackerFeaturizer`. Since, it is used by
+[`IntentTEDPolicy`](#intenttedpolicy), the target labels that it creates is the set of intents that can be
```
JEM-Mosig, line 1035: "the target labels that it creates _are_ the intents that can be..."

docs/docs/policies.mdx
```diff
@@ -689,36 +993,59 @@ default restart, label `2` a greeting and `3` indicates goodbye.
...
+expressed by a user in the context of a conversation tracker. Unlike
+other tracker featurizers, there can be multiple target labels. Hence, it pads the
+list of target labels with `-1` values on the right to return an equally sized list of target labels
```
JEM-Mosig, line 1038: Shouldn't we use the constant (or just write "a constant") here, instead of `-1`?

docs/docs/testing-your-assistant.mdx
```diff
@@ -377,6 +377,26 @@ The test script will also save a confusion matrix to a file called
...
+The test script will also generate a warnings file called `results/stories_with_warnings.yml`.
+This file contains all test stories for which [`action_unlikely_intent`](./default-actions.mdx#action_unlikely_intent)
+was predicted at any conversation turn but all actions from the original story were predicted correctly.
+However, if a test story originally included an `action_unlikely_intent` but wasn't predicted by
+the ensemble of policies, then the corresponding story will end up in `results/failed_test_stories.yml` as
+a failed story.
+
+The stories are sorted by the severity of `action_unlikely_intent`'s prediction.
+This severity is calculated by `IntentTEDPolicy` itself at prediction time.
+Higher the severity, the more unlikely is the intent and hence reviewing that particular
```
JEM-Mosig, line 391: "_The_ higher the severity..."

docs/docs/testing-your-assistant.mdx
JEM-Mosig: "You should" or "We recommend you go through..." instead of "User are encouraged to"

JEM-Mosig (approved):
Looks good! Just a few minor things.

docs/docs/policies.mdx
kedz: I would be in favor of dropping it, as it potentially adds confusion about IntentTEDPolicy predicting intents vs triggering actions.

ArjaanBuijk: Me too. It breaks the flow for the reader. I think it is very good information, but maybe put somewhere else, for those who want to learn more about the internals.

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+
+#### Prediction of `action_unlikely_intent`
+
+`IntentTEDPolicy` is invoked immediately after a user utterance and can either
+trigger `action_unlikely_intent` or abstain (in which case other policies will predict actions).
+To determine if `action_unlikely_intent` should be triggered, `IntentTEDPolicy` computes a score
+ for the user's intent in the current dialogue context and checks if this score is below a
+ certain threshold score.
+
+This threshold score is learned at training time but can be adjusted at test time
```
ArjaanBuijk, line 483: > adjusted at `test` time

Should this be: adjusted at `inference` time ?

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+  ```yaml-rasa title="config.yml"
+  policies:
+  - name: IntentTEDPolicy
+    epochs: 200
+  ```
+
+* `max_history`:
+  This parameter controls how much dialogue history the model looks at before making an inference.
+  Default `max_history` for this policy is `None`, which means that the complete dialogue history
+  since session restart is taken into account. If you want to limit the model
```
ArjaanBuijk, line 510: > since session `restart`

since session `(re)start`   ??

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+
+  ```yaml-rasa title="config.yml"
+  policies:
+  - name: IntentTEDPolicy
+    epochs: 200
+  ```
+
+* `max_history`:
+  This parameter controls how much dialogue history the model looks at before making an inference.
+  Default `max_history` for this policy is `None`, which means that the complete dialogue history
```
ArjaanBuijk, line 509: Why do we have `None` as the default?

I expect that the first thing we will do is recommend a more sensible value, like `5`. Why not make that the default?

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+    This parameter lets you configure `IntentTEDPolicy` to ignore the prediction of `action_unlikely_intent` for
+    a subset of intents. You might want to do this if you come across a certain list of intents for which there
+    are too many false warnings generated.
+
+* `tolerance`:
+    The `tolerance` parameter is a number that ranges from `0.0` to `1.0` (inclusive).
+    It helps to adjust the threshold score used during prediction of `action_unlikely_intent`
+    at inference time. Here, `0.0` means that the score threshold is the one
+    that `IntentTEDPolicy` had determined at training time.
+    A tolerance of `1.0` means that the threshold score is so low that `IntentTED` would not
```
ArjaanBuijk, line 538: Should it be `IntentTEDPolicy` instead of `IntentTED` ?

docs/docs/policies.mdx
ArjaanBuijk: I do not understand the explanation of `negative examples`, but at the same time, I wonder if it is needed to explain this here. This might be better covered somewhere else where the internal workings are explained.

docs/docs/policies.mdx
```diff
@@ -396,6 +398,308 @@ for details.
...
+|                                       |                        | intents that are _incorrect_. `IntentTEDPolicy` generates    |
+|                                       |                        | these negative examples from your training data by picking a |
+|                                       |                        | random story part and pairing it with a random intent that   |
+|                                       |                        | doesn't occur at this point.                                 |
++---------------------------------------+------------------------+--------------------------------------------------------------+
+```
+
+</details>
+
+#### Tuning the tolerance parameter
```
ArjaanBuijk, line 688: Should this tuning be done in combination with tuning of the `max_history`?
Especially if the default for max_history is None.

docs/docs/testing-your-assistant.mdx
ArjaanBuijk: Still, is there an approach to get rid of all warnings? 

Can a test story be marked to ignore the warning?

If we do not give some utitity for that, you end up looking at the same warnings all the time, which is frustrating, and you might not spot new, real warnings.

ArjaanBuijk (commented):
Looks great!! The explanations are very clear.

docs/docs/testing-your-assistant.mdx
dakshvar22: > is there an approach to get rid of all warnings?

You can run `rasa test` with `--no-warnings` flag.

> Can a test story be marked to ignore the warning?

Nope, we don't have functionality for that but seems like a very valid point. We could maybe reuse the `metadata` key of a story for users to explicitly mark it to be ignored for warnings. Something like - 

```
story: test x
metadata:
    warn: false
steps:
   - ...
```

I can't say for sure but it seems easy to do in `rasa test`. @TyDunn Do you think it makes sense to get this in the `2.8` release?
@alwx Can you please comment on the feasibility of this?

docs/docs/default-actions.mdx
dakshvar22: @alwx This is related to the [slack thread](https://rasa-hq.slack.com/archives/C01HHMR4X8S/p1622624011188400?thread_ts=1622620786.185400&cid=C01HHMR4X8S)

docs/docs/testing-your-assistant.mdx
dakshvar22: >  What I was hoping to draw reference to with the mention of the rule was the scenario where you included a rule within a larger test story that tested both that the action_unlikely_intent got triggered where you expected and then the corresponding rule got triggered. 

This makes sense. i reworded this to:

```
However, if a test story originally included an `action_unlikely_intent`, for example to ensure [a rule triggers what
happens after an `action_unlikely_intent`](./default-actions.mdx#customization-1) but the ensemble of
policies failed to do so, then the corresponding story will end up in `results/failed_test_stories.yml` as
a failed story.
```
Wdyt?

docs/docs/policies.mdx
dakshvar22: Dropped it 👍

docs/docs/policies.mdx
dakshvar22: 🙈

docs/docs/policies.mdx
dakshvar22: @ArjaanBuijk Let me try to explain `negative examples` to you and then maybe you could help us to articulate this easier.

If you had one just training story like:

```
story: greet unhappy
steps:
  - intent: greet
  - action: utter_greet
  - intent: mood_happy
  - action: goodbye
```
and you had an `affirm` intent, then a negative example will be:

input: `greet` -> `utter_greet`
output: `affirm`

This is called a negative example because acc to the training stories, `affirm` was never seen as the next intent after `greet` -> `utter_greet` conversation context.

We use these negative examples to figure out the range of values the model predicts for these negative examples constructed for each intent and `tolerance` is a way to be more / less cautious in prediction of `action_unlikely_intent`.

docs/docs/policies.mdx
dakshvar22: We have that in the code. We didn't make it flexible / customizable because we don't see a need for it to be as fine-grained as `0.01` right now based on our experiments.

docs/docs/policies.mdx
dakshvar22: Changed it to ```list of target labels with a constant value (`-1`) on the right```

docs/docs/conversation-driven-development.mdx
```diff
@@ -59,6 +59,23 @@ Your test users had at least some instruction about what the bot was intended to
 
 *   Look at conversations where an “out_of_scope” intent or fallback behavior occurred. These could indicate a potential new skill, or just a misclassified user utterance.
 *   Look for user frustration, such as requests for transfer to a human.
+*   If the assistant was trained with [`IntentTEDPolicy`](./policies.mdx#intenttedpolicy) included in the pipeline,
+    you can look for conversations where `action_unlikely_intent` is predicted at any conversation turn.
+    An `action_unlikely_intent` is predicted when the last intent expressed by the user is
+    unexpected in the current conversation context. You can also filter out such conversations by
+    running a [standalone script] which does the following:
+
+    - Fetch real conversations from a tracker store.
```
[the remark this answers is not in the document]
dakshvar22: From a tracker store.

docs/docs/default-actions.mdx
```diff
@@ -187,3 +187,39 @@ Overriding the default action for forms should **only** be used during the proce
...
 action to instruct Rasa Open Source to use the deprecated `FormAction` which is part of
 the Rasa SDK.
+
+## action_unlikely_intent
+
+Rasa Open Source triggers `action_unlikely_intent` via [`IntentTEDPolicy`](./policies.mdx#intenttedpolicy).
+You can control how often this action is predicted by tuning the [`tolerance`](./policies.mdx#intenttedpolicy)
+parameter of `IntentTEDPolicy`.
+
+### Customization
```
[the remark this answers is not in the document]
dakshvar22: `action_unlikely_intent` will be predicted however, it will be ignored for featurization by all policies which use only story data. So, effectively it will only be considered by `RulePolicy` in its predictions.

docs/docs/policies.mdx
dakshvar22: Yes, sounds more appropriate.

docs/docs/policies.mdx
dakshvar22: Just like `TED` we don't know what a sensible default is. Using the whole dialogue history does not make an assumption that only a limited amount of dialogue history is needed to make a prediction.

docs/docs/policies.mdx
dakshvar22: Ideally yes, but we haven't experimented with it as of now to recommend how exactly to do this. Maybe we can add a statement on the lines of - "If you change the `max_history` value and retrain a model, you might have to re-adjust the `tolerance` value."
Wdyt?

docs/docs/policies.mdx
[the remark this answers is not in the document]
dakshvar22: Here's the revamped description that I am thinking. Making Johannes' recommendation softer (can instead of will) but also include Chris' recommendation on using same `max_history` for `IntentTED` as for `TED`.

```
This parameter controls how much dialogue history the model looks at before making an inference.
  Default `max_history` for this policy is `None`, which means that the complete dialogue history
  since session (re)start is taken into account. If you want to limit the model
  to only see a certain number of previous
  dialogue turns, you can set `max_history` to a finite value.
  Please note that you should pick `max_history` carefully, so that the model has enough
  previous dialogue turns to create a correct prediction.
  Depending on your dataset, higher values of `max_history` can result in more frequent prediction of `action_unlikely_intent`
  as the number of unique possible conversation paths increases as more dialogue context is taken
  into account. Similarly, lowering the value of `max_history` can result in `action_unlikely_intent` being
  triggered less often but can also be a stronger indicator that the corresponding conversation path
  is highly unique and hence unexpected.
  We recommend you to set the `max_history` of `IntentTEDPolicy` equal to that of `TEDPolicy`.
```

I am advocating for equal `max_history` because max history is applied on states and not conversation turns, so `max_history - 1` will not play out very well in reality.

docs/docs/conversation-driven-development.mdx
```diff
@@ -59,6 +59,23 @@ Your test users had at least some instruction about what the bot was intended to
 
 *   Look at conversations where an “out_of_scope” intent or fallback behavior occurred. These could indicate a potential new skill, or just a misclassified user utterance.
 *   Look for user frustration, such as requests for transfer to a human.
+*   If the assistant was trained with [`IntentTEDPolicy`](./policies.mdx#intenttedpolicy) included in the pipeline,
+    you can look for conversations where `action_unlikely_intent` is predicted at any conversation turn.
+    An `action_unlikely_intent` is predicted when the last intent expressed by the user is
+    unexpected in the current conversation context. You can also filter out such conversations by
+    running a [standalone script] which does the following:
```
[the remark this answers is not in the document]
dakshvar22: I'll keep this thread open, so that I know before merging this PR.

docs/docs/command-line-interface.mdx
```diff
@@ -152,6 +152,10 @@ rasa interactive
 
 This will first train a model and then start an interactive shell session.
 You can then correct your assistants predictions as you talk to it.
+Please, note that if [`IntentTEDPolicy`](./policies.mdx#intenttedpolicy) is
+included in the pipeline, `action_unlikely_intent` can be predicted at any conversation turn.
+You can then choose to confirm or deny the action prediction, but this choice doesn't matter
+as running this action doesn't have any implications.
```
alwx, line 158: @dakshvar22 I've added this information about `rasa interactive` here

docs/docs/policies.mdx
ArjaanBuijk: @dakshvar22 ,
I re-read the description after seeing your example, and now it is now mostly clear. The one item that is not fully clear is the `score threshold of IntentTEDPolicy`.  Is that explainable in a few words?

A shot at an explanation of the `tolerance` that I would understand is as follows:

---
The `tolerance` parameter is a number that ranges from `0.0` to `1.0` (inclusive).
It helps to adjust the `threshold score` used during prediction of `action_unlikely_intent` at inference time.

During training, `IntentTED` generates "negative examples". For example, if you had one just training story like:

story: greet unhappy
steps:
  - intent: greet
  - action: utter_greet
  - intent: mood_happy
  - action: goodbye
and you had an affirm intent, then a negative example that `IntentTEDPolicy` will generate is:

input: greet -> utter_greet
output: affirm

This is called a negative example because in the training data, affirm was never seen as the next intent after greet -> utter_greet.

For each intent, `IntentTEDPolicy` uses these negative examples to figure out the range of values the model predicts for these negative examples,  and the `tolerance` is a way to be more / less cautious in prediction of `action_unlikely_intent`.
*==> Can we insert a brief explanation of the score threshold here ?*


- A tolerance of `0.0` means that the score threshold (1) is the one that `IntentTEDPolicy` had determined at training time.
- A tolerance of `1.0` means that the threshold score (1,2) is so low that `IntentTED` would not trigger `action_unlikely_intent` for any of the "negative examples" that it has encountered during training. 
---
(1) Is there a difference between `score threshold` and `threshold score` ?
(2) Still struggling to understand what happens when the tolerance is `1.0`. Probably because I do not understand the `score threshold`.

docs/docs/policies.mdx
ArjaanBuijk: Yes, that would make sense to me.

docs/docs/default-actions.mdx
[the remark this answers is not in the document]
ArjaanBuijk: @dakshvar22 ,

Do I understand correctly that:
-  `action_unlikely_intent` cannot be used in stories? 
- If you do not have a rule for it either, it is just ignored.

I did not read that anywhere, or I might have just missed that detail.

docs/docs/policies.mdx
ArjaanBuijk: Don't we know from experience that 5 is a reasonable default to start with, and then increase it when necessary?

I think we know that using the whole dialogue history is mostly not good.

docs/docs/testing-your-assistant.mdx
alwx: We can definitely do that, and it doesn't seem hard so it's more like a question of how important it is.

docs/docs/default-actions.mdx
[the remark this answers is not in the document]
dakshvar22: > `action_unlikely_intent` cannot be used in stories?

Yes, it can only appear in rules for training. At inference time, `IntentTEDPolicy` can predict it at anytime but will be ignored by policies which are trained only on stories, for example - `TEDPolicy`, `IntentTEDPolicy`, `MemoizationPolicy`, etc.. This is done to ensure that inference does not break because of a presence of `action_unlikely_intent` at inference time.

> If you do not have a rule for it either, it is just ignored.

Effectively yes. No policy's predictions are expected to be affected by its presence in that case.

docs/docs/policies.mdx
dakshvar22: I agree with you, but we cannot change the defaults in a minor version. The defaults of `TEDPolicy` and `IntentTEDPolicy` will have to be changed together to keep them at a finite (and equal) value.

docs/docs/policies.mdx
dakshvar22: Added it 👍

docs/docs/testing-your-assistant.mdx
dakshvar22: Thanks! @TyDunn It makes sense to me to do something like this. This wouldn't be release blocking though. Would you be fine pushing it post release as part of a prioritized bug fix (I know it isn't a bugfix really)?

docs/docs/default-actions.mdx
```diff
@@ -187,3 +187,44 @@ Overriding the default action for forms should **only** be used during the proce
...
+    - action: action_unlikely_intent
+    - action: ask_human_handoff
+    - intent: affirm
+    - action: trigger_human_handoff
+```
+
+Alternatively, you can also override it's behaviour as a [`custom action`](./custom-actions.mdx):
+
+```python
+class ActionUnlikelyIntent:
```
alwx, line 214: doesn't it require to extend `Action`?

docs/docs/policies.mdx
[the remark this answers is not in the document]
dakshvar22: @JEM-Mosig @kedz Do you agree with the refactored description above?

docs/docs/testing-your-assistant.mdx
dakshvar22: @kedz Could you please take a look at this again?

docs/docs/policies.mdx
dakshvar22: @ArjaanBuijk 
> Is there a difference between score threshold and threshold score

No, both are same.

> Still struggling to understand what happens when the tolerance is 1.0. Probably because I do not understand the score threshold.

The threshold score lies between the minimum and maximum scores for all negative examples (for each intent). The exact score is decided on the basis of `tolerance` parameter. If `tolerance` is `0.25` then it means that the threshold score will be selected in such a way that 25% of negative examples wouldn't result in an `action_unlikely_intent`.

Does that help clarify?

docs/docs/command-line-interface.mdx
```diff
@@ -152,6 +152,10 @@ rasa interactive
...
```
dakshvar22, line 159: @alwx Wdyt?
```suggestion
If [`IntentTEDPolicy`](./policies.mdx#intent-ted-policy) is
included in the pipeline, [`action_unlikely_intent`](./default-actions.mdx#action_unlikely_intent] can be triggered at any conversation turn. If so, the following message will be displayed: `The bot wants to run 'action_unlikely_intent' to indicate that the last user message was unexpected at this point in the conversation. Check out IntentTEDPolicy docs to learn more.`. As the message states, this is an indication that you have explored a conversation path which is unexpected according to the current set of training stories and hence adding this path to training stories is a good idea. Like other bot actions, you can choose to confirm or deny running this action.

```

docs/docs/default-actions.mdx
dakshvar22: Good point. Added.

docs/docs/default-actions.mdx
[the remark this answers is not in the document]
dakshvar22: I've included this information in the description now.

dakshvar22:
@ArjaanBuijk I've added the changelog and migration guide for the feature merge to this PR itself as it is part of the documentation. Can you please give it a review?

docs/docs/policies.mdx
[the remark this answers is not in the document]
kedz: Looks good to me!

docs/docs/policies.mdx
dakshvar22: I have explained this in more detail in "Prediction of `action_unlikely_intent`"  section.

changelog/8913.feature.md
```diff
@@ -0,0 +1,46 @@
...
+`rasa test` command has also been adapted to `UnexpecTEDIntentPolicy`:
+
+- If a test story contains `action_unlikely_intent` and the policy ensemble does not trigger it, this leads to
+  a test error (wrongly predicted action) and the corresponding story will be logged in `failed_test_stories.yml`.
+- If the story does not contain `action_unlikely_intent` and Rasa Open Source does predict it then
+  the prediction of `action_unlikely_intent` will be ignored for the evaluation (and hence not lead
+  to a prediction error) but the story will be logged in a file called `stories_with_warnings.yml`.
+
+
+`rasa data validate` command should warn if `action_unlikely_intent` is
```
ArjaanBuijk, line 38: `will warn`

changelog/8913.feature.md
```diff
@@ -0,0 +1,46 @@
...
+included in the training stories. Accordingly, `YAMLStoryWriter` and `MarkdownStoryWriter` have been updated to not dump `action_unlikely_intent` when writing stories to a file.
+
+:::caution
+The introduction of a new default action makes the release **breaking backwards compatibility of machine learning models**.
```
ArjaanBuijk, line 42: The introduction of a new default action **brakes backward compatibility of previously trained models**

docs/docs/conversation-driven-development.mdx
```diff
@@ -59,6 +59,23 @@ Your test users had at least some instruction about what the bot was intended to
...
+
+    - Fetch real conversations from a tracker store.
+    - Run `rasa test` on the fetched conversations and filter conversations containing `action_unlikely_intent`
+      in a separate warnings file. You can read more on [how to interpret these warnings](./testing-your-assistant.mdx#interpreting-the-generated-warnings).
+
+    Reviewing this subset of conversations can help you understand if real users have taken a
+    conversation path which is not present in the training data and hence "surprising"
+    for machine learning policies like `TEDPolicy`. Adding these conversation paths (with potential
+    corrections if `TEDPolicy` subsequently failed) as training stories will help in more robust action prediction
```
ArjaanBuijk, line 75: .... will **result** in more robust action prediction...

docs/docs/migration-guide.mdx
```diff
@@ -12,6 +12,33 @@ how you can migrate from one version to another.
 
 ## Rasa 2.7 to 2.8
 
+:::caution
+This is a release **breaking backwards compatibility of machine learning models**.
```
ArjaanBuijk, line 16: This release breaks backward....

docs/docs/migration-guide.mdx
```diff
@@ -12,6 +12,33 @@ how you can migrate from one version to another.
...
+It is not possible to load models trained with previous versions of Rasa Open Source. Please re-train
+your assistant before trying to use this version.
```
ArjaanBuijk, line 18: ... before using this version.

ArjaanBuijk (approved):
@dakshvar22 , @kedz .
I read it all again & found that everything is really well described now.
I proposed just a few small changes to the text.

kedz (approved):
Looks good to me!

PR #8912, opened 2021-06-17
IntentTEDPolicy: Fix when predictions should be skipped

dakshvar22:
**Proposed changes**:
- ...

### DIFF
--- a/rasa/core/policies/intent_ted_policy.py
+++ b/rasa/core/policies/intent_ted_policy.py
@@ -447,7 +447,7 @@ class IntentTEDPolicy(TEDPolicy):
         # 2. There is at least one event of type `ActionExecuted`
         #    after the last `UserUttered` event.
         if not tracker.get_last_event_for(UserUttered) or (
-            tracker.events and self._is_queried_for_first_time(tracker)
+            tracker.events and self._is_queried_before_action_executed(tracker)
         ):
             logger.debug(
                 f"Skipping predictions for {self.__class__.__name__} "
@@ -485,14 +485,14 @@ class IntentTEDPolicy(TEDPolicy):
         )
 
     @staticmethod
-    def _is_queried_for_first_time(tracker: DialogueStateTracker) -> bool:
-        """Checks if query was made for the first time after latest `UserUttered`.
+    def _is_queried_before_action_executed(tracker: DialogueStateTracker) -> bool:
+        """Checks if query was made before an `ActionExecuted` event.
 
         Should check if there is an `ActionExecuted` after the last `UserUttered`.
 
         Returns:
-            Whether the policy was queried for the first time after latest
-            `UserUttered`.
+            Whether the policy was queried after `UserUttered`
+            but before `ActionExecuted`.
         """
         applied_events = tracker.applied_events()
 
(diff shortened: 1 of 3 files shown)


rasa/core/policies/intent_ted_policy.py
```diff
@@ -482,6 +484,33 @@ def predict_action_probabilities(
             action_metadata=self._collect_action_metadata(domain, similarities),
         )
 
+    @staticmethod
+    def _is_queried_for_first_time(tracker: DialogueStateTracker) -> bool:
```
JEM-Mosig, line 488: I find the name a bit unintuitive. Maybe `_does_not_end_with_action`

JEM-Mosig: Why can't we just do one loop that goes backwards in the applied events and returns `False` once it encounters an `ActionExecuted` and `True` when it finds a `UserUttered`?

tests/core/policies/test_intent_ted_policy.py
```diff
@@ -532,6 +538,76 @@ def test_action_unlikely_intent_prediction(
...
+        should_skip: bool,
+        tmp_path: Path,
+    ):
+        caplog.set_level(logging.DEBUG)
+        loaded_policy = self.persist_and_load_policy(trained_policy, tmp_path)
+        interpreter = RegexInterpreter()
+        tracker = DialogueStateTracker(sender_id="init", slots=default_domain.slots)
+        tracker.update_with_events(tracker_events, default_domain)
+
+        _ = loaded_policy.predict_action_probabilities(
```
JEM-Mosig, line 603: Do you need the `_ = `?

JEM-Mosig (commented):
Looks good. Just some questions and a suggested name change.

kedz (commented):
+1 to @JEM-Mosig name suggestion. I also share his question about doing the check in one backward loop. But either way, it solves the bug and looks good.

rasa/core/policies/intent_ted_policy.py
dakshvar22: Regarding the name -> I would have agreed on the name if this function was inside `DialogueStateTracker` class. If it is inside `IntentTEDPolicy` it should describe an activity of `IntentTEDPolicy`. From that perspective, `_is_queried_for_first_time` made sense to me because it checks if `IntentTEDPolicy` is queried for the first time after the last UserUttered.
@kedz @JEM-Mosig Wdyt?

kedz: What about `_is_queried_before_action_executed`?

dakshvar22: Perfect! Renamed and refactored the logic.

rasa/core/policies/intent_ted_policy.py
```diff
@@ -482,6 +484,25 @@ def predict_action_probabilities(
...
+
+        Returns:
+            Whether the policy was queried after and `ActionExecuted`.
+        """
+        applied_events = tracker.applied_events()
+
+        found_action = False
+        for event in reversed(applied_events):
+            if isinstance(event, ActionExecuted):
+                found_action = True
```
kedz, line 501: Couldn't you also just return here?

dakshvar22: Ahh, now I see what you both meant 🙈 . In my head I was always thinking of trying to find if `ActionExecuted` occured after a `UserUttered` only. But i see now why it can just be simplified to checking which occurred first from the back of the list of the events - `ActionExecuted` or `UserUttered`.
I'll change it on monday morning.

PR #8929, opened 2021-06-21
Convert Intent ID Hashes from Integer to String

### LINKED ISSUE #8928: NLU Intent ID's Appear in Scientific Notation
**Rasa version**:
2.5.X

**Python version**:

**Issue**:
In some cases, the generated intent ID's get automatically converted to scientific notation when consumed by Kafka and thus make it impossible for users to track / manage these ID's from the generated Rasa events

```
  "intent": {
    "id": -1.731541806890066e+18,
    "name": "affirm",
    "confidence": 0.23743031919002533
  },
```

b-quachtran:
**Proposed changes**:
- Convert Intent ID Hashes from Integer to String
- Resolves issue #8928 where hash gets serialized into scientific notation in some cases


joejuzl:
While I think a string probably makes more sense, I'm a bit worried about the further implications of this change:
- I think this will break the event schema (`rasa/shared/utils/schemas/events.py::24`).
- Could this be a problem if a custom action server relies on it being an integer?
- Same for custom event brokers?
- Rasa X I _think_ is ok because it stores the whole parse_data as json text.

b-quachtran:
@joejuzl Do you think this solution still makes sense given the implications or if there's a better place in the code to cast it to avoid breaking changes?

joejuzl:
@b-quachtran could it be converted in the `KafkaEventBroker`?

b-quachtran:
@joejuzl That would bring up object consistency issues between event brokers though right? I think this specific customer would be okay with a temporary fix using a custom Kafka Broker, but long-term does it make sense to consider changing type ` "id": {"type": "number"}` to a string?

wochinge:
How about we use a config flag in `Kafka` now to change the ID to string and create a 3.0 issue to change it globally to string as we can do breaking changes in a major?

changelog/8929.improvement.md
```diff
@@ -0,0 +1 @@
+Converted intent ID hashes from integer to string to prevent formatting issues downstream.
```
joejuzl, line 1: Should be updated to specify kakfa broker.

rasa/core/brokers/kafka.py
```diff
@@ -200,5 +205,14 @@ def _publish(self, event: Dict[Text, Any]) -> None:
         )
         self.producer.send(self.topic, value=event, key=partition_key)
 
+    def _convert_intent_id_to_string(self, event: Dict[Text, Any]) -> Dict[Text, Any]:
+        if event.get("event", "") == "user" and "id" in event.get("parse_data", {}).get(
+            "intent", {}
+        ):
+            event["parse_data"]["intent"]["id"] = str(
```
joejuzl, line 212: Do we also need to update the id's in the `"intent_ranking"`?

rasa/core/brokers/kafka.py
```diff
@@ -107,6 +110,8 @@ def publish(
         retry_delay_in_seconds: float = 5,
     ) -> None:
         """Publishes events."""
+        if self.convert_intent_id_to_string:
```
joejuzl, line 113: Would be good to get a unit test for this.

joejuzl (changes requested):

rasa/core/brokers/kafka.py
```diff
@@ -212,6 +212,9 @@ def _convert_intent_id_to_string(self, event: Dict[Text, Any]) -> Dict[Text, Any
             event["parse_data"]["intent"]["id"] = str(
                 event["parse_data"]["intent"]["id"]
             )
+            for idx, parse_data in enumerate(event["parse_data"]["intent_ranking"]):
+                parse_data["id"] = str(parse_data["id"])
+                event["parse_data"]["intent_ranking"][idx] = parse_data
```
joejuzl, line 217: I don't think you need this line (or the `enumerate`) as you are modifying the dict in the line above.

joejuzl (approved):
Nice one! 
One, small comment, and then I think it's good to go 👍

rasa/core/brokers/kafka.py
b-quachtran: Ah so `intent_ranking` is a separate field from `intent` and contains the full breakdown of intents + ID's

joejuzl: Sorry missed this before was merged.
I mean that `parse_data` is a direct reference to `event["parse_data"]["intent_ranking"][idx]` so when you update `parse_data["id"]` you are also also updating `event["parse_data"]["intent_ranking"][idx]["id"]` so no need to re-assign  `parse_data` to `event["parse_data"]["intent_ranking"][idx]`. And thus you no longer need to know `idx`.

b-quachtran: Ah makes sense, I'll create another PR to get this updated

PR #8931, opened 2021-06-22
IntentTEDPolicy: Add severity to metadata

dakshvar22:
**Proposed changes**:
- Refactors metadata attached to the prediction of the policy.
- Metadata looks like this now:

```
Metadata schema looks like this:

{
    "query_intent": <metadata of intent that was queried>,
    "ranking": <sorted list of metadata corresponding to all intents.
                Sorting is based on predicted similarities.>
}

Each metadata dictionary looks like this:

{
    "name": <name of intent>,
    "score": <predicted similarity score>,
    "threshold": <threshold used for intent>,
    "severity": <absolute difference between score and threshold>
}
```

### DIFF
--- a/rasa/core/policies/intent_ted_policy.py
+++ b/rasa/core/policies/intent_ted_policy.py
@@ -397,8 +397,8 @@ class IntentTEDPolicy(TEDPolicy):
         self.compute_label_quantiles_post_training(model_data, label_ids)
 
     def _collect_action_metadata(
-        self, domain: Domain, similarities: np.array
-    ) -> Dict[Text, Dict[Text, float]]:
+        self, domain: Domain, similarities: np.array, query_intent: Text
+    ) -> Dict[Text, Any]:
         """Adds any metadata to be attached to the predicted action.
 
         Similarities for all intents and their thresholds are attached as metadata.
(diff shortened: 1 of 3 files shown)


rasa/core/policies/intent_ted_policy.py
```diff
@@ -397,26 +403,70 @@ def run_training(
...
+                        Sorting is based on predicted similarities.>
+        }
+
+        Each metadata dictionary looks like this:
+
+        {
+            "name": <name of intent>,
+            "score": <predicted similarity score>,
+            "threshold": <threshold used for intent>,
+            "severity": <absolute difference between score and threshold>
```
kedz, line 424: Does severity need to be here if we are additionally storing the numbers to compute it? Not sure what our policy is on the tradeoff cost for storing this info vs recomputing later.

rasa/core/policies/intent_ted_policy.py
```diff
@@ -397,26 +403,70 @@ def run_training(
...
-        Similarities for all intents and their thresholds are attached as metadata.
+        Metadata schema looks like this:
+
+        {
+            "query_intent": <metadata of intent that was queried>,
+            "ranking": <sorted list of metadata corresponding to all intents.
+                        Sorting is based on predicted similarities.>
+        }
+
+        Each metadata dictionary looks like this:
```
kedz, line 418: "Each metadata dictionary in the ranking field looks like this:"

At first it wasn't clear to me how the top schema related to the bottom.

kedz: Oh actually I see that both fields contain metadata, where query_intent is a singleton and ranking is a list. Perhaps maybe worth noting that the query_intent metadata is also in ranking?

kedz (commented):
I think in the discussion when we talked about severity, I was assuming it was only being calculated for intents which triggered `action_unlikely_intent`, but it looks like its getting computed for all intents. As its now calculated, If one intent is above the threshold but equally distant to it as an intent below the threshold, wont they be assigned the same severity? This seems confusing to me if so.

Why not just calculate severity as threshold - score? This would give intents above the threshold negative severity. Or alternatively max(0, threshold - score) and intents above the threshold have 0 severity?

JEM-Mosig (approved):
Looks good!

JEM-Mosig:
> Why not just calculate severity as threshold - score? This would give intents above the threshold negative severity. Or alternatively max(0, threshold - score) and intents above the threshold have 0 severity?

Good point! I think `threshold - score` is better.

rasa/core/policies/intent_ted_policy.py
dakshvar22: Computation of severity is a property of `IntentTEDPolicy`, so it should definitely stay inside it and added to the metadata.
On the tradeoff: Users can always use `ranking_length` parameter to reduce the ranking list if they see that exploding the memory.

dakshvar22: Thanks for bringing that up, support for `ranking_length` was missing and I added it now.

dakshvar22:
Agree about changing the severity measure. Changed it to `threshold - score`

rasa/core/policies/intent_ted_policy.py
dakshvar22: > Perhaps maybe worth noting that the query_intent metadata is also in ranking?

Added that in.

kedz (approved):
Looks good!

PR #8938, opened 2021-06-23
Use RASABOT_GITHUB_TOKEN to create issues in workflows

m-vdb:
**Proposed changes**:
- Use `RASABOT_GITHUB_TOKEN` to create issues in workflows so that subsequent workflows are triggered. (the behaviour is described [here](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token))

### DIFF
--- a/.github/workflows/ci-model-regression-on-schedule.yml
+++ b/.github/workflows/ci-model-regression-on-schedule.yml
@@ -519,7 +519,8 @@ jobs:
         if: steps.performance.outputs.is_dropped == 'true' && steps.issue-exists.outputs.result == 'false'
         uses: actions/github-script@a3e7071a34d7e1f219a8a4de9a5e0a34d1ee1293  # v4
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-model-regression-on-schedule.yml
```diff
@@ -519,7 +519,8 @@ jobs:
         if: steps.performance.outputs.is_dropped == 'true' && steps.issue-exists.outputs.result == 'false'
         uses: actions/github-script@a3e7071a34d7e1f219a8a4de9a5e0a34d1ee1293  # v4
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # do not use GITHUB_TOKEN here because it wouldn't trigger subsequent workflows
+          github-token: ${{ secrets.RASABOT_GITHUB_TOKEN }}
```
m-vdb, line 523: do we have documentation somewhere about the permissions granted to the token? here I'm assuming that this token can create issues

HotThoughts (approved):
Thank you so much for doing this 🕺

PR #8944, opened 2021-06-23
Display actual values in UserWarning re: "evaluate_every_number_of_epoch" > "epochs"

TomV:
I got a userWarning when running in 2.7 shell: 

`UserWarning: the value of 'evaluate_every_number_of_epochs' is greater than the value of 'epochs'. No evaluation will occur.`

Very odd error on a fairly generic project ( originally created using rasa 2.5).  While troubleshooting, I modified the warning to show the actual values of both epochs and evaluate_every_number_of_epochs.  

To my surprise, epochs value is set to 1 (!), all my config values are 100.  So I think there's a deeper problem (_check_evaluation_setting) is being called at the wrong point or using the wrong value for "EPOCHS"? 

Ignoring for now the bigger error, this small change to the code improves the error message to help anyone troubleshooting their own configurations.  An example error after the update includes (value=X), so it looks like this:

`UserWarning: the value of 'evaluate_every_number_of_epochs' (value=20) is greater than the value of 'epochs' (value=1). No evaluation will occur.`

The nice thing about this is that a user can play with the configuration of different pipeline element configurations (TEDPolicy, DIETClassifier, etc.) that include a config value for number of epochs to quickly pinpoint which configuration is causing the error, since you can set the value in more than one place.  Just setting my epochs values to 101, 102, etc. helps pinpoint where the the problem with the config is occuring.  

I'm a new to the rasa code base, so I'll leave it to others to figure out WHY the epoch value is set to 1 in my test.  (Open a new issue?)

**Proposed changes**:
- ...


sara-tagger:
Thanks for submitting a pull request 🚀 @yennycheung will take a look at it as soon as possible ✨

dakshvar22 (approved):
LGTM! Thanks for the contribution 🎉

dakshvar22:
@TomV Can you please fix the test: `tests/utils/test_train_utils.py::test_warning_eval_num_epochs_greater_than_epochs` and also [reformat the code with `black`](https://github.com/RasaHQ/rasa#code-style). Post that, the tests should pass and we can merge the PR. 🙏

dakshvar22:
Hi, this PR was completed as part of another PR. Hence, I am closing this one. Thanks for pointing out the issue in the first place :)

PR #8949, opened 2021-06-24
`IntentTEDPolicy`: Updated `rasa test` tests

alwx:
**Proposed changes**:
- do monkey patching inside `rasa test` tests because every time the tests are run, training will result in different model weights which will result in different predictions of `action_unlikely_intent`.
- update the imports inside `rasa.cli.test`

### DIFF
--- a/rasa/cli/test.py
+++ b/rasa/cli/test.py
@@ -9,6 +9,11 @@ from rasa.shared.exceptions import YamlException
 import rasa.shared.utils.io
 import rasa.shared.utils.cli
 from rasa.cli.arguments import test as arguments
+from rasa.core.constants import (
+    FAILED_STORIES_FILE,
+    SUCCESSFUL_STORIES_FILE,
+    STORIES_WITH_WARNINGS_FILE,
+)
 from rasa.shared.constants import (
     CONFIG_SCHEMA_FILE,
     DEFAULT_E2E_TESTS_PATH,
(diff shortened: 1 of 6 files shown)


rasa/cli/test.py
```diff
@@ -76,6 +71,12 @@ def add_subparser(
 def _print_core_test_execution_info(args: argparse.Namespace) -> None:
     output = args.out or DEFAULT_RESULTS_PATH
 
+    from rasa.core.test import (
```
wochinge, line 74: how about moving these constants to `rasa.core.constants` instead of doing a local import? And if we keep the local import: Can we move it to the top of the function instead of having it in the middle?

tests/core/test_test.py
```diff
@@ -4,10 +4,40 @@
...
 
 import rasa.core.test
+from rasa.core.policies.ensemble import SimplePolicyEnsemble
+from _pytest.monkeypatch import MonkeyPatch
 from _pytest.capture import CaptureFixture
 from rasa.core.agent import Agent
 
 
+def _probabilities_with_action_unlikely_intent_for(intent_name: Text):
+    from rasa.core.policies.policy import PolicyPrediction
```
wochinge, line 14: Why aren't we using top level imports?

tests/core/test_test.py
```diff
@@ -4,10 +4,40 @@
...
+    ) -> PolicyPrediction:
+        latest_event = tracker.events[-1]
+        if (
+            isinstance(latest_event, UserUttered)
+            and latest_event.parse_data["intent"]["name"] == intent_name
+        ):
+            # Here we return `action_unlikely_intent` if the name of the latest intent
+            # matches `intent_name`.
+            # We need to do it because every time the tests are run,
+            # training will result in different model weights which will
```
wochinge, line 30:
```suggestion
            # training will result in different model weights which might
```

tests/core/test_test.py
alwx: 👍

wochinge (approved):
💯

PR #8955, opened 2021-06-25
Slot mapping prototype

ancalita:
**Proposed changes**:
-  Explore a prototype for global slot mappings as part of this [issue](https://github.com/RasaHQ/enable_squad/issues/28).

### DIFF
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -35,9 +35,17 @@ from rasa.shared.core.constants import (
     ACTION_DEFAULT_ASK_REPHRASE_NAME,
     ACTION_BACK_NAME,
     REQUESTED_SLOT,
+    ACTION_EXTRACT_SLOTS,
 )
 from rasa.shared.exceptions import RasaException
-from rasa.shared.nlu.constants import INTENT_NAME_KEY, INTENT_RANKING_KEY
+from rasa.shared.nlu.constants import (
+    INTENT_NAME_KEY,
+    INTENT_RANKING_KEY,
+    ENTITY_ATTRIBUTE_VALUE,
+    ENTITY_ATTRIBUTE_TYPE,
+    ENTITY_ATTRIBUTE_GROUP,
+    ENTITY_ATTRIBUTE_ROLE,
+)
 from rasa.shared.core.events import (
     UserUtteranceReverted,
     UserUttered,
@@ -51,8 +59,7 @@ from rasa.shared.core.events import (
 )
 from rasa.shared.utils.schemas.events import EVENTS_SCHEMA
 from rasa.utils.endpoints import EndpointConfig, ClientResponseError
-from rasa.shared.core.domain import Domain
-
+from rasa.shared.core.domain import Domain, SlotMapping
 
 if TYPE_CHECKING:
     from rasa.shared.core.trackers import DialogueStateTracker
@@ -77,6 +84,7 @@ def default_actions(action_endpoint: Optional[EndpointConfig] = None) -> List["A
         ActionDefaultAskRephrase(),
         TwoStageFallbackAction(action_endpoint),
         ActionBack(),
+        ActionExtractSlots,
     ]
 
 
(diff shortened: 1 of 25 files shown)


rasa/core/actions/forms.py
```diff
@@ -98,11 +98,14 @@ def get_mappings_for_slot(
...
-            domain.slot_mapping_for_form(self.name()).get(
-                slot_to_fill, self.from_entity(slot_to_fill),
-            )
-        )
+        # requested_slot_mappings = self._to_list(
+        #     domain.slot_mapping_for_form(self.name()).get(
+        #         slot_to_fill, self.from_entity(slot_to_fill),
+        #     )
+        # )
+        domain_slots = domain.as_dict().get("slots")
```
wochinge, line 106: not related but why do we need `as_dict()`? Can't we just access the things directly on the `domain` object?

ancalita: I couldn't see another way to access the slot mappings for a specific slot_name - `domain.slots` is a list of `Slot` objects and I couldn't see how to 'fish' the right Slot 🤔

wochinge: mhm, whatever `as_dict` contains must be _somewhere_ in the domain object.

ancalita:
This is a brief summary of the changes made in this prototype - which is also the order I recommend reviewing in:
- `slots.py`: added `slot_mappings` argument to the `Slot` constructors & deactivated auto-fill.
- `domain.py`: 
      - enriched `SlotMapping` class to have a new type (CUSTOM), adapted validation method to get rid of forms, added two new static methods to check if the mapping intent matches tracker last message intent or mapping entity matches tracker extracted entities [These methods I lifted from forms and adapted for my purpose]
      - added new `_validate_forms` and changed `_validate_slot_mappings` used in `Domain.from_dict`
- `forms.py`: adapted `FormAction` to match changes in Domain & added mapping`condition` validation in `extract_requested_slot` [As I write this, I realise I should have done the same for `extract_other_slots` - I'll add this quickly tomorrow Tue]
- `action.py`:
      - created new default action `ActionExtractSlots`: its run method always loops through every slot and all its mappings and checks if anything is applicable to be set after each new user message; 
      - if it encounters a `condition` key in its mapping, it returns a list of empty events since that mapping is only applicable in a form [the assumption I worked with is that condition only indicates if a mapping is valid only within a certain form, therefore making `active_loop` a mandatory key, while `requested_slot` is optional. If `requested_slot` is used, then it needs to be used with `active_loop`, but this is validation work I didn't implement in the prototype]
      - had to make `output_channel` of type `Optional` for `RemoteAction` and `ActionExtractSlots` to make my bot examples run
- `processor.py`: execute `ActionExtractSlots` for every message in `MessageProcessor.handle_message()` before the next action prediction

I've tested that:
- global slot setting works outside forms for both predefined and custom mappings - the slots can be updated later in the dialogue too.
- forms work as expected & also with re-usable slots in `required_slots`
- when a condition is specified in the slot mapping, that mapping kicks in only during the correct active_loop

rasa/core/actions/action.py
```diff
@@ -884,3 +892,82 @@ def name(self) -> Text:
...
+                        if isinstance(event, SlotSet):
+                            slot_events.append(event)
+                        else:
+                            break
+                    value = None
+                else:
+                    value = None
+
+                if mapping.get("condition"):
+                    value = None
```
wochinge, line 965: why do we need this one?

rasa/core/actions/forms.py
```diff
@@ -311,6 +311,17 @@ def extract_other_slots(
                 slot_mappings = self.get_mappings_for_slot(slot, domain)
 
                 for slot_mapping in slot_mappings:
+                    condition_dict = slot_mapping.get("condition")
```
wochinge, line 314: (this is [Hungarian notation btw](https://en.wikipedia.org/wiki/Hungarian_notation))

rasa/core/actions/forms.py
```diff
@@ -145,7 +143,9 @@ def _create_unique_entity_mappings(self, domain: Domain) -> Set[Text]:
         """
         unique_entity_slot_mappings = set()
         duplicate_entity_slot_mappings = set()
```
wochinge, line 145: Why do we still need to run all of the `FormAction`. Don't we need to comment out `activate` and the first line of `def do`?

wochinge: and wouldn't the `condition` part then go into `request_next_slot` / `is_done`?

rasa/shared/core/domain.py
```diff
@@ -1922,38 +1927,87 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
+        mapping_not_intents = slot_mapping.get("not_intent", [])
+
+        intent = tracker.latest_message.intent.get("name")
+
+        intent_not_blocked = not mapping_intents and intent not in mapping_not_intents
+
+        return intent_not_blocked or intent in mapping_intents
+
+    @staticmethod
+    def entity_is_desired(
```
wochinge, line 1973: are you actually using this code?

wochinge (commented):
Some general comments. Will now try it out 🎉

wochinge:
ok, seems `rasa interactive` is broken 🙈

wochinge:
Thoughts from testing it out:
* More for an implementation proposal: We should maybe not add `SlotSet` events unless they are changing the state. Otherwise we have a bunch of useless `SlotSet` events after every user message
* having `utter_ask_{slot_name}` would be helpful for the testing (I can look in the tracker but others might not)
* I think slots are currently double extracted (related to[ this comment](https://github.com/RasaHQ/rasa/pull/8955#discussion_r660541763))

rasa/shared/core/domain.py
ancalita: Yep, it's used in `ActionExtractSlots.run` - although only once 😄

rasa/core/actions/forms.py
ancalita: > Why do we still need to run all of the FormAction. Don't we need to comment out activate and the first line of def do?
Because we kept the concept of the `form`, I didn't want to remove anything in `FormAction`, only adapted bits and pieces here

> wouldn't the condition part then go into request_next_slot / is_done?
💡 I didn't consider these methods - those could be better ways for handling slot mapping condition.

ancalita:
@wochinge many thanks for trying it out and leaving feedback 💯 
> We should maybe not add SlotSet events unless they are changing the state. 
How would we know which ones of these events change the state?

>  having utter_ask_{slot_name} would be helpful for the testing
Yes, my bad, I wasn't sure how to add this, should have asked 🙈 

> slots are currently double extracted
Not sure I follow here - you mean during a form, they are extracted once by `ActionExtractSlots` and 2nd time by the form?
I might have overlooked this part 😬 Thanks for raising, something to consider for the implementation proposal.

JEM-Mosig:
@ancalita I'll review the code later. For now I've just built a bot with it to get you some feedback. The bot implements the [Better Conditional Responses](https://www.notion.so/rasa/Taking-Apart-Forms-be897aa7fd54401bbe398702302094f6#63d46b42870b44e2b5bbda89923b0610) example from the proposal. It works great! You can find the bot and some dirty bug-fixes on the `slot-prototype-johannes` branch. 
Just a few things:

1. Slots don't have to have slot mappings, but domain loading fails with `The slot 'xyz' has no mappings defined.` if you don't specify any. I fixed this in my branch, but in a quick-and-dirty way.
2. If the custom action fails (e.g. due to bug in code), Rasa doesn't handle it well so I have to restart it.
3. I'd always use a `dummy` slot to specify one slot mapping action that does everything (see example bot in my branch). I suggest we offer some other way of specifying this instead of the `custom` slot mapping separate for each slot. (as I've mentioned before, this is fine for this first prototype, though)

ancalita:
hey @JEM-Mosig great feedback, thank you!
Regarding points 1 + 3 - I think they're related and some food for thought if we want to make mappings optional overall with this change. I assumed that because 2.0 requires slot mappings in the form, these should be mandatory per slot, but your example with the custom mappings doesn't require this. We can discuss this at length with the implementation proposal.

I'm interested in the point 2 errors - are those usage errors or bugs in the prototype that I need to pay attention to with the implementation proposal?

JEM-Mosig:
@ancalita 
> Regarding points 1 + 3 - I think they're related and some food for thought if we want to make mappings optional overall with this change. I assumed that because 2.0 requires slot mappings in the form, these should be mandatory per slot, but your example with the custom mappings doesn't require this. We can discuss this at length with the implementation proposal.

In any case you don't always need slot mappings _outside_ of a loop, so I'd say they have to be optional.

> I'm interested in the point 2 errors - are those usage errors or bugs in the prototype that I need to pay attention to with the implementation proposal?

They are basically uncaught exceptions. In the end, I think Rasa should still continue to work most of the time if the action server is not running (e.g. slots just don't get filled and you get a warning instead of a program terminating exception).

wochinge:
> How would we know which ones of these events change the state?

Basically if the new slot value is different than the current slot value.

PR #8987, opened 2021-06-30
Rename code occurrences of `IntentTEDPolicy` to `UnexpecTEDIntentPolicy`

dakshvar22:
**Proposed changes**:
- ...

### DIFF
--- a/rasa/core/policies/unexpected_intent_policy.py
+++ b/rasa/core/policies/unexpected_intent_policy.py
@@ -110,8 +110,8 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-class IntentTEDPolicy(TEDPolicy):
-    """`IntentTEDPolicy` has the same model architecture as `TEDPolicy`.
+class UnexpecTEDIntentPolicy(TEDPolicy):
+    """`UnexpecTEDIntentPolicy` has the same model architecture as `TEDPolicy`.
 
     The difference is at a task level.
     Instead of predicting the next probable action, this policy
@@ -226,7 +226,7 @@ class IntentTEDPolicy(TEDPolicy):
         # For each intent, the tolerance is the percentage of
         # negative training instances (trackers for which
         # the corresponding intent is not the correct label) that
-        # would be ignored by `IntentTEDPolicy`. This is converted
+        # would be ignored by `UnexpecTEDIntentPolicy`. This is converted
         # into a similarity threshold by identifying the similarity
         # score for the (1 - tolerance) percentile of negative
         # examples. Any tracker with a similarity score below this
@@ -552,7 +552,8 @@ class IntentTEDPolicy(TEDPolicy):
                  from getting stuck in a prediction loop.
                 For example, if the last `ActionExecuted` event
                 contained `action_unlikely_intent` predicted by
-                `IntentTEDPolicy` and if `IntentTEDPolicy` runs inference
+                `UnexpecTEDIntentPolicy` and
+                if `UnexpecTEDIntentPolicy` runs inference
                 on the same tracker, it will predict `action_unlikely_intent`
                 again which would make the dialogue manager get stuck in a
                 prediction loop.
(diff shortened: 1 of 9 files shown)


dakshvar22:
@wochinge Directly requesting a review from you here so that you don't need to take look at this in the final PR

dakshvar22:
@samsucik Added you as reviewer as well directly to this PR too so that you don't get surprised seeing these changes suddenly in the final PR 😉

samsucik (changes requested):
@dakshvar22 everything looks good but I think this line in `unexpected_intent_policy.py`:
```python
common.mark_as_experimental_feature("IntentTED Policy")
```
is one last place where IntentTED should also be renamed, right? Or maybe it's more complicated: At this point both the policy itself AND the underlying model (IntentTED) are experimental. But in the future the policy may become official while using yet another new (experimental) model 😅

dakshvar22:
Very good point Sam. Renamed this argument to `UnexpecTED Intent Policy` 😄

samsucik (approved):
Alrighty, looks good 👍

rasa/core/policies/unexpected_intent_policy.py
```diff
@@ -110,8 +110,8 @@
 logger = logging.getLogger(__name__)
```
wochinge, line 110: nice job with the file rename! 👍🏻

tests/core/policies/test_intent_ted_policy.py
```diff
@@ -15,7 +15,7 @@
     TrackerFeaturizer,
 )
```
wochinge, line 16: this file still needs renaming

tests/core/training/test_interactive.py
```diff
@@ -756,7 +757,8 @@ def test_retry_on_error_success(monkeypatch: MonkeyPatch):
             f"The bot wants to run 'action_unlikely_intent' "
             f"to indicate that the last user message was unexpected "
             f"at this point in the conversation. "
-            f"Check out IntentTEDPolicy ({DOCS_URL_POLICIES}/#intent-ted-policy) "
+            f"Check out UnexpecTEDIntentPolicy "
+            f"({DOCS_URL_POLICIES}/#unexpected-intent-policy) "
```
wochinge, line 761: the slash before the `#` looks weird, doesn't it? Same above

rasa/core/policies/unexpected_intent_policy.py
```diff
@@ -552,7 +552,8 @@ def _should_skip_prediction(tracker: DialogueStateTracker) -> bool:
                  from getting stuck in a prediction loop.
                 For example, if the last `ActionExecuted` event
                 contained `action_unlikely_intent` predicted by
-                `IntentTEDPolicy` and if `IntentTEDPolicy` runs inference
+                `UnexpecTEDIntentPolicy` and
```
wochinge, line 555: pls also fix `_metadata_filename`

wochinge: do you want to rename `class IntentTED` to `ExpecTED` as well?

wochinge (commented):
there are also some occurrences in the changelog items

tests/core/policies/test_intent_ted_policy.py
dakshvar22: Ohh man! Sorry, my bad 🤦

rasa/core/policies/unexpected_intent_policy.py
dakshvar22: > do you want to rename class IntentTED to ExpecTED as well?

No, `ExpecTED` does not make sense at all. `IntentTED` is good to indicate that the architecture is TED but for intents.

tests/core/training/test_interactive.py
dakshvar22: You are correct. It shouldn't have been in the first place. Nothing to do with renaming.

wochinge (approved):

PR #9002, opened 2021-07-01
Clarify `cache_dir` and LMTokenizer deprecation

koernerfelicia:
**Proposed changes**:
This slack [thread](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1625044540055100) highlighted some issues with the `LMFeaturizer` documentation:
1. it's not clear that the `cache_dir` can also be used to load pre-trained models locally
2. it's not clear that `LMFeaturizer` doesn't do any tokenization

I'm not married to this wording at all, please feel free to suggest big changes to it :)


aeshky (approved):
Thanks Felicia. The comments looks good to me.

koernerfelicia:
@aeshky thanks for the quick review, and as always pointing out when things are unclear! Will try and find a CSE person to give the final okay :)

docs/docs/components.mdx
```diff
@@ -212,8 +212,13 @@ now implements its behavior.
...
 
-      # An optional path to a specific directory to download and cache the pre-trained model weights.
-      # The `default` cache_dir is the same as https://huggingface.co/transformers/serialization.html#cache-directory .
+      # An optional path to a directory from which
+      # to load pre-trained model weights.
+      # If the requested model is not found in the
+      # directory, it will be downloaded and
+      # cached in this directory for future use.
+      # The default cache_dir is the same as
+      # https://huggingface.co/transformers/serialization.html#cache-directory
```
hsm207, line 221: The link:

https://huggingface.co/transformers/serialization.html#cache-directory

needs to be updated. There's no mention of cache_dir in the page linked.

docs/docs/components.mdx
```diff
@@ -793,8 +797,13 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
...
```
hsm207, line 806: Same comment about the link

hsm207 (changes requested):

docs/docs/components.mdx
koernerfelicia: Great catch, thank you! I wasn't able to find an appropriate docs page, so named the constant instead. What do you think?

hsm207: What do you think of this doc to describe `cache_dir`?

https://huggingface.co/transformers/installation.html#caching-models

koernerfelicia: That would be the perfect page, but we haven't updated our transformers to that version yet, we are still capped at `<2.12`.. In our version the default path is slightly different (`~/.cache/torch/transformers`)

I found [this](https://huggingface.co/transformers/v2.11.0/serialization.html#cache-directory), but it also isn't quite right, since we want the suffix to be `transformers`. When I do a site-specific search (`".cache/torch/" transformers site:huggingface.co/ 2.11`) nothing relevant comes up..

hsm207 (changes requested):

docs/docs/components.mdx
```diff
@@ -212,8 +212,14 @@ now implements its behavior.
...
 
-      # An optional path to a specific directory to download and cache the pre-trained model weights.
-      # The `default` cache_dir is the same as https://huggingface.co/transformers/serialization.html#cache-directory .
+      # An optional path to a directory from which
+      # to load pre-trained model weights.
+      # If the requested model is not found in the
+      # directory, it will be downloaded and
+      # cached in this directory for future use.
+      # The default value of `cache_dir` is the
+      # constant `TRANSFORMERS_CACHE`, defined in
```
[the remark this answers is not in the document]
koernerfelicia: Hmm kind of: I think this would take precedence over anything else if it were set, but if it's not set the default is different, see relevant code [here](https://github.com/huggingface/transformers/blob/v2.11.0/src/transformers/file_utils.py#L63-L76). I really don't know if this is splitting hairs, though, I'd be fine with saying environment variable if you are :)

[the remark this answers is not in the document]
koernerfelicia: But you're definitely right that "constant" isn't the right term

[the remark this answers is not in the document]
hsm207: Thanks for linking to the code. The `TRANSFORMERS_CACHE` is indeed meant to be an environment variable that the use could configure. Let's go with "environment variable" instead of "constant"

docs/docs/components.mdx
```diff
@@ -208,8 +208,15 @@ now implements its behavior.
...
-      # The `default` cache_dir is the same as https://huggingface.co/transformers/serialization.html#cache-directory .
+      # An optional path to a directory from which
+      # to load pre-trained model weights.
+      # If the requested model is not found in the
+      # directory, it will be downloaded and
+      # cached in this directory for future use.
+      # The default value of `cache_dir` can be
+      # set using the environment variable
+      # `TRANSFORMERS_CACHE`, as per the
+      # Transformers library.
```
koernerfelicia, line 219: @hsm207 sorry I dropped this -- I was out on vacation until now. What do you think of this?

hsm207: Perfect!

hsm207 (approved):

koernerfelicia:
Thanks @hsm207! Appreciate the extra time you put in to make sure we're accurate about `cache_dir`

PR #9016, opened 2021-07-04
UnexpecTEDIntentPolicy: Address review comments

dakshvar22:
**Proposed changes**:

Addresses:

- https://github.com/RasaHQ/rasa/pull/8913#discussion_r661364043
- https://github.com/RasaHQ/rasa/pull/8913#discussion_r661378213


dakshvar22:
Really not sure why code quality is failing [here](https://github.com/RasaHQ/rasa/pull/9016/checks?check_run_id=2984826522#step:10:19)

wochinge:
Seems to be due to the nesting. This fixes it:

```python
RankingCandidateMetadata = TypedDict(
    "RankingCandidateMetadata",
    {
        NAME: Text,
        SCORE_KEY: float,
        THRESHOLD_KEY: Optional[float],
        SEVERITY_KEY: Optional[float],
    },
)

UnexpecTEDIntentPolicyMetadata = TypedDict(
    "UnexpecTEDIntentPolicyMetadata",
    {
        QUERY_INTENT_KEY: Text,
        RANKING_KEY: List[RankingCandidateMetadata]
    },
)
```

rasa/core/policies/unexpected_intent_policy.py
```diff
@@ -109,6 +114,24 @@
...
+                {
+                    NAME: Text,
+                    SCORE_KEY: float,
+                    THRESHOLD_KEY: Optional[float],
+                    SEVERITY_KEY: Optional[float],
+                },
+            )
+        ],
+    },
+)
```
wochinge, line 133:
```suggestion
RankingCandidateMetadata = TypedDict(
    "RankingCandidateMetadata",
    {
        NAME: Text,
        SCORE_KEY: float,
        THRESHOLD_KEY: Optional[float],
        SEVERITY_KEY: Optional[float],
    },
)

UnexpecTEDIntentPolicyMetadata = TypedDict(
    "UnexpecTEDIntentPolicyMetadata",
    {
        QUERY_INTENT_KEY: Text,
        RANKING_KEY: List[RankingCandidateMetadata]
    },
)
```

rasa/core/policies/unexpected_intent_policy.py
```diff
@@ -4,6 +4,11 @@
 from pathlib import Path
 from typing import Any, List, Optional, Text, Dict, Type, Union, TYPE_CHECKING
 
+try:
+    from typing import TypedDict
+except ImportError:
+    from mypy_extensions import TypedDict
```
wochinge, line 10: why not doing it as in the other modules?

```
if TYPE_CHECKING:
    from typing_extensions import TypedDict
```

wochinge (commented):
🚀

rasa/core/policies/unexpected_intent_policy.py
wochinge: `mypy_extensions` is currently not even in our dependencies

dakshvar22: But don't you want to always import it somehow from one of the packages. I am not sure on when `TYPE_CHECKING` is set to `True` and when is it set to `False`, hence the question.

wochinge: `TYPE_CHECKING` is only true when we use `mypy` to analyze the code and during development in your IDE. As `typing_extension` is currently a dev package you need the `if` part (plus quotes around the type when using it). Good point that we should re-consider moving that our production packages 🤔

dakshvar22: > TYPE_CHECKING is only true when we use mypy to analyze the code and during development in your IDE

If that's the case, then you can't import it inside an `if` statement. You always want it to be imported somehow since the code needs to initialize a `TypedDict` and it will result in a runtime error if the `TypedDict` is never imported.

dakshvar22: Yep, the tests start [failing](https://github.com/RasaHQ/rasa/pull/9016/checks?check_run_id=2987874145) because of that then.

dakshvar22: Should we do - 

```
try:
    from typing import TypedDict
except ImportError:
    from typing_extensions import TypedDict
```
to keep support for python 3.6 and 3.7

dakshvar22: Hmm, I used quotes but type checks are still [failing](https://github.com/RasaHQ/rasa/pull/9016/checks?check_run_id=2988025301#step:11:16) 😞

wochinge: > If that's the case, then you can't import it inside an if statement. You always want it to be imported somehow since the code needs to initialize a TypedDict and it will result in a runtime error if the TypedDict is never imported.

You also need to put the `TypedDict` stuff in the `if`. See [here](https://github.com/RasaHQ/rasa/blob/8db9030bf7ec457e9306c95cbf7fbe811771d05f/rasa/core/test.py#L66)

wochinge (approved):
🙌🏻

PR #9046, opened 2021-07-06
8614 nlu test autoconfig

### LINKED ISSUE #8614: rasa test nlu cross validation fails with autoconfiguration
**Rasa version**: master

**Rasa SDK version** (if used & relevant): -

**Rasa X version** (if used & relevant): -

**Python version**: 3.6

**Issue**:
* create a config using the autoconfiguration feature (running `rasa init` should get you a failing config):
```
# Configuration for Rasa NLU.
# https://rasa.com/docs/rasa/nlu/components/
language: en

pipeline:
# # No configuration for the NLU pipeline was provided. The following default pipeline was used to train your model.
# # If you'd like to customize it, uncomment and adjust the pipeline.
# # See https://rasa.com/docs/rasa/tuning-your-model for more information.
#   - name: WhitespaceTokenizer
#   - name: RegexFeaturizer
#   - name: LexicalSyntacticFeaturizer
#   - name: CountVectorsFeaturizer
#   - name: CountVectorsFeaturizer
#     analyzer: char_wb
#     min_ngram: 1
#     max_ngram: 4
#   - name: DIETClassifier
#     epochs: 100
#     constrain_similarities: true
#   - name: EntitySynonymMapper
#   - name: ResponseSelector
#     epochs: 100
#     constrain_similarities: true
#   - name: FallbackClassifier
#     threshold: 0.3
#     ambiguity_threshold: 0.1
```
* run `rasa test nlu --cross-validation`

```
2021-05-05 18:40:50 INFO     rasa.cli.test  - Test model using cross validation.
InvalidConfigException: Can not train an empty pipeline. Make sure to specify a proper pipeline in the configuration using the 'pipeline' key.
```

Remove the comments (and hence not using autoconfig anymore) leads to the validation working just fine.

**Definition of Done**
- [ ] `rasa test nlu` uses autoconfig to interpolate config files with empty `pipeline` keys (e.g. by using the `TrainingDataImporter`)

twerkmeister:
**Proposed changes**:
- fixes #8614 
- nlu cross validation uses autoconfig when the provided config does not contain a pipeline (as after using `rasa init`)

**Further thoughts**
- I don't like the interface of the function chain from cli invocation to actual cross validation. The `all_args` parameter that is passed along and changed is very difficult to track. Also it has even the named parameters as duplicates in it.


rasa/model_testing.py
```diff
@@ -313,7 +314,10 @@ def perform_nlu_cross_validation(
 
     additional_arguments = additional_arguments or {}
     folds = int(additional_arguments.get("folds", 3))
-    nlu_config = rasa.nlu.config.load(config)
+
+    nlu_config = rasa.shared.importers.autoconfig.get_configuration(
```
joejuzl, line 318: I assume somewhere up the chain we use a `RasaFileImporter`. Would it possible to get the config from there, as I think that calls this method at some point?

tests/test_model_testing.py
```diff
@@ -387,3 +388,33 @@ async def test_test_does_not_use_rules(tmp_path: Path, default_agent: Agent):
...
+    config_file.write_text(config)
+    training_data_file = "data/test_moodbot/data/nlu.yml"
+    domain_file = "data/test_moodbot/domain.yml"
+
+    test_data_importer = TrainingDataImporter.load_from_dict(
+        training_data_paths=[training_data_file], domain_path=domain_file,
+    )
+    nlu_data = await test_data_importer.get_nlu_data()
+
+    # start cross validation in an outside process to monitor if it fails
```
joejuzl, line 412: What happens if you call it in the same process?

joejuzl (changes requested):

rasa/model_testing.py
twerkmeister: I changed to using TrainingDataImporter but had to change the interface of `perform_nlu_cross_validation`

tests/test_model_testing.py
twerkmeister: Well, the thing is we want to see if it trains using autoconfig. However, autoconfig has 100 epochs as a default value. That means it will run really long. To not waste time, I stop the process after the first couple seconds. Before this fix, by then a InvalidConfigException would have been raised. Does this make sense?

twerkmeister: The test is a bit different now, as I am calling rasa test directly, but the concept is the same

rasa/cli/test.py
```diff
@@ -202,7 +202,10 @@ async def run_nlu_test_async(
         config = rasa.cli.utils.get_validated_path(
             config, "config", DEFAULT_CONFIG_PATH
         )
-        perform_nlu_cross_validation(config, nlu_data, output, all_args)
+        config_importer = TrainingDataImporter.load_from_dict(config_path=config)
```
joejuzl, line 205: can we not use the `test_data_importer` from above?

joejuzl (changes requested):

rasa/cli/test.py
twerkmeister: that is not possible as at the time that importer is instantiated it is still unclear whether config is one particular config or multiple for model comparison mode. The TrainingDataImporter, however, needs a single one.

tests/cli/test_rasa_test.py
```diff
@@ -107,6 +107,35 @@ def test_test_nlu_cross_validation(run_in_simple_project: Callable[..., RunResul
...
+        "-c",
+        "config.yml",
+        "--nlu",
+        "nlu.yml",
+    ]
+
+    # we don't wanna run the cross validation for real, just want to see that it does
+    # not crash
+    try:
+        testdir.run(*args, timeout=10.0)
```
joejuzl, line 134: Does this mean the test will take 10 seconds? And if so can we make it a bit quicker?

joejuzl (changes requested):

tests/cli/test_rasa_test.py
twerkmeister: I reduced it to 8 seconds. On my machine anything less than 5 seconds was too little to catch the problem - Unfortunately, it invokes the rasa cli which takes a long time to start up. The solution I had earlier was able to do it in 3. That just had the small side effect that we used that autoconfig system directly, which also wrote back the default config to the designated config file. But there I also didn't have to change the signature of the cross eval method. So idk what's better. Almost see the first solution as slightly better. What do you think?

joejuzl: I think this way is safer as it keeps the auto config stuff encapsulated.

joejuzl (approved):

PR #9055, opened 2021-07-06
UnexpecTEDIntentPolicy: Deduplicate multiple data points with same input state

dakshvar22:
**Proposed changes**:
- Remove the need for creating multiple data points with duplicate input state but permuted version of positive label ids.


dakshvar22:
@kedz @JEM-Mosig PR ready for review :)

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -1021,7 +1017,7 @@ def training_states_labels_and_entities(
         # Mapping of example state hash to list of positive labels associated with
         # the state. Note that each individual 'label' instance is a list of ints.
```
JEM-Mosig, line 1018: In the comment it says "list of ints", but the type annotation says "List[List[Text]]".
```suggestion
        # the state. Note that each individual 'label' instance is a list of strings.
```

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -1059,26 +1055,41 @@ def training_states_labels_and_entities(
...
+        for example_state, example_entity in zip(example_states, example_entities):
+            state_hash = self._hash_example(example_state)
+            if self.remove_duplicates and state_hash in seen_states:
+                continue
+            pruned_example_states.append(example_state)
+            pruned_example_labels.append(
+                list(
+                    set(
+                        [
+                            labels[0]
```
JEM-Mosig, line 1076: Can you please add a comment explaining why `[0]`?

JEM-Mosig (approved):

rasa/core/featurizers/tracker_featurizers.py
dakshvar22: Done

dakshvar22:
@kedz I've addressed Johannes' comments but will wait for an approval from you too before I merge.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -1044,13 +1040,13 @@ def training_states_labels_and_entities(
...
                     if hashed in hashed_examples:
                         continue
                     hashed_examples.add(hashed)
 
                 # Store all positive labels associated with a training state.
-                state_hash = self._hash_example(tracker, states)
+                state_hash = self._hash_example(states)
                 state_hash_to_label_list_instances[state_hash].append(label)
 
                 example_states.append(states)
```
kedz, line 1052:
```suggestion

                if self.remove_duplicates:
                    hashed = self._hash_example(states, label)
                    if hashed in hashed_examples:
                        continue
                    hashed_examples.add(hashed)

                # Store all positive labels associated with a training state.
                state_hash = self._hash_example(states)
                state_hash_to_label_list_instances[state_hash].add(label[0])

                # Only add unique example states unless `remove_duplicates` is `False`.
                if not self.remove_duplicates or state_hash not in seen_states:
                    example_states.append(states)
                    example_entities.append(entities)
                    seen_states.add(state_hash)

                pbar.set_postfix({f"# {self.LABEL_NAME}": f"{len(example_states):d}"})

        # Collect positive labels for each state example.
        example_labels = [
            list(state_hash_to_label_list_instances[self._hash_example(state)])
            for state in example_states
        ]

        self._remove_user_text_if_intent(example_states)

        logger.debug(
            f"Created {len(example_states)} {self.LABEL_NAME} examples."
        )

        return example_states, example_labels, example_entities
```

(I can't extend the suggestion after line 1056, but this would replace everything from 1041 on in the function.)

What you have from lines 1058 on works, but IMO it seems wasteful of resources to (a) create both a large list of possibly redundant data and then a pruned list when we could just create the pruned list and (b) create `example_labels` when it isn't being used for anything.

The above suggestion would create pruned example states in one pass and simplify the creation of the pruned example labels later. Not necessary to do but I thought if we have a little more time since we have a release blocking bug we could simplify this code a bit.

This would also mean moving the creation of `seen_states` up to line 1022 and converting state_hash_to_label_list_instances to hold a set instead of a list of length 1 lists (and a name change).

If you don't do this suggestion, I still think you should remove `example_labels` since its not doing anything.

rasa/utils/tensorflow/layers_utils.py
```diff
@@ -71,14 +71,24 @@ def get_candidate_values(
...
     Returns:
         The mean of "x == y"
     """
-    return tf.reduce_mean(tf.cast(tf.math.equal(x, y), tf.float32))
+    if mask is None:
+        return tf.reduce_mean(tf.cast(tf.math.equal(x, y), tf.float32))
+    else:
+        equal_indices = tf.cast(tf.math.equal(x, y), tf.float32) * mask
+        return tf.reduce_mean(
+            tf.reduce_sum(equal_indices, axis=-1) / tf.reduce_sum(mask, axis=-1)
```
kedz, line 93: Why is this only reduced along the last axis when a mask is passed?

tests/test_memory_leak.py
```diff
@@ -156,7 +156,7 @@ def epochs(self) -> int:
 
     @property
     def max_memory_threshold_mb(self) -> float:
-        return 1200
+        return 1700
```
kedz, line 159: Shouldn't we be using less memory now? Or is this unrelated to deduplication of states?

rasa/utils/tensorflow/layers_utils.py
dakshvar22: Hmm, good question. There is an assumption that the actual mask values are on the last dimension and hence the first `reduce_mean` is needed only on that axis. The next `reduce_mean` is across all axis.

tests/test_memory_leak.py
dakshvar22: This is unrelated. This creeped into the PR from the main PR.

rasa/utils/tensorflow/layers_utils.py
kedz, line 93: As advertised in the docstring and name, this function seems agnostic to the dimensions, yet when a mask is passed in it suddenly treats the last dimension as the one of concern and makes everything else a batch dimension over which it then averages. This gives counter intuitive behavior. Based on the behavior of the unmasked case, I would expect
```python
x = [[1,2,3,4]
     [5,6,7,8]] 
y = [[1,2,3,4]
     [5,6,0,0]]
mask = [[1,1,1,1],
        [1,1,1,0]]
```
to give `6 / 7` since there are 6 total matches on 7 unmasked items.
What is currently computed however is  `((4/4) + 2/3) / 2`
             
Perhaps we should add an axis argument that gets passed so you can specify the desired behavior explicitly similar to the underlying tensorflow ops `reduce_mean` and `reduce_sum` upon which this operation is based?

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -1013,16 +1009,15 @@ def training_states_labels_and_entities(
...
         hashed_examples = set()
         # Mapping of example state hash to list of positive labels associated with
-        # the state. Note that each individual 'label' instance is a list of ints.
-        state_hash_to_label_list_instances: defaultdict[
-            int, List[List[int]]
-        ] = defaultdict(list)
+        # the state. Note that each individual 'label' instance is a list of strings.
+        state_hash_to_label_set: defaultdict[int, Set[Text]] = defaultdict(set)
+
+        seen_states = set()
```
kedz, line 1020:
```suggestion
        # Store of example hashes (of both states and labels) for removing 
        # duplicate training examples.
        hashed_examples = set()
        # Mapping of example state hash to set of positive labels associated with
        # the state. Note that each individual 'label' instance is a set of strings.
        state_hash_to_label_set: defaultdict[int, Set[Text]] = defaultdict(set)
        # Store of example state hashes so only unique states are kept when 
        # removing duplicates.
        seen_states = set()
```

Added a comment for `seen_states` and modified comments of `state_hash_to_label_set`.

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -1044,37 +1039,30 @@ def training_states_labels_and_entities(
...
+                state_hash_to_label_set[state_hash].add(label[0])
 
-                example_states.append(states)
-                example_labels.append(label)
-                example_entities.append(entities)
+                # Only add unique example states unless `remove_duplicates` is `False`.
+                if not self.remove_duplicates or state_hash not in seen_states:
+                    example_states.append(states)
+                    example_entities.append(entities)
+                    seen_states.add(state_hash)
```
kedz, line 1055:
```suggestion
                state_hash = self._hash_example(states)

                # Only add unique example states unless `remove_duplicates` is `False`.
                if not self.remove_duplicates or state_hash not in state_hash_to_label_set:
                    example_states.append(states)
                    example_entities.append(entities)

                # Store all positive labels associated with a training state.
                state_hash_to_label_set[state_hash].add(label[0])
```
Sorry last comment on this section I promise :) . I just realized we could get rid of `seen_states` completely since we are already keeping the same information in `state_hash_to_label_set`.

kedz (changes requested):
Tracker featurization code is looking good -- added a suggestion for comment fix and an optional tweak.

Have some questions/concerns about the some of the tensorflow helper functions though.

rasa/core/featurizers/tracker_featurizers.py
dakshvar22: Omitting ```Note that each individual 'label' instance is a set of strings.``` since that just mentions the data type which is already known by the type annotation.

rasa/utils/tensorflow/layers_utils.py
dakshvar22: I clarified it in the docstrings for now. We should give it a thought on how to make it more general purpose in the future (adding the axis argument).

dakshvar22:
/modeltest

```yml

include:
 - dataset: ["all"]
   config: ["all"]

```

rasa/utils/tensorflow/layers_utils.py
kedz: I'll create an issue for it.

kedz (approved):
Looks good!

PR #9062, opened 2021-07-07
fixed mypy operator error type

### LINKED ISSUE #9058: mypy: Fix `operator`

Current number of issues with this type: 44

**Definition of Done**:
- [x] enable `mypy` checks for `operator`
- [x] all existing `operator` issues are fixed

twerkmeister:
**Proposed changes**:
- fixes #9058 

- [x] enable mypy checks for operator
- [x] all existing operator issues are fixed


rasa/core/policies/rule_policy.py
```diff
@@ -928,35 +927,34 @@ def _find_action_from_loop_happy_path(
...
-
-        if should_predict_loop:
-            logger.debug(f"Predicted loop '{active_loop_name}'.")
-            return active_loop_name, LOOP_RULES + active_loop_name
-
-        # predict `action_listen` if loop action was run successfully
-        if should_predict_listen:
-            logger.debug(
-                f"Predicted '{ACTION_LISTEN_NAME}' after loop '{active_loop_name}'."
+        if active_loop_name:
```
wochinge, line 930: optional: should use guards to reduce the indent?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1162,7 +1166,7 @@ def _load_model(
         model_dir: Text,
         finetune_mode: bool = False,
     ) -> "RasaModel":
-        file_name = meta.get("file")
+        file_name = meta["file"]
```
wochinge, line 1169: should we raise a more specific error in case `file` is not in `meta`?

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -172,22 +172,13 @@ def _check_attribute_vocabulary(self, attribute: Text) -> bool:
         except (AttributeError, KeyError):
             return False
 
-    def _get_attribute_vocabulary(self, attribute: Text) -> Optional[Dict[Text, int]]:
-        """Get trained vocabulary from attribute's count vectorizer"""
+    def _get_attribute_vocabulary(self, attribute: Text) -> Dict[Text, int]:
+        """Get trained vocabulary from attribute's count vectorizer."""
+        return self.vectorizers[attribute].vocabulary_
```
wochinge, line 177: can't the attribute be not in `vectorisers`?

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -718,14 +696,11 @@ def process(self, message: Message, **kwargs: Any) -> None:
...
     def _collect_vectorizer_vocabularies(self) -> Dict[Text, Optional[Dict[Text, int]]]:
-        """Get vocabulary for all attributes"""
-
-        attribute_vocabularies = {}
-        for attribute in self._attributes:
-            attribute_vocabularies[attribute] = self._get_attribute_vocabulary(
-                attribute
-            )
-        return attribute_vocabularies
+        """Get vocabulary for all attributes."""
```
wochinge, line 699:
```suggestion
        """Gets vocabulary for all attributes."""
```

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -602,19 +593,6 @@ def _create_features(
...
         return sequence_features, sentence_features
 
-    def _get_featurized_attribute(
-        self, attribute: Text, all_tokens: List[List[Text]]
-    ) -> Tuple[
-        List[Optional[scipy.sparse.spmatrix]], List[Optional[scipy.sparse.spmatrix]]
-    ]:
-        """Return features of a particular attribute for complete data"""
-
-        if self._check_attribute_vocabulary(attribute):
```
wochinge, line 612: why can we omit this check?

rasa/nlu/utils/mitie_utils.py
```diff
@@ -65,7 +65,7 @@ def cache_key(
 
         mitie_file = component_meta.get("model", None)
         if mitie_file is not None:
-            return cls.name + "-" + str(os.path.abspath(mitie_file))
+            return str(cls.name) + "-" + str(os.path.abspath(mitie_file))
```
wochinge, line 68: optional: How about using a f-string to avoid the casts?

rasa/nlu/utils/spacy_utils.py
```diff
@@ -89,7 +89,7 @@ def cache_key(
             component_meta.get("model"), model_metadata.language, warn=False
         )
 
-        return cls.name + "-" + spacy_model_name
+        return str(cls.name) + "-" + spacy_model_name
```
wochinge, line 92: optional: should we use f-strings instead?

rasa/utils/endpoints.py
```diff
@@ -73,18 +73,19 @@ class EndpointConfig:
...
-        basic_auth: Dict[Text, Text] = None,
+        url: Text = "127.0.0.1",
+        params: Optional[Dict[Text, Any]] = None,
+        headers: Optional[Dict[Text, Any]] = None,
+        basic_auth: Optional[Dict[Text, Text]] = None,
         token: Optional[Text] = None,
         token_name: Text = "token",
         **kwargs: Any,
     ) -> None:
+        """Create an EndpointConfig instance."""
```
wochinge, line 84:
```suggestion
        """Creates an `EndpointConfig` instance."""
```

wochinge (commented):
Thanks for fixing so many `mypy` errors 🎉

rasa/core/policies/rule_policy.py
```diff
@@ -907,18 +907,17 @@ def _find_action_from_default_actions(
...
-        if default_action_name:
-            logger.debug(f"Predicted default action '{default_action_name}'.")
-            return (
-                default_action_name,
-                # create prediction source that corresponds to one of
-                # default prediction sources in `_default_sources()`
-                DEFAULT_RULES + tracker.latest_message.intent.get(INTENT_NAME_KEY),
-            )
+        intent_name = tracker.latest_message.intent.get(INTENT_NAME_KEY)
+        if intent_name:
```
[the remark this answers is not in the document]
twerkmeister: Good thinking!

rasa/nlu/classifiers/diet_classifier.py
twerkmeister: Something like Rasa Error - Corrupt Metadata or so?

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
twerkmeister: My philosophy here was that if the code is used in the wrong manner, I'd let it crash early - instead of half-heartedly returning `None` and then not knowing what to do with that or masking the error ...  🤔

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
twerkmeister: Made sense in the moment, will check again

twerkmeister: If you look into the code that calls this, you'll see that this is only called for an already trained model. The model is explicitly trained for the given attributes. I think it is really wrong here to just return an empty list for something that the user intends to be there. I think that's error obfuscation. What do you think?

twerkmeister: but seems there is still a failing test with this change ... gotta investigate :)

twerkmeister: Well good that we have these tests, I was wrong, the attributes are there in a very general fashion and if there's no data for them you just get empty vectorizers. Found a much simpler change too :)

rasa/nlu/classifiers/diet_classifier.py
twerkmeister: What do you think of the new proposal?

twerkmeister: super weird that github doesn't seem to update this preview in this case... Never seen that before
here's the current code: 
https://github.com/RasaHQ/rasa/blob/97cae315c75d27c4a813bc8cf7719c26d09a39f8/rasa/nlu/classifiers/diet_classifier.py#L1163-L1178

rasa/core/policies/ted_policy.py
```diff
@@ -199,7 +199,7 @@ class TEDPolicy(Policy):
         # If 'True' use value relative embeddings in attention
         VALUE_RELATIVE_ATTENTION: False,
         # Max position for relative embeddings
-        MAX_RELATIVE_POSITION: None,
+        MAX_RELATIVE_POSITION: 5,
```
wochinge, line 202: does this mean we can also remove [this check](https://github.com/RasaHQ/rasa/blob/b54f901f140510a8d20aadf8c6f3a415be22ec6a/rasa/utils/tensorflow/transformer.py#L59) ?

wochinge: I'm pretty sure the change is correct but I think this require's some double checking

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1162,7 +1169,12 @@ def _load_model(
         model_dir: Text,
         finetune_mode: bool = False,
     ) -> "RasaModel":
-        file_name = meta.get("file")
+        if "file" not in meta:
+            raise CorruptMetaDataException(
+                f"Metadata for trained component {cls.name}"
```
wochinge, line 1174:
```suggestion
                f"Metadata for trained component {cls.name} "
```

rasa/nlu/classifiers/diet_classifier.py
wochinge: haha, works for me 👍🏻 Looks good!

rasa/nlu/utils/mitie_utils.py
```diff
@@ -65,7 +65,7 @@ def cache_key(
 
         mitie_file = component_meta.get("model", None)
         if mitie_file is not None:
-            return cls.name + "-" + str(os.path.abspath(mitie_file))
+            return f"{cls.name}-{os.path.abspath(mitie_file)}"
```
wochinge, line 68: 🙌🏻

rasa/utils/endpoints.py
```diff
@@ -73,18 +73,19 @@ class EndpointConfig:
...
+        url: Optional[Text] = None,
+        params: Optional[Dict[Text, Any]] = None,
+        headers: Optional[Dict[Text, Any]] = None,
+        basic_auth: Optional[Dict[Text, Text]] = None,
         token: Optional[Text] = None,
         token_name: Text = "token",
         **kwargs: Any,
     ) -> None:
+        """Creates an `EndpointConfig` instance."""
         self.url = url
```
wochinge, line 85: should we immediately raise if this is `None`?

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -275,18 +258,14 @@ def _process_tokens(self, tokens: List[Text], attribute: Text = TEXT) -> List[Te
...
+        """Replace OOV words with OOV token."""
         if self.OOV_token and self.analyzer == "word":
-            vocabulary_exists = self._check_attribute_vocabulary(attribute)
-            if vocabulary_exists and self.OOV_token in self._get_attribute_vocabulary(
-                attribute
-            ):
+            attribute_vocab = self._get_attribute_vocabulary(attribute)
+            if attribute_vocab is not None and self.OOV_token in attribute_vocab:
                 # CountVectorizer is trained, process for prediction
+                attribute_vocabulary_tokens = list(attribute_vocab.keys())
```
wochinge, line 266: we can skip the `list` call, no?
```suggestion
                attribute_vocabulary_tokens = attribute_vocab.keys()
```

wochinge (commented):
Looking very good and also simplifies the code! 💯

koernerfelicia:
Hi @twerkmeister, got back today so am just looking at this for the first time now. Could you let me know how I would check that the warnings for `operator` are all gone? I ran `make types` but this raises no issues here or on `main`.

wochinge:
>  I ran make types but this raises no issues here or on main.

`operator` issues are suppressed on `main` and hence raise no error there (you can edit remove it from the ignored section in `setup.cfg` as Thomas has done in this PR) to see all errors.

And I guess they are fixed on this branch which is why you can't see them here.

rasa/core/policies/ted_policy.py
koernerfelicia: My understanding is that strictly speaking, we only need this value at all if either: `self.use_key_relative_position` or `self.use_value_relative_position` are true (see [here](https://github.com/RasaHQ/rasa/blob/b54f901f140510a8d20aadf8c6f3a415be22ec6a/rasa/utils/tensorflow/transformer.py#L82-L118)).

I was concerned that setting it to a default value (rather than `None`) might mislead people to believe it will always be relevant (and used)... but we're not very disciplined about this in the rest of the code either (e.g. we call `self._create_relative_embeddings()` even though this doesn't apply if neither of the above booleans are true).

That's obviously out of scope for this, all that to say maybe the correct thing to do would be to create another issue around cleaning this up:
- making `MAX_RELATIVE_POSITION` optional, but required when either: `self.use_key_relative_position` or `self.use_value_relative_position`
Moving things like incrementing of `MAX_RELATIVE_POSITION` into parts of the code that can only be reached if the above conditions (one of boolean true and value set) are met.

What do you think? I am happy to create and put on research board.

rasa/core/channels/console.py
```diff
@@ -60,14 +60,14 @@ def print_bot_output(
             return question
 
     if "text" in message:
-        rasa.shared.utils.cli.print_color(message.get("text"), color=color)
+        rasa.shared.utils.cli.print_color(message["text"], color=color)
```
koernerfelicia, line 63: How come we switch to subscript (rather than `get`) here but not in the loops below (for `elements`, `quick_replies`, and `custom`)?

koernerfelicia: Also, in `format_bot_output` (`rasa/core/training/interactive.py`) we check that the values are not `None`. Should we be doing this here, too?

rasa/nlu/classifiers/diet_classifier.py
koernerfelicia: Think we need this space (or one on the line below), else the message will read: `"Metadata for trained component {cls.name}is missing the 'file' attribute."`

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
```diff
@@ -248,7 +248,7 @@ def load(
                 finetuning.
             **kwargs: Any other arguments.
         """
-        file_name = meta.get("file")
+        file_name = meta["file"]
```
koernerfelicia, line 251: Should we also raise an error if file is not in the metadata (similar to [here](https://github.com/RasaHQ/rasa/pull/9062/files#diff-b75ba088655bba2a10f267458c7ce01a1d2b9d69e4447f3f548bbf00db66d208R1172-R1177))?

koernerfelicia (commented):
Love the cleanup 😍 Had a few small comments, otherwise looks quite good!

rasa/core/policies/ted_policy.py
twerkmeister: > does this mean we can also remove [this check](https://github.com/RasaHQ/rasa/blob/b54f901f140510a8d20aadf8c6f3a415be22ec6a/rasa/utils/tensorflow/transformer.py#L59) ?

Yes good point, will remove

twerkmeister: @koernerfelicia there was a discussion in the research channel with on this and we settled that it would be fine to set it to 5 even though this value should not be seen as an absolute endorsement, just like all the other hyperparameters parameters. I don't think an invalid default value makes much sense tbh. If you turn on key or value relative position why should the training crash. But obviously you should think about your hyperparameters.

twerkmeister: Hmm, just read your comment again, and understand that you also bring up a new aspect here. Users that seeing that `max_relative_position` is set to a specific value and thinking that this is what's happening. I'd be relatively comfortable taking that risk as this config parameter is definitely buried deep either in the code or the documentation. So if people go there, they are definitely more involved/advanced. They might then also see the other two relevant variables.
I think to decrease chance of user failure, I would improve the comment, saying that this is only used if either key or value relative position are turned on. What to you think @koernerfelicia?

rasa/utils/endpoints.py
twerkmeister: Apparently there's code that depends on the current behavior. Things failed when I changed the default before. So, not sure it's worth digging to deep here rn 🤔

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
twerkmeister: hm not sure what the runtime of element search is in the `dict_keys` class, but might actually make sense to change it to set!?

rasa/core/channels/console.py
twerkmeister: Fair point. I think this is due to `elements`, `quick_replies` and `custom` aren't operator issues, as they are wrapped in enumerate calls. I could change it too, or we could say it's out of scope for this issue. I guess it would come up again in another mypy issue. 🤔

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
twerkmeister: I suppose we could do that for every component. But then we also need tests for this otherwise we'll run into coverage problems. And I am thinking it's kind of a niche scenario... I mean to get there: the model zip is still intact, the overall json structure of the metadata is still intact, buuuuut the file key disappeared. I don't think that is really likely absent of willful user manipulation.

twerkmeister: Right now I am more in favor of dropping that new exception class again XD

twerkmeister:
> Love the cleanup 😍 Had a few small comments, otherwise looks quite good!

Thanks a lot for the comments :)

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
twerkmeister: @wochinge what do you think?

rasa/core/channels/console.py
twerkmeister: changed it :)

rasa/core/policies/ted_policy.py
koernerfelicia: Yes, your second interpretation is what I was trying to convey (_"Users that seeing that max_relative_position is set to a specific value and thinking that this is what's happening"_). To be clear: I'm not concerned about endorsing 5 as a default value, and I definitely agree that we should not crash. I just wouldn't want the default value being set to suggest to the user that the behaviour is also the default. To that end a comment would definitely work!

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
koernerfelicia: How come you added it for diet? Is it more likely there? (genuine question). I'm generally in favour of the symmetry, but understand how that blows up the scope, so I trust your judgment on this one 🤷‍♀️

twerkmeister: Yeah it's true it should be consistent. I added it there because Tobias saw it there and I/we didn't realize that it would be the same for all the other components as well. Now I am wondering whether it is a good idea or not

twerkmeister: I think I'll create a follow up issue for testing the consistency of metadata dicts - I think this also needs a bit of thought, e.g. whether we we would introduce a more specific type or not for the metadata. Wouldn't want to just produce a lot of code as a quick fix now.

twerkmeister: I created the issue and also changed the improper uses of get in the component load classes. Sometimes these usages would just push the error further back and show up as a `None` in a method that should have gotten the file name string. Now things crash directly where the problem is: that `file` is not in the metadata.

koernerfelicia: Sounds good to me. Thanks for creating the issue!

m-vdb:
@twerkmeister is this one ready for re-review? not sure if you know, you can re-request reviews by clicking on that icon 🔁 
Screenshot 2021-07-23 at 11 36 26

wochinge (approved):
🚀

PR #9074, opened 2021-07-08
wip (single state featurizer)

JEM-Mosig:
**Proposed changes**:
- ...


rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -30,6 +33,65 @@
...
+
+
+def convert_sparse_sequence_to_sentence_features(
+    features: List["Features"],
+) -> List["Features"]:
+    """Grabs all sparse sequence features and turns them into sparse sentence features.
+
+    That is, this function filters all sparse sequence features and then
+    turns each of these sparse sequence features into sparse sentence feature
+    by summing over their first axis, respectively.
```
JEM-Mosig, line 77: This doesn't always work. For example, the "natural" operation to combine semantic map embeddings is not a sum.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
-        self._use_regex_interpreter = False
-        self._default_feature_states = {}
-        self.action_texts = []
-        self.entity_tag_specs = []
-
-    def _create_entity_tag_specs(
-        self, bilou_tagging: bool = False
-    ) -> List[EntityTagSpec]:
-        """Returns the tag to index mapping for entities.
+        self.prepared = False
```
JEM-Mosig, line 107: Somehow, we need to get rid of the need to "prepare". The vocabulary build up should be done during initialization.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
-            }
-        else:
-            tag_id_index_mapping = {
-                tag: idx + 1  # +1 to keep 0 for the NO_ENTITY_TAG
-                for tag, idx in self._default_feature_states[ENTITIES].items()
-            }
+        # FIXME/TODO:
+        # 1. Would be cool if users would be notified that they're not training
+        #    the policy based on what their NLU pipeline returns.... (cf.
+        #    featurize_via_interpreter)
```
JEM-Mosig, line 159: Agreed

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
+        # TODO: the following this was if...: = True before, is this a problem?
+        self.use_regex_interpreter = isinstance(interpreter, RegexInterpreter)
+        self.item2index_mappings = {
+            key: item2index_mapping(items)
+            for key, items in [
+                (INTENT, domain.intents),
+                (ACTION_NAME, domain.action_names_or_texts),
+                (ENTITIES, domain.entity_states),
+                (SLOTS, domain.slot_states),
+                (ACTIVE_LOOP, domain.form_names),
```
JEM-Mosig, line 132: I think we should rename `entity_states` to `entity_feature_names` and similar for `slot_states`.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
-            )
-        ]
+            # nlu pipeline didn't create features for user or action
+            # this might happen, for example, when we have action_name in the state
+            # but it did not get featurized because only character level
+            # CountVectorsFeaturizer was included in the config.
+            name_attribute = get_name_attribute(sub_state)
+
+            if substate_type == PREVIOUS_ACTION:
+                substate_features = self.featurize_substate_via_interpreter(
```
JEM-Mosig, line 175:
```suggestion
                substate_features = self._featurize_substate_via_interpreter(
```

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
+        return state_features
+
+    def featurize_action(
+        self, action: Text, interpreter: NaturalLanguageInterpreter
+    ) -> Dict[Text, List["Features"]]:
+        """
+
+        """
+        # FIXME: if prepare_for_training has not been called then action_texts will
+        # be an empty list and this always defaults to "ACTION_NAME" - is this intended?
```
JEM-Mosig, line 222: We need to get rid of the `prepare_for_training` and put it into the `__init__` somehow.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
-        """Gets necessary information for featurization from domain.
+        exclude: Optional[List[Text]] = None,
+        keep_only_sparse_seq_converted_to_sent: Optional[List[Text]] = None,
+        fallback_to_custom_encoding: Optional[List[Text]] = None,
+        sparse: bool = False,
+    ) -> Dict[Text, List["Features"]]:
+        """Featurize all attributes associated with a sub_state.
+
+
+        # NOTE: this was `_extract_state_features`
```
JEM-Mosig, line 242: Good name change 👍

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
 
-    def _create_features(
+            for features in all_features:
+                if features is not None:
+                    attribute_to_features[attribute].append(features)
+
+        attribute_list = keep_only_sparse_seq_converted_to_sent or []
+        for attribute in attribute_list:
+            if attribute_to_features.get(attribute):
+                converted_features = convert_sparse_sequence_to_sentence_features(
```
JEM-Mosig, line 288: As mentioned above, it would be good to have a bit more flexibility here (not just `sum`).

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -40,127 +102,227 @@ class SingleStateFeaturizer:
...
+    def _featurize_attribute_via_custom_encoding(
         self, sub_state: SubState, attribute: Text, sparse: bool = False
     ) -> List["Features"]:
-        state_features = self._state_features_for_attribute(sub_state, attribute)
+        """Creates a multi-hot encoding like feature for the given attribute.
+
+        Leverages the item2index mappings computed during `prepare_training`.
+
+        FIXME/TODO: would've expected a one-hot encoding type of thing, but for
+          SLOTS the values are >1 ...(?!?)
```
JEM-Mosig, line 310: Slots can have custom encodings that depend the user's implementation. See [this](https://rasa.com/docs/rasa/domain#custom-slot-types).

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -172,137 +334,158 @@ def _create_features(
...
-            return {}
+        # NOTE: this was `_state_features_for_attribute`
 
-        output = defaultdict(list)
-        for attribute in attributes:
-            all_features = parsed_message.get_sparse_features(
-                attribute
-            ) + parsed_message.get_dense_features(attribute)
+        """
+        # FIXME: Create dataclasses for states, substates, messages,...
```
JEM-Mosig, line 343: I agree

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -172,137 +334,158 @@ def _create_features(
...
-        # if features for INTENT or ACTION_NAME exist,
-        # they are always sparse sequence features;
-        # transform them to sentence sparse features
-        if output.get(INTENT):
-            output[INTENT] = self._to_sparse_sentence_features(output[INTENT])
-        if output.get(ACTION_NAME):
-            output[ACTION_NAME] = self._to_sparse_sentence_features(output[ACTION_NAME])
 
-        return output
+class BinarySingleStateFeaturizer(SingleStateFeaturizer):
```
JEM-Mosig, line 368: We don't need to bother with this one, as it will be removed in 3.0 anyway.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -172,137 +334,158 @@ def _create_features(
...
+        name_attribute = get_name_attribute(sub_state)
+        if name_attribute:
+            return {
+                name_attribute: self._create_features(sub_state, name_attribute, sparse)
+            }
+
+        return {}
+
+
+class LabelTokenizerSingleStateFeaturizer(SingleStateFeaturizer):
```
JEM-Mosig, line 413: Same here. Deprecated.

rasa/core/featurizers/single_state_featurizer.py
```diff
@@ -172,137 +334,158 @@ def _create_features(
...
+# - The following seems to be completely separate from the SingleStateFeaturizer
+# - it implements a very similar yet different featurization
 
-        return output
 
-    def encode_state(
-        self, state: State, interpreter: NaturalLanguageInterpreter
-    ) -> Dict[Text, List["Features"]]:
-        """Encode the given state with the help of the given interpreter.
+class EntityFeaturizer:
```
JEM-Mosig, line 431: Not sure I follow. Why do you want to implement a separate class here?

JEM-Mosig (commented):
I had a first read-through. Its a good start I think!

rasa/core/featurizers/single_state_featurizer.py
ka-bu: It just helped me understand that the functionality that follows is completely independent of the rest (at least the functions itself). The logic seems to overlap and might be merged/solved with some of the above. No?

rasa/core/featurizers/single_state_featurizer.py
ka-bu: We should maybe start a notion page for those ... to keep links to places where those are used and in which way to figure out how the interfaces should look like?

rasa/core/featurizers/single_state_featurizer.py
ka-bu: Oh I see, and you do not know those values in advance and hence cannot map them to categorical "slot_value" type of things. I think it would be good to then clearly point this out somewhere (for us who build algorithms with this and might not expect that they need to scale the values and can squeeze them into some combination layer nonetheless :/).

rasa/core/featurizers/single_state_featurizer.py
ka-bu: agree!

rasa/core/featurizers/single_state_featurizer.py
ka-bu: :)

rasa/core/featurizers/single_state_featurizer.py
ka-bu: I think that is a good idea and I think there are multiple places we could reuse that (cf. my funny countvectorizer related questions :)

rasa/core/featurizers/single_state_featurizer.py
ka-bu: yes, agree! Maybe we turn part of that notion page into some keyword index (might also help to trace back State vs Message vs UtterEvent vs...)

rasa/core/featurizers/single_state_featurizer.py
ka-bu: Since this is something that depends on the kind of Features and their origin (which means this will not make separate state featurizer classes necessary), I think we'll leave this in here as TODO (for a future Feature rework or so?)

rasa/core/featurizers/single_state_featurizer.py
ka-bu: I was really confused whether the wrong interpreter might be used by accident. But I guess the reason there's no mismatch (in practice now) is in the .persist of the TrackerFeaturizer which will pickles this StateFeaturizer as a wholeand preserve the "use_regex..." flag.

rasa/core/featurizers/single_state_featurizer.py
ka-bu: Hm, just realised, if we move that to the init only, then when we reload this component (after persisting it) it won't be able to magically adjust to an updated domain.

JEM-Mosig (commented):
Just looked through the state.py again

rasa/shared/core/state.py
```diff
@@ -0,0 +1,426 @@
...
+
+
+def shallow_copy_without_empty_values(
+    given: Dict[Any, Optional[Sized]]
+) -> Dict[Any, Optional[Sized]]:
+    """
+    NOTE: this was _clean_state (with the difference that that would've dropped values
+    evaluating to False (of which we have nothing yet))
+    """
+    return {key: val for key, val in given.items() if val is None or not len(val) == 0}
```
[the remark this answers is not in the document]
ka-bu: lol, yeah that was a half-done change from "not val" to "oh I should use Sized" :D

rasa/shared/core/state.py
```diff
@@ -0,0 +1,426 @@
...
+
+    The substate will contain intent, text, and entities (if any are present).
+
+    Args:
+        tracker: dialog state tracker containing the dialog so far
+    Returns:
+        a dictionary containing intent, text and set entities
+
+    TODO: why does *the event* get to decide how it is featurized? and why is that
+      ignored in UserUttered.as_sub_state ?
```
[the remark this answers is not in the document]
ka-bu: I was mixing something up here, I think - the filtering of entities based on intent is not happening in UserUttered.as_sub_state

rasa/shared/core/state.py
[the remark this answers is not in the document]
ka-bu: fixed

rasa/shared/core/state.py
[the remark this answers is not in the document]
ka-bu: I've removed that comment

PR #9075, opened 2021-07-08
Fix interactive learning

dakshvar22:
**Proposed changes**:
- ...

### DIFF
--- a/rasa/core/training/interactive.py
+++ b/rasa/core/training/interactive.py
@@ -77,7 +77,7 @@ from rasa.core.utils import AvailableEndpoints
 from rasa.shared.importers.rasa import TrainingDataImporter
 from rasa.utils.common import update_sanic_log_level
 from rasa.utils.endpoints import EndpointConfig
-from rasa.shared.exceptions import RasaCoreException
+from rasa.shared.exceptions import InvalidConfigException
 
 # noinspection PyProtectedMember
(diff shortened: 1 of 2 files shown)


rasa/core/training/interactive.py
```diff
@@ -966,10 +969,10 @@ async def _predict_till_next_listen(
     listen = False
     while not listen:
         result = await request_prediction(endpoint, conversation_id)
-        predictions = result.get("scores")
+        predictions = result.get("scores") or []
```
wochinge, line 972: how about raising an exception in case `predictions` is empty?

wochinge (approved):
Sorry for introducing this bug!

rasa/core/training/interactive.py
dakshvar22: `RasaCoreException` or something more specific?

dakshvar22: Added `RasaCoreException`. Can we add any more specific information in the message, for example: Why such a situation may arise?

wochinge: Do we have something like an invalid config exception? I'd say that they need to retrain their model with a policy specified in the config then?

dakshvar22: Added.

wochinge (approved):
Thanks and sorry for breaking this 🙌🏻

PR #9079, opened 2021-07-08
`UnexpecTEDIntentPolicy` predictions fix

### LINKED ISSUE #9057: action_unlikely_intent prediction hides TED errors in failed_test_stories.yml
**Rasa version**: 2.7.1

**Python version**: 3.7.10

**Issue**:
When running `rasa test` with a policy ensemble that includes UnexpecTEDIntent policy, if the ensemble fails after an `action_unlikely_intent` is predicted by UnexpecTEDIntentPolicy, the action will be recorded by `rasa test` as being correct in the metrics and will not show up in `failed_test_stories.yml`.

Let’s take an example test story:
```
- story:
  - intent: greet
  - action: utter_greet
```
When a partial part of this test story:
```
  - story:
   - intent: greet
```
 is passed to the ensemble it can result in an `action_unlikely_intent` being triggered
```
- story:
  - intent: greet
  - action: action_unlikely_intent
```
Since, the last action triggered by ensemble was `action_unlikely_intent`, the ensemble should be queried again to see if other policies can actually predict `utter_greet`. So, the following story:
```
- story:
  - intent: greet
  - action: action_unlikely_intent
```
should be fed again to the ensemble. Now there are two cases possible:

**Case 1:** `utter_greet` gets predicted. So, the story looks like this - 
```
- story:
  - intent: greet
  - action: action_unlikely_intent
  - action: utter_greet
```
This story should go in `stories_with_warnings.yml`, as the expected action was correctly predicted after `action_unlikely_intent`.

**Case 2:** `utter_greet` does not get predicted but some other action gets predicted. So, the story looks like this - 
```
- story:
  - intent: greet
  - action: action_unlikely_intent
  - action: some_other_action
```
This story should end up in `failed_test_stories.yml` because `some_other_action` is predicted instead of `utter_greet`
Right now, case (2) is not being checked and can be reproduced with this example ….

```shell
# Make sure you have checked out the intent-ted branch on your rasa install
git clone https://github.com/rasahq/ited-tolerance-experiments
cd ited-tolerance-experiments

# Train a ted only ensemble
rasa train core -c ted-config.yml -s dataset1/ --out models/ted
# Train a ted+unexpecTEDIntentpolicy ensemble
rasa train core -c ited-config.yml -s dataset1/ --out models/ited
```

Run `rasa test` on the TED only ensemble
```shell
rasa test core -s test-bug/ -m models/ted --out results/ted
```
and you should see the following results:
```shell
...
2021-07-06 16:40:34 INFO     rasa.core.test  - Evaluation Results on CONVERSATION level:
2021-07-06 16:40:34 INFO     rasa.core.test  - 	Correct:          0 / 1
2021-07-06 16:40:34 INFO     rasa.core.test  - 	Accuracy:         0.000
...
2021-07-06 16:40:36 INFO     rasa.core.test  - Evaluation Results on ACTION level:
2021-07-06 16:40:36 INFO     rasa.core.test  - 	Correct:          10 / 11
2021-07-06 16:40:36 INFO     rasa.core.test  - 	F1-Score:         0.879
2021-07-06 16:40:36 INFO     rasa.core.test  - 	Precision:        0.864
2021-07-06 16:40:36 INFO     rasa.core.test  - 	Accuracy:         0.909
2021-07-06 16:40:36 INFO     rasa.core.test  - 	In-data fraction: 0.455
...
```
which are correct.

Run the ensemble with UnexpecTEDIntentPolicy 
```shell
rasa test core -s test-bug/ -m models/ited --out results/ited
```

and you get:

```shell
...
2021-07-06 16:41:36 INFO     rasa.core.test  - Evaluation Results on CONVERSATION level:
2021-07-06 16:41:36 INFO     rasa.core.test  - 	Correct:          1 / 1
2021-07-06 16:41:36 INFO     rasa.core.test  - 	Accuracy:         1.000
...
2021-07-06 16:41:37 INFO     rasa.core.test  - Evaluation Results on ACTION level:
2021-07-06 16:41:37 INFO     rasa.core.test  - 	Correct:          10 / 10
2021-07-06 16:41:37 INFO     rasa.core.test  - 	F1-Score:         1.000
2021-07-06 16:41:37 INFO     rasa.core.test  - 	Precision:        1.000
2021-07-06 16:41:37 INFO     rasa.core.test  - 	Accuracy:         1.000
2021-07-06 16:41:37 INFO     rasa.core.test  - 	In-data fraction: 0.5
...
```

Looking in the failed test stories you can see TED's error when it's run in isolation `results/ted/failed_test_stories.yml`:

```yaml
...
  - action: utter_flight_available
  - intent: deny
  - action: utter_anything_else
  - intent: deny
  - action: utter_goodbye  # predicted: utter_anything_else
```

However, the corresponding file for the combined TED+UnexpecTEDIntentPolicy (`results/ited/failed_test_stories.yml`) is empty while the action_unlikely_intent shows up in `results/ited/stories_with_warnings.yml`.

**WARNING:** This does not change the case where `action_unlikely_intent` was actually expected to be triggered at a conversation turn but was not predicted by the ensemble. Such cases should still go to `failed_test_stories.yml`.

alwx:
**Proposed changes**:
- fix #9057


rasa/core/test.py
```diff
@@ -600,47 +569,52 @@ def _collect_action_executed_predictions(
...
+        )
+    elif predicted_action_unlikely_intent:
+        partial_tracker.update(
+            ActionExecuted(
+                predicted_action,
+                prediction.policy_name,
+                prediction.max_confidence,
+                event.timestamp,
+            )
+        )
```
dakshvar22, line 610: This is the source for why tests are failing. This should only be done if `action_unlikely_intent` was actually expected. If you add it to the partial tracker even when it was unexpected, the policy predictions will get affected.

dakshvar22: Hmm, but if you don't add it to the tracker then the policy ensemble will keep predicting `action_unlikely_intent` again and again. I'll try writing the test again to see if I can simulate the condition better.

rasa/shared/core/events.py
```diff
@@ -1488,6 +1489,7 @@ def __init__(
         metadata: Optional[Dict] = None,
         action_text: Optional[Text] = None,
         hide_rule_turn: bool = False,
+        predicted_action_unlikely_intent: bool = False,
```
dakshvar22, line 1492: We need engineering advice here. My take is that we should try to avoid leaking the logic of what happened in the conversation in the past into `ActionExecuted`.
@alwx's [take](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1625828747322600?thread_ts=1625644380.222100&cid=C8LEHMTAM) is that it's quite complex / hacky to implement the functionality we need without doing this.

dakshvar22: @alwx You mentioned that this is done to handle the case when:
1. `action_unlikely_intent` was predicted.
2. It wasn't expected to be predicted.
3. Ensemble still predicted the correct action just after `action_unlikely_intent`.

So, to display cases like this:

```
- action: any_correct_action # predicted: action_unlikely_intent
```

Can't we create a new event `WarningPredicted` which also inherits from `ActionExecuted` and use it to handle this case?

dakshvar22: I'll be in meetings for the next couple of hours. Can you also please take a look at why `tests/shared/core/test_trackers.py::test_tracker_write_to_story` is failing? I think it's because of the new attribute added and other parts of the codebase not expecting it. (My original worry 😅 )

alwx: Done!

rasa/core/test.py
```diff
@@ -651,9 +617,14 @@ def _collect_action_executed_predictions(
                 prediction.max_confidence,
                 event.timestamp,
                 metadata=prediction.action_metadata,
+                predicted_action_unlikely_intent=previously_predicted_action_unlikely_intent,
```
dakshvar22, line 620: @alwx You'll need to shorten the variable name here.

rasa/core/test.py
```diff
@@ -667,6 +638,16 @@ def _collect_action_executed_predictions(
...
+                ACTION_UNLIKELY_INTENT_NAME,
+                predicted_action,
+                prediction.policy_name,
+                prediction.max_confidence,
+                event.timestamp,
+            )
+        )
     else:
         partial_tracker.update(
             ActionExecuted(
```
dakshvar22, line 653: @alwx Don't you also need to transfer the metadata in both of these `elif` / `else` constructs?

dakshvar22 (changes requested):

rasa/shared/core/events.py
```diff
@@ -1935,3 +1936,79 @@ def apply_to(self, tracker: "DialogueStateTracker") -> None:
         """Applies event to current conversation state."""
         # noinspection PyProtectedMember
         tracker._reset()
+
+
+class WronglyPredictedAction(ActionExecuted):
```
joejuzl, line 1941: I'm pretty sure we don't want to move these here. Once something is in `shared` it can be used by Rasa X and has to be supported accordingly. These are also test only events right? Not something that can actually be predicted.

rasa/shared/core/training_data/structures.py
```diff
@@ -235,6 +236,11 @@ def is_action_unlikely_intent(event: Event) -> bool:
         return (
             type(event) == ActionExecuted
             and event.action_name == ACTION_UNLIKELY_INTENT_NAME
+        ) or (
+            type(event) == WronglyPredictedAction
```
joejuzl, line 240: Do we still need this now we have `WarningPredictedAction`?

joejuzl: Also do we need any check for these test events here? They are not actually predicted ever right?

rasa/core/test.py
```diff
@@ -600,47 +570,43 @@ def _collect_action_executed_predictions(
...
             action_predictions=[predicted_action], action_targets=[expected_action]
         )
 
-    has_prediction_target_mismatch = (
-        action_executed_eval_store.has_prediction_target_mismatch()
-    )
-    if has_prediction_target_mismatch or (
-        predicted_warning and predicted_action != expected_action
+    if action_executed_eval_store.has_prediction_target_mismatch() or (
+        predicted_action_unlikely_intent and predicted_action != expected_action
```
joejuzl, line 609: Isn't this part now the equivalent of `previously_predicted_action_unlikely_intent`?

rasa/core/test.py
```diff
@@ -600,47 +570,43 @@ def _collect_action_executed_predictions(
...
     ):
         partial_tracker.update(
             WronglyPredictedAction(
```
joejuzl, line 612: If we have an predict `action_unlikely_intent` but it's not in the story, and the following action is correct, won't be now end up with two `WronglyPredictedAction` in the tracker?

rasa/core/test.py
```diff
@@ -651,9 +617,14 @@ def _collect_action_executed_predictions(
...
                 event.timestamp,
                 metadata=prediction.action_metadata,
+                predicted_action_unlikely_intent=previously_predicted_action_unlikely_intent,
             )
         )
-        if fail_on_prediction_errors and has_prediction_target_mismatch:
+        if (
+            fail_on_prediction_errors
+            and predicted_action != ACTION_UNLIKELY_INTENT_NAME
+            and expected_action != ACTION_UNLIKELY_INTENT_NAME
```
joejuzl, line 626: Doesn't this break the requirement we fail if the test story contains `ACTION_UNLIKELY_INTENT_NAME` but we don't predict it?

rasa/core/test.py
```diff
@@ -667,6 +638,16 @@ def _collect_action_executed_predictions(
                     "training stories and retrain."
                 )
             raise WrongPredictionException(error_msg)
+    elif previously_predicted_action_unlikely_intent:
```
joejuzl, line 641: is this possible given that we already have `predicted_action_unlikely_intent and predicted_action != expected_action` in the `or` of the associated `if`?

tests/core/test_test.py
```diff
@@ -105,9 +143,24 @@ async def test_testing_valid_with_non_e2e_core_model(core_agent: Agent):
...
-    moodbot_domain: Domain, train_file_name: Path
+    moodbot_domain: Domain,
+    train_file_name: Path,
+    monkeypatch: MonkeyPatch,
+    ignore_action_unlikely_intent: bool,
 ) -> Agent:
 
+    # We need `RulePolicy` to predict the correct actions
+    # in a particular conversation context as seen during training.
+    # Since it can get affected by `action_unlikely_intent` being triggered in
```
joejuzl, line 154: I don't really understand this. Does this not also cause issues using the `RulePolicy` during normal prediction?

joejuzl (changes requested):

rasa/shared/core/events.py
wochinge: +1

tests/core/test_test.py
dakshvar22: @joejuzl 
I think I can explain this better.

Firstly, we are using `RulePolicy` as an example policy to get determinism in actions predicted from the ensemble. We don't want to worry about any other action being incorrectly predicted and only want to test that `action_unlikely_intent` is correctly handled by `rasa test`. 

Now, `RulePolicy` is the only policy which does not ignore `action_unlikely_intent`'s presence in the tracker. So, for a lot of tests that are written / modified in this PR, we cannot use `RulePolicy` as it is. We want it to ignore it in some circumstances otherwise we really cannot test all the test cases reliably.

During normal prediction, this situation would have never come up in the first place because:
1. An `action_unlikely_intent` is always only predicted right after the `UserUttered` event. It won't be predicted as a follow-up action ever.
2. If a rule can be used to make a prediction after `UserUttered` event then `action_unlikely_intent` will always be dropped (priorities inside ensemble).

So, in reality rules will always work as expected during normal prediction.

Does that clarify the need for all the monkey patching here? 😅

rasa/core/test.py
alwx: No, because we make another prediction => `predicted_action` gets overwritten

rasa/core/test.py
alwx: It is possible. `prev_action_unlikely_intent` is different from `predicted_action_unlikely_intent and predicted_action != expected_action` (see comment above)

rasa/core/test.py
alwx: If we predict `action_unlikely_intent` then we ask for another prediction, and only add this `WronglyPredictedAction` if there is a mismatch in expected/predicted. This is the only case when two `WronglyPredictedAction`s will be added but that's perfectly fine because we filter out `action_unlikely_intent`s from logs.

joejuzl: Ok but if `predicted_action_unlikely_intent and predicted_action != expected_action` on line 584 is `False`, then `predicted_action` is not changed because we don't re-run so we no `predicted_action_unlikely_intent and predicted_action != expected_action` on line 609 will also always be `False`. And if line 584 is `True` then we know the next action can't be another `action_unlikely_intents` right? So can we change 609 to just `if action_executed_eval_store.has_prediction_target_mismatch()`?

tests/core/test_test.py
joejuzl: Thanks for the explanation. 

> Now, RulePolicy is the only policy which does not ignore action_unlikely_intent's presence in the tracker.

Why is this the case?

dakshvar22: Other policies are trained only on stories and `action_unlikely_intent` cannot be included in a training story because it is un-learnable. Hence, they ignore `action_unlikely_intent` during inference as well.

However, you can include `action_unlikely_intent` as a [rule to customize what should happen when `action_unlikely_intent` is triggered](https://github.com/RasaHQ/rasa/pull/8913/files#diff-507719c0f3f45420d32a4427131fb37c24cff2bdbd9868410cab35ad6552557aR197). Since, `RulePolicy` is supposed to handle everything rules it cannot ignore `action_unlikely_intent`'s presence in the tracker.

rasa/core/test.py
alwx: That's correct. Changed that.

rasa/core/test.py
alwx: Updated the condition and added a test for it (`test_action_unlikely_intent_fail_on_prediction_errors`)

tests/core/test_test.py
```diff
@@ -256,13 +256,54 @@ async def test_action_unlikely_intent_correctly_predicted(
     )
 
     result = await rasa.core.test.test(
-        str(file_name), agent, out_directory=str(tmp_path),
+        str(file_name), agent, out_directory=str(tmp_path), fail_on_prediction_errors = True
```
dakshvar22: @alwx should this be set to `True` in all the tests of this PR using `rasa.core.test.test`?

tests/core/test_test.py
```diff
@@ -256,13 +256,54 @@ async def test_action_unlikely_intent_correctly_predicted(
...
 
 
+async def test_action_unlikely_intent_fail_on_prediction_errors(
+        monkeypatch: MonkeyPatch, tmp_path: Path, moodbot_domain: Domain
+):
+    monkeypatch.setattr(
+        SimplePolicyEnsemble,
+        "probabilities_using_best_policy",
+        _probabilities_with_action_unlikely_intent_for(["mood_unhappy"]),
+    )
```
dakshvar22: @alwx Isn't `test_action_unlikely_intent_not_found` testing the cases where `action_unlikely_intent` was expected but was not found and hence a prediction error is triggered?

tests/core/test_test.py
dakshvar22: I don't think it should be set to `True` at least anywhere else. Otherwise we can't test the report as it won't be created.

alwx: For some but not for all — we sometimes also need to test if the generated report is correct, and `fail_on_prediction_errors=True` means no report will be generated.

tests/core/test_test.py
alwx: Yes, but it checks the report, and this one fails if intent was not found & there is no report that needs to be generated (which happens when `fail_on_prediction_errors=True`). I think it makes sense to have this test because the condition of whether the exception needs to be raised needs to be tested regardless of anything else.

rasa/core/test.py
```diff
@@ -600,48 +642,42 @@ def _collect_action_executed_predictions(
...
-    predicted_warning = action_executed_eval_store.predicted_warning(predicted_action)
-    if not predicted_warning:
+        )
+        prev_action_unlikely_intent = True
+        predicted_action, prediction, policy_entity_result = _run_action_prediction(
+            processor, partial_tracker, expected_action
+        )
+        action_executed_eval_store.add_to_store(
+            action_predictions=[predicted_action], action_targets=[expected_action]
+        )
```
dakshvar22, line 674: The code inside `else` is the same as these 3 lines. You can remove the `else` construct and move these 3 lines out.

rasa/core/test.py
```diff
@@ -600,48 +642,42 @@ def _collect_action_executed_predictions(
...
 
-        if (
-            prediction.policy_name
-            and predicted_action != expected_action
-            and _form_might_have_been_rejected(
-                processor.domain, partial_tracker, predicted_action
+    predicted_action_unlikely_intent = predicted_action == ACTION_UNLIKELY_INTENT_NAME
+    if predicted_action_unlikely_intent and predicted_action != expected_action:
+        partial_tracker.update(
+            WronglyPredictedAction(
```
dakshvar22, line 658: Wouldn't this and line 681 result in two successive `WronglyPredictedAction`? Could be understanding it wrongly since I am not familiar with `action_executed_eval_store`.

rasa/core/test.py
```diff
@@ -914,6 +967,22 @@ async def _collect_story_predictions(
     )
 
 
+def _filter_step_events(step: StoryStep) -> StoryStep:
+    events = []
+    for event in step.events:
+        if (
+            type(event) == WronglyPredictedAction
```
dakshvar22, line 974: Can we use `isinstance(event, WronglyPredictedAction)` to be consistent with the rest of the codebase?

rasa/shared/core/events.py
```diff
@@ -36,6 +36,7 @@
     ENTITY_LABEL_SEPARATOR,
     ACTION_SESSION_START_NAME,
     ACTION_LISTEN_NAME,
+    ACTION_UNLIKELY_INTENT_NAME,
```
dakshvar22, line 39:
```suggestion
```

tests/core/test_test.py
```diff
@@ -210,8 +275,51 @@ async def test_action_unlikely_intent_complete(
...
+
+    # We train on the above story so that RulePolicy can memorize
+    # it and we don't have to worry about other actions being
+    # predicted correctly.
+    agent = await _train_rule_based_agent(
+        moodbot_domain, file_name, monkeypatch, ignore_action_unlikely_intent=False
+    )
+
+    with pytest.raises(rasa.core.test.WrongPredictionException):
+        result = await rasa.core.test.test(
```
dakshvar22, line 304:
```suggestion
        await rasa.core.test.test(
```

tests/core/test_test.py
```diff
@@ -191,17 +251,22 @@ async def test_action_unlikely_intent_2(
...
+        out_directory=str(tmp_path),
+        fail_on_prediction_errors=True,
     )
     assert "report" in result.keys()
     assert result["report"]["conversation_accuracy"]["correct"] == 1
     assert result["report"]["conversation_accuracy"]["with_warnings"] == 0
 
 
-async def test_action_unlikely_intent_complete(
+async def test_action_unlikely_intent_fail_on_prediction_errors(
```
dakshvar22, line 269: I think from what you explained as the utility of this test, it looks like we should rename this test to not indicate that this is specific to `action_unlikely_intent` as we expect it to raise this exception on any failed prediction except `action_unlikely_intent`.

dakshvar22 (changes requested):
Mainly reviewed `rasa/core/test.py` as the shared codeowner for those files.

rasa/core/test.py
alwx: That's possible but the first `WronglyPredictedAction` will be filtered out so it works exactly as expected :)

dakshvar22 (approved):
Thanks for addressing all my comments 🚀

joejuzl (approved):
👍

Issue #8883, opened 2021-06-15
3.0 Deprecation Warning: Remove deprecated `Message` method


ancalita:
**Background**
This warning was found in:
- rasa.shared.nlu.training_data.message.py:L188
```
"""`get_combined_intent_response_key` is deprecated and ""
""will be removed in Rasa 3.0.0. ""
""Please use `get_full_intent` instead."""
```

**Definition of Done**
- [ ] resolve warning in 3.0 version
- [x] Removed the method from the codebase.
- [x] Added a changelog entry for 3.0 saying we've removed the previously deprecated method.
- [x] Removed the references in the docs

samsucik:
Can this be done right now? For 3.0 stuff, do we simply target `main`?

In any case, since the deprecated method doesn't appear anywhere in our Rasa Open Source tests or docs, the definition of done can be, I think, as simple as:
1. Removed the method from the codebase.

ancalita:
> Can this be done right now? For 3.0 stuff, do we simply target main?

I think all deprecations will be part of the prep work that can be done after the `2.8.` release, targeting `main` - @m-vdb correct me if I'm wrong about the timeline?

m-vdb:
yep 👍🏻

PR #9088, opened 2021-07-09
Deprecate Message method get_combined_intent_response_key

kedz:
**Proposed changes**:
- removes deprecated `get_combined_intent_response_key` method of `Message` class from `rasa.shared.nlu.training_data.message.py`.
- Added a changelog entry.
- fix https://github.com/RasaHQ/rasa/issues/8883

- ~[ ] added some tests for the functionality~
- ~[ ] updated the documentation~


rasa/shared/nlu/training_data/message.py
```diff
@@ -179,21 +179,11 @@ def get_full_intent(self) -> Text:
...
-            category=DeprecationWarning,
-        )
-        return self.get_full_intent()
-
     @staticmethod
     def separate_intent_response_key(
         original_intent: Text,
     ) -> Tuple[Text, Optional[Text]]:
-
+        """Split intent into main intent name and optional sub-intent name."""
```
wochinge, line 186:
```suggestion
        """Splits intent into main intent name and optional sub-intent name."""
```

wochinge: should we give a concrete example?

wochinge (approved):
🚀 
Can you please link the related issue in the pull request description so that issue and pull request are connected? 🙌🏻

rasa/shared/nlu/training_data/message.py
koernerfelicia: +1 (though I see now I was removed as reviewer :) )

jupyterjazz (approved):

rasa/shared/nlu/training_data/message.py
kedz: @joejuzl adding you since Tobias in on vacation. So I was thinking of adding this example:
```python
        """Splits intent into main intent name and optional sub-intent name.

        For example, `"FAQ/how_to_contribute"` would be split into
        `("FAQ", "how_to_contribute")`. If there is no response delimiter, the
        second tuple item is `None`, e.g. "FAQ" would be mapped to `("FAQ", None)`.
        """
```

My only hesitation with this is that we don't actually split on `"/"` but rather the constant `RESPONSE_IDENTIFIER_DELIMITER`. So there is the risk of this docstring going stale if that delimiter ever changes.

I think this alternative while accurate is somewhat confusing:
```python
        """Splits intent into main intent name and optional sub-intent name.

        For example, `f"FAQ{RESPONSE_IDENTIFIER_DELIMITER}how_to_contribute"` would be split into
        `("FAQ", "how_to_contribute")`. If there is no response delimiter, the
        second tuple item is `None`, e.g. "FAQ" would be mapped to `("FAQ", None)`.
        """
```

Do you think the first is ok to use or do you have an alternative suggestion?

jupyterjazz: Does it make sense to do both?
```yaml
"""Splits intent into main intent name and optional sub-intent name.

For example, `"FAQ/how_to_contribute"` would be split into
`("FAQ", "how_to_contribute")`.  The response delimiter can 
take different values (not just "\") and depends on the 
constant - RESPONSE_IDENTIFIER_DELIMITER.
If there is no response delimiter, the second tuple item 
is `None`, e.g. "FAQ" would be mapped to `("FAQ", None)`.
"""
```

joejuzl (approved):
🚀

rasa/shared/nlu/training_data/message.py
kedz: @jupyterjazz yeah I like that. Will make that change.

joejuzl: Sorry just saw this. We've had similar discussion in the past about how to handle constants in docstrings - it's a tricky one :) 
The only problem I have with the proposed solution is that this is a docstring for a public method, and `RESPONSE_IDENTIFIER_DELIMITER` is not really a configurable thing, so it might be slightly confusing.

Is it not possible to fstring the value into the docstrings? That way it would always be up-to-date with the constant.

Issue #8882, opened 2021-06-15
3.0 Deprecation Warning: Deprecate fallback spaCy model


ancalita:
**Background**
This warning was found in:
- rasa.nlu.utils.spacy_utils.py:L140
```
"f""SpaCy model is not properly configured! ""
 f""Please add a `model` property to `SpacyNLP`. ""
f""Will use '{spacy_model_name}' as a fallback spaCy model. ""
f""This fallback will be deprecated in Rasa 3.0"""
```

**Definition of Done**
- [x] resolve warning in 3.0 version
- [x] Removed all uses of _check_model_fallback in favour of using directly model from config. Here I assume we don't have to do any checking on model -- if a Spacy model with that name doesn't exist, then Spacy will raise an error? 🤔 Currently, we do checking but it requires hard-coding all possible Spacy model names, which then leads to issues like Support spaCy 3.1 #9068 which are just about updating that hard-coded list afaics.
- [x] Remove tests for _check_model_fallback
- [x] Removed the references in the docs

koaning:
I might be able to pick this one up. I implemented the spaCy 3.0 and wouldn't mind wrapping it up for Rasa 3.0. 

Just to check, I can make a fix with a PR to master?

samsucik:
@koaning this issue needs to be done before 3.0, while #9068 doesn't necessarily have to go into 3.0 (strictly speaking). Seeing that you assigned yourself on this one, are you actually working on it? To me it looks like an issue that research can easily pick up...

As for a definition of done, the method in question (`_check_model_fallback`) is found in the codebase and in some tests (but not in docs), so we might say:
1. Removed all uses of `_check_model_fallback` in favour of using directly `model` from config. _Here I assume we don't have to do any checking on `model` -- if a Spacy model with that name doesn't exist, then Spacy will raise an error? 🤔  Currently, we [do checking but it requires hard-coding all possible Spacy model names](https://github.com/RasaHQ/rasa/blob/300df5ac085c174521599364efabddb4000b6716/rasa/nlu/utils/spacy_utils.py#L105), which then leads to issues like #9068 which are just about updating that hard-coded list afaics._
2. Remove tests for `_check_model_fallback`

koaning:
@samsucik I'll gladly pick it up if it removes a burden from the research team. If you folks have the capacity then I certainly won't mind anybody else picking this up. 👍 

[...]

PR #9091, opened 2021-07-12
Remove spaCy 3.0 deprecation warnings

koaning:
This PR fixes https://github.com/RasaHQ/rasa/issues/8882.

### DIFF
--- a/rasa/nlu/utils/spacy_utils.py
+++ b/rasa/nlu/utils/spacy_utils.py
@@ -69,9 +69,7 @@ class SpacyNLP(Component):
             cls.defaults, component_config
         )
 
-        spacy_model_name = cls._check_model_fallback(
-            component_config.get("model"), config.language, warn=True
-        )
+        spacy_model_name = component_config.get("model")
 
         logger.info(f"Trying to load spacy model with name '{spacy_model_name}'")
 
@@ -85,9 +83,7 @@ class SpacyNLP(Component):
         cls, component_meta: Dict[Text, Any], model_metadata: "Metadata"
     ) -> Optional[Text]:
 
-        spacy_model_name = cls._check_model_fallback(
-            component_meta.get("model"), model_metadata.language, warn=False
-        )
+        spacy_model_name = component_meta.get("model")
 
         return cls.name + "-" + spacy_model_name
 
--- a/rasa/nlu/utils/spacy_utils.py
+++ b/rasa/nlu/utils/spacy_utils.py
@@ -91,57 +91,6 @@ class SpacyNLP(Component):
 
         return cls.name + "-" + spacy_model_name
 
-    @staticmethod
-    def _check_model_fallback(
-        spacy_model_name: Union[str, None], language_name: str, warn: bool = False
-    ) -> Text:
-        """This method checks if the `spacy_model_name` is missing.
-
-        If it is missing, we will attempt a fallback. This feature is a measure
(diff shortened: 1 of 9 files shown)


koaning:
I discovered some `core` tests that passed before and needed to be changed. Figured I should also add a test to make sure that we raise an appropriate error when the user forgets to add a `model` to their `SpacyNLP` config.

koaning:
@jupyterjazz I think you got auto-assigned to this PR. On my end: it's no longer a work in progress so feel free to review. Also, feel free to pass it along to somebody else if you feel the auto-tagging system got it wrong.

rasa/nlu/utils/spacy_utils.py
```diff
@@ -85,62 +82,19 @@ def cache_key(
...
-        )
+        if not component_meta.get("model"):
+            raise InvalidModelError(
+                f"Missing model configuration for `SpacyNLP` in `config.yml`.\n"
+                f"You must pass a model to the `SpacyNLP` component explicitly.\n"
+                f"For example:\n"
+                f"- name: SpacyNLP\n"
+                f"  model: en_core_web_md\n"
+                f"More informaton can be found on {DOCS_URL_COMPONENTS}#spacynlp"
+            )
```
jupyterjazz, line 94: Seems good!
Just one question though: shouldn't we also write a test for this error message?

tests/nlu/test_spacy_utils.py
```diff
@@ -5,35 +5,13 @@
...
 def test_model_raises_error_not_exist():
     """It should throw a direct error when a model doesn't exist."""
     with pytest.raises(InvalidModelError):
         SpacyNLP.create({"model": "dinosaurhead"}, RasaNLUModelConfig())
+
+
+def test_model_raises_error_no_model_given():
+    """It should throw a direct error when no model is passed."""
+    with pytest.raises(InvalidModelError):
+        SpacyNLP.create({"model": None}, RasaNLUModelConfig())
```
jupyterjazz, line 17: I'm just wondering if it makes more sense to combine these two tests cause they only differ with an argument value for "model".

rasa/nlu/utils/spacy_utils.py
koaning: This is handled by the `test_model_raises_error_no_model_given`-test.

tests/nlu/test_spacy_utils.py
koaning: Fair enough. It's two examples that could be one. I usually prefer to go parameterized once I hit `n=3` but I'll make the change.

rasa/nlu/utils/spacy_utils.py
koaning: Ah... you mean you want to test the contents of the message? We could go there but that feels like a bit of overkill.

jupyterjazz: I checked and `test_model_raises_error_no_model_given` covers the error in `load_model` and not this one. Am I missing something?
error

koaning: Yep. I just checked and fixed it. There was actually a small bug there where it gave the wrong error message. Just fixed it by changing the test and making a fix. 

Good call on your end!

koaning:
@jupyterjazz I've incorporated your requests and they led to a better PR. Does it look good to merge now?

jupyterjazz (approved):
Thanks! Looks great! 👍

Issue #8868, opened 2021-06-14
3.0 Deprecation Warning: Resolve `Domain`-related warnings


ancalita:
**Background**:
These warnings were found in:
- rasa.shared.core.domain.py:L768
```
" f""{Domain.__name__}.{Domain.action_names.__name__} ""
f""is deprecated and will be removed version 3.0.0."""
```
- rasa.shared.core.domain.py:L866
```
"f""'{self.add_categorical_slot_default_value.__name__}' is deprecated and ""
f""will be removed in Rasa Open Source 3.0.0. This method is now ""
f""automatically called when the Domain is created which makes a manual ""
f""call superfluous."""
```
- rasa.shared.core.domain.py:L892
```
"f""'{self.add_requested_slot.__name__}' is deprecated and ""
f""will be removed in Rasa Open Source 3.0.0. This method is now ""
f""automatically called when the Domain is created which makes a manual ""
f""call superfluous."""
```
- rasa.shared.core.domain.py:L930
```
"f""'{self.add_knowledge_base_slots.__name__}' is deprecated and ""
f""will be removed in Rasa Open Source 3.0.0. This method is now ""
 f""automatically called when the Domain is created which makes a manual ""
 f""call superfluous."""
```
- rasa.shared.core.domain.py:L985
```
"f""'{Domain.__name__}.{Domain.random_template_for.__class__}' ""
f""is deprecated and will be removed version 3.0.0."","
```
- rasa.shared.core.domain.py:L1993
```
"f""The definition of slot mappings in your form ""
 f""should always be preceded by the keyword `{REQUIRED_SLOTS_KEY}`. ""
f""The lack of this keyword will be deprecated in ""
 f""Rasa Open Source 3.0.0. Please see {DOCS_URL_FORMS} ""
f""for more information."""
```
- rasa.shared.core.domain.py:L723
```
"""The `forms` section in the domain used the old Rasa Open Source 1 ""
""list format to define forms. Rasa Open Source will be configured to ""
""use the deprecated `FormAction` within the Rasa SDK. If you want to ""
 ""use the new Rasa Open Source 2 `FormAction` adapt your `forms` ""
 ""section as described in the documentation. Support for the ""
""deprecated `FormAction` in the Rasa SDK will be removed in Rasa Open ""
""Source 3.0."""
```
- rasa.shared.core.slots.py:L317
```
"f""{UnfeaturizedSlot.__name__} is deprecated ""
f""and will be removed in Rasa Open Source ""
[...]

PR #9101, opened 2021-07-12
remove deprecations from `Domain`

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/8868
- enforce `Domain` schema
- Exclude the migration guide from this check as the migration guide might contain training data which once was valid but is not valid anymore. E.g. the migration guide for Rasa Open Source 2 defined a form schema which is no longer valid since Rasa Open Source 3.

### DIFF
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -701,7 +701,6 @@ class Domain:
         for form_name, form_data in forms.items():
             if form_data is not None and REQUIRED_SLOTS_KEY not in form_data:
                 forms[form_name] = {REQUIRED_SLOTS_KEY: form_data}
-        # dict with slot mappings
         return list(forms.keys()), forms, []
 
     def __hash__(self) -> int:
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -757,21 +757,6 @@ class Domain:
         return self.user_actions + self.form_names
 
     @rasa.shared.utils.common.lazy_property
-    def action_names(self) -> List[Text]:
-        """Returns action names or texts."""
-        # Raise `DeprecationWarning` instead of `FutureWarning` as we only want to
-        # notify developers about the deprecation (e.g. developers who are using the
-        # Python API or writing custom policies). End users can't change anything
-        # about this warning except making their developers change any custom code
-        # which calls this.
-        rasa.shared.utils.io.raise_warning(
-            f"{Domain.__name__}.{Domain.action_names.__name__} "
-            f"is deprecated and will be removed version 3.0.0.",
-            category=DeprecationWarning,
-        )
-        return self.action_names_or_texts
-
-    @rasa.shared.utils.common.lazy_property
     def num_actions(self) -> int:
         """Returns the number of available actions."""
         # noinspection PyTypeChecker
@@ -781,17 +766,6 @@ class Domain:
     def num_states(self) -> int:
         """Number of used input states for the action prediction."""
         return len(self.input_states)
-
-    @rasa.shared.utils.common.lazy_property
-    def retrieval_intent_templates(self) -> Dict[Text, List[Dict[Text, Any]]]:
-        """Return only the responses which are defined for retrieval intents."""
-        rasa.shared.utils.io.raise_deprecation_warning(
-            "The terminology 'template' is deprecated and replaced by "
-            "'response', call `retrieval_intent_responses` instead of "
-            "`retrieval_intent_templates`.",
-            docs=f"{DOCS_URL_MIGRATION_GUIDE}#rasa-23-to-rasa-24",
-        )
-        return self.retrieval_intent_responses
 
     @rasa.shared.utils.common.lazy_property
     def retrieval_intent_responses(self) -> Dict[Text, List[Dict[Text, Any]]]:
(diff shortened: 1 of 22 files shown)


rasa/shared/core/domain.py
```diff
@@ -702,35 +698,11 @@ def _initialize_forms(
...
-            f"The `forms` section in the domain needs to contain a dictionary. "
-            f"Instead found an object of type '{type(forms)}'.",
-            docs=DOCS_URL_FORMS,
-        )
-
-        return [], {}, []
+        for form_name, form_data in forms.items():
+            if form_data is not None and REQUIRED_SLOTS_KEY not in form_data:
+                forms[form_name] = {REQUIRED_SLOTS_KEY: form_data}
+        # dict with slot mappings
```
wochinge, line 704:
```suggestion

```

wochinge:
@m-vdb Is it possible to exclude the migration guide from the docs training data validation? I don't think we should adapt old migration guide items.

m-vdb:
not sure what you're referring to exactly 🤔 we do have [this version by version migration guide](https://rasa.com/docs/rasa/migration-guide)

wochinge:
@m-vdb I am referring to [this failing check](https://github.com/RasaHQ/rasa/pull/9101/checks?check_run_id=3046327128)

m-vdb:
Ah - I think we still need the migration guide though. So it's probably best to move the files closer to the docs (inside the /docs folder for instance), instead of removing the files and the check.

wochinge:
@m-vdb I don't want to remove the migration guide. I just want to exclude the migration guide from the docs check which checks that our documentation only contains valid training data.

m-vdb:
if you exclude the file, then the docs will be broken and won't build

m-vdb:
@wochinge there's no existing mechanism to do that yet, so feel free to update the python test to exclude the migration guide and explain why, and ask my review!

tests/docs/test_docs_training_data.py
```diff
@@ -24,6 +24,13 @@
 
 @pytest.mark.parametrize("mdx_file_path", MDX_DOCS_FILES)
 def test_docs_training_data(mdx_file_path: Path):
+    if mdx_file_path == Path("docs/docs/migration-guide.mdx"):
```
wochinge, line 27: @m-vdb That's the part which'd require review by you

m-vdb: a cleaner way could be to remove the path from the `MDX_DOCS_FILES` variable with `MDX_DOCS_FILES.remove(Path("docs/docs/migration-guide.mdx"))` after line 16.

m-vdb (approved):
sounds good, proposed an alternative for the test

rasa/shared/core/domain.py
```diff
@@ -782,17 +738,6 @@ def num_states(self) -> int:
         """Number of used input states for the action prediction."""
         return len(self.input_states)
 
-    @rasa.shared.utils.common.lazy_property
-    def retrieval_intent_templates(self) -> Dict[Text, List[Dict[Text, Any]]]:
-        """Return only the responses which are defined for retrieval intents."""
-        rasa.shared.utils.io.raise_deprecation_warning(
```
ancalita, line 788: Your eagle eyes found a deprecation warning I missed 😅 👍

rasa/shared/utils/schemas/domain.yml
```diff
@@ -44,8 +44,18 @@ mapping:
...
-    type: "any"
+    type: "map"
     required: False
+    mapping:
+      regex;([A-Za-z]+):
+        type: "map"
+        mapping:
+          required_slots:
+            type: "map"
+            required: False
```
ancalita, line 55: Should we enforce this to be `required: True` actually? 🤔 
In the docs examples I've seen, `required_slots` is always present, but not sure if giving flexibility to miss this key in forms is welcome or unwelcome.

tests/shared/core/test_domain.py
```diff
@@ -779,16 +771,6 @@ def test_check_domain_sanity_on_invalid_domain():
...
         )
 
-    with pytest.raises(InvalidDomain):
-        Domain(
-            intents={},
-            entities=[],
-            slots=[],
-            responses={},
-            action_names=[],
-            forms=["random_name", "random_name"],
```
ancalita, line 789: Why was this test removed - shouldn't we test that forms as `list` raises `InvalidDomain`?

tests/shared/core/test_domain.py
```diff
@@ -1400,24 +1409,6 @@ def test_ignored_intents_slot_mappings_invalid_domain():
         Domain.from_dict(domain_as_dict)
 
 
-def test_form_with_no_required_slots_keyword():
```
ancalita, line 1403: I think this test should stay, unless you enforce `required_slots` in the schema.

ancalita (commented):
Are you planning to open a different PR for this deprecation warning in `rasa/shared/core/slots.py:L317`? It was the only one missed from the issue.
```
"f""{UnfeaturizedSlot.__name__} is deprecated ""
f""and will be removed in Rasa Open Source ""
f""3.0. Please change the type and configure the 'influence_conversation' ""
f""flag for slot '{name}' instead."""
```

tests/shared/core/test_domain.py
wochinge: I dropped it as it's covered by the schema. Can re-add it though.

rasa/shared/utils/schemas/domain.yml
wochinge: I put `required: False` so that you can still do 

```
forms:
  my_form:
```

wochinge: (e.g. if all slot mappings are optional)

tests/shared/core/test_domain.py
wochinge: re-added

tests/shared/core/test_domain.py
wochinge: re-added

wochinge:
> Are you planning to open a different PR for this deprecation warning in rasa/shared/core/slots.py:L317? It was the only one missed from the issue.

Sorry, I've overlooked that one! Thanks the heads up 💯

ancalita (approved):
💯 
Only pending the resolution of the failed CI check on missing docstring in `CategoricalSlot`.

PR #9108, opened 2021-07-12
ignore fake credentials

HotThoughts:
**Proposed changes**:
- Ignore fake credentials

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -113,7 +113,7 @@ title = "gitleaks config"
     '''(.*?)(jpg|gif|doc|pdf|bin)$''',
     '''(go.mod|go.sum)$''',
     '''docusaurus.config.js$''',
-    '''credentials.yml$''']
+    '''rasa/data/test_moodbot/credentials.yml$''']
     regexes = [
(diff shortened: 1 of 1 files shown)


.gitleaks.toml
```diff
@@ -112,7 +112,8 @@ title = "gitleaks config"
     files = ['''^\.?gitleaks.toml$''',
     '''(.*?)(jpg|gif|doc|pdf|bin)$''',
     '''(go.mod|go.sum)$''',
-    '''docusaurus.config.js$''']
+    '''docusaurus.config.js$''',
+    '''credentials.yml$''']
```
joejuzl, line 116: Maybe we should prefix it with the test folder to be safe?

joejuzl (changes requested):

.gitleaks.toml
tczekajlo: @HotThoughts Joe has a good point, can you add a full path to the file?

HotThoughts: Updated!

joejuzl (approved):

PR #9109, opened 2021-07-12
Fix and enable mypy return check

### LINKED ISSUE #8242: mypy: Fix `return` issues

Current number of issues with this type: 18

**Definition of Done**:
- [ ] enable `mypy` checks for `return`
- [ ] all existing `return` issues are fixed

joejuzl:
**Proposed changes**:
Fixes #8242


ancalita (commented):
I didn't look into most of the research-owned files, you might need to loop in somebody from Research for a more in-depth review 😬

rasa/core/actions/forms.py
```diff
@@ -425,7 +425,7 @@ async def validate_slots(
         domain: Domain,
         output_channel: OutputChannel,
         nlg: NaturalLanguageGenerator,
-    ) -> List[Event]:
+    ) -> List[Union[SlotSet, Event]]:
```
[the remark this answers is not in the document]
joejuzl: Unfortunately not as we append both `SlotSet` and `Event` objects. I also assumed you could just use the base class but you can't, see: https://github.com/python/mypy/issues/2984#issuecomment-285721489

rasa/core/actions/action.py
```diff
@@ -690,7 +690,7 @@ async def run(
 
             events_json = response.get("events", [])
             responses = response.get("responses", [])
-            bot_messages = await self._utter_responses(
+            bot_messages: List[Event] = await self._utter_responses(
```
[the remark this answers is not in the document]
joejuzl: This is related to the other point below.  As `List` is invariant type we would then have to specify a union of `Event` and `BotUttered` all the way up.

rasa/core/policies/sklearn_policy.py
```diff
@@ -338,7 +338,7 @@ def persist(self, path: Union[Text, Path]) -> None:
     @classmethod
     def load(
         cls, path: Union[Text, Path], should_finetune: bool = False, **kwargs: Any
-    ) -> Policy:
+    ) -> Optional[Policy]:
```
wochinge, line 341: we can't do this @joejuzl as this goes against the `Policy` interface

joejuzl: haha yes true. `TEDPolicy.load` was recently changed to have `return cls()`. We can do that here too.

wochinge: yes please

poetry.lock
```diff
@@ -3157,7 +3157,7 @@ transformers = ["transformers"]
 [metadata]
 lock-version = "1.1"
 python-versions = ">=3.6,<3.9"
-content-hash = "8319a632c0f0f25e3574887786ba9fb51277f7cc7010a02f8866933065506563"
+content-hash = "80436920af7907d20d9d32586e178c2ea66bb31ebb3c85c42a7219f77e6e8301"
```
ancalita, line 3160: out of curiosity, why would `content-hash` change here? Maybe to start from the beginning, what's the purpose of this field? 🙈

rasa/core/channels/socketio.py
```diff
@@ -203,10 +204,12 @@ async def connect(
 
                 if jwt_payload:
                     logger.debug(f"User {sid} connected to socketIO endpoint.")
+                    return True
                 else:
                     return False
             else:
                 logger.debug(f"User {sid} connected to socketIO endpoint.")
+                return True
```
ancalita, line 212: Should the return type in L197 also change from `Optional[bool]` to `bool`?

rasa/nlu/components.py
```diff
@@ -945,12 +941,3 @@ def create_component(
...
             )
-
-    def create_component_from_class(self, component_class: Type[C], **cfg: Any) -> C:
-        """Create a component based on a class and a configuration.
-
-        Mainly used to make use of caching when instantiating component classes."""
-
-        component_config = {"name": component_class.name}
-
-        return self.create_component(component_config, RasaNLUModelConfig(cfg))
```
ancalita, line 956: Why was this function removed?

rasa/shared/core/domain.py
```diff
@@ -1106,7 +1107,7 @@ def _get_featurized_entities(self, latest_message: UserUttered) -> Set[Text]:
 
     def _get_user_sub_state(
         self, tracker: "DialogueStateTracker"
-    ) -> Dict[Text, Union[Text, Tuple[Text]]]:
+    ) -> Dict[Text, Union[None, Text, List[Optional[Text]], Tuple[str, ...]]]:
```
ancalita, line 1110: What does the `...` represent here?

joejuzl: It means that it can be a tuple of varying length, but all of the items will be of type `str`. See: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#tuple-types

rasa/nlu/components.py
wochinge: it was unused

poetry.lock
wochinge: It changed because I added a new `pyproject.toml` entry. The content-hash ensures `pyproject.toml` and `poetry.lock` are not out of sync

rasa/core/channels/socketio.py
wochinge: done 👍🏻 Good catch!

ancalita (approved):
🎉

PR #9111, opened 2021-07-12
gitleaks - Add example tokens to allowlist 

HotThoughts:
**The problem:**

There is a **critical** bug in `gitleaks-action v1.3` - it will always green even it found leaks. For example, [this](https://github.com/RasaHQ/rasa/commit/41aba7259103989867e3f3b4f36d1c7103ce067c). 

This left the door open for a long time until recently we [upgraded](https://github.com/RasaHQ/rasa/commit/41aba7259103989867e3f3b4f36d1c7103ce067c) `gitleaks-action` from 1.3 to 1.6. To be precise, [it fixed in v1.4.0.](https://github.com/zricethezav/gitleaks-action/releases/tag/v1.4.0)

**Proposed changes**:
- A follow up fix for #9108 
- Instead of adding *file* to allowlist/whitelist, use regex to filter *pattern* under corresponding rules
  - allows Slack tokens that contains `test`, `XXX`, `xxx`, and `286425452756`
  - allows the fake API`3e34709d01ea89032asdebfe5a74518` and APIs that contains `12345670` 

**TODOs**
- [X] Scanned the codebase
From rasa directory, scan the whole git histroy by:
```console
$ gitleaks --path=. -v --config-path=$(pwd)/.gitleaks.toml

...

INFO[0728] scan time: 12 minutes 8 seconds 837 milliseconds 454 microseconds
INFO[0728] commits scanned: 26766
WARN[0728] leaks found: 232
```
Note: all of the leaks are test tokens/example APIs. Nothing to worry about. 
- [x] Update rules for all fake credentials that should be allowed by gitleaks


HotThoughts:
Merge this PR should fix the failure in [here](https://github.com/RasaHQ/rasa/pull/9111).

.gitleaks.toml
```diff
@@ -57,8 +60,8 @@ title = "gitleaks config"
     regex = '''(?i)(api_key|apikey|secret)(.{0,20})?['|"][0-9a-zA-Z]{16,45}['|"]'''
     tags = ["key", "API", "generic"]
     [rules.allowlist]
-        paths = ['''^examples\/.*''']
-        description = "ignore examples"
+	regexes = ['''(3e34709d01ea89032asdebfe5a74518|12345670)''']
```
tczekajlo, line 63: Out of curiosity, where did you get it? 😄  I mean the tokens

HotThoughts: We use `3e34709d01ea89032asdebfe5a74518` in [docs, and tests](https://github.com/RasaHQ/rasa/search?q=3e34709d01ea89032asdebfe5a74518). 
Same for `123456701ea89032asdebfe5a74518` 👉 it starts with `12345670` so we can just match this part.

tczekajlo (approved):
LGTM 🎖️

PR #9114, opened 2021-07-13
Turned `utils.common` deprecation warning into exception

### LINKED ISSUE #8873: 3.0 Deprecation Warning: Convert warning into exception in `rasa.shared.utils.common.py`
**Background**
This warning was found in `rasa.shared.utils.common.py:L48`
```
"f""`class_from_module_path()` is expected to return a class, ""
f""but {module_path} is not one. ""
f""This warning will be converted ""
f""into an exception in {NEXT_MAJOR_VERSION_FOR_DEPRECATIONS}."""
```

**Definition of Done**
- [ ] resolve warning in 3.0

twerkmeister:
**Proposed changes**:
- fixes #8873 


rasa/shared/utils/common.py
```diff
@@ -45,13 +41,10 @@ def class_from_module_path(
         raise ImportError(f"Cannot retrieve class from path {module_path}.")
 
     if not inspect.isclass(klass):
-        rasa.shared.utils.io.raise_deprecation_warning(
+        raise TypeError(
```
ancalita, line 44: I'm wondering if it's not better to raise a `RasaException` here, wouldn't this be an usage error? 🤔 
Also the docstring should be updated with the 2nd error raised other than `ImportError`.

tests/shared/utils/test_common.py
```diff
@@ -175,17 +174,7 @@ def test_class_from_module_path_not_found(
...
-    ],
-)
-def test_class_from_module_path_ensure_class(
-    module_path: Text, outcome: bool, result: Any, recwarn: WarningsRecorder
-):
-    klass = rasa.shared.utils.common.class_from_module_path(module_path)
-    assert klass is result
-
-    assert bool(len(recwarn)) is not outcome
+def test_class_from_module_path_fails():
```
ancalita, line 177: I think we also need a test for happy path?

ancalita (commented):
We'll also need a changelog entry.

tests/shared/utils/test_common.py
twerkmeister: Happy path is already tested here:
https://github.com/RasaHQ/rasa/blob/bd4ca3f62377eab72264c5421916bcd1c2091fa0/tests/shared/utils/test_common.py#L146-L159

rasa/shared/utils/common.py
twerkmeister: Good point, RasaException does not trigger the telemetry. I guess then we would want to remove the ImportError as well

tests/shared/utils/test_common.py
ancalita: 👍 I should have looked in the file before asking 😄

ancalita (approved):
🎉

rasa/shared/utils/common.py
ancalita: Maybe it's based to revert this change for `ImportError` based on the latest failed test information 😅 
I'm actually reviewing a PR that is filtering `ImportError` exceptions from being sent to Sentry so that will be taken care of too indirectly.

twerkmeister: 😅 agreed

PR #9116, opened 2021-07-13
8569 drop py36 support

### LINKED ISSUE #8569: Drop support for Python 3.6
**Description of Problem**:
Python 3.6 end of life is [Dec 23rd 2021](https://endoflife.date/python). Our next major release could be an opportunity to drop support

**Definition of Done**:
- [x] CI workflows are updated
- [x] Documentation is updated
- [ ] Internal documentation is updated
- [x] If some library updates are possible after dropping the support for 3.6 (matplotlib, etc...) open tickets for them

twerkmeister:
**Proposed changes**:
- fixes #8569 


twerkmeister:
Hmm seems that update makes mypy discover 146 untyped call errors 😕

twerkmeister:
Seems the newer version of sanic does not have type information yet. I reverted back to 20.3.0 and the type errors are gone again

twerkmeister:
@m-vdb The python 3.6 drop is working for now without the sanic update

It seems the last missing puzzle piece is removing the python 3.6 checks from the branch rules of the repository otherwise they are stuck in `Expected — Waiting for status to be reported` :https://github.community/t/expected-waiting-for-status-to-be-reported/18001 

I don’t know if there’s anything we should take into consideration there as it will disable this check repository wide. I assume the risk is low overall, since we’ll keep all the other 3.7 and 3.8 checks in place - so it’s unlikely something would get merged in the meantime

m-vdb:
> I don’t know if there’s anything we should take into consideration there as it will disable this check repository wide

we can do that for `main` and `3.x` branches

twerkmeister:
> we can do that for `main` and `3.x` branches

ah good to know! I can't do it, since I don't have the necessary edit rights - so up to you!

changelog/8569.removal.md
```diff
@@ -0,0 +1 @@
+Removed Python 3.6 support
```
m-vdb, line 1: I think we could mention that we're removing 3.6 support because it's reaching its end of life by the end of this year:
```suggestion
Removed Python 3.6 support as [it reaches its end of life in December 2021](https://www.python.org/dev/peps/pep-0494/#lifespan).
```
WDYT?

pyproject.toml
```diff
@@ -93,7 +93,7 @@ pytz = ">=2019.1,<2022.0"
 rasa-sdk = "^2.8.0"
 colorclass = "~2.2"
 terminaltables = "~3.1.0"
-sanic = ">=19.12.2,<21.0.0"
+sanic = "20.3.0"
```
m-vdb, line 96: could we keep the range though?

m-vdb (changes requested):
Thanks @twerkmeister ! Can you also address the following comments in the code:
- [rasa.telemetry](https://github.com/RasaHQ/rasa/blob/3ac95305b47270fa62bb70ecc8742e40e6f6a8f2/rasa/telemetry.py#L686-L688)
- [rasa.utils.common](https://github.com/RasaHQ/rasa/blob/3ac95305b47270fa62bb70ecc8742e40e6f6a8f2/rasa/utils/common.py#L300-L304)
- [tests/test_telemetry.py](https://github.com/RasaHQ/rasa/blob/3ac95305b47270fa62bb70ecc8742e40e6f6a8f2/tests/test_telemetry.py#L50) (and also below in the same file)

(I found those by searching for `3.6` everywhere)

pyproject.toml
twerkmeister: thing was just when I used `poetry update` it changed to 20.13.0 or so and then we had a ton of mypy errors - I think there was some problem with the stubs. I could have a look to see if I can actually resolve that

twerkmeister:
> Thanks @twerkmeister ! Can you also address the following comments in the code:
> 
> * [rasa.telemetry](https://github.com/RasaHQ/rasa/blob/3ac95305b47270fa62bb70ecc8742e40e6f6a8f2/rasa/telemetry.py#L686-L688)
> * [rasa.utils.common](https://github.com/RasaHQ/rasa/blob/3ac95305b47270fa62bb70ecc8742e40e6f6a8f2/rasa/utils/common.py#L300-L304)
> * [tests/test_telemetry.py](https://github.com/RasaHQ/rasa/blob/3ac95305b47270fa62bb70ecc8742e40e6f6a8f2/tests/test_telemetry.py#L50) (and also below in the same file)
> 
> (I found those by searching for `3.6` everywhere)

good catch! Will do!

pyproject.toml
m-vdb: so maybe `>=19.12.2,<20.13.0`? and maybe open a follow-up issue to upgrade sanic and do this investigation there?

wochinge:
Could we please make this a priority? 🙌🏻 We actually need that for the "engine" part of the architecture revamp. We can hack it for now but it's not ideal.

twerkmeister:
Hey @tmbo, would you still know from your comment a year ago what can be done here when we drop py3.6? It's not clear whether some functionality was missing in python 3.6 that forces that function to be a generator or whether those generators can be handled differently with py 3.7 onwards? Any pointer would be help!

https://github.com/RasaHQ/rasa/blob/3ac95305b47270fa62bb70ecc8742e40e6f6a8f2/rasa/telemetry.py#L680-L749

pyproject.toml
twerkmeister: Had to use `>=19.12.2,<20.12.0` to not get mypy issues

m-vdb (approved):
💯

joejuzl (approved):
LGTM!

wochinge:
wuhuu! 🚀

PR #9117, opened 2021-07-13
Add test + fix for cache_key in SpacyNLP component. 

koaning:
This PR fixes a bug that was accidentally introduced here: https://github.com/RasaHQ/rasa/pull/9091.


koaning:
@jupyterjazz I tried running the merged branch from yesterday and I found a *very* annoying bug. 

It turns out that rewriting the unit tests added an error. This is what we were testing yesterday.

```python
# This creates a component with a model that doesn't exist.
SpacyNLP.create({"model": "dinosaurhead"})
# This creates a component with a model equal to `None`
SpacyNLP.create({"model": None})
```

But, this is what we should have tested. 

```python
# This creates a component without a model being passed in
SpacyNLP.create({})
```

We didn't test for the latter, and that's where I started getting confusing error messages today. I figured I should immediately fix.

jupyterjazz (approved):
Definitely a tricky one.
Great that you caught it :rocket:

PR #9119, opened 2021-07-13
Remove Policy deprecations and set max_history default to None.

kedz:
**Proposed changes**:

- Remove the following deprecated policy classes:
  - `FormPolicy`
  - `MappingPolicy`
  - `FallbackPolicy`
  - `TwoStageFallbackPolicy`  
  - `SklearnPolicy`
- Update the max history defaults in `MemoizationPolicy`.
- Remove/edit tests that depended on deprecated modules.
- Remove deprecated policies from docs.
- Remove any soon-to-be-broken links from CHANGELOG.mdx to docs sections for these policies. 
  
- [X] Remove deprecated policy classes.
- [x] Update the max history defaults in policies
- [x] Remove/edit tests that depended on these modules
- [X] Remove deprecated policies from docs
- [x] Remove any soon-to-be-broken links from CHANGELOG.mdx to docs sections for these policies. 


rasa/core/policies/ensemble.py
```diff
@@ -539,22 +496,6 @@ def is_not_in_training_data(
...
         return not is_memorized or max_confidence == 0.0
 
-    @staticmethod
-    def _is_not_mapping_policy(
-        policy_name: Text, max_confidence: Optional[float] = None
-    ) -> bool:
-        from rasa.core.policies.mapping_policy import MappingPolicy
-
-        is_mapping = policy_name.endswith("_" + MappingPolicy.__name__)
-        # also check if confidence is 0, than it cannot be count as prediction
```
joejuzl, line 549: Is it ok that we no longer check this?

rasa/core/policies/ensemble.py
```diff
@@ -604,18 +545,12 @@ def _pick_best_policy(
...
-            elif confidence > best_confidence:
+            if confidence > best_confidence:
                 # pick the best policy
                 best_confidence = confidence
                 best_policy_name = policy_name
 
-        if form_confidence is not None and self._is_not_mapping_policy(
-            best_policy_name, best_confidence[0]
-        ):
+        if form_confidence is not None:
```
joejuzl, line 553: Is this ever possible now?

rasa/core/policies/ensemble.py
```diff
@@ -694,43 +629,50 @@ def _best_policy_prediction(
 
         return self._pick_best_policy(predictions)
 
-    def _fallback_after_listen(
-        self, domain: Domain, prediction: PolicyPrediction
+    @staticmethod
+    def _get_prediction(
```
joejuzl, line 633: What is this new method for?

tests/core/test_agent.py
```diff
@@ -237,13 +236,12 @@ def test_form_without_form_policy(policy_config: Dict[Text, List[Text]]):
...
         )
-    assert "neither added the 'RulePolicy' nor the 'FormPolicy'" in str(execinfo.value)
+    assert "have not added the 'RulePolicy'" in str(execinfo.value)
 
 
 @pytest.mark.parametrize(
     "policy_config",
     [
-        {"policies": [{"name": FormPolicy.__name__}]},
         {"policies": [{"name": RulePolicy.__name__}]},
```
joejuzl, line 245: I guess we don't need to `parametrize` now?

rasa/core/policies/ensemble.py
```diff
@@ -790,20 +721,18 @@ def _check_policy_for_forms_available(
     if not ensemble:
         return
 
-    from rasa.core.policies.form_policy import FormPolicy
-
-    suited_policies_for_forms = (FormPolicy, RulePolicy)
+    suited_policies_for_forms = (RulePolicy)
```
joejuzl, line 724: These parenthesis are doing nothing. You would need a comma to make it a tuple, but probably a tuple is not needed now.

joejuzl (changes requested):

joejuzl:
reminder to tackle `rasa.core.policies.memoization.py:L87`:

```
"f""Please configure the max history in your configuration file, ""
f""currently 'max_history' is set to old default value of ""
f""'{max_history}'. If you want to have infinite max history ""
f""set it to 'None' explicitly. We will change the default value of ""
 f""'max_history' in the future to 'None'."","
```

rasa/core/policies/ensemble.py
kedz: This was accidentally added when I was resolving merge conflicts with main. I have removed it.

rasa/core/policies/ensemble.py
kedz: Removed the tuple and just now check that there is a rule policy instance in the config to handle forms.

rasa/core/policies/ensemble.py
kedz: As far as I can tell yes. Its only being used [here](https://github.com/RasaHQ/rasa/blob/08bf76bd70cc58784556e69c60b03570149f1119/rasa/core/policies/ensemble.py#L616-L621) as a condition in an if statement:
```python
if form_confidence is not None and self._is_not_mapping_policy(
    best_policy_name, best_confidence[0]
):
    # if mapping didn't win, check form policy predictions
    if form_confidence > best_confidence:
        best_policy_name = form_policy_name
```

Because we are also removing `FormPolicy`, `form_confidence` will always be `None` so this if condition can never evaluate to `True` anymore.

rasa/core/policies/ensemble.py
kedz: Great point! It can't happen now. Removing it.

tests/core/test_agent.py
kedz: Removing parametrization.

kedz:
/modeltest

```yaml
include:
 - dataset: ["all-core"]
   config: ["all-core"]
```

kedz:
The CI tests are now passing but our core regression tests are failing. 

Looking into this some more, I see that the problem is unrelated to this PR. It seems that `required_slots` in form definitions was optional for domain files before 2.6. After 2.6 not including was still supported but deprecated. This week, support for not including it got dropped in main as part of the move to 3.0. Now including `required_slots` is necessary or YAML validation will fail. 

Unfortunately, the two repos we run core regression tests on (finanicial-demo and helpdesk-assistant) have not had their domain files updated. I have created PRs for each of them to add this field ([here](https://github.com/RasaHQ/financial-demo/pull/146) and [here](https://github.com/RasaHQ/helpdesk-assistant/pull/55)). Once they merged core regression tests should work again.

Although none of the core policy regression tests include any of the deprecated policies removed in this PR so the core regression tests should not be affected by any of the  changes here (in theory, I wanted to run this to confirm which is how I discovered they were broken). We could merge if this PR is blocking other work and the core regression tests will get fixed when those other repos update their domain files.

rasa/core/policies/ensemble.py
```diff
@@ -790,20 +661,16 @@ def _check_policy_for_forms_available(
...
 
     if domain.form_names and not has_policy_for_forms:
         raise InvalidDomain(
-            "You have defined a form action, but have neither added the "
-            f"'{RulePolicy.__name__}' nor the '{FormPolicy.__name__}' (deprecated) to "
-            f"your policy ensemble. Either remove all forms from your domain or add "
-            f"the missing policy to your policy configuration."
+            "You have defined a form action, but have not added the "
+            f"'{RulePolicy.__name__}' to your policy ensemble. Either "
+            f"remove all forms from your domain or add the missing policy "
```
samsucik, line 672:
```suggestion
            f"remove all forms from your domain or add the '{RulePolicy.__name__}' "
```

tests/core/test_config.py
```diff
@@ -223,51 +222,35 @@ def check_memoization(p):
...
-        assert p.core_threshold == 0.7
-        assert p.priority == 2
-
-    def check_form(p):
+    def check_rule(p):
         assert p.priority == 1
 
     ensemble_dict = {
         "policies": [
             {"max_history": 5, "priority": 3, "name": "MemoizationPolicy"},
```
samsucik, line 230: Might it make sense to use the new default for `max_history` here?

tests/core/test_config.py
```diff
@@ -223,51 +222,35 @@ def check_memoization(p):
...
-                check_form(policy)
+            if isinstance(policy, RulePolicy):
+                check_rule(policy)
             else:
                 check_memoization(policy)
-        elif isinstance(policy, FallbackPolicy):
-            check_fallback(policy)
 
 
 TEST_MIGRATED_MAPPING_POLICIES = [
```
samsucik, line 256: Actually, does it make sense to also remove this test (and all the stuff around the associated `rasa.core.config.migrate_mapping_policy_to_rules`? Assuming that Rasa 3.x won't know `MappingPolicy`, this migration code shouldn't work anyway (I think) 🤔  I guess this is another topic to discuss and make a decision on: Whether to leave the 1.x->2.x migration code in 3.x at all (I think not, as it might require some bits of code that we want to remove in 3.x, such as the implementation of the removed policies).

tests/core/test_ensemble.py
```diff
@@ -184,152 +174,6 @@ def test_policy_priority():
...
-            ActionExecuted(ACTION_LISTEN_NAME),
-            utilities.user_uttered(USER_INTENT_RESTART, 1),
-        ],
-        [
-            ActionExecuted(ACTION_LISTEN_NAME),
-            utilities.user_uttered(USER_INTENT_RESTART, 1),
-        ],
-    ],
-)
-def test_mapping_wins_over_form(events: List[Event]):
```
samsucik, line 233: Just reading the names of the tests being removed, we might want to create a test case that checks RulePolicy in this way, right?

tests/core/test_training.py
```diff
@@ -69,10 +69,8 @@ async def test_training_script_without_max_history_set(
     agent = Agent.load(tmpdir)
     for policy in agent.policy_ensemble.policies:
         if hasattr(policy.featurizer, "max_history"):
-            if type(policy) == FormPolicy:
-                assert policy.featurizer.max_history == 2
-            elif type(policy) == MemoizationPolicy:
-                assert policy.featurizer.max_history == OLD_DEFAULT_MAX_HISTORY
+            if type(policy) == MemoizationPolicy:
+                assert policy.featurizer.max_history == DEFAULT_MAX_HISTORY
```
samsucik, line 73: I think we might need a more sophisticated check since the value is `None` (I don't think that `== None` works, only `is None`?).

samsucik: Perhaps sth like:
```python
if DEFAULT_MAX_HISTORY is None:
    assert policy.featurizer.max_history is DEFAULT_MAX_HISTORY
else:
    assert policy.featurizer.max_history == DEFAULT_MAX_HISTORY
```

tests/core/test_policies.py
```diff
@@ -1046,9 +400,13 @@ def test_get_training_trackers_for_policy(
...
+class PolicyWithoutInitKwargs(Policy):
+    def __init__(self, *args: Any) -> None:
+        pass
+
+    def persist(self, _) -> None:
+        pass
+
+    @classmethod
+    def _metadata_filename(cls) -> Text:
+        return "no_finetune_policy"
```
samsucik, line 412: Looks like this policy isn't used anywhere, did you mean to add some more code?

samsucik (changes requested):
Overall looks good but I think we'll need to talk about how we deal with links to removed components, as well as with migration automation code that isn't for the current major version (in this case 3.x).

CHANGELOG.mdx
```diff
@@ -667,7 +667,7 @@ https://github.com/RasaHQ/rasa/tree/main/changelog/ . -->
 
 ### Improvements
 - [#7520](https://github.com/rasahq/rasa/issues/7520): Improve the warning in case the [RulePolicy](policies.mdx#rule-policy) or the deprecated 
-  [MappingPolicy](policies.mdx#mapping-policy) are missing
+  `MappingPolicy` are missing
```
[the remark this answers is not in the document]
kedz: I was removing them at @wochinge's suggestion [here](https://github.com/RasaHQ/rasa/pull/9120#issuecomment-880014170). But I can start a discussion with CSE about this.

tests/core/test_config.py
kedz: So I looked at the code and I believe the migration code doesn't use the deprecated policies it just converts markdown to yaml so it wouldn't break anything in theory to keep it around (I think). I can also start that discussion on this.

tests/core/test_config.py
kedz: Adding a test for that.

tests/core/test_training.py
kedz: Maybe I'm missing something but I think `==` is fine here:

```shell
Python 3.7.10 (default, Feb 26 2021, 10:16:00)
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> None == None
True
>>> None is None
True
>>> None == 0
False
>>> None == -1
False
```

tests/core/test_policies.py
kedz: Ahh that should be removed. Its leftover code that should have been deleted when I was resolving a merge conflict with main. Removing it now.

tests/core/test_ensemble.py
kedz: I'm not sure. This is comparing `MappingPolicy` against `FormPolicy`, neither of which exist anymore. Not sure what I would be comparing `RulePolicy` to in this case?

tests/core/test_config.py
kedz: @wochinge says the migration code should go, but I am going to create a separate issue for removing it because it requires removing things from the cli as well 🍝 🗼 .

tests/core/test_training.py
samsucik: Okay, I've done some more reading and now agree with the `==` 🙂

tests/core/test_config.py
samsucik: Makes sense 🙂

CHANGELOG.mdx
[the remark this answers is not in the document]
samsucik: Okay, that's useful to know. Looking at Akela's thoughts: If the link checker allows, perhaps we could leave the links as they are and create a follow-up issue for when we've got legacy 2.x docs? Or we could delete the links and create a follow-up issue? 🤔

samsucik (commented):
Looks good, let's deal with the links and then we're good to go I think 🙂

CHANGELOG.mdx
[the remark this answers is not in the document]
kedz: I'm not sure how to disable the link checker so I will just aggregate the links removed in this and the other 3.0 deprecations I'm working on into an issue and we can figure out what to do with them there.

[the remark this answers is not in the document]
samsucik: Makes sense 🙂

joejuzl (approved):
I love the smell of deleted code in the morning 🥀

kedz:
@samsucik I created an issue [here](https://github.com/RasaHQ/rasa/issues/9169) to track dead docs links. With that in place can I merge?

samsucik:
@kedz I think so 😉

samsucik (approved):
🚀

PR #9173, opened 2021-07-20
Remove `processor` replicated logic in `rasa.core.test`

### LINKED ISSUE #9135: Get rid of model access in `rasa.core.test`
**Description of Problem**:
The `rasa.core.test` module has several paths where it directly access attributes on the `Agent` or even lower. This has the disadvantage that the test code is really tied to the specific model architecture. With the revamp we want to reduce these ties. This will help us to keep the interface of `Graphprocessor` small and explicit and will reduce the amount of breaking things when we start to implement the actual graph.

[Related implementation proposal page](https://www.notion.so/rasa/rasa-core-test-192bb2cb16e6430ba08c5bf53c78d6d2)

Currently the following pieces in the code directly access `Agent` stuff which we don't want it to access
1. [Tokenizing messages for the entity evaluation](https://github.com/RasaHQ/rasa/blob/60d5b32371d1764faf23edcefa8231c055869897/rasa/core/test.py#L591): **We can skip this case for now** ⚠️ We handle it once we have handled the [prediction metadata task](https://www.notion.so/rasa/Add-tokens-to-events-scope-implem-proposal-563d82d27bc64a3ea8c66ae2971b945a) which adds the token to the output of every parsed user message.
2. [Determining whether we should predict another action](https://github.com/RasaHQ/rasa/blob/60d5b32371d1764faf23edcefa8231c055869897/rasa/core/test.py#L778): My idea was to change this check to raise an exception `ActionLimitReached` when we call `processor.predict_next_action` instead of having a separate method which requires callers to know the logic
3. [2nd part of determining whether predict another action](https://github.com/RasaHQ/rasa/blob/60d5b32371d1764faf23edcefa8231c055869897/rasa/core/test.py#L806): Same we can handle this with an exception as well in the test module

**Overview of the Solution**:
I outlined some proposed ways of getting rid of these accesses. Anything else which doesn't require an additional method on the `GraphProcessor` is also fine.

**Definition of Done**:
- [x] The described locations 2) and 3) no longer replicate logic from the `processor`

ancalita:
**Proposed changes**:
- Fix #9135 


ancalita:
@alwx I was wondering if I should wrap this [section](https://github.com/RasaHQ/rasa/blob/60d5b32371d1764faf23edcefa8231c055869897/rasa/core/test.py#L674) in a try-except block too?

Also I've noticed this private `processor` [method](https://github.com/RasaHQ/rasa/blob/60d5b32371d1764faf23edcefa8231c055869897/rasa/core/processor.py#L643) - should this be adapted too, since I think there's some overlap with `predict_next_action` now?

rasa/core/test.py
```diff
@@ -795,10 +786,6 @@ async def _predict_tracker_actions(
                         "confidence": prediction.max_confidence,
                     }
                 )
-                should_predict_another_action = processor.should_predict_another_action(
-                    action_executed_result.action_predictions[0]
-                )
-                num_predicted_actions += 1
```
alwx, line 801: Cool to see you getting rid of this counter.

alwx:
@ancalita honestly, I would skip adapting the `processor` method for now (and I think it makes not much sense doing it as a part of this task) — because we'll be doing many more changes in the `test` module later and would definitly change a lot of things in `_run_action_prediction`, `_collect_action_executed_predictions`, etc

rasa/core/test.py
```diff
@@ -649,13 +649,13 @@ def _collect_action_executed_predictions(
...
-        prediction = PolicyPrediction([], policy_name=None)
-        predicted_action = "circuit breaker tripped"
-    else:
+    try:
         predicted_action, prediction, policy_entity_result = _run_action_prediction(
             processor, partial_tracker, expected_action
         )
+    except ActionLimitReached:
+        prediction = PolicyPrediction([], policy_name=None)
+        predicted_action = "circuit breaker tripped"
```
alwx, line 658: Yes, I think it makes to apply this try-except to the case when `action_unlikely_intent` is predicted and another prediction needs to be made (line 674)

alwx (changes requested):

ancalita:
@alwx A couple more things I just thought of:

- should I update the docstring for `processor.predict_next_action` now that it raises an exception?
- does the exception raised here need a message argument?
- should I add a test for this adaptation in `test_processor.py`?

alwx:
> should I update the docstring for processor.predict_next_action now that it raises an exception?

Yes, the docstring should definitely be updated as well.

> does the exception raised here need a message argument?

I think it might be useful (but probably not strictly necessary because it seems more like an improvement in the context of this issue)

> should I add a test for this adaptation in test_processor.py?

Would be amazing to add a test for this exception!

rasa/core/processor.py
```diff
@@ -374,8 +376,33 @@ def predict_next_action(
...
+        for e in reversed_events:
+            if isinstance(e, ActionExecuted):
+                if e.action_name in (ACTION_LISTEN_NAME, ACTION_SESSION_START_NAME):
+                    break
+                num_predicted_actions += 1
+
+        if self.is_action_limit_reached(
+            num_predicted_actions, should_predict_another_action
+        ):
+            raise ActionLimitReached(
```
joejuzl, line 402: Don't we need to handle this in `_predict_and_execute_next_action` now?

rasa/core/processor.py
```diff
@@ -374,8 +376,33 @@ def predict_next_action(
...
+            tracker.latest_action_name
+        )
+        reversed_events = list(tracker.events)[::-1]
+        num_predicted_actions = 0
+
+        for e in reversed_events:
+            if isinstance(e, ActionExecuted):
+                if e.action_name in (ACTION_LISTEN_NAME, ACTION_SESSION_START_NAME):
+                    break
+                num_predicted_actions += 1
```
joejuzl, line 397: Isn't this duplicate logic from what's in `_predict_and_execute_next_action`?

joejuzl (changes requested):

rasa/core/processor.py
ancalita: This [implementation proposal section](https://www.notion.so/rasa/rasa-core-test-192bb2cb16e6430ba08c5bf53c78d6d2) indicates to adapt `processor.predict_next_action` - I think mainly because `_run_action_prediction` in `rasa.core.test` module uses this method and not `processor._predict_and_execute_next_action`.

rasa/core/processor.py
ancalita: It looks similar but not quite - the while loop in `_predict_and_execute_next_action` also has a check for `_should_handle_message` which `predict_next_action` doesn't need. 

I did ask Alex[ if I should also adapt this method ](https://github.com/RasaHQ/rasa/pull/9173#issuecomment-883415430) to reduce duplicate logic from `predict_next_action` however[ it seems out of scope for this issue](https://github.com/RasaHQ/rasa/pull/9173#issuecomment-883936445).

rasa/core/processor.py
joejuzl: Sure,  but `_predict_and_execute_next_action` calls `predict_next_action` so we need to make sure it still works correctly after this change.

joejuzl: Also this ticket predates that implementation proposal right? It was created because of the revamp so we can make the interface to `GraphProcessor` smaller. (Although it of course overlaps)

rasa/core/processor.py
joejuzl: but since `_predict_and_execute_next_action` is calling `predict_next_action` we don't want the same things to be done twice right? Also we are now making changes to the processor, not just the test module, so we need to think about the broader implications.

rasa/core/processor.py
ancalita: How would you test that `_predict_and_execute_next_action` works correctly after this change, apart from using CI checks?

rasa/core/processor.py
ancalita: I'm happy to submit a different issue for this specific processor method - the original issue for this PR was focused on making sure `rasa.core.test` doesn't duplicate/expose logic from the processor. 
Otherwise, happy to accept suggestions how to adapt `processor._predict_and_execute_next_action`.

rasa/core/processor.py
ancalita: > this ticket predates that implementation proposal right?

The ticket doesn't predate the implementation proposal -  GH indicates that the ticket was created on Jul 15, that was last week Thu after the implementation proposal was submitted for review.
I also checked in advance the usages of `_predict_and_execute_next_action` and this is not used anywhere in `rasa.core.test`.

joejuzl: Sorry that was a miss-understanding - I was referring to the Test Stories implementation proposal which also lists this as a change! (But I see that you were actually linking to the Architecture revamp 👍 )

rasa/core/processor.py
ancalita: Ok, method adapted 🎉  I tried first with raising a new exception `if not should_predict_another_action`, but that broke multiple unit tests in both `test_processor` and `test_evaluation` 🙈  So I arrived at the current state. Let me know what you think.

rasa/core/processor.py
```diff
@@ -374,8 +376,23 @@ def predict_next_action(
...
+             The index of the next action and prediction of the policy.
+
+        Raises:
+            ActionLimitReached if the limit of actions to predict has been reached.
         """
+        should_predict_another_action = self.should_predict_another_action(
+            tracker.latest_action_name
+        )
+
+        if self.is_action_limit_reached(tracker, should_predict_another_action):
```
joejuzl, line 391: I still think this is confusing as we raise `ActionLimitReached` even if it's not been reached and `should_predict_another_action` is just `False`. I think the only reason this doesn't cause problems later is because we _also_ check `self.should_predict_another_action` in `_run_action`. Otherwise we would be tripping `on_circuit_break` every time we got `action_listen`!

But honestly looking through all the code, it will be quite tricky to get all of this to work cleanly - and as we said before, all of this processor implementation will go!
I think a good compromise would be to rename  `ActionLimitReached` to something more generic such as `StopPredictionActions` and then put a comment above `if self.on_circuit_break:` stating that this will only be raised in this loop if action limit is reached.

How does that sound?

tests/core/test_processor.py
```diff
@@ -1297,3 +1298,33 @@ def test_predict_next_action_with_hidden_rules():
     action, prediction = processor.predict_next_action(tracker)
     assert isinstance(action, ActionListen)
     assert not prediction.hide_rule_turn
+
+
+def test_predict_next_action_raises_exception(domain: Domain):
```
joejuzl, line 1303: Can we be specific about the exception?

joejuzl: Also would be good to test the case where the exception comes from the last action being `action_listen`, not from the limit being reached.

joejuzl (changes requested):
Thanks for making the changes 👍 
Added a couple of comments

tests/core/test_processor.py
ancalita: > Can we be specific about the exception?
Yup sure, will add the exception name; was afraid of making the line too long.

> test the case where the exception comes from the last action being action_listen
I don't think the exception ever gets triggered when the last action is action_listen, because of the `and` in this [line](https://github.com/RasaHQ/rasa/blob/ac1197b3f80071bb213d4fa66d5d24b4fc01e30e/rasa/core/processor.py#L640) - if last action is action_listen, then `should_predict_another_action` is `False` which will make `is_action_limit_reached` False as well.

rasa/core/processor.py
joejuzl: Sorry disregard this - I miss-read the implementation of `is_action_limit_reached`

joejuzl (approved):
Nice one!  🚀 
Thanks for dealing with my confusion 😅

alwx (approved):
Looks nice and clear! Great work, and thanks a lot for doing this 
🚀

Issue #6487, opened 2020-08-26
Raise proper user warnings for backwards compatibility code


m-vdb:
([Original slack thread](https://rasa-hq.slack.com/archives/C36SS4N8M/p1598437115179100))

Looking for "compatibility" in the codebase shows several examples of code that are still here for backwards compatibility purposes (until when?) but the users are not warned about them. If possible, we should introduced warnings for each of these communicate that they'll be removed in Rasa 3.0

List of things to address:
- [in rasa.model_testing](https://github.com/RasaHQ/rasa/blob/ac1197b3f80071bb213d4fa66d5d24b4fc01e30e/rasa/model_testing.py#L130-L131)
- [in rasa.model_training](https://github.com/RasaHQ/rasa/blob/ac1197b3f80071bb213d4fa66d5d24b4fc01e30e/rasa/model_training.py#L54-L55)
- [in rasa.cli.x](https://github.com/RasaHQ/rasa/blob/ac1197b3f80071bb213d4fa66d5d24b4fc01e30e/rasa/cli/x.py#L506-L515)
- [in rasa.core.utils](https://github.com/RasaHQ/rasa/blob/ac1197b3f80071bb213d4fa66d5d24b4fc01e30e/rasa/core/utils.py#L22)
- [in rasa.core.events](https://github.com/RasaHQ/rasa/blob/ac1197b3f80071bb213d4fa66d5d24b4fc01e30e/rasa/shared/core/events.py#L254-L261)

wochinge:
@m-vdb I remember you did some changes for this - can this issue be closed?

m-vdb:
no I haven't done anything for this. If you search "compatibility" in all the python files, you'll see examples of code that is backwards compatible, but it's not clear if some of these should actually be deprecated. My reasoning is that we can only keep backward compatibility for so long, so we might as well tell users things are deprecated as soon as we can? What do you think?

wochinge:
Fully agree 👍🏻

PR #9187, opened 2021-07-23
Remove outdated backwards compat code

m-vdb:
Fixes https://github.com/RasaHQ/rasa/issues/6487

**Proposed changes**:
- The code changed in this PR was introduced for backwards compat reason, but has since evolved. In all the cases I think we can get rid of the various behaviours introduced (argumenting on each of them in the PR to make the review easier)


rasa/cli/x.py
```diff
@@ -493,28 +493,14 @@ def run_locally(args: argparse.Namespace) -> None:
...
-            else:
-                raise
+        local.main(
+            args,
+            project_path,
+            args.data,
+            token=rasa_x_token,
+            config_path=config_path,
+            domain_path=domain_path,
+        )
```
m-vdb, line 503: Reason for the change: 3.0 will be supported by 1.x versions of Rasa Enterprise. Probably would have been better to warn the user in advance when introducing the backwards compatibility change. I think it's low risk to remove

rasa/core/utils.py
```diff
@@ -19,8 +19,6 @@
 from rasa.constants import DEFAULT_SANIC_WORKERS, ENV_SANIC_WORKERS
 from rasa.shared.constants import DEFAULT_ENDPOINTS_PATH
 
-# backwards compatibility 1.0.x
-# noinspection PyUnresolvedReferences
```
m-vdb, line 23: refered to [an import that no longer exists](https://github.com/RasaHQ/rasa/commit/07e2b5def944d146c5786f6d35ea23313c55ce04#diff-9c3ef27acfec367531f04a89510310550385f4d89daa83a27df0412a1c23d793R23-R27)

rasa/model_testing.py
```diff
@@ -127,10 +127,6 @@ def test_core_models(
     )
 
 
-# backwards compatibility
-test = rasa.test
```
m-vdb, line 131: Was introduced in [this PR](https://github.com/RasaHQ/rasa/pull/8141). I think that we added that in the PR for backwards compatibility _before_ we actually renamed the file, which made this change a no-op from the beginning. If we want to be a bit extra careful we could consider adding the code below, but I don't think it's necessary:
```suggestion
def test(*args, **kwargs):
    # raise a deprecation warning that this will be removed in 4.x
    return rasa.test(*args, **kwargs)
```

rasa/model_training.py
```diff
@@ -51,10 +51,6 @@ class TrainingResult(NamedTuple):
     code: int = 0
 
 
-# backwards compatibility
-train = rasa.train
```
m-vdb, line 55: Was introduced in [this PR](https://github.com/RasaHQ/rasa/pull/8141). I think that we added that in the PR for backwards compatibility _before_ we actually renamed the file, which made this change a no-op from the beginning. If we want to be a bit extra careful we could consider adding the code below, but I don't think it's necessary:
```suggestion
def train(*args, **kwargs):
    # raise a deprecation warning that this will be removed in 4.x
    return rasa.train(*args, **kwargs)
```

rasa/shared/core/events.py
```diff
@@ -247,19 +247,7 @@ def __init__(
...
-        # Needed for compatibility with Rasa versions <1.4.0. Previous versions
-        # of Rasa serialized trackers using the pickle module. For the moment,
-        # Rasa still supports loading these serialized trackers with pickle,
-        # but will use JSON in any subsequent save operations. Versions of
-        # trackers serialized with pickle won't include the `_metadata`
-        # attribute in their events, so it is necessary to define this getter
-        # in case the attribute does not exist. For more information see
-        # CHANGELOG.rst.
-        return getattr(self, "_metadata", {})
+        self.metadata = metadata or {}
```
m-vdb, line 250: Would also probably have been better to raise a deprecation warning sooner. I think it's low risk to introduce this in 3.0, since people should already be using 2.x

m-vdb:
I need to figure out why tests are currently failing

wochinge:
@m-vdb Feel free to request review from me then 👍🏻 (instead of the entire squad)

m-vdb:
@wochinge after some search I think I understand why the tests are currently failing, but I don't know how to fix this yet. Maybe you have an idea. The different event instances should have a `metadata` attribute, but for some reason, after they're deserialised via `jsonpickle`, this doesn't work (hence those attribute errors). I suspect some sort of cache in `jsonpickle`. But I'm not quite sure here 🤔 does this sound familiar?

[These tests](https://github.com/RasaHQ/rasa/blob/4a42985909413b0a599d8db7888ace8e246b0087/tests/shared/core/test_trackers.py#L144-L152) are failing, and at line 149 we deserialise events through a helper from conftest.

wochinge (commented):
Re the failing tests:
1. The old pickled dialogue files don't have the `metadata` property
2. Jsonpickle reloads the pickled dialogue without using the constructor (e.g. [here](https://docs.python.org/3/library/pickle.html#pickling-class-instances), "When a class instance is unpickled, its __init__() method is usually not invoked.")
3. There is no attribute `metadata` as we have never assigned this attribute in the first place

As far as I can see from the `events` and `tracker_store` modules we are never restoring a tracker using `pickle` or `jsonpickle` which in turn means that the jsonpickling is no functionality which we need to test. You could 
- re-dump the tracker states via `json.dumps(dialogue.as_dict())` and then use `json.loads` to the load the tracker and execute the (actual) test. 
- drop the persisted dialogues and inline them as list of events in the code. This way they are closer to the test code and we don't use code to serialize / deserialize trackers which we don't use anywhere else (I think I prefer this one)

m-vdb:
@wochinge I've addressed your comment regarding test data in this commit: f6c1026. One thing that I wasn't sure about: where to put the test data. I've moved it in `conftest.py` as other tests depend on it (as you'll see in the commit).

rasa/cli/x.py
m-vdb: 🤔 looking at the [compatibility matrix](https://rasa.com/docs/rasa-x/changelog/compatibility-matrix) it seems that we can go ahead with the change. 0.35 and below wouldn't work anyway with 3.0. The current discussion is about the latest Rasa X version supporting either 3.0 only or 3.0+2.8 (not 2.2.4).

tests/conftest.py
```diff
@@ -74,14 +81,269 @@
 }
 
 
-TEST_DIALOGUES = [
-    "data/test_dialogues/default.json",
-    "data/test_dialogues/formbot.json",
-    "data/test_dialogues/moodbot.json",
-]
+TEST_DEFAULT_DIALOGUE = Dialogue(
```
wochinge, line 84: Optional: due to the amount of lines. How about moving it a module called `test_dialogues`? 

Optional2: I think we can skip this whole `Dialogue` stuff and just have list of `Event`s here. We can then create the trackers via `DialogueStateTracker.from_events`. I made this optional as it goes beyond the scope of the PR. In my opinion we can remove the entire `Dialogue` object as it has no benefit over using `DialogueStateTracker` / the events directly

wochinge (approved):

tests/conftest.py
m-vdb: I'd go with your first option here, as it's a bit more work and I think it's a bit beyond the scope. I think `tests/dialogues.py` is enough, because naming it `test_dialogues` could be misleading (it's not a test file and shouldn't start with `test_`)

wochinge: sounds good 👍🏻

Issue #9133, opened 2021-07-15
Add fingerprinting capability to Core training data classes.


joejuzl:
**Description of Problem**:
The new architecture will rely on the ability to fingerprint all data that passes through the training graph in-order to determine what needs to be re-run when re-training.
We can start adding fingerprint capabilities to existing data types in preparation.

The scope this issue is core training data objects.

**Overview of the Solution**:

A fingerprint is a unique, stable and small representation of a piece of data. For example an MD5 hash for text. (Note: the python builtin `hash()` is not stable!)
Objects that are returned from graph nodes during training need to have a method `fingerprint` which returns a fingerprint for the object.

Below is a list of all the objects covered by this issue. Some may already have fingerprinting capability, so it is also part of this issue to confirm it works correctly, make sure there is a unit test, and make sure it is accessible from `.fingerprint()`.

- `rasa.shared.core.training_data.structures.StoryGraph`
    - This holds all the stories training data
    - A `StoryGraph` is made up of `StoryStep`s which in turn have `Checkpoint`s and `Event`s.
    - However `StoryGraph` already has a `fingerprint` method which uses the `YAMLStoryWriter` along with `rasa.shared.utils.io.deep_container_fingerprint` which should be sufficient
- `rasa.shared.core.generator.TrackerWithCachedStates`
    - Once the stories are featurized they become `TrackerWithCachedStates`
    - This contains:
        - `events: List[Event]`
        - `slots: Iterable[Slot]`
        - `_states_for_hashing: Deque[FrozenSet[Tuple[Text, FrozenSet[Tuple[Text, Tuple[Union[float, Text]]]]]]]`
    - `_states_for_hashing` is how the past states are cached in the tracker. As it is made up of numbers and text in fairly standard python structures it should be fine to fingerprint, although this functionality needs to be added.
- `rasa.shared.core.events.Event`
    - A single step in a tracker.
    - There shouldn't be anything crazy in any of the events that is hard to fingerprint
    - Although many different subclasses that all contain different things.
    - Worst case scenario we could use `as_story_string` which already gives a unique string representation for each event subclass and is a required method.
- `rasa.shared.core.slots.Slot`
    - A piece of remembered information, held in the tracker.
    - Similar to `Event` as there are lots of subclasses which are all quite different.
    - required method `as_feature` could be used for the fingerprint.
- `rasa.shared.core.domain.Domain`
[...]

PR #9189, opened 2021-07-23
Implement `fingerprint` method in Core classes

ancalita:
**Proposed changes**:
- Implements #9133 
- I implemented the fingerprint method for `Event`, `Slot` and `TrackerWithCachedStates`.
- This already existed in `Domain` and `StoryGraph` so left untouched.
- Instead added some missing tests for `Domain.fingerprint()` - `StoryGraph` had an existing test in `test_graph.py`


tests/shared/core/test_generator.py
```diff
@@ -18,3 +24,59 @@ def test_subsample_array():
...
+                {"id": 2, "name": "greet", "confidence": 0.9604260921478271},
+                {"id": 1, "name": "goodbye", "confidence": 0.01835782080888748},
+                {"id": 0, "name": "deny", "confidence": 0.011255578137934208},
+            ],
+        },
+    )
+    tracker.update(event, domain=domain)
+    container = deque()
+    container.append(tracker.freeze_current_state(domain.get_active_state(tracker)))
+    tr._states_for_hashing = container
```
ancalita, line 73: Not very pleased with accessing a private attribute here, I didn't know how to update the cached states otherwise🤔

tests/shared/core/test_events.py
```diff
@@ -614,3 +621,160 @@ def test_event_executed_comparison(
...
+                    },
+                },
+            },
+        ),
+        SessionStarted(),
+        ActionExecuted(action_name="action_listen"),
+        AgentUttered(),
+    ],
+)
+def test_event_fingerprint_consistency_across_runs(event: Event):
```
ancalita, line 741: Here I wasn't sure if to use `rasa.shared.utils.common.all_subclassess` to iterate through all the subclassess of Event in case we add Event types in the future? This would imply swapping the long list in `pytest.mark.parametrize` with a long if-elif because some events have some specific arguments that need to be passed.

rasa/shared/core/generator.py
```diff
@@ -217,6 +217,40 @@ def update(self, event: Event, skip_states: bool = False) -> None:
...
+
+        Returns:
+            fingerprint of the tracker
+        """
+        data = []
+        sender_id_hash = rasa.shared.utils.io.get_text_hash(self.sender_id)
+        data.append(sender_id_hash)
+
+        if self.domain:
+            data.append(self.domain.fingerprint())
```
joejuzl, line 231: @twerkmeister added functionality for `.fingerprint` to `deep_container_fingerprint` in https://github.com/RasaHQ/rasa/pull/9175.

rasa/shared/core/generator.py
```diff
@@ -217,6 +217,40 @@ def update(self, event: Event, skip_states: bool = False) -> None:
...
+
+        if self.slots:
+            for slot in self.slots.values():
+                data.append(slot.fingerprint())
```
joejuzl, line 235: Maybe `data` should be a `dict` rather than a list?. We probably shouldn't rely on `slots.values()` always giving the same order, and we don't want that to change the fingerprint.

rasa/shared/core/generator.py
```diff
@@ -217,6 +217,40 @@ def update(self, event: Event, skip_states: bool = False) -> None:
 
             self._append_current_state()
 
+    def fingerprint(self) -> Text:
```
joejuzl, line 220: Can any of this functionality be extracted into the parent class `DialogueStateTracker`?

rasa/shared/core/generator.py
```diff
@@ -217,6 +217,40 @@ def update(self, event: Event, skip_states: bool = False) -> None:
...
+        data.append(sender_id_hash)
+
+        if self.domain:
+            data.append(self.domain.fingerprint())
+
+        if self.slots:
+            for slot in self.slots.values():
+                data.append(slot.fingerprint())
+
+        if self._states_for_hashing:
```
joejuzl, line 237: I'm not an expert on this area of the code, but I wonder if we need to fingerprint this stuff if we just fingerprinted `self.events` instead (maybe in the base class)?

rasa/shared/core/slots.py
```diff
@@ -149,8 +149,23 @@ def persistence_info(self) -> Dict[str, Any]:
             "influence_conversation": self.influence_conversation,
         }
 
+    def fingerprint(self) -> Text:
+        """Returns a unique hash for the slot which is stable across python runs.
+
+        Returns:
+            fingerprint of the slot
+        """
+        data = [
```
joejuzl, line 158: I'm not sure this is enough. What about things like `influence_conversation`? and for subclasses like `CategoricalSlot` we have things like `self.values`.

tests/shared/core/test_domain.py
```diff
@@ -1468,3 +1468,88 @@ def test_domain_invalid_yml_in_folder():
...
+def test_domain_fingerprint_consistency_across_runs():
+    domain1 = Domain.from_yaml(
+        """
+         version: "2.0"
+         intents:
+         - greet
+         - goodbye
+         actions:
+         - action_test
+         """
```
joejuzl, line 1482: What about slots, entities, forms etc.?

tests/shared/core/test_generator.py
joejuzl: Does the cached states do anything other than just speed up certain methods? I.e. does it actually hold any unique information not encapsulated by the events and slots etc.?

joejuzl (changes requested):

rasa/shared/core/generator.py
ancalita: Seems that `sender_id` and `slots` are common attributes - should I create a fingerprint of `DialogueStateTracker` using these attributes? Then how do I combine the resulting hash of `super().fingerprint()` in `TrackerWithCachedStates` with the its other attributes (`domain`, `_states_for_hashing`)?

rasa/shared/core/generator.py
```diff
@@ -217,6 +217,40 @@ def update(self, event: Event, skip_states: bool = False) -> None:
...
             self._append_current_state()
 
+    def fingerprint(self) -> Text:
+        """Returns a unique hash for the tracker which is stable across python runs.
+
+        Returns:
+            fingerprint of the tracker
+        """
+        data = []
+        sender_id_hash = rasa.shared.utils.io.get_text_hash(self.sender_id)
```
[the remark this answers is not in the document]
ancalita: Ah true, no need to hash it then and there - I can just append it to `data` directly.

rasa/shared/core/generator.py
ancalita: I can see the PR hasn't been merged it so I can't use this added functionality at the moment. 
Would it be fine to actually change this to:
```
data.append(self.domain.as_dict())
```

joejuzl: it's only the 4 lines in this file: https://github.com/RasaHQ/rasa/pull/9175/files#diff-d0c59274cb6aad666a5addf9f758510ba9a2ecdcb48b08a16b904f90c980e017R221

So probably easier to just copy the changes across to this branch.

tests/shared/core/test_generator.py
ancalita: At a first look [here](https://github.com/RasaHQ/rasa/blob/2039cef57780f034da22efcf8c1534d12f3d0ac3/rasa/shared/core/domain.py#L1078) - seems that most of the information here should be captured in events, domain, slots. I can remove this and instead add `self.events` to fingerprintable data?

tests/shared/core/test_events.py
ancalita: Yup, can do - however a warning that not all subclassess exist in this list list, for example internal events such as `WronglyPredictedAction`. I can filter these out in the test.

tests/shared/core/test_domain.py
ancalita: I'll add them 👍

rasa/shared/core/generator.py
ancalita: Yup, removed this part and fingerprinted `self.events` instead.

rasa/shared/core/slots.py
ancalita: I've changed the fingerprinting by using `persistence_info` which should contain the extra bits from certain subclassess (such as `self.values` for `CategoricalSlot`)

rasa/shared/core/generator.py
```diff
@@ -217,6 +217,26 @@ def update(self, event: Event, skip_states: bool = False) -> None:
...
+    def fingerprint(self) -> Text:
+        """Returns a unique hash for the tracker which is stable across python runs.
+
+        Returns:
+            fingerprint of the tracker
+        """
+        data = {"sender_id": self.sender_id}
+
+        if self.domain:
+            data.update(self.domain.as_dict())
```
joejuzl, line 229: We can just add the whole `domain` as it's fingerprint-able right?

rasa/shared/core/generator.py
joejuzl: Well now we don't need the `_states_for_hashing` can we just move this whole implementation to the superclass?

tests/shared/core/test_domain.py
```diff
@@ -1468,3 +1468,92 @@ def test_domain_invalid_yml_in_folder():
     """
     with pytest.warns(UserWarning, match="The file .* your file\\."):
         Domain.from_directory("data/test_domains/test_domain_from_directory1/")
+
+
+def test_domain_fingerprint_consistency_across_runs():
+    file = """
```
joejuzl, line 1474:
```suggestion
    domain_yaml = """
```

tests/shared/core/test_events.py
```diff
@@ -614,3 +628,168 @@ def test_event_executed_comparison(
...
+
+@pytest.mark.parametrize("event_class", rasa.shared.utils.common.all_subclasses(Event))
+def test_event_subclasses_are_tested(event_class: Type[Event]):
+    subclasses = [event.__class__ for event in tested_events]
+    assert event_class in subclasses
+
+
+@pytest.mark.parametrize(
+    "event",
+    [
```
joejuzl, line 765: Can we just use `tested_events` here again?

joejuzl (changes requested):

rasa/shared/core/generator.py
ancalita: `DialogueStateTracker` doesn't have `self.domain` as attribute. Let me know how best to proceed?

rasa/shared/core/generator.py
ancalita: Would adding to the `data` dict this key-value pair work: `{'domain': self.domain.fingerprint()}`?

joejuzl: You can just do `{'domain': self.domain}`. As `deep_container_fingerprint` can now handle `.fingerprint`

tests/shared/core/test_events.py
ancalita: Ah yes, good idea!

rasa/shared/core/generator.py
joejuzl: True, good catch. 
So the normal way to handle this would be to create `DialogueStateTracker.fingerprint` that fingerprints everything apart from the domain, and then `TrackerWithCachedStates.fingerprint` would add/combine the domain to `super().fingerprint()`. e.g.

```
get_dictionary_fingerprint({"tracker": super().fingerprint(), "domain": self.domain})
```

However, I we may not even need the domain, as I _think_ it's just be used for creating the states. In which case you could just implement it on `DialogueStateTracker.fingerprint`.
@wochinge wdyt?

ancalita: Yikes, it seems `mypy` doesn't allow adding an object to the dictionary, I got this CI error:
```
Dict entry 0 has incompatible type "str": "Domain"; expected "str": "str"  [dict-item]
```
I'll wait for the final decision on whether to move this method to the parent class before making any changes.

rasa/shared/core/events.py
```diff
@@ -317,15 +318,23 @@ def as_dict(self) -> Dict[Text, Any]:
 
         return d
 
+    def fingerprint(self) -> Text:
```
wochinge, line 321: does it make sense to replace / combine this with the existing `__hash__` implementation which we have on every event anyway?

rasa/shared/core/generator.py
wochinge: > However, I we may not even need the domain, as I think it's just be used for creating the states. In which case you could just implement it on DialogueStateTracker.fingerprint.

I agree. As soon as the parent node which provides the domain has a changed output this will lead to a "fingerprint miss" anyway

ancalita: Ok, will remove the domain and move the implementation to the parent class.

rasa/shared/core/events.py
ancalita: Hmm, I steered away from anything using python `hash()` because it's unstable, hence I don't think it's a good idea to use this.

wochinge: yep, but our custom `__hash__` implementations in the events, are stable, aren't they? (they should probably not be called `__hash__` in the first place

ancalita: Not sure, looking at [`UserUttered`](https://github.com/RasaHQ/rasa/blob/071d2ef24760f738590971c044c703cc27efb567/rasa/shared/core/events.py#L474) and `SlotSet` events for example, their custom `__hash__` uses the built-in `hash()` 🤔  While a few other events like `AllSlotsReset` have [static hashes](https://github.com/RasaHQ/rasa/blob/071d2ef24760f738590971c044c703cc27efb567/rasa/shared/core/events.py#L1046).

I'm also confused because the parent class doesn't have a `__hash__` method.

wochinge: What a mess 🙈 The neat thing is that the current `__hash__` implementation excludes all properties which aren't relevant which makes me think of one edge case: Don't get `Event`s assigned timestamps when being read? And wouldn't they differ depending on when you read the training data with the events?

ancalita: > Don't get Events assigned timestamps when being read? And wouldn't they differ depending on when you read the training data with the events?

Yes, if timestamp is not specified, then it will get the current timestamp. Although not sure I understand the 2nd question, when would the events be re-instantiated? I am probably not seeing the full implications of where events get used.

However if timestamp is a problem, I can remove it from the fingerprint in the interface.
I still wouldn't use the __hash__ implementation, but if yu think it's a must, I can also spend some effort to create custom fingerprint methods for certain events?

ancalita: Yup I just tested using `__hash__` in the fingerprint and it fails the consistency test.

joejuzl: We could potentially unify `__hash__` and `fingerprint`. I.e. make `__hash__` use your new stable implementation.

tests/shared/core/test_events.py
```diff
@@ -614,3 +628,143 @@ def test_event_executed_comparison(
...
+    ),
+    WarningPredictedAction(action_name_prediction="test"),
+]
+
+
+@pytest.mark.parametrize("event", tested_events)
+def test_event_fingerprint_consistency_across_runs(event: Event):
+    f1 = event.fingerprint()
+
+    event2 = copy.copy(event)
```
joejuzl, line 751: What exactly is this testing? I think that the problem with inbuild `hash` is only apparent when starting a new python process.

tests/shared/core/test_events.py
```diff
@@ -614,3 +628,143 @@ def test_event_executed_comparison(
...
+    subclasses = [event.__class__ for event in tested_events]
+    assert event_class in subclasses
+
+
+@pytest.mark.parametrize(
+    "event", tested_events,
+)
+def test_event_fingerprint_uniqueness(event: Event):
+    f1 = event.fingerprint()
+    event.timestamp = 1
```
joejuzl, line 768: Tobias makes a good point about `timestamp`. Does this mean that when we re-read the same training data file the event fingerprints won't match?

rasa/shared/core/trackers.py
```diff
@@ -859,6 +859,23 @@ def latest_action_name(self) -> Optional[Text]:
...
+            fingerprint of the tracker
+        """
+        data = {"sender_id": self.sender_id}
+
+        if self.slots:
+            data.update(self.slots)
+
+        if self.events:
+            for event in list(self.events):
+                data.update(event.as_dict())
```
joejuzl, line 875: Event also has `fingerprint` method so can just add it directly.

joejuzl (changes requested):

rasa/shared/core/events.py
ancalita: By that do you mean delete current `__hash__` body in each individual subclass and return instead `self.fingerprint()`?
Or maybe make `__hash__` use `deep_container_fingerprint()` instead of builtin `hash`?

tests/shared/core/test_events.py
ancalita: I thought it's testing if an identical event (i.e, with the same data) returns the same fingerprint 😅 
I just tested with `copy.deepcopy` instead and the test passes.

tests/shared/core/test_events.py
ancalita: I don't know enough of what reading the training data file entails to be able to answer, but if timestamp is different every time the training data file is re-read, I will remove the timestamp from the fingerprint.

rasa/shared/core/events.py
```diff
@@ -1744,6 +1755,12 @@ def as_dict(self) -> Dict[Text, Any]:
         d["event"] = ActiveLoop.type_name
         return d
 
+    def fingerprint(self) -> Text:
```
ancalita, line 1758: Had to implement separate fingerprint method because the `as_dict` methods for `LegacyForm` and `LegacyFormValidation` update their `type_name` to be identical to their parent class `type_name` and therefore could run into a problem where 2 different event types have the same fingerprint.

rasa/shared/core/events.py
joejuzl: > By that do you mean delete current __hash__ body in each individual subclass and return instead self.fingerprint()?

Yup exactly. They are both attempting to the do the same thing, i.e. a unique representation.

For reference:
The `__hash__` method is used when you pass the object into the `hash` function and also, in turn, for things like using the object as a key in a `dictionary`, or putting it in a `set`.

rasa/shared/core/events.py
joejuzl: Can you put a comment explaining this?

rasa/shared/core/events.py
```diff
@@ -1744,6 +1755,12 @@ def as_dict(self) -> Dict[Text, Any]:
...
+        """Returns the hash of the event."""
+        d = super().as_dict()
```
joejuzl, line 1760: Maybe clearer to re-set the name here rather than calling `super()`?
It took me a second to figure out why this fixed it.

joejuzl:
> I thought it's testing if an identical event (i.e, with the same data) returns the same fingerprint 😅
I just tested with copy.deepcopy instead and the test passed

(For some reason I can't reply to this directly)

Yup I just meant that this won't test the instability of the builtin `hash`. As that only changes when starting a new python _process_.

joejuzl (changes requested):

rasa/shared/core/events.py
wochinge: > though not sure I understand the 2nd question, when would the events be re-instantiated? I am probably not seeing the full implications of where events get used.

Sorry, I'm a bit short on time, so I hope phrasing this as question works 🙈 
If you read a story file and create a tracker from the events and then do the same thing a couple of seconds later:
Is the fingerprint of both trackers the same?

ancalita: @wochinge I'll write a test for that 😄

ancalita: @wochinge Yup, timestamp does affect the fingerprint of both the trackers. I'll keep timestamp out of the event fingerprint. And also push this new test too.

ancalita: @joejuzl regarding the __hash__ implementation: would it be ok to add __hash__ to parent class and then remove the __hash__ for each individual subclass? There would be too many duplicate return lines if I amend this in every subclass 😄

ancalita: ah in the end seems I have to go down this route of duplication otherwise I get a `TypeError: unhashable type` when trying to use hash() on specific event subclasses.

joejuzl: > @joejuzl regarding the __hash__ implementation: would it be ok to add __hash__ to parent class and then remove the __hash__ for each individual subclass? There would be too many duplicate return lines if I amend this in every subclass 😄

Yes definitely.

> I get a TypeError: unhashable type when trying to use hash() on specific event subclasses.

huh that's weird. But `fingerprint` returns a `str` right?  which should also always be hashable. Worth looking into a but further I think.

ancalita: I think I found the answer [here](https://docs.python.org/3/reference/datamodel.html#object.__hash__) which is identical to Events scenario:
```
If a class that overrides __eq__() needs to retain the implementation of __hash__() from a parent class, the interpreter must be told this explicitly by setting __hash__ = <ParentClass>.__hash__.
```
So either way I still have a lot of lines to duplicate - which one do you think makes most sense? The above one or having a __hash__ implementation per subclass?

rasa/shared/core/events.py
```diff
@@ -1534,7 +1537,13 @@ def __str__(self) -> Optional[Text]:
 
     def __hash__(self) -> int:
         """Returns unique hash for event."""
-        return hash(self.__members__())
+        return int(self.fingerprint(), 16)
+
+    def fingerprint(self) -> Text:
```
ancalita, line 1542: Had to override `Event.fingerprint()` because a [story conflict test ](https://github.com/RasaHQ/rasa/blob/3f3c8830a99278578ea687bdfc3645d7c368e86b/tests/core/training/test_story_conflict.py#L139) relied on a constant hash for ACTION_UNLIKELY_INTENT ([here ](https://github.com/RasaHQ/rasa/blob/3f3c8830a99278578ea687bdfc3645d7c368e86b/rasa/core/training/story_conflict.py#L248)and [here](https://github.com/RasaHQ/rasa/blob/3f3c8830a99278578ea687bdfc3645d7c368e86b/rasa/core/training/story_conflict.py#L254)).

rasa/shared/core/trackers.py
```diff
@@ -859,6 +859,22 @@ def latest_action_name(self) -> Optional[Text]:
...
+        Returns:
+            fingerprint of the tracker
+        """
+        data: Dict[Text, Any] = {"sender_id": self.sender_id}
+
+        if self.slots:
+            data.update(self.slots)
+
+        if self.events:
+            data.update({"events": list(self.events)})
```
joejuzl, line 874:
```suggestion
            data["events"] = list(self.events)
```

joejuzl (approved):
Nice! 🚀

Issue #9185, opened 2021-07-22
Add Debug Statements to Conditional Response Variation


b-quachtran:
**Description of Problem**:

Rasa logs don't provide any information on what conditional response variations are being selected based on the slot values.

**Overview of the Solution**:

It'd be nice to have some added debug messages that indicate which response is selected based on what criteria. 

**Definition of Done**:
- [x] Debug statements added

PR #9193, opened 2021-07-23
Added Debug Messages for Conditional Response Variation

b-quachtran:
**Proposed changes**:
- Added Debug Messages for Conditional Response Variation
- Closes Issue #9185 


rasa/core/nlg/response.py
```diff
@@ -75,12 +75,20 @@ def _responses_for_utter_action(
         )
 
         if conditional_channel:
+            logger.debug(
+                "Selecting response variation with condition: "
```
ancalita, line 79: Did you intentionally choose to keep `response variation` in singular? Because actually, multiple eligible variations could be selected here then randomly picked in a different method.

rasa/core/nlg/response.py
```diff
@@ -75,12 +75,20 @@ def _responses_for_utter_action(
...
+                f"'{conditional_channel[0].get('condition')}'"
```
ancalita, line 80: To double-check on your intention here - you want to log the condition data, i.e. matched slot name and value?
If that's so perhaps we should format this a bit nicer to the user, otherwise it will come out as a list object with a dict, so for example extracting slot name and value from condition and using these in the f-string.

I would also use the constant `RESPONSE_CONDITION` for consistency.

ancalita (commented):
Thanks for the enhancement 💯

rasa/core/nlg/response.py
b-quachtran: Ah no, not intentional. I'll get that updated 👍

rasa/core/nlg/response.py
b-quachtran: Yeah, it'd be for logging slot names and values. Make sense, will update

rasa/core/nlg/response.py
```diff
@@ -177,3 +191,18 @@ def _response_variables(
         response_vars = filled_slots.copy()
         response_vars.update(kwargs)
         return response_vars
+
+    @staticmethod
+    def _format_response_conditions(response_conditions: List[Dict]) -> Text:
```
ancalita, line 196: Could you please add types for the Dict items: `List[Dict[Text, Any]]`?

rasa/core/nlg/response.py
```diff
@@ -75,12 +75,26 @@ def _responses_for_utter_action(
         )
 
         if conditional_channel:
+            formatted_response_conditions = self._format_response_conditions(
+                conditional_channel[0].get(RESPONSE_CONDITION)
```
ancalita, line 79: Here I have an edge case that might cause an inconsistency with the selection of the first text variation in the conditional responses list. Say we have 2 text variations for the same utter action: variation 1 with condition that slot A is a certain value, variation 2 with condition that slot B is another value. If a user has both slots A and B filled with these values, both response variations become eligible conditional responses. In [this line](https://github.com/RasaHQ/rasa/blob/b2b5b73c10b7a7e5f3331a0f6887ae7d6f1782fd/rasa/core/nlg/response.py#L105), rasa will pick at random one of the text variations - this could be inconsistent with the debug log message, you would say selecting response with condition A, but actually the response with condition B was shown.

To avoid this inconsistency - I actually suggest moving the debug logging in `_random_response_for` method after this:
```
if suitable_responses:
        selected_response = np.random.choice(suitable_responses)
        condition = selected_response.get(RESPONSE_CONDITION)
        if condition:
            formatted_response_condition = ...
            logger.debug(f"Selecting response variation with conditions:"....)
        return selected_response
```

Hope this makes sense, let me know if it doesn't!

Maybe if you think it's a good idea to extract the lines between `selected response` assignment and return into its own logging method 🤔

ancalita (commented):
Thank you for the changes 👍 
I've added a bit of a long comment of something I just realised could be problematic.

I would also suggest adding a unit test for the static formatting method too to make sure the debug message is what you expect it to be. Have you tested this on the CLI also?

b-quachtran:
@ancalita Yep, makes sense. Will add a unit test as well to verify formatting 👍

tests/core/nlg/test_response.py
```diff
@@ -398,3 +400,44 @@ async def test_nlg_conditional_edgecases(slots, channel, expected_response):
...
+
+async def test_nlg_conditional_response_variations_condition_logging(
+    caplog: LogCaptureFixture,
+):
+    domain = Domain.from_yaml(
+        """
+        version: "2.0"
+        responses:
+           utter_action:
+             - text: "example with channel"
```
b-quachtran, line 413:
```suggestion
             - text: "example"
```

ancalita (approved):
Looks great 🎉

Issue #9102, opened 2021-07-12
UnexpecTEDIntentPolicy doesn't train on e2e user data


JEM-Mosig:
**Rasa version**:
`intent-ted` branch (https://github.com/RasaHQ/rasa/pull/8913), b54a4be182d9c7e6d524f5f4df8dd19bbee6a693

```
Rasa Version      :         2.7.1
Minimum Compatible Version: 2.6.0
Rasa SDK Version  :         2.7.0
Rasa X Version    :         None
Python Version    :         3.8.0
Operating System  :         macOS-10.16-x86_64-i386-64bit
Python Path       :         /Users/johannes/miniconda3/envs/rasa-master/bin/python
```

**Issue**:

When trained on
```
- story: sad path 1
  steps:
  - intent: greet
  - action: utter_greet
  - user: "blah"
  - bot: "xyz"
  - action: utter_did_that_help
  - intent: affirm
  - action: utter_happy
```
`UnexpecTEDIntentPolicy` fails to train.

**Error (including full traceback)**:
```
Traceback (most recent call last):
  File "/Users/johannes/miniconda3/envs/rasa-master/bin/rasa", line 5, in <module>
    main()
  File "/Users/johannes/rasa/rasa/rasa/__main__.py", line 117, in main
    cmdline_arguments.func(cmdline_arguments)
  File "/Users/johannes/rasa/rasa/rasa/cli/train.py", line 59, in <lambda>
    train_parser.set_defaults(func=lambda args: run_training(args, can_exit=True))
  File "/Users/johannes/rasa/rasa/rasa/cli/train.py", line 91, in run_training
    training_result = train_all(
  File "/Users/johannes/rasa/rasa/rasa/api.py", line 109, in train
    return rasa.utils.common.run_in_loop(
  File "/Users/johannes/rasa/rasa/rasa/utils/common.py", line 296, in run_in_loop
    result = loop.run_until_complete(f)
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/Users/johannes/rasa/rasa/rasa/model_training.py", line 108, in train_async
    return await _train_async_internal(
  File "/Users/johannes/rasa/rasa/rasa/model_training.py", line 288, in _train_async_internal
    await _do_training(
  File "/Users/johannes/rasa/rasa/rasa/model_training.py", line 352, in _do_training
    await _train_core_with_validated_data(
[...]

PR #9203, opened 2021-07-26
Make UnexpecTEDIntentPolicy compatible with E2E data

dakshvar22:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/9102
- The approach taken here: Filter incoming training trackers to consist of only those trackers which have no `UserUttered` event with intent undefined or an `ActionExecuted` event with action name undefined.


kedz:
When does this happen: `ActionExecuted` events with no action_name.
Is that because of an E2E prediction?

rasa/core/policies/unexpected_intent_policy.py
```diff
@@ -397,6 +398,37 @@ def compute_label_quantiles_post_training(
...
+        Args:
+            trackers: All trackers available for training.
+
+        Returns:
+            Trackers which should be used for training.
+        """
+        trackers_for_training = []
+        for tracker in trackers:
+            tracker_compatible = True
+            for event in tracker.events:
```
kedz, line 422: Why not `tracker.applied_events()` here? Is this because we are assuming training trackers don't don't have things like restarts or event rollbacks?

dakshvar22:
It's the debatable case that was discussed on the disentangling doc as well. When you create an E2E story like - 
```
story:
  - user: "Hello, i'm looking for a play station"
  - bot: "Which version"
```

See the usage of `bot` instead of `action`. Whenever a training story like this is added, `ActionExecuted` events don't have an `action_name` but they have `action_text`.

rasa/core/policies/unexpected_intent_policy.py
dakshvar22: Yeah very very unlikely that training trackers will have those. But even if they do, they will be applied inside the tracker featurizer. So, if anything we are looking at more number of events over here.

dakshvar22: No strong opinions here TBH. `applied_events` would also work. It didn't cross my mind honestly. 😄

dakshvar22: Let me change it to applied_events actually. Thanks for pointing it out.

kedz (approved):
This PR looks good!

kedz:
@dakshvar22 This is beyond the scope of this PR but is it worth considering whether to hide E2E turns from `UnexpecTEDIntentPolicy` rather than removing them from training data? I'm worried that this current setup will cause `UnexpecTEDIntentPolicy` to do unpredictable things at run time after an E2E turn because it's input state features will include featurizations of user or bot utterances and it will not have seen those at training time or if these turns are hidden from the policy at run time, the states will still be unusual given the training data censoring. WDYT?

dakshvar22:
Let me first disable auto-merge 😅

dakshvar22:
@kedz I think you have raised a good point. Hiding E2E turns might be a good idea however I think we should think more about this problem and potential solutions when either of these features are promoted to first class features. Otherwise, we'll add a lot of code for interaction between two experimental features without being sure about either which does not seem like the best idea. I'll create a separate issue for this so that we don't lose track of it.

kedz:
@dakshvar22 💯 Thanks for creating the issue!

PR #9211, opened 2021-07-27
Fix create issue token error in CI - model regression

HotThoughts:
**The problem:**
- Report description from [here](https://github.com/RasaHQ/rasa/blob/b2b5b73c10b7a7e5f3331a0f6887ae7d6f1782fd/.github/workflows/ci-model-regression-on-schedule.yml#L445) is a multiline string that contains certain symbols (left quote, single quote). For multiline string in JavaScript, we can quote it use left quotes, but we have to make sure the string itself does not contain any left quotes or add a \ in front of `. 

**Proposed changes:**
- Fix the `SyntaxError: Invalid or unexpected token`  error by replacing all left quotes with \\`. See the [workflow run](https://github.com/RasaHQ/rasa/pull/9211/checks?check_run_id=3185682449) and the [created issue](https://github.com/RasaHQ/rasa/issues/9225).


tczekajlo (approved):
LGTM 🎖️

PR #9222, opened 2021-07-28
getting rid of model access in `rasa.nlu.test`

### LINKED ISSUE #9136: Get rid of model access in `rasa.nlu.test`
**Description of Problem**:
The `rasa.nlu.test` module has several paths where it directly access attributes on the `processor`'s `pipeline` or even lower. This has the disadvantage that the test code is really tied to the specific model architecture. With the revamp we want to reduce these ties. This will help us to keep the interface of `Graphprocessor` small and explicit and will reduce the amount of breaking things when we start to implement the actual graph.

[Related implementation proposal page](https://www.notion.so/rasa/rasa-nlu-test-408a6a1ac76642d499c54faf5a271d03)

Locations where currently access things which are too low level:

1. [`remove_pretrained_extractors`](https://github.com/RasaHQ/rasa/blob/e1d2b6c4a37f7ed0dda93efbacd386f01556d119/rasa/nlu/test.py#L1398): We can instead remove the predictions of pretrained entity extractors from the predictions themselves (basically excluding them from the assert)
2. [`is_classifier_present`](https://github.com/RasaHQ/rasa/blob/e1d2b6c4a37f7ed0dda93efbacd386f01556d119/rasa/nlu/test.py#L1363): We can e.g. look at a single test prediction to see if an intent was added
3. [`is_response_selector_present`](https://github.com/RasaHQ/rasa/blob/e1d2b6c4a37f7ed0dda93efbacd386f01556d119/rasa/nlu/test.py#L1373): We can e.g. look at a single test prediction to see if the response selector added some result
4. [`get_available_response_selector_types`](https://github.com/RasaHQ/rasa/blob/e1d2b6c4a37f7ed0dda93efbacd386f01556d119/rasa/nlu/test.py#L1384): See related Slack thread [here](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1624542769423100). We can basically use the same approach as [here](https://github.com/RasaHQ/rasa/blob/38bb8e98167508229bd90e3fd2a6fa4758300b2e/rasa/core/test.py#L464)
5. [`is_entity_extractor_present`](https://github.com/RasaHQ/rasa/blob/e1d2b6c4a37f7ed0dda93efbacd386f01556d119/rasa/nlu/test.py#L1357): We can e.g. look at the test data and the test predictions to see what extractors are present (or whether one is present)
6. [`get_entity_extractors`]: (https://github.com/RasaHQ/rasa/blob/e1d2b6c4a37f7ed0dda93efbacd386f01556d119/rasa/nlu/test.py#L1332): We can e.g. look at the test data and the test predictions to see what extractors are present (or whether one is present)

**Overview of the Solution**:
I outlined some proposed ways of getting rid of these accesses. Anything else which doesn't require an additional method on the `GraphProcessor` is also fine.

**Examples** (if relevant):

**Blockers** (if relevant):

**Definition of Done**:
- [ ] The described locations no longer inspect or manipulate the interpreter directly
- [ ] A researcher has reviewed and approved the changes

twerkmeister:
**Proposed changes**:
- fixes #9136 


twerkmeister:
I am running out of time on this one ~ currently going with the proposed approach in the issue. I think it's definitely possible and also not too difficult to change on the surface. I have made the change for the entity extraction. However, what I am seeing now is that some of the more integrated tests relied on results being there simply by virtue of the components being present. These tests probably might be needed to be redesigned to actually do some work and not just look like they are doing it. Or we are happy with the complex procedure not crashing. Something to consider.

twerkmeister:
So this can be reviewed now - it seems to work reasonably well - however it's difficult to determine whether there will be any hiccups for users running the eval script. The evaluation builds fairly complex and nested data structures whose content determines what metrics are calculated. Now that we change part of this logic to work retroactively, inspecting the results, instead of our message processor, sometimes there are surprises, especially in small, demo-sized datasets:

1. We cannot distinguish anymore whether there is no entity extractor or whether the entity extractor didn't find anything
2. Train / test splits can lead to entities being underrepresented in either train/test and lead to missing results
3. For entity extractors, low epoch numbers can lead to non-recognition even on the training set

To get around some of these problems, I inspect the training data instead of the results, i.e. asking "are there any entities in the training data?" as opposed to "did any component find any entities?". Yet at the same time this might lead to problems when no entity extractor is found or present, as the current eval procedure depends on extractors being there.

rasa/nlu/test.py
```diff
@@ -1319,94 +1308,34 @@ def get_eval_data(
...
-
-    return response_selector_types
-
-
-def remove_pretrained_extractors(pipeline: List[Component]) -> List[Component]:
-    """Remove pre-trained extractors from the pipeline.
-
-    Remove pre-trained extractors so that entities from pre-trained extractors
-    are not predicted upon parsing.
+def get_active_entity_extractors(
```
wochinge, line 1318: This is the root of all problems, right? What do you think of changing the predictions of entity extractors to do return `{"name": "<name of entity extractor", "entities": [...]}`? We could keep the change only with NLU for now and keep the `UserUttered` part the same.

tests/nlu/test_evaluation.py
```diff
@@ -410,17 +408,11 @@ def test_run_cv_evaluation(
             "pipeline": [
                 {"name": "WhitespaceTokenizer"},
                 {"name": "CountVectorsFeaturizer"},
-                {"name": "DIETClassifier", EPOCHS: 2},
+                {"name": "DIETClassifier", EPOCHS: 25},
```
wochinge, line 411: can we instead set the random seed?

twerkmeister: potentially, but the difference here is really just 5 seconds or so. The training data is extremely small

rasa/nlu/test.py
twerkmeister: Yeah that makes sense! Then you would know reliably what extractors are active in the graph 💡

wochinge: @dakshvar22 What do you think of this suggestion?

rasa/nlu/test.py
```diff
@@ -1246,54 +1245,44 @@ def get_eval_data(
...
     intent_results, entity_results, response_selection_results = [], [], []
 
-    response_labels = [
+    response_labels = {
         e.get(INTENT_RESPONSE_KEY)
         for e in test_data.intent_examples
         if e.get(INTENT_RESPONSE_KEY) is not None
-    ]
+    }
     intent_labels = [e.get(INTENT) for e in test_data.intent_examples]
```
wochinge, line 1253: optional: Should this also be a `set` then?
```suggestion
    intent_labels = {e.get(INTENT) for e in test_data.intent_examples}
```

rasa/nlu/test.py
```diff
@@ -1246,54 +1245,44 @@ def get_eval_data(
...
-        is_entity_extractor_present(interpreter) and len(test_data.entities) > 0
-    )
+    should_eval_intents = len(set(intent_labels)) >= 2
+    should_eval_response_selection = len(response_labels) >= 2
+    should_eval_entities = len(test_data.entity_examples) > 0
 
     for example in tqdm(test_data.nlu_examples):
         result = interpreter.parse(example.get(TEXT), only_output_properties=False)
-
+        remove_entities_of_extractors(result, PRETRAINED_EXTRACTORS)
```
wochinge, line 1260: optional: can this be a protected function?
```suggestion
        _remove_entities_of_extractors(result, PRETRAINED_EXTRACTORS)
```

rasa/nlu/test.py
wochinge: can this be protected?
```suggestion
def _get_active_entity_extractors(
```

rasa/nlu/test.py
```diff
@@ -1319,94 +1308,34 @@ def get_eval_data(
...
-    Args:
-        pipeline: the pipeline
 
-    Returns:
-        Updated pipeline
-    """
-    pipeline = [c for c in pipeline if c.name not in PRETRAINED_EXTRACTORS]
-    return pipeline
+def remove_entities_of_extractors(
+    result_dict: Dict[Text, Any], extractor_names: Set[Text]
```
wochinge, line 1331: optional: Seems a little bit less vague that `result_dict` to me
```suggestion
    nlu_parse_result: Dict[Text, Any], extractor_names: Set[Text]
```

rasa/nlu/test.py
```diff
@@ -1636,14 +1564,12 @@ def cross_validate(
     import rasa.nlu.config
 
     if isinstance(nlu_config, (str, Dict)):
-        # nlu_config = rasa.nlu.config.load(nlu_config)
```
wochinge, line 1639: 💯

rasa/nlu/test.py
```diff
@@ -1959,7 +1864,7 @@ def _compute_entity_metrics(
     from rasa.model_testing import get_evaluation_metrics
 
     entity_metric_results: EntityMetrics = defaultdict(lambda: defaultdict(list))
-    extractors = get_entity_extractors(interpreter)
+    extractors = get_active_entity_extractors(entity_results)
```
wochinge, line 1867: can we please remove the unused `interpreter` param for the function?

rasa/nlu/test.py
wochinge: @twerkmeister I think it's fine to keep it like this for now but can we please create a follow up issue? Then we can either change the entity predictions or allow some basic introspection (e.g. the model metadata has the graph schema which would allow us iterating over classes from the outside)

tests/nlu/test_evaluation.py
```diff
@@ -448,9 +440,12 @@ def test_run_cv_evaluation(
         assert all(key in extractor_evaluation for key in ["errors", "report"])
 
 
-def test_run_cv_evaluation_with_response_selector(monkeypatch: MonkeyPatch):
+@pytest.mark.timeout(
```
wochinge, line 443: how about a test where there are no entities in the test set / none are predicted?

rasa/nlu/test.py
wochinge: thanks!

wochinge:
Ready for another review?

joejuzl (approved):
Looks good! Nice we can remove the mocking stuff from the unit tests.

Mazyod:
https://github.com/RasaHQ/rasa/pull/9222#discussion_r698413443

Is it clear that the proposed change here affects the behavior of the eval function and is out of scope of the PR (namely, decoupling model access from nlu.test)?

Since Training Data can have multiple sentences, but a single intent, using `set` made sure there are more 1 unique intent, and removing it just means there are more than one training sentence.

I am just concerned this seemingly innocuous change isn't thought through as a behavior change. If it was indeed and intended change, my apologies in advance.

Issue #9176, opened 2021-07-20
Rasa 3.0 Architecture Revamp: Core graph entities


joejuzl:
**Background**:
Context [here](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e).

**Description**:
This ticket encompasses the elements that will allow us to instantiate and run a graph.

- `GraphRunner`: Instantiates and runs a `GraphSchema`
- `GraphNode`: A single node in an `InstantiatedGraph`.
- `GraphComponent`: An interface for all GraphComponents which sit in within a `GraphNode`
- `GraphComponentOutput` An interface for anything a `GraphComponent` and `GraphNode` can return. (May be empty/not needed).
- `ExecutionContext`: A typeddict/namedtuple/similar object that is passed to each GraphComponent when executing
- `GraphSchema`: A yaml(?) schema to define the allowed keys/structure of a `GraphSchema`
- `GraphNodeHook` An interface for classes that can define graph node hooks.

**Definition of Done**:
- [ ] Entities are implemented.
- [ ] Entities are unittested.
- [ ] Entities are functionally tested using test components.

PR #9224, opened 2021-07-28
Implement core graph entities and runner

joejuzl:
**Proposed changes**:
Fixes: #9176.

Things to check:
- Are we testing all the different run cases (plus edge cases)?
- Does the directory structure make sense?
- Some TODOs littered about.


rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @abstractmethod
+    def supported_languages(self) -> List[Text]:
+        """Determines which languages this component can work with."""
+        ...
+
+    @abstractmethod
+    def required_packages(self) -> List[Text]:
+        """Any extra python dependencies required for this component to run."""
+        ...
```
wochinge, line 74: does it make sense that we offer a default implementation here?

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+@dataclass
+class ExecutionContext:
+    """Holds information about a single graph run."""
+
+    graph_schema: GraphSchema = field(repr=False)
+    model_id: Text
+    diagnostic_data: bool = False
```
wochinge, line 83: optional: A bit more in line with `is_finetuning`
```suggestion
   should_add_diagnostic_data: bool = False
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+            self._component_class, self._constructor_name
+        )(self._component_config, self._execution_context, **kwargs)
+
+    def parent_node_names(self) -> List[Text]:
+        """The names of the parent nodes of this node."""
+        return list(self._inputs.values())
+
+    def __call__(self, *inputs_from_previous_nodes: List[Any]) -> Dict[Text, Any]:
+        """This method is called when the node executes in the graph."""
+        received_inputs = dict(ChainMap(*inputs_from_previous_nodes))
```
wochinge, line 160: is there a nice way than this? 🙈

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        self._component: GraphComponent = getattr(  # type: ignore[no-redef]
+            self._component_class, self._constructor_name
+        )(self._component_config, self._execution_context, **kwargs)
+
+    def parent_node_names(self) -> List[Text]:
+        """The names of the parent nodes of this node."""
+        return list(self._inputs.values())
+
+    def __call__(self, *inputs_from_previous_nodes: List[Any]) -> Dict[Text, Any]:
```
wochinge, line 158:
```suggestion
    def __call__(self, *inputs_from_previous_nodes: List[Dict[Text, Any]]) -> Dict[Text, Any]:
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+            constructor_kwargs = rasa.shared.utils.common.minimal_kwargs(
+                kwargs, self._constructor_fn
+            )
+            self._load_component(constructor_kwargs)
+
+        run_kwargs = rasa.shared.utils.common.minimal_kwargs(kwargs, self._fn)
+        logger.debug(
+            f"Node {self._node_name} running "
+            f"{self._component_class.__name__}.{self._fn_name} "
+            f"with kwargs: {run_kwargs}"
```
wochinge, line 175:
```suggestion
            f"with kwargs: {run_kwargs}."
```

rasa/engine/runner/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    def run(
+        self,
+        inputs: Optional[Dict[Text, Any]] = None,
+        targets: Optional[List[Text]] = None,
+    ) -> Dict[Text, Any]:
+        """Runs the instantiated graph with the given inputs and targets.
+
+        Args:
+            inputs: Input nodes to be added to the graph.
+                These can be referenced by name
```
wochinge, line 36: optional: weird formatting?

rasa/engine/runner/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+                in the "needs" key of a node in the schema.
+            targets: Nodes who's output is needed and must always run.
```
wochinge, line 38:
```suggestion
            targets: Nodes whose output is needed and must always run.
```

rasa/engine/runner/dask.py
```diff
@@ -0,0 +1,100 @@
...
+
+        run_targets = targets if targets else self._targets
+
+        logger.debug(
+            f"Running graph with inputs: {inputs}, targets: {targets} "
+            f"and {self._execution_context}."
+        )
+
+        try:
+            # TODO: set dask scheduler via config?
```
wochinge, line 87: in my opinion fine to skip this for now. We can still do so if we need it

rasa/engine/runner/dask.py
```diff
@@ -0,0 +1,100 @@
...
+            return dict(ChainMap(*dask_result))
+        except RuntimeError as e:
+            raise GraphRunError("Error running runner.") from e
+
+    @staticmethod
+    def _add_inputs_to_graph(inputs: Optional[Dict[Text, Any]], graph: Any,) -> None:
+        for input_name, input_value in inputs.items():
+            if input_value in graph.keys():
+                raise GraphRunError(
+                    f'Input value "{input_value}" clashes with a node name.'
```
wochinge, line 98: nit: I'd switch single and double quotes to be consistent with (most of) the codebase

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+        return list(self._inputs.values())
+
+    def __call__(self, *inputs_from_previous_nodes: List[Any]) -> Dict[Text, Any]:
+        """This method is called when the node executes in the graph."""
+        received_inputs = dict(ChainMap(*inputs_from_previous_nodes))
+        kwargs = {}
+        for input, input_node in self._inputs.items():
+            kwargs[input] = received_inputs[input_node]
+
+        if not self._component:
```
wochinge, line 165: optional: same thing but maybe easier to understand?
```suggestion
        if not self._eager:
```

rasa/engine/runner/dask.py
```diff
@@ -0,0 +1,100 @@
...
+
+        try:
+            # TODO: set dask scheduler via config?
+            dask_result = dask.get(run_graph, run_targets)
+            return dict(ChainMap(*dask_result))
+        except RuntimeError as e:
+            raise GraphRunError("Error running runner.") from e
+
+    @staticmethod
+    def _add_inputs_to_graph(inputs: Optional[Dict[Text, Any]], graph: Any,) -> None:
```
wochinge, line 94: can we improve the typing of `graph`?

tests/engine/runner/test_dask.py
```diff
@@ -0,0 +1,334 @@
...
+    graph_schema: GraphSchema = {
+        "add": SchemaNode(
+            needs={"i1": "first_input", "i2": "second_input"},
+            uses=AddInputs,
+            fn="add",
+            constructor_name="create",
+            config={},
+            eager=eager,
+        ),
+        "subtract_2": SchemaNode(
```
wochinge, line 148:
```suggestion
        "substract_2": SchemaNode(
```

wochinge (commented):
just had a short look but it looks very good 🚀 💯 
I'll have another look first thing in the morning where I play around a bit with it 👍🏻

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
+from __future__ import annotations
+from abc import ABC, abstractmethod
+from collections import ChainMap
+from dataclasses import dataclass, field
+from typing import Any, Callable, Dict, List, Optional, Text, Type
+
+from rasa.engine.runner.dask import logger
```
wochinge, line 7: causes a circular import

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+import rasa.shared.utils.common
+
+
```
wochinge, line 10:
```suggestion
logger = logging.getLogger(__name__)
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
+from __future__ import annotations
```
wochinge, line 1: sweet, didn't know about this. Will use that in mine, too.

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+
+    needs: Dict[Text, Text]
+    uses: Type[GraphComponent]
+    constructor_name: Text
+    fn: Text
+    config: Dict[Text, Any]
+    eager: bool = False
+    is_target: bool = False
+    is_input: bool = False
+    resource_name: Optional[Text] = None  # TODO: type Resource
```
wochinge, line 25: will be provided by my `ModelStorage` stuff

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+GraphSchema = Dict[Text, SchemaNode]
+
+
+class GraphComponent(ABC):
+    """Interface for any component which will run in a graph."""
+
+    # TODO: This doesn't enforce that it exists in subclasses..
```
wochinge, line 34: maybe something like a `classproperty`? Or (as this requires writing a custom decorator) an abstract static method?

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @classmethod
+    @abstractmethod
+    def create(
+        cls, config: Dict, execution_context: ExecutionContext
+    ) -> GraphComponent:
+        """Creates a new graph component.
+
+        Args:
+            config: This config overrides the `default_config`
```
wochinge, line 45:
```suggestion
            config: This config overrides the `default_config`.
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+            execution_context: Information about the current graph run.
+
+        Returns: An instantiated GraphComponent
```
wochinge, line 48: a bit further up you wrote `graph component`. I'd decide for one for consistency
```suggestion
        Returns: An instantiated `GraphComponent`
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+class GraphComponent(ABC):
+    """Interface for any component which will run in a graph."""
+
+    # TODO: This doesn't enforce that it exists in subclasses..
+    default_config: Dict[Text, Any]
+
+    @classmethod
+    @abstractmethod
+    def create(
+        cls, config: Dict, execution_context: ExecutionContext
```
wochinge, line 40:
```suggestion
        cls, config: Dict[Text, Any], execution_context: ExecutionContext
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+            config: This config overrides the `default_config`
+            execution_context: Information about the current graph run.
+
+        Returns: An instantiated GraphComponent
+        """
+        ...
+
+    @classmethod
+    def load(cls, config: Dict, execution_context: ExecutionContext) -> GraphComponent:
+        """The load method is for creating a component using persisted data.
```
wochinge, line 54: nit: This doesn't follow our descriptive docstring convention
```suggestion
        """Creates a component using a persisted version of itself.
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Args:
+            config: This config overrides the `default_config`
```
wochinge, line 57:
```suggestion
            config: This config overrides the `default_config`.
```

rasa/engine/graph.py
joejuzl: haha oops. I moved things around quite a lot using pycharm's refactor functionality - forgot to create a new logger.

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+            execution_context: Information about the current graph run.
+
+        Returns: An instantiated, loaded GraphComponent
+        """
+        return cls.create(config, execution_context)
+
+    @abstractmethod
+    def supported_languages(self) -> List[Text]:
+        """Determines which languages this component can work with."""
```
wochinge, line 70: should we state that returning an empty list means that all languages are supported? Or is that `None`?

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+        self._component: Optional[GraphComponent] = None
+        if self._eager:
+            self._load_component()
+
+    def _load_component(
+        self, additional_kwargs: Optional[Dict[Text, Any]] = None
+    ) -> None:
+        kwargs = additional_kwargs if additional_kwargs else {}
+
+        logger.debug(
```
joejuzl, line 147: I was tempted to create a custom logger in the graph node which automatically adds some formatting and information - do you think it's worth it?

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+        """Any extra python dependencies required for this component to run."""
+        ...
+
+
+@dataclass
+class ExecutionContext:
+    """Holds information about a single graph run."""
+
+    graph_schema: GraphSchema = field(repr=False)
+    model_id: Text
```
wochinge, line 84: what's the difference between `field(repr=False)` and just omitting it?

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+    diagnostic_data: bool = False
+    is_finetuning: bool = False
+
+
+class GraphNode:
+    """Instantiates and runs a `GraphComponent` within a graph.
+
+    A `GraphNode` is a wrapper for a `GraphComponent` that allows it to be executed
+    In the context of a graph. It is responsible for instantiating the component at the
```
wochinge, line 93:
```suggestion
    A `GraphNode` is a wrapper for a `GraphComponent` that allows it to be executed
    in the context of a graph. It is responsible for instantiating the component at the
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+        execution_context: ExecutionContext,
+    ) -> None:
+        """Initializes `GraphNode`.
+
+        Args:
+            node_name: The name of the node in the schema.
+            component_class: The class to be instantiated and run.
+            constructor_name: The method used to instantiate the component.
+            component_config: Config to be passed to the component.
+            fn_name: The function to be run when the node executes.
```
wochinge, line 116:
```suggestion
            fn_name: The function on the instantiated `component` to be run when the node executes.
```

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+        self._component: GraphComponent = getattr(  # type: ignore[no-redef]
+            self._component_class, self._constructor_name
+        )(self._component_config, self._execution_context, **kwargs)
+
+    def parent_node_names(self) -> List[Text]:
+        """The names of the parent nodes of this node."""
+        return list(self._inputs.values())
+
+    def __call__(self, *inputs_from_previous_nodes: List[Any]) -> Dict[Text, Any]:
+        """This method is called when the node executes in the graph."""
```
wochinge, line 161: nit: please use descriptive docstrings

rasa/engine/graph.py
```diff
@@ -0,0 +1,196 @@
...
+        received_inputs = dict(ChainMap(*inputs_from_previous_nodes))
+        kwargs = {}
+        for input, input_node in self._inputs.items():
```
wochinge, line 164: as this shadows the built-in Python one
```suggestion
        for input_name, input_node in self._inputs.items():
```

rasa/engine/graph.py
wochinge: optional: might be worth adding a docstring here?

rasa/engine/runner/__init__.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Optional, Text
+
+from rasa.engine.graph import ExecutionContext, GraphSchema
+
+
+class GraphRunner(ABC):
+    """A GraphRunner is responsible for running a GraphSchema."""
```
wochinge, line 9:
```suggestion
    """A GraphRunner is responsible for running a `GraphSchema`."""
```

rasa/engine/runner/dask.py
```diff
@@ -0,0 +1,100 @@
...
+            for node_name, schema_node in graph_schema.items()
+        }
+
+    def run(
+        self,
+        inputs: Optional[Dict[Text, Any]] = None,
+        targets: Optional[List[Text]] = None,
+    ) -> Dict[Text, Any]:
+        """Runs the graph (see parent class for full docstring)."""
+        run_graph = {
```
wochinge, line 71: optional: Should we add a link / example how the dask graph format looks like?

tests/engine/runner/test_dask.py
```diff
@@ -0,0 +1,334 @@
...
+
+    runner = DaskGraphRunner(
+        graph_schema=graph_schema, execution_context=execution_context
+    )
+    results = runner.run(inputs={"first_input": 3, "second_input": 4})
+    assert results["subtract_2"] == 5
+
+
+@pytest.mark.parametrize("eager", [True, False])
+def test_target_override(eager: bool):
```
wochinge, line 138:
```suggestion
def test_eager_loading(eager: bool):
```

tests/engine/runner/test_dask.py
```diff
@@ -0,0 +1,334 @@
...
+    }
+    runner = DaskGraphRunner(
+        graph_schema=graph_schema,
+        execution_context=ExecutionContext(graph_schema=graph_schema, model_id="1"),
+    )
+    results = runner.run()
+    assert results["provide"] == 1
+
+
+def test_no_target():
```
wochinge, line 217: should we also test the case where we have an unused node in the graph (neither target nor parent of another node)?

rasa/engine/runner/dask.py
wochinge: and can we give some pointers how to potentially resolve this?

rasa/engine/graph.py
```diff
@@ -0,0 +1,198 @@
...
+        self, additional_kwargs: Optional[Dict[Text, Any]] = None
+    ) -> None:
+        kwargs = additional_kwargs if additional_kwargs else {}
+
+        logger.debug(
+            f"Node {self._node_name} loading "
+            f"{self._component_class.__name__}.{self._constructor_name} "
+            f"with config: {self._component_config}, and kwargs: {kwargs}."
+        )
+        self._component: GraphComponent = getattr(  # type: ignore[no-redef]
```
wochinge, line 152: * should we catch the case when the component for some reason doesn't have the method
* Do we need to handle the case where something goes wrong during the instantiation of a component?

wochinge (commented):
It's a lot of nit picks, sorry, but as this will be the heart of the codebase I payed extra attention 🤓 

> Are we testing all the different run cases (plus edge cases)?

I was wondering if it would make sense to test the `GraphNode` in isolation as well? Currently we are only testing it indirectly via the `graph_runner`.

> Does the directory structure make sense?

Good to me 👍🏻

rasa/engine/graph.py
wochinge: Great idea! I'd take a note of it in the implementation proposal and defer it to later. I'd keep the `engine` PR as small and simple as possible, merge to `main` and then add the bells and whistles bit by bit

rasa/engine/runner/__init__.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+from abc import ABC, abstractmethod
```
wochinge, line 2: I'd prefer to move this a named module instead of having it in `__init__` . I think `__init__` should be empty except maybe some imports

rasa/engine/graph.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @classmethod
+    def load(cls, config: Dict, execution_context: ExecutionContext) -> GraphComponent:
+        """The load method is for creating a component using persisted data.
+
+        Args:
+            config: This config overrides the `default_config`
+            execution_context: Information about the current graph run.
+
+        Args:
```
wochinge, line 60: this part of the docstring is duplicate

Should we mention what the default implementation does (point to `create`)?

rasa/engine/graph.py
```diff
@@ -0,0 +1,198 @@
...
+        kwargs = additional_kwargs if additional_kwargs else {}
+
+        logger.debug(
+            f"Node {self._node_name} loading "
+            f"{self._component_class.__name__}.{self._constructor_name} "
+            f"with config: {self._component_config}, and kwargs: {kwargs}."
+        )
+        self._component: GraphComponent = getattr(  # type: ignore[no-redef]
+            self._component_class, self._constructor_name
+        )(self._component_config, self._execution_context, **kwargs)
```
wochinge, line 154: do you think it's safer to pass the first args via keywords instead of as args?

tests/engine/runner/test_dask.py
```diff
@@ -0,0 +1,334 @@
...
+        return ["en"]
+
+    def required_packages(self) -> List[Text]:
+        return []
+
+    def add(self, i1: int, i2: int) -> int:
+        return i1 + i2
+
+
+class SubstractByX(GraphComponent):
```
wochinge, line 33:
```suggestion
class SubtractByX(GraphComponent):
```

rasa/engine/graph.py
wochinge: I think you don't have to specify `List` similar as you don't have to specify `Dict` for `kwargs`
```suggestion
    def __call__(self, *inputs_from_previous_nodes: Dict[Text, Any]) -> Dict[Text, Any]:
```

rasa/engine/graph.py
joejuzl: Empty list?

rasa/engine/graph.py
joejuzl: haha it's your original way from the prototype. 
I actually think this might be nicest way though - unless you can think of a better way?
Could also extract to a function to make it clearer?

rasa/engine/graph.py
joejuzl: aha yes true. weird one.

tests/engine/runner/test_dask.py
joejuzl: Not sure what "substract" is 😁

rasa/engine/graph.py
wochinge: yep

rasa/engine/graph.py
wochinge: Let's extract it to a function yeah :-D

tests/engine/runner/test_dask.py
wochinge: yeah, I learned about that today 😆

tests/engine/runner/test_dask.py
joejuzl: the point of this test was to test passing `targets` into `run`. I just made it test for eager and not-eager too.

rasa/engine/graph.py
joejuzl: > should we catch the case when the component for some reason doesn't have the method

I didn't do this because it will be checked by the validation. However we could have another check here for added safety.

> Do we need to handle the case where something goes wrong during the instantiation of a component?

Wouldn't this require a very broad try catch?

tests/engine/runner/test_dask.py
joejuzl: Even though I named the class that way....

rasa/engine/graph.py
wochinge: > Wouldn't this require a very broad try catch?
 
yeah 🤷🏻  You mean we should just let it raise and fail then?

rasa/engine/graph.py
joejuzl: I couldn't find a way to get the typing to work for an abstract class property. We could do it as a method, but then it just means we have to call it rather than access as attribute.

rasa/engine/graph.py
joejuzl: I think `None` for all. Should we have a default?

rasa/engine/graph.py
joejuzl: It just means it's omitted from the auto-generated `__repr__` method.

rasa/engine/graph.py
wochinge: `None` for all sounds good. Yeah, let's go with the current default (which is `None` for for all)

rasa/engine/graph.py
wochinge: haha, that makes actually a lot of sense. Thanks for the explanation!

rasa/engine/graph.py
wochinge: that would be fine to me 👍🏻

tests/engine/runner/test_dask.py
```diff
@@ -0,0 +1,271 @@
...
+    graph_schema: GraphSchema = {
+        "provide": SchemaNode(
+            needs={},
+            uses=ProvideX,
+            fn="provide",
+            constructor_name="create",
+            config={},
+            is_target=True,
+        ),
+        "provide_2": SchemaNode(  # This will not execute
```
wochinge, line 156: this will not execute or just not return its output?

tests/engine/test_graph_node.py
```diff
@@ -0,0 +1,195 @@
...
+    if eager:
+        assert create_mock.called
+    else:
+        assert not create_mock.called
+
+    assert not run_mock.called
+
+    node()
+
+    assert create_mock.called
```
wochinge, line 108: should we assert that it was only called once?

wochinge (approved):
Left a few more tiny comments but in my opinion ready to go to the integration branch 🚀 💯

rasa/engine/runner/interface.py
```diff
@@ -0,0 +1,41 @@
...
+
+        Returns: Instantiated GraphRunner
+        """
+        ...
+
+    @abstractmethod
+    def run(
+        self,
+        inputs: Optional[Dict[Text, Any]] = None,
+        targets: Optional[List[Text]] = None,
```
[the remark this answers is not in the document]
joejuzl: Most of the time, because each target node will specify it's a target. This is just for if we want to run different subsets of the graph. I think we added it because we thought we might just tokenize or something.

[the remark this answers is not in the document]
wochinge: Oh, I see. Makes sense 👍🏻

Issue #9178, opened 2021-07-20
Rasa 3.0 Architecture Revamp: Caching and fingerprinting


joejuzl:
**Background**:
Context [here](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e).

**Description**:
This covers the elements that allow graphs to fingerprint and cache their outputs.
This does not cover the pruning of the graph after a fingerprint run.

- `Cache`: Interface for the cache.
- `LocalCache`: Local disk implementation of `Cache`
- `FingerprintCalculator`: Calculates node run fingerprints.
- `GraphFingerprintComponent`: Replaces a `GraphComponent` during a fingerprint run.
- `CachedComponent`: Replaces a `GraphComponent` so as to use a cached output.
- `Cacheable`: Interface for `GraphComponent` output which is cacheable.
- `GraphComponentTrainingOuput`:  Interface for `GraphComponent` output which is fingerprintable.
- `CachingHook`: Implementation of `GraphNodeHook` that enables caching during a training run.

**Definition of Done**:
- [ ] Entities are implemented.
- [ ] Entities are unittested.
- [ ] Entities are functionally tested using test components.

wochinge:
implemented `Cache` / `LocalCache` (didn't use an interface for now as we don't have other subclasses atm) and `Cacheable` on `3.0-architecture-revamp/9178/caching`. Will probably focus on integrating the `ModelStorage` with your branch tomorrow before continuing the cache stuff (somewhat builds on `ModelStorage`)

PR #9234, opened 2021-07-29
3.0 architecture revamp/9178/caching

wochinge:
**Proposed changes**:
- implement `Cache`, `LocalCache`, `Cacheable` of https://github.com/RasaHQ/rasa/issues/9178

### DIFF
--- a/rasa/engine/caching.py
+++ b/rasa/engine/caching.py
@@ -1,0 +1,450 @@
+from __future__ import annotations
+
+import abc
+import logging
+import os
+import shutil
+import tempfile
+from datetime import datetime
+from pathlib import Path
+from typing import Text, Any, Optional, Tuple, List
+
+from packaging import version
+from sqlalchemy.engine import URL
+
+from sqlalchemy.exc import OperationalError
+from typing_extensions import Protocol, runtime_checkable
+
+import rasa
+import rasa.model
+import rasa.utils.common
+import rasa.shared.utils.common
+from rasa.constants import MINIMUM_COMPATIBLE_VERSION
+import sqlalchemy as sa
+import sqlalchemy.orm
+from sqlalchemy.ext.declarative import declarative_base, DeclarativeMeta
+
+from rasa.engine.storage.storage import ModelStorage
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_CACHE_LOCATION = Path(".rasa", "cache")
+DEFAULT_CACHE_NAME = "cache.db"
+DEFAULT_CACHE_SIZE_MB = 1000
+
+CACHE_LOCATION_ENV = "RASA_CACHE_DIRECTORY"
+CACHE_DB_NAME_ENV = "RASA_CACHE_NAME"
+CACHE_SIZE_ENV = "RASA_MAX_CACHE_SIZE"
+
+
+class TrainingCache(abc.ABC):
+    """Stores training results in a persistent cache.
+
+    Used to minimize re-retraining when the data / config didn't change in between
+    training runs.
+    """
+
+    @abc.abstractmethod
+    def cache_output(
+        self,
+        fingerprint_key: Text,
+        output: Any,
+        output_fingerprint: Text,
+        model_storage: ModelStorage,
+    ) -> None:
+        """Adds the output to the cache.
+
+        If the output is of type `Cacheable` the output is persisted to disk in addition
+        to its fingerprint.
+
+        Args:
+            fingerprint_key: The fingerprint key serves as key for the cache. Graph
+                components can use their fingerprint key to lookup fingerprints of
+                previous training runs.
+            output: The output. The output is only cached to disk if it's of type
+                `Cacheable`.
+            output_fingerprint: The fingerprint of their output. This can be used
+                to lookup potentially persisted outputs on disk.
+            model_storage: Required for caching `Resource` instances. E.g. `Resource`s
+                use that to copy data from the model storage to the cache.
+        """
+
+    ...
+
+    @abc.abstractmethod
+    def get_cached_output_fingerprint(self, fingerprint_key: Text) -> Optional[Text]:
+        """Retrieves fingerprint of output based on fingerprint key.
+
+        Args:
+            fingerprint_key: The fingerprint serves as key for the lookup of output
+                fingerprints.
+
+        Returns:
+            The fingerprint of a matching output or `None` in case no cache entry was
+            found for the given fingerprint key.
+        """
+        ...
+
+    @abc.abstractmethod
+    def get_cached_result(
+        self, output_fingerprint_key: Text, node_name: Text, model_storage: ModelStorage
+    ) -> Optional[Cacheable]:
+        """Returns a potentially cached output result.
+
+        Args:
+            output_fingerprint_key: The fingerprint key of the output serves as lookup
+                key for a potentially cached version of this output.
+            node_name: The name of the graph node which wants to use this cached result.
+            model_storage: The current model storage (e.g. used when restoring
+                `Resource` objects so that they can fill the model storage with data).
+
+        Returns:
+            `None` if no matching result was found or restored `Cacheable`.
+        """
+        ...
+
+
+@runtime_checkable
+class Cacheable(Protocol):
+    """Protocol for cacheable graph component outputs.
+
+    We only cache graph component outputs which are `Cacheable`. We only store the
+    output fingerprint for everything else.
+    """
+
+    def to_cache(self, directory: Path, model_storage: ModelStorage) -> None:
+        """Persists `Cacheable` to disk.
+
+        Args:
+            directory: The directory where the `Cacheable` can persist itself to.
+            model_storage: The current model storage (e.g. used when caching `Resource`
+                objects.
+        """
+        ...
+
+    @classmethod
+    def from_cache(
+        cls, node_name: Text, directory: Path, model_storage: ModelStorage
+    ) -> Cacheable:
+        """Loads `Cacheable` from cache.
+
+        Args:
+            node_name: The name of the graph node which wants to use this cached result.
+            directory: Directory containing the persisted `Cacheable`.
+            model_storage: The current model storage (e.g. used when restoring
+                `Resource` objects so that they can fill the model storage with data).
+
+        Returns:
+            Instantiated `Cacheable`.
+        """
+        ...
+
+
+class LocalTrainingCache:
+    """Caches training results on local disk (see parent class for full docstring)."""
+
+    Base: DeclarativeMeta = declarative_base()
+
+    class CacheEntry(Base):
+        """Stores metadata about a single cache entry."""
+
+        __tablename__ = "cache_entry"
+
+        fingerprint_key = sa.Column(sa.String(), primary_key=True)
+        output_fingerprint_key = sa.Column(sa.String(), nullable=False, index=True)
+        last_used = sa.Column(sa.DateTime(timezone=True), nullable=False)
+        rasa_version = sa.Column(sa.String(255), nullable=False)
+        result_location = sa.Column(sa.String())
+        result_type = sa.Column(sa.String())
+
+    def __init__(self) -> None:
+        """Creates cache.
+
+        The `Cache` setting can be configured via environment variables.
+        """
+        self._cache_location = Path(
+            os.environ.get(CACHE_LOCATION_ENV, DEFAULT_CACHE_LOCATION)
+        )
+
+        self._max_cache_size = float(
+            os.environ.get(CACHE_SIZE_ENV, DEFAULT_CACHE_SIZE_MB)
+        )
+        self._cache_database_name = os.environ.get(
+            CACHE_DB_NAME_ENV, DEFAULT_CACHE_NAME
+        )
+
+        if not self._cache_location.exists() and not self._is_disabled():
+            logger.debug(
+                f"Creating caching directory '{self._cache_location}' because "
+                f"it doesn't exist yet."
+            )
+            self._cache_location.mkdir(parents=True)
+
+        self._sessionmaker = self._create_database()
+
+        self._drop_cache_entries_from_incompatible_versions()
+
+    def _create_database(self) -> sqlalchemy.orm.sessionmaker:
+        if self._is_disabled():
+            # Use in-memory database as mock to avoid having to check `_is_disabled`
+            # everywhere
+            database = ""
+        else:
+            database = str(self._cache_location / self._cache_database_name)
+
+        # Use `future=True` as we are using the 2.x query style
+        engine = sa.create_engine(
+            URL.create(drivername="sqlite", database=database), future=True
+        )
+        self.Base.metadata.create_all(engine)
+
+        return sa.orm.sessionmaker(engine)
+
+    def _drop_cache_entries_from_incompatible_versions(self) -> None:
+        incompatible_entries = self._find_incompatible_cache_entries()
+
+        for entry in incompatible_entries:
+            self._delete_cached_result(entry)
+
+        self._delete_incompatible_entries_from_cache(incompatible_entries)
+
+        logger.debug(
+            f"Deleted {len(incompatible_entries)} from disk as their version "
+            f"is older than the minimum compatible version "
+            f"('{MINIMUM_COMPATIBLE_VERSION}')."
+        )
+
+    def _find_incompatible_cache_entries(self) -> List[LocalTrainingCache.CacheEntry]:
+        with self._sessionmaker() as session:
+            query_for_cache_entries = sa.select(self.CacheEntry)
+            all_entries: List[LocalTrainingCache.CacheEntry] = session.execute(
+                query_for_cache_entries
+            ).scalars().all()
+
+        return [
+            entry
+            for entry in all_entries
+            if version.parse(MINIMUM_COMPATIBLE_VERSION)
+            > version.parse(entry.rasa_version)
+        ]
+
+    def _delete_incompatible_entries_from_cache(
+        self, incompatible_entries: List[LocalTrainingCache.CacheEntry]
+    ) -> None:
+        incompatible_fingerprints = [
+            entry.fingerprint_key for entry in incompatible_entries
+        ]
+        with self._sessionmaker.begin() as session:
+            delete_query = sa.delete(self.CacheEntry).where(
+                self.CacheEntry.fingerprint_key.in_(incompatible_fingerprints)
+            )
+            session.execute(delete_query)
+
+    @staticmethod
+    def _delete_cached_result(entry: LocalTrainingCache.CacheEntry) -> None:
+        if entry.result_location and Path(entry.result_location).is_dir():
+            shutil.rmtree(entry.result_location)
+
+    def cache_output(
+        self,
+        fingerprint_key: Text,
+        output: Any,
+        output_fingerprint: Text,
+        model_storage: ModelStorage,
+    ) -> None:
+        """Adds the output to the cache (see parent class for full docstring)."""
+        if self._is_disabled():
+            return
+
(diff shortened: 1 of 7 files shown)


rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+
+        Args:
+            fingerprint_key: The fingerprint key serves as key for the cache. Graph
+                components can use their fingerprint key to lookup fingerprints of
+                previous training runs.
+            output: The output. The output is only cached to disk if it's of type
+                `Cacheable`.
+            output_fingerprint: The fingerprint of they output. This can be used
+                to lookup potentially persisted outputs on disk.
+            model_storage: Required for caching `Cacheable` instances. E.g. `Resource`s
```
joejuzl, line 196:
```suggestion
            model_storage: Required for caching `Resource` instances. E.g. `Resource`s
```

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+                use that to copy data from the model storage to the cache.
+        """
+        if self._is_disabled():
+            return
+
+        cache_dir, output_type = None, None
+        if isinstance(output, Cacheable):
+            cache_dir, output_type = self._cache_output_to_disk(output, model_storage)
```
joejuzl, line 204: If the next step fails should we clean this up? Otherwise it will never be removed right?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+    class CacheEntry(Base):
+        """Stores metadata about a single cache entry."""
+
+        __tablename__ = "cache_entry"
+
+        # `create_sequence` is needed to create a sequence for databases that
+        # don't autoincrement Integer primary keys (e.g. Oracle)
+        fingerprint_key = sa.Column(sa.String(), primary_key=True)
+        output_fingerprint_key = sa.Column(sa.String(), nullable=False, index=True)
+        last_used = sa.Column(sa.DateTime(), nullable=False)
```
joejuzl, line 90: `timezone=True` so we know it's `utc`?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+            return path_to_cached, module
+        except Exception as e:
+            logger.warning(
+                f"Failed to find module for cached output of type "
+                f"'{result_type}'. Error:\n{e}"
+            )
+            return None, None
+
+
+def _directory_size_in_mb(path: Path) -> float:
```
joejuzl, line 366: Is there no existing util to do this?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+            oldest_cache_item = (
+                session.execute(query_for_least_recently_used_entry).scalars().first()
+            )
+
+            # TODO: Can't happen
+            if not oldest_cache_item:
+                return
+
+            self._delete_cached_result(oldest_cache_item)
+            delete_query = sa.delete(self.CacheEntry).where(
```
joejuzl, line 269: Is this duplicate functionality of `_delete_incompatible_entries_from_cache`?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+                logger.debug(f"Caching output of type '{type(output)}' succeeded.")
+            except Exception as e:
+                logger.error(
+                    f"Caching output of type '{type(output)}' failed with the "
+                    f"following error:\n{e}"
+                )
+                return None, None
+
+            output_size = _directory_size_in_mb(temp_dir)
+            if output_size > self._max_cache_size:
```
joejuzl, line 236: I wonder if we also want a `max_cache_entry_size`?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+
+            return None
+
+    def get_cached_result(
+        self, output_fingerprint_key: Text
+    ) -> Tuple[Optional[Path], Optional[Type[Cacheable]]]:
+        """Returns a potentially cached output result.
+
+        Args:
+            output_fingerprint_key:
```
joejuzl, line 309: empty description

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+
+        Returns:
+            - Path to directory containing the cached content or `None` in case no
+            persisted data was found
```
joejuzl, line 313:
```suggestion
            persisted data was found.
```

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def _get_cached_result(
+        self, output_fingerprint_key: Text
+    ) -> Tuple[Optional[Path], Optional[Text]]:
+        with self._sessionmaker.begin() as session:
+            query = sa.select(
+                self.CacheEntry.result_location, self.CacheEntry.result_type
+            ).where(
+                self.CacheEntry.output_fingerprint_key == output_fingerprint_key,
+                self.CacheEntry.result_location != sa.null(),
```
joejuzl, line 339: Do we need an index on `result_location` then?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+            )
+
+            match = session.execute(query).first()
+
+            if match:
+                return Path(match.result_location), match.result_type
```
joejuzl, line 345: I wonder if it's worth having a container for this given we pass this tuple around a lot?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+
+    @staticmethod
+    def _load_from_cache(
+        path_to_cached: Path, result_type: Text
+    ) -> Tuple[Optional[Path], Optional[Type[Cacheable]]]:
+        try:
+            module = rasa.shared.utils.common.class_from_module_path(result_type)
+            assert isinstance(module, Cacheable)
+
+            return path_to_cached, module
```
joejuzl, line 357: Why don't we call `module.from_cache(path_to_cached)` or similar here?

joejuzl (changes requested):
Nice one! 🚀 
Will review the tests too soon.

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+
+    Base: DeclarativeMeta = declarative_base()
+
+    class CacheEntry(Base):
+        """Stores metadata about a single cache entry."""
+
+        __tablename__ = "cache_entry"
+
+        # `create_sequence` is needed to create a sequence for databases that
+        # don't autoincrement Integer primary keys (e.g. Oracle)
```
wochinge, line 87:
```suggestion
```

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+        incompatible_entries = self._find_incompatible_cache_entries()
+
+        for entry in incompatible_entries:
+            self._delete_cached_result(entry)
+
+        self._delete_incompatible_entries_from_cache(incompatible_entries)
+
+        logger.debug(
+            f"Deleted {len(incompatible_entries)} from disk as their version "
+            f"is smaller than the minimum compatible version "
```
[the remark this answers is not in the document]
wochinge: I went for `older`, okay?

rasa/engine/caching.py
wochinge: unfortunately not 😬

rasa/engine/caching.py
wochinge: In the sequence diagram this is done by the caller. This has the advantage that we don't have to load everything into memory right away but only when we need it.

rasa/engine/caching.py
wochinge: what added benefit would it provide?

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+                fingerprint_key=fingerprint_key,
+                output_fingerprint_key=output_fingerprint,
+                last_used=datetime.utcnow(),
+                rasa_version=rasa.__version__,
+                result_location=cache_dir,
+                result_type=output_type,
+            )
+            session.add(cache_entry)
+
+    def _is_disabled(self) -> bool:
```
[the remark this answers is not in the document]
wochinge: It names the check which makes it easier to understand it's purpose.

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+        with self._sessionmaker() as session:
+            query_for_cache_entries = sa.select(self.CacheEntry)
+            all_entries: List[TrainingCache.CacheEntry] = session.execute(
+                query_for_cache_entries
+            ).scalars().all()
+
+        return [
+            entry
+            for entry in all_entries
+            if version.parse(MINIMUM_COMPATIBLE_VERSION)
```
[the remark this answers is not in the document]
wochinge: it's doable but makes the query very complicated and might not be super robust (e.g. think of edge cases like alphas, RCs, etc. ). I'd go for this now and maybe opt for a more sophisticated solution down the road (if we need it)

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+            f"is smaller than the minimum compatible version "
+            f"('{MINIMUM_COMPATIBLE_VERSION}')."
+        )
+
+    def _find_incompatible_cache_entries(self) -> List[TrainingCache.CacheEntry]:
+        with self._sessionmaker() as session:
+            query_for_cache_entries = sa.select(self.CacheEntry)
+            all_entries: List[TrainingCache.CacheEntry] = session.execute(
+                query_for_cache_entries
+            ).scalars().all()
```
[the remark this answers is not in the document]
wochinge: Related to upcoming changes in sqlachemy 2:

See [here](https://docs.sqlalchemy.org/en/14/orm/queryguide.html#selecting-orm-entities-and-attributes)

> When selecting a list of single-element rows containing ORM entities, it is typical to skip the generation of Row objects and instead receive ORM entities directly, which is achieved using the Result.scalars() method:

rasa/engine/caching.py
wochinge: Due to [this](https://www.z-car.com/blog/mysql/mysql-should-you-put-an-index-on-a-boolean-field-to-help-query-performance) and the fact that the (indexed) fingerprint key already limits the result to 0 or 1 entries I wouldn't add one

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+        ]
+
+    def _delete_incompatible_entries_from_cache(
+        self, incompatible_entries: List[TrainingCache.CacheEntry]
+    ) -> None:
+        incompatible_fingerprints = [
+            entry.fingerprint_key for entry in incompatible_entries
+        ]
+        with self._sessionmaker.begin() as session:
+            delete_query = sa.delete(self.CacheEntry).where(
```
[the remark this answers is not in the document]
wochinge: We can but that leads to `n` queries instead of a single one which can be optimized by the database.

rasa/engine/caching.py
wochinge: The other one deletes using a list of fingerprint keys where this one uses a single one. We could indeed deduplicate it but I'm wondering whether this doesn't decrease the readability

rasa/engine/caching.py
wochinge: We only pass it to the calling method and then to whoever is calling that. How about a named tuple?

rasa/engine/caching.py
wochinge: I'll turn around the order that should boil down to the same effect 👍🏻

rasa/engine/caching.py
```diff
@@ -0,0 +1,373 @@
...
+        """Stores metadata about a single cache entry."""
+
+        __tablename__ = "cache_entry"
+
+        # `create_sequence` is needed to create a sequence for databases that
+        # don't autoincrement Integer primary keys (e.g. Oracle)
+        fingerprint_key = sa.Column(sa.String(), primary_key=True)
+        output_fingerprint_key = sa.Column(sa.String(), nullable=False, index=True)
+        last_used = sa.Column(sa.DateTime(), nullable=False)
+        rasa_version = sa.Column(sa.String(255), nullable=False)
```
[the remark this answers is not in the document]
joejuzl: 👍  (I thought we were going to do version filtering in sql)

rasa/engine/caching.py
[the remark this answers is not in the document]
joejuzl: sorry that was unclear - I was asking about the feature of disabling the cache in-general.
Is it something we need, and is the best way to do it by using `max_cache_size`?

rasa/engine/caching.py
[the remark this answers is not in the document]
joejuzl: yep fair enough. I guess if you wanted to do it properly SQLy you would have a column for each part of the version (and maybe version in separate table).

rasa/engine/caching.py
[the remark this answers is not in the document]
joejuzl: ah I see. Was hard to find any information on the `scalars` method.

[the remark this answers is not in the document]
joejuzl: [here](https://docs.sqlalchemy.org/en/14/glossary.html#term-2.0-style) it mentions using `future=True`. Do we not have to do that too?

rasa/engine/caching.py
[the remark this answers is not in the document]
joejuzl: I thought you could do bulk deletion given a list of ORM objects.

rasa/engine/caching.py
joejuzl: I guess the question is, do we want to cache something which fills the entire cache?
Another alternative would be to allow a certain proportion of the max cache size. But maybe not needed.

rasa/engine/caching.py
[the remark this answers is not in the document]
wochinge: > ah I see. Was hard to find any information on the scalars method.

Yeah, I was struggling with that, too.

> here it mentions using future=True. Do we not have to do that too?

Added it 👍🏻

rasa/engine/caching.py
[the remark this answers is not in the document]
wochinge: nope. You'd need a for loop

rasa/engine/caching.py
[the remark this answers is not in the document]
wochinge: That's how we've stated it in the [implementation proposal](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#97534007cfc24aa38e3b4d54e7d35c84). I think it's useful when e.g. running with docker images which for some reason have no permissions to write to disk

rasa/engine/caching.py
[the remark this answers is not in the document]
wochinge: yeah, probably. You'd need to make sure that you'd handle alphas, Rcs. etc correctly though as well 😬

rasa/engine/caching.py
wochinge: I'd go for the simple LRU mechanism without any secret sauce to it for now. I am explicitly excluding the database from the size calculation now and added some handling for potential infinite loops

rasa/engine/caching.py
joejuzl: I.e. when we non-eagerly initialise the CachedGraphComponent?

rasa/engine/caching.py
joejuzl: might be worth making it a public util method and unit testing it.

rasa/engine/caching.py
wochinge: yep

rasa/engine/caching.py
joejuzl: 👍  fine as is.

rasa/engine/caching.py
wochinge: handled 👍🏻

rasa/engine/caching.py
joejuzl: named tuple sounds good 👍

rasa/engine/caching.py
wochinge: will do 👍🏻

rasa/engine/caching.py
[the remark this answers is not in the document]
joejuzl: I must be thinking about one of the different/older sqlalchemy query modes. It's annoying how many different ways there are to use it....

rasa/engine/caching.py
wochinge: done

wochinge:
@joejuzl Should we introduce a `Cache` interface or is it fine to omit it for now? I don't have a strong opinion on it

joejuzl:
> @joejuzl Should we introduce a Cache interface or is it fine to omit it for now? I don't have a strong opinion on it

I also don't have a strong opinion 😬

rasa/engine/caching.py
wochinge: used a dataclass and as this allowed me to simplify some checks

wochinge:
then let's go without one for now 👍🏻

joejuzl:
> then let's go without one for now 👍🏻

Sorry - but just had a thought - maybe having an interface will make it easier for testing, as we can have a test cache we use when testing other components?

tests/engine/test_caching.py
```diff
@@ -0,0 +1,358 @@
...
+        cached.cached_type.from_cache("xy", cached.cache_directory, model_storage)
+        == output
+    )
+
+
+def test_get_cached_result_with_miss(tmp_path: Path, temp_cache: TrainingCache):
+    model_storage = ModelStorage(tmp_path)
+
+    # Cache something
+    temp_cache.cache_output(uuid.uuid4().hex, None, uuid.uuid4().hex, model_storage)
```
joejuzl, line 81: Maybe have a value other than `None` so we know it's not being returned later?

tests/engine/test_caching.py
```diff
@@ -0,0 +1,358 @@
...
+
+    cache.cache_output(fingerprint_key1, output1, output_fingerprint1, model_storage)
+
+    # Cache only metadata (`output` is not `Cacheable`)
+    fingerprint_key2 = uuid.uuid4().hex
+    output_fingerprint2 = uuid.uuid4().hex
+    cache.cache_output(fingerprint_key2, None, output_fingerprint2, model_storage)
+
+    # Pretend we updated Rasa Open Source to a no longer compatible version
+    monkeypatch.setattr(rasa.engine.caching, "MINIMUM_COMPATIBLE_VERSION", "99999.9.9")
```
joejuzl, line 188: Would be cool to have one which _isn't_ deleted too, so we know we're not just nuking the whole thing.

joejuzl (changes requested):

rasa/utils/common.py
```diff
@@ -322,6 +322,31 @@ async def call_potential_coroutine(
...
+
+    Returns:
+        Directory size in MiB.
+    """
+    filenames_to_exclude = filenames_to_exclude or []
+    size = 0.0
+    for root, _dirs, files in os.walk(path):
+        for filename in files:
+            if filename in filenames_to_exclude:
+                # Exclude database from sum
```
joejuzl, line 342: not always a database right?

rasa/engine/caching.py
```diff
@@ -0,0 +1,443 @@
...
+    @staticmethod
+    def _load_from_cache(
+        path_to_cached: Path,
+        result_type: Text,
+        node_name: Text,
+        model_storage: ModelStorage,
+    ) -> Optional[Cacheable]:
+        try:
+            module = rasa.shared.utils.common.class_from_module_path(result_type)
+            assert isinstance(module, Cacheable)
```
joejuzl, line 435: Probably shouldn't use `assert` in prod code.

joejuzl (approved):
Couple of small comments

Issue #8924, opened 2021-06-21
Add tokens to events


joejuzl:
**Description of Problem**:
Background: https://github.com/RasaHQ/rasa/issues/8533
Proposed solution: https://www.notion.so/rasa/Sending-tokens-with-events-to-Rasa-X-05fe8895e4aa416e8ca8e3e1f339bc91

Summary:
We don't want to re-calculate tokens in Rasa X so we want to send the message tokens along with the event through the event broker.

**Overview of the Solution**:
- Add the text tokens to the events by updating `output_properties`
- Convert them to a serialisable form

**Related Issues**:
Adding model identifier to events: https://github.com/RasaHQ/rasa/issues/8914

**Definition of Done**:
- [x] UserUttered with text tokens are serializable.
- [x] All UserUttered events have text tokens attached when pushed to the event broker.
- [ ] Integration tested.

ancalita:
@joejuzl might need a quick crash course in integration testing 😬 to make sure I understand what's expected to be completed. I also left a few questions in the Notion page + double-checking on deserialisation of tokens, if that's something i need to take care of anywhere in Rasa? Or that's something that happens at Rasa-X end?

joejuzl:
> double-checking on deserialisation of tokens, if that's something i need to take care of anywhere in Rasa? Or that's something that happens at Rasa-X end?

It will be up to Rasa X to read the tokens from the events that it consumes from the event broker.

> might need a quick crash course in integration testing 

We have so few integration tests right now. Currently only for one broker, and it just checks initial connection...
Thinking about it now I'm not 100% sure if adding integration tests should be in scope of this issue.
For now I think just unit and manual test (with local Rasa X instance!)

Also have a look at `tests/integration_tests/core/brokers/test_pika.py` , and if we do decide to do more tests we can have a call to discuss it properly.

ancalita:
@joejuzl Quick q on running a manual test with `rasa x`: When I initialised the development set-up, I created separate virtual envs for rasa-oss and rasa x. Will the `rasa x` command run in the rasa-oss env or should I `pip install rasa-x` in this environment too?

joejuzl:
@joejuzl Quick q on running a manual test with rasa x: When I initialised the development set-up, I created separate virtual envs for rasa-oss and rasa x. Will the rasa x command run in the rasa-oss env or should I pip install rasa-x in this environment too?

[...]

PR #9239, opened 2021-07-30
Add text tokens while parsing the message

ancalita:
**Proposed changes**:
- Implements #8924 


rasa/nlu/model.py
```diff
@@ -464,7 +464,7 @@ def parse(
         data = self.default_output_attributes()
         data[TEXT] = text
 
-        message = Message(data=data, time=timestamp)
+        message = Message(data=data, time=timestamp, output_properties={"text_tokens"})
```
joejuzl, line 467: Maybe we should extract these output properties into a constant?

rasa/shared/core/events.py
```diff
@@ -458,6 +458,10 @@ def __init__(
         }
 
         if parse_data:
+            if "text_tokens" in parse_data:
```
joejuzl, line 461: Do we have a constant for the `"text_tokens"` key?

joejuzl (changes requested):
Could we also have tests for tracker stores and event brokers, to make sure that the serialisation of the tokens doesn't break anything there?

rasa/shared/core/events.py
ancalita: Yup, it's [here](https://github.com/RasaHQ/rasa/blob/819c339f3f5cdf32593cc725ff7d8f4a8253ea3a/rasa/nlu/constants.py#L55) - would replacing this with `TOKENS_NAMES.get(rasa.shared.nlu.constants.TEXT)` be alright?

ancalita:
@joejuzl I spent quite a bit of time on Friday figuring out how to mock the event broker in order to be able to check the message after being published. Seems it cannot be done without actually connecting to an instance of the event broker, so I tried to write an integration test using the existing pika integration test - however locally this test fails via PyCharm when trying to connect, I get a `ConnectionRefused` error. Is the right route the one outlined in the [README](https://github.com/RasaHQ/rasa#running-the-integration-tests)?

I might also need some guidance how to write the integration tests for the other types of event brokers.

Re. the tracker stores - here I should aim for unit tests or also integration tests?

rasa/shared/core/events.py
```diff
@@ -458,6 +458,10 @@ def __init__(
...
+                self.parse_data["text_tokens"] = [
+                    (t.start, t.end, t.text) for t in parse_data.pop("text_tokens")
```
wochinge, line 463: should we use a dict instead of a tuple to make the single parts easier identifiable? Cause `json.dumps` will make the tuple a list and then you might up with something like `[1,2, "3"]` which is not exactly intuitive (the list is more memory efficient though)

joejuzl: We also don't necessarily need the actual test of the token. It might be worth asking Rasa X folks if they will need it.

ancalita: Ok sure, I'll ask if they need the actual text.
What would the dict look like in both cases though:
- with text: `{t.text: [t.start, t.end]}`
- without text: `{t.start: t.end}`?

wochinge: mhm, I personally don't like dictionaries where I don't know the keys (unless I'd always just iterate over the entire thing anyway) because they are super awkward to access. I'd probably even be verbose. If they don't need the text I'd go for the tuple (that's clear enough in that case I think) (`t.start, t.end`)

joejuzl: Yeah I'd say either tuple e.g. `(1,5)` or if we need text then a full dict e.g. `{"start": 1, "end": 5, "text" "hello"}`

changelog/8924.improvement.md
```diff
@@ -0,0 +1 @@
+Send user message tokens along with the `UserUttered` event through the event broker to Rasa X.
```
jupyterjazz, line 1: Since we only send token indices and not values...
```suggestion
Send indices of user message tokens along with the `UserUttered` event through the event broker to Rasa X.
```

rasa/shared/nlu/training_data/message.py
```diff
@@ -27,8 +28,12 @@
 if typing.TYPE_CHECKING:
     from rasa.shared.nlu.training_data.features import Features
 
+TEXT_TOKENS = TOKENS_NAMES.get(rasa.shared.nlu.constants.TEXT)
+
 
 class Message:
+    """Holds NLU training data."""
```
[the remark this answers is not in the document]
ancalita: I quickly jotted a description to stop the linter from complaining, I'm happy with any suggestions to improve this class definition.

tests/core/test_processor.py
```diff
@@ -1328,3 +1328,25 @@ def test_predict_next_action_raises_limit_reached_exception(domain: Domain):
...
+        TemplatedNaturalLanguageGenerator(mood_agent.domain.responses),
+    )
+    tracker = await processor.log_message(message)
+    event = tracker.get_last_event_for(event_type=UserUttered)
+    event_tokens = event.as_dict().get("parse_data").get("text_tokens")
+
+    assert event_tokens[0][0] == 0
+    assert event_tokens[0][1] == 5
+    assert event_tokens[1][0] == 6
+    assert event_tokens[1][1] == 11
```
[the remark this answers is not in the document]
ancalita: To clarify, you're suggesting to calculate the right hand side of the assertion? Sure, can definitely do this.

changelog/8924.improvement.md
ancalita: Good catch! This was a relic of initial commits - some changes were decided upon just today 😄

tests/core/test_processor.py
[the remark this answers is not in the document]
jupyterjazz: Yes, in order to make clearer what these numbers are and where they're coming from.

rasa/shared/nlu/training_data/message.py
```diff
@@ -4,6 +4,7 @@
 import copy
 
 import rasa.shared.utils.io
+from rasa.nlu.constants import TOKENS_NAMES
```
ancalita, line 7: @joejuzl this import actually causes the test checking that the `shared` package is independent to fail.
I could potentially add a constant for TEXT_TOKENS in `rasa/shared/nlu/constants.py`?

tests/core/test_tracker_stores.py
```diff
@@ -876,3 +877,63 @@ def test_tracker_store_connection_error(config: Dict, domain: Domain):
...
+    tracker_store.save(tracker)
+
+    retrieved_tracker = tracker_store.retrieve(sender_id=sender_id)
+    event = retrieved_tracker.get_last_event_for(event_type=UserUttered)
+    event_tokens = event.as_dict().get("parse_data").get("text_tokens")
+
+    assert event_tokens[0][0] == indices[0][0]
+    assert event_tokens[0][1] == indices[0][1]
+    assert event_tokens[1][0] == indices[1][0]
+    assert event_tokens[1][1] == indices[1][1]
```
jupyterjazz, line 916: I think you can check this in one line - `assert event_tokens == indices` - if you change `indices.append((start, end))` to `indices.append([start, end])`

rasa/shared/nlu/training_data/message.py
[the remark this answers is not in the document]
ka-bu: How about something like...
```
Container for data that describes one conversation state.

More precisely, the data describes a set of attributes that also can be used to describe at most 
one use and one action sub-state. For example, these attributes can be `TEXT`  and 
`INTENT` (describing a user utterance) or `ACTION_NAME` (describing a bot action). The container 
includes raw information (`self.data`) as well as features (`self.features`) for each such attribute. 
Moreover, the container can be tagged with a timestamp and can keep track about information 
on a specific  subset of attributes (`self.output_properties`).
```
using the [notion of state](https://www.notion.so/rasa/Events-States-1a6a181807d9421e9bd019deb6161f55#e04e8ce334754e82b31e7d4480a3796e) as it is at the moment and the term "features" for real (numeric unreadable) representations (which I think is sometimes also used for "attributes" - not that the terms "features" and "attributes" are really different -  but but we use them for very different things and should maybe use them like this (to avoid "featurized features" and instead have "features = featurization of (raw) attributes").

rasa/shared/core/events.py
ancalita: To follow-up: Practice CDD confirmed they don't need the text, so I kept the tuple type `(t.start, t.end)`.

jupyterjazz (approved):
Thanks for addressing the comments. Looks good for me :rocket:

rasa/shared/nlu/training_data/message.py
```diff
@@ -29,6 +30,14 @@
 
 
 class Message:
+    """Container for data that describes one conversation state.
+
+    The data describes a set of attributes that can also be used to describe
+    at most one use and one action sub-state. For example, these attributes
```
jupyterjazz, line 36: I think this is a typo
```suggestion
    at most one user and one action sub-state. For example, these attributes
```

ancalita: let me double-check with @ka-bu - in your Message docstring did you mean `use` or `user`?

rasa/nlu/model.py
```diff
@@ -464,7 +465,10 @@ def parse(
         data = self.default_output_attributes()
         data[TEXT] = text
 
-        message = Message(data=data, time=timestamp)
+        output_properties = {TEXT_TOKENS}
+        message = Message(
+            data=data, time=timestamp, output_properties=output_properties
+        )
```
jupyterjazz, line 471: Since we're not getting `TEXT_TOKENS` from `TOKENS_NAMES` anymore, should we change it like this?
```suggestion
        message = Message(
            data=data, time=timestamp, output_properties={TEXT_TOKENS}
        )
```

rasa/shared/nlu/training_data/message.py
ka-bu: sorry that was a typo, I meant "user" :) a state has sub-states describing e.g. a user utterance or a bot action

rasa/shared/nlu/training_data/message.py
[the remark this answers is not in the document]
joejuzl: I find "describes one conversation state" to be misleading, as this only captures the data of a single turn in a conversation, not the state of the conversation itself.

I also think a key piece of information about this class is it is used to pass data through the NLU pipeline and/or in end-to-end policies and that it's real purpose is to represent a user (or bot) _message_. We have other things which describe conversation state that are distinct from this as they are used in different places e.g. states from tracker featurization.

tests/core/test_tracker_stores.py
```diff
@@ -876,3 +877,60 @@ def test_tracker_store_connection_error(config: Dict, domain: Domain):
...
+def prepare_token_serialisation(
+    tracker_store: TrackerStore,
+    response_selector_interpreter: Interpreter,
+    sender_id: Text,
+):
+    text = "Good morning"
+    tokens = text.split()
+    current_index = 0
+    indices = []
+    for t in tokens:
```
joejuzl, line 891: You could use the `WhitespaceTokenizer` to do this for you?

joejuzl (changes requested):

rasa/shared/nlu/training_data/message.py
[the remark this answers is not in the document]
ka-bu: Regarding conversation state vs turns: True, but you could also put information for both, user and action substate in there at the moment (and nothing would keep us from doing so). E.g. the logic in the TrackerFeaturizers even relies on the fact that the attributes of those substates don't overlap.

> We have other things which describe conversation state

Also true but at the moment it is also that sub-states are and must be of the same shape and form as what is passed to data as "message_data" and that substate names must correspond to the attribues (of a message).

Agree that I'm too focused here on describing what is there/could be done with it as opposed to motivating how it should be used though ^^. Feel free to rephrase.

How about...
```
Container for data that can be used to describe a conversation turn.

The turn is described by a set of attributes such as e.g. `TEXT`  and  `INTENT` when describing a user utterance
or e.g. `ACTION_NAME` for describing a bot action. The container includes raw information (`self.data`) as well as features (`self.features`) for each such attribute. 
Moreover, the message has a timestamp and can keep track about information on a specific  subset of attributes (`self.output_properties`).
```
@joejuzl ?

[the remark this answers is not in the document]
joejuzl: Sounds good 👍

joejuzl (approved):
Nice!

PR #9251, opened 2021-08-02
Add missing severity to vulnerability report

virtualroot:
**Proposed changes**:
- Overwrite `lazy-actions/gitrivy` default of `HIGH,CRITICAL` with all severities


erohmensing (approved):
🙏🏻 we don't want UNKNOWN? (that's fine, just asking!)

erohmensing:
ah I see you've copied what we have for `rasa`, make sense 👍🏻

PR #9266, opened 2021-08-04
Fix doc push

m-vdb:
**Proposed changes**:
- Fix "Publish Documentation" workflow
- I've also re-run the fixed workflow locally to make sure the docs are updated. Build is running [here](https://app.netlify.com/teams/rasahq/builds/610ab3c915a9c90007571183)

cc @aeshky as you ran the release of 2.8.2 and uncovered the issue 😄 thanks!

### DIFF
--- a/scripts/push_docs_to_branch.sh
+++ b/scripts/push_docs_to_branch.sh
@@ -48,7 +48,7 @@ then
     EXISTING_VERSION=$NEW_VERSION
 fi
 
-# install yarn dependency in the temp directory
+# install yarn dependencies in the temp directory
 cd $TMP_DOCS_FOLDER/docs && yarn install && cd - || exit 1
 
 if [ ! -z "$EXISTING_VERSION" ]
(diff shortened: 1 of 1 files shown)


scripts/push_docs_to_branch.sh
```diff
@@ -48,6 +48,9 @@ then
     EXISTING_VERSION=$NEW_VERSION
 fi
 
+# install yarn dependency in the temp directory
+cd $TMP_DOCS_FOLDER/docs && yarn install && cd - || exit 1
```
m-vdb, line 52: no idea why it didn't fail before, maybe some sort of global cache in `yarn`?! not an expert

m-vdb:
I've already requested someone from infra -- let's keep Tobias out of this one 😄

scripts/push_docs_to_branch.sh
mprazz: I have no clue either tbh :sweat_smile: I'm way less aware of the `rasa` side compared to the `rasa-x` side :grimacing:

mprazz (approved):
LGTM :medal_military:

mprazz:
if it works now, that's the important part -- doing a yarn install is not a destructive action so I don't think we should be overly concerned about this.

scripts/push_docs_to_branch.sh
m-vdb: I do think we'll have the same pb for Rasa X 😄

PR #9269, opened 2021-08-04
3.0 architecture revamp/integration branch

joejuzl:
**Proposed Changes**
Implement the "engine" part of the [architecture revamp](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e)
- datastructures to represent graphs and graph nodes
- logic to run graphs
- backends which provide persistence for graph components
- backend for caching training results in between
- logic for pruning the training graph based on the caching results


wochinge:
@ka-bu @alwx Can you please review the following parts (and their corresponding tests) 🙌🏻 

@alwx :
- `runner/`
- `pyproject.toml`
- `caching.py`
- `storage/`
- `exceptions.py`
- `loader.py`

@ka-bu :
- `training/`
- `graph.py`
- `validation.py`

rasa/engine/training/graph_trainer.py
```diff
@@ -0,0 +1,226 @@
...
+                pruned_schema, target_node_name, fingerprint_run_outputs
+            )
+
+        return pruned_schema.minimal_graph_schema()
+
+    def _walk_and_prune(
+        self,
+        schema: GraphSchema,
+        current_node_name: Text,
+        fingerprint_run_outputs: Dict[Text, Union[FingerprintStatus, Any]],
```
ka-bu, line 170: Why is the "Any" needed?

ka-bu (commented):
Just some small questions/comments so far. Will look through training/hooks, validation and tests (and everything else again ^^) on Monday.

pyproject.toml
```diff
@@ -95,7 +95,7 @@ colorclass = "~2.2"
 terminaltables = "~3.1.0"
 sanic = ">=19.12.2,<20.12.0"
 sanic-cors = "^0.10.0b1"
-sanic-jwt = ">=1.3.2,<2.0"
+sanic-jwt = ">=1.3.2,<1.7.0"
```
wochinge, line 98: the newly released 1.7 seems incompatible with our code: https://github.com/RasaHQ/rasa/pull/9269/checks?check_run_id=3321485704

rasa/engine/graph.py
```diff
@@ -0,0 +1,479 @@
...
+                themselves.
+            resource: Resource locator for this component which can be used to persist
+                and load itself from the `model_storage`.
+            execution_context: Information about the current graph run.
+            kwargs: Output values from previous nodes might be passed in as `kwargs`.
+
+        Returns:
+            An instantiated, loaded `GraphComponent`.
+        """
+        return cls.create(config, model_storage, resource, execution_context)
```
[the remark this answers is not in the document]
wochinge: Using `create` allows us to enforce the signature via an interface which we can't with constructors. And the ability to use different constructors is e.g. used for training (create) vs inference (`load` a trained component)

rasa/engine/graph.py
```diff
@@ -0,0 +1,479 @@
...
+            for dependency in target_dependencies:
+                required += self._all_dependencies_schema([dependency])
+
+        return required
+
+
+class GraphComponent(ABC):
+    """Interface for any component which will run in a graph."""
+
+    # TODO: This doesn't enforce that it exists in subclasses..
```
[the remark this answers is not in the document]
wochinge: yeah, I think that makes sense. What do you think @joejuzl ?

rasa/engine/graph.py
```diff
@@ -0,0 +1,479 @@
...
+
+        return GraphSchema(
+            {
+                node_name: node
+                for node_name, node in self.nodes.items()
+                if node_name in dependencies
+            }
+        )
+
+    def _all_dependencies_schema(self, targets: List[Text]) -> List[Text]:
```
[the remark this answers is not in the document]
wochinge: mhm, we are currently doing that by just running the graph. I can add a check to the graph validation code 👍🏻

rasa/engine/graph.py
```diff
@@ -0,0 +1,479 @@
...
+        resource: Resource,
+        execution_context: ExecutionContext,
+        **kwargs: Any,
+    ) -> GraphComponent:
+        """Creates a component using a persisted version of itself.
+
+        If not overridden this method merely calls `create`.
+
+        Args:
+            config: This config overrides the `default_config`.
```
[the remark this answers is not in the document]
wochinge: I think we should improve the docstring. What's meant is `config: The config for this graph component. This is the default config of the component merged config specified by the user`. (does that make more sense?) . What we essentially do is `config={**default_config, **user_config}`

rasa/engine/graph.py
```diff
@@ -0,0 +1,479 @@
...
+        inputs_from_previous_nodes: Tuple[Dict[Text, Any]]
+    ) -> Dict[Text, Any]:
+        return dict(ChainMap(*inputs_from_previous_nodes))
+
+    def __call__(self, *inputs_from_previous_nodes: Dict[Text, Any]) -> Dict[Text, Any]:
+        """Calls the `GraphComponent` run method when the node executes in the graph.
+
+        Args:
+            *inputs_from_previous_nodes: The output of all parent nodes. Each is a
+                dictionary with a single item mapping the node's name to its output.
```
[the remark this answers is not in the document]
wochinge: yeah, it could 👍🏻 I personally think either is good. Would you prefer a tuple?

rasa/engine/training/graph_trainer.py
```diff
@@ -0,0 +1,226 @@
...
+        logger.debug("Running the train graph in fingerprint mode.")
+        fingerprint_run_outputs = fingerprint_graph_runner.run()
+
+        pruned_training_schema = self._prune_schema(
+            train_schema, fingerprint_run_outputs
+        )
+        return pruned_training_schema
+
+    def _create_fingerprint_schema(self, train_schema: GraphSchema) -> GraphSchema:
+        fingerprint_schema = copy.deepcopy(train_schema)
```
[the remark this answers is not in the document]
wochinge: Could do but that would only apply to custom graph schemas in my opinion. Our `default.v1` recipe which converts our current config format to a graph schema shouldn't create any schemas with nodes which can be skipped by default

rasa/engine/training/graph_trainer.py
```diff
@@ -0,0 +1,226 @@
...
+            # any input data to the graph. This means we can prune according to what
+            # has actually changed.
+            if not schema_node.is_input:
+                FingerprintComponent.replace_schema_node(schema_node, self._cache)
+        return fingerprint_schema
+
+    def _prune_schema(
+        self,
+        schema: GraphSchema,
+        fingerprint_run_outputs: Dict[Text, Union[FingerprintStatus, Any]],
```
ka-bu, line 139: .... Isn't the `Any` always a `Cachable` here?

ka-bu (commented):
Looks good to me :) Just some more minor comments / questions

rasa/engine/graph.py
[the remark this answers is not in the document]
ka-bu: yep, that sounds better :)

rasa/engine/graph.py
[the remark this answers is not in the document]
ka-bu: I think if it's always a tuple of two things then it should be a Tuple

rasa/engine/validation.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+KEYWORDS_EXPECTED_TYPES: Dict[Text, TypeAnnotation] = {
+    "resource": Resource,
+    "execution_context": ExecutionContext,
+    "model_storage": ModelStorage,
+    "config": Dict[Text, Any],
+}
+
+
```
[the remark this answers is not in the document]
wochinge: It's relatively heavyweight code which doesn't keep much state. Also it gives us a bit more flexibility when and where to use it when we have it implemented separately. 

>  E.g. _get_fn looks like it should be a classmethod in GraphComponent

I'd like to avoid mixing implementation with interface definitions


> Does it make sense to extend this in the future by some _validate_component step that triggers the graph component to expect it's context and decide whether everything is fine (eg. check that featurizers xy is present if it relies on it).

I think that could be a cool idea 👍🏻 as far as I know we don't need it at the moment though. Hence, I'd defer to add it when we need it.

rasa/engine/storage/resource.py
```diff
@@ -0,0 +1,70 @@
...
+class Resource:
+    """Represents a persisted graph component in the graph."""
+
+    name: Text
+
+    @classmethod
+    def from_cache(
+        cls, node_name: Text, directory: Path, model_storage: ModelStorage
+    ) -> Resource:
+        """Loads a `Resource` from the cache.
```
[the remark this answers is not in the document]
wochinge: It moves the resource from the cache to the current `model_storage`. So `from_cache` makes sense as it "initializes the `Resource` from the cached directory"

rasa/engine/training/graph_trainer.py
wochinge: Nope. You could say that `Any` is always a `Fingerprintable` though

rasa/engine/caching.py
```diff
@@ -0,0 +1,450 @@
...
+            model_storage: The current model storage (e.g. used when restoring
+                `Resource` objects so that they can fill the model storage with data).
+
+        Returns:
+            Instantiated `Cacheable`.
+        """
+        ...
+
+
+class LocalTrainingCache:
```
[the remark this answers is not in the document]
wochinge: You meant that `Storable` would be used for caching and persisting and the information whether something is cacheable would be provided separately?

rasa/engine/storage/resource.py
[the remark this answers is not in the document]
ka-bu: Oh ok I see, the above suggestion doesn't make sense - should be "from directory". And the "Cache" only plays a role insofar as we assume the directory has been created by the cache.

rasa/engine/training/graph_trainer.py
```diff
@@ -0,0 +1,226 @@
...
+            execution_context=ExecutionContext(graph_schema=pruned_training_schema),
+            hooks=hooks,
+        )
+
+        logger.debug("Running the pruned train graph with real node execution.")
+
+        graph_runner.run()
+
+        domain = Domain.from_path(domain_path)
+        model_metadata = self._model_storage.create_model_package(
```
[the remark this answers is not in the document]
wochinge: > Is the "I want to change these parameters for this prediction run" feature already included somewhere?

Not here (at the moment). That should be done in the graph loader.
This one is really just in case just want to use a model directly after training 

>  but there's no good reason to separate those 🤔

Why not? Loading really just loads it while graph trainer does so much more stuff (fingerprinting, pruning, packaging)🤔

rasa/engine/graph.py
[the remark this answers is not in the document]
wochinge: Makes sense to me, what do you think @joejuzl ?

rasa/engine/training/graph_trainer.py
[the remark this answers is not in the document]
ka-bu: > Why not?
I just didn't want to cause too much unnecessary rework with my comments :D Conceptually, I'd think it would be great to separate those two things

rasa/engine/training/graph_trainer.py
ka-bu: Hm, but `CachedComponent` has a `Cachable` `self._output` and `replace_schema_node` configures the class to be a `CachedComponent`  and the config to have output set to something that should be.... `Cachable` then?

rasa/engine/training/graph_trainer.py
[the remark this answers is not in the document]
wochinge: Ah, sorry, I've misunderstood you. I thought you'd prefer to merge them. So you're saying that the graph trainer should not return a loaded model at all to not duplicate work from the "graph loader"?

rasa/engine/caching.py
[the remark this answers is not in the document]
wochinge: I think we should definitely keep the `Storable` idea in mind. This was one of the areas where Joe and I struggled the most to find some good abstractions. Will need to think some more about this.

rasa/engine/training/graph_trainer.py
wochinge: `_prune_schema` operates on the entire graph after the fingerprint run. During the fingerprint run we run `FingerprintComponent` (which return `FingerprintStatus` objects) **and** input nodes (which return `Any`/ `Fingerprintable`). Input nodes usually load files and return something which not necessarily has to be cacheable. Only `fingerprintable`. after `_prune_schema` we either only have `CachedComponents` with a `Cacheable` output in the graph or a mix of input, `Cacheable` and other nodes which need running.

rasa/engine/graph.py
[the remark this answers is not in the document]
joejuzl: yup sounds good to me 👍

rasa/engine/graph.py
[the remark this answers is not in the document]
joejuzl: 👍

rasa/engine/training/graph_trainer.py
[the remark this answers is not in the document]
ka-bu: yep - but I guess there's no need to e.g. have two predict_graph definitions (in that case separating those two things would be nice)

rasa/engine/training/graph_trainer.py
ka-bu: That workflow makes sense to me. What I'm struggling with is that `self._output` of a `CachedComponent` should always be `Cacheable` but it is ok that `replace_schema_node(cls, node: SchemaNode, output: Any)` uses `Any` (and what `replace_schema_node` does is just describe how a `CachedComponent` should be initialised... with an output of type `Any` and not something `Cacheable`). .... So should the type annotations in `CachedComponent` just be `Any` or am I missing something here?

joejuzl: You are right, that is inconsistent. The thing is, we use CachedComponent for things where the output was never actually in the cache (see `rasa/engine/training/graph_trainer.py:221`).
So actually `CachedComponent._output` should just be `Any`. 
It doesn't really care if the value came from a cache or not. I guess the component could also have been called something like `ConstantValueComponent`.

wochinge:
@ka-bu Implemented your review comments on this PR https://github.com/RasaHQ/rasa/pull/9369

rasa/engine/training/graph_trainer.py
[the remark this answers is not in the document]
wochinge: I think it's a convenience functionality for now that the trainer returns the instantiated model. We should check in the end if we are actually using this or if we can just return `None`

rasa/engine/training/graph_trainer.py
ka-bu: Since the tests run with non-cacheable things there should also be nothing downstream that expects them to have "from_cache" / where that output might have to be loaded from disk? - We could also call it something like `OutputInMemoryComponent` (doesn't look nice but maybe closer to what it does? 🤔 )

pyproject.toml
```diff
@@ -148,7 +151,6 @@ pydoc-markdown = "^4.1.6"
 pytest-timeout = "^1.4.2"
 mypy = "^0.910"
 bandit = "^1.6.3"
-typing-extensions = "^3.7.4"
```
alwx, line 151: Why do we need to move it from dev dependencies?

rasa/engine/training/graph_trainer.py
wochinge: Or `PrecomputedValueProvider` (as we call all the other things `*Provider` as well e.g. `TrainingDataProvider` or `DomainProvider`). Or `InmemoryProvider`?

rasa/engine/storage/storage.py
```diff
@@ -0,0 +1,139 @@
...
+from rasa.engine.storage.resource import Resource
+from rasa.shared.core.domain import Domain
+from rasa.engine.graph import GraphSchema
+
+
+logger = logging.getLogger(__name__)
+
+
+class ModelStorage(abc.ABC):
+    """Serves as storage backend for `GraphComponents` which need persistance."""
```
alwx, line 19:
```suggestion
    """Serves as storage backend for `GraphComponents` which need persistence."""
```

rasa/engine/caching.py
```diff
@@ -0,0 +1,450 @@
...
+
+            while (
+                rasa.utils.common.directory_size_in_mb(
+                    self._cache_location,
+                    filenames_to_exclude=[self._cache_database_name],
+                )
+                + output_size
+                > self._max_cache_size
+            ):
+                self._drop_least_recently_used_item()
```
alwx, line 331: Can it also be that the `self._max_cache_size` is too small or the `output_size` is too big and dropping least recently used item couldn't help freeing up the space needed to store in cache?

rasa/engine/caching.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def _get_cached_result(
+        self, output_fingerprint_key: Text
+    ) -> Tuple[Optional[Path], Optional[Text]]:
+        with self._sessionmaker.begin() as session:
+            query = sa.select(
+                self.CacheEntry.result_location, self.CacheEntry.result_type
+            ).where(
+                self.CacheEntry.output_fingerprint_key == output_fingerprint_key,
+                self.CacheEntry.result_location != sa.null(),
```
alwx, line 416: More of a question: why do we need `self.CacheEntry.result_location != sa.null()` condition? I understand that when the result location is empty then there is basically no cached output but why it's the case and in which situations can it happen?

pyproject.toml
wochinge: it's now a prod dependency as we need it for our dataclasses

rasa/engine/caching.py
wochinge: 👍🏻 That's covered by [this line](https://github.com/RasaHQ/rasa/blob/c141b28ccda13d24b7a9ae8db65f3880aa9d0890/rasa/engine/caching.py#L315)

rasa/engine/caching.py
wochinge: Every `GraphComponent` output has to be `Fingerprintable` but not every `GraphComponent` output is `Cacheable`. This means that we sometimes only store the fingerprints of the output but not the cached output. This is e.g. allows cases like this

```
# Train
A (not cachable, no target node) -> B (cacheable, target node)

# Fingerprint run
A (fingerprint didn't change)  -> B (fingerprint didn't change)

# This leads to the "train graph" only consisting of a cached output of B 
# although we don't have a cached version of A
B (using cached version)

```

Does that make sense?

rasa/engine/training/graph_trainer.py
wochinge: @ka-bu @joejuzl I think this is the last open question 🎉

rasa/engine/caching.py
alwx: 👍

rasa/engine/caching.py
alwx: Yes, makes sense! Thanks for the explanation

alwx:
The caching / storage part seems relatively simple and clear. Great job! 
(will take a closer look at the rest of it today but mostly for my own education purposes)

alwx (approved):

rasa/engine/training/graph_trainer.py
joejuzl: Personally I would probably just leave it as is - as I think in most situations it's the most understandable name. However I don't feel super strongly.

wochinge: @ka-bu What do you think? I feel the same as Joe (we can also always rename it somewhen later)

ka-bu: I'd prefer one of the Cache-free *provider names cause my brain is strongly like "caching = fingerprinting+store"

joejuzl: But we only use these after the fingerprint run, i.e. in a cached run. It's just that it can be wrapping the output of an input node directly to stop running the input nodes twice.

wochinge: went with `PrecomputedValueProvider`

wochinge:
Enabled auto merge 🔥

rasa/engine/training/graph_trainer.py
wochinge: That's true but I think the new name still makes it purpose very clear.

Issue #9068, opened 2021-07-07
Support spaCy 3.1


koaning:
SpaCy 3.1 [just got released](https://explosion.ai/blog/spacy-v3-1) and it supports a few new languages and features. It'd make sense for us to update for Rasa 3.0.

koaning:
Related: https://github.com/RasaHQ/rasa/issues/8882

TyDunn:
@koaning are you working on this?

koaning:
It's on my todo. I figured with all the Rasa 3.0 stuff that's happening it was low priority and best delayed but I can pick it up at a moments notice.

TyDunn:
So you would do it as part of 3.0 or no? We should factor it into the action plan, if you intend to it to be in the release

koaning:
Ah, pardon! I should've been more explicit. Yes! 

Let's put it into the release, it shouldn't be too much work. I might be able to wrap it up tomorrow. 

Just to confirm, I can push to master, or should I push to one of the 3.0 branches?

TyDunn:
I think you can just push to `main` but let's double check with @m-vdb

m-vdb:
yes, push to `main`!

PR #9284, opened 2021-08-06
Upgrade to spaCy 3.1

koaning:
Fixes https://github.com/RasaHQ/rasa/issues/9068 by updating poetry file and running `poetry update`.


joejuzl:
Do we envision using any of the functionality in Rasa OS that is only in 3.1?
Just asking because the current version constraint (`^3.0`) does allow for 3.1.

koaning:
I suppose only indirectly. The main reason for adding it is because the spaCy folks will have removed my bugs and there's a new language available.

joejuzl (approved):

Issue #9181, opened 2021-07-20
Rasa 3.0 Architecture Revamp: `GraphSchema` validation.


joejuzl:
**Background**:
Context [here](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e).

**Description**:
This issue regards validating a graph schema.

This includes:
- The run function for each GraphComponent has all its required parameters provided by its parents.
- The return value type of a node matches the function argument type of its children.
- The GraphComponent class can be loaded.
- All required packages for the GraphComponent are installed.
- The node config matches the GraphComponent config.

This does **not** include:
- Any recipe level validation.
- Any training data based validation.

**Blockers** (if relevant):
- https://github.com/RasaHQ/rasa/issues/9176

**Definition of Done**:
- [ ] Implemented.
- [ ] Unittested.
- [ ] Functionally tested using test components.

wochinge:
we also need to validate the usage of any reserved keywords like `resource` (see discussion [here](https://github.com/RasaHQ/rasa/pull/9233#discussion_r681686999))

PR #9294, opened 2021-08-06
implement graph schema validation

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9181
- Add validation for graph schemas
  - [x] The run and constructor methods exist
  - [x] The run function for each GraphComponent has all its required parameters provided by its parents.
  - [x] The constructor function for any non-eager GraphComponent has all its required parameters provided by its parents.
  - [x] The return value type of a node matches the function argument type of its children.
  - [x] All reserved parameter names are of the correct type. e.g. `resource` is a `Resource`.
  - [x] All return values for training graph implement `GraphComponentTrainingOutput`
  - [x] All required packages for the GraphComponent are installed.
  - [x] The components all support the language.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,6 +124,7 @@ aio-pika = "^6.7.1"
 pyTelegramBotAPI = "^3.7.3"
 dask = "2021.7.2"
 typing-extensions = "^3.7.4"
+typing-utils = "^0.1.0"
 
 [tool.poetry.dev-dependencies]
 pytest-cov = "^2.10.0"
--- a/rasa/engine/exceptions.py
+++ b/rasa/engine/exceptions.py
@@ -8,3 +8,7 @@ class GraphComponentException(Exception):
 
 class GraphSchemaException(Exception):
     """Represents errors when dealing with `GraphSchema`s."""
+
+
+class GraphSchemaValidationException(Exception):
+    """Indicates that the given graph schema is invalid."""
--- a/rasa/engine/graph.py
+++ b/rasa/engine/graph.py
@@ -8,7 +8,12 @@ from dataclasses import dataclass, field
 import logging
 from typing import Any, Callable, Dict, List, Optional, Text, Type, Tuple
 
-from rasa.engine.exceptions import GraphComponentException, GraphSchemaException
+from typing_extensions import runtime_checkable, Protocol
+
+from rasa.engine.exceptions import (
+    GraphComponentException,
+    GraphSchemaException,
+)
 import rasa.shared.utils.common
 from rasa.engine.storage.resource import Resource
 
@@ -17,6 +22,15 @@ if typing.TYPE_CHECKING:
 
 
 logger = logging.getLogger(__name__)
+
+
+@runtime_checkable
+class Fingerprintable(Protocol):
+    """Protocol which each node output has to implement in the train graph."""
+
+    def fingerprint(self) -> Text:
+        """Returns a unique stable fingerprint for itself."""
+        ...
 
 
 @dataclass
@@ -147,14 +161,16 @@ class GraphComponent(ABC):
         """
         return cls.create(config, model_storage, resource, execution_context)
 
-    def supported_languages(self) -> Optional[List[Text]]:
+    @staticmethod
+    def supported_languages() -> Optional[List[Text]]:
         """Determines which languages this component can work with.
 
         Returns: A list of supported languages, or `None` to signify all are supported.
         """
         return None
 
-    def required_packages(self) -> List[Text]:
+    @staticmethod
+    def required_packages() -> List[Text]:
         """Any extra python dependencies required for this component to run."""
         return []
 
--- a/rasa/engine/validation.py
+++ b/rasa/engine/validation.py
@@ -121,7 +121,7 @@ def _validate_required_packages(node: SchemaNode, node_name: Text) -> None:
 
 def _get_parameter_information(
     node_name: Text, uses: Type[GraphComponent], method_name: Text
-) -> Tuple[Dict[Text, ParameterInfo], Optional[Type]]:
+) -> Tuple[Dict[Text, ParameterInfo], TypeAnnotation]:
     fn = _get_fn(node_name, uses, method_name)
 
     type_hints = _get_type_hints(node_name, uses, fn)
--- a/rasa/engine/validation.py
+++ b/rasa/engine/validation.py
@@ -20,17 +20,20 @@ from rasa.engine.storage.resource import Resource
 from rasa.engine.storage.storage import ModelStorage
 
 
+TypeAnnotation = Union[TypeVar, Text, Type]
+
+
 @dataclasses.dataclass
 class ParameterInfo:
     """Stores metadata about a function parameter."""
 
-    type_annotation: Union[TypeVar, Text, Type]
+    type_annotation: TypeAnnotation
     # `True` if we have a parameter like `**kwargs`
     is_variable_length_keyword_arg: bool
     has_default: bool
 
 
-KEYWORDS_EXPECTED_TYPES: Dict[Text, Union[TypeVar, Text, Type]] = {
+KEYWORDS_EXPECTED_TYPES: Dict[Text, TypeAnnotation] = {
     "resource": Resource,
     "execution_context": ExecutionContext,
     "model_storage": ModelStorage,
@@ -146,7 +149,7 @@ def _get_parameter_information(
 
 def _get_type_hints(
     node_name: Text, uses: Type[GraphComponent], fn: Callable
-) -> Dict[Text, Union[TypeVar, Text, Type]]:
+) -> Dict[Text, TypeAnnotation]:
     try:
         return typing.get_type_hints(fn)
     except NameError as e:
@@ -183,7 +186,7 @@ def _validate_run_fn(
     node_name: Text,
     node: SchemaNode,
     run_fn_params: Dict[Text, ParameterInfo],
-    run_fn_return_type: Union[TypeVar, Text, Type],
+    run_fn_return_type: TypeAnnotation,
     is_train_graph: bool,
 ) -> None:
     _validate_types_of_reserved_keywords(run_fn_params, node_name, node, node.fn)
@@ -329,7 +332,7 @@ def _validate_parent_return_type(
     node: SchemaNode,
     parent_name: Text,
     parent: SchemaNode,
-    required_type: Union[TypeVar, Text, Type],
+    required_type: TypeAnnotation,
 ) -> None:
     _, parent_return_type = _get_parameter_information(
         parent_name, parent.uses, parent.fn
--- a/rasa/engine/validation.py
+++ b/rasa/engine/validation.py
@@ -1,0 +1,342 @@
+import inspect
+import logging
+import typing
+from typing import Optional, Callable, Text, Tuple, Dict, Type, Any, Set, Union, TypeVar
+
+import dataclasses
+
+import rasa.utils.common
+import typing_utils
+
+from rasa.engine.exceptions import GraphSchemaValidationException
+from rasa.engine.graph import (
+    GraphSchema,
+    GraphComponent,
+    SchemaNode,
+    Fingerprintable,
+    ExecutionContext,
+)
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+
+
+@dataclasses.dataclass
+class ParameterInfo:
+    """Stores metadata about a function parameter."""
+
+    type_annotation: Union[TypeVar, Text, Type]
+    # `True` if we have a parameter like `**kwargs`
+    is_variable_length_keyword_arg: bool
+    has_default: bool
+
+
+KEYWORDS_EXPECTED_TYPES: Dict[Text, Union[TypeVar, Text, Type]] = {
+    "resource": Resource,
+    "execution_context": ExecutionContext,
+    "model_storage": ModelStorage,
+    "config": Dict[Text, Any],
+}
+
+
+def validate(
+    schema: GraphSchema, language: Optional[Text], is_train_graph: bool
+) -> None:
+    """Validates a graph schema.
+
+    This tries to validate that the graph structure is correct (e.g. all nodes pass the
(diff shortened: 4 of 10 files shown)


rasa/engine/exceptions.py
```diff
@@ -8,3 +8,7 @@ class GraphComponentException(Exception):
 
 class GraphSchemaException(Exception):
     """Represents errors when dealing with `GraphSchema`s."""
+
+
+class GraphSchemaValidationException(Exception):
```
wochinge, line 13: probably worth distinguishing the errors a little bit better

rasa/engine/graph.py
```diff
@@ -19,6 +23,12 @@
 logger = logging.getLogger(__name__)
 
 
+@typing.runtime_checkable
+class Fingerprintable(typing.Protocol):
```
wochinge, line 27: remove this once `GraphTrainer` is merged

wochinge:
@joejuzl PR is here. I'll do some more polishing on Monday.

rasa/engine/exceptions.py
wochinge: decided to not to do it for  as we don't need finer distinction at the moment

pyproject.toml
```diff
@@ -124,6 +124,7 @@ aio-pika = "^6.7.1"
 pyTelegramBotAPI = "^3.7.3"
 dask = "2021.7.2"
 typing-extensions = "^3.7.4"
+typing-utils = "^0.1.0"
```
wochinge, line 127: not perfect but as it's only used in location I think it's okay for now

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+    "config": Dict[Text, Any],
+}
+
+
+def validate(
+    schema: GraphSchema, language: Optional[Text], is_train_graph: bool
+) -> None:
+    """Validates a graph schema.
+
+    This checks that the graph structure is correct (e.g. all nodes pass the correct
```
joejuzl, line 44: Maybe "checks that the graph structure is correct" is too big a claim?

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+    things into each other) and validates the single graph components.
```
joejuzl, line 45:
```suggestion
    things into each other) and validates the individual graph components.
```

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+
+        _validate_needs(
+            node_name, node, schema, create_fn_params, run_fn_params,
+        )
+
+
+def _validate_interface_usage(node_name: Text, node: SchemaNode) -> None:
+    if not issubclass(node.uses, GraphComponent):
+        raise GraphSchemaValidationException(
+            f"Node '{node_name}' uses class '{node.uses.__name__}'. This module does "
```
joejuzl, line 82:
```suggestion
            f"Node '{node_name}' uses class '{node.uses.__name__}'. This class does "
```

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+
+    Args:
+        schema: The schema which needs validating.
+        language: Used to validate if all components support the language the assistant
+            is used in. If the language is `None`, all components are assumed to be
+            compatible.
+        is_train_graph: Whether the graph is used for training.
+
+    Raises:
+        GraphSchemaValidationException: If the validation failed.
```
joejuzl, line 55: Very Optional: Did you consider collecting all the errors and returning/displaying them all at the end rather than the user having to do one-by-one?

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        _validate_constructor(node_name, node, create_fn_params)
+
+        _validate_needs(
+            node_name, node, schema, create_fn_params, run_fn_params,
+        )
+
+
+def _validate_interface_usage(node_name: Text, node: SchemaNode) -> None:
+    if not issubclass(node.uses, GraphComponent):
```
joejuzl, line 80: Unfortunately doesn't look at types or we wouldn't have to check the keyword parameters later...

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+
+def _get_fn(node_name: Text, uses: Type[GraphComponent], method_name: Text) -> Callable:
+    fn = getattr(uses, method_name, None)
+    if fn is None:
+        raise GraphSchemaValidationException(
+            f"Node '{node_name}' uses graph component '{uses.__name__}' which does not "
+            f"have the specified "
+            f"method '{method_name}'. Please make sure you're either using "
+            f"the right graph component or specifying a valid method "
+            f"for this component."
```
joejuzl, line 166: "...for the 'fn' and 'constructor_name' options".

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+            f"Failed to retrieve type annotations for component "
+            f"'{uses.__name__}' due to error:\n{e}"
+        )
+        raise GraphSchemaValidationException(
+            f"Node '{node_name}' uses graph component '{uses.__name__}' which has "
+            f"type annotations in its method '{fn.__name__}' which failed to be "
+            f"retrieved. Please make sure remove any forward "
+            f"reference by removing the quotes around the type "
+            f"(e.g. 'def foo() -> \"int\"' becomes 'def foo() -> int'. and make sure "
+            f"all type annotations can be resolved during runtime."
```
joejuzl, line 154: Should we mention they may need `from __future__ import annotations`?

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+from rasa.engine.storage.storage import ModelStorage
+from rasa.nlu import components
+
+
+@dataclasses.dataclass
+class ParameterInfo:
+    """Stores metadata about a function parameter."""
+
+    type_annotation: Any
+    is_kwargs: bool
```
joejuzl, line 27: Can we explain this with a comment or clearer name. It took me a while to realise this means `**kwargs` rather than any singular keyword argument.

joejuzl: Could we be caught out if someone uses `*args`?

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+
+def _validate_constructor(
+    node_name: Text, node: SchemaNode, create_fn_params: Dict[Text, ParameterInfo],
+) -> None:
+    _validate_types_of_reserved_keywords(
+        create_fn_params, node_name, node, node.constructor_name
+    )
+
+    for param_name in _required_args(create_fn_params):
+        if node.eager:
```
joejuzl, line 246: This doesn't need to be run in every iteration.

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+    node: SchemaNode,
+    graph: GraphSchema,
+    create_fn_params: Dict[Text, ParameterInfo],
+    run_fn_params: Dict[Text, ParameterInfo],
+) -> None:
+    available_args, has_kwargs = _get_available_args(
+        node, create_fn_params, run_fn_params
+    )
+
+    for param_name, parent_name in node.needs.items():
```
joejuzl, line 273: What about if we have an "available_args" which is not provided by the "needs"? Do we not need to check in that direction too?

tests/engine/test_validation.py
```diff
@@ -0,0 +1,511 @@
...
+
+
+def test_graph_component_is_no_graph_component():
+    class MyComponent:
+        def other(self) -> TrainingData:
+            pass
+
+    schema = create_test_schema(uses=MyComponent)
+
+    with pytest.raises(GraphSchemaValidationException):
```
joejuzl, line 69: Is it worth checking some of the error message in all these tests just to confirm it's failing for the correct reason?

tests/engine/test_validation.py
```diff
@@ -0,0 +1,511 @@
...
+        def run(self) -> Any:
+            pass
+
+    schema = create_test_schema(uses=MyComponent)
+
+    with pytest.raises(GraphSchemaValidationException):
+        validation.validate(schema, language=None, is_train_graph=True)
+
+
+def test_graph_output_is_not_fingerprintable_None():
```
joejuzl, line 112: can we parameterize the `test_graph_output_is_not_fingerprintable_...` ones?

tests/engine/test_validation.py
```diff
@@ -0,0 +1,511 @@
...
+        ) -> GraphComponent:
+            pass
+
+    schema = create_test_schema(uses=MyComponent)
+
+    with pytest.raises(GraphSchemaValidationException):
+        validation.validate(schema, language=None, is_train_graph=True)
+
+
+def test_graph_constructor_execution_context_wrong_type():
```
joejuzl, line 236: could parameterize the `test_graph_constructor_..._wrong_type` tests?

tests/engine/test_validation.py
```diff
@@ -0,0 +1,511 @@
...
+        def supported_languages() -> Optional[List[Text]]:
+            return supported_languages
+
+    schema = create_test_schema(uses=MyComponent)
+
+    validation.validate(schema, language=current_language, is_train_graph=False)
+
+
+@pytest.mark.parametrize(
+    "required_packages", [["pytorch"], ["tensorflow", "kubernetes"]]
```
joejuzl, line 290: Guess we can now never use pytorch! 😝

joejuzl (changes requested):
Super cool!! Really glad we are doing this is v1 - I think it will make a huge difference 💯 
And really great unit testing 👍

rasa/engine/validation.py
wochinge: mhm, interesting idea 🤔 I haven't considered that. Might also obfuscate the code a lot as some validations require certain information to be present from earlier validations (e.g. checking type hints requires a successful retrieval of the function and its type hints in the first place). I'd go for this (simpler) solution for now as most graphs should be generated from the `Recipe`s anyway.

rasa/engine/validation.py
wochinge: Well, as the `run fn` is somewhat arbitrary as well as the `loads` signature the protocol check would never really help us 😬

rasa/engine/validation.py
```diff
@@ -0,0 +1,324 @@
...
+        raise GraphSchemaValidationException(
+            f"Node '{node_name}' uses a component '{node.uses.__name__}' whose method "
+            f"'{node.fn}' does not return a fingerprintable "
+            f"output. This is required for caching. Please make sure you're "
+            f"using a return type which implements the "
+            f"'{Fingerprintable.__name__}' protocol."
+        )
+
+
+def _validate_types_of_reserved_keywords(
```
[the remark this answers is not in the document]
wochinge: yep, but that wouldn't work with custom components. I didn't find a way to cleanly call mypy from Python code 😬

rasa/engine/validation.py
wochinge: That's done in `_validate_constructor` and `_validate_run_fn`

tests/engine/test_validation.py
wochinge: unfortunately not. The type annotations are unfortunately not dynamically called. I tried to parametrize types and put that in the type annotations of `MyComponent` but that doesn't work. We could try to dynamically set them via something like `MyComponent.run.__annnotations__["return"] = xy` but that in my opinion makes the tests very unreadable and very hacky.

tests/engine/test_validation.py
wochinge: same as https://github.com/RasaHQ/rasa/pull/9294/files#r685316330

rasa/engine/validation.py
wochinge: > Can we explain this with a comment or clearer name. It took me a while to realise this means **kwargs rather than any singular keyword argument.

did both 👍🏻 

> Could we be caught out if someone uses *args?
 Write

Will add a test for this.

wochinge: I'm just ignoring variable length positional arguments now. Warning about them doesn't necessarily make sense as one might use them as a way to forbid any positional arguments (as described [here](https://stackoverflow.com/a/50565247))

tests/engine/test_validation.py
wochinge: I avoided that as you might have to adapt the tests for every little string change then 🤔 Will have a look

wochinge: done

rasa/engine/validation.py
```diff
@@ -0,0 +1,342 @@
...
+)
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+
+
+@dataclasses.dataclass
+class ParameterInfo:
+    """Stores metadata about a function parameter."""
+
+    type_annotation: Union[TypeVar, Text, Type]
```
joejuzl, line 27: Might be worth having `Union[TypeVar, Text, Type]` as a variable now we use it quite often.

rasa/engine/validation.py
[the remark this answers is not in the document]
joejuzl: > yep, but that wouldn't work with custom components.

Why not? It would enforce that they implement the methods of the protocol with correct parameters and types.

> I didn't find a way to cleanly call mypy from Python code 😬

Yeah I also looked when first looking at protocols and couldn't find anything nice.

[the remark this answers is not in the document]
wochinge: > Why not? It would enforce that they implement the methods of the protocol with correct parameters and types.

True, but it would only work for `create` and `load` as the `run` fn is dynamic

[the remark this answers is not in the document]
joejuzl: Yup true - but we never have these keywords in `run` right?
Anyway pointless debate - this is good 👍

rasa/engine/validation.py
wochinge: done

rasa/engine/validation.py
joejuzl: ah yes - ofc!

joejuzl (approved):
👍👍👍

Issue #7956, opened 2021-02-15
macOS Big Sur 11.2.1 - installing virtual environment and dependencies compiler issues


twerkmeister:
**Rasa version**: master branch

**pyenv version**: 1.2.22

**Operating system** (windows, osx, ...): macOS Big Sur 11.2.1 

**Issue**: For me there were a couple of compiler problems in getting the virtual env and dependencies for development running.

1. `pyenv install` only worked for certain versions: `pyenv install 3.7.6` as shown in the readme won't work.`pyenv install 3.7.9` [turned out to work in the end among a few others](https://github.com/pyenv/pyenv/issues/1764#issuecomment-758400362)

```
> pyenv install 3.7.6
python-build: use openssl@1.1 from homebrew
python-build: use readline from homebrew
Downloading Python-3.7.6.tar.xz...
-> https://www.python.org/ftp/python/3.7.6/Python-3.7.6.tar.xz
Installing Python-3.7.6...
python-build: use readline from homebrew
python-build: use zlib from xcode sdk

BUILD FAILED (OS X 11.2.1 using python-build 20180424)

Inspect or clean up the working tree at /var/folders/g8/frp_8d7s2639spgcxqyd3kpw0000gn/T/python-build.20210215142044.19884
Results logged to /var/folders/g8/frp_8d7s2639spgcxqyd3kpw0000gn/T/python-build.20210215142044.19884.log

Last 10 log lines:
/usr/local/include/pthread.h:197:34: note: expanded from macro '_PTHREAD_SWIFT_IMPORTER_NULLABILITY_COMPAT'
        defined(SWIFT_CLASS_EXTRA) && (!defined(SWIFT_SDK_OVERLAY_PTHREAD_EPOCH) || (SWIFT_SDK_OVERLAY_PTHREAD_EPOCH < 1))
                                        ^
197 warnings generated.
197 warnings generated.
198 warnings generated.
197 warnings generated.
197 warnings generated.
197 warnings generated.
197 warnings generated.
```

proposed improvement: Possibly update Readme to use 3.7.9 right away, or add a note for macOS users.

2. dependency grpcio will run into clang errors during installation when executing `make install`.  I tried a bunch of stuff. [In the end a simple environment variable helped:](https://github.com/grpc/grpc/issues/24677#issuecomment-775458281)
`export SYSTEM_VERSION_COMPAT=1`

proposed improvement: Add a note for macOS users.

[...]

PR #9296, opened 2021-08-07
Add a support documentation for the macOS compiler issues #7956

ErickGiffoni:
Since quite a few macOS users were having some compiler trouble when getting the virtual env and dependencies for development running, I thought **Rasa should have** a document to support them.

I was able to write that new document by giving error examples and possible solutions for them that worked for me.
My intention, if the Rasa team agree, is that **the new support document will be improved as other problems should appear**.

Also, the **README should point** to this new document **so it can be easily found** by those in need.

**Proposed changes**:

- Add a new document: Development Internals Support for macOS users -> regarding #7956 
- Update README with a link to the new **macOS support guide**

**Status**:
- [ ] updated the changelog 
- [ ] reformat files using `black`

### DIFF
--- a/macOS-support.md
+++ b/macOS-support.md
@@ -1,0 +1,170 @@
+# Development Internals Support for macOS users
+
+When trying to get virtual env and dependencies running, macOS users may
+
+experience some compiler problems.
+
+If that is your case, this guide is intended to help you get things working.
+
+It was built based on common problems that were reported to the Rasa repository.
+
+It is possible that your problem is not listed here yet, but try reading the
+
+bellow because it might give you an insight.
+
+
+## Problems with pyenv
+
+
+### 1. ```pyenv install 3.7.6``` failed
+
+
+If you tried the command above and it failed, you probably got something like
+
+this:
+
+
+Error example 1
+```
+> pyenv install 3.7.6
+python-build: use openssl@1.1 from homebrew
+python-build: use readline from homebrew
+Downloading Python-3.7.6.tar.xz...
+-> https://www.python.org/ftp/python/3.7.6/Python-3.7.6.tar.xz
+Installing Python-3.7.6...
+python-build: use readline from homebrew
+python-build: use zlib from xcode sdk
+
+BUILD FAILED (OS X 11.2.1 using python-build 20180424)
+
+Inspect or clean up the working tree at /var/folders/g8/frp_8d7s2639spgcxqyd3kpw0000gn/T/python-build.20210215142044.19884
+Results logged to /var/folders/g8/frp_8d7s2639spgcxqyd3kpw0000gn/T/python-build.20210215142044.19884.log
+
+Last 10 log lines:
+/usr/local/include/pthread.h:197:34: note: expanded from macro '_PTHREAD_SWIFT_IMPORTER_NULLABILITY_COMPAT'
+        defined(SWIFT_CLASS_EXTRA) && (!defined(SWIFT_SDK_OVERLAY_PTHREAD_EPOCH) || (SWIFT_SDK_OVERLAY_PTHREAD_EPOCH < 1))
+                                        ^
+```
+
+or even this:
+
+
+Error example 2
+```
+[...]
+./Modules/posixmodule.c:8436:15: error: implicit declaration of function 'sendfile' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
+        ret = sendfile(in, out, offset, &sbytes, &sf, flags);
+              ^
+1 error generated.
+make: *** [Modules/posixmodule.o] Error 1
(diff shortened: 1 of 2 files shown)


macOS-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+experience some compiler problems.
+
+If that is your case, this guide is intended to help you get things working.
+
+It was built based on common problems that were reported to the Rasa repository.
+
+It is possible that your problem is not listed here yet, but try reading the
+
+bellow because it might give you an insight.
```
mrbaguvix, line 13:
```suggestion
below because it might give you an insight.
```

macOS-support.md
```diff
@@ -0,0 +1,170 @@
+# Development Internals Support for macOS users
+
+When trying to get virtual env and dependencies running, macOS users may
+
+experience some compiler problems.
+
+If that is your case, this guide is intended to help you get things working.
```
mrbaguvix, line 7:
```suggestion
If that is the case, this guide is intended to help you get things working.
```

macOS-support.md
```diff
@@ -0,0 +1,170 @@
+# Development Internals Support for macOS users
```
mrbaguvix, line 1:
```suggestion
# Development Internal Support for macOS users
```

macOS-support.md
```diff
@@ -0,0 +1,170 @@
...
+
+it like this:
+
+
+```$ pyenv install --patch 3.7.6 < <(curl -sSL https://github.com/python/cpython/commit/8ea6353.patch)```
+
+
+The above should work just fine, but if life doesn't go easy on you, it's possible
+
+that an error like the one bellow showed up:
```
mrbaguvix, line 78:
```suggestion
that an error like the one below showed up:
```

mrbaguvix (changes requested):

ErickGiffoni (commented):
Thank you @mrbaguvix David for reviewing!

twerkmeister:
Hey @ErickGiffoni, thanks for putting in so much effort with this PR and apologies that the fixes for the issues we had found back a while didn't make it into the proper places yet. I wasn't aware that the issue was indeed so prevalent.

Unfortunately, I think this Mac OS guide outlines solving the problem in fairly difficult manner and demands a lot from users. I understand you care about the problem and it cost you a lot of energy and time. Yet, given I had outlined in my issue to simply change the version in the readme + adding a one line note about the env variable, I think this is the right way to go in terms of user experience and not expecting users to go through a much longer and more difficult guide. Sorry again that you had to tackle this when there was already better knowledge out there.

@m-vdb let's prioritise this change so that not more people run into this. Should only be a 1 ~ as it's changing a few lines in the docs

ErickGiffoni:
Thanks for the feedback, @twerkmeister .
Although I understand that the MacOS guide requires a lot from users, I think it would be of great value for Rasa having something like it for those in pain. We could change the version in the README and add a note about the env variable because this is really a better user experience way. In this case, the macOS guide would exist in order to have something more, a plus. That’s my ideia for you.

Thanks again, you all are an amazing team. Me and my friends here will continue to search for other issues in order to contribute 🙌🏻😁

samsucik:
Hey @ErickGiffoni, I can see that there are some additions to the `ted_policy`, `test_ted_policy` and `test_unexpected_intent_policy` files, which look like leftovers from a git merge. Could you, please, clean up those cases? Ideally, your PR shouldn't need to make any changes to that code and shouldn't need a review from the Research team 🙂

ErickGiffoni:
> Hey @ErickGiffoni, I can see that there are some additions to the `ted_policy`, `test_ted_policy` and `test_unexpected_intent_policy` files, which look like leftovers from a git merge. Could you, please, clean up those cases? Ideally, your PR shouldn't need to make any changes to that code and shouldn't need a review from the Research team 🙂

Yes, I can. Sorry for the delay and the little mess. It was an unexpected error.

ErickGiffoni:
Hey, @samsucik , I'm having a little trouble here. It says I'm up to date with branch ```main```, but I think during the merge process, while I was trying to do what you asked, I accidentally changed those ```tests``` files.
Would you give me a hint or something ?

samsucik:
@ErickGiffoni I think the simplest (but not elegant) way is to just take the file contents from the `main` branch and copy-paste them in order to make the files on your branch the same...

ErickGiffoni:
@samsucik thank you, that worked just fine !
Is everything in order now ?

samsucik:
@ErickGiffoni yeah, looks good -- there are no changes to the codebase. Thanks! 🙂

samsucik:
Also, ping @twerkmeister @m-vdb since this PR now has nothing to do with the research team and is back to being only about our docs (as per your earlier comments) 🙂

PR #9304, opened 2021-08-09
8705 response selector 2.8.x fixes

ka-bu:
**Proposed changes**:
- added a check to ensure that either a label is defined or entity recognition is enabled in `DIETClassifier`
   * consequently changed the `_uses_sequential_features_for_input_text` to True in `DIETClassifier`
- removed the setdefault by removing check from `_get_sequence_feature_length`
    * replaced the missing check by renaming the function to `_get_sequence_feature_length_or_zeros` to make behaviour explicit
 - added test for "unsupported labels" check in function that returns one example per label
 - added check for sentence labels in train
    * added note on why this cannot help to detect `LexicalSyntacticFeaturizer` when it is combined with others - which works but doesn't exactly do what one would expect due to the missing sentence features - hence we only catches that weird edge case with `LexicalSyntacticFeaturizer` and nothing else
        * tried to get featurizer information from `component_config` and the `partial_pipeline` of component but both are empty when I try to run the tweaked version of the repsonseselectorbot

- (I think the `label_data` logic might be simplified - because presence of default label features already indicates what we check again when using the label_data, but I'm not going to touch that now again 😬 ) 


rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -571,7 +587,7 @@ def _create_label_data(
         needs_sentence_features = self._needs_sentence_features_for_labels()
         if not sentence_features and (not sequence_features or needs_sentence_features):
 
-            if sequence_features and not sentence_features and needs_sentence_features:
+            if not sequence_features and needs_sentence_features:
```
dakshvar22, line 590:
```suggestion
            if sequence_features and needs_sentence_features:
```

dakshvar22: At least the warning suggests that this should be the condition 😅

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -775,19 +793,24 @@ def _collect_features(
...
             ),
         )
-        if self._uses_sequence_features_for_text() and all(
-            not feats[TEXT] for feats in features
+
+        # NOTE: This check is a last santiy check but it doesn't solve every problem.
+        # We need the list of featurizers for this component and need to raise an
+        # error if LexicalSyntacticFeaturizer is included on it's own - or together
+        # with other featurizers. The latter is not checked by the following. Otherwise,
+        # we'll run into problems (especially if entity recognition is enabled).
```
dakshvar22, line 805: Not a helpful comment. We should either write what problem we run into or remove this part of the comment.

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -775,19 +793,24 @@ def _collect_features(
...
+        if self._needs_sentence_features_for_input_text() and any(
+            all(f.type != FEATURE_TYPE_SENTENCE for f in feats[TEXT])
+            for feats in features
```
dakshvar22, line 808: I guess `any` is extra here?

rasa/nlu/selectors/response_selector.py
```diff
@@ -635,17 +635,17 @@ def load(
     def _uses_sequence_features_for_text(self) -> bool:
```
dakshvar22, line 635: Shouldn't this be `_uses_sequence_features_for_input_text`?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -461,26 +470,31 @@ def _tag_id_index_mapping_for(
 
         return tag_id_dict
 
-    def _uses_sequence_features_for_text(self) -> bool:
+    def _uses_sequence_features_for_input_text(self) -> bool:
```
dakshvar22, line 473: Please make it static

dakshvar22 (changes requested):

rasa/nlu/selectors/response_selector.py
ka-bu: ouch, good catch

rasa/nlu/classifiers/diet_classifier.py
ka-bu: this must stay this way so child classes can overwrite this and use e.g. self.component_config to make this decision

dakshvar22: Ahh okay.. yes that makes sense.

rasa/nlu/classifiers/diet_classifier.py
ka-bu: another ouch 🤦‍♀️  wanted to remove "not sentence_features" here....

JEM-Mosig (approved):
Looks good! Just some style issue about "uses" vs "needs".

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -461,26 +470,31 @@ def _tag_id_index_mapping_for(
...
         by simply summing over sequence+sentence features.
         """
-        return (
-            self.component_config[NUM_TRANSFORMER_LAYERS] > 0
-            or self.component_config[ENTITY_RECOGNITION]
-            or self.label_attribute is not None
-        )
+        return True
+
+    def _needs_sentence_features_for_input_text(self) -> bool:
```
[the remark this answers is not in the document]
ka-bu: because _uses is optional (i.e. we drop sequence features which we always assume to be there if the architecture won't use them - i.e. results won't differ if you changed those) and _needs means they need this to function properly and we have a problem if they're missing

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -775,19 +793,26 @@ def _collect_features(
...
                 None
-                if self._uses_sequence_features_for_text()
+                if self._uses_sequence_features_for_input_text()
                 else FEATURE_TYPE_SENTENCE
             ),
         )
-        if self._uses_sequence_features_for_text() and all(
-            not feats[TEXT] for feats in features
+
+        # NOTE: This check is a last santiy check but it doesn't solve every problem.
```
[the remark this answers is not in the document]
ka-bu: I don't know how to check this just from the configs / component attributes. Tried partial pipeline and config, but couldn't extract the info about the featurizers from there (it's all empty, not sure why).

Issue #9274, opened 2021-08-05
Adapt `TEDPolicy` to new `GraphComponent` interface


wochinge:
**Overview of the Solution**:
Adapt the `TEDPolicy` to the new `GraphComponent` interface.
See the [implementation proposal](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e) for further details.

* adapt `create` and `load` functions to new signatures
* implement `required_packages` and `supported_languages` 
* adapt persistence
* add default param for `interpreter` so graph doesn't need to fill it (this is a workaround)
* adapt how config is passed into the component (and default config)

**Definition of Done**:
- [ ] TEDPolicy adapted to the new GraphComponent interface 
- [ ] Unit tests are adapted to adapted of policy
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old policy implementation)

PR #9306, opened 2021-08-09
3.0 architecture revamp/9274/adapt ted policy

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9274
- adapt `Policy` interface to `GraphComponent` interface
  - adapt persistence to use `ModelStorage` and `Resource`s
  - add required parameters to `create`, `load` and the constructor
  - move featurizer instantiation within policy (was done in `PolicyEnsemble` before)
- adapt `TEDPolicy` to `GraphComponent interface
  - adapt persistence to use `ModelStorage` and `Resource`s
  - add required parameters to `create`, `load` and the constructor
- adapt `PolicyTestCollection` to change `Policy` interface
- adapt all `TEDPolicy` unit tests to work with new policy
- get rid of config persistence as well as interpolating of default config as this done as part of the `GraphNode`

**Things you shouldn't have a look at** ⛔ 
- please don't have a look at the modules prefixed with `_` as they only contain the old but duplicated code


rasa/core/policies/policy.py
```diff
@@ -272,43 +295,6 @@ def _featurize_for_prediction(
             == SupportedData.ML_DATA,
         )
 
-    def train(
```
wochinge, line 275: I deleted them since there is no rule what a graph component receives as parameter during training or running (this only depends on the `needs` section)

rasa/core/policies/policy.py
```diff
@@ -114,17 +121,37 @@ def _create_featurizer(
...
-        **kwargs: Any,
     ) -> None:
         """Constructs a new Policy object."""
+        if featurizer is None:
+            featurizer = _featurizer_from_policy_config(config)
         self.__featurizer = self._create_featurizer(featurizer)
-        self.priority = priority
-        self.finetune_mode = should_finetune
+
+        self.priority = config.get("priority", DEFAULT_POLICY_PRIORITY)
```
wochinge, line 135: use constants for this

rasa/core/policies/policy.py
```diff
@@ -114,17 +121,37 @@ def _create_featurizer(
 
     def __init__(
         self,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
         featurizer: Optional[TrackerFeaturizer] = None,
```
wochinge, line 128: I kept the featurizer separate as its type changes (in the config it's a `Dict` and then we later instantiate it)

rasa/core/policies/policy.py
```diff
@@ -354,49 +344,55 @@ def _metadata_filename(cls) -> Text:
...
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+        **kwargs: Any,
+    ) -> "Policy2":
+        """Loads a trained policy (see parent class for full docstring)."""
+        config = {}
+        featurizer = None
+
+        with model_storage.read_from(resource) as path:
```
wochinge, line 372: todo: handle exception

rasa/core/policies/ted_policy.py
```diff
@@ -654,17 +661,15 @@ def train(
         self,
         training_trackers: List[TrackerWithCachedStates],
         domain: Domain,
-        interpreter: NaturalLanguageInterpreter,
-        **kwargs: Any,
-    ) -> None:
+        interpreter: NaturalLanguageInterpreter = RegexInterpreter(),
```
wochinge, line 664: add code comment

rasa/core/policies/ted_policy.py
```diff
@@ -970,78 +979,90 @@ def _load_model_utilities(cls, model_path: Path) -> Dict[Text, Any]:
...
-        **kwargs: Any,
-    ) -> "TEDPolicy":
-        """Loads a policy from the storage.
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> TEDPolicy2:
+        """Loads a policy from the storage (see parent class for full docstring)."""
+        try:
```
wochinge, line 988: how do we feel about indentation levels?

rasa/core/policies/ted_policy.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
wochinge, line 1: 💡 this allows use to get rid of our quoted [forward references](https://www.python.org/dev/peps/pep-0484/#forward-references). E.g. consider the following:

```python
class Foo:
   def __init__(self) -> "Foo":  # the quoted thing is a forward reference as we can't resolve `Foo` yet
     pass
```

becomes 

```python
from __future__ import annotations

class Foo:
   def __init__(self) -> Foo:
     pass
```

rasa/core/policies/ted_policy.py
```diff
@@ -156,7 +164,7 @@ class TEDPolicy(Policy):
     """
 
     # please make sure to update the docs when changing a default parameter
-    defaults = {
+    default_config = {
```
wochinge, line 167: 💡 all `GraphComponents` have a `default_config`.

rasa/core/policies/policy.py
wochinge: re-add

rasa/core/policies/policy.py
```diff
@@ -354,49 +344,55 @@ def _metadata_filename(cls) -> Text:
...
-
-        Args:
-            path: Path to persist policy to.
-        """
-        # not all policies have a featurizer
-        if self.featurizer is not None:
-            self.featurizer.persist(path)
+    def persist(self) -> None:
+        """Persists the policy to storage."""
+        with self._model_storage.write_to(self._resource) as path:
```
wochinge, line 349: 💡 `GraphComponent`s can persist themselves using their `resource` locator. The `model_storage` will provide a path for them which they can write to. See `load` if you want to see how to read from a `model_storage`

rasa/core/policies/policy.py
```diff
@@ -354,49 +344,55 @@ def _metadata_filename(cls) -> Text:
...
-        logger.info(
-            f"Couldn't load metadata for policy '{cls.__name__}'. "
-            f"File '{metadata_file}' doesn't exist."
-        )
-        return cls()
+    def load(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
```
wochinge, line 364: 💡 When we instantiate the `GraphComponent` for inference we hand them a `resource` which they can use to retrieve their trained version from the `model_storage`

rasa/core/policies/policy.py
wochinge, line 128: 💡 I've decided to keep the `TrackerFeaturizer` separate as it's type changes: in the config it's a dictionary with config parameters - here it's an instantiated Python object. Keeping it out of the config also avoids that we persist the featurizer twice (as part of the config and in the `persist` method).

rasa/core/policies/ted_policy.py
```diff
@@ -688,16 +693,20 @@ def train(
                 f"file to avoid this warning.",
                 category=UserWarning,
             )
-            return
+            return self._resource
```
wochinge, line 696: 💡 We return the `resource` after the training so that potential children nodes can receive the resource and then load the policy from the `model_storage` (we don't make use of this for policies but [we e.g. do that for featurization in the "nlu part" of the graph](https://www.notion.so/rasa/Rasa-3-0-Architecture-Scoping-Document-7dfb4454b436459f9eee0ec70d07aec7#5fdfa64cfa79436c8c5a1d06150b7565)).

tests/core/policies/test_ted_policy.py
```diff
@@ -12,8 +12,12 @@
     MaxHistoryTrackerFeaturizer,
```
wochinge, line 12: 💡 Please don't be intimidated by the amount of changes in the test file. This was only some extra work we had to do for the first policy due to our `PolicyTestCollection`. For your policy you will only have to migrate the `test_xy_policy` part.

tests/core/policies/test_ted_policy.py
```diff
@@ -712,14 +840,23 @@ class TestTEDPolicyWithRelativeAttentionMaxHistoryOne(TestTEDPolicy):
...
     def create_policy(
-        self, featurizer: Optional[TrackerFeaturizer], priority: int
+        self,
+        featurizer: Optional[TrackerFeaturizer],
+        priority: int,
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
     ) -> Policy:
+        # TODO: seems like a duplicate of the test above??
```
wochinge, line 850: @dakshvar22 Is this duplicate on purpose?

dakshvar22: `max_history` is defined as `1` for this class whereas the above one will use the default `max_history=3` defined in `PolicyTestCollection`.

wochinge: oh dear. thanks! 🙈

rasa/core/policies/policy.py
ka-bu: 👍  I think that also avoids the problem that the config doesn't contain the domain (so if we wanted to use the config to re-instantiate the `TrackerFeaturizer` instead of loading it from disk and passing it here as a `TrackerFeaturizer`, then we'd need to add the domain to the config before persisting and adapt the `TrackerFeaturizers` constructor to handle the domain - but we could do that later when we disentangle :) )

ka-bu: Wait, do we even need to pass it here? If we're using the constructor during training, this is None. If we use it during inference, then the featurizer is accessible via the model_storage, no?

rasa/core/policies/_ted_policy.py
```diff
@@ -0,0 +1,1094 @@
+# WARNING: This module will be dropped before Rasa Open Source 3.0 is released.
+#          Please don't do any changes in this module and rather adapt `TEDPolicy2` from
+#          the regular `rasa.core.policies.ted_policy` module. This module is a
+#          workaround to defer breaking changes due to the architecture revamp in 3.0.
+# flake8: noqa
```
wochinge, line 5: 💡 Our time is too valuable to fix docstring errors in modules which we will drop anyway

rasa/core/policies/_ted_policy.py
```diff
@@ -0,0 +1,1094 @@
+# WARNING: This module will be dropped before Rasa Open Source 3.0 is released.
```
wochinge, line 1: 💡 I added a warning to avoid that anything does changes in this module.

rasa/core/policies/policy.py
wochinge: @ka-bu The policies would use their `load`  method to instantiate themselves (and their featurizers) from disk. The actual constructor should then no longer have to read anything from disk.

Not sure I'm understanding your previous comment though 👀 The `Domain` would be a separate input to `train` or `predict_next_action_probabilities` - fed by a parent graph node.

Silverquillgrp:
Please discontinue these emails!!!


Lorraine (Lou) Y. Pawlivsky-Love
***@***.***
tel 778 350 4949
http://www.silverquill.org

wochinge:
@Silverquillgrp Sorry, I've no clue why GitHub subscribed you this PR. Commenting probably made it even worse 🙈 You can unsubscribe with the button shown in the screenshot
Screenshot 2021-08-10 at 16 39 22

Another reason why you might get notifications for this repository is maybe that you clicked on `watch` (if you scroll all they way up then there is a button on the top right corner)

rasa/core/policies/policy.py
ka-bu: @wochinge In that first comment I was just trying to go through what we would need to do in case we wanted to put everything in that that config (i.e. not the `TrackerFeaturizer` object in the config but extending the config by a dictionary that can be used to instantiate the `TrackerFeaturizer`). And I guess that idea doesn't make much sense at the moment because we don't have anything to configure there now 🤔  . And I think before asking more questions I should have a look at how/where custom TrackerFeaturizers would be instantiated (I was looking no further as `_standard_featurizer` so far), I will do that tomorrow.
(And in the second comment I was just confusing this here with the approach of loading and then passing e.g. the internal `model` to the constructor again - forget that :)) 
Thanks!

rasa/core/policies/ted_policy.py
ka-bu: Would it make sense to switch to something like Hydra for the configs in the long term or is there any reason we cannot do that? (would give us lots of type and consistency checks for free and should make hierarchical configs easier to manage)

Silverquillgrp:
Thanks...I was under the impression that I was participating in the Apple letter about privacy concerns....cheers, Lorraine

rasa/core/policies/ted_policy.py
```diff
@@ -688,16 +688,20 @@ def train(
...
 
         self.run_training(model_data, label_ids)
 
+        self.persist()
```
wochinge, line 695: 💡 We always persist as part of the training

wochinge:
@aeshky @ka-bu @joejuzl Could you please give this PR a code review so that I can merge it? 🙌🏻

rasa/core/featurizers/tracker_featurizers.py
```diff
@@ -509,7 +509,7 @@ def persist(self, path: Union[Text, Path]) -> None:
         )
 
     @staticmethod
-    def load(path: Text) -> Optional["TrackerFeaturizer"]:
+    def load(path: Union[Text, Path]) -> Optional["TrackerFeaturizer"]:
```
joejuzl, line 512: I really wish we just stuck to one throughout the code base 😂

rasa/core/policies/policy.py
```diff
@@ -354,49 +379,50 @@ def _metadata_filename(cls) -> Text:
...
-        if self.featurizer is not None:
-            self.featurizer.persist(path)
+            file = Path(path) / self._metadata_filename()
 
-        file = Path(path) / self._metadata_filename()
-
-        rasa.shared.utils.io.create_directory_for_file(file)
-        rasa.shared.utils.io.dump_obj_as_json_to_file(file, self._metadata())
+            rasa.shared.utils.io.create_directory_for_file(file)
+            rasa.shared.utils.io.dump_obj_as_json_to_file(file, self._metadata())
```
joejuzl, line 392: Did we find any reason why metadata should be treated differently from config?

rasa/core/policies/policy.py
```diff
@@ -602,3 +628,71 @@ def confidence_scores_for(
     results[idx] = value
 
     return results
+
+
+class InvalidPolicyConfig(RasaException):
+    """Exception that can be raised when policy config is not valid."""
+
+
+# TODO: Test the shit out of this
```
joejuzl, line 637: haha - will this be part of this PR?

rasa/core/policies/ted_policy.py
```diff
@@ -319,36 +327,35 @@ def _standard_featurizer(max_history: Optional[int] = None) -> TrackerFeaturizer
...
+        fake_features: Optional[Dict[Text, List[Features]]] = None,
         entity_tag_specs: Optional[List[EntityTagSpec]] = None,
-        should_finetune: bool = False,
-        **kwargs: Any,
     ) -> None:
         """Declares instance variables with default values."""
         self.split_entities_config = rasa.utils.train_utils.init_split_entities(
-            kwargs.get(SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE),
-            self.defaults.get(
+            config.get(SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE),
```
joejuzl, line 341: Don't we already handle using default config in `GraphNode`?

rasa/core/policies/ted_policy.py
```diff
@@ -319,36 +327,35 @@ def _standard_featurizer(max_history: Optional[int] = None) -> TrackerFeaturizer
...
 
-        # TODO: check if the else statement can be removed.
+        super().__init__(
+            config, model_storage, resource, execution_context, featurizer=featurizer
+        )
+
+        # TODO: check if this statement can be removed.
         #  More context here -
         #  https://github.com/RasaHQ/rasa/issues/5786#issuecomment-840762751
-        if not featurizer:
```
joejuzl, line 342: This is now covered by `_featurizer_from_policy_config` in `super().__init__`?

rasa/core/policies/ted_policy.py
```diff
@@ -379,10 +386,10 @@ def model_class() -> Type["TED"]:
     def _metadata_filename(cls) -> Optional[Text]:
         return "ted_policy"
 
-    def _load_params(self, **kwargs: Dict[Text, Any]) -> None:
-        new_config = rasa.utils.train_utils.check_core_deprecated_options(kwargs)
+    def _load_params(self, config: Dict[Text, Any]) -> None:
+        new_config = rasa.utils.train_utils.check_core_deprecated_options(config)
         self.config = rasa.utils.train_utils.override_defaults(
-            self.defaults, new_config
+            self.default_config, new_config
```
joejuzl, line 392: Again, isn't this covered in `GraphNode`?

rasa/core/policies/ted_policy.py
joejuzl: Can we test for this?

joejuzl (changes requested):

rasa/core/policies/policy.py
wochinge: yeah, unfortunately. Should've been tested before in my opinion 😬

rasa/core/featurizers/tracker_featurizers.py
wochinge: once we've done the revamp we can prioritize using `Path` everywhere 😆

rasa/core/policies/ted_policy.py
wochinge: that's tested by the `PolicyTestCollection`. There are several tests which train and and then load from the persisted thing.

rasa/core/policies/policy.py
wochinge: yeah, some policies use this to store additional stuff, e.g. the [`RulePolicy` ](https://github.com/RasaHQ/rasa/blob/f78a6a17d5146248500630a2397ee099a24e0f5c/rasa/core/policies/rule_policy.py#L1212) or [MemoizationPolicy](https://github.com/RasaHQ/rasa/blob/107caacd1a25fafac4232e9916875eb9c66bbf9f/rasa/core/policies/memoization.py#L245). 

There is definitely a huge overlap with the config. We should consider this when migrating these policies (doesn't matter for `TEDPolicy`)

rasa/core/policies/ted_policy.py
wochinge: done

rasa/core/policies/ted_policy.py
wochinge: done

rasa/core/policies/policy.py
wochinge: done

rasa/core/policies/ted_policy.py
wochinge: yep because I move the `super()` part before this part.

rasa/core/policies/ted_policy.py
```diff
@@ -112,13 +117,17 @@
...
 
-if TYPE_CHECKING:
-    from rasa.shared.nlu.training_data.features import Features
+from rasa.shared.nlu.training_data.features import Features
 
 
 logger = logging.getLogger(__name__)
 
+# This is a workaround around until we have all components migrated to `GraphComponent`.
+# (see the top of the file for more information).
```
joejuzl, line 128: There is no information at top of this file.

rasa/core/policies/_ted_policy.py
```diff
@@ -0,0 +1,1112 @@
...
+#          Please don't do any changes in this module and rather adapt `TEDPolicy2` from
```
joejuzl, line 2:
```suggestion
#          Please don't do any changes in this module and rather adapt `TEDPolicyGraphComponent` from
```

joejuzl (changes requested):

rasa/core/policies/ted_policy.py
wochinge: 🤦🏻

rasa/core/policies/policy.py
joejuzl: Yep we can probably split it between config and normal persisted data.

rasa/core/policies/policy.py
```diff
@@ -272,40 +309,30 @@ def _featurize_for_prediction(
...
-    def train(
-        self,
-        training_trackers: List[TrackerWithCachedStates],
-        domain: Domain,
-        interpreter: NaturalLanguageInterpreter,
-        **kwargs: Any,
-    ) -> None:
-        """Trains the policy on given training trackers.
+    @abc.abstractmethod
+    def train(self, **kwargs: Any,) -> Resource:
```
joejuzl, line 313: I've just thought don't all policies take `domain` and `training_trackers`?
If so can this be:

```
def train(
     self,
     training_trackers: List[TrackerWithCachedStates],
     domain: Domain,
     **kwargs: Any,
) -> Resource
```

tests/core/policies/test_ted_policy.py
```diff
@@ -74,14 +80,21 @@ def get_checkpoint_dir_path(train_path: Path, ted_pos: Optional[int] = 0) -> Pat
...
-def test_diagnostics():
+def test_diagnostics(
+    default_model_storage: ModelStorage, default_execution_context: ExecutionContext
+):
     domain = Domain.from_yaml(DOMAIN_YAML)
-    policy = TEDPolicy()
+    policy = TEDPolicy(
+        TEDPolicy.default_config.copy(),
+        default_model_storage,
+        Resource(""),
```
joejuzl, line 95:
```suggestion
        Resource("TEDPolicy"),
```

I know it doesn't make a huge difference, but just to make it slightly less likely something weird happens down the line.

joejuzl (approved):
Nice work 👍

rasa/core/policies/policy.py
wochinge: they do now, but maybe one in the future doesn't. I'd keep it fully flexible for now.

ka-bu (changes requested):
looks good to me :) just some small requests and questions

rasa/constants.py
```diff
@@ -38,3 +38,5 @@
 ENV_GPU_CONFIG = "TF_GPU_MEMORY_ALLOC"
 ENV_CPU_INTER_OP_CONFIG = "TF_INTER_OP_PARALLELISM_THREADS"
 ENV_CPU_INTRA_OP_CONFIG = "TF_INTRA_OP_PARALLELISM_THREADS"
+
+EPOCH_OVERRIDE = "epoch_override"
```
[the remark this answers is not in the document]
wochinge: mhm. I didn't see it as specifically related to `TensorFlow`. Any component which has epochs might use it, right? As this is a general flag for finetuning I think it should also be a general constant. The tensorflow constants seem to specific for me in this case.

rasa/core/policies/policy.py
```diff
@@ -114,17 +131,37 @@ def _create_featurizer(
...
+        resource: Resource,
+        execution_context: ExecutionContext,
         featurizer: Optional[TrackerFeaturizer] = None,
-        priority: int = DEFAULT_POLICY_PRIORITY,
-        should_finetune: bool = False,
-        **kwargs: Any,
     ) -> None:
         """Constructs a new Policy object."""
+        if featurizer is None:
+            featurizer = _featurizer_from_policy_config(config)
```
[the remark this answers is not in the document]
wochinge: I'll have a look. To be honest, I've just copied this over from `ensemble.py`

rasa/core/policies/policy.py
```diff
@@ -602,3 +630,77 @@ def confidence_scores_for(
...
+    return featurizer_func, featurizer_config
+
+
+def _get_state_featurizer_from_dict(
+    featurizer_config: Dict[Text, Any]
+) -> Tuple[Callable[..., SingleStateFeaturizer], Dict[Text, Any]]:
+    from rasa.core import registry
+
+    # featurizer can have only 1 state featurizer
+    if len(featurizer_config["state_featurizer"]) > 1:
```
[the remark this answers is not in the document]
wochinge: It's an iterable `Dict`. Don't ask me why it's a list of configs when we only allow 1 🤦🏻 That's how the current config format works. I think changing that is out of the scope of this adaption.

rasa/core/policies/policy.py
```diff
@@ -602,3 +630,77 @@ def confidence_scores_for(
...
+            ) = _get_state_featurizer_from_dict(featurizer_config)
+
+            # override featurizer's state_featurizer
+            # with real state_featurizer class
+            featurizer_config["state_featurizer"] = state_featurizer_func(
+                **state_featurizer_config
+            )
+        else:
+            # Removes empty state featurizer things from the config
+            featurizer_config.pop("state_featurizer", None)
```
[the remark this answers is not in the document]
wochinge: done

rasa/core/policies/policy.py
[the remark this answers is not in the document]
wochinge: done

rasa/core/policies/policy.py
wochinge: Yeah, it gives us the flexibility to add extra data but also removes the flexibility to remove parameters.
E.g. we once had a `MappingPolicy` which only required the domain https://github.com/RasaHQ/rasa/blob/2.8.x/rasa/core/policies/mapping_policy.py

rasa/constants.py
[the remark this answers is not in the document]
ka-bu: > Any component which has epochs might use it, right?
I like that idea but we only use it for the tf models (ehm, and only it TED it seems 😅 ) (and although I like that idea, unifying the settings might also not be ideal because sklearn users would expect that to be called "iter")

rasa/core/policies/policy.py
```diff
@@ -354,49 +381,50 @@ def _metadata_filename(cls) -> Text:
...
 
-            data.update(kwargs)
+                config.update(kwargs)
 
-            return cls(**data)
+        except (ValueError, FileNotFoundError, FileIOException):
+            logger.info(
+                f"Couldn't load metadata for policy '{cls.__name__}' as the persisted "
+                f"metadata couldn't be loaded."
+            )
```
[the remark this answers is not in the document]
wochinge: I just did what we did before as I didn't want to refactor but merely adapt the policies 🤷🏻‍♂️ I think someone probably put `info` in case this happens because the policy didn't receive the necessary training data to train itself. In that case `error`  during prediction time could be misleading

rasa/core/policies/policy.py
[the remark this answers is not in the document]
ka-bu: I know it's not really part of this PR - feel free to ignore 😬

rasa/core/policies/ted_policy.py
```diff
@@ -319,36 +328,32 @@ def _standard_featurizer(max_history: Optional[int] = None) -> TrackerFeaturizer
...
+        super().__init__(
+            config, model_storage, resource, execution_context, featurizer=featurizer
+        )
+
         self.split_entities_config = rasa.utils.train_utils.init_split_entities(
-            kwargs.get(SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE),
-            self.defaults.get(
-                SPLIT_ENTITIES_BY_COMMA, SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE
-            ),
+            config[SPLIT_ENTITIES_BY_COMMA], SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE
```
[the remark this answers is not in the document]
wochinge: `_load_params` definitely overrides deprecated options etc. I think further refactoring is somewhat out of the scope of the adaption

rasa/core/policies/ted_policy.py
wochinge: I think you're absolutely right 👍🏻 Due to the current state of our featurizers I think it makes sense to keep it like this _for now_ though because
- We currently only have one `Resource` per node (related to your question). As soon as featurizers would become their own graph nodes they'd also get their own `Resource` and we can then adapt them as required. Currently they'd end up writing to the same directory anyway.
- For the moment there is no need to expose this information to the featurizers. I agree that we should do this mid/long term but as long as they are embedded within policies we in my opinion don't need to expose the outer structure to them which also
- keeps the diff and changes smaller. In my opinion we should reach a stable system with the minimum of required changes first and then decide which bells and whistles we add. 

Does that make sense? What do you think?

rasa/constants.py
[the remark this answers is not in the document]
wochinge: Mhm 🤔 But even if sklearn calls it differently we wouldn't expose this implementation detail to our users / outer layers, right? Will move it over though 👍🏻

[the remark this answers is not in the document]
wochinge: actually I think you're right. good idea to move it 👍🏻

rasa/core/policies/policy.py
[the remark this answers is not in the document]
wochinge: already did it (this change was small enough 😆 )

rasa/core/policies/policy.py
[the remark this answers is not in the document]
ka-bu: Oh ok, interesting - Yeah there are lots of cases where training is silently skipped :/

rasa/core/policies/policy.py
```diff
@@ -139,8 +126,8 @@ def __init__(
     ) -> None:
         """Constructs a new Policy object."""
         if featurizer is None:
-            featurizer = _featurizer_from_policy_config(config)
-        self.__featurizer = self._create_featurizer(featurizer)
+            featurizer = self._create_featurizer(config)
+        self.__featurizer = featurizer
```
ka-bu, line 130:
```suggestion
            self.__featurizer = self._create_featurizer(config)
        else:
            self.__featurizer = copy.deep_copy(featurizer)
```

ka-bu (changes requested):
just one last tiny thing

rasa/core/policies/policy.py
wochinge: Do we actually need the `copy.deepcopy`? It seemed like an unused relict to me as we are not sharing the featurizers between policies anyway 🤷🏻

ka-bu: I have no clue 😅  I know you mentioned that some of those are set by `ensemble.py` but I never looked into how they do that

rasa/core/policies/policy.py
ka-bu: I think if we ever change the parameters drastically (or re-introduce policies that really need no training), then we should to think about new interfaces anyway, don't we? (because e.g. `MappingPolicy` effectively didn't need any training at all and would never appear in our graphs with a train step and would never require training data - it does still require the whole conversation / trackers for inference though)

rasa/core/policies/ted_policy.py
ka-bu: Yep makes sense. You're right, it's internals of TED and so TED has to make sure it uses that correctly and we never use it in such a dynamic way that adding automatic checks elsewhere makes much sense. Not sure that featurizers will ever become graph components (With the disentangling+sequence-generators/mini-batch work, it should even be more lightweight and more tightly coupled with TED).

rasa/core/policies/policy.py
ka-bu: @aeshky , what do you think?

wochinge: Added them to `train` and `predict_action_probabilities` 👍🏻

ka-bu (approved):
👍   Some final idea that can be ignored: To get rid of some indentation levels in `load` we could maybe create an outer `load` that does the try/catch for specific errors and opens the context and an inner `_load` that just receives `model_path`. But it might be annoying to agree what kind of exception the inner/outer thing can raise/has to catch.

rasa/core/policies/policy.py
wochinge: Haha, I'd dump it for now as the line was super old and no test fails.

PR #9317, opened 2021-08-10
Make it possible to use `or` with slot values

### LINKED ISSUE #8933: Make it possible to OR slot values - Implementation
**Description of Problem**:
Background: https://github.com/RasaHQ/rasa/issues/7715
Proposed Solution: https://www.notion.so/rasa/Make-it-possible-to-OR-slot-values-4abdf1b6632c4bb2881d2057b58ca335

Summary:
We want to be able to use `or` for `slot_was_set` steps in stories.
It currently only works for `intent` steps.

**Todo**:
- updating the story schema `rasa/shared/utils/schemas/stories.yml`
- updating `YAMLStoryReader`
    - `_parse_or_statement`
- updating `StoryStepBuilder`
    - adding a `slot_was_set` equivalent to `add_user_messages`
- make sure `rasa.shared.core.training_data.visualization` is updated for Rasa X story visualisation

**Example Story**:
```
- story: tell name bob or joe
  steps:
  - intent: greet
  - action: utter_greet
  - intent: tell_name
  - or:
    - slot_was_set:
        - name: joe
    - slot_was_set:
        - name: bob
  - action: utter_generic_hi
```

**Definition of Done**:
- [ ] `or` functionality added for `slot_was_set` events
- [ ] Rasa X visualisation compatibility  
- [ ] Tests are added
- [ ] Feature described the docs
- [ ] Feature mentioned in the changelog

alwx:
**Proposed changes**:
- fix #8933


rasa/shared/core/training_data/story_reader/story_reader.py
```diff
@@ -113,6 +115,14 @@ def _add_event(self, event_name: Text, parameters: Dict[Text, Any]) -> None:
                 "Unknown event '{}'. It is Neither an event "
                 "nor an action).".format(event_name)
             )
+
+        return parsed_events
+
+    def _add_event(self, event_name: Text, parameters: Dict[Text, Any]) -> None:
+        # add 'name' only if event is not a SlotSet,
+        # because there might be a slot with slot_key='name'
```
ancalita, line 123: I think these comments can be removed now that you've extracted the code into `_parse_events()`

rasa/shared/core/training_data/story_reader/story_step_builder.py
```diff
@@ -74,31 +74,47 @@ def add_user_messages(
               Default value is `False`, which preserves the expected behavior
               of the reader.
         """
+        self.add_events(messages, is_used_for_training)
+
+    def add_events(
```
ancalita, line 79: Some of the comments preserved here might need to accommodate slot events, at the moment they only mention intents.

rasa/shared/core/training_data/structures.py
```diff
@@ -169,11 +170,16 @@ def _bot_string(story_step_element: Event) -> Text:
         return f"    - {story_step_element.as_story_string()}\n"
 
     @staticmethod
-    def _or_string(story_step_element: Sequence[Event], e2e: bool) -> Text:
+    def _or_string(story_step_element: Sequence[Event], e2e: bool) -> Optional[Text]:
         for event in story_step_element:
-            if not isinstance(event, UserUttered):
+            # OR statement can also contain `slot_was_set`, and
+            # we're going to ignore this events when representing
```
ancalita, line 176: Only trying to understand why is it better to ignore the `SlotSet` event?

rasa/shared/utils/schemas/stories.yml
```diff
@@ -103,6 +103,8 @@ mapping:
                 sequence:
                 - type: "map"
                   mapping: *intent_and_entities
+                - type: "map"
+                  mapping: *slot_was_set_seq
```
ancalita, line 107: What does the `*` do here? I've tried looking this up in pykwalify docs but couldn't find anything.

ancalita (commented):
Looks great 💯  I've only left a few nit-picking comments, nothing too major.
I was also wondering if the task regarding `rasa.shared.core.training_data.visualization` modification in the Definition of Done was set aside? or something you have to follow-up on?

rasa/shared/utils/schemas/stories.yml
alwx: This is basically how linking is done. You can check other mappings in this file (for example, `mapping: *intent_and_entities` just above this)

rasa/shared/core/training_data/story_reader/yaml_story_reader.py
```diff
@@ -338,26 +338,47 @@ def _validate_that_utterance_is_in_domain(self, utterance: UserUttered) -> None:
...
-                    utterances.append(utterance)
+                    events.append(utterance)
+            elif KEY_CHECKPOINT_SLOTS in item.keys():
+                if isinstance(item, dict):
+                    for key, value in item.items():
+                        parsed_events = self._parse_events(
+                            SlotSet.type_name, {key: value}
+                        )
+                        events.extend(parsed_events)
+                elif isinstance(item, str):
```
joejuzl, line 355: Is it possible for `item` to be a str if this is nested under `elif KEY_CHECKPOINT_SLOTS in item.keys()`?

rasa/shared/utils/schemas/stories.yml
joejuzl: for example:
```
block1: &myblock
  a: 1
  b: 2
block2:
  subblock: *myblock
```

is equivalent to:
```
block1:
  a: 1
  b: 2
block2:
  subblock:
    a: 1
    b: 2
```

tests/shared/core/training_data/story_reader/test_yaml_story_reader.py
```diff
@@ -552,6 +552,46 @@ def test_or_statement_if_not_training_mode():
...
+      - intent: tell_name
+      - or:
+        - slot_was_set:
+            - name: joe
+        - slot_was_set:
+            - name: bob
+      - action: utter_generic_hi
+    """
+
+    reader = YAMLStoryReader(is_used_for_training=False)
```
joejuzl, line 571: Can we have a test where ` is_used_for_training=True` so we can see the actual `or` unfolding? i.e how the output would look for actual training.

joejuzl (changes requested):
Thanks for this!
I think it would be really cool if we could test this through the `TrainingDataGenerator`.
So we can see that the actual trackers are created as we expect (`test_common_story_reader.py` is probably the most similar).

Also we need to make sure that story visualisation still works correctly (it's used by Rasa X).

rasa/shared/core/training_data/structures.py
alwx: fixed that, was a stupid assumption

tests/shared/core/training_data/story_reader/test_common_story_reader.py
```diff
@@ -297,3 +297,42 @@ async def test_yaml_slot_different_types(stories_file: Text, domain: Domain):
...
+        intent={INTENT_NAME_KEY: "simple", "confidence": 1.0},
+        parse_data={
+            "text": "/simple",
+            "intent_ranking": [{"confidence": 1.0, INTENT_NAME_KEY: "simple"}],
+            "intent": {"confidence": 1.0, INTENT_NAME_KEY: "simple"},
+            "entities": [],
+        },
+    )
+    assert tracker.events[2] == ActionExecuted("utter_default")
+    assert tracker.events[3] == SlotSet(key="name", value="joe")
```
ancalita, line 338: There's a missing newline at the end here.
Also, would index 4 in `tracker.events` also be a `SlotSet` event, the one with `value = bob`?

ancalita (commented):
Looking good 👍 I think the visualisation module also needs updating.

alwx:
Just checked it: `slot_was_set` doesn't affect the visualization at all — so nothing needs to be adapted in `rasa.shared.core.training_data.visualization`

tests/shared/core/training_data/story_reader/test_common_story_reader.py
alwx: That's a good assumption but no — however, I am not sure why. My assumption was that there should be two trackers and `trackers[0].events[3] == SlotSet(key="name", value="joe")` and `trackers[1].events[3] == SlotSet(key="name", value="bob")` but that doesn't happen.
I was thinking that's an issue but it doesn't happen with or expressions when the intents (not `SlotSet`s) are used, and it was like that before all these changes — so I now wonder if it's the bug or I just got that all wrong.
Maybe somebody can clarify? @joejuzl or @wochinge or maybe even someone from research

wochinge: Very good catch @ancalita ! It should be two trackers - one with "joe" and one with "bob". If you do the `or` with intents you also get two trackers. This might require some changes in the `TrackerGenerator`

joejuzl: This still needs to be handled right @alwx ?

joejuzl (changes requested):

tests/shared/core/training_data/story_reader/test_common_story_reader.py
alwx: Should be good to go now — I rebased the PR: there is no `common_story_reader.py` anymore, and the test to check if story steps get copied or not has already been added to `test_yaml_story_reader.py` (e.g. `test_or_statement_with_slot_was_set`).

wochinge: I think the rebasing just deleted this test. We still need a test which asserts that the generator creates two trackers out of the story with the `or` slot. So you basically need a test which asserts that `training.load_data` causes 2 trackers

alwx: Finally added a test for this! (`test_or_statement_story_with_or_slot_was_set`)

tests/core/test_tracker_stores.py
```diff
@@ -398,8 +398,8 @@ def test_db_url_with_query_from_endpoint_config(tmp_path: Path):
         driver: my-driver
         another: query
     """
-    f = tmp_path / "tmp_config_file.yml"
-    f.write_text(endpoint_config)
+    stories_path = tmpdir / "stories.yml"
```
ancalita, line 401: Not sure why these modifications were needed? In addition to the code quality failings, the test also fails.

tests/shared/core/training_data/test_structures.py
```diff
@@ -87,6 +87,38 @@ def test_as_story_string_or_statement():
...
+
+    steps = reader.read_from_parsed_yaml(yaml_content)
+
+    assert len(steps) == 1
+
+    assert (
+        steps[0].as_story_string()
+        == """
+## hello world
+* slot{"slot_was_set": [{"name": "joe"}]} OR slot{"slot_was_set": [{"name": "bob"}]}
```
ancalita, line 116: It seems that actually the CI expects this:
`slot{"name": "joe"} OR slot{"name": "bob"}` 🤔

alwx: Fixed this

tests/core/test_tracker_stores.py
alwx: I guess it was some post-merge stuff, sorry 🙈
Removed it!

tests/shared/core/training_data/test_structures.py
```diff
@@ -87,6 +87,38 @@ def test_as_story_string_or_statement():
...
+      steps:
+      - or:
+        - slot_was_set:
+            - name: joe
+        - slot_was_set:
+            - name: bob
+      - action: some_action
+    """
+
+    reader = YAMLStoryReader(is_used_for_training=False)
```
ancalita, line 105: Do we need a test here when the `is_used_for_training` flag is `True` too?

alwx: It's tested in another test (`test_or_statement_with_slot_was_set_is_used_for_training`) — in this case it makes no sense 'cause it will break the logic.

ancalita: What do you mean by it will break the logic? You mean the method `as_story_string()` can't be called for each of the 2 steps?

alwx: Because when you specify that `is_used_for_training=True` then checkpoints are getting added and trackers are getting copied (see `add_stories` method in `StoryStepBuilder`, line 110). As a result, `reader.read_from_parsed_yaml(yaml_content)` will create multiple trackers and calling `.as_story_string` on them will lead to a result like that which is meaningless:

```
## hello world
    - slot{"name": "joe"}
> GENR_OR_b2ae7
```

However, this result is important when these trackers are used for training, and that case is already tested in `test_or_statement_with_slot_was_set_is_used_for_training`

ancalita: Great, thanks for the deep-dive!

ancalita (approved):
LGTM 🙌🏼

rasa/shared/core/training_data/story_reader/story_step_builder.py
```diff
@@ -74,31 +74,47 @@ def add_user_messages(
...
+        self, events: List[Event], is_used_for_training: bool = True
+    ) -> None:
+        """Adds next story steps with the specified list of events.
+
+        Args:
+            events: Events that need to be added.
+            is_used_for_training: Identifies if it's a part of OR statement.
```
joejuzl, line 86: Do we need this anymore if we no longer support MD?

rasa/shared/core/training_data/story_reader/story_step_builder.py
```diff
@@ -74,31 +74,47 @@ def add_user_messages(
...
-                    t.add_events(messages)
+                    t.add_events(events)
                 return
 
-            # If there are multiple different intents the
+            # If there are multiple different events the
             # user can use the express the same thing
             # we need to copy the blocks and create one
             # copy for each possible message
             generated_checkpoint = self._generate_checkpoint_name_for_or_statement(
```
joejuzl, line 110: Does this method work properly with non `UserUttered` events?

rasa/shared/core/training_data/story_reader/story_step_builder.py
```diff
@@ -143,9 +159,7 @@ def _next_story_steps(self) -> List[StoryStep]:
         ]
         return current_turns
 
-    def _generate_checkpoint_name_for_or_statement(
-        self, messages: List[UserUttered]
-    ) -> str:
+    def _generate_checkpoint_name_for_or_statement(self, messages: List[Event]) -> str:
```
joejuzl, line 162: This method refers to `messages_texts_or_intents` - does this work for all events?

rasa/shared/core/training_data/structures.py
joejuzl: Is this comment incorrect now then?

joejuzl (changes requested):

rasa/shared/core/training_data/story_reader/story_step_builder.py
alwx: The `__str__` method is definied for all the events so it does because it requires only that. However, the names should be changed.

tests/shared/core/training_data/story_reader/test_yaml_story_reader.py
```diff
@@ -647,27 +575,6 @@ def test_or_statement_story_with_or_slot_was_set(domain: Domain):
...
-    "file,warning",
-    [
-        ("data/test_yaml_stories/test_base_retrieval_intent_story.yml", None),
-        (
-            "data/test_yaml_stories/non_test_full_retrieval_intent_story.yml",
-            UserWarning,
-        ),
-    ],
-)
-async def test_story_with_retrieval_intent_warns(
```
joejuzl, line 574: Do we not expect this warning any more?

tests/shared/core/training_data/test_structures.py
```diff
@@ -56,37 +56,6 @@ def test_session_start_is_not_serialised(domain: Domain):
...
-
-    reader = YAMLStoryReader(is_used_for_training=False)
-    yaml_content = rasa.shared.utils.io.read_yaml(stories)
-
-    steps = reader.read_from_parsed_yaml(yaml_content)
-
-    assert len(steps) == 1
-
-    assert (
-        steps[0].as_story_string()
```
joejuzl, line 81: Is this no longer used?

tests/shared/core/training_data/test_structures.py
```diff
@@ -102,21 +71,12 @@ def test_as_story_string_or_statement_with_slot_was_set():
...
 
     steps = reader.read_from_parsed_yaml(yaml_content)
 
-    assert len(steps) == 1
-
-    assert (
-        steps[0].as_story_string()
-        == """
-## hello world
-* slot{"name": "joe"} OR slot{"name": "bob"}
```
joejuzl: ditto

joejuzl (changes requested):

tests/shared/core/training_data/story_reader/test_yaml_story_reader.py
alwx: Returned it back — in this case we can actually just remove the optional argument for `YAMLStoryReader` to test exactly the same warning.

tests/shared/core/training_data/test_structures.py
alwx: It isn't. `is_used_for_training=False` can no longer be used so the tracker(s) generated by `StoryReader` can only be used for training.

tests/shared/core/training_data/test_structures.py
alwx: Same as above.

tests/shared/core/training_data/test_structures.py
joejuzl: So can we delete `as_story_string()` as part of removing mark down support?

joejuzl (approved):
💯

tests/shared/core/training_data/test_structures.py
alwx: It still seems to be used by the interactive learning so I would skip that for now.

joejuzl: So should we keep the test then?

PR #9339, opened 2021-08-12
Graph Loader

joejuzl:
Implement the graph loader.

### DIFF
--- a/tests/engine/test_loader.py
+++ b/tests/engine/test_loader.py
@@ -1,0 +1,96 @@
+from datetime import datetime
+from pathlib import Path
+
+from _pytest.tmpdir import TempPathFactory
+
+import rasa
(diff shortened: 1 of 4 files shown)


tests/engine/test_loader.py
```diff
@@ -0,0 +1,96 @@
...
+                fn="run_inference",
+                constructor_name="load",
+                config={},
+                is_target=True,
+                resource=Resource("train"),
+            ),
+        }
+    )
+
+    before_train_time = datetime.utcnow()
```
wochinge, line 67: how about using `freezegun`?

wochinge (approved):
Sweet 💯 🚀

tests/engine/test_loader.py
joejuzl: done

Issue #9276, opened 2021-08-05
Implement `TrackerGenerator`


wochinge:
**Overview of the Solution**:
Implement a new graph component called `TrackerGenerator`.
This component converts the training `StoryGraphs` provided by the `StoryGraphProvider` using the domain provided by `DomainProvider` to `TrackerWithCachedStates` (the parent node components do not have to be implemented).

See the [implementation proposal](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#8bc7b59d86184ab3899cde2cf44a403d) for further details.

**Definition of Done**:
- [ ] `TrackerGenerator` is implemented
- [ ] Component is unit tested

PR #9361, opened 2021-08-16
3.0 architecture revamp/9276/tracker generator

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9276


rasa/components/providers/training_tracker_provider.py
```diff
@@ -0,0 +1,53 @@
...
+class TrainingTrackerProvider(GraphComponent):
+    """Provides training trackers to policies based on training stories."""
+
+    def __init__(self, config: Dict[Text, Any]) -> None:
+        self._config = config
+
+    @staticmethod
+    def get_default_config() -> Dict[Text, Any]:
+        return {
+            "remove_duplicates": True,
```
ka-bu, line 21: I think these parameters should (only) be defined in `TrainingDataGenerator` and here we should just pass them on (via some `TrainingDataGenerator.get_default_config()`

And why don't we make `TrainingDataGenerator` a GraphComponent itself?  `generate_trackers` wouldn't need the instantiation of a `TrainingDataGenerator` then (and it doesn't need to be instantiated multiple times because configs don't change / cannot be specified in `generate_trackers` anyway.

tests/components/providers/test_training_tracker_provider.py
```diff
@@ -0,0 +1,55 @@
...
+)
+from rasa.shared.core.training_data.structures import StoryGraph
+
+
+@pytest.mark.parametrize(
+    "config, expected_trackers",
+    [
+        ({}, 507),
+        ({"augmentation_factor": 0}, 7),
+        ({"use_story_concatenation": False}, 7),
```
ka-bu, line 21: The `expected_trackers` output is really a test of the `TrainingDataGenerator` (and not so much of any functionality inside the TrainingTrackerProvider). I would expect these to be tests in `TrainingDataGenerator`. ... Not really easy to test what is done (only) here then. We could mock the TrainingDataGenerator and test that the constructor and generate is used 😅  . Maybe another reason to make `TrainingDataGenerator` a graph component. Wdyt?

ka-bu (commented):
I have some questions here.

rasa/components/providers/training_tracker_provider.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class TrainingTrackerProvider(GraphComponent):
+    """Provides training trackers to policies based on training stories."""
+
+    def __init__(self, config: Dict[Text, Any]) -> None:
+        self._config = config
+
+    @staticmethod
+    def get_default_config() -> Dict[Text, Any]:
```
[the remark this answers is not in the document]
wochinge: I think we should in general try to use less inheritance for code sharing and rather more composition. And as the class should know it's super class we can still do `{**ParentClass.get_default_config(), **{'additional_parameter' : ...}}`

rasa/components/providers/training_tracker_provider.py
wochinge: That's a good idea 👍🏻 Will do so 👍🏻

rasa/components/providers/training_tracker_provider.py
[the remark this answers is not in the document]
ka-bu: True, we can do that (I'm just super lazy 😄 ).  And we don't have to use the config concat for the algos now. I think that might become a nice task on its own combined with your idea of unifying the parameters (like the epoch override).

rasa/components/providers/training_tracker_provider.py
wochinge: @ka-bu Actually I now remember why I didn't do this. Doing this would mean to create a dependency to `rasa.engine` in the `rasa.shared` package which in turn would mean that we'd need to move the entire engine package to `rasa.shared` (as `rasa.shared` is not allowed to import anything outside of `rasa.shared` which I wanted to avoid. 

> I think these parameters should (only) be defined in TrainingDataGenerator

It's somewhat fine as we have the unit test to ensure these params are correct, right?

ka-bu: Ah ok, so we need the wrapper here. But I'd still move the parameters/tests. Because otherwise we're only half-way respecting the engine/shared border (and the wrapper just begin a wrapper), don't we? Because those parameters are internals of `TrainingDataGenerator` and if we change the internals there, we'd have to adapt the parameters here *and* the unit tests here as well. 
(And if we want tests here, we could mock.patch the TrainingDataGenerator in the tests then and just ensuring that creation and generate are called - cause that's all that this wrapper should do / we know "here")

wochinge: I think the tests are valid as they test whether we are actually wrapping the `TrainingDataGenerator` correctly. if we'd mock the tracker generator then we can't catch potentially breaking changes (such as if somebody renames the params). 

How would you move the params over due `TrainingDataGenerator`? I fail to see an elegant way to do so besides inspecting the method signatures which in my opinion goes to far.

ka-bu: My workaround would be to adapt `TrainingDataGenerator` so that it looks like a GraphComponent in some ways but we don't call it like that, i.e. has a `config` ,  instead of `self.<param>` do `self.config[<param>]` , adding properties to account for the missing `self.domain` for example ,.... and yeah that is a lot of changes. ... Agree that without that we should keep the tests. Then let's just keep it as is (?).

ka-bu (approved):

joejuzl (approved):

rasa/components/providers/training_tracker_provider.py
wochinge: I gave it another thought and decided to keep it as it is for now. I think you've very good points but I don't think it's the best use of our resources for the moment. Also I'd like to keep the changes to existing things minimal so we have less to look through in case some of the model regression tests or something else is giving weird results

PR #9369, opened 2021-08-17
3.0 architecture revamp/integration branch review comments kathrin

wochinge:
**Proposed changes**:
- return tuple from `GraphNode` instead of `{"node_name": value"}`
- improve docstring
- validate for cycles in the schema
- change `defaults` to `get_default_config` (from property to method)

### DIFF
--- a/rasa/engine/graph.py
+++ b/rasa/engine/graph.py
@@ -167,7 +167,8 @@ class GraphComponent(ABC):
         If not overridden this method merely calls `create`.
 
         Args:
-            config: This config overrides the `default_config`.
+            config: The config for this graph component. This is the default config of
+                the component merged with config specified by the user.
             model_storage: Storage which graph components can use to persist and load
                 themselves.
             resource: Resource locator for this component which can be used to persist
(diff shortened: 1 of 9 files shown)


rasa/engine/validation.py
```diff
@@ -82,6 +96,29 @@ def validate(
         )
 
 
+def _validate_cycle(schema: GraphSchema) -> None:
+    for target_name in schema.target_names:
```
wochinge, line 100: this would technically not catch cycles if there is an unused node in the graph 🤔 Wanted to avoid walking the graph to often (which would happen if we do this for every node)

rasa/engine/graph.py
```diff
@@ -361,24 +371,19 @@ def parent_node_names(self) -> List[Text]:
         """The names of the parent nodes of this node."""
         return list(self._inputs.values())
 
-    @staticmethod
```
wochinge, line 364: I think deleting this proves Kathrin's idea was great 💯 😄

joejuzl: Agreed 🔥

ka-bu (approved):
👍

Issue #5786, opened 2020-05-07
Max_history parameter should be carried over to the featurizer definition in TEDPolicy


akelad:
If you have your configuration defined like this, with the featurizer defined explicitly:

*   name: TEDPolicy
     max_history: 5
     featurizer:
*   name: MaxHistoryTrackerFeaturizer
     state_featurizer:
*   name: LabelTokenizerSingleStateFeaturizer

The `max_history` parameter won't get carried over to the `MaxHistoryTrackerFeaturizer`. This is kind of strange behaviour, because if you specify the policy like this:

*   name: TEDPolicy
     max_history: 5

The policy loads up `MaxHistoryTrackerFeaturizer` internally with `max_history: 5`.

See forum post for source of confusion: [https://forum.rasa.com/t/running-oom-in-tedpolicy-with-900-stories-1-10-0/28230/6](https://forum.rasa.com/t/running-oom-in-tedpolicy-with-900-stories-1-10-0/28230/6)

chkoss:
Is this rather an issue for @RasaHQ/production-squad or for research squad?

wochinge:
@Ghostvv Does it actually make sense to do this when we use Genie's featurizer for everything? If this should still be done, then I think we can handle this + I think it's even good if we can get our feet wet with easier issues like this one.

Ghostvv:
it still makes sense to do, we'll have `max_history` parameter anyhow

joaoCeilandia:
Me and some friends where working on this issue and wue found that this behavior happens because when creating a featurizer it is instantiated as a `TrackerFeaturizer`  and when no featurizer is instantiated, it is created as a `MaxHistoryTrackerFeaturizer`. this code shows this:
![image](https://user-images.githubusercontent.com/31783143/111869900-41e00c00-8960-11eb-8d41-24db99e3df83.png)

So the only way to make `max_history` be inside the featurizer is by don't instantiating one. 

we tought about make a change by our own to try to fix this but as we don't know what is wanted for this issue( e.g.  if we could make a non-instanciated featurizer just became a `TrackerFeaturizer` or make a intanciated one became a `MaxHistoryFeaturizer` ), we came ask for some directioning.

@akelad @Ghostvv @wochinge @chkoss

dakshvar22:
[...]

PR #9383, opened 2021-08-17
Adapted memoization policies for graph

### LINKED ISSUE #9344: Adapt `MemoizationPolicy` and `AugmentedMemoizationPolicy` to new `GraphComponent` interface
**Overview of the Solution**:
Adapt the `MemoizationPolicy` and `AugmentedMemoizationPolicy` to the new `GraphComponent` interface.

* adapt persistence
* add default param for `interpreter` so graph doesn't need to fill it (this is a workaround)
* adapt how config is passed into the component (and default config)

**Definition of Done**:
- [ ] `MemoizationPolicy` adapted to the new GraphComponent interface 
- [ ] `AugmentedMemoizationPolicy` adapted to the new GraphComponent interface 
- [ ] Unit tests are adapted to adaption of policy
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old policy implementation)

twerkmeister:
**Proposed changes**:
- fixes #9344 

important context: https://www.notion.so/rasa/Migrating-policies-to-the-new-model-architecture-819812d37d784e4d98347d74ea4844fc#369e558d39b74a5096081a0408c23665


rasa/core/policies/memoization.py
```diff
@@ -45,44 +65,42 @@ class MemoizationPolicy(Policy):
...
+        # please make sure to update the docs when changing a default parameter
+        return {
+            "enable_feature_string_compression": True,
+            "use_nlu_confidence_as_score": False,
+            POLICY_PRIORITY: MEMOIZATION_POLICY_PRIORITY,
+            POLICY_MAX_HISTORY: DEFAULT_MAX_HISTORY,
+        }
 
-    USE_NLU_CONFIDENCE_AS_SCORE = False
+    required_packages = []
```
wochinge, line 79: you actually don't need that need one (it's implemented as staticmethod as well but it has a default)
```suggestion
```

rasa/core/policies/memoization.py
```diff
@@ -45,44 +65,42 @@ class MemoizationPolicy(Policy):
...
-            featurizer: tracker featurizer
-            priority: the priority of the policy
-            max_history: maximum history to take into account when featurizing trackers
-            lookup: a dictionary that stores featurized tracker states and
-                predicted actions for them
-        """
+        """Initialize the policy."""
         if not featurizer:
-            featurizer = self._standard_featurizer(max_history)
+            featurizer = self._standard_featurizer()
```
wochinge, line 99: I actually made that part of the super class so that you don't need to take care of it 👍🏻

rasa/core/policies/memoization.py
```diff
@@ -45,44 +65,42 @@ class MemoizationPolicy(Policy):
...
 
-        super().__init__(featurizer, priority, **kwargs)
-
-        self.max_history = self.featurizer.max_history
-        self.lookup = lookup if lookup is not None else {}
+        super().__init__(config, model_storage, resource, execution_context, featurizer)
+        self.config = config
```
wochinge, line 102: should we actually store the config as part of the super class? 🤔

rasa/core/policies/memoization.py
```diff
@@ -239,20 +256,56 @@ def predict_action_probabilities(
         return self._prediction(result)
 
     def _metadata(self) -> Dict[Text, Any]:
-        return {
-            "priority": self.priority,
-            "max_history": self.max_history,
-            "lookup": self.lookup,
-        }
+        return {"config": self.config, "lookup": self.lookup}
```
wochinge, line 259: can we please drop the `config` here as we store it with the graph schema? The only thing we actually need to store is the lookup table

rasa/core/policies/memoization.py
```diff
@@ -45,44 +65,42 @@ class MemoizationPolicy(Policy):
...
-    ENABLE_FEATURE_STRING_COMPRESSION = True
+    @staticmethod
+    def get_default_config() -> Dict[Text, Any]:
+        """Returns the default config (see parent class for full docstring)."""
+        # please make sure to update the docs when changing a default parameter
+        return {
+            "enable_feature_string_compression": True,
+            "use_nlu_confidence_as_score": False,
+            POLICY_PRIORITY: MEMOIZATION_POLICY_PRIORITY,
+            POLICY_MAX_HISTORY: DEFAULT_MAX_HISTORY,
```
wochinge, line 76: I actually forgot that part for `TEDPolicy` 😬 Great catch 👍🏻

wochinge (commented):
Can we please also adapt the unit tests to the new GraphComponent versions of the policy? 🙌🏻

rasa/core/policies/memoization.py
twerkmeister: I think we should yeah

rasa/core/policies/memoization.py
twerkmeister: In the super class it looks like the config is loaded from the metadata file too

koernerfelicia: [here](https://github.com/RasaHQ/rasa/blob/fdb0413301dcd3021ae3742d0d91528dc94a9f8d/rasa/core/policies/policy.py#L444) ?

wochinge: Very good point. How about nuking the entire `load` implementation of the parent class and make it an abstract method so that subclasses have to implement it themselves?

rasa/core/policies/memoization.py
```diff
@@ -239,20 +250,56 @@ def predict_action_probabilities(
...
+        """Loads a trained policy (see parent class for full docstring)."""
+        config = {}
+        featurizer = None
+        lookup = None
+
+        try:
+            with model_storage.read_from(resource) as path:
+                metadata_file = Path(path) / cls._metadata_filename()
+                metadata = json.loads(rasa.shared.utils.io.read_file(metadata_file))
+                config = metadata["config"]
```
wochinge, line 277: let's please use passed in `config` instead of the persisted one

wochinge:
@twerkmeister Can we please also adapt the unit tests to the new GraphComponent versions of the policies? 🙌🏻

twerkmeister:
1. Adjusted the unit tests
2. Adjusted the load function to not load the config from disc, but didn't remove it entirely

twerkmeister:
should also fix #5786

rasa/core/policies/memoization.py
```diff
@@ -239,20 +249,54 @@ def predict_action_probabilities(
...
+        try:
+            with model_storage.read_from(resource) as path:
+                metadata_file = Path(path) / cls._metadata_filename()
+                metadata = json.loads(rasa.shared.utils.io.read_file(metadata_file))
+                lookup = metadata["lookup"]
+
+                if (Path(path) / FEATURIZER_FILE).is_file():
+                    featurizer = TrackerFeaturizer.load(path)
+
+                config.update(kwargs)
```
wochinge, line 280: why do we need this line?

rasa/core/policies/policy.py
```diff
@@ -123,6 +126,7 @@ def __init__(
         featurizer: Optional[TrackerFeaturizer] = None,
     ) -> None:
         """Constructs a new Policy object."""
+        self.config = config
```
wochinge, line 129: ?
```suggestion
        self._config = config
```

rasa/core/policies/policy.py
```diff
@@ -147,18 +151,25 @@ def create(
...
-            return cls._standard_featurizer()
+            return self._standard_featurizer()
+
+        if (
+            POLICY_MAX_HISTORY not in featurizer_configs[0]
+            and POLICY_MAX_HISTORY in policy_config
+        ):
+            featurizer_configs[0][POLICY_MAX_HISTORY] = policy_config[
+                POLICY_MAX_HISTORY
+            ]
```
wochinge, line 168: how about passing the `max_history` separately to `_get_featurizer_from_config`? This would avoid weird errors in case the `featurizer_configs` are not a list

rasa/core/policies/policy.py
```diff
@@ -147,18 +151,25 @@ def create(
...
 
         if not featurizer_configs:
-            return cls._standard_featurizer()
+            return self._standard_featurizer()
+
+        if (
+            POLICY_MAX_HISTORY not in featurizer_configs[0]
+            and POLICY_MAX_HISTORY in policy_config
+        ):
+            featurizer_configs[0][POLICY_MAX_HISTORY] = policy_config[
```
wochinge, line 166: I don't think all `TrackerFeaturizer` support this (namely `FullDialogueTrackerFeaturizer` )

rasa/core/policies/ted_policy.py
```diff
@@ -332,12 +332,13 @@ def get_default_config() -> Dict[Text, Any]:
             # ingredients in a recipe, but it doesn't make sense for the parts of
             # an address
             SPLIT_ENTITIES_BY_COMMA: SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE,
+            # Max history of the policy, unbounded by default
+            POLICY_MAX_HISTORY: DEFAULT_MAX_HISTORY,
```
wochinge, line 336: 🙌🏻 Could you please also add the policy priority here? I forgot to do so 🙌🏻 🙈

tests/core/test_policies.py
```diff
@@ -38,10 +42,19 @@
...
+    InvalidPolicyConfig,
+    PolicyGraphComponent,
+)
 from rasa.core.policies.rule_policy import RulePolicy
 from rasa.core.policies.ted_policy import TEDPolicy
-from rasa.core.policies.memoization import AugmentedMemoizationPolicy, MemoizationPolicy
+from rasa.core.policies.memoization import (
+    AugmentedMemoizationPolicyGraphComponent,
+    MemoizationPolicyGraphComponent,
+)
```
wochinge, line 56: how about doing this to avoid an unnecessary diff?
```suggestion
from rasa.core.policies.memoization import (
    AugmentedMemoizationPolicyGraphComponent as AugmentedMemoizationPolicy,
    MemoizationPolicyGraphComponent as MemoizationPolicy,
)
```

tests/core/test_policies.py
```diff
@@ -379,28 +420,42 @@ def create_policy(
...
+            config, model_storage, resource, execution_context, featurizer
+        )
 
-    def test_featurizer(self, trained_policy: Policy, tmp_path: Path):
+    def test_featurizer(
+        self,
+        trained_policy: EitherPolicy,
+        resource: Resource,
+        model_storage: ModelStorage,
+        tmp_path: Path,
```
wochinge, line 435: no longer needed
```suggestion
```

tests/core/test_policies.py
```diff
@@ -379,28 +420,42 @@ def create_policy(
...
+        execution_context: ExecutionContext,
+    ):
         assert isinstance(trained_policy.featurizer, MaxHistoryTrackerFeaturizer)
         assert trained_policy.featurizer.state_featurizer is None
-        trained_policy.persist(str(tmp_path))
-        loaded = trained_policy.__class__.load(str(tmp_path))
+        config = self._config(DEFAULT_POLICY_PRIORITY)
+        trained_policy.persist()
```
wochinge, line 441: shouldn't be required as the policy should already be persisted during training
```suggestion
```

tests/core/test_policies.py
```diff
@@ -379,28 +420,42 @@ def create_policy(
...
     def test_memorise(
         self,
-        trained_policy: MemoizationPolicy,
+        trained_policy: MemoizationPolicyGraphComponent,
         default_domain: Domain,
         stories_path: Text,
     ):
         trackers = train_trackers(default_domain, stories_path, augmentation_factor=20)
-        trained_policy.train(trackers, default_domain, RegexInterpreter())
+        if isinstance(trained_policy, PolicyGraphComponent):
```
wochinge, line 455: why do we need this? All memoization policies should be migrated, no?

rasa/core/policies/memoization.py
twerkmeister: not sure ~ if you don't know either, I'll remove it!

rasa/core/policies/policy.py
twerkmeister: I was wondering why featurizer configs is a list at all?

rasa/core/policies/ted_policy.py
twerkmeister: will do

rasa/core/policies/policy.py
twerkmeister: alright

tests/core/test_policies.py
twerkmeister: Is that really the new default behavior, i.e. to persist automatically when trained? If so I haven't implemented it yet!

twerkmeister: The fixture isn't doing persistence either

tests/core/test_policies.py
twerkmeister: It's needed as we are subclassing PolicyTestCollection. But I actually just checked that only Memoization and Ted extend that test suite - So basically I could remove all of that old code non-graph code right?

rasa/core/policies/memoization.py
wochinge: Let's remove it 👍🏻

rasa/core/policies/policy.py
wochinge: no clue 🤷🏻  I guess some historic reason. I've just moved the code for this but didn't do any changes to the current config

tests/core/test_policies.py
wochinge: > Is that really the new default behavior, i.e. to persist automatically when trained? If so I haven't implemented it yet!

yep (see e.g. `TEDPolicy` or `DIETClassifier`)

tests/core/test_policies.py
wochinge: indeed 👍🏻

twerkmeister: Ah yikes, now I did a ton of changes but actually UnexpecTED extends TestTedPolicy, but isn't converted yet -.-

twerkmeister: ah wait seems that was finished an hour ago

rasa/core/policies/memoization.py
```diff
@@ -239,20 +251,52 @@ def predict_action_probabilities(
         return self._prediction(result)
 
     def _metadata(self) -> Dict[Text, Any]:
-        return {
-            "priority": self.priority,
-            "max_history": self.max_history,
-            "lookup": self.lookup,
-        }
+        return {"lookup": self.lookup}
```
wochinge, line 254: How about dumping `_metadata` from the superclass as it's not overridden by anything?

rasa/core/policies/memoization.py
```diff
@@ -239,20 +251,52 @@ def predict_action_probabilities(
...
+            with model_storage.read_from(resource) as path:
+                metadata_file = Path(path) / cls._metadata_filename()
+                metadata = json.loads(rasa.shared.utils.io.read_file(metadata_file))
+                lookup = metadata["lookup"]
+
+                if (Path(path) / FEATURIZER_FILE).is_file():
+                    featurizer = TrackerFeaturizer.load(path)
+
+        except (ValueError, FileNotFoundError, FileIOException):
+            logger.info(
```
wochinge, line 283: `TEDPolicy` and other policies do a `warning`. How about changing this here, too, to make it consistent?
```suggestion
            logger.warning(
```

rasa/core/policies/memoization.py
```diff
@@ -239,20 +251,52 @@ def predict_action_probabilities(
...
 
+    @classmethod
+    def load(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+        **kwargs: Any,
+    ) -> "MemoizationPolicyGraphComponent":
```
wochinge, line 268: please don't use forward references (you can remove them by importing `from __future__ import annotations`)
```suggestion
    ) -> MemoizationPolicyGraphComponent:
```

rasa/core/policies/policy.py
wochinge: @twerkmeister Seems this is still open?

rasa/core/policies/ted_policy.py
```diff
@@ -332,12 +338,15 @@ def get_default_config() -> Dict[Text, Any]:
...
+            # Determines the importance of policies, higher values take precedence
+            POLICY_PRIORITY: DEFAULT_POLICY_PRIORITY,
         }
 
-    @staticmethod
-    def _standard_featurizer(max_history: Optional[int] = None) -> TrackerFeaturizer:
+    def _standard_featurizer(self) -> TrackerFeaturizer:
         return MaxHistoryTrackerFeaturizer(
-            SingleStateFeaturizer(), max_history=max_history
+            SingleStateFeaturizer(), max_history=self.config.get(POLICY_MAX_HISTORY)
```
wochinge, line 349: we can even do this, no?
```suggestion
            SingleStateFeaturizer(), max_history=self.config[POLICY_MAX_HISTORY]
```

rasa/core/policies/ted_policy.py
```diff
@@ -360,13 +369,6 @@ def __init__(
             config[SPLIT_ENTITIES_BY_COMMA], SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE
         )
 
-        # TODO: check if this statement can be removed.
-        #  More context here -
-        #  https://github.com/RasaHQ/rasa/issues/5786#issuecomment-840762751
-        max_history = config.get(POLICY_MAX_HISTORY)
```
wochinge, line 366: 🚀

tests/core/test_policies.py
```diff
@@ -6,7 +6,7 @@
 import numpy as np
```
wochinge, line 6: thanks for cleaning this up! There are two more `TODO`s in the code which I believe can be removed, right? (Line 353 and 310)

tests/core/test_policies.py
```diff
@@ -426,9 +425,12 @@ def test_memorise(
...
         )
-        trained_policy.train(
-            trackers_no_augmentation, default_domain, RegexInterpreter()
-        )
+        if isinstance(trained_policy, PolicyGraphComponent):
+            trained_policy.train(trackers_no_augmentation, default_domain)
+        else:
+            trained_policy.train(
+                trackers_no_augmentation, default_domain, RegexInterpreter()
+            )
```
wochinge, line 433: we can drop this `if`, can't we?

wochinge (commented):
Sweet! Thanks a lot for all this cleanup work!

rasa/core/policies/policy.py
twerkmeister: Yes indeed, I didn't do it here because it would mean another 100+ line changes because it would ripple through all the policies. In Ted policy alone there are like 50 accesses to `self.config`. Let's do it in a focused, separate PR just after this to not clutter this one here?

rasa/core/policies/memoization.py
twerkmeister: It's referenced in the persist method. I think it would be good to keep it for the sake of describing the interface too

twerkmeister:
@wochinge wow thanks for your laser sharp reviewing 😁 

laser_eyes

rasa/core/policies/policy.py
wochinge: sounds good 👍🏻

rasa/core/policies/memoization.py
wochinge: but other policies are overriding that anyway. How about moving it from `Policy` then to `MemoizationPolicy`? Especially since it's not really part of the interface anymore (all the policies have to call `persist` themselves anyway)

twerkmeister: ok moved the specific implementation

twerkmeister: Hm, not sure it was really a good thing. I think after all it's a decent default implementation and gives some useful structure to future policies, even though it might not be used by more than 1 policy rn

wochinge: I see your point. Just wondering if it actually provides a useful structure as they still have to remember to call it / some policies might not need persistence anyway

twerkmeister: alright I removed those in the PolicyGraphComponent

rasa/core/policies/policy.py
```diff
@@ -176,10 +179,17 @@ def _create_featurizer(cls, policy_config: Dict[Text, Any]) -> TrackerFeaturizer
...
+        if (
+            isinstance(featurizer, MaxHistoryTrackerFeaturizer)
+            and POLICY_MAX_HISTORY in policy_config
+            and POLICY_MAX_HISTORY not in featurizer_config
+        ):
+            featurizer.max_history = policy_config[POLICY_MAX_HISTORY]
+        return featurizer
+
+    def _standard_featurizer(self) -> MaxHistoryTrackerFeaturizer:
+        """Initializes the standard featurizer for this policy."""
```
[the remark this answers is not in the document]
twerkmeister: I think that logic doesn't quite work as Memoization doesn't use SingleStateFeaturizer. However, it was possible to make it a bit more simple by making the Policy class have the behavior that uses the config max history too so that way it was identical to TED and TED's method could be removed.

rasa/core/policies/policy.py
```diff
@@ -147,18 +151,17 @@ def create(
         """Creates a new untrained policy (see parent class for full docstring)."""
         return cls(config, model_storage, resource, execution_context)
 
-    @classmethod
-    def _create_featurizer(cls, policy_config: Dict[Text, Any]) -> TrackerFeaturizer:
+    def _create_featurizer(self, policy_config: Dict[Text, Any]) -> TrackerFeaturizer:
```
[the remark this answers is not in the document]
twerkmeister: yup

rasa/core/policies/policy.py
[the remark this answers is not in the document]
twerkmeister: you could obviously make it even more abstract by definining a standard tracker feautrizer class and a standard state featurizer class per component, then you could make their behavior all equal - not sure it's worth it at this point though

joejuzl (approved):
🚀 🚀 🚀

PR #9385, opened 2021-08-17
Version support policy

m-vdb:
**Proposed changes**:
- Add section in README to document actively maintained versions
- Add new doc page to document actively maintained versions
- Update instructions for releases

After this PR is merge, an announcement should be sent on Slack `#product-engineering`


TyDunn (approved):

docs/docs/actively-maintained-versions.mdx
```diff
@@ -0,0 +1,19 @@
...
+- 1.10
+- 2.0
+- 2.1
+- 2.2
+- 2.3
+- 2.4
+- 2.5
+- 2.6
+- 2.7
+- 2.8
```
akelad, line 19: I'm worried that we'll forget to update this as we release new minors - can we not just put a 2.x, rather then listing them all explicitly?

akelad (commented):
one question

docs/docs/actively-maintained-versions.mdx
m-vdb: even if we mention it explicitly in the instructions to prepare a release? I thought that more explicit content would be better for users

akelad: well I guess the question is whether people actually read those instructions (I don't most of the time because I have done it 1000 times). Is there a way to just automate this?

m-vdb: yeah there's a way sure 😅 but I won't have the time to do it this week. Let's pick your solution and see if we want the full list or not later. It's a good first iteration. I'll update the PR

PR #9387, opened 2021-08-18
Fix script to install docs: skip node_modules

m-vdb:
**Proposed changes**:
- Fix script that updates docs for a minor. Addresses issues like [this one](https://github.com/RasaHQ/rasa/runs/3351536720?check_suite_focus=true)

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -6,6 +6,7 @@ on:
     - 'main'
     tags:
     - '**'
+  pull_request:
 
 # SECRETS
 # - GH_DOCS_WRITE_KEY: generated locally, added to github repo (public key)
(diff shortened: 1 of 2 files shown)


HotThoughts:
Wow, this was fast 🔥 

Could we add the trigger `pull_request:` [here](https://github.com/RasaHQ/rasa/blob/ce298c3e71284416dd8fe9fa674da5750ee4f8f7/.github/workflows/documentation.yml#L3) so the workflow runs in this PR and we can validate the fix? And of course we will remove it before merging.

m-vdb:
done with 7adcaba1245 👍🏻 workflow running here: https://github.com/RasaHQ/rasa/runs/3361067002?check_suite_focus=true

m-vdb:
done in: b4dac18e 👍🏻 workflow run: https://github.com/RasaHQ/rasa/pull/9387/checks?check_run_id=3361120558

m-vdb:
done 👍🏻 workflow run: https://github.com/RasaHQ/rasa/pull/9387/checks?check_run_id=3361172679

m-vdb:
@HotThoughts good call requesting me to run this in workflow. `find` options on MacOS and Ubuntu are definitely different... I'm hitting a wall here, I am not sure which type of regex I should use with `find` so that it works in the CI. I've followed [this doc](https://manpages.ubuntu.com/manpages/xenial/man7/regex.7.html) and [this SO answer](https://stackoverflow.com/a/36781850/999076), and tried out: `-regextype posix-egrep`, `-regextype egrep` and `-regextype posix-extended`, without success. (and of course without the option it also fails). Do you have any idea of what option I should use? If not, who could I ask?

HotThoughts:
@m-vdb 😞 Not only `find`, `sed` also has special syntax on MacOS (yesterday I learned hah). The syntax differences between Linux and Unix really make everyone's life harder. 😢 

The [`find` man page](https://linux.die.net/man/1/find) says

> -regex pattern
  File name matches regular expression pattern. This is a match on the whole path, not a search. For example, to match a file named './fubar3', you can use the regular expression '.*bar.' or '.*b.*3', but not 'f.*r3'. The regular expressions understood by find are by default **Emacs Regular Expressions**, but this can be changed with the -regextype option.

Emacs's regular expression syntax requires us to escape `|` and `()` match group with `\`, so I did that. 

Pushed my changes to this PR. The workflow run [succeeded](https://github.com/RasaHQ/rasa/pull/9387/checks?check_run_id=3365013987). Nice!

HotThoughts (approved):
💯 Good to go

m-vdb:
You rock! thanks 🚀 🙏🏻

PR #9394, opened 2021-08-18
Convert DIETClassifier to be a GraphComponent

### LINKED ISSUE #9275: Adapt `DIETClassifier` to new `GraphComponent` interface
**Overview of the Solution**:
Adapt the `DIETClassifier ` to the new `GraphComponent` interface.

* adapt persistence

**Definition of Done**:
- [ ] DIETClassifier adapted to the new GraphComponent interface
- [ ] Unit tests are adapted to adapted policy
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old policy implementation)

joejuzl:
**Proposed changes**:
fix #9275 

This the equivalent of #9306 but for the `DIETClassifier`

Adapted classes:
* ~`Component` -> `NLUGraphComponent`~
  * ~New graph component interface for nlu components~
  * ~Lots of code is deleted~
  * ~Methods changed to match `GraphComponent` interface~
* ~`IntentClassifier` -> `IntentClassifierGraphComponent`~
  * ~New (empty) base class for all intent classifiers~
* `EntityExtractor` -> ~`EntityExtractorGraphComponent`~ `EntityExtractorMixin`
  * ~New base class for all entity classifiers~ Now just a container for some shared functionality.
  * only small changes
* `DIETClassifier` -> `DIETClassifierGraphComponent`
  * adapt default config
  * adapt `required_packages` and `supported_languages`  
  * adapt persistence

- All the tests in `test_diet_classifier.py` had to be changed so as to not use the component_builder/train methods. Helper fixtures are added to do the pre-processing required (tokenizing and featurization).

**Things you shouldn't have a look at** ⛔ 
- please don't have a look at the modules prefixed with `_` as they only contain the old but duplicated code

### DIFF
--- a/rasa/engine/graph.py
+++ b/rasa/engine/graph.py
@@ -260,6 +260,7 @@ class ExecutionContext:
     model_id: Optional[Text] = None
     should_add_diagnostic_data: bool = False
     is_finetuning: bool = False
+    node_name: Optional[Text] = None
 
 
 class GraphNode:
@@ -322,7 +323,9 @@ class GraphNode:
         self._model_storage = model_storage
         self._existing_resource = resource
 
-        self._execution_context: ExecutionContext = execution_context
+        self._execution_context: ExecutionContext = dataclasses.replace(
+            execution_context, node_name=self._node_name
+        )
 
         self._hooks: List[GraphNodeHook] = hooks if hooks else []
 
--- a/rasa/engine/graph.py
+++ b/rasa/engine/graph.py
@@ -260,6 +260,7 @@ class ExecutionContext:
     model_id: Optional[Text] = None
     should_add_diagnostic_data: bool = False
     is_finetuning: bool = False
+    # This is set by the `GraphNode` before it is passed to the `GraphComponent`.
     node_name: Optional[Text] = None
 
 
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -5,7 +5,6 @@ from collections import defaultdict
 from pathlib import Path
 
 import numpy as np
-import os
 import scipy.sparse
 import tensorflow as tf
 
@@ -282,7 +281,7 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
                 f" We will change the default value of '{EPOCHS}' in the future to 1. "
             )
 
-        self.component_config: Dict[Text, Any] = config
+        self.component_config = config
         self._model_storage = model_storage
         self._resource = resource
         self._execution_context = execution_context
@@ -1025,7 +1024,6 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
                 message.set(ENTITIES, entities, add_to_output=True)
 
             if out and self._execution_context.should_add_diagnostic_data:
-                # TODO: JUZL: use node name
                 message.add_diagnostic_data(
                     self._execution_context.node_name, out.get(DIAGNOSTIC_DATA)
                 )
@@ -1201,7 +1199,7 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
         finetune_mode: bool = False,
     ) -> "RasaModel":
         file_name = cls.__name__
-        tf_model_file = os.path.join(model_path, file_name + ".tf_model")
+        tf_model_file = model_path / f"{file_name}.tf_model"
 
         label_key = LABEL_KEY if config[INTENT_CLASSIFICATION] else None
         label_sub_key = LABEL_SUB_KEY if config[INTENT_CLASSIFICATION] else None
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -1027,7 +1027,7 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
             if out and self._execution_context.should_add_diagnostic_data:
                 # TODO: JUZL: use node name
                 message.add_diagnostic_data(
-                    self.__class__.__name__, out.get(DIAGNOSTIC_DATA)
+                    self._execution_context.node_name, out.get(DIAGNOSTIC_DATA)
                 )
 
         return messages
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -377,11 +377,6 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
             self.component_config
         )
 
-    @staticmethod
-    def required_packages() -> List[Text]:
-        """Any extra python dependencies required for this component to run."""
-        return ["tensorflow"]
-
     @classmethod
     def create(
         cls,
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -378,9 +378,9 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
             self.component_config
         )
 
-    # package safety checks
-    @classmethod
-    def required_packages(cls) -> List[Text]:
+    @staticmethod
+    def required_packages() -> List[Text]:
+        """Any extra python dependencies required for this component to run."""
         return ["tensorflow"]
 
     @classmethod
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -264,6 +264,62 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
             MODEL_CONFIDENCE: SOFTMAX,
         }
 
+    def __init__(
+        self,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+        index_label_id_mapping: Optional[Dict[int, Text]] = None,
+        entity_tag_specs: Optional[List[EntityTagSpec]] = None,
+        model: Optional[RasaModel] = None,
+        sparse_feature_sizes: Optional[Dict[Text, Dict[Text, List[int]]]] = None,
+    ) -> None:
+        """Declare instance variables with default values."""
+        if EPOCHS not in config:
+            rasa.shared.utils.io.raise_warning(
+                f"Please configure the number of '{EPOCHS}' in your configuration file."
+                f" We will change the default value of '{EPOCHS}' in the future to 1. "
+            )
+
+        self.component_config: Dict[Text, Any] = config
+        self._model_storage = model_storage
+        self._resource = resource
+        self._execution_context = execution_context
+
+        self._check_config_parameters()
+
+        # transform numbers to labels
+        self.index_label_id_mapping = index_label_id_mapping or {}
+
+        self._entity_tag_specs = entity_tag_specs
+
+        self.model = model
+
+        self.tmp_checkpoint_dir = None
+        if self.component_config[CHECKPOINT_MODEL]:
+            self.tmp_checkpoint_dir = Path(rasa.utils.io.create_temporary_directory())
+
+        self._label_data: Optional[RasaModelData] = None
+        self._data_example: Optional[Dict[Text, Dict[Text, List[FeatureArray]]]] = None
+
+        self.split_entities_config = rasa.utils.train_utils.init_split_entities(
+            self.component_config[SPLIT_ENTITIES_BY_COMMA],
+            SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE,
+        )
+
+        self.finetune_mode = self._execution_context.is_finetuning
+        self._sparse_feature_sizes = sparse_feature_sizes
+
+        if not self.model and self.finetune_mode:
+            raise rasa.shared.exceptions.InvalidParameterException(
+                f"{self.__class__.__name__} was instantiated "
+                f"with `model=None` and `finetune_mode=True`. "
+                f"This is not a valid combination as the component "
+                f"needs an already instantiated and trained model "
+                f"to continue training in finetune mode."
+            )
+
     # init helpers
     def _check_masked_lm(self) -> None:
         if (
@@ -326,62 +382,6 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
     @classmethod
     def required_packages(cls) -> List[Text]:
         return ["tensorflow"]
-
-    def __init__(
-        self,
-        config: Dict[Text, Any],
-        model_storage: ModelStorage,
-        resource: Resource,
-        execution_context: ExecutionContext,
-        index_label_id_mapping: Optional[Dict[int, Text]] = None,
-        entity_tag_specs: Optional[List[EntityTagSpec]] = None,
-        model: Optional[RasaModel] = None,
-        sparse_feature_sizes: Optional[Dict[Text, Dict[Text, List[int]]]] = None,
-    ) -> None:
-        """Declare instance variables with default values."""
-        if EPOCHS not in config:
-            rasa.shared.utils.io.raise_warning(
-                f"Please configure the number of '{EPOCHS}' in your configuration file."
-                f" We will change the default value of '{EPOCHS}' in the future to 1. "
-            )
-
-        self.component_config: Dict[Text, Any] = config
-        self._model_storage = model_storage
-        self._resource = resource
-        self._execution_context = execution_context
-
-        self._check_config_parameters()
-
-        # transform numbers to labels
-        self.index_label_id_mapping = index_label_id_mapping or {}
-
-        self._entity_tag_specs = entity_tag_specs
-
-        self.model = model
-
-        self.tmp_checkpoint_dir = None
-        if self.component_config[CHECKPOINT_MODEL]:
-            self.tmp_checkpoint_dir = Path(rasa.utils.io.create_temporary_directory())
-
-        self._label_data: Optional[RasaModelData] = None
-        self._data_example: Optional[Dict[Text, Dict[Text, List[FeatureArray]]]] = None
-
-        self.split_entities_config = rasa.utils.train_utils.init_split_entities(
-            self.component_config[SPLIT_ENTITIES_BY_COMMA],
-            SPLIT_ENTITIES_BY_COMMA_DEFAULT_VALUE,
-        )
-
-        self.finetune_mode = self._execution_context.is_finetuning
-        self._sparse_feature_sizes = sparse_feature_sizes
-
-        if not self.model and self.finetune_mode:
-            raise rasa.shared.exceptions.InvalidParameterException(
-                f"{self.__class__.__name__} was instantiated "
-                f"with `model=None` and `finetune_mode=True`. "
-                f"This is not a valid combination as the component "
-                f"needs an already instantiated and trained model "
-                f"to continue training in finetune mode."
-            )
 
     @classmethod
     def create(
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -3,7 +3,6 @@ import copy
 import logging
 from collections import defaultdict
 from pathlib import Path
-import typing
 
 import numpy as np
 import os
@@ -1082,40 +1081,61 @@ class DIETClassifierGraphComponent(GraphComponent, EntityExtractorMixin):
         resource: Resource,
         execution_context: ExecutionContext,
         **kwargs: Any,
+    ) -> DIETClassifierGraphComponent:
+        """Loads a policy from the storage (see parent class for full docstring)."""
+        try:
+            with model_storage.read_from(resource) as model_path:
+                return cls._load(
+                    model_path, config, model_storage, resource, execution_context
+                )
+        except ValueError:
+            logger.warning(
+                f"Failed to load {cls.__class__.__name__} from model storage. Resource "
+                f"'{resource.name}' doesn't exist."
+            )
+            return cls(config, model_storage, resource, execution_context)
+
+    @classmethod
+    def _load(
+        cls,
+        model_path: Path,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
     ) -> "DIETClassifierGraphComponent":
         """Loads the trained model from the provided directory."""
-        with model_storage.read_from(resource) as model_path:
-            (
-                index_label_id_mapping,
-                entity_tag_specs,
-                label_data,
-                data_example,
-                sparse_feature_sizes,
-            ) = cls._load_from_files(model_path)
+        (
+            index_label_id_mapping,
+            entity_tag_specs,
+            label_data,
+            data_example,
+            sparse_feature_sizes,
+        ) = cls._load_from_files(model_path)
 
-            config = train_utils.update_confidence_type(config)
-            config = train_utils.update_similarity_type(config)
-            config = train_utils.update_deprecated_loss_type(config)
+        config = train_utils.update_confidence_type(config)
+        config = train_utils.update_similarity_type(config)
+        config = train_utils.update_deprecated_loss_type(config)
 
-            model = cls._load_model(
-                entity_tag_specs,
-                label_data,
-                config,
-                data_example,
-                model_path,
-                finetune_mode=execution_context.is_finetuning,
-            )
+        model = cls._load_model(
+            entity_tag_specs,
+            label_data,
+            config,
+            data_example,
+            model_path,
+            finetune_mode=execution_context.is_finetuning,
+        )
 
-            return cls(
-                config=config,
-                model_storage=model_storage,
-                resource=resource,
-                execution_context=execution_context,
-                index_label_id_mapping=index_label_id_mapping,
-                entity_tag_specs=entity_tag_specs,
-                model=model,
-                sparse_feature_sizes=sparse_feature_sizes,
-            )
+        return cls(
+            config=config,
+            model_storage=model_storage,
+            resource=resource,
+            execution_context=execution_context,
+            index_label_id_mapping=index_label_id_mapping,
+            entity_tag_specs=entity_tag_specs,
+            model=model,
+            sparse_feature_sizes=sparse_feature_sizes,
+        )
 
     @classmethod
     def _load_from_files(
--- a/rasa/nlu/extractors/extractor.py
+++ b/rasa/nlu/extractors/extractor.py
@@ -52,7 +52,8 @@ class EntityExtractorMixin(abc.ABC):
     """
 
     @property
-    def name(self):
+    def name(self) -> Text:
+        """Returns the name of the class."""
         return self.__class__.__name__
 
     def add_extractor_name(
--- a/rasa/nlu/extractors/extractor.py
+++ b/rasa/nlu/extractors/extractor.py
@@ -45,10 +45,11 @@ class EntityTagSpec(NamedTuple):
 
 
 class EntityExtractorMixin(abc.ABC):
-    """Entity extractors are components which extract entities.
+    """Provides functionality for components that do entity extraction.
 
-    They can be placed in the pipeline like other components, and can extract
-    entities like a person's name, or a location.
+    Inheriting from this class will add utility functions for entity extraction.
+    Entity extraction is the process of identifying and extracting entities like a
+    person's name, or a location from a message.
     """
 
     @property
--- a/tests/nlu/classifiers/test_diet_classifier.py
+++ b/tests/nlu/classifiers/test_diet_classifier.py
@@ -86,9 +86,6 @@ def create_diet(
 def create_train_load_and_process_diet(
     nlu_data_path: Text,
     create_diet: Callable[..., DIETClassifier],
-    default_model_storage: ModelStorage,
-    default_execution_context: ExecutionContext,
-    default_diet_resource: Resource,
     train_load_and_process_diet: Callable[..., Message],
 ) -> Callable[..., Message]:
     def inner(
@@ -112,11 +109,7 @@ def create_train_load_and_process_diet(
 
 @pytest.fixture()
 def train_load_and_process_diet(
-    nlu_data_path: Text,
-    create_diet: Callable[..., DIETClassifier],
-    default_model_storage: ModelStorage,
-    default_execution_context: ExecutionContext,
-    default_diet_resource: Resource,
+    nlu_data_path: Text, create_diet: Callable[..., DIETClassifier],
 ) -> Callable[..., Message]:
     def inner(
         diet: DIETClassifier,
--- a/tests/nlu/classifiers/test_diet_classifier.py
+++ b/tests/nlu/classifiers/test_diet_classifier.py
@@ -632,7 +632,7 @@ async def test_process_gives_diagnostic_data(
         diagnostic_data = processed_message.get(DIAGNOSTIC_DATA)
 
         # DIETClassifier should add attention weights
-        name = "DIETClassifier"
+        name = "DIETClassifierGraphComponent"
         assert isinstance(diagnostic_data, dict)
         assert name in diagnostic_data
         assert "attention_weights" in diagnostic_data[name]
--- a/tests/nlu/classifiers/test_diet_classifier.py
+++ b/tests/nlu/classifiers/test_diet_classifier.py
@@ -1,17 +1,20 @@
+import copy
 from pathlib import Path
 
 import numpy as np
 import pytest
 from unittest.mock import Mock
-from typing import List, Text, Dict, Any
+from typing import Callable, List, Optional, Text, Dict, Any
 from _pytest.monkeypatch import MonkeyPatch
 
-import rasa.model
+from rasa.engine.graph import ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.nlu import registry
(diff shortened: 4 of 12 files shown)


rasa/nlu/components.py
```diff
@@ -460,347 +466,75 @@ def name(self) -> Text:
...
-        """
-        # Check language supporting
-        language = config.language
-        if not cls.can_handle_language(language):
-            # check failed
-            raise UnsupportedLanguageError(cls.name, language)
-
-        return cls(component_config)
-
-    def provide_context(self) -> Optional[Dict[Text, Any]]:
```
joejuzl, line 597: Context is needed by components such as the spacy components, so that they can communicate with each other. This will be replaced by distinct edges in the graph.

rasa/nlu/components.py
```diff
@@ -460,347 +466,75 @@ def name(self) -> Text:
...
-        return d
+    @abc.abstractmethod
+    def persist(self) -> None:
+        """Persists this component to disk for future loading."""
+        ...
 
     def __eq__(self, other: Any) -> bool:
         return self.__dict__ == other.__dict__
 
-    def prepare_partial_processing(
```
joejuzl, line 702: Of this partial processing stuff is now obsolete as we run the pipeline in the graph.

rasa/nlu/components.py
```diff
@@ -460,347 +466,75 @@ def name(self) -> Text:
...
-
-        if self.partial_processing_context is not None:
-            for component in self.partial_processing_pipeline:
-                component.process(message, **self.partial_processing_context)
-        else:
-            logger.info("Failed to run partial processing due to missing pipeline.")
-        return message
-
     @classmethod
-    def can_handle_language(cls, language: Hashable) -> bool:
```
joejuzl, line 745: This is all handled in schema validation now.

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -323,21 +334,28 @@ def required_packages(cls) -> List[Text]:
 
     def __init__(
         self,
-        component_config: Optional[Dict[Text, Any]] = None,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
```
ka-bu, line 340: I wonder if there's a good way to get rid of the extra kwargs in a reasonable way. I mean those are only used in `load`. So we could move that there / o a separate function that might include more checks regarding "does the config we're using work with the model that we loaded" ? ( e.g. config might say embedding dim x but our loaded model has dim y then we know something went really really wrong ) Maybe that is done elsewhere ... but not having those extra args here would look nice / feel cleaner to me. What do you think?

ka-bu (commented):
Half done, will finish tomorrow :) Looks good so far! Mainly just questions

rasa/nlu/components.py
```diff
@@ -460,347 +466,75 @@ def name(self) -> Text:
...
-        ] = rasa.utils.train_utils.override_defaults(self.defaults, component_config)
+    def __init__(
+        self,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> None:
+        """Initializes a new NLU Component."""
+        self.component_config: Dict[Text, Any] = config
```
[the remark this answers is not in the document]
wochinge: should this be protected?
```suggestion
        self._component_config: Dict[Text, Any] = config
```

rasa/nlu/components.py
```diff
@@ -460,347 +466,75 @@ def name(self) -> Text:
...
-        pass
-
-    def train(
-        self,
-        training_data: TrainingData,
-        config: Optional[RasaNLUModelConfig] = None,
-        **kwargs: Any,
-    ) -> None:
+    @abc.abstractmethod
+    def train(self, training_data: TrainingData, **kwargs: Any,) -> None:
```
wochinge, line 486: do we make this required? Some components such as `Duckling` or some pretrained featurizers don't need to implement train

wochinge:
```suggestion
    def train(self, training_data: TrainingData, **kwargs: Any,) -> Resource:
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -821,12 +839,7 @@ def preprocess_train_data(self, training_data: TrainingData) -> RasaModelData:
     def _check_enough_labels(model_data: RasaModelData) -> bool:
         return len(np.unique(model_data.get(LABEL_KEY, LABEL_SUB_KEY))) >= 2
 
-    def train(
-        self,
-        training_data: TrainingData,
-        config: Optional[RasaNLUModelConfig] = None,
-        **kwargs: Any,
-    ) -> None:
+    def train(self, training_data: TrainingData, **kwargs: Any,) -> None:
```
wochinge, line 842:
```suggestion
    def train(self, training_data: TrainingData, **kwargs: Any,) -> Resource:
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -984,150 +999,139 @@ def _predict_entities(
...
+                [tag_spec._asdict() for tag_spec in self._entity_tag_specs]
+                if self._entity_tag_specs
+                else []
+            )
+            rasa.shared.utils.io.dump_obj_as_json_to_file(
+                model_path / f"{file_name}.entity_tag_specs.json", entity_tag_specs
+            )
 
-        return {"file": file_name}
+            return None
```
wochinge, line 1065:
```suggestion
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -984,150 +999,139 @@ def _predict_entities(
...
         **kwargs: Any,
-    ) -> "DIETClassifier":
+    ) -> "DIETClassifierGraphComponent":
         """Loads the trained model from the provided directory."""
-        if not meta.get("file"):
-            logger.debug(
-                f"Failed to load model for '{cls.__name__}'. "
-                f"Maybe you did not provide enough training data and no model was "
-                f"trained or the path '{os.path.abspath(model_dir)}' doesn't exist?"
+        with model_storage.read_from(resource) as model_path:
```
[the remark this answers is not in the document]
wochinge: Yeah, I think we should handle that exception

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1215,7 +1222,7 @@ def _load_model_class(
             data_signature=model_data_example.get_signature(),
             label_data=label_data,
             entity_tag_specs=entity_tag_specs,
-            config=copy.deepcopy(meta),
+            config=copy.deepcopy(config),
```
wochinge, line 1225: why the `deepcopy`?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -49,6 +50,116 @@
...
+    return Resource("DIET")
+
+
+@pytest.fixture()
+def create_diet(
+    default_model_storage: ModelStorage,
+    default_execution_context: ExecutionContext,
+    default_diet_resource: Resource,
+) -> Callable[..., DIETClassifierGraphComponent]:
+    def inner(config: Dict[Text, Any], load=False) -> DIETClassifierGraphComponent:
```
wochinge, line 64:
```suggestion
    def inner(config: Dict[Text, Any], load: bool = False) -> DIETClassifierGraphComponent:
```

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -49,6 +50,116 @@
...
+
+        if expect_intent:
+            assert classified_message.data["intent"]["name"]
+
+        loaded_diet = create_diet(diet.component_config, load=True)
+
+        classified_message2 = loaded_diet.process([message2])[0]
+
+        assert classified_message2.fingerprint() == classified_message.fingerprint()
+        return classified_message
```
wochinge, line 158: how about making this all about training the classifier moving the asserting of a message (and possible the processing of the message) to a separate function?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -57,7 +168,9 @@ def test_compute_default_label_features():
         Message(data={TEXT: "test d"}),
     ]
 
-    output = DIETClassifier._compute_default_label_features(label_features)
+    output = DIETClassifierGraphComponent._compute_default_label_features(
```
wochinge, line 171: how about doing the import hack `import `DIETClassifierGraphComponent` as `DIETClassifier` to avoid unnecessary diffs now and in the future?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -357,29 +368,17 @@ def as_pipeline(*components):
     ],
 )
 async def test_softmax_normalization(
-    component_builder,
-    tmp_path,
     classifier_params,
     data_path: Text,
     output_length,
     output_should_sum_to_1,
+    create_train_load_and_process_diet: Callable[..., Message],
```
wochinge, line 375: really sweet solution with the fixture 💯

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -743,52 +610,39 @@ async def test_doesnt_checkpoint_with_zero_eval_num_examples(
...
-    assert name in diagnostic_data
-    assert "attention_weights" in diagnostic_data[name]
-    assert isinstance(diagnostic_data[name].get("attention_weights"), np.ndarray)
-    assert "text_transformed" in diagnostic_data[name]
-    assert isinstance(diagnostic_data[name].get("text_transformed"), np.ndarray)
+    default_execution_context.should_add_diagnostic_data = should_add_diagnostic_data
+    processed_message = create_train_load_and_process_diet({EPOCHS: 1})
+
+    if should_add_diagnostic_data:
+        """Tests if processing a message returns attention weights as numpy array."""
```
wochinge, line 633:
```suggestion
        # Tests if processing a message returns attention weights as numpy array.
```

wochinge (commented):
Nice work with all the subclasses and stuff and rewriting the tests 💯

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -111,7 +120,9 @@
 POSSIBLE_TAGS = [ENTITY_ATTRIBUTE_TYPE, ENTITY_ATTRIBUTE_ROLE, ENTITY_ATTRIBUTE_GROUP]
 
 
-class DIETClassifier(IntentClassifier, EntityExtractor):
+class DIETClassifierGraphComponent(
+    IntentClassifierGraphComponent, EntityExtractorGraphComponent
```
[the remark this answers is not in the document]
joejuzl: > This makes me wonder if we want to drop the GraphComponent suffix later?

Yup our plan is to revert the names back in the "integration" phase.

> Also IntentClassifier and EntityExtractor don't contain anything related to GraphComponent really, right? (IntentClassifier definitely doesn't - 😭 ) So how about dropping those GraphComponents there and instead here DIETClassifierGraphComponent(GraphComponent,IntentClassifier, EntityExtractor) ?

Yes I think this is a very good point. In fact I think we can remove `NLUGraphComponent` (formally `Component`) completely, as it doesn't really provide anything anymore - is just another layer of subclassing (same with `IntentClassifier`)

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -984,150 +999,139 @@ def _predict_entities(
...
+        """Persist this model into the passed directory."""
         import shutil
 
         if self.model is None:
-            return {"file": None}
+            return None
 
-        model_dir_path = Path(model_dir)
-        tf_model_file = model_dir_path / f"{file_name}.tf_model"
+        with self._model_storage.write_to(self._resource) as model_path:
```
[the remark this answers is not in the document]
joejuzl: Yes, exactly.

rasa/nlu/extractors/extractor.py
```diff
@@ -40,7 +46,7 @@ class EntityTagSpec(NamedTuple):
     num_tags: int
 
 
-class EntityExtractor(Component):
+class EntityExtractorGraphComponent(NLUGraphComponent, abc.ABC):
```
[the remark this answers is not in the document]
joejuzl: `EntityExtractor` is really just a collection of functions. It could probably just be a module tbh. To make the conversion easier I can just make it into a mixin, i.e. remove it's superclasses, so it's just used for other classes to gain this functionality.

rasa/nlu/classifiers/diet_classifier.py
joejuzl: You mean `**kwargs: Any,`? 
This is needed in-case we want an output from a parent node to be used in the loading of a child node.

rasa/nlu/classifiers/diet_classifier.py
joejuzl: Because it was always like that before 😛

rasa/nlu/extractors/extractor.py
```diff
@@ -78,24 +86,6 @@ def add_processor_name(self, entity: Dict[Text, Any]) -> Dict[Text, Any]:
 
         return entity
 
-    def init_split_entities(self) -> Dict[Text, bool]:
```
joejuzl, line 81: This function is pointless, we can just call `rasa.utils.train_utils.init_split_entities` directly, like in `TEDPolicy`.

rasa/nlu/components.py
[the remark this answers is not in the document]
joejuzl: I think it's potentially useful to be able to access the config? And it's not something that solely inside the object

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -984,150 +999,139 @@ def _predict_entities(
...
-    def process(self, message: Message, **kwargs: Any) -> None:
+    def process(self, messages: List[Message], **kwargs: Any) -> List[Message]:
         """Augments the message with intents, entities, and diagnostic data."""
-        out = self._predict(message)
+        for message in messages:
+            out = self._predict(message)
 
-        if self.component_config[INTENT_CLASSIFICATION]:
-            label, label_ranking = self._predict_label(out)
+            if self.component_config[INTENT_CLASSIFICATION]:
```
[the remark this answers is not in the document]
joejuzl: 👍 I'll do it as the final commit.

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: hmm but where would the loading go? because it currently checks that the persisting and loading does not change the prediction at all.

joejuzl: Also I'm worried adding more fixture functions will make it harder to follow, given they all have to have `inner` to have access to the fixtures... I think this structure works well for all the tests now - and maybe we can re-asses in the future if it is inflexible for whatever reason?

rasa/nlu/classifiers/diet_classifier.py
ka-bu: Not really, more like in no extra arguments in the intialisation, which should all either be None or set to something, and a separate function for initialising the DIET classifier using the loaded resources. 🤔  
.... Sorry, I realize how that there's no benefit in that if don't have those extra checks. That would be just one init and the four lines of "overwrite this with that". Please just ignore my spam here ^^.

rasa/nlu/components.py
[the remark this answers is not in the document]
ka-bu: It will definitely be useful if we can access that in the extra _validate steps..... 😁  But that doesn't mean we shouldn't protect the configs from modifications 🤔

rasa/engine/graph.py
```diff
@@ -260,6 +260,7 @@ class ExecutionContext:
     model_id: Optional[Text] = None
     should_add_diagnostic_data: bool = False
     is_finetuning: bool = False
+    node_name: Optional[Text] = None
```
wochinge, line 263: should we add a comment that this is filled in by the `GraphNode` later on?

rasa/engine/graph.py
```diff
@@ -322,7 +323,9 @@ def __init__(
         self._model_storage = model_storage
         self._existing_resource = resource
 
-        self._execution_context: ExecutionContext = execution_context
+        self._execution_context: ExecutionContext = dataclasses.replace(
```
wochinge, line 326: 🚀

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -124,139 +130,195 @@ class DIETClassifier(IntentClassifier, EntityExtractor):
...
+        sparse_feature_sizes: Optional[Dict[Text, Dict[Text, List[int]]]] = None,
+    ) -> None:
+        """Declare instance variables with default values."""
+        if EPOCHS not in config:
+            rasa.shared.utils.io.raise_warning(
+                f"Please configure the number of '{EPOCHS}' in your configuration file."
+                f" We will change the default value of '{EPOCHS}' in the future to 1. "
+            )
+
+        self.component_config: Dict[Text, Any] = config
```
wochinge, line 285: nit: why do we need to restate the type? 
```suggestion
        self.component_config = config
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -316,58 +378,21 @@ def _check_config_parameters(self) -> None:
             self.component_config
         )
 
-    # package safety checks
-    @classmethod
-    def required_packages(cls) -> List[Text]:
+    @staticmethod
+    def required_packages() -> List[Text]:
+        """Any extra python dependencies required for this component to run."""
         return ["tensorflow"]
```
wochinge, line 384: funny that `TEDPolicy` doesn't do this. Might be worth adding there actually for consistency.

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -984,150 +1008,160 @@ def _predict_entities(
...
+            if self.component_config[ENTITY_RECOGNITION]:
+                entities = self._predict_entities(out, message)
 
-            message.set(ENTITIES, entities, add_to_output=True)
+                message.set(ENTITIES, entities, add_to_output=True)
 
-        if out and DIAGNOSTIC_DATA in out:
-            message.add_diagnostic_data(self.unique_name, out.get(DIAGNOSTIC_DATA))
+            if out and self._execution_context.should_add_diagnostic_data:
+                # TODO: JUZL: use node name
```
wochinge, line 1028:
```suggestion
```

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1162,16 +1195,16 @@ def _load_model(
...
+        config: Dict[Text, Any],
         data_example: Dict[Text, Dict[Text, List[FeatureArray]]],
-        model_dir: Text,
+        model_path: Path,
         finetune_mode: bool = False,
     ) -> "RasaModel":
-        file_name = meta["file"]
-        tf_model_file = os.path.join(model_dir, file_name + ".tf_model")
+        file_name = cls.__name__
+        tf_model_file = os.path.join(model_path, file_name + ".tf_model")
```
wochinge, line 1204: nit
```suggestion
        tf_model_file = model_path / f"{file_name}.tf_model"
```

rasa/nlu/classifiers/diet_classifier.py
wochinge: lol, should we drop it?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -984,150 +999,139 @@ def _predict_entities(
 
         return entities
 
-    def process(self, message: Message, **kwargs: Any) -> None:
+    def process(self, messages: List[Message], **kwargs: Any) -> List[Message]:
         """Augments the message with intents, entities, and diagnostic data."""
-        out = self._predict(message)
+        for message in messages:
```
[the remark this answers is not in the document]
joejuzl: As of now predict for DIET will only work on 0 or 1 message. However the featurizors will predict on more than one - not sure if batches will help there though.

rasa/nlu/extractors/extractor.py
```diff
@@ -40,13 +44,18 @@ class EntityTagSpec(NamedTuple):
     num_tags: int
 
 
-class EntityExtractor(Component):
+class EntityExtractorMixin(abc.ABC):
     """Entity extractors are components which extract entities.
 
     They can be placed in the pipeline like other components, and can extract
     entities like a person's name, or a location.
     """
```
wochinge, line 52: the docstring needs some adapting, hm?

rasa/nlu/classifiers/diet_classifier.py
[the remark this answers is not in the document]
ka-bu: I'm not sure I understand what you mean? When we use model_inference on a bunch of examples at once (i.e. on rasa data model build from multiple messages atm), then model_inference will perform inference for all examples in parallel, whereas by passing one example after the other, we can't do that. - Anyway, nothing we need to do here now, just something that might speed everything up later :)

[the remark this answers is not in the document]
ka-bu: Oh wait, we really just have 0 or 1 message.... I somehow thought the whole history might be passed through this. Oh, no, those are persisted in the tracker store. Damn. 😅

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -224,102 +299,36 @@ async def test_train_persist_load_with_different_settings_non_windows(
             "intent_split_symbol": "+",
         },
         {"name": "CountVectorsFeaturizer"},
-        {"name": "DIETClassifier", MASKED_LM: True, EPOCHS: 1},
     ]
-    await _train_persist_load_with_different_settings(
-        pipeline, component_builder, tmp_path, should_finetune=False
-    )
-    await _train_persist_load_with_different_settings(
```
wochinge, line 232: seems we are dropping the `should_finetune=True` tests, aren't we?

tests/nlu/classifiers/test_diet_classifier.py
wochinge: mhm, I imagined something like the snippet below. In my opinion that would make the single unit tests more verbose but easier to read as you don't have to jump through multiple fixtures in order to understand what's going on. However, I've seen that it was more or less done similarly before so fine to me.

```python
classifier = create_diet(...)
path = train_diet(classifier, pipeline, ...)
loaded_classifier = load_diet(...)

prediction1 = classifier.process(...)
prediction2 = classifier.process(...)
```

wochinge (commented):
Just commenting due to the `should_finetune=True/False` tests for now. But besides stuff seems fine (might make sense that @ka-bu has another look at the tests as well)

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: The problem is that you need to pre-process the messages before training and prediction. And for prediction you need to re-use the featurizors that were trained before. So there would be a lot of passing around things like pipelines.

wochinge: yeah, then skip it 👍🏻

rasa/nlu/classifiers/diet_classifier.py
joejuzl: I had a slight worry that it was done because the model manipulated the config (i.e. like `DIETClassifier` does). And didn't want some strange hard to find bug to pop up later before of this change.... but we could YOLO and change it?

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: There weren't actually testing finetuning as they never trained again on the loaded model. There are tests lower in the file which actually testing finetuning properly.

wochinge: but they test if the classifier can be loaded correctly with the given config and in finetune mode, no?

joejuzl: But nothing really different happens until you actually start training and `adjust_for_incremental_training` is called. And all of that is tested in other tests. 
I can add it back in, but I think it kind of gives a false expectation on what the test is actually doing.

rasa/nlu/classifiers/diet_classifier.py
ka-bu: Seems DIET just uses the load inherited from RasaModel atm which just does: https://github.com/RasaHQ/rasa/blob/08bf76bd70cc58784556e69c60b03570149f1119/rasa/utils/tensorflow/models.py#L422
And DIET2BOW and DIET2DIET don't change anything either (but do the copy in model_load call as well). So .... :D  

(In general, I'd do a copy in all __init__s etc. but I think that should go into that refactor configs task one day :D)

joejuzl: The config actually ends up going all the way into `tf.keras.models.Model` via kwargs - although I have no idea if it does anything with them (probably not).

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: I've `parametrize`'d these tests to do it again.

wochinge: instead of parametrizing it - could we just try to load it twice? This would avoid one training and be great for the CI speed

wochinge (approved):
🚀

rasa/engine/graph.py
```diff
@@ -260,6 +260,8 @@ class ExecutionContext:
     model_id: Optional[Text] = None
```
ka-bu, line 260: Just "discovered" the `node_name` and now I'm wondering: Should we call this `node_id` then? And do we even need this? Isn't `node_name` used like an id? Don't both have to be unique because we do things like `message.add_diagnostic_data(self._execution_context.node_name, out.get(DIAGNOSTIC_DATA))` later ?  🤔

joejuzl: The thing is we call it `node_name` everywhere through-out the rest of the engine and graph code - so maybe this is not the best place to change it.

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: done

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -49,6 +52,120 @@
...
+        config: Dict[Text, Any], load: bool = False, finetune: bool = False
+    ) -> DIETClassifier:
+        if load:
+            constructor = DIETClassifier.load
+        else:
+            constructor = DIETClassifier.create
+
+        default_execution_context.is_finetuning = finetune
+        return constructor(
+            config={**DIETClassifier.get_default_config(), **config},
```
ka-bu, line 76: Think we can drop this, GraphNode should take care of filling in defaults, no?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -49,6 +52,120 @@
 from rasa.utils.tensorflow.model_data_utils import FeatureArray
 
 
+@pytest.fixture()
+def default_diet_resource() -> Resource:
+    return Resource("DIET")
+
+
+@pytest.fixture()
+def create_diet(
```
ka-bu, line 61: I like the fixture+inner constructs :) Except this one, that looks like a a complicated way of having Resource("DIET") as the default 😅

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -49,6 +52,120 @@
...
+        )
+
+    return inner
+
+
+@pytest.fixture()
+def create_train_load_and_process_diet(
+    nlu_data_path: Text,
+    create_diet: Callable[..., DIETClassifier],
+    default_model_storage: ModelStorage,
```
ka-bu, line 89: Shouldn't we be able to drop e.g. `default_model_storage` here, as we already have `create_diet`?

ka-bu: (same below)

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -449,126 +436,85 @@ async def test_inner_linear_normalization(
...
-    parse_data = loaded.parse("hello")
+    parse_data = parsed_message.data
     intent_ranking = parse_data.get("intent_ranking")
 
     # check that the output was not normalized
     mock.normalize.assert_not_called()
 
     # check that the output was correctly truncated
-    assert len(intent_ranking) == output_length
+    assert len(intent_ranking) == LABEL_RANKING_LENGTH
```
ka-bu, line 456: was this maybe a parameter because `LABEL_RANKING_LENGTH` is just a default?

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -449,126 +436,85 @@ async def test_inner_linear_normalization(
...
     monkeypatch.setattr(train_utils, "normalize", mock.normalize)
 
-    _config = RasaNLUModelConfig({"pipeline": pipeline})
-    (trained_model, _, persisted_path) = await rasa.nlu.train.train(
-        _config,
-        path=str(tmpdir),
-        data="data/test/many_intents.yml",
-        component_builder=component_builder,
+    parsed_message = create_train_load_and_process_diet(
+        {LOSS_TYPE: "margin", RANDOM_SEED: 42, EPOCHS: 1},
```
ka-bu, line 446: Btw, created an item on the "do this later" list for being able to set the epochs to 0 (for example) here - because it would suffice to work with a randomly initialised model here.

tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -578,114 +524,58 @@ async def test_train_tensorboard_logging(
...
     )
 
-    await rasa.nlu.train.train(
-        _config,
-        path=str(tmp_path),
-        data=nlu_data_path,
-        component_builder=component_builder,
-        fixed_model_name=model_name,
-    )
+    with default_model_storage.read_from(default_diet_resource) as model_dir:
```
ka-bu, line 535: shouldn't we rather grab the self._resource from the model directly?

ka-bu (commented):
Just some small questions/suggestions to make the tests that already look so much cleaner and shorter now 🥳   maybe even another tiny bit shorter :)

rasa/engine/graph.py
joejuzl: oh yes sorry - it's exactly the same. We are just now passing it into the graph components.

rasa/nlu/classifiers/diet_classifier.py
joejuzl: I guess it's not actually, as it's a dependency of rasa. Shall I remove it?

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: This is not using the `GraphNode`, just the component directly so we have to do it.

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: Without it we would also have to pass in `default_model_storage` and `default_execution_context` everytime.

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: good catch

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: I think these ended up only having one item in paramaterize after recent deprecation removals - so I just removed the pointless `paramaterize`.

tests/nlu/classifiers/test_diet_classifier.py
joejuzl: Could do, although it's a protected member, and we know this will be the same one.

rasa/engine/graph.py
ka-bu: thanks, just wasn't sure I understood this right 🥳

tests/nlu/classifiers/test_diet_classifier.py
ka-bu: Hm, `model_storage` and `default_execution_contexts` could also have defaults? ... But I see that one day we might want to test different storages/contexts, so that'll be handy then :)

tests/nlu/classifiers/test_diet_classifier.py
ka-bu: As a lazy person I don't like having to look into create_diet here to find out they're the same :)  - but I also agree that doing that via protected variable isn't nice and there isn't really a reason to introduce some getter because the models have the .load already 🤔 . So I'm fine leaving it as is.

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -316,58 +377,21 @@ def _check_config_parameters(self) -> None:
...
-
-        if not self.model and self.finetune_mode:
-            raise rasa.shared.exceptions.InvalidParameterException(
-                f"{self.__class__.__name__} was instantiated "
-                f"with `model=None` and `finetune_mode=True`. "
-                f"This is not a valid combination as the component "
-                f"needs an already instantiated and trained model "
-                f"to continue training in finetune mode."
-            )
+    @classmethod
```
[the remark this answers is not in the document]
ka-bu: Aaaah, sorry, I mixed up create vs load

tests/nlu/classifiers/test_diet_classifier.py
ka-bu: Oh right, thanks! Would be cool if we had a way to consistently and automatically do that for GraphComponents as well, will add that to the notion list

ka-bu (approved):

PR #9395, opened 2021-08-18
Removed RegexInterpreter as a Default when NLU Interpreter Created

### LINKED ISSUE #9319: RegexInterpreter Used When NLU Model Not Present is Causing Intent Prediction Issues
_Rasa version_ : 2.7.1

 _Operating system_ (windows, osx, ...): Linux

 _Issue_ : When the NLU model is still being loaded into memory, [this case for the NLU interpreter object| [https://github.com/RasaHQ/rasa/blob/main/rasa/core/interpreter.py#L32](https://github.com/RasaHQ/rasa/blob/main/rasa/core/interpreter.py#L32)] will not apply and the RegexInterpreter gets used as the default. As a result, the user's raw input text gets predicted as the intent, e.g. `intent=Intent(name=Hello I needed help to follow up on my extension I needed for my bill, confidence=1.0)`.

This is causing problems for customers that are using production data to analyze intent distribution since the predicted intent ends up being assigned as the raw text.

b-quachtran:
**Proposed changes**:
- Removed use of RegexInterpreter as the default interpreter when endpoint config is `None`
- Closes #9319  

### DIFF
--- a/rasa/core/interpreter.py
+++ b/rasa/core/interpreter.py
@@ -171,7 +171,9 @@ def _create_from_endpoint_config(
 ) -> rasa.shared.nlu.interpreter.NaturalLanguageInterpreter:
     """Instantiate a natural language interpreter based on its configuration."""
 
-    if endpoint_config.type is None or endpoint_config.type.lower() == "http":
+    if endpoint_config and (
+        endpoint_config.type is None or endpoint_config.type.lower() == "http"
+    ):
         return RasaNLUHttpInterpreter(endpoint_config=endpoint_config)
     else:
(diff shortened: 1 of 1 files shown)


wochinge (commented):
Can you please add a test which verifies the new behavior? I'd like to avoid that we break this again in the future 🙈

b-quachtran:
@wochinge Yes for sure, still need to fix the currently failing test cases as well 😅

b-quachtran:
Huh I think it auto-requested a review, my bad it's not close to ready yet

PR #9402, opened 2021-08-19
Fix some wording issues on TEDPolicy documentation

usc-m:
**Proposed changes**:
- add some missing words to help make the TEDPolicy documentation a bit more readable

### DIFF
--- a/docs/docs/policies.mdx
+++ b/docs/docs/policies.mdx
@@ -87,7 +87,7 @@ a multi-task architecture for next action prediction and entity
 recognition. The architecture consists of several transformer encoders which are shared for both tasks.
 A sequence of entity labels is predicted through a Conditional Random Field (CRF) tagging layer on top of the
 user sequence transformer encoder output corresponding to the input sequence of tokens.
-For the next action prediction, the dialogue transformer encoder output and system action labels are embedded into a
+For the next action prediction, the dialogue transformer encoder output and the system action labels are embedded into a
 single semantic vector space. We use the dot-product loss to maximize the similarity with the target label and
 minimize similarities with negative samples.
 
--- a/docs/docs/policies.mdx
+++ b/docs/docs/policies.mdx
@@ -84,10 +84,10 @@ Doing so can lead to unexpected and undesired bot behavior.
 
 The Transformer Embedding Dialogue (TED) Policy is
 a multi-task architecture for next action prediction and entity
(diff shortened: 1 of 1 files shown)


docs/docs/policies.mdx
```diff
@@ -84,10 +84,10 @@ Doing so can lead to unexpected and undesired bot behavior.
 
 The Transformer Embedding Dialogue (TED) Policy is
 a multi-task architecture for next action prediction and entity
-recognition. The architecture several transformer encoders which are shared for both tasks.
+recognition. The architecture consists of several transformer encoders which are shared for both tasks.
 A sequence of entity labels is predicted through a Conditional Random Field (CRF) tagging layer on top of the
 user sequence transformer encoder output corresponding to the input sequence of tokens.
-For the next action prediction the dialogue transformer encoder output and system action labels are embedded into a
+For the next action prediction, the dialogue transformer encoder output and system action labels are embedded into a
```
usc-m, line 90: "Dialogue transformer encoder output and system action labels" is also a bit of a mouthful, though I'm not certain what would be the best way to change it. Any ideas?

b-quachtran: Hmm, what about "transformer encoding and action labels"

usc-m: I think my specific issue is that it feels a bit ambiguous at current - not totally clear on my first pass reading it if "system action labels" belong to the "dialogue transformer encoder". I think your suggestion is a little more clear in that regard but I'm wondering if we're losing important context by dropping the name of the specific transformer.

Perhaps "For the next action prediction, the system action labels and the output of the dialogue transformer encoder are embedded into a single semantic vector space"? Does that read more clearly without causing confusion between the two things?

wochinge:
@b-quachtran I let you handle the review 🙌🏻

docs/docs/policies.mdx
b-quachtran: The action labels themselves are separate from the transformer encoding, but are related in the sense that labels predicted at state N are computed using the encoding derived from state N-1. 

Yeah dialogue can be left in to avoid confusion and avoid ambiguity since a different transformer architecture is used for DIET.

b-quachtran (approved):

docs/docs/policies.mdx
b-quachtran: It reads clearly with the "the" added 👍

Issue #9277, opened 2021-08-05
Implement `Recipe` to convert current config to `GraphSchema`


wochinge:
**Overview of the Solution**:
Implement the `Recipe` class which converts the current config to a train and predict `GraphSchema` using the given CLI params. This component loads all the classes and uses the  `@DefaultV1.register(...)` decorator to create the graph structure from it.

See the [implementation proposal](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e) for further details.

**Blockers** (if relevant):
Existing components needs to be adapted to new graph interface.

**Definition of Done**:
- [ ] Recipe is tested with different configurations (default configuration, configurations with custom components / policies, auto configuration)
- [ ] `Recipe` interface is implemented
- [ ] all default Rasa Open Source components / policies are decorated with the `@DefaultV1.register(...)`
- [ ] the configuration is validated
    - Check pipeline is not empty
    - Checks only one tokenizer is present

wochinge:
We could consider omitting Duckling (and other pretrained extractors) from the `train_graph` 🤔 See https://github.com/RasaHQ/rasa/issues/3389#issuecomment-887308951

wochinge:
omit e2e featurization when we only train a core model

m-vdb:
@wochinge and @joejuzl to split this issue

wochinge:
Moved this back to `Backlog` as the core functionality is implemented but now have to until all things have been changed to be `GraphComponents` before wrapping this up

PR #9403, opened 2021-08-19
implement DefaultV1 Recipe

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9277
- implement recipe which converts current config to graph schemas
- logic to convert regular configs to train and predict graph schemas
- interpolates passed CLI params in to the graph components config
- handles Core only / NLU only training
- implemented decorator which allows to register classes with the `DefaultV1Recipe`

**Open Todos**
- [ ] wait until all graph components have been adapted to `GraphComponent` interface and then register all these classes with the decorator
- [ ] change the current `isinstance` checks to check the component type via the registered `ComponentType`


wochinge:
@joejuzl that's what I have so far

wochinge:
The new edge case about "rule_only_data" would still need to be considered (see [here](https://rasa-hq.slack.com/archives/C01HMTEGX37/p1629877242006400))

joejuzl:
@wochinge reminder: `JiebaTokenizer` persists and loads. (Although we may be able to avoid it.)

wochinge:
reminder to pass language to Duckings model config if ducklings model config is not filled

wochinge:
`EntitySynonymMapper` doesn't add anything during training which is required by subsequent components

wochinge:
we also need autoconfig in there

rasa/core/policies/_ensemble.py
```diff
@@ -74,6 +72,8 @@ def __init__(
         self._set_rule_only_data()
 
     def _set_rule_only_data(self) -> None:
+        from rasa.core.policies.rule_policy import RulePolicy
```
wochinge, line 75: Changes here were needed to avoid cyclic imports

Issue #9332, opened 2021-08-12
implement `DomainProvider`


wochinge:
**Overview of the Solution**:
We need to implement the `DomainProvider`. This component receives an instantiated `TrainingDataImporter` and returns the result of `importer.get_domain(...)` 

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#24ef56343a554fe88748651864fe9e77) shows the component's position in the graph.

**Input to the component**
- `TrainingDataImporter`

**Output:** 
* a `Domain` object

**Blockers**
* https://github.com/RasaHQ/rasa/pull/9184

**Definition of Done**:
- [x] Component is implemented and unit tested

ancalita:
@wochinge to double-check - this component should be placed in `rasa/engine/training/components.py`?
Or should I wait for you to merge the tracker-generator PR and then I can create a module just for `DomainProvider` and place it in `rasa/components/providers`? I could prioritise the RulePolicy issue I have instead 🤔

wochinge:
good question! @joejuzl I'd actually like to keep all these components out of the engine. Hence, I've created a new top level package with components [here](https://github.com/RasaHQ/rasa/pull/9361/files). What do you think?

joejuzl:
> good question! @joejuzl I'd actually like to keep all these components out of the engine. Hence, I've created a new top level package with components here. What do you think?

Makes sense to me - although maybe `graph_components` dir name?

wochinge:
sounds good 👍🏻 Will adapt my PR accordingly

PR #9419, opened 2021-08-20
Implement `DomainProvider` graph component

ancalita:
**Proposed changes**:
- implements #9332 

### DIFF
--- a/rasa/graph_components/providers/domain_provider.py
+++ b/rasa/graph_components/providers/domain_provider.py
@@ -20,7 +20,7 @@ class DomainProvider(GraphComponent):
         """Creates domain provider."""
         self._model_storage = model_storage
         self._resource = resource
-        self.domain = domain
+        self._domain = domain
 
     @classmethod
     def create(
@@ -31,6 +31,18 @@ class DomainProvider(GraphComponent):
         execution_context: ExecutionContext,
     ) -> DomainProvider:
         """Creates component (see parent class for full docstring)."""
+        return cls(model_storage, resource)
+
+    @classmethod
+    def load(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+        **kwargs: Any,
+    ) -> DomainProvider:
+        """Creates provider using a persisted version of itself."""
         with model_storage.read_from(resource) as resource_directory:
             domain = Domain.from_path(resource_directory)
         return cls(model_storage, resource, domain)
@@ -38,7 +50,7 @@ class DomainProvider(GraphComponent):
     def persist(self, domain: Domain) -> None:
         """Persists domain to model storage."""
         with self._model_storage.write_to(self._resource) as resource_directory:
-            domain.persist(resource_directory.joinpath("domain.yml"))
+            domain.persist(resource_directory / "domain.yml")
 
     def provide_train(self, importer: TrainingDataImporter) -> Domain:
         """Generates loaded Domain of the bot during training."""
@@ -48,4 +60,4 @@ class DomainProvider(GraphComponent):
 
     def provide_inference(self) -> Optional[Domain]:
         """Provides the domain during inference."""
-        return self.domain
+        return self._domain
--- a/rasa/graph_components/providers/domain_provider.py
+++ b/rasa/graph_components/providers/domain_provider.py
@@ -53,7 +53,7 @@ class DomainProvider(GraphComponent):
             domain.persist(resource_directory / "domain.yml")
 
     def provide_train(self, importer: TrainingDataImporter) -> Domain:
-        """Generates loaded Domain of the bot during training."""
+        """Provides domain from training data during training."""
         domain = importer.get_domain()
         self.persist(domain)
         return domain
--- a/rasa/graph_components/providers/domain_provider.py
+++ b/rasa/graph_components/providers/domain_provider.py
@@ -41,11 +41,11 @@ class DomainProvider(GraphComponent):
             domain.persist(resource_directory.joinpath("domain.yml"))
 
     def provide_train(self, importer: TrainingDataImporter) -> Domain:
-        """Generates loaded Domain of the bot."""
+        """Generates loaded Domain of the bot during training."""
         domain = importer.get_domain()
         self.persist(domain)
         return domain
 
-    def provide_inference(self):
+    def provide_inference(self) -> Optional[Domain]:
         """Provides the domain during inference."""
         return self.domain
--- a/rasa/graph_components/providers/domain_provider.py
+++ b/rasa/graph_components/providers/domain_provider.py
@@ -39,7 +39,6 @@ class DomainProvider(GraphComponent):
         return cls(config)
 
     @staticmethod
-    def generate_domain(config_path: Text, domain_path: Text) -> Domain:
+    def generate_domain(importer: TrainingDataImporter) -> Domain:
         """Generates loaded Domain of the bot."""
-        importer = TrainingDataImporter.load_from_config(config_path, domain_path)
         return importer.get_domain()
--- a/rasa/graph_components/providers/domain_provider.py
+++ b/rasa/graph_components/providers/domain_provider.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+from typing import Dict, Text, Any
+
+from rasa.engine.graph import GraphComponent, ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.shared.core.domain import Domain
+from rasa.shared.importers.importer import TrainingDataImporter
+
+
+class DomainProvider(GraphComponent):
+    """Provides domain during training and inference time."""
+
+    def __init__(self, config: Dict[Text, Any]) -> None:
+        """Creates domain provider from config."""
+        self._config = config
+
+    @staticmethod
+    def get_default_config() -> Dict[Text, Any]:
+        """Returns default configuration (see parent class for full docstring)."""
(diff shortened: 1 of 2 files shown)


rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,44 @@
...
+    def get_default_config() -> Dict[Text, Any]:
+        """Returns default configuration (see parent class for full docstring)."""
+        return {
+            "remove_duplicates": True,
+            "unique_last_num_states": None,
+            "augmentation_factor": 50,
+            "tracker_limit": None,
+            "use_story_concatenation": True,
+            "debug_plots": False,
+        }
```
twerkmeister, line 28: Where do these config parameters come from? They don't seem to be related to domains

rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Dict, Text, Any
+
+from rasa.engine.graph import GraphComponent, ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.shared.core.domain import Domain
+from rasa.shared.importers.importer import TrainingDataImporter
+
+
+class DomainProvider(GraphComponent):
```
twerkmeister, line 11: Reading the documentation I saw that the DomainProvider is marked as a persisting component. However, I also don't fully understand what should be persisted here and why ~ @wochinge is that persisting still relevant or did we remove that anyway by making the TrainingDataImporter fingerprint random?

twerkmeister (commented):
Found two things :)

rasa/graph_components/providers/domain_provider.py
ancalita: 😄 should have added a comment that because I wasn't sure if this component should have a config, I left these ones in from the template of `TrainingTrackerProvider` in hope to find out more during the review. I can remove this method if it's preferred to have an empty config.

wochinge: You can remove it as this one doesn't require a config 👍🏻

rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,44 @@
...
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> DomainProvider:
+        """Creates component (see parent class for full docstring)."""
+        return cls(config)
+
+    @staticmethod
+    def generate_domain(importer: TrainingDataImporter) -> Domain:
```
wochinge, line 42: Could we call this `provide`? Makes it a bit simpler to generate the graph in the recipe if similar things have similar methods
```suggestion
    def provide(importer: TrainingDataImporter) -> Domain:
```

rasa/graph_components/providers/domain_provider.py
wochinge: @twerkmeister Very good point! This component needs to be able to persist as this component also provides the domain during inference (our policies need the domain during inference). So the code should roughtly look like this:

```python

class DomainProvider:

   def __init__(domain: Optional):
    ...
    
   def provide_train(self, importer):
      domain = importer.get_domain()
      self.persist(self._resource, self._model_storage, domain)
      return domain
      
   def load(...):
     with model_storage(...):
        domain = read_domain_from_file)
     return cls(domain)
     
   def provide_inference(self):
      return self.domain
```

wochinge (changes requested):

rasa/graph_components/providers/domain_provider.py
ancalita: @wochinge which approach should I take for `persist` method 🤔 :
```
with self._model_storage.write(...) as directory:
        self.domain.persist(directory)
```
OR

```
domain_yaml = self.domain.as_yaml()
self._resource.to_cache(Path(domain_yaml), self._model_storage)
```

wochinge: the first one please 🙌🏻 

The second one should ideally never be called manually. That's done by the the `LocalTrainingCache` in the `TrainingHook`

rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,51 @@
...
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> DomainProvider:
+        """Creates component (see parent class for full docstring)."""
+        with model_storage.read_from(resource) as resource_directory:
+            domain = Domain.from_path(resource_directory)
+        return cls(model_storage, resource, domain)
```
wochinge, line 36: this should be done in `load`. `create` is called during the training and in that case the `model_storage` won't contain an entry for this resource (it should even raise in this case 🤔 )

rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def persist(self, domain: Domain) -> None:
+        """Persists domain to model storage."""
+        with self._model_storage.write_to(self._resource) as resource_directory:
+            domain.persist(resource_directory.joinpath("domain.yml"))
```
wochinge, line 41: optional: slightly fancier and (maybe) easier to read
```suggestion
            domain.persist(resource_director / "domain.yml")
```

rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(
+        self,
+        model_storage: ModelStorage,
+        resource: Resource,
+        domain: Optional[Domain] = None,
+    ) -> None:
+        """Creates domain provider."""
+        self._model_storage = model_storage
+        self._resource = resource
+        self.domain = domain
```
wochinge, line 23:
```suggestion
        self._domain = domain
```

tests/test_graph_components/providers/test_domain_provider.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_domain_provider_generates_and_persists_domain(
+    default_model_storage: ModelStorage,
+    default_execution_context: ExecutionContext,
+    tmp_path: Path,
+    config_path: Text,
+    domain_path: Text,
+):
+    resource = Resource.from_cache("xy", tmp_path, default_model_storage)
```
wochinge, line 19: You can just use the constructor. The `from_cache` parts is basically always off limits as this is only called by the `TrainingHook`
```suggestion
    resource = Resource("some resource")
```

tests/test_graph_components/providers/test_domain_provider.py
```diff
@@ -0,0 +1,37 @@
...
+        default_model_storage,
+        resource,
+        default_execution_context,
+    )
+    assert isinstance(component, DomainProvider)
+
+    importer = TrainingDataImporter.load_from_config(config_path, domain_path)
+    domain = component.provide_train(importer)
+
+    assert isinstance(domain, Domain)
```
wochinge, line 31: can we also test `provide_inference`?

rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,63 @@
...
+            domain = Domain.from_path(resource_directory)
+        return cls(model_storage, resource, domain)
+
+    def persist(self, domain: Domain) -> None:
+        """Persists domain to model storage."""
+        with self._model_storage.write_to(self._resource) as resource_directory:
+            domain.persist(resource_directory / "domain.yml")
+
+    def provide_train(self, importer: TrainingDataImporter) -> Domain:
+        """Generates loaded Domain of the bot during training."""
```
wochinge, line 56: As it doesn't "generate" anything I'd go with something like this? What do you think? 
```suggestion
        """Provides domain from training data during training."""
```

tests/test_graph_components/providers/test_domain_provider.py
```diff
@@ -0,0 +1,43 @@
...
+
+    with default_model_storage.read_from(resource) as d:
+        match = list(d.glob("**/domain.yml"))
+        assert len(match) == 1
+        assert match[0].is_file()
+        assert domain.fingerprint() == Domain.from_path(match[0]).fingerprint()
```
wochinge, line 35: can we maybe compare this to the content of the `domain` fixture? E.g. this wouldn't fail if we somehow persist the wrong domain to the `model_storage` and return the wrong domain in `provide_train` at the same time

rasa/graph_components/providers/domain_provider.py
```diff
@@ -0,0 +1,63 @@
...
+        resource: Resource,
+        execution_context: ExecutionContext,
+        **kwargs: Any,
+    ) -> DomainProvider:
+        """Creates provider using a persisted version of itself."""
+        with model_storage.read_from(resource) as resource_directory:
+            domain = Domain.from_path(resource_directory)
+        return cls(model_storage, resource, domain)
+
+    def persist(self, domain: Domain) -> None:
```
wochinge, line 50: as it's not used outside of the class
```suggestion
    def _persist(self, domain: Domain) -> None:
```

wochinge (approved):
🎸

rasa/graph_components/providers/domain_provider.py
ancalita: I also wanted to ask should I separate the tests to improve the coverage - atm that check fails because of coverage decrease?

wochinge: this test seems to be wrong - I just checked and you have a 100% test coverage. I fail to see how you could test better 😆 I'd just ignore the check for as it's not a required check.

Issue #9326, opened 2021-08-12
implement `NLUMessageConverter`


wochinge:
**Overview of the Solution**:
We need to implement the `NLUMessageConverter`. This component runs during inferences and takes a `UserMessage` object and converts it into a list of `Message` objects which can then be processed by our "NLU" components (tokenizer, featurizer, classifiers).

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#48704bc28dab486dad98e9ebd3ab9077) shows the component's position in the graph.

**Input to the component**
`Optional[UserMessage]` (the message might be `None` in case we are predicting actions after actions (in contrast to predicting the first action after a user message).

**Output:** 
A list of `Message` objects. The list has either length 1 (in case `UserMessage is not None`) or length 0 (in case `UserMessage is None`).

**Definition of Done**:
- [x] Component is implemented and unit tested

ancalita:
@wochinge would this need a separate directory? or a module in dir `graph_components` is enough?

wochinge:
yeah, how about making a `converts` package within `graph_components`?

Issue #9327, opened 2021-08-12
implement `NLUPredictionToHistoryAdder`


wochinge:
**Overview of the Solution**:
[`NLUPredictionToHistoryAdder`](https://www.notion.so/rasa/NLUPredictionToHistoryAdder-10f22c1e118a4a09ae8291028d554f2a) is a component which 
- runs during inference
- and adds the NLU prediction (`Message` object with text tokens, intent, entities) to the `DialogueStateTracker` by adding a `UserUttered` event. The tracker containing the `UserUttered` event will then be used by the policies to make a prediction.

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#48704bc28dab486dad98e9ebd3ab9077) shows the component's position in the graph.

**Input to the component**
- the `List[Message]` which contains the NLU prediction. Note that the list might be an empty list in case we are predicting an action after an action (in contrast to predicting the first action after a user message).
- the `DialogueStateTracker` retrieved from the tracker store
- the current `Domain` (required for adding stuff to the tracker)
- the original `UserMessage` object which contains information like channels etc. 

**Output:** 
The potentially updated `DialogueStateTracker` (in case we received an empty list of `Message`s we don't need to update it)

**Definition of Done**:
- [ ] Component is implemented and unit tested

PR #9420, opened 2021-08-20
Implement NLUMessageConverter

ancalita:
**Proposed changes**:
- implements #9326 

### DIFF
--- a/rasa/graph_components/converters/nlu_message_converter.py
+++ b/rasa/graph_components/converters/nlu_message_converter.py
@@ -35,13 +35,9 @@ class NLUMessageConverter(GraphComponent):
         if message:
             data = {
                 TEXT: message.text,
-                "input_channel": message.input_channel,
                 "message_id": message.message_id,
                 "metadata": message.metadata,
             }
-
-            if message.parse_data:
-                data["parse_data"] = message.parse_data
 
             return [Message(data=data)]
 
--- a/rasa/graph_components/converters/nlu_message_converter.py
+++ b/rasa/graph_components/converters/nlu_message_converter.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+from typing import Dict, Text, Any, Optional, List
+
+from rasa.core.channels.channel import UserMessage
+
+from rasa.engine.graph import GraphComponent, ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.shared.nlu.constants import TEXT, INTENT, ENTITIES
+from rasa.shared.nlu.training_data.message import Message
+
+
+class NLUMessageConverter(GraphComponent):
+    """Converts the user message into a NLU Message object."""
+
+    def __init__(self, config: Dict[Text, Any]) -> None:
+        """Creates converter from config."""
+        self._config = config
+
+    @staticmethod
+    def get_default_config() -> Dict[Text, Any]:
+        """Returns default configuration (see parent class for full docstring)."""
+        return {
+            "remove_duplicates": True,
+            "unique_last_num_states": None,
+            "augmentation_factor": 50,
+            "tracker_limit": None,
+            "use_story_concatenation": True,
+            "debug_plots": False,
+        }
+
+    @classmethod
+    def create(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> NLUMessageConverter:
+        """Creates component (see parent class for full docstring)."""
+        return cls(config)
+
+    @staticmethod
+    def convert_user_message(message: Optional[UserMessage]) -> List[Optional[Message]]:
+        """Converts user message into Message object.
+
+        Returns:
+            List containing only one instance of Message.
+            Else empty list if user message is None.
+        """
+        if message:
+            data = dict()
+            data[TEXT] = message.text
+            data["input_channel"] = message.input_channel
+            data["message_id"] = message.message_id
+            data["metadata"] = message.metadata
+
+            if message.parse_data:
+                data[INTENT] = message.parse_data[INTENT]
+                data[ENTITIES] = message.parse_data[ENTITIES]
+                data["parse_data"] = message.parse_data
+
+            return [Message(data=data)]
+
+        return []
--- a/rasa/graph_components/converters/nlu_message_converter.py
+++ b/rasa/graph_components/converters/nlu_message_converter.py
@@ -6,28 +6,12 @@ from rasa.core.channels.channel import UserMessage
 from rasa.engine.graph import GraphComponent, ExecutionContext
 from rasa.engine.storage.resource import Resource
 from rasa.engine.storage.storage import ModelStorage
-from rasa.shared.nlu.constants import TEXT, INTENT, ENTITIES
+from rasa.shared.nlu.constants import TEXT
 from rasa.shared.nlu.training_data.message import Message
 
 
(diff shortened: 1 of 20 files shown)


rasa/graph_components/converters/nlu_message_converter.py
```diff
@@ -0,0 +1,65 @@
...
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> NLUMessageConverter:
+        """Creates component (see parent class for full docstring)."""
+        return cls(config)
+
+    @staticmethod
+    def convert_user_message(message: Optional[UserMessage]) -> List[Optional[Message]]:
```
wochinge, line 44:
```suggestion
    def convert_user_message(message: Optional[UserMessage]) -> List[Message]:
```

rasa/graph_components/converters/nlu_message_converter.py
```diff
@@ -0,0 +1,65 @@
...
+        """Converts user message into Message object.
+
+        Returns:
+            List containing only one instance of Message.
+            Else empty list if user message is None.
+        """
+        if message:
+            data = dict()
```
wochinge, line 52: optional: maybe something like this is slightly easier to read?

```python
data = {
  TEXT: message.text,
  "input_channel": message.input_channel,
  "message_id: message.message_id,
  ...
}
```

rasa/graph_components/converters/nlu_message_converter.py
```diff
@@ -0,0 +1,65 @@
...
+            data[TEXT] = message.text
+            data["input_channel"] = message.input_channel
```
wochinge, line 54: Do we actually need the input channel here? [This code](https://github.com/RasaHQ/rasa/blob/f30fade97c9fd50b7156aa7d8513bc2fc3527a78/rasa/core/processor.py#L589-L591) is currently not doing that. From my intuition it's fine if the [`NLUPredictionToHistoryAdder`](https://github.com/RasaHQ/rasa/issues/9327) takes care of passing the `input_channel` to the `UserUttered` event.

rasa/graph_components/converters/nlu_message_converter.py
```diff
@@ -0,0 +1,65 @@
...
+            data["message_id"] = message.message_id
+            data["metadata"] = message.metadata
+
+            if message.parse_data:
```
wochinge, line 58: There should never be a `message.parse_data` at this point as the `NLUMessageConverter` comes before the graph section which adds tokens, features, intent / entity predictions. So you always should just have a plain `UserMessage` object which you need to convert to `Message` object which then passes through what used to be the `NLU pipeline"

rasa/graph_components/converters/nlu_message_converter.py
```diff
@@ -0,0 +1,65 @@
...
+    def get_default_config() -> Dict[Text, Any]:
+        """Returns default configuration (see parent class for full docstring)."""
+        return {
+            "remove_duplicates": True,
+            "unique_last_num_states": None,
+            "augmentation_factor": 50,
+            "tracker_limit": None,
+            "use_story_concatenation": True,
+            "debug_plots": False,
+        }
```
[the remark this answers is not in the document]
ancalita: I wasn't actually sure what should be the default config for this component so I was hoping to find out during the review 😄 So it's actually an empty `{ }` and therefore fine to delete these methods?

rasa/graph_components/converters/nlu_message_converter.py
ancalita: if this method can return an empty list, why are we removing the `Optional` part? 🤔 Or does `Optional` allow for list of `None` and that's why it should be removed?

rasa/graph_components/converters/nlu_message_converter.py
ancalita: Good point! I got confused by the fact that `parse_data` is an existing attribute of `UserMessage` - why is this even present if this is at the start of the NLU pipeline?

rasa/graph_components/converters/nlu_message_converter.py
ancalita: I wasn't sure how to check what does `Message.data` contain to be sure if `input_channel` is relevant or not 🤔

rasa/graph_components/converters/nlu_message_converter.py
```diff
@@ -0,0 +1,48 @@
...
+        """Converts user message into Message object.
+
+        Returns:
+            List containing only one instance of Message.
+            Else empty list if user message is None.
+        """
+        if message:
+            data = {
+                TEXT: message.text,
+                "input_channel": message.input_channel,
```
wochinge, line 38: this one can be dropped, no?

tests/test_graph_components/converters/test_nlu_message_converter.py
```diff
@@ -0,0 +1,39 @@
...
+
+    if text:
+        message = UserMessage(text=text)
+        nlu_message = component.convert_user_message(message)
+        assert len(nlu_message) == 1
+        assert isinstance(nlu_message[0], Message)
+
+        assert nlu_message[0].get("text") == "Hi"
+        assert nlu_message[0].get("metadata") is None
+    else:
```
wochinge, line 36: optional: This is in my opinion a separate unit test as it tests for a different aspect (e.g. https://blog.colinbreck.com/in-defence-of-testing-one-thing-per-test-method/)

rasa/graph_components/converters/nlu_message_converter.py
wochinge: seems to be a lazy hack which is only used by this code https://github.com/RasaHQ/rasa/blob/a1afc66ee482955f89bfc4318b4cd3907f8e2378/rasa/server.py#L1038

rasa/graph_components/converters/nlu_message_converter.py
wochinge: I'd check the line in the existing code where `Message` is created. There it doesn't add the `input_channel` to `Message`)

tests/test_graph_components/converters/test_nlu_message_converter.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+    if text:
+        message = UserMessage(text=text)
+        nlu_message = component.convert_user_message(message)
+        assert len(nlu_message) == 1
+        assert isinstance(nlu_message[0], Message)
+
+        assert nlu_message[0].get("text") == "Hi"
+        assert nlu_message[0].get("metadata") is None
```
wochinge, line 35: can we please also have a test with metadata? I know that some community members use that to pass extra information into their NLU components

ancalita: sure, what kind of key-value info goes into metadata?

wochinge: any 🙈 That's completely up to the bot developer. So metadata would be just random dict basically.

rasa/graph_components/converters/nlu_message_converter.py
ancalita: > I'd check the line in the existing code where Message is created. 

Are you referring to this [line](https://github.com/RasaHQ/rasa/blob/7f0bb29b1a40e237931f6f431813e08654bea1fb/rasa/nlu/model.py#L471)? 
Ok, yup, will drop the `input_channel`.

wochinge: yep exactly 👍🏻

wochinge (approved):
💯

PR #9440, opened 2021-08-24
domain without responses provider

### LINKED ISSUE #9333: implement `DomainWithoutResponsesProvider`
**Overview of the Solution**:
We need to implement the `DomainWithoutResponsesProvider `. This provides domain but without responses. Used to avoid retraining policies when the response text changes.

**Input to the component**
- a `Domain`

**Output:** 
* same `Domain` but with nuked `responses` (set to empty list)

**Definition of Done**:

ka-bu:
**Proposed changes**:
- closes #9333 

### DIFF
--- a/rasa/graph_components/providers/domain_without_response_provider.py
+++ b/rasa/graph_components/providers/domain_without_response_provider.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+from typing import Dict, Text, Any
+import copy
+
+from rasa.engine.graph import GraphComponent, ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.shared.core.domain import Domain
+
+
+class DomainWithoutResponseProvider(GraphComponent):
+    """Provides domain without information about responses."""
+
+    @classmethod
+    def create(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> DomainWithoutResponseProvider:
+        """Creates component (see parent class for full docstring)."""
+        return cls()
+
+    def recreate_without_responses(self, domain: Domain) -> Domain:
+        """Recreates the given domain but acts as if responses have not been specified.
+
+        Args:
+            domain: A domain.
+
+        Returns:
+            Domain that has been created from the same parameters as the given domain
+            but with an empty set of responses.
+        """
+        responses = dict()
+        domain = copy.deepcopy(domain)
+        domain.responses = responses
+        domain.user_actions = domain._custom_actions
+        domain.action_names_or_texts = (
+            domain._combine_user_with_default_actions(domain.user_actions)
+            + [
+                form_name
+                for form_name in domain.form_names
+                if form_name not in domain._custom_actions
+            ]
+            + domain.action_texts
+        )
+
+        return domain
--- a/tests/test_graph_components/providers/test_domain_without_responses_provider.py
+++ b/tests/test_graph_components/providers/test_domain_without_responses_provider.py
@@ -1,0 +1,60 @@
+from typing import Text
+
+import pytest
+from rasa.graph_components.providers.domain_without_response_provider import (
+    DomainWithoutResponseProvider,
+)
+from rasa.engine.graph import ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.shared.core.domain import KEY_E2E_ACTIONS, KEY_RESPONSES, Domain
+
+
+@pytest.mark.parametrize(
+    "domain_yml",
+    [
+        ("data/test_domains/travel_form.yml"),
+        ("data/test_domains/mixed_retrieval_intents.yml"),
+    ],
+)
+def test_copy_without_responses(
+    default_model_storage: ModelStorage,
+    default_execution_context: ExecutionContext,
+    domain_yml: Text,
+):
+
+    component = DomainWithoutResponseProvider.create(
+        {"arbitrary-unused": 234},
+        default_model_storage,
+        Resource("xy"),
+        default_execution_context,
+    )
+
+    domain = Domain.from_file(path=domain_yml,)
+    domain_copy = component.recreate_without_responses(domain=domain,)
+
+    given_as_dict = domain.as_dict()
+    copy_as_dict = domain_copy.as_dict()
+
+    # all configurations not impacted by responses stay intact
+    assert sorted(given_as_dict.keys()) == sorted(copy_as_dict.keys())
+    for key in given_as_dict.keys():
+        if key not in [KEY_RESPONSES, KEY_E2E_ACTIONS]:
+            assert given_as_dict[key] == copy_as_dict[key]
+
+    # Note: The given domains do contain responses and some actions for which no
+    # responses are defined...
+    assert domain.responses, "choose a different test config"
+    assert domain.user_actions, "choose a different test config"
+    assert set(domain.user_actions).difference(
+        domain.responses
+    ), "choose a different test config"
+
+    # Assert that the recreated copy does not contain any response information
+    assert not domain_copy.responses
+    assert all(action in domain._custom_actions for action in domain_copy.user_actions)
+    assert set(domain_copy.action_names_or_texts) < set(domain.action_names_or_texts)
+    assert all(
+        response not in domain_copy.action_names_or_texts
+        for response in set(domain.responses).difference(domain._custom_actions)
+    )


tests/test_graph_components/providers/test_domain_without_responses_provider.py
```diff
@@ -0,0 +1,60 @@
...
+):
+
+    component = DomainWithoutResponseProvider.create(
+        {"arbitrary-unused": 234},
+        default_model_storage,
+        Resource("xy"),
+        default_execution_context,
+    )
+
+    domain = Domain.from_file(path=domain_yml,)
```
ancalita, line 33: For better readability I would rename this to `original_domain` and replace `given` occurences with `original`.

tests/test_graph_components/providers/test_domain_without_responses_provider.py
```diff
@@ -0,0 +1,60 @@
...
+    # all configurations not impacted by responses stay intact
+    assert sorted(given_as_dict.keys()) == sorted(copy_as_dict.keys())
+    for key in given_as_dict.keys():
+        if key not in [KEY_RESPONSES, KEY_E2E_ACTIONS]:
+            assert given_as_dict[key] == copy_as_dict[key]
+
+    # Note: The given domains do contain responses and some actions for which no
+    # responses are defined...
+    assert domain.responses, "choose a different test config"
+    assert domain.user_actions, "choose a different test config"
```
ancalita, line 48: I'm not entirely sure what role the `choose a different test config` string plays in the assertion? 🤔

tests/test_graph_components/providers/test_domain_without_responses_provider.py
```diff
@@ -0,0 +1,60 @@
...
+        domain.responses
+    ), "choose a different test config"
+
+    # Assert that the recreated copy does not contain any response information
+    assert not domain_copy.responses
+    assert all(action in domain._custom_actions for action in domain_copy.user_actions)
+    assert set(domain_copy.action_names_or_texts) < set(domain.action_names_or_texts)
+    assert all(
+        response not in domain_copy.action_names_or_texts
+        for response in set(domain.responses).difference(domain._custom_actions)
```
ancalita, line 59: I don't think you need the `difference` method here, since in theory `responses` and `_custom_actions` should be disjoint? Also I'd maybe add a similar assertion to check that `responses` are not in `domain_copy.user_actions`?

ancalita (commented):
Great job in also removing responses from more hidden places like `domain.user_actions` 💯 
I've left a few smallish comments, mainly to improve readability.

tests/test_graph_components/providers/test_domain_without_responses_provider.py
ka-bu: Just in case someone decides to adapt the .ymls we have in the test_domain folder and changes them so that they don't have the nice properties that I would like to use for testing here - then these asserts will function as a reminder that these tests have become outdated

tests/test_graph_components/providers/test_domain_without_responses_provider.py
ka-bu: I think, `_custom_actions` and `responses` could overlap - because to create `user_actions` we de-duplicate the concatenation of both. But maybe I'm overlooking some earlier check here?

ka-bu: > Also I'd maybe add a similar assertion to check that responses are not in domain_copy.user_actions?
good point, will add that

ancalita: > _custom_actions and responses could overlap

According to [this](https://github.com/RasaHQ/rasa/blob/6b26632e7785544cc3d486c2628b832a8c474dc5/rasa/shared/core/domain.py#L217), only custom actions listed in the `actions` section of the domain fill the `action_names` argument  in the constructor which is what `self._custom_actions` gets assigned. But I don't know where else this constructor gets used.

ka-bu: That is a nice and confusing (for me) observation 😅 . 

> But I don't know where else this constructor gets used.

Agree ^^. Then let's leave the ".difference" in there since the `_combine_with_responses` suggests that might happen somehow and the constructor would allow it - even if the methods loading those yamls would fix that (at the moment) for us?

tests/test_graph_components/providers/test_domain_without_responses_provider.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.mark.parametrize(
+    "domain_yml",
+    [
+        ("data/test_domains/travel_form.yml"),
+        ("data/test_domains/mixed_retrieval_intents.yml"),
+    ],
+)
+def test_copy_without_responses(
```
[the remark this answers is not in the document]
ka-bu: done

tests/test_graph_components/providers/test_domain_without_responses_provider.py
ka-bu: also done

tests/test_graph_components/providers/test_domain_without_responses_provider.py
ka-bu: I've made this more verbose/explicit and added a bit on how to resolve it

tests/test_graph_components/providers/test_domain_without_responses_provider.py
ancalita: Yup, all good 👍🏼

rasa/graph_components/providers/domain_without_response_provider.py
```diff
@@ -0,0 +1,49 @@
...
+            domain: A domain.
+
+        Returns:
+            Domain that has been created from the same parameters as the given domain
+            but with an empty set of responses.
+        """
+        responses = dict()
+        domain = copy.deepcopy(domain)
+        domain.responses = responses
+        domain.user_actions = domain._custom_actions
```
ancalita, line 38: This might be nitpicking, but shouldn't protected attributes like `_custom_actions` not be accessible outside the class? The alternative would be the more convoluted way:
```
custom_actions = domain.as_dict.get(KEY_ACTIONS)
```

ka-bu: No, you're right I shouldn't be using this. Same with `_combine_user_with_default_actions`. Maybe the best solution is to add it to the domain 😬 . (And the even better solution would be to let the domain remember which parameters where used to instantiate it, I think, but I won't touch that now 😄 ).

ancalita: Something to add to the ideas doc 👍🏼 In the meantime, wouldn't `set(domain.action_names_or_texts).difference(responses)` work as a way to remove responses without using `_combine_user_with_default_actions`? Would probably need to be bumped a line or two before you assign to `responses` the empty dict.

ka-bu: Yep, same with `user_actions` - but maybe it's really better to just move this to Domain so we don't miss that during the disentangling work (especially because in the current form it is very clearly 1:1 what the initialisation would do) and then we don't need to know/foresee how `action_names_or_texts` and the other things (will) look like ?

ancalita: By moving to Domain, you mean having methods in Domain that disentangle responses from `user_actions` and `action_names_or_texts`?

ka-bu: I mean adding a "def recreate_without_responses(self)"

ancalita: Oh right, yes, that would be the cleanest! @wochinge would this be out of scope for this PR or fine to action it here?

wochinge: yep, I'd prefer that, too 👍🏻 Wouldn't a quick way to dump responses also be to

```python
serialized = domain.as_dict()
del serialized["responses"]
return Domain.from_dict(serialized)
```

ka-bu: Ohhhh, perfect, yes :) Had expected `as_dict` to return all/most attributes as a dict but that doesn't contain any of the funny mixed lists

rasa/graph_components/providers/domain_without_response_provider.py
```diff
@@ -0,0 +1,36 @@
...
+        """Recreates the given domain but acts as if responses have not been specified.
+
+        Args:
+            domain: A domain.
+
+        Returns:
+            Domain that has been created from the same parameters as the given domain
+            but with an empty set of responses.
+        """
+        serliaized_domain = domain.as_dict()
```
ancalita, line 34: tiny typos here `serialised`.

ka-bu: doh, serouisely compliacated wrods :D thanks!

rasa/graph_components/providers/domain_without_response_provider.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
+from typing import Dict, Text, Any
+
+from rasa.engine.graph import GraphComponent, ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.shared.core.domain import KEY_RESPONSES, Domain
+
+
+class DomainWithoutResponseProvider(GraphComponent):
```
wochinge, line 10: nit pick
```suggestion
class DomainWithoutResponsesProvider(GraphComponent):
```

tests/test_graph_components/providers/test_domain_without_responses_provider.py
```diff
@@ -0,0 +1,74 @@
...
+        if key not in [KEY_RESPONSES, KEY_E2E_ACTIONS]:
+            assert original_dict[key] == modified_dict[key]
+
+    # Note: The given domains do contain responses and some actions for which no
+    # responses are defined...
+    reminder = (
+        f"This test needs to be adapted. The domain yamls have been changed so that "
+        f"they no longer contain some actions and some responses - which do not "
+        f"coincide (i.e. not all responses are listed as actions). Hence "
+        f"{domain_yml} is no longer a good test case here. Please Remove the yaml from "
```
wochinge, line 51:
```suggestion
        f"{domain_yml} is no longer a good test case here. Please remove the yaml from "
```

tests/test_graph_components/providers/test_domain_without_responses_provider.py
```diff
@@ -0,0 +1,74 @@
...
+        f"the parameterization or replace it with an different yaml."
+    )
+    assert original_domain.responses, reminder
```
wochinge, line 54: Not sure I'm understanding the `reminder` correctly - why is `{domain_yml}` no longer a good test case?

rasa/graph_components/providers/domain_without_response_provider.py
```diff
@@ -0,0 +1,36 @@
...
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> DomainWithoutResponseProvider:
+        """Creates component (see parent class for full docstring)."""
+        return cls()
+
+    def recreate_without_responses(self, domain: Domain) -> Domain:
```
wochinge, line 24: another nit: can we rename this to `provide` so that all providers have a similar signature? Makes it easier to generate the recipes 🙈

wochinge:
```suggestion
    def provide(self, domain: Domain) -> Domain:
```

tests/test_graph_components/providers/test_domain_without_responses_provider.py
ka-bu: cause I want some responses to only appear in the responses but not listed as actions , otherwise the checks will fail later

ancalita (approved):
🎉

Issue #9329, opened 2021-08-12
implement `ProjectProvider`


wochinge:
**Overview of the Solution**:
We need to implement the `ProjectProvider`. This component receives a path to a domain and a list of training paths. It uses the `TrainingDataImporter` to load this data and then returns the `TrainingDataImporter` . This component only runs during training.

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#24ef56343a554fe88748651864fe9e77) shows the component's position in the graph.

**Input to the component**
- path to a `Domain` (maybe a list is actually better)
- a list of file paths to training data

**Output:** 
* an instantiated `TrainingDataImporter`

**Blockers**
* https://github.com/RasaHQ/rasa/pull/9184

**Definition of Done**:
- [ ] Component is implemented and unit tested
- [ ] Works with `RasaFileImporter` and `MultiprojectImporter`

PR #9441, opened 2021-08-24
implement project provider

twerkmeister:
**Proposed changes**:
- Adresses #9329 
- Minimal version; got some open qs
  - is this the right direction overall? 
  - @wochinge ~ in the issue you wrote "maybe a list [of Domain paths] is actually better" ~ Not sure what should be achieved with this
  - Do I understand the multi importer thing right that I would first need to read in the config and then determine whether the multi importer and the imports key have been defined? And then reinstantiate the importer with MultiProjectImporter? 


rasa/graph_components/providers/project_provider.py
```diff
@@ -0,0 +1,43 @@
...
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> ProjectProvider:
+        """Creates component (see parent class for full docstring)."""
+        return cls(config)
+
+    def provide(self) -> TrainingDataImporter:
+        """Provides the domain during inference."""
```
ancalita, line 38: Don't forget to change the docstring 😄

ancalita (commented):
I think it definitely looks in the right direction 🚀 

Regarding your question on `MultiProjectImporter`, I think that check for `importers` key is done [here](https://github.com/RasaHQ/rasa/blob/6217f1046448a7e8f8e8e3c8487f4b1b8715070e/rasa/shared/importers/importer.py#L145). That should already instantiate the right importer, maybe then the config of component needs to be altered with what's in `imports`🤔? I'm not very confident with how this importer works to help here more.

In addition to your existing test, a test for a custom importer might be worth creating to make sure that still works correctly too.

wochinge:
> wochinge ~ in the issue you wrote "maybe a list [of Domain paths] is actually better" ~ Not sure what should be achieved with this

Rasa actually supports training with multiple domains. However the importer only handles a single file or a single directory which might change in the future. Anyway, handling this out of scope I'd say.

twerkmeister:
Makes sense! Much appreciated @ancalita @wochinge

data/test_multiproject/config.yml
```diff
@@ -0,0 +1,24 @@
...
+    epochs: 100
+  - name: MemoizationPolicy
+  - name: RulePolicy
+
+importers:
+- name: MultiProjectImporter
+
+imports:
+  - projects/ChitchatBot
+  - projects/GreetBot
```
wochinge, line 24:
```suggestion
  - projects/GreetBot

```

data/test_multiproject/projects/ChitchatBot/data/nlu.yml
```diff
@@ -0,0 +1,14 @@
...
+    - are you a bot?
+    - are you a human?
+    - am I talking to a bot?
+    - am I talking to a human?
+- intent: tell_joke
+  examples: |
+    - tell me joke
+    - tell me sth funny
+    - can you tell me a joke?
+    - what's your best joke?
```
wochinge, line 14:
```suggestion
    - what's your best joke?

```

rasa/graph_components/providers/project_provider.py
```diff
@@ -0,0 +1,45 @@
...
+from rasa.shared.constants import (
+    DEFAULT_CONFIG_PATH,
+    DEFAULT_DOMAIN_PATH,
+    DEFAULT_DATA_PATH,
+)
+from rasa.shared.importers.importer import TrainingDataImporter
+
+
+class ProjectProvider(GraphComponent):
+    """Provides domain during training and inference time."""
```
wochinge, line 16:
```suggestion
    """Provides domain and training data during training and inference time."""
```

wochinge (approved):
🚀

Issue #7362, opened 2020-11-25
Optional arguments to rasa data convert are not optional


kaiogu:
**Rasa version**:
2.0.6
**Rasa SDK version** (if used & relevant):

**Rasa X version** (if used & relevant):

**Python version**:
3.8.0
**Operating system** (windows, osx, ...):
ubuntu 18.04
**Issue**:
Arguments stated to be optional by the `rasa data convert xxx` help are not actually optional.

**Error (including full traceback)**:
```
❯ rasa data convert nlu
usage: rasa data convert nlu [-h] [-v] [-vv] [--quiet] -f {json,md,yaml} --data DATA --out OUT [-l LANGUAGE]
rasa data convert nlu: error: the following arguments are required: -f/--format, --data, --out
```
Contrast this with the help of the same command in which these arguments are stated to be optional and to have defaults:
```
❯ rasa data convert nlu --help
usage: rasa data convert nlu [-h] [-v] [-vv] [--quiet] -f {json,md,yaml} --data DATA --out OUT [-l LANGUAGE]

optional arguments:
  -h, --help            show this help message and exit
  -f {json,md,yaml}, --format {json,md,yaml}
                        Output format the training data should be converted into. Note: currently training data can be
                        converted to 'yaml' format only from 'md' format (default: yaml)
  --data DATA           Path to the file or directory containing Rasa NLU data. (default: data)
  --out OUT             File (for `json` and `md`) or existing path (for `yaml`) where to save training data in Rasa
                        format. (default: converted_data)
  -l LANGUAGE, --language LANGUAGE
                        Language of data. (default: en)
```

**Command or request that led to error**:
```
❯ rasa data convert nlu
❯ rasa data convert nlg
❯ rasa data convert core
```

**Content of configuration file (config.yml)** (if relevant):
```yml

```

**Content of domain file (domain.yml)** (if relevant):
```yml

```

**Definition of Done**
For all `rasa data convert` commands:
- [ ] set `required=False` (or don't set it) if a default was provided

sara-tagger:
Thanks for the issue, @amn41 will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

wochinge:
The issue is that we mark things as `required` but give them defaults at the same type. Argparse treats everything with a default as `required=False` as optional but then still fails if not provided 🤦🏻

PR #9458, opened 2021-08-25
Fix non-optional optional args on rasa data convert sub-commands

usc-m:
**Proposed changes**:
- Addresses #7362 
- `argparse` interprets `required` as "this argument must be set by user input", and we would like it to mean "this argument must have a value set (by default or user input)", so adapt how we pass in a value for `required` on this basis
-- We could possibly just drop `required` on our wrapper function entirely as it seems to be used in exactly one place to set it `True` and a default is also specified at the same time. Is that preferable?

- [] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)


rasa/cli/arguments/default_arguments.py
```diff
@@ -109,7 +109,10 @@ def add_data_param(
         nargs="+",
         type=str,
         help=f"Paths to the files or directories containing {data_type} data.",
-        required=required,
+        # The desired behaviour is that required indicates if this argument must
+        # have a value, but argparse interprets it as "must have a value
+        # from user input", so we toggle it only if our default is not set
+        required=required and default is None,
```
wochinge, line 115: I think we have the same problem for `--out OUT` as well, right?

wochinge (commented):
Congrats to your first Rasa PR 🚀 🚀 🚀 Glad you to have you onboard!

rasa/cli/arguments/default_arguments.py
usc-m: Good catch, looks like it has the same logic but isn't actively an issue because nothing is setting `required` to `True` - I'll update in the same way

usc-m:
That should be those changes mirrored onto `--out` - has no effect on tests since it doesn't change the output, but may help us catch errors in future

PR #9470, opened 2021-08-27
Adapt `SklearnIntentClassifier` to `GraphComponent` interface

### LINKED ISSUE #9406: Adapt `SklearnIntentClassifier` to new `GraphComponent` interface
**Overview of the Solution**:
Adapt the `SklearnIntentClassifier ` to the new `GraphComponent` interface.

* adapt persistence

**Definition of Done**:
- [x] `SklearnIntentClassifier` adapted to the new GraphComponent interface
- [x] Unit tests are adapted to adapted component
- [x] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old component implementation)

ancalita:
**Proposed changes**:
- closes #9406 


rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -1,72 +1,92 @@
...
         le: Optional["sklearn.preprocessing.LabelEncoder"] = None,
     ) -> None:
         """Construct a new intent classifier using the sklearn framework."""
         from sklearn.preprocessing import LabelEncoder
 
-        super().__init__(component_config)
+        self.component_config = config
+        self._model_storage = model_storage
+        self._resource = resource
+        self._execution_context = execution_context
```
wochinge, line 68: do we actually need to save this one?

rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -83,14 +103,8 @@ def transform_labels_num2str(self, y: np.ndarray) -> np.ndarray:
 
         return self.le.inverse_transform(y)
 
-    def train(
-        self,
-        training_data: TrainingData,
-        config: Optional[RasaNLUModelConfig] = None,
-        **kwargs: Any,
-    ) -> None:
+    def train(self, training_data: TrainingData, **kwargs: Any,) -> Resource:
```
wochinge, line 106: let's pass in `num_threads` via the config and drop the `kwargs`. Seems cleaner to me.

rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -225,40 +244,54 @@ def predict(self, X: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
-            return cls(meta, classifier, encoder)
-        else:
-            return cls(meta)
+        try:
+            with model_storage.read_from(resource) as model_dir:
+                file_name = cls.__name__
+                classifier_file = list(model_dir.glob(f"**/{file_name}_classifier.pkl"))
+
+                if classifier_file:
+                    assert len(classifier_file) == 1
```
wochinge, line 276: let's not have an `assert` in production code. How about raising an exception which is then caught by the block in 292?

tests/nlu/classifiers/test_sklearn_classifier.py
```diff
@@ -0,0 +1,72 @@
...
+from rasa.nlu import registry
+from rasa.nlu.classifiers.sklearn_intent_classifier import (
+    SklearnIntentClassifierGraphComponent,
+)
+from rasa.nlu.config import RasaNLUModelConfig
+from rasa.shared.nlu.training_data.training_data import TrainingData
+
+
+@pytest.fixture()
+def training_data(nlu_data_path: Text):
```
wochinge, line 19:
```suggestion
def training_data(nlu_data_path: Text) -> TrainingData:
```

tests/nlu/classifiers/test_sklearn_classifier.py
```diff
@@ -0,0 +1,72 @@
...
+):
+    return SklearnIntentClassifierGraphComponent.create(
+        SklearnIntentClassifierGraphComponent.get_default_config(),
+        default_model_storage,
+        Resource("sklearn"),
+        default_execution_context,
+    )
+
+
+def test_persist_and_load(
```
wochinge, line 35: can we add a test where loading the component from the model storage fails and we initialize an empty `SklearnClassifier`?

rasa/nlu/classifiers/sklearn_intent_classifier.py
ancalita: True, `execution_context` is not required, I'll remove it; the others are needed.

rasa/nlu/classifiers/sklearn_intent_classifier.py
ancalita: Sure, what would be the exception? 
Here my intention was to make sure that `len` is not greater than 1. Perhaps I can raise a debug warning that more than 1 file were found? In theory it shouldn't be more than 1 file, maybe I'm overcomplicating matters.

wochinge: how about just throwing a `ValueError`? Your `except ValueError` would then catch it.

ancalita: Thinking more about this approach - it would return a new component instance when actually files were found that could be used to load the component. Maybe I'm imagining a scenario that actually would never happen (e.g. have more than 1 file found) 🤔 I will just remove the assertion, WDYT?

ka-bu: were not testing this anywhere else - we're always relying on the assumption that only persist() writes to that directory - so I think it's ok to remove the check for "== 1" :)  ... we could also create some constants for classifier_file_name and encoder_file_name and just use those here and in persist (because we also rely on the model storage/resource to be a unique location for this component evyerwhere)

rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -165,49 +182,51 @@ def _create_classifier(
...
 
-    def process(self, message: Message, **kwargs: Any) -> None:
+    def process(self, messages: List[Message]) -> List[Message]:
         """Return the most likely intent and its probability for a message."""
+        for message in messages:
+            if not self.clf:
+                # component is either not trained or didn't
+                # receive enough training data
+                intent = None
+                intent_ranking = []
```
wochinge, line 192: should we pull this out of the loop?

rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -225,40 +244,53 @@ def predict(self, X: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
-            classes = io_utils.json_unpickle(encoder_file)
-            encoder = LabelEncoder()
-            encoder.classes_ = classes
-            return cls(meta, classifier, encoder)
-        else:
-            return cls(meta)
+        try:
+            with model_storage.read_from(resource) as model_dir:
+                file_name = cls.__name__
+                classifier_file = list(model_dir.glob(f"**/{file_name}_classifier.pkl"))
```
wochinge, line 273: why do we actually need to glob this at all? Don't we exactly know where to load it from as we explicitly hardcode the file names?

tests/nlu/classifiers/test_sklearn_classifier.py
```diff
@@ -0,0 +1,94 @@
...
+    default_sklearn_intent_classifier.train(training_data)
+
+    loaded = SklearnIntentClassifierGraphComponent.load(
+        SklearnIntentClassifierGraphComponent.get_default_config(),
+        default_model_storage,
+        Resource("sklearn"),
+        default_execution_context,
+    )
+
+    predicted = copy.copy(training_data)
```
wochinge, line 66: I think a `copy` is not enough - you'll need a `deepcopy`

rasa/nlu/classifiers/sklearn_intent_classifier.py
ancalita: And then return the list with unmodified messages?

rasa/nlu/classifiers/sklearn_intent_classifier.py
ancalita: Oh right - would it be enough to just replace this with `model_dir / f"{file_name}_classifier.pkl"` ?

rasa/nlu/classifiers/sklearn_intent_classifier.py
wochinge: mhm, actually that makes sense 👍🏻 Let's keep it in then

rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -1,72 +1,92 @@
...
+            "num_threads": 1,
+        }
 
     def __init__(
         self,
-        component_config: Optional[Dict[Text, Any]] = None,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
```
wochinge, line 59:
```suggestion
```

rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -225,40 +244,53 @@ def predict(self, X: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
-            encoder.classes_ = classes
-            return cls(meta, classifier, encoder)
-        else:
-            return cls(meta)
+        try:
+            with model_storage.read_from(resource) as model_dir:
+                file_name = cls.__name__
+                classifier_file = model_dir / f"{file_name}_classifier.pkl"
+
+                if classifier_file:
```
wochinge, line 275:
```suggestion
                if classifier_file.exists():
```

rasa/nlu/classifiers/sklearn_intent_classifier.py
```diff
@@ -225,40 +244,53 @@ def predict(self, X: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
...
+                        config,
+                        model_storage,
+                        resource,
+                        execution_context,
+                        classifier,
+                        encoder,
+                    )
+        except ValueError:
+            logger.warning(
+                f"Failed to load {cls.__name__} from model storage. Resource "
```
wochinge, line 293:
```suggestion
                f"Failed to load '{cls.__name__}' from model storage. Resource "
```

wochinge (approved):
Left a few more comment but looks great otherwise 🚀

Issue #9330, opened 2021-08-12
implement `NLUTrainingDataProvider`


wochinge:
**Overview of the Solution**:
We need to implement the `NLUTrainingDataProvider`. This component receives an instantiated `TrainingDataImporter` and returns the result of `importer.get_nlu_data(language)` 

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#24ef56343a554fe88748651864fe9e77) shows the component's position in the graph.

**Input to the component**
- `TrainingDataImporter`

**Component Config**
- the language
- whether it persists `True` / `False`

**Output:** 
* a `TrainingData` object

**Blockers**
* https://github.com/RasaHQ/rasa/pull/9184

**Definition of Done**:
- [ ] Component is implemented and unit tested

PR #9474, opened 2021-08-30
3.0 architecture revamp/9330/nlu training data provider

aeshky:
**Proposed changes**:
- implemented NLUTrainingDataProvider according to #9330 


rasa/graph_components/providers/nlu_training_data_provider.py
```diff
@@ -0,0 +1,87 @@
...
+    @classmethod
+    def create(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> NLUTrainingDataProvider:
+        """Creates component (see parent class for full docstring)."""
+        return cls(model_storage, resource)
```
aeshky, line 42: Is it okay that we have unused arguments in some of the class methods (create and load)? Do we plan to extend the class and use them? Or are they there to maintaini a consistent interface regardless?

rasa/graph_components/providers/nlu_training_data_provider.py
```diff
@@ -0,0 +1,87 @@
...
+        **kwargs: Any,
+    ) -> NLUTrainingDataProvider:
+        """Creates provider using a persisted data."""
+        with model_storage.read_from(resource) as resource_directory:
+            training_data = load_data(
+                resource_name=os.path.join(
+                    str(resource_directory), DEFAULT_TRAINING_DATA_OUTPUT_PATH
+                ),
+                language=config["language"],
+            )
```
aeshky, line 60: I modelled my implementation on `DomainProvider`. However, unlike the `Domain` class `TrainingData` did not have a `from_file` method. Did I use the correct nlu training data loader?

rasa/graph_components/providers/nlu_training_data_provider.py
```diff
@@ -0,0 +1,87 @@
...
+        return cls(model_storage, resource, training_data)
+
+    def _persist(self, training_data: TrainingData) -> None:
+        """Persists NLU training data to model storage."""
+        with self._model_storage.write_to(self._resource) as resource_directory:
+            training_data.persist(
+                dir_name=str(resource_directory),
+                filename=DEFAULT_TRAINING_DATA_OUTPUT_PATH,
+            )
```
aeshky, line 69: I used the constant `DEFAULT_TRAINING_DATA_OUTPUT_PATH` from `training_data.py` instead of hardcoding the nlu file name the way `DomainProvider` [does](https://github.com/RasaHQ/rasa/blob/33267a951f924ed17250a8e3b1c4f136b1670c16/rasa/graph_components/providers/domain_provider.py#L53). Is this okay, or should I hardcode it to be consistent?

I noticed that the interface for [`persist` in `TrainingData`](https://github.com/RasaHQ/rasa/blob/33267a951f924ed17250a8e3b1c4f136b1670c16/rasa/shared/nlu/training_data/training_data.py#L431) is differnt to [the one for `Domain`](https://github.com/RasaHQ/rasa/blob/33267a951f924ed17250a8e3b1c4f136b1670c16/rasa/shared/core/domain.py#L1436). Is this okay?

tests/test_graph_components/providers/test_nlu_training_data_provider.py
```diff
@@ -0,0 +1,56 @@
...
+    # get data and persist it using config
+    data_2 = provider_1.provide(config_2, importer)
+
+    # load the provider
+    provider_2 = NLUTrainingDataProvider.load(
+        config_2, default_model_storage, resource, default_execution_context
+    )
+    data_3 = provider_2.provide(config_1)
+
+    assert data_2.fingerprint() != data_3.fingerprint()
```
aeshky, line 56: This assertion is supposed to be `==` but that's failing. After I persist the data then load it, the actions are missing. Am I missing something obvious? Or maybe I'm using the data loader incorrectly?

rasa/graph_components/providers/nlu_training_data_provider.py
```diff
@@ -0,0 +1,84 @@
...
+    def provide(
+        self, config: Dict[Text, Any], importer: Optional[TrainingDataImporter] = None,
+    ) -> TrainingData:
+        """Provides nlu training data during training."""
+        if config == {}:  # use default config if config is empty
+            config = self.get_default_config()
+        if importer is None:
+            training_data = self._training_data
+        else:
+            training_data = importer.get_nlu_data(language=config["language"])
```
aeshky, line 81: I wanted `self._training_data` to be accessible, so I made `importer` optional, and return `self._training_data` when no importer is passed. Not sure if this is the desired behaviour or if I should do something different.

rasa/graph_components/providers/nlu_training_data_provider.py
```diff
@@ -0,0 +1,84 @@
...
+    ) -> None:
+        """Creates NLU training data provider."""
+        self._model_storage = model_storage
+        self._resource = resource
+        self._training_data = training_data
+
+    @classmethod
+    def get_default_config(cls) -> Dict[Text, Any]:
+        """Get Default class configuration"""
+        return {"language": "en", "persist": False}
```
aeshky, line 32: I put `language` and `persist` in the configuration instead of passing them as arguments to the methods. Is this what we want?

rasa/graph_components/providers/nlu_training_data_provider.py
ka-bu: > are they there to maintaini a consistent interface regardless?

Yep, exactly :) This makes it easy to create every graph component in exactly the same way. If a component isn't configurable or doesn't need to persist anything (because nothing needs to be "trained"), then they can just leave those parameters unused (e.g. the `graph_component/converter/nlu_message_converter.py` doesn't need anything).

(Although I think in the future this will change a little - at least the "config" should be always passed on to enable checks - i.e. if the component can't be configured, we shouldn't be able to attempt to configure it without raising a warning)

rasa/graph_components/providers/nlu_training_data_provider.py
ka-bu: yep, that matches what Tobias requested. The persist option was a bit confusing for me at first - but I think it is here because importing the nlu data from disk can be a bit more costly than it looks (because we might have to parse the stories and extract the nlu data from there (cf. [importers here](https://github.com/RasaHQ/rasa/blob/2e82c9caab989348d57e74d918cc4286f85778a5/rasa/shared/importers/importer.py#L482)))

ka-bu: and yep, the component should store those parameters in it's internal "config" and then use that config wherever needed (which makes sense for all parameters that e.g. don't depend on other components or on the training of other components - in contrast, the training data that is provided by this component will be expected as parameter in the method of another component but will not be passed as a config parameter)

rasa/graph_components/providers/nlu_training_data_provider.py
ka-bu: > hardcoding the nlu file name the way DomainProvider does. Is this okay, or should I hardcode it to be consistent?
that's ok  - the names don't really matter (they only need to be unique within a "resource" ^^)

> interface ... is different
that is ok as well - domain and training data don't have to implement the method in the same way (if there was a common "persistable" interface that they implemented, then they should look the same) 

(@joejuzl , we could also create a cacheable version of TrainingData and let the caching mechanism take care of persisting, but that would mean that we cannot make persisting optional as we can via this approach here where we can switch that on/off via the config... 🤔  Which is not what we want, right?)

rasa/graph_components/providers/nlu_training_data_provider.py
ka-bu: ... so the config needs to be passed on to the constructor here!

rasa/graph_components/providers/nlu_training_data_provider.py
```diff
@@ -0,0 +1,87 @@
...
+                filename=DEFAULT_TRAINING_DATA_OUTPUT_PATH,
+            )
+
+    def provide(
+        self, config: Dict[Text, Any], importer: Optional[TrainingDataImporter] = None,
+    ) -> TrainingData:
+        """Provides nlu training data during training."""
+
+        if config == {}:  # use default config if config is empty
+            config = self.get_default_config()
```
ka-bu, line 77: You can move this to the constructor and there use something like `self._config = {**self.default_config(), **config}` which will fill the `self._config` with the default parameters first and then overwrite default values with the ones from the `config` passed to the constructor

rasa/graph_components/providers/nlu_training_data_provider.py
ka-bu: I think the other components (will) might always pass on an importer. But if we constructed this instance (self) via the `load` method, then there will be `self._training_data` that can be passed on. And if we didn't create this instance via `load`, then `self._training_data` will be `None`. So you can use the information whether `self._training_data` is None to decide whether the importer has to be used.

Downside of that approach is that if the data on disk changes, then a training data provider loaded from disk won't notice those differences (it will load the version of training data persisted elsewhere and not load the new data).

.... @joejuzl ?

tests/test_graph_components/providers/test_nlu_training_data_provider.py
```diff
@@ -0,0 +1,56 @@
...
+    provider_1 = NLUTrainingDataProvider.create(
+        NLUTrainingDataProvider.get_default_config(),
+        default_model_storage,
+        resource,
+        default_execution_context,
+    )
+    assert isinstance(provider_1, NLUTrainingDataProvider)
+
+    # check the data provided is as expected
+    data_0 = provider_1.provide({}, importer)
```
ka-bu, line 39: I think if you'd add the "language" to the first parameter (config) here, then both results should be the same

ka-bu: ... oh, thought this line was failing, but it is the assert below :D strange..

ka-bu: ... is "en" the default somewhere? 🤔

tests/test_graph_components/providers/test_nlu_training_data_provider.py
ka-bu: Think it should be `data_3 = provider_2.provide(config_**2**)` ... but if that fixes it the above should've also failed 😅

rasa/graph_components/providers/nlu_training_data_provider.py
ka-bu: The `GraphNode` which which wraps/instantiates the `GraphComponent` will take care of this for us :) (see: [here](https://github.com/RasaHQ/rasa/blob/5bb331ff7ba75ad564478cbc71fd1a46bce5ee02/rasa/engine/graph.py#L315))

rasa/graph_components/providers/nlu_training_data_provider.py
```diff
@@ -0,0 +1,84 @@
...
+from rasa.shared.importers.importer import TrainingDataImporter
+from rasa.shared.nlu.training_data.training_data import (
+    TrainingData,
+    DEFAULT_TRAINING_DATA_OUTPUT_PATH,
+)
+from rasa.shared.nlu.training_data.loading import load_data
+
+
+class NLUTrainingDataProvider(GraphComponent):
+    """Provides NLU training data during training and inference time."""
```
joejuzl, line 16: This will only be used during training.

rasa/graph_components/providers/nlu_training_data_provider.py
joejuzl: The reason persist is in the config is because it is controlled by a cli flag `--persist-nlu-data`. So it will change the config of the component, rather than come from a different component.

Is it's true it essentially means we bundle the nlu training data in with the model data in the model archive.

joejuzl: I wonder if we should have a default language though 🤔

rasa/graph_components/providers/nlu_training_data_provider.py
joejuzl: So this is a weird one... but we don't actually need a `load` for this component!
We only persist to keep the legacy functionality of `--persist-nlu-data` which saves the nlu data with the model. We would never actually load this data.

rasa/graph_components/providers/nlu_training_data_provider.py
joejuzl: We only persist so that the data is in the model package at the end. I think any caching would be separate from this.

rasa/graph_components/providers/nlu_training_data_provider.py
joejuzl: As we are not going to load the data, I don't think we need to even have `self._training_data`.

joejuzl (changes requested):

rasa/graph_components/providers/nlu_training_data_provider.py
ka-bu: So we just need to persist during "provide()" and not think about anything else, right? (Cause users could just switch of that flag to persist that data again and again)

aeshky: > I wonder if we should have a default language though 🤔

I was reluctant about putting english as default. I can remove it.

tests/test_graph_components/providers/test_nlu_training_data_provider.py
aeshky: It was. I changed it.

tests/test_graph_components/providers/test_nlu_training_data_provider.py
aeshky: I think the language was not the issue. I think the loading function isn't loading the whole data 🤔 I changed the test to check that I can load the persisted data and that it is an instace of TrainingData. Fingerprint still not matching though.

tests/test_graph_components/providers/test_nlu_training_data_provider.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+
+    # check the default configuration is as expected
+    config_1 = NLUTrainingDataProvider.get_default_config()
+    assert "language" not in config_1
+    assert config_1["persist"] is False
+
+    # create a provider
+    provider_1 = NLUTrainingDataProvider.create(
+        {"language": "en", "persist": True},
```
joejuzl, line 36: Can we also test when `"persist"` is `False`?

joejuzl (changes requested):
Looking good! 
Just one comment on the tests.

tests/test_graph_components/providers/test_nlu_training_data_provider.py
```diff
@@ -51,4 +51,24 @@ def test_nlu_training_data_provider(
...
+
+    # create a provider with persist == False
+    NLUTrainingDataProvider.create(
+        {"language": "en", "persist": False},
+        default_model_storage,
+        resource,
+        default_execution_context,
+    )
+
+    # check the data was not persisted
```
joejuzl, line 69: Don't we need to call `provide` to check if this really isn't persisting?
Maybe we could paramaterize the test?

joejuzl (changes requested):

tests/test_graph_components/providers/test_nlu_training_data_provider.py
joejuzl: bump

aeshky: Oh yeah you're right.

aeshky: Okay, now fixed!

joejuzl (approved):

aeshky:
Thanks @joejuzl!

PR #9475, opened 2021-08-30
Fix #9302: Featurised slots sometimes appear in general rules

### LINKED ISSUE #9302: Rule policies will not fire if unrelated featurized slots are set
_Product feedback submission from @rgstephens_

**Who is the feedback from?**

Multiple customers

**What is the feedback?**

Rule policies will not fire if unrelated featurized slots are set. This behavior is counterintuitive and not documented.

**What problem are they trying to solve?**

A simple rule like this should work if a featurized slot is set:

```yml
- rule: greet
 steps:
 - intent: greet
 - action: utter_greet
```

[Repo with more details](https://github.com/rgstephens/rasa-rule-slot-issue)

[Slack thread](https://rasa-hq.slack.com/archives/C0148JNFXLG/p1628271530006800)

**Recommendation**
- Assign two researcher to this issue

**Possible Inelegant Workaround**
- Add a rule for every possible slot state

**Possible Solutions**
- Make the state of slot featurization explicit in the rules (e.g. none or any)

**Definition of Done**
In order of urgency:
- [ ] See whether suggested workaround will resolve immediate customer problem
- [ ] Understand customer context: who does this affect? 
- [ ] Identify consequences: what would change if this suggested behavior change were implemented?
- [ ] Identify possible alternative next steps to be refined and scoped later, likely to include more explicit definition of rule behavior

Instead of the above steps, @samsucik carried out an investigation to pin-point the bug and, while doing so, proposed a simple fix (implemented in #9475), which has been verified to fix the issue on all examples provided by @rgstephens. The discussion inspired [related improvements to our docs](https://github.com/RasaHQ/rasa/issues/9521) and [another bug report](https://github.com/RasaHQ/rasa/issues/9483).

@samsucik to coordinate who is doing what on this issue
@kedz assigned reviewer

samsucik:
Over in #9302, I identified a bug in the `CategoricalSlot` featurisation code: If the slot `influences_conversation` but is currently unset (i.e. set to `None`), it can still get featurised as if it was set to an actual value. This happens when one of the allowed values for the slot is `"none"` (or `"None"`, or `"nOnE"`, etc.) due to [this line](https://github.com/RasaHQ/rasa/blob/0d97d427fd342b3f0fab4eaabc3d5169249dbf61/rasa/shared/core/slots.py#L372).

**Proposed changes**:
I'm proposing a fix: To treat unset slots in the abovementioned case the same way as we treat categorical slots for which featurising the slot fails. For those cases, we return an all-zero array as the slot's featurisation. Note that, in the future, we might want to distinguish between these cases and return different features for each.

In order to implement the fix, I need to disallow the Pythonic `None` as a valid value for categorical slots. Currently, one can list `null` in `domain.yml` among the possible values for such a slot. That `null` gets translated into `"none"` within `Domain`'s Python code. Which, in turn, during the featurisation of the slot, matches the cases where the slot's value is the Pythonic `None` (i.e. the slot is unset). In this sense, we are allowing people to use the special `unset` value (represented by `null`/`None`) as a regular value for categorical slots. I think this isn't good.

I was considering returning an empty feature array (`[]`) but then decided not to do this. We return this value in `Slot.as_feature` for slots that don't influence the conversation, but here we're talking about _conversation-influencing yet unset_ slots.

I also concluded that we need to talk in the docs more about unset slots. Or at least to collect all relevant info in one place, which I'm trying to do here (to be actually put into our docs _in a separate issue_):
- a slot is originally unset (unless it has `initial_value` defined)
- this means its value is the Pythonic `None`
- where such a slot `influences_conversation`, models know that it's unset (i.e. its featurisation differs from when it's set to some value)
- in Python (custom actions), we unset a slot by setting it to `None`; in stories/rules, we write `slot_name: null`
- categorical slots can't have the Pythonic `None` as a possible legit value. `None` (in `domain.yml` written as `null`) is reserved for when the slot is unset

In addition to unset slots, we also need to make it clear in the docs that **featurisation of categorical slots is case-insensitive** and involves casting to `str` (and we might re-think these design decisions in the future).

One more related issue to be clearly communicated in the docs is slot-agnostic vs slot-dependent behaviour: If, for some given context, a rule prescribes certain next steps irrespective of slots, and another rule (or a story) for the same context specifies different behaviour while conditioning on slots, then these two rules/stories are in conflict. A good example of this is reported [here](https://github.com/RasaHQ/rasa/issues/9302#issuecomment-900873505) and elaborated on [here](https://github.com/RasaHQ/rasa/issues/9302#issuecomment-907291732).


kedz:
> In addition to unset slots, we also need to make it clear in the docs that featurisation of categorical slots is case-insensitive (and we might re-think this design decision in the future).

Also possibly worth mentioning that the value is also cast to a str type before lower casing and testing for str equality.

samsucik:
> Also possibly worth mentioning that the value is also cast to a str type before lower casing and testing for str equality.

Yeah, that makes sense. I mean, the way we do the value checking isn't exactly obvious, so making it explicit can't hurt.

samsucik:
I've now added tests and checked that all of them fail with the fix disabled and pass with the fix included. @kedz the last bit I'll add before a review is a changelog entry 🙂

samsucik:
@alwx fyi I've been discussing bits of this fix with Joe [over here](https://rasa-hq.slack.com/archives/C01HHMR4X8S/p1630335253041400).

rasa/shared/core/slots.py
```diff
@@ -343,9 +343,21 @@ def __init__(
...
             name, initial_value, value_reset_delay, auto_fill, influence_conversation
         )
-        self.values = [str(v).lower() for v in values] if values else []
+        if values and None in values:
+            rasa.shared.utils.io.raise_warning(
+                f"Categorical slot '{self.name}' has `null` listed as a possible value,"
+                f" which translates to `None` in Python. This value is reserved for"
+                f" when a slot is not set and should not be used as a normal value."
+                f" Rasa will proceed, not treating `null` as a possible normal value"
+                f" for the '{self.name}' slot.",
```
kedz, line 352:
```suggestion
                f"Categorical slot '{self.name}' has `null` listed as a possible value,"
                f" which translates to `None` in Python. This value is reserved for"
                f" when a slot is not set and should not be specified as a value in the slot definition."
                f" Rasa will ignore `null` as a possible value for the '{self.name}' slot."
                f" Consider renaming this value, for example, `unset`, or explicitly specifying "
                f" the string by using quotation marks, `\"null\"`.",
```

`normal value` is not defined anywhere and might be confusing to the average user. I reworded to avoid use of that phrase. I also added a suggestion at the end for how they could fix this warning.

kedz (approved):
Looks good! Added a minor comment about wording of a warning message.

alwx (approved):
The fix looks good for me. Thanks!

Issue #9336, opened 2021-08-12
implement `ConfigValidator`


wochinge:
**Overview of the Solution**:
We need to implement the `ConfigValidator `. This component receives the current training data as `TrainingDataImporter` and validates it against the current train graph schema (accessible via `ExecutionContext`). Implements the logic from [here](https://github.com/RasaHQ/rasa/blob/e0a6a4a0829f510fd33657c918192ebe3f5ba06e/rasa/nlu/components.py#L245).  

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#24ef56343a554fe88748651864fe9e77) shows the component's position in the graph.

**Update**:
- this should also check [this](https://github.com/RasaHQ/rasa/blob/9fdbca3a1e7b408fa2999d8e394520185c7a0fe0/rasa/core/policies/ensemble.py#L118)
- and [this](https://github.com/RasaHQ/rasa/blob/9fdbca3a1e7b408fa2999d8e394520185c7a0fe0/rasa/core/policies/ensemble.py#L247)
- consider checking all graph components which have a `validate_against_domain` to validate them against the domain (see [here](https://github.com/RasaHQ/rasa/pull/9576/files#r705073921))

**Input to the component**
- `TrainingDataImporter`

**Output:** 
* `none`
* raises exception in case config doesn't match training data

**Blockers**
* https://github.com/RasaHQ/rasa/pull/9184

**Definition of Done**:
- [ ] Component is implemented and unit tested

ErickGiffoni:
Hello ! Is there anyone working on this issue ?

I’m with 2 other developers here and we’d like to contribute, if possible!

wochinge:
Hi @ErickGiffoni . Thanks for your initiative 🙌🏻 

This issue is part of an architecture revamp which we are currently undertaking with relatively tight deadlines. I'd suggest that you'd rather take one of the issues from the [contributor board](https://github.com/orgs/RasaHQ/projects/23).

ErickGiffoni:
> Hi @ErickGiffoni . Thanks for your initiative 🙌🏻
> 
> This issue is part of an architecture revamp which we are currently undertaking with relatively tight deadlines. I'd suggest that you'd rather take one of the issues from the [contributor board](https://github.com/orgs/RasaHQ/projects/23).

Thanks for your quick feedback!

[...]

PR #9481, opened 2021-08-31
ensemble graph component

### LINKED ISSUE #9346: Create `GraphPolicyEnsemble` graph component
**Overview of the Solution**:
We need to implement the `GraphPolicyEnsemble` (maybe a different better name?!). This component runs during inferences and selects the winning prediction from all the policy components.
Most of the functionality can come from the current `SimplePolicyEnsemble`.

- Leave the old `PolicyEnsemble` and `SimplePolicyEnsemble` for now as they will be required for the current tests to work.
- The new component `GraphPolicyEnsemble` will use a lot of the functionality of `SimplePolicyEnsemble` but almost none of what is in `PolicyEnsemble` (the superclass).
- The functionality needs to be adapted to receive the inputs directly into the run function (which will be similar to `probabilities_using_best_policy`).
- The class needs to inherit from `GraphComponent` and implement all required methods.
- `GraphPolicyEnsemble` should be unit tested - lots of this can be adapted from `test_ensemble.py`, however the tests will need to be changed to pass predictions, rather than policies into the ensemble.

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#48704bc28dab486dad98e9ebd3ab9077) shows the component's position in the graph.

**Input to the component**
A variable number of `PolicyPrediction`s from parent policy nodes. Will need to be captured by `**kwargs`.

**Output:**
The winning `PolicyPrediction`

**Definition of Done**:

ka-bu:
**Proposed changes**:
- closes #9346 


ka-bu:
@joejuzl ,
-  the `ensemble.py` contains the list of things that I think we might need to add and how
- any idea why some of the integration tests seem to have a problem with `RulePolicy` (and others don't)?

@JEM-Mosig 
- could you please have a look at the new `DefaultPolicyPredictionEnsemble` and the tests for that?

many thanks :)

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,315 @@ def _check_for_important_policies(self) -> None:
...
-        # because some of them might add events
-        rejected_action_name = None
-        last_action_event = next(
-            (
-                event
-                for event in reversed(tracker.events)
-                if isinstance(event, (ActionExecutionRejected, ActionExecuted))
-            ),
-            None,
+        does_not_contain_no_user_predictions = all(
```
wochinge, line 476: can we please remove this double negation? Makes the code in my opinion hard to understand

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,315 @@ def _check_for_important_policies(self) -> None:
...
-                f"Execution of '{rejected_action_name}' was rejected. "
-                f"Setting its confidence to 0.0 in all predictions."
-            )
-            for prediction in predictions.values():
-                prediction.probabilities[
-                    domain.index_for_action(rejected_action_name)
-                ] = 0.0
-
-        return self._pick_best_policy(predictions)
+        return max(
```
wochinge, line 490: somewhat neat but not sure if more readable 🤔

tests/core/test_ensemble.py
```diff
@@ -1,566 +1,212 @@
...
-    prediction = policy_ensemble_0.probabilities_using_best_policy(
-        tracker, domain, RegexInterpreter()
-    )
-    assert prediction.policy_name == "policy_{}_{}".format(i, type(priority_2).__name__)
-    assert prediction.probabilities == priority_2_result.probabilities
 
-    i = 0  # index of priority_2 in ensemble_1
-    prediction = policy_ensemble_1.probabilities_using_best_policy(
-        tracker, domain, RegexInterpreter()
+def test_default_predict__excludes_rejected_action(
```
wochinge, line 68:
```suggestion
def test_default_predict_excludes_rejected_action(
```

tests/core/test_ensemble.py
```diff
@@ -1,566 +1,212 @@
...
-
-    tracker = DialogueStateTracker.from_events(
-        "rule triggering tracker",
-        evts=[
-            ActionExecuted(ACTION_LISTEN_NAME),
-            UserUttered(text="hello", intent={"name": "greet"}),
-            ActionExecuted(ACTION_UNLIKELY_INTENT_NAME),
-        ],
+    # expected events
+    expected_events = set(
```
wochinge, line 192: this is somewhat tautological testing, isn't it? How about hardcoding the expected events instead of computing them in the test?

rasa/core/policies/ensemble.py
wochinge: maybe `contains_no_form_predictions`?

rasa/core/policies/_policy.py
```diff
@@ -428,3 +428,139 @@ def format_tracker_states(states: List[Dict]) -> Text:
                     formatted_states.append(state_formatted)
 
         return "\n".join(formatted_states)
+
+
+class PolicyPrediction:
```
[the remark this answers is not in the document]
ka-bu: it's a copy of the previous policy prediction - in the new one (`PolicyPrediction2`), priorities are just ints (not tuples containing 1 int) and the max. probability is not called "max. confidence"

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,315 @@ def _check_for_important_policies(self) -> None:
             )
 
     @staticmethod
-    def check_domain_ensemble_compatibility(
-        ensemble: Optional["PolicyEnsemble"], domain: Optional[Domain]
+    def assert_compatibility_with_domain(
```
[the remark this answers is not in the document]
ka-bu: yep - I'll rename them to "validate_" so it's consistent and we know what to look out for later

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,315 @@ def _check_for_important_policies(self) -> None:
...
+          ensemble: a list of policies
+          domain: the common domain
+        Raises:
+          `InvalidPolicyEnsembleConfig`: if this ensemble cannot be applied to
+            the given ensemble of policies
+        """
+        super().validate(ensemble=ensemble, domain=domain)
+        # TODO: shouldn't we raise in case priorities are not unique since this might
+        # lead to unexpected results in edge cases (swap order in your config)
+        self.warn_priorities_not_unique(ensemble=ensemble)
```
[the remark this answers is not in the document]
ka-bu: Even then there's the case where they're equally confident and a tie breaker is needed - it makes sense to use priorities for that. If the order of policy predictions (wrt policies who made it) never changes, that order could be used as tie breaker but that seems to inexplicit (and would fail in that very strange case, where you move things around in your config and assume that has no effect).

rasa/core/policies/_policy.py
```diff
@@ -428,3 +428,139 @@ def format_tracker_states(states: List[Dict]) -> Text:
...
+        """
+        if not isinstance(other, PolicyPrediction):
+            return False
+
+        return (
+            self.probabilities == other.probabilities
+            and self.policy_name == other.policy_name
+            and self.policy_priority == other.policy_priority
+            and self.events == other.events
+            and self.optional_events == other.events
```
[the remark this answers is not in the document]
ka-bu: ouch, that's a mean one :D

rasa/core/policies/ensemble.py
ka-bu: agree - I just didn't properly understand how to rephrase "no user prediction"

rasa/core/policies/ensemble.py
ka-bu: just the max or the whole function? I can pull it out and name it argmax

tests/core/test_ensemble.py
ka-bu: I'm not sure what you mean here. The "computed" events are included in the `best_prediction`. I can move the "expected" stuff before the "combine" step to emphasise that?

ka-bu: (and maybe rename `winner_idx` to `expected winner idx`)

rasa/core/policies/ensemble.py
```diff
@@ -1,31 +1,81 @@
...
+       instead of parametrising it's behavior)
+NOTE: differences to previous version
+-  previously there was a comment there saying
+    # find rejected action before running the policies
+    # because some of them might add events
+   this seems to be outdated because the policy prediction contains the event lists
+   and because I could not find policies adding events while they predict
+   --> we do *not* move this to a component that is executed before the policies
+   but leave this snippet in the ensemble's predict
+-  refactored the way the "best prediction" is determined because that decision
```
joejuzl, line 36: I know it's very tempting to refactor old code as we do these migrations. But I really think now is not the ideal time. We have _a huge_ amount of changes as part of 3.0, and I think our aim should be to minimize unneeded changes as much as possible. The more things we change, the harder it will be to debug if things don't work properly when we start the integration change, remember these new migrated components aren't being tested together yet! 
It will be much easier to manage and test refactors once we have everything up and running.

ka-bu: Sorry I should've left the ensembling part itself as un-changed as possible. - But the first part on what is described as change there would mean that if we don't change that we'd need a new component to mimic what was done before.

tests/core/test_ensemble.py
ka-bu: done, better now?

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +220,318 @@ def _check_for_important_policies(self) -> None:
...
+        execution_context: ExecutionContext,
+    ) -> GraphComponent:
+        """Creates a new instance (see parent class for full docstring)."""
+        return cls()
 
-    @classmethod
-    def from_dict(cls, policy_configuration: Dict[Text, Any]) -> List[Policy]:
-        import copy
+    def __str__(self) -> Text:
+        return f"{self.__class__.__name__}()"
```
JEM-Mosig, line 381: This is so generic, shouldn't it be part of the `PolicyPredictionEnsemble` base class?

rasa/core/policies/ensemble.py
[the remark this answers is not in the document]
JEM-Mosig: In principle, we shouldn't compare confidences of two distinct ML policies in the first place. 🙈

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +220,318 @@ def _check_for_important_policies(self) -> None:
...
+          domain: the common domain
+        Raises:
+          `InvalidPolicyEnsembleConfig`: if this ensemble cannot be applied to
+            the given ensemble of policies
+        """
+        super().validate(ensemble=ensemble, domain=domain)
+        # TODO: shouldn't we raise in case priorities are not unique since this might
+        # lead to unexpected results in edge cases (swap order in your config)
+        self.warn_priorities_not_unique(ensemble=ensemble)
+        # TODO: shouldn't we assert that there's at most one rule policy here?
```
JEM-Mosig, line 397: Not sure. A customer might build their own special rule policy that can work in parallel with ours.

rasa/core/policies/ensemble.py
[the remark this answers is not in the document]
JEM-Mosig: I think for now we should just emulate exactly whatever we have in Rasa 2.8.x.

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +220,318 @@ def _check_for_important_policies(self) -> None:
...
-                    domain.index_for_action(rejected_action_name)
-                ] = 0.0
-
-        return self._pick_best_policy(predictions)
+        # grab the index of a prediction whose tuple of scores is >= all the
+        # tuples of scores for any other prediction
+        arg_max = max(
+            list(range(len(predictions))), key=lambda idx: scores(predictions[idx])
+        )
+        return arg_max
```
JEM-Mosig: Just checking: Is this exactly equivalent to the 2.8.x procedure?

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +220,318 @@ def _check_for_important_policies(self) -> None:
...
+        if tracker.latest_action_name == ACTION_LISTEN_NAME:
+            was_end_to_end_prediction = best.is_end_to_end_prediction
+            if was_end_to_end_prediction:
+                logger.debug("Made prediction using user text.")
+            else:
+                logger.debug("Made prediction without user text")
+            event = DefinePrevUserUtteredFeaturization(was_end_to_end_prediction)
+            events.append(event)
+            logger.debug(f"Added `{event}` event.")
+        return events
```
JEM-Mosig, line 535: Can we somehow move this right into the policy? Or would this be more messy?

rasa/core/policies/memoization.py
```diff
@@ -25,7 +25,7 @@
 )
 from rasa.core.featurizers.tracker_featurizers import FEATURIZER_FILE
 from rasa.shared.exceptions import FileIOException
-from rasa.core.policies.policy import PolicyPrediction, PolicyGraphComponent
```
JEM-Mosig, line 28: I know its just a placeholder, but maybe it'd be cleaner to have something like
```
class PolicyPrediction1:
  # old implementation

class PolicyPrediction2:
  # new implementation

PolicyPrediction = PolicyPrediction2
```

rasa/core/policies/policy.py
```diff
@@ -608,16 +625,16 @@ def __eq__(self, other: Any) -> bool:
         )
 
     @property
-    def max_confidence_index(self) -> int:
-        """Gets the index of the action prediction with the highest confidence.
+    def max_probability_index(self) -> int:
+        """Gets the index of the action with the highest predicted probability.
```
JEM-Mosig, line 629: Let's call it "confidence" or "confidence score" throughout.

tests/core/test_ensemble.py
```diff
@@ -1,566 +1,214 @@
...
-            result,
-            self.__class__.__name__,
-            policy_priority=self.priority,
-            is_end_to_end_prediction=self.is_end_to_end_prediction,
-            is_no_user_prediction=self.is_no_user_prediction,
-            events=self.events,
-            optional_events=self.optional_events,
-        )
+@pytest.fixture
+def default_ensemble(
```
JEM-Mosig, line 41: Shouldn't those be in `conftest.py`?

rasa/core/policies/ensemble.py
JEM-Mosig: Can we add some tests that compare outputs of both old and new policy ensembles?

JEM-Mosig (changes requested):
Again, code changed while I was reviewing, so you may already have addressed some of my comments. My main ask is that we should have tests that ensure that the old and new implementations make exactly the same decisions.

rasa/core/policies/memoization.py
ka-bu: We've used "2" or "GraphComponent" suffix pretty consistently and added #TODOs where stuff needs to be removed (both suffixes will disappear when we refactor/rename in the end)

rasa/core/policies/ensemble.py
ka-bu: The functionality from this function was basically [pick_best_policy](https://github.com/RasaHQ/rasa/blob/0aa9e6c2357dc133d73e870e0d53966dca88fdfa/rasa/core/policies/_ensemble.py#L501) before. There are no tests for that `pick_best_policy` alone in the [tests](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py).

What we had in the old tests: 
- test that policy with higher priority wins if probablities are equal [here](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py#L148)
- end to end predictions supersede others [here](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py#L329)
- no user prediction supersedes others - even if there is an end to end prediction [here](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py#L354)
- tests that check whether optional/mandatory events have been applied (`applies_*_events_*` )
- tests that check whether the featurization definition even thas been added ([here and following test](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py#L434))

All of these have been replaced by equivalent tests ( and we *must* replace them because those tests worked with Policies whereas this new component is supposed to work with PolicyPredictions only)

What is *not* covered:
- the tests related to UnexpecTEDIntent - which rather seem to test what kind of prediction UnexpecTEDIntent makes in a certain situation (because of this [comment](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py#L562))
- ( similar here](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py#L503)

rasa/core/policies/_policy.py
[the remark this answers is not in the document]
ka-bu: revert or keep?

tests/core/test_ensemble.py
ka-bu: This fixture is just used here so far

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +220,318 @@ def _check_for_important_policies(self) -> None:
...
+            predictions: a list of policy predictions that include "probabilities"
+              which are non-negative but *do not* necessarily up to 1
+            tracker: dialogue state tracker holding the state of the conversation,
+              which may influence the combination of predictions as well
 
-        self._add_package_version_info(metadata)
+        Returns:
+            a single prediction
+        """
+        ...
```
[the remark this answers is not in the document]
ka-bu: It's already marked as abstractmethod - afaik, think we use "..." everywhere in the new graph components in this case

rasa/core/policies/ensemble.py
ka-bu: There's not really a point in moving that there because the child class won't use str(super()) and you'll never instantiate the (abstract) base class (I should give that one an "ABC" though :D)

rasa/core/policies/ensemble.py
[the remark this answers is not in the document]
ka-bu: >In principle, we shouldn't compare confidences of two distinct ML policies in the first place. 🙈

And I still think we shouldn't call these things confidences 😅   - but I'll revert the probability stuff because Daksh just reminded me that it's not even consistently discrete probability distributions  but can be an array of probabilities .... 😬

rasa/core/policies/ensemble.py
ka-bu: I added this because `rule_only_data` is picked from the first rule policy that is found which seemed odd. -- But I can remove this and we'll add a note in the rulepolicy or whatever is distributing that rule_policy. (FYI @wochinge )

rasa/core/policies/ensemble.py
ka-bu: Oh, I see - "choose_events" is a poor name for this as it is adding additional events maybe and we could pull the  `DefinePrevUserUtteredFeaturization` out .... Or, do you mean the whole "_choose_events"?

rasa/core/policies/_policy.py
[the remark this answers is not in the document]
wochinge: Priorities are also one `int` [here](https://github.com/RasaHQ/rasa/blob/19fb70a05e3637efb9512a942c56cd12ad2c738e/rasa/core/policies/policy.py#L511), aren't they?

I like the rename. The only issue is that `PolicyPrediction` might be used by users who have custom policies. Means we'd have to make sure to add this change to the migration guide or make this non breaking

rasa/core/policies/ensemble.py
```diff
@@ -1,31 +1,81 @@
...
+   --> we do *not* move this to a component that is executed before the policies
+   but leave this snippet in the ensemble's predict
+-  refactored the way the "best prediction" is determined because that decision
+   process was a bit scattered and in-explicit and changed PolicyPrediction2 such that
+   priority is an int not a tuple (?!)
+- `is_not_in_training_data` has been removed
+   -->  can be replaced by an check that determines whether the policy is
+        some kind of memoization policy *inside the tests* (via its name...)
+NOTE/FIXME: questions
+-  Where are the events from the (winning) predictions applied? We only add things to
```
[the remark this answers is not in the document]
wochinge: can we remove `fixme` note then?

rasa/core/policies/ensemble.py
```diff
@@ -1,31 +1,81 @@
...
+    # because some of them might add events
+   this seems to be outdated because the policy prediction contains the event lists
+   and because I could not find policies adding events while they predict
+   --> we do *not* move this to a component that is executed before the policies
+   but leave this snippet in the ensemble's predict
+-  refactored the way the "best prediction" is determined because that decision
+   process was a bit scattered and in-explicit and changed PolicyPrediction2 such that
+   priority is an int not a tuple (?!)
+- `is_not_in_training_data` has been removed
+   -->  can be replaced by an check that determines whether the policy is
```
wochinge: That would be great! Can we do this as part of this PR or create an issue so we don't forget?

rasa/core/policies/ensemble.py
```diff
@@ -34,53 +84,130 @@
...
-        policies: List[Policy],
-        action_fingerprints: Optional[Dict[Any, Dict[Text, List]]] = None,
-    ) -> None:
-        self.policies = policies
-        self.date_trained = None
+def propagate_specification_of_rule_only_data(ensemble: List[Policy]) -> None:
+    """Extracts information on rule only data and propagates it to all policies.
 
-        self.action_fingerprints = action_fingerprints or {}
+    TODO: remove this once there's a corresponding new graph component taking
```
wochinge, line 57: I'd dump it now. Will request review for this later today from you 👍🏻

rasa/core/policies/_policy.py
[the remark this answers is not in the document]
ka-bu: just reverted it

rasa/core/policies/ensemble.py
```diff
@@ -34,53 +84,130 @@
...
+    else:
+        rule_only_data = dict()
+    for policy in ensemble:
+        policy.set_shared_policy_states(rule_only_data=rule_only_data)
 
-        self._check_priorities()
-        self._check_for_important_policies()
 
-        self._set_rule_only_data()
+def validate_warn_rule_based_data_unused_or_missing(
```
wochinge: Mhm, actually we can't do this check as part of the `RulePolicy` as it wouldn't fire in case users don't add this. This would probably go into [this component](https://github.com/RasaHQ/rasa/issues/9336)

wochinge: (updated the issue description accordingly)

rasa/core/policies/ensemble.py
```diff
@@ -34,53 +84,130 @@
...
 
-    def _check_for_important_policies(self) -> None:
+    Observe that policies predict "mandatory" as well as "optional"
+    events. The ensemble decides which of the optional events should
+    be passed on.
+    """
 
-        if not any(isinstance(policy, RulePolicy) for policy in self.policies):
+    @classmethod
+    def validate_ensemble_valid(cls, ensemble: List[Policy], domain: Domain) -> None:
```
wochinge, line 134: I think we can dump this once we move this to the recipe validation.

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,316 @@ def _check_for_important_policies(self) -> None:
...
-            Whether or not any of the supplied training trackers contain rule-based
-            data.
+        Raises:
+            `InvalidDomain` exception if the given domain is incompatible with the
+            given ensemble
         """
-        return any(tracker.is_rule_tracker for tracker in training_trackers)
+        # TODO: does it make sense to add `validate_against_domain`
+        # to other policies?
+        RulePolicy.validate_against_domain(ensemble, domain)
```
wochinge: I think this should be done by the `RulePolicy` itself during training

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,316 @@ def _check_for_important_policies(self) -> None:
...
-                f"your pipeline.",
-                docs=DOCS_URL_RULES,
-            )
-        elif not is_rules_consuming_policy_available and contain_rule_trackers:
-            rasa.shared.utils.io.raise_warning(
-                f"Found rule-based training data but no policy supporting rule-based "
-                f"data. Please add `{RulePolicy.__name__}` or another rule-supporting "
-                f"policy to the `policies` section in `{DEFAULT_CONFIG_PATH}`.",
-                docs=DOCS_URL_RULES,
+        if domain.form_names and not contains_rule_policy:
```
wochinge, line 199: I think this should be done by this https://github.com/RasaHQ/rasa/issues/9336

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,316 @@ def _check_for_important_policies(self) -> None:
...
+        )
 
-        self.date_trained = datetime.now().strftime("%Y%m%d-%H%M%S")
+        # combine the resulting predictions (with probabilities that do not
+        # necessarily sum up to one anymore)
+        final_prediction = self.combine_predictions(predictions, tracker)
 
-    def probabilities_using_best_policy(
-        self,
+        logger.debug(f"Predicted next action using {final_prediction.policy_name}.")
```
wochinge, line 232:
```suggestion
        logger.debug(f"Predicted next action using '{final_prediction.policy_name}'.")
```

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,316 @@ def _check_for_important_policies(self) -> None:
...
-                f"Every featurizer can only have 1 state "
-                f"featurizer but one of the featurizers uses "
-                f"{len(featurizer_config['state_featurizer'])}."
-            )
-        state_featurizer_config = featurizer_config["state_featurizer"][0]
-        state_featurizer_name = state_featurizer_config.pop("name")
-        state_featurizer_func = registry.state_featurizer_from_module_path(
-            state_featurizer_name
-        )
+    def validate(self, ensemble: List[Policy], domain: Domain) -> None:
```
wochinge, line 333: can we dump the whole method then? I think it's good to add later but as far as I can see it's not a check we have currently and gives just one more thing to take care of for the architecture revamp otherwise

rasa/core/policies/ensemble.py
wochinge: Let's keep it out of this PR and focus on migrating the ensemble in this PR. I'd not even create a note - if even then an issue.

rasa/core/policies/ensemble.py
```diff
@@ -91,590 +218,316 @@ def _check_for_important_policies(self) -> None:
...
 
-        policy_events = []
-        for policy_name, prediction in predictions.items():
-            policy_events += prediction.events
+        # make a copy and swap out the events - and drop action metadata
+        # (TODO: what is action meta data and why was it ignored?)
+        final = copy.copy(predictions[best_idx])
+        final.events = events
+        final.optional_events = []
+        final.action_metadata = None
```
wochinge, line 407: `UnexpecTED` is using this and it should definitely not be ignored 🙈

wochinge: great catch 👍🏻

wochinge:
@ka-bu Do we already have an issue for all the `Recipe` validations? Would be great to open one and then move all the code todos there and remove them from the code itself

rasa/core/policies/_policy.py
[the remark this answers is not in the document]
ka-bu: oh forgot to reload and didn't see your comment :D  -- I doubt that users access the internals of PolicyPredictions (they'd need to write custom ensembles for that) because the constructor does not change -- but Daksh reminded me that it might be used elsewhere

rasa/core/policies/ensemble.py
ka-bu: nice, yeah that makes sense :)

rasa/core/policies/ensemble.py
ka-bu: > @ka-bu Do we already have an issue for all the `Recipe` validations? Would be great to open one and then move all the code todos there and remove them from the code itself

nope, I'll remove all the validate_ functions now and open an issue

rasa/core/policies/ensemble.py
ka-bu: Sorry that comment is confusing. The old implementation instantiates a new prediction with everything except the `action_metadata`. So this new version is the same as before - I was just wondering if there *should be* something.  ....So shouldn't this be `None`?

ka-bu:
> @ka-bu Do we already have an issue for all the `Recipe` validations? Would be great to open one and then move all the code todos there and remove them from the code itself

@wochinge , issue is [here](https://github.com/RasaHQ/rasa/issues/9573)

rasa/core/policies/ensemble.py
ka-bu: 1. Do we want to add a separate component to extract the rejected action before processing the policies, or do we all agree that policies do not alter trackers during their process - @joejuzl  @JEM-Mosig  ? 
2. Do we want to keep the "best prediction computation" as is - or (and this is what it boils down to in the end) replace the two `if` statements by something readable (see comment/explanation below) (because the rest is really just reducing the iterative filtering of lists into a single filter/loop)?  Remember that the tests still need to change anyway because we're not working with policies but predictions. @joejuzl @JEM-Mosig ?

rasa/core/policies/ensemble.py
wochinge: Exactly. The `action_metadata` shouldn't be `None`. Instead the action metadata should be passed over to the final `PolicyPrediction`

rasa/core/policies/memoization.py
wochinge: @JEM-Mosig This would basically invalidate the git history for the migrated classes. By duplicating we preserve at least large parts of the history

rasa/core/policies/ensemble.py
ka-bu: (and the reason I'm talking about str(super()) is that other ensembles might have configs that we'd like to printout)

rasa/core/policies/ensemble.py
ka-bu: @JEM-Mosig  - which tests are you missing and does the above explanation help?

rasa/core/policies/ensemble.py
joejuzl: I would stick to the bare minimum required to convert to a `GraphComponent`. We can always refactor/improve the code later, but right now we are very time-constrained and in the middle of a huge change.

ka-bu: Ok so, @joejuzl @wochinge  regarding...
1. As said before - this will either looks similar but change the logic (because rejection is done after predicting) or not change the logic but require more changes (i.e. a component that rejects actions before the policies predict). @JEM-Mosig  will think about this tomorrow and confirm (as far as I see that comment regarding "this needs to be done before" is obsolete - otherwise the whole "let's put events into policy predictions" wouldn't be there I think)
2. Just reverted the to the old if-logic with necessary adaptions (i.e. list of predictions instead of dictionary of policies). Tests stay as "changed" as they are - because there's no more need to create policies that predict hard-coded constants, since this new component works with predictions. @JEM-Mosig  thinks those changes are ok and will double check those tests that as well.

rasa/core/policies/ensemble.py
ka-bu: added this as TODO for later here: https://www.notion.so/rasa/Cleanup-ensemble-logic-e3a4b699fc2c4c75a8d77613ae31b526

rasa/core/policies/ensemble.py
ka-bu: fixed this in the reverted version

ka-bu:
@wochinge  - I had (mentioned in the notes and) skipped [these](https://github.com/RasaHQ/rasa/blob/main/tests/core/test_ensemble.py#L503) tests requiring a trained `UnexpecTEDIntentPolicy`. Did I miss a good way to replace those?

rasa/core/policies/ensemble.py
```diff
@@ -538,49 +183,56 @@ def _pick_best_policy(
...
             The winning policy prediction.
         """
-        # find rejected action before running the policies
-        # because some of them might add events
-        rejected_action_name = None
+        # NOTE: previously this method did find rejected action before running the
+        # policies "because some of them might add events"
+        # - if this comment is obsolete and no prediction modifies trackers during
+        #   predictions, then the following is fine
+        # - if this comment is not obsolete then we have to invent a new
```
joejuzl, line 230: I don't understand why this would need to be done in a different component?

rasa/core/policies/policy.py
```diff
@@ -598,7 +596,7 @@ def __eq__(self, other: Any) -> bool:
             and self.policy_name == other.policy_name
             and self.policy_priority == other.policy_priority
             and self.events == other.events
-            and self.optional_events == other.events
```
joejuzl, line 601: was this just a bug before?

rasa/core/policies/ensemble.py
ka-bu: Because this new component works with predictions and hence all policies have run "predict" already. The old code tried to identify the rejected action *before* the policies run predict "because some of them might add events". => We can't do that inside the component that runs after all predicts.

rasa/core/policies/policy.py
ka-bu: yep, @wochinge  fixed that

rasa/core/policies/ensemble.py
ka-bu: Reference: [in this line the "predict" happened - after the rejected event has been extracted](https://github.com/RasaHQ/rasa/blob/main/rasa/core/policies/ensemble.py#L599)

joejuzl: Is it possible to remove the events retroactively?

ka-bu: If there was a way to distinguish between the events that the policies might add during predict and the ones that where there before, then yep. Can we distinguish that by the timestamp? 

But ideally this is an obsolete comment from a time where we did not have `PolicyPredictions` that could hold events. 

As far as I see:
- TED only uses tracker.latest_action_name and nothing else
- unexpecTEDintent uses past_states and some other getter from the tracker
- memoization modifies trackers - but only "init_copies" of trackers and those copies are never passed on / do not replace the actual tracker
- rule policy only gets information - or modifies an "init_copy" that is *not* passed on
and I expect Johannes to confirm that in the review (which is why I implemented that action rejected stuff here - and left the comments so that he is aware and double checks)

ka-bu: ( Oh and if that is not the case and all policies could modify the tracker any time they want during predict, then (in addition to that extra component) the predict graph would have to look sequential and pass tracker from one policy to the next in the same order as the for loop in the ensemble before if we wanted to replicate that old behaviour 1:1, I guess 🤔  )

tests/core/test_ensemble.py
```diff
@@ -1,566 +1,204 @@
...
 
-    # prediction cannot be action_unlikely_intent for sure because
-    # the last event is not of type UserUttered and that's the
-    # first condition for `UnexpecTEDIntentPolicy` to make a prediction
-    assert (
-        moodbot_domain.action_names_or_texts[np.argmax(prediction.probabilities)]
-        != ACTION_UNLIKELY_INTENT_NAME
+    # ... not quite there yet, because old implementation creates a policy with
+    # best_policy.priority as priority and the first one is a tuple which then
+    # becomes a tuple with a tuple with an int, so...
```
JEM-Mosig, line 198: Can you clarify what you mean by "first one" here? I see in the code what happens, but the comment is confusing. Maybe replace the comment by a description of why the priority becomes a tuple upon `combine_predictions`.

JEM-Mosig: Actually, where does it become a tuple?

JEM-Mosig (approved):
Looks good! Just one thing about `policy_priority` that confuses me.

tests/core/test_ensemble.py
ka-bu: Inside `PolicyPrediction` - the constructor puts that into a tuple of length 1 😅

rasa/core/policies/ensemble.py
JEM-Mosig: To the best of my knowledge, no policy actually modifies the tracker directly. But it would be good to write some tests for the Policy class that checks this. Maybe a customer has a custom policy that does such shenanigans? I don't think we should allow it.

ka-bu: That's a good point. We should add tests for that.  -- Agree that we shouldn't allow it. Those side effects would mess with all the nice things we could do with the graph.

rasa/core/policies/policy.py
wochinge: I also caused it 🤦🏻

rasa/core/policies/ensemble.py
```diff
@@ -1,525 +1,145 @@
...
+    Observe that policies predict "mandatory" as well as "optional"
+    events. The ensemble decides which of the optional events should
+    be passed on.
+    """
 
-    def probabilities_using_best_policy(
+    @abstractmethod
+    def combine_predictions(
         self,
+        predictions: List[PolicyPrediction],
```
wochinge, line 56: Just realized something: Unless we do a separate `GraphComponent` to merge the predictions into a list or we do some magic in the `GraphComponent` then we can't have several `GraphComponents` return things for the same input params of another one. An easy way to would be to solve this via `kwargs` (and maybe filter for `isinstance(value, PolicyPrediction)`

rasa/core/policies/ensemble.py
```diff
@@ -1,525 +1,145 @@
...
+        """Creates a new instance (see parent class for full docstring)."""
+        return cls()
 
-class SimplePolicyEnsemble(PolicyEnsemble):
-    """Default implementation of a `Policy` ensemble."""
+    def __str__(self) -> Text:
+        return f"{self.__class__.__name__}()"
 
     @staticmethod
-    def is_not_in_training_data(
```
wochinge, line 469: I think we need to keep this function somewhere as [this](https://github.com/RasaHQ/rasa/blob/d7eacd7e2ed8e413896e967a23d3b94fcd18d85c/rasa/core/test.py#L825) code is using it

rasa/core/policies/ensemble.py
```diff
@@ -1,525 +1,145 @@
...
-            state_featurizer_name
-        )
-
-        return state_featurizer_func, state_featurizer_config
-
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> GraphComponent:
```
wochinge, line 111:
```suggestion
    ) -> DefaultPolicyPredictionEnsemble:
```

rasa/core/policies/ensemble.py
```diff
@@ -590,52 +207,56 @@ def _best_policy_prediction(
...
+            tracker: dialogue state tracker holding the state of the conversation
+            domain: the common domain
 
         Returns:
-            The best policy prediction.
+            The "best" prediction.
         """
-        winning_prediction = self._best_policy_prediction(tracker, domain, interpreter)
+        if not predictions:
+            raise InvalidConfigException(
```
wochinge, line 249: optional: Can we please add test for this?

wochinge (commented):
Sorry for the late review!

rasa/core/policies/ensemble.py
ka-bu: oh, I missed that, yep 😬  thanks!

rasa/core/policies/ensemble.py
ka-bu: that function and related code could use some rework I think 😅  -- for now I've just updated the docstr (which was promising the opposite of what happens :D)

rasa/core/policies/ensemble.py
ka-bu: Added a new "combine_predictions_from_kwargs" that just calls the other function. And although the motivation for this function is really the graph, I decided not to make the base class a `GraphComponent`, but maybe we should 🤔 . What do you think?

rasa/core/policies/ensemble.py
```diff
@@ -1,525 +1,203 @@
...
-    def _emit_rule_policy_warning(
-        self, training_trackers: List[DialogueStateTracker]
-    ) -> None:
-        """Emit `UserWarning`s about missing rule-based data."""
-        is_rules_consuming_policy_available = (
-            self._policy_ensemble_contains_policy_with_rules_support()
-        )
-        contain_rule_trackers = self._training_trackers_contain_rule_trackers(
-            training_trackers
+        predictions = [
```
wochinge, line 104: 🙌🏻

rasa/core/policies/ensemble.py
wochinge: Tbh, the base class currently has no use at all 🤷🏻 (it never had - I think it was ought to be customizable but it de facto isn't 🙈 ). Up to you if you want to merge them as part of this as part of this PR

rasa/core/policies/ensemble.py
wochinge: thanks!

wochinge (approved):
🚀

rasa/core/policies/ensemble.py
ka-bu: I'll leave the base class here -- I think we could use some separate ensembles (instead of trying to squeeze all "reasonable" behaviour into one ensemble). And judging from the discussions on RulePolicy vs TED vs things extracted from regex interpreter - I think this could happen soon 😅

PR #9484, opened 2021-08-31
Add validation pass for 'active_loop' forms

### LINKED ISSUE #8760: Warn user if they have not defined a form but used `active_loop: form` in training data
**Description of Problem**:
This issue is one of two* created from this user issue [here.](https://github.com/RasaHQ/rasa/issues/8729)

It seems if `active_loop: some_form` is specified in a story but `some_form` is not specified in the domain, TEDPolicy crashes due to an empty tensor.

(see user issue for more info)

**Overview of the Solution**:
We should warn the user if they have not defined the form but included `active_loop: some_form` in their stories or rules, maybe during `rasa data validate` or before they start training. Note that this is not the same as an action not being defined, which we already check and warn for.

*The [other issue](https://github.com/RasaHQ/rasa/issues/8759) is to handle this more gracefully.

**Definition of Done**:
- [ ] Tests are added
- [ ] Users are warned in `data validate` if they haven't defined a form but have set the active loop to that form in their training data (stories and rules)

usc-m:
**Proposed changes**:
- Adds a validation pass for `active_loop` to ensure that referenced forms exist (resolves #8760)


alwx (approved):
🚀

Looks good, thanks!

PR #9205, opened 2021-07-26
8705 response selector (2.8.x)

ka-bu:
**Proposed changes**:
- see #9010 

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [x] updated the documentation
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #9494, opened 2021-09-01
Regex "Classifier"

### LINKED ISSUE #9328: implement `RegexClassifier`
**Overview of the Solution**:
We need to implement the [`RegexClassifier`](https://www.notion.so/rasa/RegexClassifier-3f2d3912a89e468fbb76497626017d8a). That's a graph component which only runs during inference and which resembles the functionality of `RegexInterpreter` . If message starts with `/ ` it overrides the prediction of the classifier with intent + entities 

[This figure](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#48704bc28dab486dad98e9ebd3ab9077) shows the component's position in the graph.

**Input to the component**
`List[Message]`: This list is either empty or has length 1 contains the parsing results of the previous "NLU" graph components (tokens, intents, entities)

**Output:** 
A list of `Message` objects. If the message started with `/` the `RegexClassifier` should override previous prediction results.

**Definition of Done**:

ka-bu:
**Proposed changes**:
- closes #9328 


ka-bu:
Why is the `RegexInterpreter` in `shared`? -- If it is because of the functionality that is now included in this "handler" component, then I guess I should move it back to `shared` and create a graph component in `rasa.nlu` that just wraps the class from `shared` (i.e. adds a `create`).

alwx:
@ka-bu I guess it was only because other stuff from `shared` depends on it (for example, `rasa.shared.importers.importer`).

rasa/nlu/regex_message_handler.py
```diff
@@ -0,0 +1,292 @@
...
+                f"Failed to interpret some parts. "
+                f"Continuing without {match.group('rest')}. ",
+                docs=DOCS_URL_STORIES,
+            )
+
+        # Add the results to the message.
+        # NOTE: In the original implementation the message TEXT was set to the
+        # text from which the regex was extracted. However, if the regex matches
+        # then we know there is *no* text included in that "TEXT".
+        # TODO: .. is `output_properties` (cf. Message) even used anywhere?
```
alwx, line 144: I don't think it's used now but again, `Message` is a part of `shared` page so it might be in use by third-party apps.

rasa/nlu/regex_message_handler.py
```diff
@@ -0,0 +1,292 @@
...
+                        # end indices of the *substring that matches the entity regex*
+                        # ... but it should be indices of tokens to which these
+                        # entities belong. There are no such tokens. Hence, I think we
+                        # should just skip those indices. But can we, or will something
+                        # complain? If something expects indices (despite not being
+                        # able to use them because there's no text), then we could
+                        # instead use more explicit "unknown" indicators (e.g. -1)
+                        # ENTITY_ATTRIBUTE_START: sidx,
+                        # ENTITY_ATTRIBUTE_END: eidx,  # can't be more specific
+                        ENTITY_ATTRIBUTE_VALUE: entity_value,
```
alwx, line 237: IMO it's fine to just skip them in this case. That's one of the reasons for this architectural change — to get rid of redundant/useless stuff.

rasa/nlu/regex_message_handler.py
```diff
@@ -0,0 +1,292 @@
...
+        If no confidence is specified, then this method returns the maximum
+        confidence `1.0`.
+        If a confidence is specified but extraction fails, then this method defaults
+        to a confidence of `0.0`.
+
+        Args:
+          match: a match produced by `self.pattern`
+          domain: the domain
+        Returns:
+          some confidence value
```
alwx, line 255: The docstring format seems to not be correct (see other docstrings).

rasa/nlu/regex_message_handler.py
```diff
@@ -0,0 +1,292 @@
...
+            try:
+                confidence = float(confidence_str)
+            except ValueError:
+                confidence = 0.0
+                raise ValueError(
+                    f"Expected confidence to be a non-negative decimal number but "
+                    f"found {confidence}. Continuing with 0.0 instead."
+                )
+            # NOTE: This is new. Before, we would've just passed on this value,
+            # whatever it was.
```
alwx, line 271: Curious why it's needed.

alwx (changes requested):

rasa/nlu/regex_message_handler.py
ka-bu: I can add the text back in - but then all models working "end to end" will see that weird expression? 🤔  Whom could we ask?

rasa/nlu/regex_message_handler.py
ka-bu: @dakshvar22 -- FYI, we had talked about this 'classifier' before you want on holiday -- we'll remove the TEXT and remove the START/END since no downstream model would do anything meaningful with it (because if it does something with those indices, it would also expect tokens and tokens/text, but those aren't there)

rasa/nlu/regex_message_handler.py
ka-bu: The default is 1.0 which somehow implies that this should be bounded by 1.

rasa/nlu/regex_message_handler.py
ka-bu: @JEM-Mosig this might have to do something with #9587 - in addition to TED just being TED and being "only_e2e" in that case ... what do you think?

ka-bu: > I can add the text back in - but then all models working "end to end" will see that weird expression? 🤔 

Came to the conclusion [here](https://github.com/RasaHQ/rasa/issues/9587#issuecomment-916061233) that for various reasons (e.g. how the (only) ensemble works) models shouldn't see those messages -- if and only if a `RulePolicy` is contained in the ensemble. So looks to me like it would be better to remove this here.

rasa/nlu/regex_message_handler.py
ka-bu: Daksh just confirmed this is reasonable :)

rasa/nlu/regex_message_handler.py
ka-bu: looked ok for me - but I've replaced the whitespaces with tabs anyway - better now?

rasa/nlu/regex_message_handler.py
```diff
@@ -0,0 +1,277 @@
...
+
+        # If it doesn't match, then (potentially) something went wrong, because the
+        # message text did start with the special prefix -- however, a user might
+        # just have decided to start their text this way.
+        if not match:
+            logger.warning(f"Failed to parse intent end entities from '{user_text}'.")
+            return message
+
+        # Extract attributes from the match - and validate it via the domain.
+        intent_name = RegexMessageHandler._parse_intent_name(match, domain)
```
wochinge, line 123: ?
```suggestion
        intent_name = self._parse_intent_name(match, domain)
```

rasa/nlu/regex_message_handler.py
```diff
@@ -0,0 +1,283 @@
...
+    PREDICTED_CONFIDENCE_KEY,
+    TEXT,
+)
+import rasa.shared.utils.io
+from rasa.shared.nlu.training_data.message import Message
+
+logger = logging.getLogger(__name__)
+
+
+class RegexMessageHandler(GraphComponent):
```
wochinge, line 31: should we put this into the `nlu.classifier` package as it's kind of like the `DIETClassifier` ?

ka-bu: good idea - Btw, I just had a look at the entity featurization (for #9205), and I think no *ML* policy does anything with the entities extracted here 😅   (since there are no tokens - which means all the "entities to index" functions that are included in `SingleStateFeaturizer` won't do anything with these entities). Rule/Memoization should be able to handle them 🤔  ....  Anyway, will move to classifier! 😄

wochinge: lol 🙈 cc @JEM-Mosig As we have considered removing the entity featurization with the global slot mappings

rasa/nlu/classifiers/regex_message_handler.py
```diff
@@ -0,0 +1,283 @@
...
+        # convert them into the list of dictionaries that we expect
+        entities: List[Dict[Text, Any]] = []
+        for entity_type, entity_values in parsed_entities.items():
+            if not isinstance(entity_values, list):
+                entity_values = [entity_values]
+            for entity_value in entity_values:
+                entities.append(
+                    {
+                        ENTITY_ATTRIBUTE_TYPE: entity_type,
+                        ENTITY_ATTRIBUTE_VALUE: entity_value,
```
wochinge, line 229: mhm, seems the current `RegexInterpreter` doesn't handle entity `role` / `group`. Guess it's fine to omit this from here then as well for now

tests/nlu/classifiers/test_regex_message_handler.py
```diff
@@ -0,0 +1,266 @@
...
+from typing import Dict, Optional, Text, List, Any
+import json
+import numpy as np
+
+from rasa.shared.nlu.training_data.message import Message
+from rasa.shared.core.domain import Domain
+from rasa.engine.storage.resource import Resource
+from rasa.engine.graph import ExecutionContext
+from rasa.engine.storage.storage import ModelStorage
+from rasa.nlu.regex_message_handler import RegexMessageHandler
```
wochinge, line 11:
```suggestion
from rasa.nlu.classifiers.regex_message_handler import RegexMessageHandler
```

rasa/nlu/classifiers/regex_message_handler.py
```diff
@@ -0,0 +1,283 @@
...
+                f"Expected entities from {domain.entities} "
+                f"but found {unknown_entity_types}. "
+                f"Continuing without unknown entity types. ",
+                docs=DOCS_URL_STORIES,
+            )
+            parsed_entities = {
+                key: value
+                for key, value in parsed_entities
+                if key not in unknown_entity_types
+            }
```
wochinge, line 218: opt: this doesn't seem to be tested atm

tests/nlu/classifiers/test_regex_message_handler.py
```diff
@@ -0,0 +1,266 @@
...
+        action_names=[],
+        forms={},
+    )
+
+    parsed_messages = regex_message_handler.process([message], domain)
+
+    assert parsed_messages[0] == message
+
+
+# TODO: this isn't done anymore - what was that good for?
```
wochinge, line 263: mhm, good question. I think we should still add the `TEXT` to the returned message

rasa/nlu/classifiers/regex_message_handler.py
```diff
@@ -0,0 +1,283 @@
...
+            PREDICTED_CONFIDENCE_KEY: confidence,
+        }
+        intent_ranking = [
+            {INTENT_NAME_KEY: intent_name, PREDICTED_CONFIDENCE_KEY: confidence,}
+        ]
+        message_data = {}
+        message_data[INTENT] = intent_data
+        message_data[INTENT_RANKING_KEY] = intent_ranking
+        message_data[ENTITIES] = entities
+        return Message(message_data, output_properties=set(message_data.keys()))
```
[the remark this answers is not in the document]
ka-bu: .... which tokens?

rasa/nlu/classifiers/regex_message_handler.py
ka-bu: damn, yep, sorry

rasa/nlu/classifiers/regex_message_handler.py
ka-bu: yeah, someone would need to adapt the pattern for that as well

rasa/nlu/classifiers/regex_message_handler.py
```diff
@@ -0,0 +1,283 @@
...
+import logging
+import re
+from typing import Any, Dict, Match, Optional, Pattern, Text, List
+
+from rasa.engine.graph import GraphComponent, ExecutionContext
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
+from rasa.shared.core.domain import Domain
+
+# FIXME: https://rasa.com/docs/rasa/stories/ doesn't actually explain this format?...
```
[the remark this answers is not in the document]
ka-bu: done

rasa/nlu/classifiers/regex_message_handler.py
ka-bu: (because currently, the pattern is {<type> : <value>} not {"type": type, ...})

tests/nlu/classifiers/test_regex_message_handler.py
ka-bu: Why? - If used in combination with `TED`, we can run into the (edge case) that this weird text is interpreted as text. Meanwhile, `RulePolicies` don't care about the text because they work with the extracted intent.

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
wochinge: the ones from the `Tokenizer`. Should be technically just one 1 token but would be still great to not omit it

tests/nlu/classifiers/test_regex_message_handler.py
wochinge: We're for example doing [this](https://github.com/RasaHQ/rasa/blob/dc116ec17b9b2df71c5a0f29da9cfff573a1c183/rasa/graph_components/adders/nlu_prediction_to_history_adder.py#L54). Even we if replace that with `original_message.text` we'd still have a `UserUttered` with the text which would get featurized by TED. We should in my opinion be carefully with omitting the text as I think too many things downstream depend on the text (e.g. Rasa X). 

Fully agree that this is not ideal and that we need to separate the user's raw input from the prediction instead of merging both into one event but I don't think we can fix this until we re-organize the events

ka-bu: The `.get()` will return `None` and a `UserUttered` is able to cope with a text set to `None`. The policy featurization pipeline can deal with missing texts as well. So that doesn't look like a problem to me there.

> think too many things downstream depend on the text (e.g. Rasa X).
That on the other hand, sounds like a good point 😬 

I'll add it back in and just open an issue...

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
ka-bu: Where did that happen in the old version? The regex interpreter didn't return tokens

tests/nlu/classifiers/test_regex_message_handler.py
ka-bu: @wochinge , (apart from the issue with keeping/removing text) to avoid confusion: The test below the `TODO` comment linked above shows that regex interpreter dealt with messages that didn't start with the "/" prefix and prepended that if it wasn't there. -- However in `nlu/run.py` and in `processor.py` the regex interpreter is only used if it does start with the "/" prefix. Which is why in the new version, the "/" is mandatory and not followed by a "?" in the pattern. Long story short: That linked test there will be removed anyway and I'll add a "text is still there" check to tests :D

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
wochinge: @ancalita Did we skip that on purpose back then or did we just miss it?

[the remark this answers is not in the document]
ka-bu: If there ever where tokens, then it's a bit strange that the indices where chosen like [this](https://github.com/RasaHQ/rasa/blob/7625fcaec9a5d1f47dbcc093a405e9369e7261fc/rasa/shared/nlu/interpreter.py#L120) 🤔  ... which is removed atm as well 😬  (because policies etc won't use those anyway - but I'll add them back in with the text now because RasaX)

tests/nlu/classifiers/test_regex_message_handler.py
wochinge: Makes sense to make it mandatory! 

> I'll add it back in and just open an issue...

Thanks 🙌🏻 And sorry for the caution 😞

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
wochinge: The `RegexInterpreter` definitely never had tokens. This is a very recent changes. If we don't add tokens everytime we need to communicate that to Christof and his team

rasa/nlu/classifiers/regex_message_handler.py
wochinge: oooh - the test actually fails earlier because your jsons use single quotes and jsons _have_ to have double quotes

ka-bu: oh, great catch - I'll add expected messages / match parameters to those pytest.warns

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
ancalita: Hmm, good question on why `RegexInterpreter` doesn't add tokens, honestly it didn't occur to me during implementation 🙈 
I was following this [guide ](https://www.notion.so/rasa/Sending-tokens-with-events-to-Rasa-X-05fe8895e4aa416e8ca8e3e1f339bc91)which @joejuzl prepared, would you know why tokens are only added by the `nlu.model.Interpreter`?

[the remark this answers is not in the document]
ka-bu: It can't the messages it gets look like "/<intentname>@<confidencevalue><entity-dict>" - there is no text :)

Due to the worries regarding RasaX, I've just added back in:
- "text" attribute - which contains that malformed message and will need to be caught by TED later then (will open an issue if we merge this eventually :D)
- "start/end" positions for entities - which are (just as before) the start/end of the match of the entity regex

tests/nlu/classifiers/test_regex_message_handler.py
ka-bu: no problem at all ^^

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
joejuzl: @ancalita it was not intentional to not add tokens in the `RegexInterpreter` I just didn't think of it in the scope/implementation proposal. Does it even make sense though? What would the tokens even be? Just a single one fo the intent?

rasa/nlu/regex_message_handler.py
JEM-Mosig: MemoizationPolicy can definitely distinguish by entity. RulePolicy doesn't.

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
wochinge: > What would the tokens even be? Just a single one fo the intent?

I thought just whatever the tokenizer gives us for this 🤷🏻 We can also omit - in that case Rasa X should fallback to just use the entire text as token? Do we have an issue for Rasa X where we can put that information?

wochinge (approved):
The last piece 🚀 💯

rasa/nlu/classifiers/regex_message_handler.py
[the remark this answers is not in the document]
wochinge: I'll add this to [this information](https://github.com/RasaHQ/rasa-x/issues/4704)

ka-bu:
@wochinge  , test-core-featurizers, ubuntu-latest, 3.8 fails with this mitie error.  ....?
```
_____________ ERROR collecting tests/nlu/utils/test_mitie_utils.py _____________
ImportError while importing test module '/home/runner/work/rasa/rasa/tests/nlu/utils/test_mitie_utils.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.8.11/x64/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/nlu/utils/test_mitie_utils.py:10: in <module>
    import mitie
E   ModuleNotFoundError: No module named 'mitie'
```

wochinge:
I re-ran. I hope it's just some flakiness. From the logs it appears as if the dependency was installed correctly

ka-bu:
> I re-ran. I hope it's just some flakiness. From the logs it appears as if the dependency was installed correctly

should've thought of that - thanks!

PR #9513, opened 2021-09-02
Handle syslog writing from the command line

PaulNewtech:
**Proposed changes**:
- This change allows you to give the system log handler as a parameter to the` rasa run` command line


rasa/core/utils.py
```diff
@@ -40,14 +40,18 @@ def configure_file_logging(
...
         return
-
-    formatter = logging.Formatter("%(asctime)s [%(levelname)-5.5s]  %(message)s")
-    file_handler = logging.FileHandler(
-        log_file, encoding=rasa.shared.utils.io.DEFAULT_ENCODING
-    )
-    file_handler.setLevel(logger_obj.level)
-    file_handler.setFormatter(formatter)
-    logger_obj.addHandler(file_handler)
+    if log_file == "/dev/log":
```
joejuzl, line 43: Is there a way we can confirm this is a syslog file? Relying on the path seems a bit brittle, especially cross-platform.

rasa/core/utils.py
```diff
@@ -40,14 +40,18 @@ def configure_file_logging(
...
+        syslog_handler = logging.handlers.SysLogHandler(address=log_file)
+        syslog_handler.setLevel(logger_obj.level)
+        logger_obj.addHandler(syslog_handler)
```
joejuzl, line 46: These two lines could be moved out of the `if` and `else`.

rasa/core/utils.py
```diff
@@ -40,14 +40,18 @@ def configure_file_logging(
...
+    else:
+        formatter = logging.Formatter("%(asctime)s [%(levelname)-5.5s]  %(message)s")
```
joejuzl, line 48: Do we want a formatter for `syslog_handler` too? Or does it handle that itself?

joejuzl (changes requested):
Thanks for this submission! 
Added a couple of comments.

rasa/core/utils.py
PaulNewtech: Thanks for the review
I just changed this behavior by adding a new option to the command line. This way, it will still be a syslog file.
Thank you for the suggestion.

rasa/core/utils.py
PaulNewtech: this comment was relevant, but is not anymore with the new commits

rasa/core/utils.py
PaulNewtech: thank for this comment, I took this into account by adding a specific formatter that shows the pid for syslog

PaulNewtech:
Hello joejuzl,
Thank you for the review. I took your comments into account. Now the syslog is directly handled by a new arg in the command line. 
I look forward to see what you think about this new version.

rasa/core/utils.py
```diff
@@ -30,24 +30,32 @@
...
 ) -> None:
     """Configure logging to a file.
 
     Args:
         logger_obj: Logger object to configure.
         log_file: Path of log file to write to.
+        use_syslog: Add syslog as a logger
     """
-    if not log_file:
+    if not log_file and not use_syslog:
```
joejuzl, line 42: Do we need this if both the subsequent statements are `if`s?

rasa/utils/common.py
```diff
@@ -148,6 +149,15 @@ def update_sanic_log_level(log_file: Optional[Text] = None) -> None:
         logger.addHandler(file_handler)
         error_logger.addHandler(file_handler)
         access_logger.addHandler(file_handler)
+    if use_syslog:
+        formatter = logging.Formatter("%(asctime)s [%(levelname)-5.5s] [%(process)d]"
+                                      " %(message)s")
+        syslog_handler = logging.handlers.SysLogHandler(address='/dev/log')
```
joejuzl, line 155:
```suggestion
        syslog_handler = logging.handlers.SysLogHandler(address=log_file)
```

joejuzl (changes requested):
Looking good! Thanks for making those changes, I think the CLI flag makes sense.
Can you add a changelog please?

Also, the CI will now run, so you may have to fix some tests and code style issues.

rasa/core/utils.py
PaulNewtech: You are right. I fixed that

PaulNewtech: I did this way to be as close as possible from the original file. But if you say that the None return has no impact, I can fix that

rasa/utils/common.py
PaulNewtech: This would totally break the possibility to log in syslog
The idea is that you can log in a log file and in syslog in parallel.

PaulNewtech: I see what you meant about cross plateform subject. 
I'm making a change to use the default communication with the syslog server. I'll make a new request once it's done.

PaulNewtech:
Hi Joe,

I just added a few things to the pull request.

As you said the use of "/dev/log" was a problem with system compatibility. Using the UDP communication with the local syslog server is the way[ you can manage this OS issue.]( https://docs.python.org/3/library/logging.handlers.html#logging.handlers.SysLogHandler)
That's why I decided to had the capability to specify which server, and port your syslog server is listening on and the protocol that is used. There is now those 3 flag more.

It will be very useful for production purposes.

I also manage a few pipeline error.

I will do the changelog after your response :)

Looking forward to reading from you.

rasa/cli/arguments/run.py
```diff
@@ -55,6 +55,30 @@ def add_server_arguments(parser: argparse.ArgumentParser) -> None:
...
+    parser.add_argument(
+        "--use-syslog",
+        action="store_true",
+        help="Add syslog as a log handler",
+    )
+    parser.add_argument(
+        "--syslog-address",
+        type=str,
+        default=constants.DEFAULT_SYSLOG_HOST,
+        help="Adress of the syslog server. --use-sylog flag is required"
```
joejuzl, line 67:
```suggestion
        help="Address of the syslog server. --use-sylog flag is required"
```

joejuzl (changes requested):
Looking good!
I think it was a good decision to allow the full syslog server configuration 👍

joejuzl:
Hi @PaulNewtech 
It looks like some CI checks are failing.
You will need to make sure that Code Quality passes and all the tests.
I will approve and run the workflow, make sure to check the job statuses before merging master again so you can see what needs to be fixed.
Let me know if you need any help.

PaulNewtech:
Hi @joejuzl ,

I hope it will be good with this fix. 
But the error log of the test are not very explicit.

joejuzl:
The code quality test is failing. Try running `make lint` to check locally. (hint: you will need to run the black formatter on your code)

Also for tests you can run `make test` see:  https://github.com/RasaHQ/rasa#running-the-tests

PaulNewtech:
Hi @joejuzl 
I did the required change for the black linter.
The code climate issue to fix seems to have nothing to do with my changes (it was more than 25 lines before it)

joejuzl:
Please make sure to run `make lint` to check all the linting steps pass.

> The code climate issue to fix seems to have nothing to do with my changes (it was more than 25 lines before it)

Codeclimate is not a required check right now - so that's ok.

PaulNewtech:
Thank you !
It's done

PaulNewtech:
Hey @joejuzl , 
I just made changes to the tests to be compliant with the new CLI flags.
Thank you

joejuzl (approved):
🚀 🚀 🚀

joejuzl:
Thanks for putting in all the work to get it merged! 🔥

PaulNewtech:
Thank you @joejuzl  !

PR #9523, opened 2021-09-03
3.0 architecture revamp/9450/lexical syntactic

### LINKED ISSUE #9450: Migrate `LexicalSyntacticFeaturizer` to GraphComponent interface
**Overview of the Solution**:
Adapt the `LexicalSyntacticFeaturizer ` to the new `GraphComponent ` interface.

* adapt persistence
* split `train` into `train` (trains the featurizer) and `process_training_data` (adds features to training data for subsequent classifiers etc.)

**Definition of Done**:
- [ ] LexicalSyntacticFeaturizer adapted to the new GraphComponent interface
- [ ] Unit tests are adapted to adapted component
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old component implementation)

ka-bu:
**Proposed changes**:
- closes #9450


rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +85,409 @@ def required_components(cls) -> List[Type[Component]]:
...
-        value = self.function_dict[feature](token)
-        if value is None:
-            logger.debug(
-                f"Invalid value '{value}' for feature '{feature}'."
-                f" Feature is ignored."
+        Returns:
+          the given list of messages which have been modified in-place
+        """
+        for message in messages:
+            self.process(message)
```
koernerfelicia, line 367:
```suggestion
            self.process_message(message)
```

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +85,409 @@ def required_components(cls) -> List[Type[Component]]:
...
+        for anchor in range(len(tokens)):
+
             # in case of an even number we will look at one more word before,
             # e.g. window size 4 will result in a window range of
             # [-2, -1, 0, 1] (0 = current word in sentence)
-            window_size = len(configured_features)
+            window_size = len(self._feature_config)
             half_window_size = window_size // 2
             window_range = range(-half_window_size, half_window_size + window_size % 2)
+            assert len(window_range) == window_size
```
koernerfelicia, line 300: Can this happen outside of the loop? (not dep. on anchor)

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +85,409 @@ def required_components(cls) -> List[Type[Component]]:
...
+                if not mapping:
+                    continue
+                feature_idx = mapping.get(feature_value, -1)
+                if feature_idx > -1:
+                    rows.append(token_idx)
+                    cols.append(feature_idx)
+        rows = np.array(rows)
+        cols = np.array(cols)
+        data = np.ones(len(rows))
+        return scipy.sparse.coo_matrix((data, (rows, cols)), shape=shape)
```
koernerfelicia, line 438: 🙌

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: oh, good point! will move that

rasa/nlu/featurizers/dense_featurizer/dense_featurizer.py
```diff
@@ -0,0 +1,55 @@
...
+            # if features are all zero, then we must continue with zeros
+            if not any(is_non_zero_vector):
+                dense_sequence_features = np.zeros([1, shape[-1]])
+
+        if pooling_operation == MEAN_POOLING:
+            return np.mean(dense_sequence_features, axis=0, keepdims=True)
+        elif pooling_operation == MAX_POOLING:
+            return np.max(dense_sequence_features, axis=0, keepdims=True)
+        else:
+            raise ValueError(
```
wochinge, line 51: Should be a subclass of `RasaException` right?

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
 
-        # makes sure the alias name is set
-        # Necessary for `unique_name` to be defined
-        self.component_config = component_config
-        component_config.setdefault(FEATURIZER_CLASS_ALIAS, self.unique_name)
+FeatureType = TypeVar("FeatureType")
 
-        super().__init__(component_config)
 
+class Featurizer2(Generic[FeatureType]):
```
wochinge, line 28: 🚀

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
-        # take only non zeros feature vectors into account
-        non_zero_features = np.array([f for f in features if f.any()])
+    def get_default_config() -> Dict[Text, Any]:
+        """Returns the component's default config."""
+        return {FEATURIZER_CLASS_ALIAS: None}
 
-        # if features are all zero just return a vector with all zeros
-        if non_zero_features.size == 0:
-            return np.zeros([1, features.shape[-1]])
+    def __init__(self, config: Optional[Dict[Text, Any]] = None) -> None:
```
wochinge, line 36: is the config actually `Optional`?

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
-            return np.mean(non_zero_features, axis=0, keepdims=True)
+        Args:
+          config: configuration
+        """
+        super().__init__()
+        self.validate_config(config)
+        self._config = config
+        # TODO: we can't use COMPONENT_INDEX anymore to create a unique name
+        # on the fly --> take care of that in graph validation
+        self._identifier = self._config.get(FEATURIZER_CLASS_ALIAS, None)
```
wochinge, line 47: We can get the `node_name` from the `ExecutionContext`

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
 
-        if pooling_operation == MAX_POOLING:
-            return np.max(non_zero_features, axis=0, keepdims=True)
+    @property
+    def identifier(self) -> Text:
```
wochinge, line 50: do we actually need this or can we make `identifier` just a public attribute?

wochinge: also: Who accesses this public attribute? Isn't it enough to keep it protected?

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
+    # TODO: validate config for this component in the context of a recipe
 
-class SparseFeaturizer(Featurizer):
-    pass
+    @classmethod
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
+        """Validates that the featurizer is configured properly."""
+        identifier = config.get(FEATURIZER_CLASS_ALIAS, [])
+        if not identifier:
+            raise ValueError(
```
wochinge, line 65: This should be a subclass of `RasaException`

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
+                f"Expected the config to map key `{FEATURIZER_CLASS_ALIAS}` "
+                f"to some name only found {config}"
```
wochinge, line 67:
```suggestion
                f"to some name only found {config}."
```

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
+            )
+
+    def validate_compatibility_with_tokenizer(self, tokenizer: Tokenizer) -> None:
+        """Validate that the featurizer is compatible with the given tokenizer."""
+        # NOTE: this wasn't done before, there was just a comment
+        # TODO: add (something like) this to graph validation
+        pass
+
+    # @abstractmethod
+    # def train(self, training_data: TrainingData,) -> None:
```
wochinge, line 77: I wouldn't as not all of them train

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
+        """Validates that the featurizer is configured properly."""
+        identifier = config.get(FEATURIZER_CLASS_ALIAS, [])
+        if not identifier:
+            raise ValueError(
+                f"Expected the config to map key `{FEATURIZER_CLASS_ALIAS}` "
+                f"to some name only found {config}"
+            )
+
+    def validate_compatibility_with_tokenizer(self, tokenizer: Tokenizer) -> None:
```
wochinge, line 70: Cool idea! Can we instead of the comment have an issue for this or put in the document with follow-up ideas?

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,142 @@
...
+        for type, features in [
+            (FEATURE_TYPE_SEQUENCE, sequence),
+            (FEATURE_TYPE_SENTENCE, sentence),
+        ]:
+            if features is not None:
+                wrapped_feature = Features(features, type, attribute, self.identifier,)
+                message.add_features(wrapped_feature)
+
+    @staticmethod
+    def validate_featurizers_compatible(featurizers: Iterable[Featurizer2]) -> None:
```
wochinge, line 128: By what is this called?

rasa/utils/io.py
```diff
@@ -172,11 +172,13 @@ def validate(document: Document) -> None:
     return FunctionValidator
 
 
-def json_unpickle(file_name: Union[Text, Path]) -> Any:
+def json_unpickle(file_name: Union[Text, Path], keys: bool = False) -> Any:
     """Unpickle an object from file using json.
 
     Args:
         file_name: the file to load the object from
+        keys: If set to True then jsonpickle will encode non-string dictionary keys
```
wochinge, line 180:
```suggestion
        keys: If set to `True` then jsonpickle will encode non-string dictionary keys
```

rasa/utils/io.py
```diff
@@ -186,22 +188,24 @@ def json_unpickle(file_name: Union[Text, Path]) -> Any:
...
 
 
-def json_pickle(file_name: Union[Text, Path], obj: Any) -> None:
+def json_pickle(file_name: Union[Text, Path], obj: Any, keys: bool = False) -> None:
     """Pickle an object to a file using json.
 
     Args:
         file_name: the file to store the object to
         obj: the object to store
+        keys: If set to True then jsonpickle will encode non-string dictionary keys
```
wochinge, line 200:
```suggestion
        keys: If set to `True` then jsonpickle will encode non-string dictionary keys
```

tests/shared/utils/test_io.py
```diff
@@ -343,6 +358,17 @@ def test_create_directory_for_file(tmp_path: Path):
     assert os.path.exists(os.path.dirname(file))
 
 
+def test_write_utf_8_yaml_file(tmp_path: Path):
+    """This test makes sure that dumping a yaml doesn't result in Uxxxx sequences
+    but rather directly dumps the unicode character."""
+
+    file_path = str(tmp_path / "test.yml")
```
wochinge, line 365: do we even need the `str` cast?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -153,157 +356,137 @@ def _map_features_to_indices(
...
+                )
+                loaded_featurizer = cls(
+                    config=config,
+                    model_storage=model_storage,
+                    resource=resource,
+                    execution_context=execution_context,
+                )
+                loaded_featurizer._set(
+                    feature_to_idx_dict, check_consistency_with_config=True
+                )
```
wochinge, line 473: can't we pass this via the constructor instead of calling an extra method?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,82 +85,269 @@ def required_components(cls) -> List[Type[Component]]:
...
+        self._model_storage = model_storage
+        self._resource = resource
+        self._execution_context = execution_context
+        # featurizer specific
+        self._feature_config = self._config.get(self.FEATURES, [])
+        self._feature_to_idx_dict = {}
+        self._number_of_features = -1
+        # where the negative value indicates that feature_to_idx hasn't been set yet
+
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
```
wochinge, line 156:
```suggestion
    @classmethod
    def validate_config(cls, config: Dict[Text, Any]) -> None:
```

wochinge: do we have tests for the validation?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,82 +85,269 @@ def required_components(cls) -> List[Type[Component]]:
...
+            )
+        except TypeError as e:
+            raise InvalidConfigException(message) from e
+        if configured_feature_names.difference(cls.SUPPORTED_FEATURES):
+            raise InvalidConfigException(message)
+
+    def validate_compatibility_with_tokenizer(self, tokenizer: Tokenizer) -> None:
+        """Validate a configuration for this component in the context of a recipe."""
+        # NOTE: this wasn't done before, there was just a comment
+        # TODO: add (something like) this to graph validation
```
wochinge, line 186: I think if we do this than we should do it in the `Recipe`. The graph schema validation doesn't know / doesn't care about the types of the components.

rasa/nlu/featurizers/dense_featurizer/dense_featurizer.py
ka-bu: I could make this a `InvalidConfigException` ? Though this one feels like a `ValueError` (from which one would re-raise a `InvalidConfigException` only if this is really defined by a config later - and not if it is a hard coded static internal usage)

rasa/nlu/featurizers/featurizer.py
ka-bu: argh, nope :)

rasa/nlu/featurizers/featurizer.py
ka-bu: Ah, nice - will use that as a default to overwrite the configured alias

rasa/nlu/featurizers/featurizer.py
ka-bu: Had thought that the validation will need it - before then adding the validate_* functions and making this obsolete. Will remove it

rasa/nlu/featurizers/featurizer.py
ka-bu: That's a good point

rasa/nlu/featurizers/featurizer.py
ka-bu: Was thinking the graph should do a static checks of the configs. All of these validate_* functions should be class functions (one is not, will fix that now) that can be used during those checks.

rasa/nlu/featurizers/featurizer.py
ka-bu: Added [here](https://www.notion.so/rasa/graph-Validation-static-check-of-configs-bfc65b6d97714fd099938cdbabca0f53) - Might also be included in some of the upcoming Recipe tasks, no? Cause some of the checks have been done in the version we're replacing here (most of all in the ensemble.py)

tests/shared/utils/test_io.py
ka-bu: I just moved that code - don't think so, will remove it

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: We could - this is illustrating the separation I meant before. Let's say we train/persist/load multiple attributes (i.e. internal vocabulary + internal model). If we put vocabulary and model into the __init__ both with "Optional" that looks like I could instantiate this component with some model but vocabulary = None which probably makes no sense (otherwise we wouldn't train and persist both). In contrast, this _set (started calling this _load in other PR), emphasises that both need to be set (cause nothing "Optional"). 
Besides, it would also enable us to put a generic "create" into the graph component that just "__init__"s with the same parameters - which means all GraphComponents just need to implement "__init__" as usual (with super().__init__() taking care of graph stuff) and we get the "create" via inheritance.

... I know this is nitpicking but what do you think :D

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: oh, forgot those - will add now 👍

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: Sounds good - I didn't think much about Recipe vs graph schema :)

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: We could also streamline the load() a bit with this... because all classes will do something like:
```
   instance = cls(config, model_storage, execution_context,....)
   try: 
     with  ... context ...:
           # load resource xy 
           # call _set/_load with all loaded resources (and check whether they conflict with the given config)
   except:
         # complain the model cannot be loaded - maybe because it wasn't trained (which is how it currently works for most components)
    return instance
 ```

Comment (ka-bu) [reply rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py:473]: ... that said, I can also just put those in the __init__ for now :)

File: rasa/nlu/featurizers/dense_featurizer/dense_featurizer.py (hunk 4 above)
Comment (wochinge) [reply rasa/nlu/featurizers/dense_featurizer/dense_featurizer.py:51]: From where do we actually call this? As long as it's an `InvalidConfigException` in the end I'm fine 👍🏻

File: rasa/nlu/featurizers/featurizer.py (hunk 13 above)
Comment (wochinge) [reply rasa/nlu/featurizers/featurizer.py:128]: I think if we do this then this should be done by the Recipes as they know which role which component has in the graph.  I'd lean towards doing all of this later and prioritizing having a minimal running version later. Once this all works we can work on adding more validations etc.

File: rasa/nlu/featurizers/featurizer.py (hunk 12 above)
Comment (wochinge) [reply rasa/nlu/featurizers/featurizer.py:70]: Yep, can do 👍🏻

**Comment by dakshvar22** [comment]
@ka-bu Came across [this ](https://github.com/RasaHQ/rasa/pull/9523/files#diff-89d6559996ad9aeed863473bc182ede3771a8b895590eb4884edcfe706989dc3R48). Are we actually doing max pooling anywhere in the codebase of featurizers?

File: rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py (hunk 17 above)
Comment (wochinge) [reply rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py:473]: > ... I know this is nitpicking but what do you think :D

Haha, In my opinion this is fine. I'd prefer to rather forbid calling the `__init__` directly and always go via `create` / `load` 🤔 I'd go for `__init__` for now to be consistent to how it's done in the other components and then revisit this later.

> because all classes will do something like:

Actually just the ones which train 🤔

File: rasa/nlu/featurizers/dense_featurizer/dense_featurizer.py (hunk 4 above)
Comment (ka-bu) [reply rasa/nlu/featurizers/dense_featurizer/dense_featurizer.py:51]: some of the dense featurizers will use these

File: rasa/nlu/featurizers/featurizer.py (hunk 13 above)
Comment (ka-bu) [reply rasa/nlu/featurizers/featurizer.py:128]: Yep, but the validate_* logic should remain in the classes, no?

File: tests/shared/utils/test_io.py (hunk 16 above)
Comment (ka-bu) [reply tests/shared/utils/test_io.py:365]: Left it as is - thinking maybe this was tested with str on purpose (it is in shared)

File: rasa/nlu/featurizers/featurizer.py (hunk 8 above)
Comment (ka-bu) [reply rasa/nlu/featurizers/featurizer.py:50]: Kept it eventually - Reason is that if we put the identifier as "origin" into the features, then we should be able to identify the featurizer with that string somehow

File: rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py (hunk 17 above)
Comment (ka-bu) [reply rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py]: > Actually just the ones which train 🤔
oh, true - or the ones who "train (but I don't train and just load and persist something downloaded from somehwere)"

**Comment by ka-bu** [comment]
> @ka-bu Came across [this ](https://github.com/RasaHQ/rasa/pull/9523/files#diff-89d6559996ad9aeed863473bc182ede3771a8b895590eb4884edcfe706989dc3R48). Are we actually doing max pooling anywhere in the codebase of featurizers? I think we only do mean pooling everywhere to aggregate sentence level features, isn't it?

@dakshvar22  I haven't checked. But it looks like such a natural / basic operation, that I'd leave this in here regardless of whether it is used at the moment, no?

**Comment by dakshvar22** [comment]
I'd not add it if it is not used by any of the featurizers inside Rasa Open Source to reduce adding any bits of unused code (@wochinge this is an active recommendation for the codebase, right?). To make it easier for people to customize the pooling behavior we can move it to a separate function which I think you already have.

**Comment by ka-bu** [comment]
> I'd not add it if it is not used by any of the featurizers inside Rasa Open Source to reduce adding any bits of unused code (@wochinge this is an active recommendation for the codebase, right?). To make it easier for people to customize the pooling behavior we can move it to a separate function which I think you already have.

Same code was included in featurizer.py before. It has just moved to dense_featurizer. Is/was this the separate function you mean?

File: rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py [hunk 20]
```diff
@@ -254,6 +254,12 @@ def train(self, training_data: TrainingData,) -> Resource:
         """
         feature_to_idx_dict = self._create_feature_to_idx_dict(training_data)
         self._set_feature_to_idx_dict(feature_to_idx_dict=feature_to_idx_dict)
+        if not self._feature_to_idx_dict:
+            rasa.shared.utils.io.raise_warning(
+                "No lexical syntactic features could be extracted from the training"
+                "data. In order for this component to work you need to define "
+                "`features` that can be found in the given training data."
```
wochinge, line 261: I think it somewhere in Notion states to use single quotes 🙈 
```suggestion
                "No lexical syntactic features could be extracted from the training "
                "data. In order for this component to work you need to define "
                "'features' that can be found in the given training data."
```

rasa/nlu/featurizers/dense_featurizer/dense_featurizer.py
```diff
@@ -0,0 +1,56 @@
+from rasa.shared.exceptions import InvalidConfigException
+from typing import Text
+import numpy as np
+
+from rasa.nlu.featurizers.featurizer import Featurizer2
+from rasa.utils.tensorflow.constants import MEAN_POOLING, MAX_POOLING
+
+
+class DenseFeaturizer2(Featurizer2[np.ndarray]):
```
wochinge, line 9: as it doesn't implement all abstract methods?
```suggestion
class DenseFeaturizer2(Featurizer2[np.ndarray], abc.ABC):
```

rasa/nlu/featurizers/featurizer.py
wochinge: but can't we just use `self._identifier` in that case? What's the extra benefit from the `identifier`?

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,141 @@
...
+    @abstractmethod
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
+        """Validates that the component is configured properly."""
+        ...
+
+    @classmethod
+    @abstractmethod
+    def validate_compatibility_with_tokenizer(cls, tokenizer: Tokenizer) -> None:
+        """Validate that the featurizer is compatible with the given tokenizer."""
+        # TODO: add (something like) this to recipe validation
```
wochinge, line 68: should we create a new issue where we collect what we need to validate as part of the recipe? I'd probably split that from the v1 recipe implementation as this is already quite a lot of code and it's not required for the integration phase.

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,141 @@
...
+        ...
 
-class SparseFeaturizer(Featurizer):
-    pass
+    def process_training_data(self, training_data: TrainingData) -> TrainingData:
+        """Processes the training examples in the given training data in-place.
+
+        Args:
+          training_data: the training data
+        Returns:
```
wochinge, line 88:
```suggestion

        Returns:
```

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,141 @@
...
+        for type, features in [
+            (FEATURE_TYPE_SEQUENCE, sequence),
+            (FEATURE_TYPE_SENTENCE, sentence),
+        ]:
+            if features is not None:
+                wrapped_feature = Features(features, type, attribute, self.identifier,)
+                message.add_features(wrapped_feature)
+
+    @staticmethod
+    def validate_configs_compatible(
```
wochinge, line 118: nit: should we move this up so it's bundled with the rest of the `validate_` methods?

rasa/utils/io.py
```diff
@@ -186,22 +188,24 @@ def json_unpickle(file_name: Union[Text, Path]) -> Any:
...
 
 
-def json_pickle(file_name: Union[Text, Path], obj: Any) -> None:
+def json_pickle(file_name: Union[Text, Path], obj: Any, keys: bool = False) -> None:
     """Pickle an object to a file using json.
 
     Args:
         file_name: the file to store the object to
         obj: the object to store
+        keys: If set to `True` then jsonpickle will encode non-string dictionary keys
```
wochinge, line 200: optional: Is a name like `encode_non_string_keys` more precise than? Just `keys` seems kind of vague

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -1,66 +1,84 @@
...
+END_OF_SENTENCE = "EOS"
+BEGIN_OF_SENTENCE = "BOS"
 
-    Moves with a sliding window over every token in the user message and creates
-    features according to the configuration.
+FEATURES = "features"
+
+
+class LexicalSyntacticFeaturizerGraphComponent(SparseFeaturizer2, GraphComponent):
+    """Extracts and encodes lexical syntactic features.
```
wochinge, line 51: thanks for these amazing docstrings 🙌🏻 🚀

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -153,157 +372,130 @@ def _map_features_to_indices(
...
-
-            final_sequence_features = Features(
-                sequence_features,
-                FEATURE_TYPE_SEQUENCE,
-                TEXT,
-                self.component_config[FEATURIZER_CLASS_ALIAS],
+            sentence_features = self._map_tokens_to_raw_features(tokens)
+            sparse_matrix = self._map_raw_features_to_indices(sentence_features)
+            self.add_features_to_message(
+                # FIXME: create sentence feature and make `sentence` non optional
```
wochinge, line 403: should we create an issue for this?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,82 +87,283 @@ def required_components(cls) -> List[Type[Component]]:
...
+        """
+        # Note that this will only sort the top level keys - and we keep
+        # doing it to ensure consistently with what was done before)
+        feature_vocabulary = OrderedDict(sorted(feature_vocabulary.items()))
 
-        for feature_name, feature_values in feature_vocabulary.items():
-            feature_to_idx_dict[feature_name] = {
-                str(feature_value): feature_idx
+        # create the nested mapping
+        feature_to_idx_dict: Dict[Tuple[int, Text], Dict[Hashable, int]] = {}
```
wochinge, line 362: Pycharm gives me type warning in this line and 365. Can we please fix them?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,82 +87,283 @@ def required_components(cls) -> List[Type[Component]]:
...
+        feature_vocabulary: Dict[Text, Set[Hashable]] = dict()
         for example in training_data.training_examples:
-            tokens = example.get(TOKENS_NAMES[TEXT])
-            if tokens:
-                all_features.append(self._tokens_to_features(tokens))
+            tokens = example.get(TOKENS_NAMES[TEXT], [])
+            sentence_features = self._map_tokens_to_raw_features(tokens)
+            for token_features in sentence_features:
+                for position_and_feature_name, feature_value in token_features.items():
+                    feature_vocabulary.setdefault(position_and_feature_name, set()).add(
```
wochinge, line 285: also getting some type warnings by PyCharm here and in 289

wochinge (commented):
> Yep, but the validate_* logic should remain in the classes, no?

Unfortunately I can't respond in the comment for some reason. @joejuzl and I wanted to use protocols but classes are probably as good for now. Makes sense to keep the validation logic this in the class 👍🏻 

Thanks for refactoring this @ka-bu ! Just for the sake of reviewing: Can we have separate PRs in the future for migrating and refactorings so it's easier for the reviewers to see what's going on?

wochinge:
@koernerfelicia Are you okay with the changes to the ML related bits? I haven't looked at them in depth (probably also the wrong person for this)

rasa/nlu/featurizers/featurizer.py
ka-bu: actually yeah, we can 👍  - we don't really need that to be available now and maybe will never because we leave the validate_ here

ka-bu:
> Can we have separate PRs in the future for migrating and refactorings so it's easier for the reviewers to see what's going on?

sorry, yeah, thought this would still be a quick one 😅 

@dakshvar22 you were familiar with the lexical syntactic featurizer. Spotted anything that deviates now? Any comments on the sentence vectors? Do we want to create an issue for that?

rasa/nlu/featurizers/featurizer.py
ka-bu: good idea

rasa/nlu/featurizers/featurizer.py
ka-bu: put that below the rest on purpose because it's not validating a single featurizer config but *only* whether they're compatible with each other

rasa/utils/io.py
ka-bu: 👍

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: ... with lots of ` instead of ' :D

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: I was reading this as code (" When including code inside a docstring, use back-ticks ("`") to surround the text (like you would do in Markdown). "). The examples in notion use ' around e.g. printouts of constants. I guess here I could just remove the ` entirely - but ` should be ok, no?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: should be fixed now - input was still Set[Text] instead of Set[Hashable]

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: 👍   forgot to adapt some key here, they're not tuples of window index and feature name

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: oh, hashable is not enough here

ka-bu: casted all raw feature values to str and replaced Hashable by Text (because we need "Comparable" as well)

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
```diff
@@ -1,197 +1,288 @@
...
+        # simple example 1
         (
             "hello goodbye hello",
+            None,
+            [["BOS", "upper"], ["BOS", "EOS", "prefix2", "digit"], ["EOS", "low"],],
             [
                 [0.0, 0.0, 0.0, 0.0, 1.0, 1.0, 0.0, 1.0, 0.0, 1.0, 1.0, 0.0, 1.0],
                 [0.0, 1.0, 1.0, 1.0, 0.0, 1.0, 0.0, 1.0, 1.0, 0.0, 0.0, 1.0, 1.0],
                 [1.0, 0.0, 1.0, 1.0, 0.0, 0.0, 1.0, 1.0, 0.0, 1.0, 0.0, 0.0, 0.0],
-                [1.0, 1.0, 2.0, 2.0, 1.0, 2.0, 1.0, 3.0, 1.0, 2.0, 1.0, 1.0, 2.0],
```
dakshvar22, line 27: These were CLS token features that are no longer computed?

dakshvar22 (commented):
Added some questions / comments..

Specific questions from slack:

> do we want to add some sentence featurizer?

What kind of a sentence featurizer do you mean? Do you mean add sentence features like a max pool of all token features? Fair to look into it as a separate issue I'd say as this should just be a refactoring issue I believe.

> Are we aware and ok with e.g. a “prefix5” being a string of length < 5?

I think this has come up in the past (not as something that is hurting the performance but just as an observation) but we haven't dug deeper into the consequences. Fair to create an issue for investigation 👍

dakshvar22:
> It is used here: You can specify mean or max pooling for spacy featurizer

Ahh right! So much flexibility 😅 
Sounds good to have the code in there for it then.

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
ka-bu: they were never computed - the old comparison had a "[:-1]" to exclude these and checked whether sentence embeddings are None

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
...
         "suffix2": lambda token: token.text[-2:],
         "suffix1": lambda token: token.text[-1:],
-        "pos": lambda token: token.data.get(POS_TAG_KEY)
-        if POS_TAG_KEY in token.data
-        else None,
+        "pos": lambda token: token.data.get(POS_TAG_KEY, None),
         "pos2": lambda token: token.data.get(POS_TAG_KEY, [])[:2]
-        if "pos" in token.data
+        if POS_TAG_KEY in token.data
         else None,
```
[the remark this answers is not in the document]
ka-bu: we're not using the validate_* functions yet, so I'd keep that for now - but good point :)

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
...
+        if feature_name not in cls.SUPPORTED_FEATURES:
+            raise ValueError(
+                f"Configured feature '{feature_name}' not valid. Please check "
+                f"'{DOCS_URL_COMPONENTS}' for valid configuration parameters."
+            )
+        if feature_name == END_OF_SENTENCE:
+            return str(token_position == num_tokens - 1)
+        if feature_name == BEGIN_OF_SENTENCE:
+            return str(token_position == 0)
+        return str(cls._FUNCTION_DICT[feature_name](token))
```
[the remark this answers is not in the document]
ka-bu: That check was just a  logger.debug message which was pretty confusing imho because Nones where *not* ignored

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,133 @@
...
+
+        Args:
+          config: configuration
+          name: a name that can be used as identifier, in case the configuration does
+            not specify an `alias`
+        """
+        super().__init__()
+        self.validate_config(config)
+        self._config = {**self.get_default_config(), **config}
+        self._identifier = self._config.get(FEATURIZER_CLASS_ALIAS, name)
```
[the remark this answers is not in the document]
ka-bu: I can add a check to validate_config that avoid people setting `FEATURIZER_CLASS_ALIAS` to `None` 👍

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,133 @@
...
+    @classmethod
+    @abstractmethod
+    def validate_compatibility_with_tokenizer(cls, tokenizer: Tokenizer) -> None:
+        """Validate that the featurizer is compatible with the given tokenizer."""
+        # TODO: add (something like) this to recipe validation
+        # TODO: replace tokenizer by config of tokenizer to enable static check
+        ...
+
+    @abstractmethod
+    def process(self, messages: List[Message]) -> List[Message]:
```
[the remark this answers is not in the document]
ka-bu: because graph :)

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
         "suffix3": lambda token: token.text[-3:],
         "suffix2": lambda token: token.text[-2:],
         "suffix1": lambda token: token.text[-1:],
-        "pos": lambda token: token.data.get(POS_TAG_KEY)
-        if POS_TAG_KEY in token.data
-        else None,
```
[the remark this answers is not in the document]
ka-bu: if `POS_TAG_KEY` is not in ` token.data `, then `token.data.get(POS_TAG_KEY, None)` will return None
if `POS_TAG_KEY` i in ` token.data `, then `token.data.get(POS_TAG_KEY, None)` will return that element
... the new version is equivalent (and avoid searching for the POS_TAG_KEY key twice)

ka-bu:
> > do we want to add some sentence featurizer?
> 
> What kind of a sentence featurizer do you mean? Do you mean add sentence features like a max pool of all token features? Fair to look into it as a separate issue I'd say as this should just be a refactoring issue I believe.

👍  

> > Are we aware and ok with e.g. a “prefix5” being a string of length < 5?
> 
> I think this has come up in the past (not as something that is hurting the performance but just as an observation) but we haven't dug deeper into the consequences. Fair to create an issue for investigation 👍

and/or maybe just update the docs to point out that one does not need to add prefix2, prefix3, ... and so on 

will you create the (research) issues then?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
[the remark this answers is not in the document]
dakshvar22: Ohh my bad, sorry I didn't see you had a `None` as the default value.

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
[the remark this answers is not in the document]
dakshvar22: Okay should we add a TODO to simplify it once we use the validate_* function s?

rasa/nlu/featurizers/featurizer.py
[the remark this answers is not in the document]
dakshvar22: Is there an issue to keep a note of this?

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
wochinge: Oops. Backticks are the way to go then 👍🏻

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,133 @@
...
+
+    @classmethod
+    @abstractmethod
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
+        """Validates that the component is configured properly."""
+        ...
+
+    @classmethod
+    @abstractmethod
+    def validate_compatibility_with_tokenizer(cls, tokenizer: Tokenizer) -> None:
```
wochinge, line 57: tokenizer or `Type[Tokenizer]`? I'd actually prefer the later

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,133 @@
...
+            for config in featurizer_configs
+            if FEATURIZER_CLASS_ALIAS in config
+        )
+        if not alias_counter:  # no alias found
+            return
+        if alias_counter.most_common(1)[0][1] > 1:
+            raise InvalidConfigException(
+                f"Expected the featurizers to have unique names but found "
+                f" (name, count): {alias_counter.most_common()}. "
+                f"Please update your recipe such that each featurizer has a unique "
```
wochinge, line 131: actually that's part of the config then. The recipes will just pass that one. The names passed in via ExecutionContext are always unique
```suggestion
                f"Please update your config such that each featurizer has a unique "
```

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,133 @@
...
+            not specify an `alias` (or this `alias` is None)
+        """
+        super().__init__()
+        self.validate_config(config)
+        self._config = {**self.get_default_config(), **config}
+        self._identifier = self._config[FEATURIZER_CLASS_ALIAS] or name
+
+    @classmethod
+    @abstractmethod
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
```
wochinge, line 51: should we call this as part of the constructor? There is actually no need for the recipe to do this as the the component has everything it needs

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
...
+    @classmethod
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
+        """Validates that the component is configured properly."""
+        if FEATURES not in config:
+            return  # will be replaced with default
+        feature_config = config[FEATURES]
+        message = (
+            f"Expected configuration of `features` to be a list of lists that "
+            f"that contain names of lexical and syntactic features "
+            f" (i.e. {cls.SUPPORTED_FEATURES})."
```
wochinge, line 168:
```suggestion
            f" (i.e. {cls.SUPPORTED_FEATURES}). "
```

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
...
+                for pos_config in feature_config
+                for feature_name in pos_config
+            )
+        except TypeError as e:
+            raise InvalidConfigException(message) from e
+        if configured_feature_names.difference(cls.SUPPORTED_FEATURES):
+            raise InvalidConfigException(message)
+
+    @classmethod
+    def validate_compatibility_with_tokenizer(
```
wochinge, line 183: this has a different signature than the parent class.

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
...
-    def process(self, message: Message, **kwargs: Any) -> None:
-        self._create_sparse_features(message)
+        Args:
+          training_data: the training data
+        """
+        feature_to_idx_dict = self._create_feature_to_idx_dict(training_data)
+        self._set_feature_to_idx_dict(feature_to_idx_dict=feature_to_idx_dict)
+        if not self._feature_to_idx_dict:
+            rasa.shared.utils.io.raise_warning(
+                "No lexical syntactic features could be extracted from the training"
```
wochinge, line 257:
```suggestion
                "No lexical syntactic features could be extracted from the training "
```

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
...
+        Returns:
+           a nested mapping that maps from tuples of positions (in the window) and
+           supported feature names to "raw feature to index" mappings, i.e.
+           mappings that map the respective raw feature values to unique indices
+           (where `unique` means unique with respect to all indices in the
+           *nested* mapping)
+        """
+        # Note that this will only sort the top level keys - and we keep
+        # doing it to ensure consistently with what was done before)
+        feature_vocabulary = OrderedDict(sorted(feature_vocabulary.items()))
```
wochinge, line 357: I'm still getting type warnings in 360 and 364. I think you need to add types to this variable

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +87,415 @@ def required_components(cls) -> List[Type[Component]]:
...
+        """
+        for message in messages:
+            self._process_message(message)
+        return messages
 
-    def _create_sparse_features(self, message: Message) -> None:
-        """Convert incoming messages into sparse features."""
-        import scipy.sparse
+    def _process_message(self, message: Message) -> None:
+        """Featurize the given message in-place.
```
wochinge, line 385:
```suggestion
        """Featurizes the given message in-place.
```

dakshvar22:
> diet doing something particularly clever with the last sequence token instead of the non-existend sentence token)

Doesn't DIET create a dummy vector of zeros if sentence features are missing? I think you added ( / are adding) it in the PR meant to fix the response selector, right?

dakshvar22:
> will you create the (research) issues then?

Would be great if you can create as a follow up of this PR and send it to the research inbox then. 🙂

rasa/nlu/featurizers/featurizer.py
ka-bu: added to #9573

rasa/nlu/featurizers/featurizer.py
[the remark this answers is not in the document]
ka-bu: It is covered in #9573 as a thing that needs to be validated

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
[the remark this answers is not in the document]
ka-bu: np :)

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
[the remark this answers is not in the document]
ka-bu: 👍  just added a reminder to #9573

rasa/nlu/featurizers/featurizer.py
ka-bu: was thinking that we can check this before actually instantiating anything

ka-bu: but yeah I should add this to the constructor definitely as well!

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: oh ok, will do

ka-bu:
> > will you create the (research) issues then?
> 
> Would be great if you can create as a follow up of this PR and send it to the research inbox then. 🙂

done #9574 :)  (not sure about tags?)

rasa/nlu/featurizers/featurizer.py
ka-bu: and actually I do that already :)

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: done

rasa/nlu/featurizers/featurizer.py
ka-bu: is now fixed to `validate_compatibility_with_tokenizer(config, tokenizer_type)` (and added a test I forgot 😬 )

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: done -> #9574

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,135 @@
...
+        """Validates that the component is configured properly."""
+        ...
+
+    @classmethod
+    @abstractmethod
+    def validate_compatibility_with_tokenizer(
+        cls, config: Dict[Text, Any], tokenizer_type: Type[Tokenizer]
+    ) -> None:
+        """Validate that the featurizer is compatible with the given tokenizer."""
```
wochinge, line 60:
```suggestion
        """Validates that the featurizer is compatible with the given tokenizer."""
```

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +88,424 @@ def required_components(cls) -> List[Type[Component]]:
...
+        Args:
+          feature_name: the name of a supported feature
+          token: the token from which we want to extract the feature
+          token_position: the position of the token inside the tokenized text
+          num_tokens: the total number of tokens in the tokenized text
+        Returns:
+          the raw feature value as text
+        """
+        if feature_name not in cls.SUPPORTED_FEATURES:
+            raise ValueError(
```
wochinge, line 118:
```suggestion
            raise InvalidConfigException(
```

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +88,424 @@ def required_components(cls) -> List[Type[Component]]:
...
+    @classmethod
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
+        """Validates that the component is configured properly."""
+        if FEATURES not in config:
+            return  # will be replaced with default
+        feature_config = config[FEATURES]
+        message = (
+            f"Expected configuration of `features` to be a list of lists that "
+            f"that contain names of lexical and syntactic features "
+            f" (i.e. {cls.SUPPORTED_FEATURES}). "
```
wochinge, line 169:
```suggestion
            f"(i.e. {cls.SUPPORTED_FEATURES}). "
```

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +88,424 @@ def required_components(cls) -> List[Type[Component]]:
...
+                    f" Will continue without these features."
+                )
+
+    def train(self, training_data: TrainingData) -> Resource:
+        """Trains the featurizer.
 
-    def process(self, message: Message, **kwargs: Any) -> None:
-        self._create_sparse_features(message)
+        Args:
+          training_data: the training data
```
wochinge, line 256: return value is missing from docstring

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +88,424 @@ def required_components(cls) -> List[Type[Component]]:
...
+           supported feature names to "raw feature to index" mappings, i.e.
+           mappings that map the respective raw feature values to unique indices
+           (where `unique` means unique with respect to all indices in the
+           *nested* mapping)
+        """
+        # Note that this will only sort the top level keys - and we keep
+        # doing it to ensure consistently with what was done before)
+        ordered_feature_vocabulary: OrderedDict[
+            Tuple[int, Text], Set[Text]
+        ] = OrderedDict(sorted(feature_vocabulary.items()))
```
wochinge, line 364: pycharm actually still complains but it's a false one and should be clear from the type definitions 🤦🏻

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
wochinge: Thanks! added it to the Research project 👍🏻

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
```diff
@@ -69,241 +88,424 @@ def required_components(cls) -> List[Type[Component]]:
...
+                return cls(
+                    config=config,
+                    model_storage=model_storage,
+                    resource=resource,
+                    execution_context=execution_context,
+                    feature_to_idx_dict=feature_to_idx_dict,
+                )
+        except ValueError:
+            logger.warning(
+                f"Failed to load {cls.__class__.__name__} from model storage. Resource "
```
wochinge, line 491:
```suggestion
                f"Failed to load `{cls.__class__.__name__}` from model storage. Resource "
```

wochinge (approved):
💯 please double check with @koernerfelicia / @dakshvar22 but thumbs up from me 👍🏻

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
wochinge: I dunno. I think it's just not smart enough. Fine for now

wochinge:
Thanks for creating the recipe validation issue!

rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py
ka-bu: done

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
ka-bu: ... so actually yeah, probably at some point there was some "sum" aggregation it seems :D

dakshvar22: > so actually yeah, probably at some point there was some "sum" aggregation it seems

Sorry, I'm confused. So, we should add the "sum" aggregation or we shouldn't? :)

dakshvar22: Okay I see what you mean.. The test originally just excluded the last row of the expected array anyways 🤦

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
```diff
@@ -1,197 +1,312 @@
...
-            ]
-        }
+def test_feature_computation(
+    create_lexical_syntactic_featurizer: Callable[
+        [Dict[Text, Any]], LexicalSyntacticFeaturizerGraphComponent
+    ],
+    sentence: Text,
+    part_of_speech: Optional[List[Text]],
+    feature_config: List[List[Text]],
+    expected_features: List[List[int]],
```
dakshvar22, line 105: Since one of the test cases is just `List[int]`

```suggestion
    expected_features: List[Union[int, List[int]]],
```

dakshvar22 (approved):
LGTM, thanks for answering all my questions 👍

rasa/nlu/featurizers/featurizer.py
```diff
@@ -1,48 +1,135 @@
...
+        """Instantiates a new featurizer.
+
+        Args:
+          config: configuration
+          name: a name that can be used as identifier, in case the configuration does
+            not specify an `alias` (or this `alias` is None)
+        """
+        super().__init__()
+        self.validate_config(config)
+        self._config = {**self.get_default_config(), **config}
```
joejuzl, line 46: This shouldn't be needed as long as the subclass incorporates this classes default_config into it's own right?

wochinge: good catch 👍🏻

rasa/nlu/featurizers/featurizer.py
joejuzl: Does this make sense in the base featurizer class though? There are featurizers which don't do any config validation.

rasa/nlu/featurizers/featurizer.py
ka-bu: you mean because the graph node node does L.46 already? yep - feel free to remove (even though I'd still vote for that to happen in components as it keep confusing me that you could instantiate a component with an empty config and *not* get one with the default config :D)

tests/nlu/featurizers/test_lexical_syntactic_featurizer.py
ka-bu: yep, I was surprised as well :)

PR #9540, opened 2021-09-06
Use concurrency group to cancel workflows

HotThoughts:
**Proposed changes**:
- Use GitHub's internal concurrency feature to cancel workflows under the same concurrency group


erohmensing (approved):
Thank ya! I'm thinking [this one](https://github.com/RasaHQ/rasa/blob/main/.github/workflows/documentation.yml) could probably make use of it too!

HotThoughts:
@erohmensing done 😎

PR #9548, opened 2021-09-06
Build out NLUPredictionToHistoryAdder

usc-m:
**Proposed changes**:
- Build out NLUPredictionToHistoryAdder (#9327 )


rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,53 @@
...
+        for message in predictions:
+            # Need to find out where to get time from
+            # Message contains time but it is an int, we want a float
+            # What is domain for?
+            user_utterance = UserUttered(
+                message.data.get(TEXT),
+                message.data.get(INTENT),
+                message.data.get(ENTITIES),
+                message.data,
+                None,
```
usc-m, line 46: Is this fine to leave as None - it's the timestamp. `Message` gives us an `int` timestamp but this expects `float`, is there a conversion?

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,53 @@
...
+                original_message.input_channel,
+                message.data.get("message_id"),
+                message.data.get("metadata"),
+            )
+            user_utterance.apply_to(dialogue_tracker)
```
usc-m, line 51: Is this the correct way of adding the utterance to the tracker? This one doesn't seem to use the `Domain` so I'm not sure what purpose that serves

usc-m:
Still TODO: 
- Write out the tests (check empty list of messages, list of some messages) - I don't think this needs to be fancy and involve a lot of the graph machinery. Is it sufficient to just create one of these components and interact with it directly or should I wrap it in a Node? 
- Document and remove comments addressed to self

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
wochinge: @usc-m I'd exactly do what we currently do [here](https://github.com/RasaHQ/rasa/blob/f58acdea8408563c428282d3159b11d3973acacd/rasa/core/processor.py#L598). This means
- use `tracker.update(event, domain)`
- don't pass a timestamp 

Might also be worth to copy over the logging statements from the `processor` code.

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,70 @@
...
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> NLUPredictionToHistoryAdder:
+        """Creates component (see parent class for full docstring)."""
+        return cls()
+
+    def add(
+        self,
+        predictions: List["Message"],
```
wochinge, line 35: Can we please not use forward references? We check the graph schema using the type annotations and forward references tend to break that
```suggestion
        predictions: List[Message],
```

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,70 @@
...
+        tracker: "DialogueStateTracker",
```
wochinge, line 36:
```suggestion
        tracker: DialogueStateTracker,
```

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,70 @@
...
+        domain: Domain,
+        original_message: UserMessage,
+    ) -> "DialogueStateTracker":
```
wochinge, line 39:
```suggestion
    ) -> DialogueStateTracker:
```

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,70 @@
...
+        for message in predictions:
+            user_utterance = UserUttered(
+                message.data.get(TEXT),
+                message.data.get(INTENT),
+                message.data.get(ENTITIES),
+                message.data,
+                None,
+                original_message.input_channel,
+                message.data.get("message_id"),
+                message.data.get("metadata"),
```
wochinge, line 61: I'd actually prefer to get this from the `original_message` I'm not a 100% sure if this currently kept during NLU processing

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,70 @@
...
+            )
+            tracker.update(user_utterance, domain)
+
+        logger.debug(
```
wochinge, line 65: can we also take the slot debugging message over?

tests/test_graph_components/adders/test_nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,54 @@
...
+
+    assert len(tracker.events) == len(messages)
+    for i, message in enumerate(messages):
+        assert isinstance(tracker.events[i], UserUttered)
+        assert tracker.events[i].text == message.data.get(TEXT)
+        assert tracker.events[i].intent == message.data.get(INTENT)
+        assert tracker.events[i].entities == message.data.get(ENTITIES, [])
+        assert tracker.events[i].input_channel == original_message.input_channel
+        assert tracker.events[i].message_id == message.data.get("message_id")
+        assert tracker.events[i].metadata == message.data.get("metadata", {})
```
wochinge, line 54: optional: should we give the expected `UserUttered` events as part of the fixture? Doing it like this seems bit like tautological testing as you're re-using the code from the production code

usc-m: Yep, can redo it like that - I did think it was a bit tautological but I was trying to reduce the amount of test data that needed to be altered if the code changed, but since there's only a couple of messages it's probably not a big deal

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
usc-m: Gotcha, was not aware of that

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
usc-m: Can do - I didn't know if those would be the same so I erred on the side of caution and assumed the one attached to the `message` was more likely to be correct for that message (thinking the `original_message` might be out of date for some messages in the batch). In this case are the predictions all derived from the one `original_message` or are they different messages but we assume that they share channel and metadata?

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
usc-m: It might be a nice, small refactor to move the slot logging code somewhere else since it seems like a useful helper function to have - but I'm not sure if extracting it to some utils file somewhere is in scope for this PR? So I'm planning to just take the code as-is

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,78 @@
...
+            tracker: The tracker the predictions should be attached to
+            domain: The domain of the model.
+            original_message: An original message from the user with
+                extra metadata to annotate the predictions (e.g. channel)
+
+        Returns:
+            The original tracker updated with events created from the predictions
+        """
+        for message in predictions:
+            user_utterance = UserUttered(
```
ancalita, line 53: Optional: rename to `user_event` to avoid future confusion?

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
ancalita: +1 for removing timestamp and using keyword arguments for `input_channel`, `message_id`, `metadata`

tests/test_graph_components/adders/test_nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,65 @@
...
+):
+    component = NLUPredictionToHistoryAdder.create(
+        {**NLUPredictionToHistoryAdder.get_default_config()},
+        default_model_storage,
+        Resource("test"),
+        default_execution_context,
+    )
+
+    tracker = DialogueStateTracker("test", None)
+    domain = Domain.from_file(path="data/test_domains/travel_form.yml")
```
ancalita, line 53: You could also use the domain fixture instead.

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,78 @@
...
+                message.data.get(TEXT),
+                message.data.get(INTENT),
+                message.data.get(ENTITIES),
+                message.data,
```
ancalita, line 57: I actually think it would be best to leave this out because`UserUttered.parse_data` would get assigned [here](https://github.com/RasaHQ/rasa/blob/e152bba0f5c28ccb21041d2bcb8beb9786c00d48/rasa/shared/core/events.py#L450) with the correct metadata from the provided `original_message`, otherwise it might get overwritten [here](https://github.com/RasaHQ/rasa/blob/e152bba0f5c28ccb21041d2bcb8beb9786c00d48/rasa/shared/core/events.py#L461) if metadata in `message.data` is `None`?

ancalita (commented):
Left a few suggestions 👍🏼

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
wochinge: I think nothing else should call this. So copying it here is fine. I'd say it's part of the PR as it's required to preserve the existing behavior in the graph.

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
wochinge: Oh I see 👍🏻 It should always be just an empty list or a list with one message. Lists with multiple messages are actually not possible for this component.

tests/test_graph_components/adders/test_nlu_prediction_to_history_adder.py
usc-m: I did not know this existed - thanks!

rasa/graph_components/adders/nlu_prediction_to_history_adder.py
usc-m: That's a good spot - I wasn't sure if the `parse_data` was going to contain other info that needed to be preserved which is why I did it that way but that's a good point. I'll change it out since we're trying to preserve the `metadata` on the `original_message` deliberately so that's probably the priority

tests/test_graph_components/adders/test_nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,65 @@
...
+
+    assert len(tracker.events) == len(messages)
+    for i, message in enumerate(messages):
+        assert isinstance(tracker.events[i], UserUttered)
+        assert tracker.events[i].text == expected[i].text
+        assert tracker.events[i].intent == expected[i].intent
+        assert tracker.events[i].entities == expected[i].entities
+        assert tracker.events[i].input_channel == expected[i].input_channel
+        assert tracker.events[i].message_id == expected[i].message_id
+        assert tracker.events[i].metadata == expected[i].metadata
```
wochinge, line 65: How about this? Maybe in combination with something like `@freezegun.freeze_time("2018-01-01")`?
```suggestion
        assert tracker.events[i] == expected[i]
```

wochinge (approved):
Nice! 💯

tests/test_graph_components/adders/test_nlu_prediction_to_history_adder.py
usc-m: `freezegun` seems neat - I'm going to have to read up on that

tests/test_graph_components/adders/test_nlu_prediction_to_history_adder.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.mark.parametrize(
+    "messages,expected,input_channel",
+    [
+        ([], [], "slack"),
+        (
+            [
+                Message(
+                    {TEXT: "message 1", INTENT: {}, ENTITIES: [{}], "message_id": "1",}
+                ),
+                Message(
```
ancalita, line 27: Based on the comment from @wochinge that `List[Message]` can have 1 object max, should the test be updated to reflect this, i.e. remove 2nd message?

usc-m: Makes sense, also would want to drop the `metadata` one in light of the fact that isn't propagated through the `Message` any more

ancalita: Yup,also it would be a good idea to add metadata in your test when you do the `UserMessage` instantiation in L56?

ancalita (approved):
🙌🏼

PR #9562, opened 2021-09-07
Migrate jieba tokenizer for the graph

joejuzl:
**Proposed changes**:
Fixes: #9500 


rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -1,107 +1,51 @@
...
 
-class Tokenizer(Component):
+class TokenizerGraphComponent(GraphComponent, abc.ABC):
     """Base class for tokenizers."""
 
-    def __init__(self, component_config: Dict[Text, Any] = None) -> None:
-        """Construct a new tokenizer using the WhitespaceTokenizer framework."""
-        super().__init__(component_config)
-
+    def __init__(self, config: Dict[Text, Any]) -> None:
```
wochinge, line 30: optional: how about saving the config as part of the parent class?

rasa/nlu/tokenizers/tokenizer.py
```diff
@@ -1,107 +1,51 @@
...
-        self.intent_split_symbol = self.component_config.get("intent_split_symbol", "_")
+        self.intent_split_symbol = config["intent_split_symbol"]
         # token pattern to further split tokens
-        token_pattern = self.component_config.get("token_pattern", None)
+        token_pattern = config["token_pattern"]
         self.token_pattern_regex = None
         if token_pattern:
             self.token_pattern_regex = re.compile(token_pattern)
 
+    @abc.abstractmethod
```
wochinge, line 42: 🙌🏻

rasa/nlu/tokenizers/jieba_tokenizer.py
```diff
@@ -75,25 +103,25 @@ def tokenize(self, message: Message, attribute: Text) -> List[Token]:
...
+    ) -> JiebaTokenizerGraphComponent:
+        """Loads a custom dictionary from model storage."""
+        relative_dictionary_path = config["dictionary_path"]
 
-        # get real path of dictionary path, if any
+        # If a custom dictionary path is in the config we know that it should have
+        # been saved to the model storage.
         if relative_dictionary_path is not None:
-            dictionary_path = os.path.join(model_dir, relative_dictionary_path)
+            with model_storage.read_from(resource) as resource_directory:
```
wochinge, line 118: let's handle the `ValueError` here

rasa/nlu/tokenizers/jieba_tokenizer.py
```diff
@@ -75,25 +103,25 @@ def tokenize(self, message: Message, attribute: Text) -> List[Token]:
...
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
         **kwargs: Any,
-    ) -> "JiebaTokenizer":
-        """Loads trained component (see parent class for full docstring)."""
-        relative_dictionary_path = meta.get("dictionary_path")
+    ) -> JiebaTokenizerGraphComponent:
+        """Loads a custom dictionary from model storage."""
+        relative_dictionary_path = config["dictionary_path"]
```
wochinge, line 113: what guarantees that this is relative? I think we should be able to handle absolute paths as well (e.g. just loading everything in the `resource_directory` instead of doing weird name stuff)

tests/nlu/tokenizers/test_jieba_tokenizer.py
```diff
@@ -36,18 +50,62 @@ def test_jieba(text, expected_tokens, expected_indices):
...
+    # The dictionary has not been persisted yet.
+    with pytest.raises(ValueError):
+        JiebaTokenizerGraphComponent.load(
+            {**JiebaTokenizerGraphComponent.get_default_config(), **component_config},
+            default_model_storage,
+            resource,
+            default_execution_context,
+        )
+
+    tk.persist()
```
wochinge, line 90: shouldn't this be called as part of `train`?

rasa/nlu/tokenizers/jieba_tokenizer.py
joejuzl: This is just what they called it before. It's not really relative, it's just not in the model storage. If the dictionary path is set in the config it means they are using a custom dictionary and we would have saved it to the model storage. 
However we could also just always try to load whatever is in the model storage.

tests/nlu/tokenizers/test_jieba_tokenizer.py
joejuzl: This is what I was asking about in the slack thread. There isn't really a `train`, only a `process_training_data`. We could save it as part of that method, but it makes it a bit different from the usual patten. I don't really feel strongly either way though.

rasa/nlu/tokenizers/jieba_tokenizer.py
wochinge: > However we could also just always try to load whatever is in the model storage.
 Write

I'd just do that as we'd only move things to the model storage which were part of `dictionary_path`

tests/nlu/tokenizers/test_jieba_tokenizer.py
wochinge: same. Probably having separate things is cleaner tbh.

joejuzl: ok I will leave it separate for now, but it's a quick change later if we find it's easier w.r.t to the recipe (or whatever) to have them combined.

wochinge: Can we do a train and load test then?

rasa/nlu/tokenizers/jieba_tokenizer.py
joejuzl: However if there was no `dictionary_path` in the config we won't have created the resource so we will get a `ValueError`. And this error could signify either we forgot to persist, or we just didn't use custom dictionaries. Whereas if we keep this check, we know something is wrong if there is a custom dictionary path _and_ no resource.

tests/nlu/tokenizers/test_jieba_tokenizer.py
joejuzl: I'm not sure what you mean. This test does, `process_training_data`, `persist`, `load` then `process` which is the full lifecycle.

rasa/nlu/tokenizers/jieba_tokenizer.py
```diff
@@ -102,14 +134,9 @@ def copy_files_dir_to_dir(input_dir: Text, output_dir: Text) -> None:
...
-
-        # copy custom dictionaries to model dir, if any
-        if self.dictionary_path is not None:
-            target_dictionary_path = os.path.join(model_dir, file_name)
-            self.copy_files_dir_to_dir(self.dictionary_path, target_dictionary_path)
-
-            return {"dictionary_path": file_name}
-        else:
-            return {"dictionary_path": None}
+    def persist(self) -> None:
```
wochinge, line 137: don't we need to call this as part of `train`?

tests/nlu/tokenizers/test_jieba_tokenizer.py
wochinge: sorry, all good 🙌🏻

rasa/nlu/tokenizers/jieba_tokenizer.py
joejuzl: didn't we cover this here? https://github.com/RasaHQ/rasa/pull/9562#discussion_r703605128 😁 
(or am I missing something?)

wochinge: Wait - but something has to call `persist` then? I thought we'd add a `train` to the tokenizer where call `persist`

joejuzl: Oh you mean a `train` method which just calls `persist`. I was assuming we would call `persist` directly as it wont be called in the same way anyway. i.e. wont have any inputs or outputs.

wochinge: Ooooh, yes,we can also do that. For consistency (and less edge conditions in the recipe) a `train` would be nicer but whatever 👍🏻

wochinge (approved):
🔥

PR #9563, opened 2021-09-07
Adapt `CRFEntityExtractor` to graph component interface

### LINKED ISSUE #9506: Migrate `CRFEntityExtractor` to `GraphComponent` interface
**Overview of the Solution**:
Adapt the `CRFEntityExtractor ` to the new `GraphComponent` interface.

* adapt `create` and `load` functions to new signatures.
* implement `required_packages` and `supported_languages`.
* adapt persistence.
* split `train` into `train` and `process_training_data` (process the training data for the next components in the pipeline).
* support multiple messages in `process` (**make sure to handle empty messages**).
* Inherit from both `GraphComponent` _and_ `EntityExtractorMixin` (instead of the old `EntityExtractor`).

**Process to avoiding breaking the CI**
Use the approach showcased in the [TEDPolicy migration](https://github.com/RasaHQ/rasa/pull/9306) to avoid breaking the test suite.
Main points:
* Rename old files with a prefix i.e. from `some_component.py` to `_some_component.py`.
* make the changes in the current file e.g. `some_component.py`.
* In `some_component.py` import the old component from `_some_component.py` and assign like `SomeComponent = SomeComponent`
* Rename the _new_ component in `_some_component.py` to `SomeComponentGraphComponent`
* Adapt the old unit tests, i.e. `test_some_component.py` to test `SomeComponentGraphComponent`.

**Definition of Done**:
- [x] `CRFEntityExtractor` adapted to the new `GraphComponent` interface.
- [x] Unit tests are adapted to adapted component.
- [x] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old component implementation).

ancalita:
**Proposed changes**:
- closes #9506 


tests/nlu/extractors/test_crf_entity_extractor.py
```diff
@@ -93,7 +96,7 @@ async def test_train_persist_with_different_configurations(
     pipeline = [
         {"name": "SpacyNLP", "model": "en_core_web_md"},
         {"name": "SpacyTokenizer"},
-        {"name": "CRFEntityExtractor"},
+        {"name": "CRFEntityExtractorGraphComponent"},
```
ancalita, line 99: This test failed locally due to `ImportError: Cannot retrieve class from path CRFEntityExtractorGraphComponent.`
I'd appreciate any ideas how to adapt this one 🙏🏼

alwx: by the way, `test_train_persist_load_with_composite_entities` also needs to be adapted

alwx: checking why it doesn't work :)

alwx (changes requested):
1. In order to fix the issue with `test_train_persist_with_different_configurations` you need to add `CRFEntityExtractorGraphComponent` to `component_classes` in `rasa.nlu.registry`. However, that will also require changes to how we register components. Currently we do mapping from a component's name to its class to allow name based lookup:
```
registered_components: Dict[Text, Type[Component]] = {
    c.name: c for c in component_classes
}
```
This approach won't work with `GraphComponent`s just because there is no `c.name` property and because the types won't match.

2. `test_train_persist_load_with_composite_entities` also needs to work with `CRFEntityExtractorGraphComponent`

tests/nlu/extractors/test_crf_entity_extractor.py
wochinge: This is actually expected. You have migrated the component and it's no longer compatible with the old model architecture. You've to do something similar as Joe did [here](https://github.com/RasaHQ/rasa/blob/5bb331ff7ba75ad564478cbc71fd1a46bce5ee02/tests/nlu/classifiers/test_diet_classifier.py#L122). Or once the `SpacyTokenizer` is also migrated you could just run them one after another

ancalita: Thank you both for your suggestions 🙏🏼  The unit tests have been adapted and are now passing 🎉

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -121,20 +83,70 @@ def required_components(cls) -> List[Type[Component]]:
         "digit": lambda crf_token: crf_token.text.isdigit(),
         "pattern": lambda crf_token: crf_token.pattern,
         "text_dense_features": (
-            lambda crf_token: CRFEntityExtractor._convert_dense_features_for_crfsuite(
+            lambda crf_token: CRFEntityExtractorGraphComponent._convert_dense_features_for_crfsuite(
```
ancalita, line 86: I'm unsure how to modify this as to pass the code quality test 😬

wochinge: haha 😁 I'd do it like [here](https://github.com/RasaHQ/rasa/blob/562eef087924e5fe796151689e080b7f65156b06/rasa/shared/importers/autoconfig.py#L112) and add a `# noqa: E501`

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -121,20 +83,70 @@ def required_components(cls) -> List[Type[Component]]:
...
+        resource: Resource,
+        execution_context: ExecutionContext,
         entity_taggers: Optional[Dict[Text, "CRF"]] = None,
     ) -> None:
-
-        super().__init__(component_config)
+        self.component_config = config
+        self._model_storage = model_storage
+        self._resource = resource
+        self._execution_context = execution_context
```
wochinge, line 149: is this one actually needed?

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -155,24 +169,25 @@ def _validate_configuration(self) -> None:
...
-                "No training examples with entities present. Skip training"
-                "of 'CRFEntityExtractor'."
-            )
-            return
+    @staticmethod
+    def required_packages() -> List[Text]:
+        """Any extra python dependencies required for this component to run."""
+        return ["sklearn_crfsuite", "sklearn"]
 
+    def process_training_data(
```
wochinge, line 187: This method is actually no needed during training time as entities are already annotated in the training data

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -306,57 +346,51 @@ def _tag_confidences(
...
 
-        if not file_names:
-            logger.debug(
-                f"Failed to load model for 'CRFEntityExtractor'. "
-                f"Maybe you did not provide enough training data and no model was "
-                f"trained or the path '{os.path.abspath(model_dir)}' doesn't exist?"
-            )
-            return cls(component_config=meta)
+                for file_name in file_names:
+                    name = file_name.name.split(".")[-2]
```
wochinge, line 371:
```suggestion
                    name = file_name.stem
```

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -306,57 +346,51 @@ def _tag_confidences(
...
-                file_name = f"{file_name}.{name}.pkl"
-                model_file_name = os.path.join(model_dir, file_name)
-                joblib.dump(entity_tagger, model_file_name)
-                file_names[name] = file_name
-
-        return {"files": file_names}
+        with self._model_storage.write_to(self._resource) as model_dir:
+            if self.entity_taggers:
+                for name, entity_tagger in self.entity_taggers.items():
+                    file_name = self.__class__.__name__
```
wochinge, line 391: I think we can dump the class name as this is always the same

rasa/nlu/extractors/crf_entity_extractor.py
wochinge: see my other comment in `persist`

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -186,8 +201,27 @@ def train(
 
         dataset = [self._convert_to_crf_tokens(example) for example in entity_examples]
 
+        return dataset
+
+    def train(self, training_data: TrainingData,) -> Resource:
+        """Train the extractor on a data set."""
```
wochinge, line 207:
```suggestion
        """Trains the extractor on a data set."""
```

tests/nlu/extractors/test_crf_entity_extractor.py
```diff
@@ -1,47 +1,60 @@
...
-    _config = RasaNLUModelConfig({"pipeline": pipeline, "language": "en"})
+    tokenizer = WhitespaceTokenizer()
+    tokenizer.train(training_data)
 
-    (trainer, trained, persisted_path) = await rasa.nlu.train.train(
-        _config,
-        path=str(tmp_path),
-        data="data/test/demo-rasa-composite-entities.yml",
-        component_builder=component_builder,
+    crf_extractor = CRFEntityExtractorGraphComponent.create(
```
wochinge, line 33: optional: You could a fixture like I did [here](https://github.com/RasaHQ/rasa/blob/fc27727128b108c871394e375e219a7e27fabd35/tests/nlu/tokenizers/test_mitie_tokenizer.py#L17) to shorten this

wochinge (commented):
A few comments - looking great otherwise 🚀 💯

rasa/nlu/extractors/crf_entity_extractor.py
ancalita: It was previously part of training, I only extracted this section into this new method. I think it's required because training works with `CRFTokens`, not `TrainingData` objects 🤔

wochinge: Sure, but can't we just move the code from `process_training_data` to `train` instead and dump `process_training_data`? Having `process_training_data` makes it look like anything after it would need this information

ancalita: Aaa, in other words, keep it as it used to be 😄 The only reason I had extracted was due to the [issue description](https://github.com/RasaHQ/rasa/issues/9506), but will revert this change.

rasa/nlu/extractors/crf_entity_extractor.py
```diff
@@ -121,20 +83,69 @@ def required_components(cls) -> List[Type[Component]]:
...
         self,
-        component_config: Optional[Dict[Text, Any]] = None,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
         entity_taggers: Optional[Dict[Text, "CRF"]] = None,
     ) -> None:
-
-        super().__init__(component_config)
+        """Declare instance variables with default values."""
```
wochinge, line 145: I don't understand the docstring: which default values?

wochinge (approved):
💯 Great work!

tests/nlu/extractors/test_crf_entity_extractor.py
```diff
@@ -176,11 +223,12 @@ def test_crf_use_dense_features(spacy_nlp: Any):
     ],
 )
 def test_most_likely_entity(
+    crf_entity_extractor: Callable[[Dict[Text, Any]], CRFEntityExtractorGraphComponent],
     entity_predictions: List[Dict[Text, float]],
     expected_label: Text,
     expected_confidence: float,
 ):
-    crf_extractor = CRFEntityExtractor({"BILOU_flag": True})
```
wochinge, line 183: we wanna keep that config, don't we?

rasa/nlu/extractors/crf_entity_extractor.py
ancalita: Ah I copied this over from another component, I'll change the docstring instead.

tests/nlu/extractors/test_crf_entity_extractor.py
ancalita: Oups, missed this! Will correct.

alwx (approved):
🚀

PR #9565, opened 2021-09-07
3.0 architecture revamp/9507/mitie entity extractor

### LINKED ISSUE #9507: Migrate `MitieEntityExtractor` to `GraphComponent` interface
**Overview of the Solution**:
Adapt the `MitieEntityExtractor ` to the new `GraphComponent` interface.

* adapt `create` and `load` functions to new signatures.
* implement `required_packages` and `supported_languages`.
* adapt persistence.
* convert `train` into `process_training_data` (this component does not train).
* support multiple messages in `process` (**make sure to handle empty messages**).
* Inherit from both `GraphComponent` _and_ `EntityExtractorMixin` (instead of the old `EntityExtractor`).
* see adapt MitieNLP for graph #9443 and 3.0 architecture revamp/9407/mitie intent classifier adaption #9444 for how to get mitie model

**Process to avoiding breaking the CI**
Use the approach showcased in the [TEDPolicy migration](https://github.com/RasaHQ/rasa/pull/9306) to avoid breaking the test suite.
Main points:
* Rename old files with a prefix i.e. from `some_component.py` to `_some_component.py`.
* make the changes in the current file e.g. `some_component.py`.
* In `some_component.py` import the old component from `_some_component.py` and assign like `SomeComponent = SomeComponent`
* Rename the _new_ component in `_some_component.py` to `SomeComponentGraphComponent`
* Adapt the old unit tests, i.e. `test_some_component.py` to test `SomeComponentGraphComponent`.

**Definition of Done**:
- [ ] `MitieEntityExtractor` adapted to the new `GraphComponent` interface.
- [ ] Unit tests are adapted to adapted component.
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old component implementation).

ka-bu:
**Proposed changes**:
- closes #9507 


rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -1,89 +1,130 @@
...
+        Args:
+            config: The configuration.
+            model_storage: Storage which graph components can use to persist and load
+                themselves.
+            resource: Resource locator for this component which can be used to persist
+                and load itself from the `model_storage`.
+            ner: Mitie named entity extractor
+        """
+        # graph component
+        self._config = {**self.get_default_config(), **config}
```
joejuzl, line 71: This is already handled by the `GraphNode` (rasa/engine/graph.py:315)

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -1,89 +1,130 @@
...
+
+        Args:
+            config: The configuration.
+            model_storage: Storage which graph components can use to persist and load
+                themselves.
+            resource: Resource locator for this component which can be used to persist
+                and load itself from the `model_storage`.
+            ner: Mitie named entity extractor
+        """
+        # graph component
```
joejuzl, line 70: Do we need this comment?

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -1,89 +1,130 @@
...
+        self._config = {**self.get_default_config(), **config}
+        self._model_storage = model_storage
+        self._resource = resource
+        self.validate_config(self._config)
+        # extractor
```
joejuzl, line 75: Maybe a better named variable would make this comment redundant?

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -1,89 +1,130 @@
...
         import mitie
 
-        model_file = kwargs.get("mitie_file")
-        if not model_file:
-            raise Exception(
-                "Can not run MITIE entity extractor without a "
-                "language model. Make sure this component is "
-                "preceeded by the 'MitieNLP' component."
-            )
+        trainer = mitie.ner_trainer(str(mitie_model.model_path))
```
joejuzl, line 124: aha so we can't actually re-use the loaded model? At this point the path could just be in the config I guess...

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -133,50 +187,105 @@ def _prepare_mitie_sample(training_example: Message) -> Any:
...
+        tokens = message.get(TOKENS_NAMES[TEXT])
+
+        entities = []
+        token_texts = [token.text for token in tokens]
+        mitie_entities = self._ner.extract_entities(
+            token_texts, mitie_model.word_feature_extractor,
         )
-        extracted = self.add_extractor_name(ents)
-        message.set(ENTITIES, message.get(ENTITIES, []) + extracted, add_to_output=True)
+        for e in mitie_entities:
```
joejuzl, line 237: Where is this new code coming from?

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -133,50 +187,105 @@ def _prepare_mitie_sample(training_example: Message) -> Any:
...
-        classifier_file = os.path.join(model_dir, file_name)
-        if os.path.exists(classifier_file):
-            extractor = mitie.named_entity_extractor(classifier_file)
-            return cls(meta, extractor)
-        else:
-            return cls(meta)
-
-    def persist(self, file_name: Text, model_dir: Text) -> Optional[Dict[Text, Any]]:
+        try:
+            with model_storage.write_to(resource) as model_path:
```
joejuzl, line 266:
```suggestion
            with model_storage.read_from(resource) as model_path:
```

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -133,50 +187,105 @@ def _prepare_mitie_sample(training_example: Message) -> Any:
...
+                ner_file = model_path / cls.MITIE_RESOURCE_FILE
+                if not ner_file.exists():
+                    raise FileNotFoundError(
+                        f"Expected a MITIE extractor file at {ner_file}."
+                    )
+                ner = mitie.named_entity_extractor(str(ner_file))
+                return MitieEntityExtractorGraphComponent(
+                    config, model_storage, resource, ner=ner
+                )
+        except FileNotFoundError:
```
joejuzl, line 276: Should we catch `ValueError` too?

tests/nlu/extractors/test_mitie_entity_extractor.py
```diff
@@ -0,0 +1,181 @@
...
+        message.data[ENTITIES] = entities
+        if with_trainable_examples:
+            message.data[INTENT] = "must have intent otherwise not an NLU example"
+        else:
+            pass  # not adding an intent is sufficient to make this a "core example"
+        messages_with_entities.append(message)
+
+    # turn them into training data
+    training_data = TrainingData(messages_with_entities)
+    training_data.training_examples = messages_with_entities
```
joejuzl, line 110: Do we need this line?

joejuzl (changes requested):

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: I can remove the comments

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: i think "ner" is fine - the comments where just to sort out that there is stuff that a common graph component base class could do

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: we use it in process

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: this isn't new - it's `extract_entities` in the old code (without underscore)

tests/nlu/extractors/test_mitie_entity_extractor.py
ka-bu: nope :)

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: the old implementation doesn't - and mitie.py actually raises a generic `Exception` ... I've just added an except that catches that mitie exception

rasa/nlu/extractors/mitie_entity_extractor.py
joejuzl: oh ok I see, the re-ordering of the methods makes it hard to see what has changed.

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -133,50 +185,114 @@ def _prepare_mitie_sample(training_example: Message) -> Any:
...
+                    raise FileNotFoundError(
+                        f"Expected a MITIE extractor file at {ner_file}."
+                    )
+                ner = mitie.named_entity_extractor(str(ner_file))
+                return MitieEntityExtractorGraphComponent(
+                    config, model_storage, resource, ner=ner
+                )
+        except (FileNotFoundError, Exception) as e:
+            if not isinstance(e, FileNotFoundError):
+                is_mitie_exception = not (
```
joejuzl, line 276: Why do we need this? Can't we just have different `except` clauses?

joejuzl (changes requested):

rasa/nlu/extractors/mitie_entity_extractor.py
joejuzl: `model_storage.read_from` throws it if the resource doesn't exist yet, and we've been handling it in all the other `load`s.

ka-bu: ah ok, wasn't aware of that 👍

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: In both cases I want to create the default instance

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: oh yeah sorry about that - the extract is not public anymore so I thought it was strange to have that first and before the place where it's used

rasa/nlu/extractors/mitie_entity_extractor.py
joejuzl: It just seems a bit brittle to condition on error message text from code we don't control. I'm also a bit confused, why is this called `is_mitie_exception` if there is the `not` it seems like that exception comes from mitie? Or am I missing something?

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -133,50 +185,114 @@ def _prepare_mitie_sample(training_example: Message) -> Any:
...
-            return cls(meta)
+        try:
+            with model_storage.read_from(resource) as model_path:
+                ner_file = model_path / cls.MITIE_RESOURCE_FILE
+                if not ner_file.exists():
+                    raise FileNotFoundError(
+                        f"Expected a MITIE extractor file at {ner_file}."
+                    )
+                ner = mitie.named_entity_extractor(str(ner_file))
+                return MitieEntityExtractorGraphComponent(
```
joejuzl, line 271:
```suggestion
                return cls(
```

rasa/nlu/extractors/mitie_entity_extractor.py
```diff
@@ -133,50 +185,114 @@ def _prepare_mitie_sample(training_example: Message) -> Any:
...
+                    raise
 
-    def persist(self, file_name: Text, model_dir: Text) -> Optional[Dict[Text, Any]]:
+            rasa.shared.utils.io.raise_warning(
+                f"Failed to load {cls.__name__} from model storage. "
+                f"This can happen if the model could not be trained because regexes "
+                f"could not be extracted from the given training data - and hence "
+                f"could not be persisted. Error: {e}."
+            )
+            return MitieEntityExtractorGraphComponent(config, model_storage, resource)
```
joejuzl, line 290:
```suggestion
            return cls(config, model_storage, resource)
```

joejuzl (changes requested):

rasa/nlu/extractors/mitie_entity_extractor.py
ka-bu: oh yep that not was too much :) (had that statement in the if before) - yes it is brittle but it only catches things with that exact message (but I can also remove it if we don't expect any errors from mitie directly)

joejuzl: another option is to have a separate `try except` just around `mitie.named_entity_extractor(str(ner_file)), which raises a specific warning. And then just handle `ValueError` in this one (similarly to how we do in other places we use `model_storage.read_from`)

ka-bu: you mean raising a ValueError from the Exception (in case that comes from mitie and then handle that)? (just updated code accordingly)

joejuzl: I was thinking this:

```python
        try:
            with model_storage.read_from(resource) as model_path:
                ner_file = model_path / cls.MITIE_RESOURCE_FILE
                if not ner_file.exists():
                    raise FileNotFoundError(
                        f"Expected a MITIE extractor file at {ner_file}."
                    )

                try:
                    ner = mitie.named_entity_extractor(str(ner_file))
                except Exception:
                        f"Mitie wasn't able to load the MITIE extractor from "
                        f"{ner_file}."
                    )
                    return cls(config, model_storage, resource, ner=ner)

        except ValueError as e:
                f"Failed to load {cls.__name__} from model storage. "
                f"This can happen if the model could not be trained because regexes "
                f"could not be extracted from the given training data - and hence "
                f"could not be persisted. Error: {e}."
            )
            return cls(config, model_storage, resource)

```

ka-bu: Wouldn't this also catch keyboard interrupts happening during loading of the mitie model?

joejuzl: No, `KeyboardInterrupt` inherits from `BaseException` but not `Exception`.

joejuzl: tbh though I would be fine with not catching the mitie exception. Better to fail fast than suppress an error that shouldn't happen.

ka-bu: I'll just remove it - we're not catching this stuff anywhere else I think. I was just triggered to look into it when you mentioned value error and then was wondering if that could come from there.

> inherits from BaseException

oh, nice that they made that... exception :D

joejuzl: > oh, nice that they made that... exception :D

😂

joejuzl (approved):
💯

Issue #9345, opened 2021-08-13
Adapt `RulePolicy` to new `GraphComponent` interface


joejuzl:
**Overview of the Solution**:
Adapt the `RulePolicy` to the new `GraphComponent` interface.
See the [implementation proposal](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e) for further details.

* adapt `create` and `load` functions to new signatures
* implement `required_packages` and `supported_languages`
* adapt persistence
* add default param for `interpreter` so graph doesn't need to fill it (this is a workaround)
* adapt how config is passed into the component (and default config)

**Blockers**
- Adapt `MemoizationPolicy` #9344

**Definition of Done**:
- [ ] `RulePolicy` adapted to the new GraphComponent interface
- [ ] Unit tests are adapted to adaption of policy
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old policy implementation)

ancalita:
@joejuzl i see that that the Memoization Policy PR opened by Thomas is still in progress - should I branch off his branch and start working on this issue?

joejuzl:
> @joejuzl i see that that the Memoization Policy PR opened by Thomas is still in progress - should I branch off his branch and start working on this issue

Is there another issue you can pick up in the meantime?

ancalita:
> Is there another issue you can pick up in the meantime?

@joejuzl yup will start the `DomainProvider` & `NLUMessageConverter` ones.

PR #9576, opened 2021-09-08
migrate `RulePolicy`

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9345
- migrate `RulePolicy`
- implement `RuleOnlyDataProvider`
- (rename `test_graph_components` to `graph_components`)


rasa/core/policies/rule_policy.py
```diff
@@ -107,74 +116,69 @@ def supported_data() -> SupportedData:
...
+        self._check_for_contradictions = config["check_for_contradictions"]
 
-        rule_policy = next(
-            (p for p in ensemble.policies if isinstance(p, RulePolicy)), None
-        )
-        if not rule_policy or not rule_policy._enable_fallback_prediction:
-            return
+        self._rules_sources = defaultdict(list)
 
+    def _validate_against_domain(self, domain: Optional[Domain]) -> None:
```
ka-bu, line 175: why is/was domain optional? 🤔

rasa/core/policies/ted_policy.py
```diff
@@ -994,20 +1000,32 @@ def load(
         model_storage: ModelStorage,
         resource: Resource,
         execution_context: ExecutionContext,
+        rule_only_data: Optional[Dict[Text, Any]] = None,
```
ka-bu, line 1003: Hm, this is ok because we inspect all methods including the `load` and hence will notice that it is `rule_only_data` that needs to be provided and not a kwarg, right?

rasa/core/policies/rule_policy.py
ka-bu: why is/was domain optional? 🤔

rasa/core/policies/memoization.py
```diff
@@ -223,18 +230,14 @@ def _prediction_result(
         return result
 
     def predict_action_probabilities(
-        self,
-        tracker: DialogueStateTracker,
-        domain: Domain,
-        precomputations: Optional[MessageContainerForCoreFeaturization] = None,
```
ka-bu, line 229: `tests/core/test_policies.py` will fail without this - it assumes all `predict_action_probabilities` look alike

tests/core/policies/test_rule_policy.py
```diff
@@ -38,29 +42,68 @@
...
+@pytest.fixture()
+def policy_with_config(
+    default_model_storage: ModelStorage,
+    default_execution_context: ExecutionContext,
+    resource: Resource,
+) -> Callable[..., RulePolicy]:
+    def inner(config: Optional[Dict[Text, Any]] = None) -> RulePolicy:
+        config = config or {}
+        policy = RulePolicy.create(
+            {**RulePolicy.get_default_config(), **config},
```
ka-bu, line 89: I've eventually put that into the constructor (or constructor of base class) to avoid having to thing about that in tests later. Seems odd to me that we have the `get_default_config` but the default config is not enforced by the constructor.

tests/core/policies/test_rule_policy.py
```diff
@@ -642,49 +678,49 @@ def test_contradicting_rules_and_stories():
...
+    lookup = trained_rule_policy.lookup
 
-    policy = RulePolicy(
-        priority=priority,
-        lookup=lookup,
-        core_fallback_threshold=core_fallback_threshold,
-        core_fallback_action_name=core_fallback_action_name,
-        enable_fallback_prediction=enable_fallback_prediction,
+    loaded = RulePolicy.load(
+        RulePolicy.get_default_config(),
```
ka-bu, line 702: Btw, It's a bit weird that we don't store the old config. One could trainand save something with e.g. embeddingdim=100 and during load decide to set that with embeddingdim=200. 🤔

tests/core/policies/test_rule_policy.py
```diff
@@ -3064,6 +3035,12 @@ def test_include_action_unlikely_intent():
             "casd", evts=conversation_events, slots=domain.slots
         ),
         domain,
-        RegexInterpreter(),
     )
     test_utils.assert_predicted_action(prediction, domain, utter_2)
+
+
+def test_invalid_fallback_action_name(policy_with_config: Callable[..., RulePolicy]):
```
ka-bu, line 3042: 👍

rasa/core/policies/memoization.py
ka-bu: ah, I see, you just removed that parameter everywhere in the tests :)

ka-bu (approved):
oh, that's what happens when someone answers to a comment that is removed then  :D  

Looks good :) Just have some small questions and not quite sure about the `precomputations` removal (might we need that later again because we'll want to test the lookup table?).

tests/core/policies/test_rule_policy.py
wochinge: That's done[ by the `GraphNode`. ](https://github.com/RasaHQ/rasa/blob/5bb331ff7ba75ad564478cbc71fd1a46bce5ee02/rasa/engine/graph.py#L316). What do you mean by enforcing it?

tests/core/policies/test_rule_policy.py
wochinge: This is stored as part of the graph schema. The individual graph components don't have to store it themselves

rasa/core/policies/ted_policy.py
wochinge: 🤦🏻 This actually only works during training and not in this case 🙈 Will fix it

wochinge: I thought I'm clever with the `load` thing but that actually only works during training. Passing it via `predict_action_probabilities` now

wochinge:
> about the precomputations removal (might we need that later again because we'll want to test the lookup table?).

Yep, removed that as only some policies actually need that (e.g. it was unused by memoization)

tests/core/policies/test_rule_policy.py
ka-bu: I mean I can instantiate components and then get an instance that has none of the default parameters set that I expect - unless we consistently use `self._config.get('whatever', self.get_default_config()['whatever'])`.

wochinge: yep, but that's only in the unit tests, isn't it? Can't happen in real-life cause of the `GraphNode`

ka-bu: True - still would expect this to happen inside the graph component - same with the sanity checks for the configs (i.e. if you pass in a parameter that isn't there, the component should be able to complain)

rasa/core/policies/memoization.py
```diff
@@ -192,14 +193,20 @@ def _recall_states(self, states: List[State]) -> Optional[Text]:
...
+        rule_only_data: Optional[Dict[Text, Any]],
     ) -> Optional[Text]:
         """Finds the action based on the given states.
 
         Args:
             states: List of states.
             tracker: The tracker.
             domain: The Domain.
+            rule_only_data: Slots and loops which are specific to rules and hence
+                should be excluded from during the featurization.
```
ka-bu, line 209:
```suggestion
            rule_only_data: Slots and loops which are specific to rules and hence
                should be ignored by this policy.
```
because we use "featurization" for too many things (e.g. in this case the "featurization" will not create numeric `Features` but just `State`s)

tests/core/policies/test_rule_policy.py
wochinge: > the component should be able to complain

I'd still can do this (several components actually do)

> True - still would expect this to happen inside the graph component

What are the advantages for this? In my opinion it helps to remove duplicate logic in the components (overriding with default, storing config)

rasa/core/policies/policy.py
```diff
@@ -372,15 +372,16 @@ def predict_action_probabilities(
         self,
         tracker: DialogueStateTracker,
         domain: Domain,
-        precomputations: Optional[MessageContainerForCoreFeaturization] = None,
+        rule_only_data: Optional[Dict[Text, Any]] = None,
```
ka-bu, line 375: This is changing the common policy interface here so that it matches what is needed in `Memoization` and `RulePolicy` - but  `TEDPolicy` needs the precomputations and we need to be able to find that parameter via inspection I think. So I think we should add both (or remove this part of the common policy interface - in which case we would have to remove the related tests in `test_policy.py` which assume this common interface).

rasa/core/policies/rule_policy.py
```diff
@@ -767,23 +761,26 @@ def train(
...
 
         Args:
             training_trackers: The list of the trackers.
             domain: The domain.
-            interpreter: Interpreter which can be used by the polices for featurization.
+
+        Returns:
+            The resource which can be used to load the trained policy.
         """
+        self._validate_against_domain(domain)
```
ka-bu, line 775: This is now caught when we start training the component. Before, it was caught before, cf. [here](https://github.com/RasaHQ/rasa/blob/0882404686e085cb6cbc58e369240d8c1c962c62/rasa/core/agent.py#L411). If we want to replicate that behaviour, then I think we should make this check a class method and trigger it during whatever validation runs before any training.

ka-bu (changes requested):

rasa/core/policies/rule_policy.py
wochinge: I moved it here as it "kind of" replicates the previous behavior (just throws later) and we have the loaded `Domain` at hand. Technically we don't have a loaded domain before running the graph (we could make that an exception but I'd probably wouldn't do so unless we have to). 

The other chance would be to make `validate_against_domain` part of a protocol and then do it as part of [this issue](https://github.com/RasaHQ/rasa/issues/9336). I think this would probably be the easiest and cleanest way - what do you think? (also added a note there about this). If it's okay for you I'd keep the check here for now to preserve the basic functionality of this and then potentially revisit this as part of https://github.com/RasaHQ/rasa/issues/9336

rasa/core/policies/policy.py
wochinge: I added this to the interface because _every_ policy requires `rule_only_data` while only policies with end-to-end support require `precomputations`. The graph is flexible enough to provide data only to the components which need it and in my opinion the `kwargs` cover for this case.

> we need to be able to find that parameter via inspection I think

In this case we can just inspect `TEDPolicy` which has this param in its signature, can't we?


> we would have to remove the related tests in test_policy.py

done 👍🏻

rasa/core/policies/rule_policy.py
ka-bu: sounds good - think what we're also missing is all the "required component" stuff and checks for that, no?

rasa/core/policies/policy.py
ka-bu: >The graph is flexible enough to provide data only to the components which need it and in my opinion the kwargs cover for this case.

Should we then adapt TEDPolicy etc to implement the exact same interface and pass/retrieve the precomputations as/from kwargs?

rasa/core/policies/rule_policy.py
wochinge: Yep, will add it to the recipe validation stuff 👍🏻

wochinge: (you actually already added it there 🎉 )

rasa/core/policies/policy.py
wochinge: In my opinion it's fine to keep it in the signature and also have it explicit. Also doesn't violate the interface contract as it has a default

ka-bu (approved):

Issue #9456, opened 2021-08-25
Migrate `LanguageModelFeaturizer` to GraphComponent interface


wochinge:
**Overview of the Solution**:
Adapt the `LanguageModelFeaturizer ` to the new `GraphComponent ` interface.
See the [implementation proposal](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e) for further details.

* adapt `create` ~and `load`~ functions new signatures
* implement `required_packages` and `supported_languages` 
* ~adapt persistence~
* rename `train` to `process_training_data` (this featurizer only adds features but doesn't require training)
* support multiple messages in `process` (**make sure to handle empty messages**)

**Definition of Done**:
- [ ] LanguageModelFeaturizer adapted to the new GraphComponent interface
- [ ] Unit tests are adapted to adapted component
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old component implementation)

aeshky:
The `LanugageModelFeaturizer` doesn't currently have a `load` method. However, reading the [description](https://github.com/RasaHQ/rasa/blob/051b7077825d2379736be14c5e6bdfd87e0abd9e/rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py#L46), the class is designed to always load a pretrained model. In this case, should I replace the existing `create` method with `load`, and then not create a `create` method?

wochinge:
It's also okay to just adapt the `create` one as the default implementations of the `GraphComponent`s `load` just points to `create` (see [here](https://github.com/RasaHQ/rasa/blob/5bb331ff7ba75ad564478cbc71fd1a46bce5ee02/rasa/engine/graph.py#L181))

aeshky:
[`GraphComponent`](https://github.com/RasaHQ/rasa/blob/5bb331ff7ba75ad564478cbc71fd1a46bce5ee02/rasa/engine/graph.py#L129) doesn't have a `name` method like the old [`Component`](https://github.com/RasaHQ/rasa/blob/5bb331ff7ba75ad564478cbc71fd1a46bce5ee02/rasa/nlu/components.py#L452). Do we want to add one?

wochinge:
No, that's not needed

aeshky:
The original ML featurizer doesn't have a persist method. Should I add one, given that we only load pretrained models?

wochinge:
[...]

PR #9586, opened 2021-09-09
3.0 architecture revamp/9456/language model featurizer2

aeshky:
**Proposed changes**:
- Adapted `LanguageModelFeaturizer` according to https://github.com/RasaHQ/rasa/issues/9456


rasa/nlu/featurizers/dense_featurizer/_lm_featurizer.py
```diff
@@ -0,0 +1,802 @@
+# flake8: noqa
+# WARNING: This module will be dropped before Rasa Open Source 3.0 is released.
+#          Please do not make changes to this module, and instead makes changes to
+#          `LanguageModelFeaturizerGraphComponent`
+#          This module is a workaround to defer breaking changes due to the architecture
+#          revamp in 3.0.
```
aeshky, line 6: This is the old `lm_featurizer.py` but renamed. The only change is the highlighted comment.

aeshky:
@ka-bu do you have capacity to review? I tagged you because I wanted to make sure I correctly used your featurizer classes 😁

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -42,58 +45,102 @@
...
-        super(LanguageModelFeaturizer, self).__init__(component_config)
+        """Initializes LanguageModelFeaturizer with the model
+        specified in the config."""
+        super(LanguageModelFeaturizerGraphComponent, self).__init__(
+            execution_context.node_name, config
+        )
+        self._config = {**self.get_default_config(), **config}
+        self._model_storage = model_storage
+        self._resource = resource
+        self._execution_context = execution_context
```
wochinge, line 73: these are actually not used anywhere, are they?
```suggestion
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -42,58 +45,102 @@
...
-        return [Tokenizer]
-
     def __init__(
         self,
-        component_config: Optional[Dict[Text, Any]] = None,
-        skip_model_load: bool = False,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
```
wochinge, line 63:
```suggestion
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -42,58 +45,102 @@
...
+
+    @staticmethod
+    def required_packages() -> List[Text]:
+        """Returns the extra python dependencies required for LanguageModelFeaturizer."""
+        return ["transformers"]
+
+    @staticmethod
+    def required_components(cls) -> List[Type[Component]]:
+        """Returns packages to be installed."""
+        return [Tokenizer]
```
wochinge, line 133: we no longer need this one
```suggestion
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -160,7 +207,7 @@ def _load_model_instance(self, skip_model_load: bool) -> None:
     def cache_key(
         cls, component_meta: Dict[Text, Any], model_metadata: Metadata
     ) -> Optional[Text]:
-        """Cache the component for future use.
+        """Caches the component for future use.
```
wochinge, line 210: 🚀

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -42,58 +45,102 @@
...
+        return cls(config, model_storage, resource, execution_context)
 
-    def _load_model_metadata(self) -> None:
-        """Load the metadata for the specified model and sets these properties.
+    @staticmethod
+    def supported_languages() -> Optional[List[Text]]:
+        """Returns list of supported languages."""
+        # return ["en"]
+        # None means all languages are supported
+        return None  # TODO: check which language are supported
```
aeshky, line 123: I couldn't figure out what languages were supported in the original implementation 🤔 I left it as `None` to mean all languages are supported. If this is correct, I'll remove `# return ["en"]` and the `# TODO`. 
Maybe the support comes from Hugging face and not us?

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -754,13 +796,19 @@ def train(
...
                 batch_start_index += batch_size
 
-    def process(self, message: Message, **kwargs: Any) -> None:
-        """Process an incoming message by computing its tokens and dense features.
+    def process(self, messages: List[Message]) -> None:
+        """Processes a list of incoming message by computing their
+        tokens and dense features."""
+        for message in messages:
+            if message:
+                self._process_message(message)
```
wochinge, line 804: we can shorten `_set_lm_features` using `add_features_to_message` from the parent class, can't we?

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -754,13 +796,19 @@ def train(
...
 
-        Args:
-            message: Incoming message object
+    def _process_message(self, message: Message, **kwargs: Any) -> None:
```
wochinge, line 806: as it's unused
```suggestion
    def _process_message(self, message: Message) -> None:
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -754,13 +796,19 @@ def train(
                     self._set_lm_features(batch_docs[index], ex, attribute)
                 batch_start_index += batch_size
 
-    def process(self, message: Message, **kwargs: Any) -> None:
-        """Process an incoming message by computing its tokens and dense features.
+    def process(self, messages: List[Message]) -> None:
```
wochinge, line 799:
```suggestion
    def process(self, messages: List[Message]) -> List[Message]:
```

wochinge: you have to return the messages at the end of it

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky, line 804: Specs say: **make sure to handle empty messages**
Is the `if` statement sufficient or do we mean something else?

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: same in `process_training_data`

wochinge: is it actually possible to unify `process` and `process_training_data`? 🤔

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -1,21 +1,49 @@
...
+    return Resource("LanguageModelFeaturizerGraphComponent")
+
+
+@pytest.fixture
+def create_language_model_featurizer(
+    default_model_storage: ModelStorage,
+    resource_language_model_featurizer,
+    default_execution_context: ExecutionContext,
+) -> Callable[[Dict[Text, Any]], LanguageModelFeaturizerGraphComponent]:
+    def inner(config: Dict[Text, Any]):
```
wochinge, line 38:
```suggestion
    def inner(config: Dict[Text, Any]) -> Callable[[Dict[Text, Any]], LanguageModelFeaturizerGraphComponent]:
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky, line 799: Are we happy with the name `process`? There is also `process_training_data`. Maybe we want to call this `process_messages`?

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -54,37 +82,77 @@ def create_pretrained_transformers_config(
...
+    ],
+):
+    """ Test create function as a callable
+    This notation will be used for the remaining tests """
+    x = create_language_model_featurizer(
+        LanguageModelFeaturizerGraphComponent.get_default_config()
+    )
+    assert isinstance(x, LanguageModelFeaturizerGraphComponent)
+
+
```
wochinge, line 155: I'd challenge if we need this tests as this should be covered by the other tests implicitly

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -1,21 +1,49 @@
...
+@pytest.fixture
+def create_language_model_featurizer(
+    default_model_storage: ModelStorage,
+    resource_language_model_featurizer,
+    default_execution_context: ExecutionContext,
+) -> Callable[[Dict[Text, Any]], LanguageModelFeaturizerGraphComponent]:
+    def inner(config: Dict[Text, Any]):
+        return LanguageModelFeaturizerGraphComponent.create(
+            config=config,
```
wochinge, line 40:
```suggestion
            config={**LanguageModelFeaturizerGraphComponent.get_default_config(), **config},
```

wochinge (commented):
💯

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: Let's stay consistent with the other ones and keep this `process` for now. Refactoring and improving naming is a great thing to do but we should focus on the migration itself now and don't do too many things at once.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: the loop handles empty lists as well so you're fine 👍🏻

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: Seems there was no check before so you can just omit the `supported_languages` implementation in favor of the one in the parent class

aeshky:
Looks like some of my tests are failing. The ones I'm able to run locally on my machine are passing. I'll look into the CI ones.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky: Oh yeah. I must have copied it from one of the other classes. Oops!

tests/nlu/featurizers/test_lm_featurizer.py
aeshky: Sure. I actually wrote the callable one because I was trying to figure out how to use `Callable` 😂

tests/nlu/featurizers/test_lm_featurizer.py
aeshky: I do this in the [constructor](https://github.com/RasaHQ/rasa/blob/6e12ff8a6fdaff3ac1c0de86a55a8b69d5cedb38/rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py#L70). Do I need it here too?

wochinge: oooh. Doesn't need to be done in the constructor as the graph will do this on the outside (see [here](https://github.com/RasaHQ/rasa/blob/5bb331ff7ba75ad564478cbc71fd1a46bce5ee02/rasa/engine/graph.py#L316))

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky: The execution context is used in the superclass constructor call. Should I keep it and remove the others?

ka-bu: yep, you can remove the unused ones from the `init` (they only need to stay in the `create` because we need that function to implement the interface defined by the GraphComponent)

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -171,17 +211,12 @@ def cache_key(
         weights = component_meta.get("model_weights", {})
 
         return (
-            f"{cls.name}-{component_meta.get('model_name')}-"
+            f"{cls.__name__}-{component_meta.get('model_name')}-"
```
ka-bu, line 214: Think we can remove the `cache_key` method since all the fingerprinting / caching is now done by the graph

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
ka-bu: Yeah we could (I was also wondering if I should do that for the ConveRT featurizer but because of all the other stuff I had changed, I left it as is eventually 😅 ). 

But it doesn't matter at the moment - if we use this either with 1 or 0 messages only.

And I still don't really understand (or keep forgetting) why it is 0 or 1 message. If there were 0 new messages, then we should get the same output from the pipeline... ? - Does it have to do with the special restart events? But even those extract those things from the message 🤔

aeshky: It's currently passing by reference and modifying messages in place. Should modify and return a copy instead? 🤔

aeshky: >is it actually possible to unify process and process_training_data? 🤔

Need to go through the code to understand what's it's doing. 
@ka-bu is your response to tobi about this too?

wochinge: If it's too much effort unifying this then let's skip it.

> you have to return the messages at the end of it
It's currently passing by reference and modifying messages in place, isn't it? Should modify and return a copy instead? 🤔

Modifying in-place is good for now. But we still need to return the things as output to work with the graph (every node needs to have input and output). So no copy is needed

> If there were 0 new messages, then we should get the same output from the pipeline... ? - Does it have to do with the special restart events? But even those extract those things from the message 🤔

We have the 0 message case when you predict an action after having just predicted an action, e.g:
```
- User: "Hi"
- action: utter_greet  <-- in this case we have list with message [1]
- action: action_listen  <-- in this case we predict with an empty list of messages. All NLU components just skip in this case
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -716,13 +716,13 @@ def _get_docs_for_batch(
 
         return batch_docs
 
-    def train(
+    def process_training_data(
         self,
         training_data: TrainingData,
         config: Optional[RasaNLUModelConfig] = None,
         **kwargs: Any,
     ) -> None:
```
aeshky, line 724: @wochinge what should this function return?

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
ka-bu: > We have the 0 message case when you predict an action after having just predicted an action, e.g:

- If the policies do not see that new action (- action: utter_greet) then that means they only see (- User: "Hi") and they will predict the exact same action (- action: utter_greet) again and again. 
- If the policies see that new action (- action: utter_greet), then that means a message containing ACTION_NAME utter_greet has arrived, no?
Or do you mean the prediction *after* (- action: action_listen) - but that doesn't need to be passed through the graph anyway, no?

ka-bu: @aciel - the response was about replacing how process works (i.e. process message by message) by how process_training_data works (i.e. processing a bunch of messages in parallel at a time - at most 64 to be precise ^^). It's not much effort unifying it but it's pointless if process is only called with 1 or 0 messages at a time.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
ka-bu: just the training data that you just modified in that function :)  -- in the graph definition, each of these functions becomes kind of a node with the parameters being the input and the output of the function begin passed on to another node (i.e. we cannot tell the graph that something is just modified in place, we must return a value and it can then pass on this value to whatever other node is definied in the graph definition)

aeshky:
Is it possible to run the code quality checks locally?

ka-bu:
> Is it possible to run the code quality checks locally?

yep, you can run: `make formatter`, `make types`, `make lint` (cf. Makefile in main folder)

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -42,57 +38,83 @@
...
+    def required_packages() -> List[Text]:
+        """Returns the extra python dependencies required."""
+        return ["transformers"]
+
+    def _check_language_is_supported(self) -> None:
+        if "language" in self._config:
+            language = self._config["language"]
+            supported_languages = self.supported_languages()
+            if supported_languages is not None and language not in supported_languages:
+                raise UnsupportedLanguageError(self.__name__, language)
```
ka-bu, line 114: @wochinge the language parameter and checks can be removed here as well, right? -- btw, what do we do when the supported language effectively depends on the config (e.g. because a model trained for spanish is loaded... (I have no idea if they can differ here but seems reasonable that they maybe will differ))?

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -42,57 +38,83 @@
...
+        return {
+            **DenseFeaturizer2.get_default_config(),
+            # name of the language model to load.
+            "model_name": "bert",
+            # Pre-Trained weights to be loaded(string)
+            "model_weights": None,
+            # an optional path to a specific directory to download
+            # and cache the pre-trained model weights.
+            "cache_dir": None,
+            "skip_model_load": False,
```
ka-bu, line 74: we can drop the `skip_model_load` parameter and instead skip the model loading in the tests (cause this parameter must only be used there) the same way we skip e.g. the url validation [here](https://github.com/RasaHQ/rasa/blob/542c8d8fa6c29c8c55aa38f29ab50c13f204cfa9/tests/nlu/featurizers/test_convert_featurizer.py#L103)

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -754,15 +753,20 @@ def train(
...
-        Args:
-            message: Incoming message object
-        """
-        # process of all featurizers operates only on TEXT and ACTION_TEXT attributes,
-        # because all other attributes are labels which are featurized during training
-        # and their features are stored by the model itself.
+    def process(self, messages: List[Message]) -> List[Message]:
+        """Processes messages by computing tokens and dense features."""
+        for message in messages:
+            if message:
```
ka-bu, line 761: You can remove the "if" here (because messages won't evaluate to False and) because the handling of empty lists is achieved by the for loop already:  " for message in []: # do xy " won't ever execute xy :)

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -780,17 +785,9 @@ def _set_lm_features(
...
-            attribute,
-            self.component_config[FEATURIZER_CLASS_ALIAS],
-        )
-        message.add_features(final_sequence_features)
-        final_sentence_features = Features(
-            sentence_features,
-            FEATURE_TYPE_SENTENCE,
-            attribute,
-            self.component_config[FEATURIZER_CLASS_ALIAS],
+        self.add_features_to_message(
```
ka-bu, line 788: 🥳

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -42,57 +38,83 @@
...
-        if not cls.can_handle_language(language):
-            # check failed
-            raise UnsupportedLanguageError(cls.name, language)
-        return cls(component_config)
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> "LanguageModelFeaturizerGraphComponent":
```
wochinge, line 97: Can we please remove all forward references? This requires a `from __future__ import annotations` import at the top to avoid errors
```suggestion
    ) -> LanguageModelFeaturizerGraphComponent:
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: Good point 👍🏻 I think the best would probably be to add a config param to `supported_languages`? (cc @joejuzl ). Then the `supported_languages` can decide based on its config.

wochinge: @aeshky We can dump the entire `_check_language_is_supported` check for now as we do this as part of the [graph schema validation](https://github.com/RasaHQ/rasa/blob/262b5d1dcae599b56e21a81e421f7462699b9a1c/rasa/engine/validation.py#L133). 

Makes sense to add the extended check (having a dynamic `supported_languages` using the config) in the future but for now this is in my opinion not needed as we haven't done it before and we focus on the migration now

rasa/nlu/registry.py
```diff
@@ -30,7 +30,7 @@
 from rasa.nlu.featurizers.sparse_featurizer.count_vectors_featurizer import (
     CountVectorsFeaturizer,
 )
-from rasa.nlu.featurizers.dense_featurizer.lm_featurizer import LanguageModelFeaturizer
+from rasa.nlu.featurizers.dense_featurizer._lm_featurizer import LanguageModelFeaturizer
```
wochinge, line 33: this change shouldn't be needed. Let's do it as it's done [here](https://github.com/RasaHQ/rasa/blob/4eda9168a6231ae104531c231205c78b0bd1fb97/rasa/nlu/featurizers/sparse_featurizer/lexical_syntactic_featurizer.py#L42)

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -54,34 +85,46 @@ def create_pretrained_transformers_config(
...
 
 
-def train_texts(
-    texts: List[Text], model_name: Text, model_weights: Text
+def process_training_text(
+    texts: List[Text],
+    model_name: Text,
+    model_weights: Text,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 93:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -54,34 +85,46 @@ def create_pretrained_transformers_config(
...
 
 
-def process_texts(
-    texts: List[Text], model_name: Text, model_weights: Text
+def process_messages(
+    texts: List[Text],
+    model_name: Text,
+    model_weights: Text,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 114:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -312,31 +355,41 @@ def evaluate_message_shape_values(
...
+    def test_lm_featurizer_shapes_in_process_training_data(
         self,
         model_name: Text,
         model_weights: Text,
         texts: List[Text],
         expected_shape: List[Tuple[int]],
         expected_sequence_vec: List[List[float]],
         expected_cls_vec: List[List[float]],
+        create_language_model_featurizer: Callable[
+            [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 367:
```suggestion
            [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -312,31 +355,41 @@ def evaluate_message_shape_values(
...
+    def test_lm_featurizer_shapes_in_process_messages(
         self,
         model_name: Text,
         model_weights: Text,
         texts: List[Text],
         expected_shape: List[Tuple[int]],
         expected_sequence_vec: List[List[float]],
         expected_cls_vec: List[List[float]],
+        create_language_model_featurizer: Callable[
+            [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 386:
```suggestion
            [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -501,24 +555,38 @@ def check_subtokens(
...
 
-    def test_lm_featurizer_number_of_sub_tokens_train(
+    def test_lm_featurizer_num_sub_tokens_process_training_data(
         self,
         model_name: Text,
         model_weights: Text,
         texts: List[Text],
         expected_number_of_sub_tokens: List[List[float]],
+        create_language_model_featurizer: Callable[
+            [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 565:
```suggestion
            [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -501,24 +555,38 @@ def check_subtokens(
...
 
-    def test_lm_featurizer_number_of_sub_tokens_process(
+    def test_lm_featurizer_num_sub_tokens_process_messages(
         self,
         model_name: Text,
         model_weights: Text,
         texts: List[Text],
         expected_number_of_sub_tokens: List[List[float]],
+        create_language_model_featurizer: Callable[
+            [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 582:
```suggestion
            [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -527,10 +595,15 @@ def test_lm_featurizer_number_of_sub_tokens_process(
     [(20, "bert", False), (1000, "bert", True), (1000, "xlnet", False)],
 )
 def test_sequence_length_overflow_train(
-    input_sequence_length: int, model_name: Text, should_overflow: bool
+    input_sequence_length: int,
+    model_name: Text,
+    should_overflow: bool,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 602:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -557,9 +630,12 @@ def test_long_sequences_extra_padding(
     actual_sequence_lengths: List[int],
     model_name: Text,
     padding_needed: bool,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 634:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -591,8 +667,13 @@ def test_input_padding(
     max_sequence_length_model: int,
     resulting_length: int,
     padding_added: bool,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 671:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -614,18 +695,21 @@ def test_log_longer_sequence(
     model_weights: Text,
     should_overflow: bool,
     caplog: LogCaptureFixture,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 699:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -636,9 +720,16 @@ def test_log_longer_sequence(
     [(256, 512, 256), (700, 700, 700), (700, 512, 512)],
 )
 def test_attention_mask(
-    actual_sequence_length: int, max_input_sequence_length: int, zero_start_index: int
+    actual_sequence_length: int,
+    max_input_sequence_length: int,
+    zero_start_index: int,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 727:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -664,15 +755,17 @@ def test_lm_featurizer_correctly_handle_whitespace_token(
     text: Text,
     tokens: List[Tuple[Text, int]],
     expected_feature_tokens: List[Tuple[Text, int]],
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], create_language_model_featurizer
```
wochinge, line 759:
```suggestion
        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
```

wochinge (commented):
A few more comments but looking great otherwise!

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky: Cool. Should we do this for the other components too? It looks like [Mitie](https://github.com/RasaHQ/rasa/blob/a6b0ac95fc70745eef228c0710003d7f6506c2af/rasa/nlu/featurizers/dense_featurizer/mitie_featurizer.py#L66) is using a forward reference, and [Spacy](https://github.com/RasaHQ/rasa/blob/a6b0ac95fc70745eef228c0710003d7f6506c2af/rasa/nlu/featurizers/dense_featurizer/spacy_featurizer.py#L58) is returning a `GraphComponent`.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky: >Makes sense to add the extended check (having a dynamic supported_languages using the config) in the future

I added a `TODO` for this.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky: Maybe I misunderstood: in the requirements we say to handle empty messages (#9456).
I though this meant that some messages will be `None`. In this case, the loop will not skip them but the `if` statement will. Or is that not the required behaviour? 🤔

aeshky: I mean not the whole list is empty, but some individual messages are empty.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky: I'll try 🐒🩹

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: returning a `GraphComponent` (that's the parent class is fine). Great catch with Mitie 🙈

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: Messages won't be empty 👍🏻 With "empty messages" we meant empty lists. Sorry for the vagueness!

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
wochinge: Can we please rather create an issue @aeshky ? Todos in the code tend to never get done and just make the code messy

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
aeshky: Oh, no problem!

aeshky:
Okay I think I addressed all of your comments @wochinge and @ka-bu 🙂

wochinge:
already on it 🏃🏻

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -39,60 +41,77 @@
...
+            # name of the language model to load.
+            "model_name": "bert",
+            # Pre-Trained weights to be loaded(string)
+            "model_weights": None,
+            # an optional path to a specific directory to download
+            # and cache the pre-trained model weights.
+            "cache_dir": None,
+            "language": None,
+            # TODO: the supported language should depend on the language the model
+            #  was pre-trained on. In the future add this check to config validation.
```
wochinge, line 79:
```suggestion
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -39,60 +41,77 @@
...
+        }
 
     @classmethod
-    def required_components(cls) -> List[Type[Component]]:
-        """Packages needed to be installed."""
-        return [Tokenizer]
+    def validate_config(cls, config: Dict[Text, Any]) -> None:
```
wochinge, line 83: @ka-bu What do you think of doing this with Python Protocols in the future? Too many of the components implement this with `pass`. I think using protocols instead would make this more flexible

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -501,24 +556,38 @@ def check_subtokens(
...
         model_name: Text,
         model_weights: Text,
         texts: List[Text],
         expected_number_of_sub_tokens: List[List[float]],
+        create_language_model_featurizer: Callable[
+            [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
+        ],
     ):
-        messages = train_texts(texts, model_name, model_weights)
+        """ Tests the number of sub tokens when calling the function
```
wochinge, line 569:
```suggestion
        """Tests the number of sub tokens when calling the function
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -501,24 +556,38 @@ def check_subtokens(
...
         model_name: Text,
         model_weights: Text,
         texts: List[Text],
         expected_number_of_sub_tokens: List[List[float]],
+        create_language_model_featurizer: Callable[
+            [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
+        ],
     ):
-        messages = process_texts(texts, model_name, model_weights)
+        """ Tests the number of sub tokens when calling the function
```
wochinge, line 586:
```suggestion
        """Tests the number of sub tokens when calling the function
```

tests/nlu/featurizers/test_lm_featurizer.py
```diff
@@ -527,11 +596,18 @@ def test_lm_featurizer_number_of_sub_tokens_process(
...
+    model_name: Text,
+    should_overflow: bool,
+    create_language_model_featurizer: Callable[
+        [Dict[Text, Any]], LanguageModelFeaturizerGraphComponent
+    ],
+    monkeypatch: MonkeyPatch,
 ):
-    component = LanguageModelFeaturizer(
-        {"model_name": model_name}, skip_model_load=True
+    monkeypatch.setattr(
```
wochinge, line 607: Way better than having production code changed for `pytest` 🚀

wochinge: one thing: should we move the monkeypatching into the fixture to avoid repeating this?

wochinge (approved):
💯

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
ka-bu: I'm not sure. I like the idea of protocols, but the nice thing about the abstract method is that it forces everyone to think about this and implement it. With the protocol we might miss some essential validation without even noticing it (although that's not really a problem with `validate_config` - was just thinking of the other `validate_..` where we replaced tokenizer by tokenizer type or so). 
And if we use the `validate_config` in all components eventually then it doesn't make a difference really whether it's a protocol or there's a version in `GraphComponent` that just passes, no?

wochinge: Good points 👍🏻 Let's just see how this develops and find the right abstractions then 👍🏻

wochinge:
🚀

PR #9588, opened 2021-09-09
migrate `RegexFeaturizer`

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9451
- migrate `RegexFeaturizer`

### DIFF
--- a/rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
+++ b/rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
@@ -241,11 +241,8 @@ class RegexFeaturizerGraphComponent(SparseFeaturizer2, GraphComponent):
         try:
             with model_storage.read_from(resource) as model_dir:
                 patterns_file_name = model_dir / "patterns.pkl"
-                if patterns_file_name.exists():
-                    known_patterns = rasa.shared.utils.io.read_json_file(
-                        patterns_file_name
-                    )
-        except ValueError:
+                known_patterns = rasa.shared.utils.io.read_json_file(patterns_file_name)
+        except (ValueError, FileNotFoundError):
             logger.warning(
                 f"Failed to load `{cls.__class__.__name__}` from model storage. "
                 f"Resource '{resource.name}' doesn't exist."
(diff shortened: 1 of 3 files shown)


rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
```diff
@@ -102,38 +111,41 @@ def _merge_new_patterns(self, new_patterns: List[Dict[Text, Text]]) -> None:
...
         else:
             self.known_patterns = patterns_from_data
 
+        self._persist()
+        return self._resource
+
+    def process_training_data(self, training_data: TrainingData) -> TrainingData:
+        """Processes the training examples (see parent class for full docstring)."""
         for example in training_data.training_examples:
             for attribute in [TEXT, RESPONSE, ACTION_TEXT]:
```
wochinge, line 134: didn't use the thing from the parent class due to the extra attributes

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
```diff
@@ -230,48 +229,51 @@ def _features_for_patterns(
...
-        if patterns_file_name.exists():
-            known_patterns = rasa.shared.utils.io.read_json_file(patterns_file_name)
 
-        return RegexFeaturizer(
-            meta, known_patterns=known_patterns, finetune_mode=should_finetune,
-        )
+        try:
+            with model_storage.read_from(resource) as model_dir:
+                patterns_file_name = model_dir / "patterns.pkl"
+                if patterns_file_name.exists():
```
ka-bu, line 244: Don't we want to warn if that resource isn't there?   ( Sometimes I've added a "this might be because it wasn't trained" - because I had kept the "do not persist if not trained" approach as it was before)

ka-bu (approved):
Just one small question, otherwise looks good :)

rasa/nlu/featurizers/sparse_featurizer/regex_featurizer.py
wochinge: There wasn't one before that's why haven't added one. Will add a debug message

wochinge: the case actually can't happen. So handling it as `FileNotFoundError` now

PR #9593, opened 2021-09-09
Adapt `ResponseSelector` to `GraphComponent` interface

### LINKED ISSUE #9448: Migrate `ResponseSelector` to `GraphComponent`
**Overview of the Solution**:
Adapt the `ResponseSelector ` to the new `GraphComponent` interface.

* adapt persistence

**Definition of Done**:
- [x] ResponseSelector adapted to the new GraphComponent interface
- [ ] Unit tests are adapted to adapted ResponseSelector
- [x] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old ResponseSelector implementation)

ancalita:
**Proposed changes**:
- Closes #9448 

### DIFF
--- a/rasa/nlu/selectors/response_selector.py
+++ b/rasa/nlu/selectors/response_selector.py
@@ -313,17 +313,6 @@ class ResponseSelectorGraphComponent(DIETClassifierGraphComponent):
             sparse_feature_sizes=sparse_feature_sizes,
         )
 
-    @classmethod
-    def create(
-        cls,
-        config: Dict[Text, Any],
-        model_storage: ModelStorage,
-        resource: Resource,
-        execution_context: ExecutionContext,
-    ) -> ResponseSelectorGraphComponent:
-        """Creates a new untrained component (see parent class for full docstring)."""
-        return cls(config, model_storage, resource, execution_context)
-
     @property
     def label_key(self) -> Text:
         """Returns label key."""
@@ -331,10 +320,12 @@ class ResponseSelectorGraphComponent(DIETClassifierGraphComponent):
 
     @property
     def label_sub_key(self) -> Text:
+        """Returns label sub_key."""
         return LABEL_SUB_KEY
 
     @staticmethod
     def model_class(use_text_as_label: bool) -> Type[RasaModel]:
+        """Returns model class."""
         if use_text_as_label:
             return DIET2DIET
         else:
--- a/rasa/nlu/selectors/response_selector.py
+++ b/rasa/nlu/selectors/response_selector.py
@@ -580,6 +580,7 @@ class ResponseSelectorGraphComponent(DIETClassifierGraphComponent):
 
             if (
                 self._execution_context.should_add_diagnostic_data
+                and out
                 and DIAGNOSTIC_DATA in out
             ):
                 message.add_diagnostic_data(
--- a/rasa/nlu/selectors/response_selector.py
+++ b/rasa/nlu/selectors/response_selector.py
@@ -578,7 +578,10 @@ class ResponseSelectorGraphComponent(DIETClassifierGraphComponent):
 
             self._set_message_property(message, prediction_dict, selector_key)
 
-            if out and DIAGNOSTIC_DATA in out:
+            if (
+                self._execution_context.should_add_diagnostic_data
+                and DIAGNOSTIC_DATA in out
+            ):
                 message.add_diagnostic_data(
                     self._execution_context.node_name, out.get(DIAGNOSTIC_DATA)
                 )
--- a/rasa/nlu/selectors/response_selector.py
+++ b/rasa/nlu/selectors/response_selector.py
@@ -7,8 +7,6 @@ import tensorflow as tf
 
 from typing import Any, Dict, Optional, Text, Tuple, Union, List, Type
 
-from rasa.utils import train_utils
-
 from rasa.engine.graph import ExecutionContext
 from rasa.engine.storage.resource import Resource
 from rasa.engine.storage.storage import ModelStorage
@@ -654,29 +652,12 @@ class ResponseSelectorGraphComponent(DIETClassifierGraphComponent):
         **kwargs: Any,
     ) -> ResponseSelectorGraphComponent:
         """Loads the trained model from the provided directory."""
+        model: ResponseSelectorGraphComponent = super().load(
+            config, model_storage, resource, execution_context, **kwargs
+        )
+
         try:
             with model_storage.read_from(resource) as model_path:
-                (
-                    index_label_id_mapping,
-                    entity_tag_specs,
-                    label_data,
-                    data_example,
-                    sparse_feature_sizes,
-                ) = cls._load_from_files(model_path)
-
-                config = train_utils.update_confidence_type(config)
-                config = train_utils.update_similarity_type(config)
-                config = train_utils.update_deprecated_loss_type(config)
-
-                model = cls._load_model(
-                    entity_tag_specs,
-                    label_data,
-                    config,
-                    data_example,
-                    model_path,
-                    finetune_mode=execution_context.is_finetuning,
-                )
-
                 file_name = cls.__name__
                 responses = rasa.shared.utils.io.read_json_file(
                     model_path / f"{file_name}.responses.json"
@@ -684,18 +665,9 @@ class ResponseSelectorGraphComponent(DIETClassifierGraphComponent):
                 all_retrieval_intents = rasa.shared.utils.io.read_json_file(
                     model_path / f"{file_name}.retrieval_intents.json"
                 )
-                return cls(
-                    config,
-                    model_storage,
-                    resource,
-                    execution_context,
-                    index_label_id_mapping,
-                    entity_tag_specs,
-                    model,
-                    all_retrieval_intents,
-                    responses,
-                    sparse_feature_sizes,
-                )
+                model.responses = responses
+                model.all_retrieval_intents = all_retrieval_intents
+                return model
         except ValueError:
             logger.warning(
                 f"Failed to load {cls.__name__} from model storage. Resource "
--- a/tests/nlu/selectors/test_selectors.py
+++ b/tests/nlu/selectors/test_selectors.py
@@ -400,6 +400,8 @@ async def test_process_gives_diagnostic_data(
     for component in loaded_pipeline:
         component.train(training_data)
 
+    default_execution_context.should_add_diagnostic_data = True
+
     response_selector = create_response_selector(
         config_params, default_model_storage, default_execution_context
     )
--- a/tests/nlu/selectors/test_selectors.py
+++ b/tests/nlu/selectors/test_selectors.py
@@ -314,13 +314,13 @@ def _train_persist_load_with_different_settings(
     for component in loaded_pipeline:
         component.train(training_data)
 
-    if should_finetune:
-        default_execution_context.is_finetuning = True
-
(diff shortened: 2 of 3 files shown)


tests/nlu/selectors/test_selectors.py
```diff
@@ -268,105 +294,176 @@ async def test_train_model_checkpointing(
...
-    )
-    await _train_persist_load_with_different_settings(
-        pipeline, component_builder, tmpdir, True
+    config_params = {EPOCHS: 1}
+
+    _train_persist_load_with_different_settings(
+        pipeline, config_params, default_model_storage, default_execution_context, False
     )
 
+    # _train_persist_load_with_different_settings(
```
ancalita, line 360: This part of the test fails with `InvalidParameterException` raised by `DietClassifierGraphComponent` in L314  - unsure how to adapt this further 🙈

tests/nlu/selectors/test_selectors.py
```diff
@@ -752,34 +910,57 @@ async def test_sparse_feature_sizes_decreased_incremental_training(
...
+        Resource("response_selector"),
+        default_execution_context,
+    )
+
+    classified_message2 = loaded_selector.process([message2])[0]
+
+    assert classified_message2.fingerprint() == classified_message.fingerprint()
+
+    # if should_raise_exception:
+    #     with pytest.raises(Exception) as exec_info:
```
ancalita, line 951: also this part of the test fails because no Exception is raised and I wasn't sure how to trigger this in the new architecture either 😬

ancalita:
@wochinge @aeshky this test `test_train_model_checkpointing` was not edited because I didn't know how best to update it, it seems a lot of files are dumped in one location, but that's not the approach in current architecture. Any guidance would be helpful 🙏🏼

rasa/nlu/selectors/response_selector.py
```diff
@@ -289,15 +304,29 @@ def __init__(
...
+    def create(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> ResponseSelectorGraphComponent:
+        """Creates a new untrained component (see parent class for full docstring)."""
+        return cls(config, model_storage, resource, execution_context)
+
```
wochinge, line 326: there is actually no need for this right? we can just use the one from the parent class
```suggestion
```

rasa/nlu/selectors/response_selector.py
```diff
@@ -490,97 +519,101 @@ def _resolve_intent_response_key(
...
-                RESPONSE_SELECTOR_RESPONSES_KEY: label_responses,
-                RESPONSE_SELECTOR_RESPONSE_TEMPLATES_KEY: label_responses,
-                PREDICTED_CONFIDENCE_KEY: top_label[PREDICTED_CONFIDENCE_KEY],
-                INTENT_RESPONSE_KEY: label_intent_response_key,
-                RESPONSE_SELECTOR_UTTER_ACTION_KEY: utter_action_key,
-                RESPONSE_SELECTOR_TEMPLATE_NAME_KEY: utter_action_key,
-            },
-            RESPONSE_SELECTOR_RANKING_KEY: label_ranking,
-        }
+            if out and DIAGNOSTIC_DATA in out:
```
wochinge, line 592: there is a flag for the diagnostic thing on `execution_context` which we should use instead

rasa/nlu/selectors/response_selector.py
```diff
@@ -624,26 +656,66 @@ def _instantiate_model_class(self, model_data: RasaModelData) -> "RasaModel":
...
+    ) -> ResponseSelectorGraphComponent:
         """Loads the trained model from the provided directory."""
-        model: ResponseSelector = super().load(
-            meta, model_dir, model_metadata, cached_component, **kwargs
-        )
-        if not meta.get("file"):
-            return model
-
-        model.responses = meta.get("responses", {})
-        model.all_retrieval_intents = meta.get("all_retrieval_intents", [])
```
wochinge, line 641: why can't we just use the old approach and just load these 2 extra things from the model_storage?

wochinge (commented):
looking into the failing stuff now 👍🏻 

Nice work with this one 💯

tests/nlu/selectors/test_selectors.py
```diff
@@ -268,105 +294,176 @@ async def test_train_model_checkpointing(
...
+    importer = RasaFileImporter(
+        training_data_paths=["data/examples/rasa/demo-rasa.yml"]
+    )
+    training_data = importer.get_nlu_data()
+
+    for component in loaded_pipeline:
+        component.train(training_data)
+
+    if should_finetune:
+        default_execution_context.is_finetuning = True
```
wochinge, line 318: this was done after the training. Now you're doing it _before_ the training which makes it fail.

Finetuning takes an already trained model and improves it. So instead of starting the training from scratch, it uses an already trained model and somewhat "continues" training. Which is why you're getting the exception: You can't finetune without having a trained model first. Does that make sense?

tests/nlu/selectors/test_selectors.py
```diff
@@ -752,34 +910,57 @@ async def test_sparse_feature_sizes_decreased_incremental_training(
...
-                model_to_finetune=loaded,
-            )
-        assert "Sparse feature sizes have decreased" in str(exec_info.value)
-    else:
-        (_, trained, _) = await rasa.nlu.train.train(
-            _config,
-            path=str(tmpdir),
-            data=iter2_path,
-            component_builder=component_builder,
-            model_to_finetune=loaded,
```
wochinge, line 783: this triggers a training with a model to finetune. So you need to set `is_finetuning=True` for the response selector (I think you need to use the `loaded_selector` in the outcommented parts)

rasa/nlu/selectors/response_selector.py
ancalita: Ah I see, then I should replace the current if statement with a `if self._execution_context.should_add_diagnostic_data`?
I also had a question - the default of this flag is `False`, how would it change to `True`?

rasa/nlu/selectors/response_selector.py
ancalita: To double-check, you mean that I don't need to persist these 2 extra things separately and I can load them from `model`? I actually wasn't sure if the persisted model would contain both `responses` and `all_retrieval_intents`?

tests/nlu/selectors/test_selectors.py
ancalita: Got it, perfect TIL 😁 ! Reshuffled this after the training bit and it passed now.

rasa/nlu/selectors/response_selector.py
wochinge: that's all manual in the first place so somebody has to specify that when making predictions (same as it's now probably)

rasa/nlu/selectors/response_selector.py
wochinge: No, I just meant using the old but with the changes `model_storage` changes and the other constructor call?

tests/nlu/selectors/test_selectors.py
```diff
@@ -207,166 +233,242 @@ def test_resolve_intent_response_key_from_label(
...
-        == training_data.responses[
+        == response_selector_training_data.responses[
             util.intent_response_key_to_template_key(resolved_intent_response_key)
         ]
     )
 
 
-async def test_train_model_checkpointing(
-    component_builder: ComponentBuilder, tmpdir: Path
+def test_train_model_checkpointing(
```
ancalita, line 242: Uncertain if this adaptation captures the initial test intention 😬

rasa/nlu/selectors/response_selector.py
ancalita: Ahaa, took a while but I think I understand now your suggestion, it also removes a lot of duplicate code 👍🏼

aeshky:
Looks good 💯 but have yet to review the tests 😬

tests/nlu/selectors/test_selectors.py
usc-m: So I think the intention here is that when checkpointing is enabled, it seems to write out checkpoint files to a temporary directory and then move them to the model storage later into a 'checkpoints' directory (from looking at `DIETClassifierGraphComponent`), and this tests that that happens. 

So I'm not sure it's sufficient to check that just the model is loaded correctly, you'd need to check that the directory exists and has some stuff in it (though the original comments seem to suggest the number of things in there is a bit variable so you might want to look at that for inspiration). I'm not sure if this is necessary if the underlying class tests it, but I'm not sure I have enough context on that.

usc-m:
Looks really good 🎉 Tried to clarify your question about the confusing test, there's a similar test on the `DIETClassifierGraphComponent` which I think pretty much can just be reused (makes me wonder if we need a second copy of it really - suppose it's maybe a bit safer to check the derived class hasn't changed anything there)

tests/nlu/selectors/test_selectors.py
wochinge: I think it'd be great to inspect the `model_storage` whether it contains the checkpoint files 👍🏻

tests/nlu/selectors/test_selectors.py
```diff
@@ -427,29 +544,42 @@ async def test_margin_loss_is_not_normalized(
...
-    (trained_model, _, persisted_path) = await rasa.nlu.train.train(
-        _config,
-        path=str(tmp_path),
-        data=data_path,
-        component_builder=component_builder,
+    importer = RasaFileImporter(training_data_paths=["data/test_selectors"])
+    training_data = importer.get_nlu_data()
+
+    for component in loaded_pipeline:
+        component.train(training_data)
```
wochinge: this bit of code is used multiple times in the tests. How about extracting it into a helper function / fixture?

wochinge (commented):
Looks great - just two more small comments

tests/nlu/selectors/test_selectors.py
ancalita: Yup, I actually just tidied up the tests this morning by extracting 2 helper methods, have just pushed a few minutes ago 👍🏼

tests/nlu/selectors/test_selectors.py
ancalita: I modified the test to resemble the one for `DIETClassifier` - however, the `checkpoints` directory doesn't exist, but the files are there 🤔 . I've commented out the 2 lines that fail in the test.

ancalita: I found why that part of the test fails but not how to fix it though. The part of custom config I pass when creating the response selector (e.g. CHECKPOINT_MODEL default set to `True`) doesn't overwrite the default config for some reason that I yet don't understand 🤔

wochinge: instead of doing files > 4 you just wanna check if the `checkpoint` dir exists, right? And that exists for me when running the test 🤔

wochinge: https://github.com/RasaHQ/rasa/blob/f78a6a17d5146248500630a2397ee099a24e0f5c/rasa/utils/train_utils.py#L201

ancalita: Ahaa, this test config needed settings for `EVAL_NUM_EXAMPLES` & `EVAL_NUM_EPOCHS` to pass. All amended now 🎉

usc-m (approved):

Issue #9449, opened 2021-08-25
Migrate `CountVectorsFeaturizer` to `GraphComponent` interface


wochinge:
**Overview of the Solution**:
Adapt the `CountVectorsFeaturizer ` to the new `GraphComponent` interface.
See the [implementation proposal](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e) for further details.

* adapt `create` and `load` functions to new signatures
* implement `required_packages` and `supported_languages` 
* adapt persistence
* split `train` into `train` (trains the featurizer) and `process_training_data` (adds features to training data for subsequent classifiers etc.)
* support multiple messages in `process` (**make sure to handle empty messages**)

Blockers:
- https://github.com/RasaHQ/rasa/issues/9416

**Definition of Done**:
- [ ] CountVectorsFeaturizer adapted to the new GraphComponent interface
- [ ] Unit tests are adapted to adapted component
- [ ] Other tests (integration tests, etc.) don't break (because they'd (for now) still use the old component implementation)

joejuzl:
Reminder: We need to adapt CountVectorsFeaturizer to always create a vocabulary for end-to-end even if there was no end-to-end training data
See: https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#8863a24281c64d6e8b9de161c9a8d7ee

wochinge:
This is actually no longer needed as we changed it so that the "core" model is re-using the trained "nlu" featurizers

aeshky:
>Blockers:
>Implement Spacy model provider graph component #9416

Is this actually blocked? 🤔

PR #9595, opened 2021-09-09
migrate `CountVectorsFeaturizer`

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9449
- migrate `CountVectorsFeaturizer`


rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -593,53 +614,20 @@ def _get_featurized_attribute(
...
-                message.add_features(final_sentence_features)
-
     def train(
-        self,
-        training_data: TrainingData,
-        cfg: Optional[RasaNLUModelConfig] = None,
-        **kwargs: Any,
-    ) -> None:
-        """Train the featurizer.
+        self, training_data: TrainingData, spacy_nlp: Optional[SpacyModel] = None,
```
ka-bu, line 618: looks like `spacy_nlp` should not be optional (or handled below)

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -798,37 +777,63 @@ def _create_independent_vocab_vectorizers(
...
+                for attribute in vectorizers:
+                    ftr.vectorizers[attribute]._validate_vocabulary()
 
-        share_vocabulary = meta["use_shared_vocab"]
+                return ftr
 
-        if share_vocabulary:
-            vectorizers = cls._create_shared_vocab_vectorizers(
-                meta, vocabulary=vocabulary
+        except (ValueError, FileNotFoundError, FileIOException):
```
ka-bu, line 817: (This reminds me that I definitely shouldn't use ValueErrors for the config validation - otherwise the ValueError here could be from the constructor complaining about a faulty config 😬 .)

ka-bu (changes requested):

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
wochinge: I made it optional as Spacy might not always be present. I'm handling the `None` case below

ka-bu: oh 👀  - completely overlooked that 

another maybe stupid followup question: this is modifying `self.OOV_words` during `train` - could the config that we pass during `load` contain `OOV_words` and hence overwrite this result?

wochinge: very good point! We'll need to persist it separately

wochinge: thanks for the catch ❤️ Updated it

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
```diff
@@ -50,91 +53,95 @@ class CountVectorsFeaturizer(SparseFeaturizer):
...
-        self.lowercase = self.component_config["lowercase"]
+        self.lowercase = self._config["lowercase"]
 
         # use the lemma of the words or not
-        self.use_lemma = self.component_config["use_lemma"]
+        self.use_lemma = self._config["use_lemma"]
 
     def _load_vocabulary_params(self) -> None:
-        self.OOV_token = self.component_config["OOV_token"]
+        self.OOV_token = self._config["OOV_token"]
```
ka-bu, line 142: I think it would be cleaner if we pass "oov_tokens" and "oov_words" to the `init` and inside `init` use `_load_vocabulary_params` only if both are None (and otherwise set the oov words and tokens to the passed parameters). -- And if not, it would be cool to at least leave some notes somewhere that what happens here does not destroy the loaded oov words/tokens.

ka-bu (changes requested):

rasa/nlu/featurizers/sparse_featurizer/count_vectors_featurizer.py
wochinge: done

ka-bu (approved):
👍

PR #9607, opened 2021-09-10
Migrate EntitySynonymMapper to graph architecture

usc-m:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9504
- Migrate EntitySynonymMapper to graph architecture


rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -1,41 +1,65 @@
...
 import rasa.shared.utils.io
+from rasa.nlu.extractors._entity_synonyms import EntitySynonymMapper
+from rasa.engine.storage.resource import Resource
+from rasa.engine.storage.storage import ModelStorage
 
+# This is a workaround around until we have all components migrated to `GraphComponent`.
+EntitySynonymMapper = EntitySynonymMapper
 
-class EntitySynonymMapper(EntityExtractor):
-    @classmethod
```
usc-m, line 18: I wasn't sure if this needed kept?

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -1,41 +1,65 @@
...
-    ) -> None:
+    @classmethod
+    def create(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+        synonyms: Optional[Dict[Text, Any]] = None,
+    ) -> EntitySynonymMapperComponent:
```
usc-m, line 57: I seem to get a lot of errors about this not being defined yet locally - seems to really want me to use forward references, but not sure what I can do to avoid them like the other components?

tests/nlu/extractors/test_entity_synonyms.py
```diff
@@ -11,7 +11,7 @@ def test_entity_synonyms():
         {"entity": "test", "value": "china", "start": 0, "end": 6},
     ]
     ent_synonyms = {"chines": "chinese", "NYC": "New York City"}
-    EntitySynonymMapper(synonyms=ent_synonyms).replace_synonyms(entities)
+    EntitySynonymMapperComponent(synonyms=ent_synonyms).replace_synonyms(entities)
```
usc-m, line 14: The test after this one seems to use the old component architecture and I'm not quite sure what to do with it - is that an important part of the test or should I strip it out and essentially replicate this?

rasa/nlu/extractors/entity_synonyms.py
usc-m: This refers to the actual class method below, not the class name itself

rasa/nlu/extractors/_entity_synonyms.py
```diff
@@ -0,0 +1,120 @@
+import os
```
wochinge, line 1: can we please add the comment on top here which we have for the other migrated components?

rasa/nlu/extractors/entity_synonyms.py
wochinge: yeah, this one is no longer needed 👍🏻

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -1,41 +1,65 @@
...
+    """Maps entities to their synonyms if they appear in the training data."""
+
+    SYNONYM_FILENAME = "synonyms.json"
 
     def __init__(
         self,
-        component_config: Optional[Dict[Text, Any]] = None,
+        config: Optional[Dict[Text, Any]] = None,
+        model_storage: ModelStorage = None,
+        resource: Resource = None,
```
wochinge, line 30: no need for defaults - these will always be passed
```suggestion
        config: Dict[Text, Any],
        model_storage: ModelStorage,
        resource: Resource,
```

rasa/nlu/extractors/entity_synonyms.py
wochinge: you need a `from __future__ import annotations` at the top of your file

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -44,59 +68,62 @@ def train(
...
                 self.add_entities_if_synonyms(entity_val, str(entity.get("value")))
 
-    def process(self, message: Message, **kwargs: Any) -> None:
+        if self._resource:
+            self._persist()
 
-        updated_entities = message.get(ENTITIES, [])[:]
-        self.replace_synonyms(updated_entities)
-        message.set(ENTITIES, updated_entities, add_to_output=True)
+    def process(self, messages: List[Message]) -> None:
```
wochinge, line 74: we need to return the processed message even though they are modified in place to make it explicit that the graph node passes an output into the next component (otherwise the data flow would be implicit again)
```suggestion
    def process(self, messages: List[Message]) -> List[Message]:
```

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -44,59 +68,62 @@ def train(
...
-                entity_synonyms_file, self.synonyms, separators=(",", ": ")
-            )
-            return {"file": file_name}
-        else:
-            return {"file": None}
+        if self._synonyms:
+            with self._model_storage.write_to(self._resource) as storage:
+                entity_synonyms_file = os.path.join(
+                    storage, EntitySynonymMapperComponent.SYNONYM_FILENAME
+                )
```
wochinge, line 87: `storage` is a `Path` object which allows you to do fancy things like this (see [here](https://docs.python.org/3/library/pathlib.html))
```suggestion
                entity_synonyms_file = storage / EntitySynonymMapperComponent.SYNONYM_FILENAME
                
```

tests/nlu/extractors/test_entity_synonyms.py
wochinge: Oh dear - the test is indeed very confusing. It currently gets a pipeline, then extracts the one component which we need and then trains it 😮‍💨 I think you can drop all of this and just create a new `EntitySynonymMapperComponent` and call train on it

wochinge (commented):
Looking very good 🚀

wochinge:
at @usc-m Added a "fix https://github.com/RasaHQ/rasa/issues/9504" to the pull request description so that issue and PR are linked

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -1,102 +1,142 @@
...
+            self._add_entities_if_synonyms(key, value)
 
         for example in training_data.entity_examples:
             for entity in example.get(ENTITIES, []):
                 entity_val = example.get(TEXT)[entity["start"] : entity["end"]]
-                self.add_entities_if_synonyms(entity_val, str(entity.get("value")))
+                self._add_entities_if_synonyms(entity_val, str(entity.get("value")))
+
+        if self._resource:
+            self._persist()
```
wochinge, line 73: as the `resource` is not `Optional`
```suggestion
        self._persist()
```

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -1,102 +1,142 @@
...
-
-        entity_synonyms_file = os.path.join(model_dir, file_name)
-        if os.path.isfile(entity_synonyms_file):
-            synonyms = rasa.shared.utils.io.read_json_file(entity_synonyms_file)
-        else:
-            synonyms = None
-            rasa.shared.utils.io.raise_warning(
-                f"Failed to load synonyms file from '{entity_synonyms_file}'.",
-                docs=DOCS_URL_TRAINING_DATA + "#synonyms",
+        with model_storage.read_from(resource) as storage:
```
wochinge, line 115: this might throw a `ValueError` in case their is no persisted data for this `Resource`. Can we please handle that (e.g. see how the other components do it)?

wochinge (commented):
just a few more comments. Ready to go otherwise 🚀

rasa/nlu/extractors/entity_synonyms.py
```diff
@@ -1,102 +1,152 @@
...
 
         for example in training_data.entity_examples:
             for entity in example.get(ENTITIES, []):
                 entity_val = example.get(TEXT)[entity["start"] : entity["end"]]
-                self.add_entities_if_synonyms(entity_val, str(entity.get("value")))
+                self._add_entities_if_synonyms(entity_val, str(entity.get("value")))
+
+        self._persist()
+
+    def process(self, messages: List[Message]) -> List[Message]:
```
wochinge, line 77: one more thing: `process_training_data` was removed from the parent class 😬 Could you please add this here then (and probably add a short test)? 

Should be these lines:

```python
def process_training_data(self, training_data: TrainingData) -> TrainingData:
    """Processes the training examples in the given training data in-place.

    Args:
      training_data: the training data

    Returns:
      same training data after processing
    """
    self.process(training_data.training_examples)
    return training_data
```

wochinge (approved):
One comment due to some changes on `main`. Nice and clean change 💯

PR #9609, opened 2021-09-12
Resolve hardcoded entities key

Mazyod:
**Proposed changes**:
- Replace hardcoded "entities" key with its corresponding const `ENTITIES`.


ancalita (commented):
Good spot and thank you for your contribution.
Could you please update your branch with latest changes from `main` and then we can proceed with merging this.

ancalita (approved):
🎉

PR #9629, opened 2021-09-14
Fix for failed_test_stories.yml not printing the correct message

### LINKED ISSUE #9279: failed_test_stories.yml is not printing the correct message
**Rasa version**: 2.8.0

**Rasa SDK version** (if used & relevant): N/A

**Rasa X version** (if used & relevant): N/A

**Python version**: 3.8

**Operating system** (windows, osx, ...): Linux

**Issue**: failed_test_stories.yml includes a test story that is actually correct

The bot is attached.

When running `rasa test` , you will see that this test story is marked as a failed test story, but it does not say why.

**file: test_stories.yml**
```yaml
- story: play ping pong
  steps:
  - user: |
      hello there!
    intent: greet
  - action: utter_greet
  - user: |
      play ping pong
    intent: play_game
  - action: utter_lets_play

```

**file: failed_test_stories.yml**
```yaml
version: "2.0"
stories:
- story: play ping pong (./tests/test_stories.yml)
  steps:
  - intent: greet
  - action: utter_greet
  - intent: play_game
  - action: utter_lets_play
```

In previous versions of Rasa, it would print the reason why it failed:

**file: failed_test_stories.yml**
```yaml
version: "2.0"
stories:
- story: play ping pong 
  steps:
  - intent: greet
  - action: utter_greet
  - intent: play_game  # predicted: play_game: play [ping pong](game)
  - action: utter_lets_play
```

It all works when updating the test story to include the entity extraction, like this:

**file: test_stories.yml  (with corrected version)**
```yaml
- story: play ping pong
  steps:
  - user: |
      hello there!
    intent: greet
  - action: utter_greet
  - user: |
      play [ping pong](game)
    intent: play_game
  - action: utter_lets_play
```

[the_model.zip](https://github.com/RasaHQ/rasa/files/6941250/the_model.zip)

alwx:
**Proposed changes**:
- fix #9279


rasa/core/test.py
```diff
@@ -373,7 +374,10 @@ def inline_comment(self) -> Optional[Text]:
         """A comment attached to this event. Used during dumping."""
         from rasa.shared.core.events import format_message
 
-        if self.predicted_intent != self.intent["name"]:
+        if (
+            self.predicted_intent != self.intent["name"]
+            or self.predicted_entities != self.target_entities
```
ancalita, line 379: I'm wondering if this shouldn't go into an `elif` and using `inline_comment_for_entity` in that elif branch? This solution causes a different test to fail with correct intent but wrong entity and it's perhaps clearer if the comment points to the entity as the source of failure rather than intent.
Probably off-topic, I also found it strange that this static method is not used anywhere 🤔

alwx: `inline_comment_for_entity` is actually used by `YAMLStoryWriter` (in `process_user_utterance` method).
Regarding the first suggestion: not sure how it can help with that test but still thinking of it :(

alwx: Found another way of doing it. Please, re-check!

rasa/shared/core/training_data/story_writer/yaml_story_writer.py
```diff
@@ -263,6 +244,25 @@ def process_user_utterance(
                     entities.append(entity["entity"])
             result[KEY_ENTITIES] = entities
 
+        if hasattr(user_utterance, "inline_comment"):
+            comment = user_utterance.inline_comment(force_add=not entities)
```
ancalita, line 248: Looks great 💯 I think my only suggestion would be to rename this argument into something that would clarify instantly what its purpose is.

alwx: Tried to find one, I think it suits better than the previous one.

ancalita (approved):
🎉

PR #9636, opened 2021-09-14
Use asyncio.run instead of loop.run_until_complete

### LINKED ISSUE #9236: Py36 drop: consider using asyncio.run instead of loop.run_until_complete()
Python 3.7 introduced [`asyncio.run`](https://docs.python.org/3/library/asyncio-task.html#asyncio.run):
> This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once.

**Definition of done**
- [ ] Evaluate if and where this should be replaced in the code
- [ ] Implement needed changes

usc-m:
**Proposed changes**:
- Fix #9236 

### DIFF
--- a/rasa/api.py
+++ b/rasa/api.py
@@ -104,9 +104,8 @@ def train(
         An instance of `TrainingResult`.
     """
     from rasa.model_training import train_async
-    import rasa.utils.common
 
-    return rasa.utils.common.run_in_loop(
+    return asyncio.run(
         train_async(
             domain=domain,
             config=config,
@@ -121,7 +120,9 @@ def train(
             model_to_finetune=model_to_finetune,
             finetuning_epoch_fraction=finetuning_epoch_fraction,
         ),
-        loop,
+        # TODO: asyncio.run doesn't let you set a custom loop,
+        # do we want to keep this parameter?
+        # loop,
     )
 
 
--- a/rasa/core/agent.py
+++ b/rasa/core/agent.py
@@ -289,11 +289,11 @@ def create_agent(model: Text, endpoints: Text = None) -> "Agent":
     from rasa.core.tracker_store import TrackerStore
     from rasa.core.utils import AvailableEndpoints
     from rasa.core.brokers.broker import EventBroker
-    import rasa.utils.common
+    import asyncio
 
     _endpoints = AvailableEndpoints.read_endpoints(endpoints)
 
-    _broker = rasa.utils.common.run_in_loop(EventBroker.create(_endpoints.event_broker))
+    _broker = asyncio.run(EventBroker.create(_endpoints.event_broker))
     _tracker_store = TrackerStore.create(_endpoints.tracker_store, event_broker=_broker)
     _lock_store = LockStore.create(_endpoints.lock_store)
 
--- a/rasa/utils/common.py
+++ b/rasa/utils/common.py
@@ -268,6 +268,7 @@ class RepeatedLogFilter(logging.Filter):
         return False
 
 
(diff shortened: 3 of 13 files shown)


tests/core/test_nlg.py
```diff
@@ -56,6 +56,9 @@ async def generate(request):
     return app
 
 
+# TODO: Not sure if these ones can be replaced
+# pytest-sanic adds a fixture which is what is being used here
+# Not certain this won't change behaviour
```
usc-m, line 61: Not sure there's as much value changing the tests, seems like using the specific test fixtures may give us less headaches in the long run than using asyncio.run. What do you think?

rasa/utils/common.py
```diff
@@ -268,6 +268,7 @@ def filter(self, record: logging.LogRecord) -> bool:
         return False
 
 
+# TODO: This is now unused
```
usc-m, line 271: I can probably just drop this function but I'm not sure if this constitutes part of our public API and therefore needs to go through the whole deprecation process first

rasa/api.py
```diff
@@ -121,7 +120,9 @@ def train(
             model_to_finetune=model_to_finetune,
             finetuning_epoch_fraction=finetuning_epoch_fraction,
         ),
-        loop,
+        # TODO: asyncio.run doesn't let you set a custom loop,
```
usc-m, line 123: This is the only call I could find to the old `run_in_loop` function that actually let you specify a specific `loop` - and none of the invocations of this use it. However, I'm pretty sure this function (`train`) is a public API so removing a parameter might require some delicate handling. What do you think

rasa/core/agent.py
```diff
@@ -289,11 +289,11 @@ def create_agent(model: Text, endpoints: Text = None) -> "Agent":
     from rasa.core.tracker_store import TrackerStore
     from rasa.core.utils import AvailableEndpoints
     from rasa.core.brokers.broker import EventBroker
-    import rasa.utils.common
+    import asyncio
 
     _endpoints = AvailableEndpoints.read_endpoints(endpoints)
 
-    _broker = rasa.utils.common.run_in_loop(EventBroker.create(_endpoints.event_broker))
+    _broker = asyncio.run(EventBroker.create(_endpoints.event_broker))
```
wochinge, line 296: Will this actually work? Pretty sure we already have a running event loop at this point so `asyncio.get_event_loop` seems to be more fitting, right?

rasa/utils/common.py
wochinge: No, you can safely drop it as it's just a `util` function outside of the `rasa.shared` package (everything within `rasa.shared` needs extra caution as it might be used by downstream dependencies such as Rasa X)

tests/core/test_nlg.py
wochinge: yes, agree 👍🏻

rasa/api.py
wochinge: Good point - I think dropping is fine as long as point that out in the changelog 👍🏻

wochinge (commented):
💯

rasa/core/agent.py
usc-m: Tests seem to pass so it does work but it might be more efficient to use the helper to run it in the existing loop if that exists, I might do an experiment to see if doing that actually takes the `get_event_loop` path

wochinge: that would be great cause I'm pretty sure there should already be an existing loop

PR #9638, opened 2021-09-14
finetuning validator

### LINKED ISSUE #9335: implement `FineTuningValidator`
**Overview of the Solution**:
We need to implement the `FineTuningValidator`. This is a component at the beginning of the graph which receives all training data and raises an exception in case `is_finetuning=True` and finetuning is not possible (e.g. because new labels were added). In case we are doing a regular training (and not finetuning) this persists the necessary information extracted from the training data to be able to validate when initialized via `load` whether we can finetune.

The check whether the model is finetunable should be loosely based upon [this current code piece](https://github.com/RasaHQ/rasa/blob/f78a6a17d5146248500630a2397ee099a24e0f5c/rasa/model.py#L548).

**Input to the component**
- `TrainingDataImporter`

**Output:** 
* `None`
* raises an exception if finetuning can't be done

**Blockers**
* https://github.com/RasaHQ/rasa/pull/9184

**Definition of Done**:

ka-bu:
**Proposed changes**:
- closes #9335 


ka-bu:
@joejuzl  We need the actual NLU training data in this validator and we'll also need the loaded data in for the `ConfigValidator` (NLU data as well as story data, where the story data is needed to replicate the validations previously done in `_ensemble.py`). Which means we load the NLU data 3 times (2 x in validators, 1 x in provider) and story data 2 times.  Is it ok if I adapt this component such that it takes `TrainingData` as input? ... Or am I missing something? 🤔

joejuzl:
@ka-bu 
The `CombinedDataImporter` (which should always be in the importer chain) caches the methods such as `get_nlu_data` so it should be fine.

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+
+    This is a component at the beginning of the graph which receives all training data
+    and raises an exception in case is_finetuning=True and finetuning is not possible
+    (e.g. because new labels were added). In case we are doing a regular training
+    (and not finetuning) this persists the necessary information extracted from the
+    training data to be able to validate when initialized via load whether we can
+    finetune.
+
+    Finetuning is possible if, compared to the initial training phase, it holds that
+    1. the configuration (excpet for "epoch" keys) does not change
```
joejuzl, line 35:
```suggestion
    1. the configuration (except for "epoch" keys) does not change
```

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+        self._fingerprints: Dict[Text, Text] = fingerprints or {}
+
+    def validate_nlu_only(
+        self, importer: TrainingDataImporter, config: Dict[Text, Any]
+    ) -> None:
+        """Validates whether we can finetune the NLU part when finetuning is enabled.
+
+        Args:
+            importer: a training data importer
+            config: the configuration (including the recipe)
```
joejuzl, line 79: What does "including the recipe" mean?

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+            fingerprints: a dictionary of fingerprints generated by a
+                `FineTuningValidator`
+        """
+        self._execution_context = execution_context
+        self._model_storage = model_storage
+        self._resource = resource
+        self._fingerprints: Dict[Text, Text] = fingerprints or {}
+
+    def validate_nlu_only(
+        self, importer: TrainingDataImporter, config: Dict[Text, Any]
```
joejuzl, line 73: We can get the config from the `importer` right?

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+    def _validate(
+        self,
+        importer: TrainingDataImporter,
+        config: Dict[Text, Any],
+        nlu: bool = True,
+        core: bool = True,
+    ) -> None:
+        """Validate whether the finetuning setting conflicts with other settings.
+
+        Note that this validation always takes into account the configuration of
```
joejuzl, line 123: I feel this is implementation details and does not need to be in a docstring.

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+        if finetuning and not self._fingerprints:
+            raise InvalidConfigException(
+                f"Finetuning is enabled but the {self.__class__.__name__} "
+                f"does not remember seeing a training run. Ensure that you have "
+                f"trained your model at least once (with finetuning disabled) "
+                f"and ensure that the  {self.__class__.__name__} is part of the "
+                f"training graph. "
+            )
+
+        # check the rasa version
```
joejuzl, line 150: If a comment is needed, a nicely named method would be preferable in my opinion.

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+        self._fingerprints[FINGERPRINT_VERSION] = rasa.__version__
+
+        # check the configuration
+
+        self._compare_raise_update(
+            fingerprint_key=FINGERPRINT_CONFIG_WITHOUT_EPOCHS_KEY,
+            new_fingerprint=self._get_fingerprint_of_config_without_epochs(config),
+            error_message=(
+                "Cannot finetune after making certain changes to the config. "
+                "The only changes that are allowed are"
```
joejuzl, line 172: Unfinished sentence.

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+        """Returns a fingerprint of a version of the given domain without responses.
+
+        Args:
+            domain: a domain
+        Returns:
+            fingerprint
+        """
+        # Do a copy of the domain to not change the actual domain (shallow is enough)
+        domain = copy.copy(domain)
+        # don't include the response texts in the fingerprint.
```
joejuzl, line 261: I don't think we need these comments - this is obvious given the name of the method and the line of code.

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+        # Note: Since this involves importing nlu data, it is probably the most
+        # costly step and hence happens last.
+
+        # NOTE: If there's a consistency check between domain and training data
+        # that ensures domain and NLU training data are consistent, then we can
+        # drop this check.
+
+        if nlu:
+            nlu_data = importer.get_nlu_data()
+            domain = importer.get_domain()
```
joejuzl, line 201: Is this used?

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -0,0 +1,341 @@
...
+                    "for finetuning contains at least one example for every label "
+                    "(i.e. intent, action name, ...) that was included in the NLU "
+                    "data used for training the model which we attempt to finetune "
+                    "now. Moreover, you must not add labels that were not included "
+                    "during training before. "
+                ),
+            )
+
+        # persist
+        if not finetuning:
```
joejuzl, line 217: What if we want to finetune a second time?

tests/graph_components/validators/test_finetuning_validator.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+@pytest.mark.parametrize("nlu, core", [(True, False), (False, True), (True, True)])
+def test_validate_after_changing_response_text_in_domain(
+    get_validation_method: Callable[..., ValidationMethodType],
+    project: Text,
+    nlu: bool,
+    core: bool,
+):
+    # NOTE: test_model/test_validate_changed_response_text
```
joejuzl, line 114: What does this mean?

tests/graph_components/validators/test_finetuning_validator.py
```diff
@@ -0,0 +1,434 @@
...
+    config2 = copy.deepcopy(config1)
+    replacements = 0
+    for key in config2:
+        for sub_config in config2[key]:
+            if "epochs" in sub_config:
+                sub_config["epochs"] = sub_config["epochs"] + 5
+                replacements += 1
+    assert (
+        replacements > 0
+    ), 'Please update the test such that the config used here contains "epoch" keys.'
```
joejuzl, line 237: Do we need to test our tests? 😂

tests/graph_components/validators/test_finetuning_validator.py
```diff
@@ -0,0 +1,434 @@
...
+
+@pytest.mark.parametrize(
+    "nlu, core,key",
+    [
+        (nlu, core, key)
+        for nlu, core in [(True, False), (False, True), (True, True)]
+        for key in [INTENT, ACTION_NAME]
+    ],
+)
+def test_validate_after_removing_or_adding_intent_or_action_name(
```
joejuzl, line 331: Do we have a test where we add a message that is not a new label?

joejuzl (changes requested):

rasa/graph_components/validators/finetuning_validator.py
ka-bu: oh, I missed that, yep :)

rasa/graph_components/validators/finetuning_validator.py
ka-bu: I found this a bit confusing which is why I added it explicitly. If I want to finetune the NLU model and than re-train the core part with that fine tuned model, then I must adapt the core part only after finetuning (even though that should have no effect on the nlu part), no?

rasa/graph_components/validators/finetuning_validator.py
ka-bu: 👍  they're left-overs from the old version

rasa/graph_components/validators/finetuning_validator.py
ka-bu: nope, copy&paste mistake - domain is not needed for nlu

rasa/graph_components/validators/finetuning_validator.py
ka-bu: then we can use the same fingerprints again - they haven't changed the first time we finetuned

tests/graph_components/validators/test_finetuning_validator.py
ka-bu: This is just for us / for the review to see where a check came from

tests/graph_components/validators/test_finetuning_validator.py
ka-bu: kind of :) - I don't see any harm in checking this because the config that is modified here is defined in another function and so you have to know that this might need an update (otherwise it will just silently pass without checking anything)

tests/graph_components/validators/test_finetuning_validator.py
ka-bu: don't think that was included before - will add one

rasa/graph_components/validators/finetuning_validator.py
ka-bu: Actually the version could change. Although, if the updated version is compatible with the last version but incompatible with the oldest version, that would be strange. I can still update it...

rasa/graph_components/validators/finetuning_validator.py
joejuzl: ohhhh I see what you mean now. Then yes I agree, this makes sense to keep.

joejuzl (approved):
Nice! 🚀

Issue #9364, opened 2021-08-16
3.0 slot mappings - rasa-oss implementation pII: add new default action + adapt `forms.py` + add logic in `processor.py`


ancalita:
**Description of Problem**:
Background: This is provided in [Taking Apart Forms](https://www.notion.so/rasa/Taking-Apart-Forms-be897aa7fd54401bbe398702302094f6) and [Slot Mappings](https://www.notion.so/rasa/Slot-Mappings-and-Forms-60e94e1c26724cfbba0a97c394f68b99).

Summary: The approach of filling slots in 2.0 includes an implicit mechanism (via [auto-fill](https://rasa.com/docs/rasa/domain#slot-auto-fill)) and an explicit mechanism (via forms). This makes it challenging to keep information up-to-date throughout the course of a conversation, as well as requires slot mapping duplication across forms if multiple forms should fill the same slot.

**Overview of the Solution**:

- The full extent of proposed changes can be found in the [implementation proposal](https://www.notion.so/rasa/Global-Slot-Mappings-in-3-0-7dc9b0b89f7a4b89b77fb8866655be61).
- This issue is only concerned with the sections requiring adaptations to [`action.py`](https://www.notion.so/rasa/Global-Slot-Mappings-in-3-0-7dc9b0b89f7a4b89b77fb8866655be61#d3a71a50c12e40ff9860a523e2f1c89f), [`forms.py`](https://www.notion.so/rasa/Global-Slot-Mappings-in-3-0-7dc9b0b89f7a4b89b77fb8866655be61#731a419a4c0e48b18d15abe274d217b3) and [`processor.py`](https://www.notion.so/rasa/Global-Slot-Mappings-in-3-0-7dc9b0b89f7a4b89b77fb8866655be61#5e9d41a7d076407bac9f90462e5b0d56).

**Definition of Done**:
- [x] add new default action for extracting slots
- [x] adapt `forms.py`
- [x] add logic in `processor.py`
- [x] add unit tests

PR #9644, opened 2021-09-15
Adapt `slot.py` and `domain.py` to accept global slot mappings

### LINKED ISSUE #9363: 3.0 slot mappings - rasa-oss implementation pI: adapt `slots.py` and `domain.py` in `rasa.shared.core`
**Description of Problem**:
Background: This is provided in [Taking Apart Forms](https://www.notion.so/rasa/Taking-Apart-Forms-be897aa7fd54401bbe398702302094f6) and [Slot Mappings](https://www.notion.so/rasa/Slot-Mappings-and-Forms-60e94e1c26724cfbba0a97c394f68b99).

Summary: The approach of filling slots in 2.0 includes an implicit mechanism (via [auto-fill](https://rasa.com/docs/rasa/domain#slot-auto-fill)) and an explicit mechanism (via forms). This makes it challenging to keep information up-to-date throughout the course of a conversation, as well as requires slot mapping duplication across forms if multiple forms should fill the same slot.

**Overview of the Solution**:
- The full extent of proposed changes can be found in the [implementation proposal](https://www.notion.so/rasa/Global-Slot-Mappings-in-3-0-7dc9b0b89f7a4b89b77fb8866655be61).
- This issue is only concerned with the sections requiring adaptations to [`slots.py`](https://www.notion.so/rasa/Global-Slot-Mappings-in-3-0-7dc9b0b89f7a4b89b77fb8866655be61#97f13a61479b45d5bea9e98010f70750) and [`domain.py`](https://www.notion.so/rasa/Global-Slot-Mappings-in-3-0-7dc9b0b89f7a4b89b77fb8866655be61#49d8cbaefd82419db4af78752e518298).

**Definition of Done**:
- [x] make relevant changes in `slots.py`
- [x] make relevant changes in `domain.py`
- [x] create an issue for adding utility module to migrate from 2.0 format to 3.0 format
- [x] add & adapt unit tests

ancalita:
**Proposed changes**:
- Closes #9363 
- Also had to adapt `forms.py` in order to make the unit tests pass, kept the changes to a minimum, so that the next PR will look into considerable more change to this module as planned in issue #9364 

### DIFF
--- a/data/test_domains/conditional_response_variations.yml
+++ b/data/test_domains/conditional_response_variations.yml
@@ -12,8 +12,12 @@ slots:
     values:
       - primary
       - secondary
+    mappings:
+      - type: from_text
   can_withdraw:
     type: bool
+    mappings:
+      - type: custom
 
 responses:
   utter_withdraw:
--- a/data/test_domains/conditional_response_variations.yml
+++ b/data/test_domains/conditional_response_variations.yml
@@ -13,7 +13,7 @@ slots:
       - primary
       - secondary
     mappings:
-      - type: from_text
+      - type: custom
   can_withdraw:
     type: bool
     mappings:
--- a/data/test_domains/duplicate_entities.yml
+++ b/data/test_domains/duplicate_entities.yml
@@ -9,15 +9,19 @@ slots:
   cuisine:
     type: text
     mappings:
-    - type: from_text
+      - type: from_entity
+        entity: cuisine
   location:
     type: text
     mappings:
-    - type: from_text
+      - type: from_entity
+        entity: location
 
 entities:
  - name
  - name
+ - cuisine
+ - location
 
 responses:
   utter_greet:
--- a/data/test_domains/duplicate_entities.yml
+++ b/data/test_domains/duplicate_entities.yml
@@ -8,8 +8,12 @@ intents:
 slots:
   cuisine:
     type: text
+    mappings:
+    - type: from_text
   location:
     type: text
+    mappings:
+    - type: from_text
 
 entities:
  - name
--- a/examples/rules/domain.yml
+++ b/examples/rules/domain.yml
@@ -20,8 +20,13 @@ slots:
   some_slot:
     type: text
     influence_conversation: false
+    mappings:
+    - type: from_entity
+      entity: some_slot
   detailed_faq:
     type: bool
+    mappings:
+    - type: custom
 
 actions:
 - utter_explain_some_slot
--- a/examples/rules/domain.yml
+++ b/examples/rules/domain.yml
@@ -21,8 +21,21 @@ slots:
     type: text
     influence_conversation: false
     mappings:
-    - type: from_entity
-      entity: some_slot
+      # Slot mappings can be defined in the domain.
+      # You can also implement custom slot mappings in your validate function for the
+      # `Form` by returning the desired slot events.
+      # The slot mappings follow the same syntax as currently in the SDK implementation.
+      - type: from_entity
+        entity: some_slot
+      # Example of a slot mapping which extracts the slot value from the message text if
+      # the intent is `greet`.
+      # - type: from_text
+      #   intent: greet
+      # Example of a slot mapping which sets the slot to `my value` in case the message
+      # has an intent `greet`
+      # - type: from_intent
+      #   intent: greet
+      #   value: "my value"
   detailed_faq:
     type: bool
     mappings:
@@ -50,22 +63,7 @@ actions:
 forms:
   loop_q_form:
     required_slots:
-      some_slot:
-      # Slot mappings can be defined in the domain.
-      # You can also implement custom slot mappings in your validate function for the
-      # `Form` by returning the desired slot events.
-      # The slot mappings follow the same syntax as currently in the SDK implementation.
-      - type: from_entity
-        entity: some_slot
-      # Example of a slot mapping which extracts the slot value from the message text if
-      # the intent is `greet`.
-      # - type: from_text
-      #   intent: greet
-      # Example of a slot mapping which sets the slot to `my value` in case the message
-      # has an intent `greet`
-      # - type: from_intent
-      #   intent: greet
-      #   value: "my value"
+      - some_slot
 
 session_config:
   session_expiration_time: 60  # value in minutes
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -1239,9 +1239,9 @@ class Domain:
                     for entity in entities
                     if any(
                         [
-                            entity.get("entity")
-                            == mapping.get("entity", "no_entity_found")
+                            entity.get("entity") == mapping.get("entity")
                             for mapping in slot.mappings
+                            if mapping.get("type") == "from_entity"
                         ]
                     )
                 ]
@@ -1755,15 +1755,6 @@ class Domain:
 
         return any(key in content for key in ALL_DOMAIN_KEYS)
 
-    def slot_mapping_for_form(self, form_name: Text) -> List[Text]:
-        """Deprecated method in 3.0."""
-        rasa.shared.utils.io.raise_deprecation_warning(
-            "Slot mappings defined in the form section of the domain have been "
-            "deprecated in 3.0.0.",
-            "4.0.0",
-        )
-        return self.required_slots_for_form(form_name)
-
     def required_slots_for_form(self, form_name: Text) -> List[Text]:
         """Retrieve the list of required slot names for a form defined in the domain.
 
@@ -1976,7 +1967,7 @@ def _validate_slot_mappings(domain_slots: Dict[Text, Any]) -> None:
     for slot_name, properties in domain_slots.items():
         if not isinstance(properties, dict):
             raise InvalidDomain(
-                f"The properties of slot {slot_name} were specified as "
+                f"The properties of slot '{slot_name}' were specified as "
                 f"'{type(properties)}'. They need to be specified as dictionary."
             )
         mappings = properties.get("mappings")
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -329,9 +329,6 @@ class Domain:
         slot_dict = copy.deepcopy(slot_dict)
         # Don't sort the slots, see https://github.com/RasaHQ/rasa-x/issues/3900
         for slot_name in slot_dict:
-            # remove auto_fill if it exists in the domain
-            slot_dict[slot_name].pop("auto_fill", None)
-
             slot_type = slot_dict[slot_name].pop("type", None)
             slot_class = Slot.resolve_by_type(slot_type)
 
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -1221,7 +1221,7 @@ class Domain:
         return states
 
     def slots_for_entities(self, entities: List[Dict[Text, Any]]) -> List[SlotSet]:
-        """Creates slot events for entities if auto-filling is enabled.
+        """Creates slot events for entities if from_entity mapping matches.
 
         Args:
             entities: The list of entities.
@@ -1229,14 +1229,20 @@ class Domain:
         Returns:
             A list of `SlotSet` events.
         """
+        # TODO: revisit/remove method after this is fixed:
+        #  https://github.com/RasaHQ/rasa/issues/9364
         if self.store_entities_as_slots:
             slot_events = []
             for slot in self.slots:
-                # if slot.auto_fill:
                 matching_entities = [
                     entity.get("value")
                     for entity in entities
-                    if entity.get("entity") == slot.name
+                    if any(
+                        [
+                            entity.get("entity") == mapping.get("entity", None)
+                            for mapping in slot.mappings
+                        ]
+                    )
                 ]
                 if matching_entities:
                     if slot.type_name == "list":
@@ -1975,14 +1981,17 @@ def _validate_slot_mappings(domain_slots: Dict[Text, Any]) -> None:
         mappings = properties.get("mappings")
 
         if mappings is None:
-            raise InvalidDomain(f"The slot '{slot_name}' has no mappings defined.")
+            raise InvalidDomain(
+                f"The slot '{slot_name}' has no mappings defined. "
+                f"Please see {DOCS_URL_SLOTS} for more information."
+            )
 
         if not isinstance(mappings, list):
             raise InvalidDomain(
                 f"The slot mappings for slot '{slot_name}' in "
                 f"the domain have type '{type(mappings)}'. "
                 f"It is required to provide a list of slot "
-                f"mappings. Please see the documentation "
+                f"mappings. Please see {DOCS_URL_SLOTS} "
                 f"for more information."
             )
 
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -1224,19 +1224,17 @@ class Domain:
         if self.store_entities_as_slots:
             slot_events = []
             for slot in self.slots:
-                if slot.auto_fill:
-                    matching_entities = [
-                        entity.get("value")
-                        for entity in entities
-                        if entity.get("entity") == slot.name
-                    ]
-                    if matching_entities:
-                        if slot.type_name == "list":
-                            slot_events.append(SlotSet(slot.name, matching_entities))
-                        else:
-                            slot_events.append(
-                                SlotSet(slot.name, matching_entities[-1])
-                            )
+                # if slot.auto_fill:
+                matching_entities = [
+                    entity.get("value")
+                    for entity in entities
+                    if entity.get("entity") == slot.name
+                ]
+                if matching_entities:
+                    if slot.type_name == "list":
+                        slot_events.append(SlotSet(slot.name, matching_entities))
+                    else:
+                        slot_events.append(SlotSet(slot.name, matching_entities[-1]))
             return slot_events
         else:
             return []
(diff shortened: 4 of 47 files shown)


rasa/shared/core/domain.py
```diff
@@ -1424,8 +1431,8 @@ def cleaned_domain(self) -> Dict[Text, Any]:
         for slot in domain_data[KEY_SLOTS].values():
             if slot["initial_value"] is None:
                 del slot["initial_value"]
-            if slot["auto_fill"]:
-                del slot["auto_fill"]
+            # if slot["auto_fill"]:
+            del slot["auto_fill"]
```
ancalita, line 1435: Now that `auto-fill` can only be `False` - I wasn't sure if to delete it or not? To make my life easy with test adaptation, I went for deletion, but not sure what would make more sense for Rasa-X.

rasa/shared/core/domain.py
```diff
@@ -1219,19 +1228,17 @@ def slots_for_entities(self, entities: List[Dict[Text, Any]]) -> List[SlotSet]:
...
-                        if entity.get("entity") == slot.name
-                    ]
-                    if matching_entities:
-                        if slot.type_name == "list":
-                            slot_events.append(SlotSet(slot.name, matching_entities))
-                        else:
-                            slot_events.append(
-                                SlotSet(slot.name, matching_entities[-1])
-                            )
+                # if slot.auto_fill:
```
ancalita, line 1231: This was a temp fix to prevent quite a few tests from failing since this method is used in several other parts of the codebase. I will revisit the usage of this method once the 2nd issue of the implementation is completed too to see if this can be adapted/removed.

tests/test_validator.py
```diff
@@ -346,27 +346,27 @@ def test_verify_form_slots_invalid_domain(tmp_path: Path):
...
+             last_name:
                 type: text
+                mappings:
+                - type: from_text
         """
     )
     importer = RasaFileImporter(domain_path=domain)
-    validator = Validator.from_importer(importer)
     with pytest.warns(UserWarning) as w:
-        validity = validator.verify_form_slots()
```
ancalita, line 363: I've realised that by adding a new `InvalidDomain` raise in `_validate_forms()` in `domain.py` for these kind of mismatches that this validator method `verify_form_slots()`  is now redundant. Should I remove it as part of this PR?

examples/rules/domain.yml
```diff
@@ -20,8 +20,26 @@ slots:
   some_slot:
     type: text
     influence_conversation: false
+    mappings:
+      # Slot mappings can be defined in the domain.
+      # You can also implement custom slot mappings in your validate function for the
```
usc-m, line 25: Is this comment now correct if we're changing how custom slot mappings are implemented?

rasa/shared/core/domain.py
usc-m: Probably makes sense to delete for now but longer run we probably will want to just drop the key entirely if it has no information content at all.

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1798,116 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
         if not isinstance(mapping, dict):
             raise InvalidDomain(
                 f"Please make sure that the slot mappings for slot '{slot_name}' in "
-                f"your form '{form_name}' are valid dictionaries. Please see "
-                f"{DOCS_URL_FORMS} for more information."
+                f"your domain are valid dictionaries. Please see "
+                f"the documentation for more information."
```
usc-m, line 1802: Maybe we want to refer to the docs URL for domains here? I assume that's where this info will live down the line (though could be subject to change at this stage I suppose)

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1798,116 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
         mapping_type = mapping.get("type")
         required_keys = validations.get(mapping_type)
 
         if required_keys is None:
             raise InvalidDomain(
-                f"Your form '{form_name}' uses an invalid slot mapping of type "
+                f"Your domain uses an invalid slot mapping of type "
                 f"'{mapping_type}' for slot '{slot_name}'. Please see "
-                f"{DOCS_URL_FORMS} for more information."
+                f"the documentation for more information."
```
usc-m, line 1820: Same here with the docs URL

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1798,116 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
 
         for required_key in required_keys:
             if mapping.get(required_key) is None:
                 raise InvalidDomain(
                     f"You need to specify a value for the key "
                     f"'{required_key}' in the slot mapping of type '{mapping_type}' "
-                    f"for slot '{slot_name}' in the form '{form_name}'. Please see "
-                    f"{DOCS_URL_FORMS} for more information."
+                    f"for slot '{slot_name}'. Please see the "
+                    f"documentation for more information."
```
usc-m, line 1829: And here

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1798,116 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
+                if condition.get("active_loop") == active_loop_name:
+                    active_loop_match = True
+                    break
+
+        if active_loop_match:
+            form_ignored_intents = domain.forms[active_loop_name].get(
+                IGNORED_INTENTS, []
+            )
+            if not isinstance(form_ignored_intents, list):
+                form_ignored_intents = [form_ignored_intents]
```
usc-m, line 1851: Could just set `ignored_intents` directly here - if not a list, wrap it in a list and set it, else just set it? Saves a call to `extend`, which reads to me like the list is going to grow beyond just the values collected in this pass.

rasa/shared/core/slots.py
```diff
@@ -27,16 +27,18 @@ class Slot:
     def __init__(
         self,
         name: Text,
+        mappings: List[Dict[Text, Any]],
         initial_value: Any = None,
         value_reset_delay: Optional[int] = None,
-        auto_fill: bool = True,
+        auto_fill: bool = False,
```
usc-m, line 33: Is it worth maybe leaving this as True? That would ensure existing code which uses the deprecated functionality gets the deprecation warning (which also changes the value to False, so we shouldn't have any issues correctness wise). I'm not sure how best to handle this - it looks like existing code should just work and leaving it set to False would mean new code written doesn't trigger the deprecation warning, but if the semantics are changing would it be worthwhile warning users? Is that maybe something we'd do after they've migrated to the new version on first run (excluding the migration guide)?

tests/test_validator.py
usc-m: Does `rasa data validate` produce a warning under the same conditions as `InvalidDomain` is raised? If so yeah probably fine to get rid of it or replace it with a call to something else

usc-m (commented):
Looks good 💯 some minor things before I approve, mostly optional - just want to see what you think so we don't miss it later

examples/rules/domain.yml
ancalita: This will probably have to be adapted after the rasa-sdk changes, because it refers to `FormValidationAction` and that interface will be changed to work outside Forms too. I will make a note about this comment in this file 👍🏼

rasa/shared/core/domain.py
ancalita: I actually think the information about slot mappings will be moved to the Domain/Slots page so will update accordingly. I've already made a note to revisit all these Docs links once the docs draft is ready.

tests/test_validator.py
ancalita: Yes, running `rasa data validate` will continue to result in error, because the domain will be empty (so the check that will fail will be [this](https://github.com/RasaHQ/rasa/blob/954107f4c6e84e506d387b65dc978452421e2acd/rasa/cli/data.py#L201)) and the warning message displayed is the one I updated this test with.

The reason why the domain is initialised as empty if `InvalidDomain` gets raised is found [here](https://github.com/RasaHQ/rasa/blob/954107f4c6e84e506d387b65dc978452421e2acd/rasa/shared/importers/rasa.py#L67).

rasa/shared/core/domain.py
joejuzl: This means we are always auto-filling slots from entities though right? Shouldn't this just be entirely replaced by the mappings now?

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1799,117 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
+    def _get_ignored_intents(
+        mapping: Dict[Text, Any], domain: Domain, active_loop_name: Text,
+    ) -> List[Text]:
+        mapping_conditions = mapping.get("conditions")
+        active_loop_match = False
+        ignored_intents = []
+
+        if mapping_conditions:
+            for condition in mapping_conditions:
+                if condition.get("active_loop") == active_loop_name:
```
joejuzl, line 1843: Can there be multiple conditions with the same `active_loop`?

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1799,117 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
+                    active_loop_match = True
+                    break
+
+        if active_loop_match:
+            form_ignored_intents = domain.forms[active_loop_name].get(
```
joejuzl, line 1848: Do we still need form level ignored intents?

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1799,117 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
+    ) -> bool:
+        """Checks whether user intent matches slot mapping intent specifications."""
+        mapping_intents = mapping.get("intent", [])
+        mapping_not_intents = mapping.get("not_intent", [])
+
+        active_loop_name = tracker.active_loop_name
+        if active_loop_name:
+            mapping_not_intents = set(
+                mapping_not_intents
+                + SlotMapping._get_ignored_intents(mapping, domain, active_loop_name)
```
joejuzl, line 1870: Doesn't this just match on `active_loop`? Do we know that the other conditions have matched?

rasa/shared/core/domain.py
```diff
@@ -1789,38 +1799,117 @@ def validate(mapping: Dict[Text, Any], form_name: Text, slot_name: Text) -> None
...
+            )
+
+        intent = tracker.latest_message.intent.get("name")
+
+        intent_not_blocked = not mapping_intents and intent not in mapping_not_intents
```
joejuzl, line 1875: why `not mapping_intents`?

rasa/shared/core/slots.py
```diff
@@ -45,9 +47,19 @@ def __init__(
...
         self._value_reset_delay = value_reset_delay
+        if auto_fill:
+            rasa.shared.utils.io.raise_deprecation_warning(
+                "Slot auto-fill has been deactivated in 3.0 and replaced by a"
+                " new explicit mechanism to specify slot mappings. "
+                "Please refer to the docs to learn more.",
+                "4.0.0",
+                docs=None,
+            )
+            auto_fill = False
```
joejuzl, line 62: Is this really a deprecation warning if we are just setting it to `False` whenever it is `True`?

rasa/shared/utils/schemas/domain.yml
```diff
@@ -51,7 +51,9 @@ mapping:
         type: "map"
         mapping:
           required_slots:
-            type: "map"
+            type: "seq"
```
joejuzl, line 54: Do we also need to add a section for the new global mappings?

joejuzl (changes requested):

rasa/shared/core/domain.py
ancalita: >  This means we are always auto-filling slots from entities though right?

Only when the entity and slot names match.

>  Shouldn't this just be entirely replaced by the mappings now?

Oh right, you mean modify the implementation of this method to find the matching mapping (i.e, via `SlotMapping.entity_is_desired`?

data/test_domains/conditional_response_variations.yml
```diff
@@ -12,8 +12,12 @@ slots:
     values:
       - primary
       - secondary
+    mappings:
+      - type: from_text
   can_withdraw:
     type: bool
+    mappings:
+      - type: custom
```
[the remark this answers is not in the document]
ancalita: Yep, this means that the slot should be set according to a custom mapping defined in a custom action (most likely the adapted `FormValidationAction`). Potentially it can also be set by a custom policy, although I'm not too sure yet how the default action would access that 🤔

data/test_domains/duplicate_entities.yml
```diff
@@ -8,8 +8,12 @@ intents:
 slots:
   cuisine:
     type: text
+    mappings:
+    - type: from_text
   location:
     type: text
+    mappings:
+    - type: from_text
```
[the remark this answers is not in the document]
ancalita: Yup, in this case, that's what will happen. However, because this particular domain didn't have too many details, I decided to use this mapping type since it seemed the easier route (this mapping type doesn't require other keys than `type`). Happy to change this to `custom`?

rasa/shared/core/domain.py
ancalita: In theory, yes, there could be if `requested_slot` key is used too, but with different values. 
But here the intention was only to retrieve the `ignored_intents` of a form if these exist and the form is activated.

rasa/shared/core/domain.py
ancalita: I actually only realised last week that this wasn't discussed during the implementation proposal phase - if to maintain this key in forms, move it under each slot name or under mapping conditions? I proceeded with keeping in forms, but happy to start this conversation today with main stakeholders to find out the best course of action.

rasa/shared/core/domain.py
ancalita: I don't think the other conditions are required for retrieving `ignored_intents` since this flag would apply regardless to all slots if the form is activated.

rasa/shared/core/domain.py
ancalita: I actually lifted this line from `FormAction` - I think the intention here was two-fold:
1) if mappings has no intent listed -> then this mapping is applicable to any intent, as long as latest intent is not listed as `not_intent` (therefore intent is `unblocked`).
2) if mappings has an intent listed -> then this mapping is applicable only to a specified intent, therefore is `blocked`.

rasa/shared/core/slots.py
ancalita: At the moment because devs expect `auto-fill` to be by default `True` (therefore, not required to explicitly add `auto-fill: true` in the domain), the deprecation warning won't be triggered; it would only be raised only if someone would set it to `True` in the domain. @usc-m [suggested ](https://github.com/RasaHQ/rasa/pull/9644#discussion_r711117715)to keep the default auto-fill to be True as it used to be and raise the deprecation warning. WDYT?

rasa/shared/utils/schemas/domain.yml
ancalita: Ah yes, definitely! I'll add a section here right away.

rasa/shared/core/slots.py
ancalita: I've actioned this so that auto-fill remains `True` by default and raises deprecation warning, however, it makes some tests to fail because it increase the overall number of warnings. Ok to keep it like this and edit the tests with explanatory comments where necessary?

ancalita: Example CI run here: https://github.com/RasaHQ/rasa/pull/9644/checks?check_run_id=3652137803

rasa/shared/core/domain.py
ancalita: Dropped the key entirely after we decided to remove the functionality entirely but warn about the removal.

data/test_domains/duplicate_entities.yml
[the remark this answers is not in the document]
joejuzl: I reckon if it hasn't come from a form we should have it as `custom` right?

rasa/shared/utils/schemas/domain.yml
```diff
@@ -43,6 +43,40 @@ mapping:
...
+            required: False
+          initial_value:
+            type: "any"
+            required: False
+          mappings:
+           type: "seq"
+           required: True
+           allowempty: False
+           sequence:
+              - type: "map"
```
joejuzl, line 78: Sorry to be a pain - but we could also restrict this right? e.g. `active_loop` etc.

joejuzl (changes requested):
looking good - just a few more comments.

rasa/shared/core/domain.py
```diff
@@ -1739,27 +1755,29 @@ def is_domain_file(filename: Text) -> bool:
 
         return any(key in content for key in ALL_DOMAIN_KEYS)
 
-    def slot_mapping_for_form(self, form_name: Text) -> Dict[Text, Any]:
-        """Retrieve the slot mappings for a form which are defined in the domain.
+    def slot_mapping_for_form(self, form_name: Text) -> List[Text]:
```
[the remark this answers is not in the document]
ancalita: Not anymore, I've replaced all its calls with  the new `required_slots_for_form` method. All good to remove then?

joejuzl (approved):
Nice work! 💯

Issue #8459, opened 2021-04-13
Utilize specific retrieval intents in story testing


mvielkind:
**Description of Problem**:
When testing stories containing retrieval intents only the base retrieval intent (i.e. `chitchat` or `faq`) are utilized. If you define retrieval sub-intents (i.e. `chitchat/ask_name` or `chitchat/ask_weather`) in your story the testing will ignore the added detail of the specific retrieval sub-intents, which can raise confusing and unexpected results from tests.

**Overview of the Solution**:
In our testing output from the `ResponseSelector` should be utilized to determine the sub-intent of the retrieval intent instead of just using the top-level intent. The example below demonstrates the current behavior along with the expected behavior.

**Examples** (if relevant):
This is the test story:

```
- story: response selector
  steps:
  - user: |
      what is your name?
    intent: chitchat/ask_name
  - action: utter_chitchat/ask_name
```

After `rasa test` the story above will output the following:
```
  steps:
  - intent: chitchat
  - action: utter_chitchat/ask_name  # predicted: utter_chitchat
  - action: action_listen  # predicted: action_default_fallback
```

The test story contains differing levels of specificity for the retrieval intents than is provided by the test results where the test seems to ignore the specific retrieval intent mentioned in the story.

An expected output for the same test would be the following where the test considers the specific retrieval intents in the test story.
```
steps:
  - intent: chitchat/ask_name [specificity in intent step for response selector]
  - action: utter_chitchat/ask_name  # predicted: utter_chitchat/goodbye [specificity in prediction]
  - action: action_listen  # predicted: action_default_fallback
 ```

**Definition of Done**:
- [ ] Story tests utilize the `ResponseSelector` output

**Comment by alwx** [comment]
After some investigation I found out that there is no easy way of doing this.
The problem is that the prediction is correct — `MessageProcessor.predict_next_action` only returns the name of the predicted intent which is indeed `utter_chitchat` in this particular case:
```
  - action: utter_chitchat/ask_name  # predicted: utter_chitchat
 ```
 
Since there are no retrieval intents in the domain, the only way to understand that `utter_chitchat/ask_name` should be used is by executing the `utter_chitchat` action — in this case, `MessageProcessor._run_action` will be called, and when executing the action, we also execute `ResponseSelector.process` method which makes the correct prediction and executes the right action.
However, we don't run actions when doing testing — which is exactly what we're trying to solve here: https://github.com/RasaHQ/rasa/issues/8691

[...]

PR #9657, opened 2021-09-17
Utilizing retrieval intents in test stories

jupyterjazz:
PR for issue #8459 
**Proposed changes**:
- Failed test stories will display full retrieval intents.
- Retrieval intents will be extracted during action prediction in test stories so that we won't have unnecessary mismatches anymore.

### DIFF
--- a/data/test_retrieval_intents_in_test_stories/config.yml
+++ b/data/test_retrieval_intents_in_test_stories/config.yml
@@ -1,0 +1,14 @@
+language: en
+pipeline:
+- name: WhitespaceTokenizer
+- name: CountVectorsFeaturizer
+- name: DIETClassifier
+  epochs: 5
+- name: ResponseSelector
+  epochs: 5
+policies:
+- name: TEDPolicy
+  random_seed: 42
+  epochs: 5
+- name: MemoizationPolicy
+  max_history: 5
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -4,25 +4,21 @@ import logging
 from typing import List, Text, Optional, Dict, Any, TYPE_CHECKING
 
 import aiohttp
-
 import rasa.core
-from rasa.core.policies.policy import PolicyPrediction
-
-from rasa.shared.core import events
 from rasa.core.constants import DEFAULT_REQUEST_TIMEOUT
-
+from rasa.core.policies.policy import PolicyPrediction
 from rasa.nlu.constants import (
     RESPONSE_SELECTOR_DEFAULT_INTENT,
     RESPONSE_SELECTOR_PROPERTY_NAME,
     RESPONSE_SELECTOR_PREDICTION_KEY,
     RESPONSE_SELECTOR_UTTER_ACTION_KEY,
 )
-
 from rasa.shared.constants import (
     DOCS_BASE_URL,
     DEFAULT_NLU_FALLBACK_INTENT_NAME,
     UTTER_PREFIX,
 )
+from rasa.shared.core import events
 from rasa.shared.core.constants import (
     USER_INTENT_OUT_OF_SCOPE,
     ACTION_LISTEN_NAME,
@@ -37,8 +33,7 @@ from rasa.shared.core.constants import (
     ACTION_BACK_NAME,
     REQUESTED_SLOT,
 )
-from rasa.shared.exceptions import RasaException
-from rasa.shared.nlu.constants import INTENT_NAME_KEY, INTENT_RANKING_KEY
+from rasa.shared.core.domain import Domain
 from rasa.shared.core.events import (
     UserUtteranceReverted,
     UserUttered,
@@ -50,10 +45,10 @@ from rasa.shared.core.events import (
     Restarted,
     SessionStarted,
 )
+from rasa.shared.exceptions import RasaException
+from rasa.shared.nlu.constants import INTENT_NAME_KEY, INTENT_RANKING_KEY
 from rasa.shared.utils.schemas.events import EVENTS_SCHEMA
 from rasa.utils.endpoints import EndpointConfig, ClientResponseError
-from rasa.shared.core.domain import Domain
-
 
 if TYPE_CHECKING:
     from rasa.shared.core.trackers import DialogueStateTracker
@@ -359,6 +354,34 @@ class ActionRetrieveResponse(ActionBotResponse):
         """Resolve the name of the intent from the action name."""
         return action_name.split(UTTER_PREFIX)[1]
 
+    def get_full_retrieval_name(
+        self, tracker: "DialogueStateTracker"
+    ) -> Optional[Text]:
+        """Extracts retrieval intent from response selector and
+        returns complete action utterance name."""
+        if RESPONSE_SELECTOR_PROPERTY_NAME not in tracker.latest_message.parse_data:
+            return None
+
+        response_selector_properties = tracker.latest_message.parse_data[
+            RESPONSE_SELECTOR_PROPERTY_NAME
+        ]
+
+        if (
+            self.intent_name_from_action(self.action_name)
+            in response_selector_properties
+        ):
+            query_key = self.intent_name_from_action(self.action_name)
+        elif RESPONSE_SELECTOR_DEFAULT_INTENT in response_selector_properties:
+            query_key = RESPONSE_SELECTOR_DEFAULT_INTENT
+        else:
+            return None
+
+        selected = response_selector_properties[query_key]
+        full_retrieval_utter_action = selected[RESPONSE_SELECTOR_PREDICTION_KEY][
+            RESPONSE_SELECTOR_UTTER_ACTION_KEY
+        ]
+        return full_retrieval_utter_action
+
     async def run(
         self,
         output_channel: "OutputChannel",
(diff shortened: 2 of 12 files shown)


rasa/core/test.py
```diff
@@ -550,8 +552,20 @@ def _collect_user_uttered_predictions(
...
+                    RESPONSE_SELECTOR_PROPERTY_NAME
+                ]
+            }
+            if RESPONSE_SELECTOR_PROPERTY_NAME in predicted
+            else None
+        )
         end_to_end_user_utterance = EndToEndUserUtterance(
-            event.text, event.intent, event.entities
+            text=event.text,
+            intent=event.intent,
```
wochinge, line 566: We seem to mix labeled data here with data from the prediction. Shouldn't the retrieval intent be `intent_gold`?

tests/core/test_test.py
```diff
@@ -123,6 +124,41 @@ async def test_testing_warns_if_action_unknown(
     assert "which is not part of the training data / domain" in output
 
 
+async def test_testing_with_utilizing_retrieval_intents(
+    intent_retrieval_agent: Agent,
+    intent_retrieval_bot_test_stories: Path,
+    intent_retrieval_bot_test_results: Path,
+):
+    if not os.path.exists(intent_retrieval_bot_test_results):
```
wochinge, line 132: small hack as it's a `Path`
```suggestion
    if not intent_retrieval_bot_test_results.exists():
```

tests/core/test_test.py
```diff
@@ -123,6 +124,41 @@ async def test_testing_warns_if_action_unknown(
...
+        stories=intent_retrieval_bot_test_stories,
+        agent=intent_retrieval_agent,
+        e2e=True,
+        out_directory=intent_retrieval_bot_test_results,
+        disable_plotting=True,
+        warnings=False,
+    )
+    failed_stories_path = os.path.join(
+        intent_retrieval_bot_test_results, "failed_test_stories.yml"
+    )
```
wochinge, line 145:
```suggestion
    failed_stories_path = intent_retrieval_bot_test_results / "failed_test_stories.yml"
```

tests/core/test_test.py
```diff
@@ -123,6 +124,41 @@ async def test_testing_warns_if_action_unknown(
...
+    with open(failed_stories_path, "r") as stream:
+        failed_stories = yaml.safe_load(stream)
```
wochinge, line 147: can we please use our existing utils for this?

tests/core/test_test.py
```diff
@@ -123,6 +124,41 @@ async def test_testing_warns_if_action_unknown(
...
+    # check that the intent is shown correctly in the failed test stories file
+    intent, full_intent = "faq", "faq/is-this-legit"
+    target_intents = [full_intent, full_intent, intent]
+    for index, story in enumerate(failed_stories["stories"]):
+        assert story["steps"][0]["intent"] == target_intents[index]
+    # check that retrieval intent for actions is retrieved correctly
+    # and only when it's needed
```
wochinge, line 154: do you also want to check that the comment is the right one?

wochinge (commented):
Great testing and clean PR :100: Mostly questions at this point from me

data/test_retrieval_intents_in_test_stories/config.yml
```diff
@@ -0,0 +1,16 @@
+language: en
```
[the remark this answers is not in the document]
jupyterjazz: Do you mean just config or the whole bot itself? I think re-using existing bots is a good idea. I should be able to modify data/test_response_selector_bot directory and use it for testing 🤖

data/test_retrieval_intents_in_test_stories/config.yml
```diff
@@ -0,0 +1,16 @@
...
+- name: DIETClassifier
+  random_seed: 42
+  epochs: 5
+- name: ResponseSelector
+  random_seed: 42
+  epochs: 5
+policies:
+- name: TEDPolicy
+  random_seed: 42
+  epochs: 5
```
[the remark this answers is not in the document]
jupyterjazz: In other words, let's just have a trained model and use that for testing?

rasa/core/test.py
jupyterjazz: At this point, target and predicted intents are the same (there's no mismatch) so I don't think we're mixing anything. 
`intent_gold` is for labeled intent and is not a prediction. Here we need response selector prediction in order to extract it in the next step.

data/test_retrieval_intents_in_test_stories/config.yml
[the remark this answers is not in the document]
jupyterjazz: Got it, `MemoizationPolicy` should be enough for my tests.

data/test_retrieval_intents_in_test_stories/config.yml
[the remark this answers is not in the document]
jupyterjazz: Of course you meant config 😄 but what about modifying an existing bot instead of adding a new one?

rasa/core/test.py
```diff
@@ -600,18 +614,48 @@ def _get_e2e_entity_evaluation_result(
...
+    """Get the name of predicted action.
+
+    If the action is instance of `ActionRetrieveResponse`, we need to return full
+    action name with its retrieval intent (e.g. utter_faq/is-this-legit).
+    The only case when we should not do it is when an expected action given in
+    a test story is a retrieval action but it's not specified in the test story.
+    To illustrate this, we're basically avoiding this unnecessary mismatch:
+    utter_faq (expected) != utter_faq/is-this-legit (predicted).
+    In this case or if the action isn't instance of `ActionRetrieveResponse`,
+    the function returns only the action name (e.g. utter_faq).
```
ka-bu, line 631: How are we computing metrics for these results? Do we still count "utter_faq/wrong-response" as correct regarding the intent classification then? Do we have separate metrics regarding the response selection? ... Doesn't really look like it or maybe I'm looking at the wrong modules.

jupyterjazz: This code concerns only `run_core_test` part. Response selector is evaluated in the `run_nlu_test` and doesn't depend on this code. 

However, now that you pointed this out, I looked at the response selector evaluation and I think there are some problems... I'll look into it today.

tests/core/test_test.py
jupyterjazz: It looks difficult to test that, YAML readers don't extract comments, but what I'm testing there is exactly what goes to the comments. So, I think it's not necessary, but if you think we need that I can figure out and add it somehow.

data/test_response_selector_bot/tests/test_stories.yml
```diff
@@ -0,0 +1,30 @@
...
+  - action: utter_chitchat/ask_name
+
+- story: test 1 - test retrieving full utter action when predicted wrongly
+  steps:
+  - user: |
+      What is your name?
+    intent: chitchat/ask_name
+  - action: utter_chitchat/ask_weather
+
+- story: test 2 - test that works well without retrieval intents 1
```
wochinge, line 18: shouldn't this one fail due to the wrong `utter_goodbye`?

rasa/core/actions/action.py
```diff
@@ -359,6 +354,37 @@ def intent_name_from_action(action_name: Text) -> Text:
         """Resolve the name of the intent from the action name."""
         return action_name.split(UTTER_PREFIX)[1]
 
+    def get_full_retrieval_name(
+        self, tracker: "DialogueStateTracker"
+    ) -> Optional[Text]:
+        """Returns full retrieval name for the action.
+
+        Extracts retrieval intent from response selector and
+        returns complete action utterance name.
```
wochinge, line 363: Can we please follow the Google spec for docstrings? I think not following it exactly is only okay for one liners

tests/core/test_test.py
```diff
@@ -127,6 +128,49 @@ async def test_testing_warns_if_action_unknown(
     assert "which is not part of the training data / domain" in output
 
 
+async def test_testing_with_utilizing_retrieval_intents(
+    response_selector_agent: Agent,
+    response_selector_test_stories: Path,
+    response_selector_results: Path,
+):
+    if not response_selector_results.exists():
+        os.mkdir(response_selector_results)
```
wochinge, line 137:
```suggestion
        response_selector_results.mkdir()
```

tests/core/test_test.py
```diff
@@ -127,6 +128,49 @@ async def test_testing_warns_if_action_unknown(
...
+        agent=response_selector_agent,
+        e2e=True,
+        out_directory=response_selector_results,
+        disable_plotting=True,
+        warnings=False,
+    )
+    failed_stories_path = response_selector_results / "failed_test_stories.yml"
+    failed_stories = read_yaml(read_file(failed_stories_path, "utf-8"))
+    # check that the intent is shown correctly in the failed test stories file
+    target_intents = {
```
wochinge, line 150: optional: This test is a bit hard to read - could we just create another test file with the expected output and then compare these files?

data/test_response_selector_bot/tests/test_stories.yml
jupyterjazz: Yes, it does fail. And I'm checking `failed_test_stories.yml` to see if it failed correctly.

wochinge: should we fix the comment then?

tests/core/test_test.py
jupyterjazz: You mean writing these dictionaries (`target_intents` and `target_actions`) into a file and reading it during this test?

wochinge: I meant having a `"expected_failed_test_stories.yml" which you can just compare to the `failed_test_stories.yml` from these tests by comparing the strings instead of doing the workarounds you're currently doing

PR #9667, opened 2021-09-17
3.0 architecture revamp/9336/config validator

### LINKED ISSUE #9336: implement `ConfigValidator`

ka-bu:
**Proposed changes**:
- closes #9336 


rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+            )
+
+        if training_data.entity_examples and self._component_types.isdisjoint(
+            TRAINABLE_EXTRACTORS
+        ):
+            rasa.shared.utils.io.raise_warning(
+                "You have defined training data consisting of entity examples, but "
+                "your NLU pipeline does not include an entity extractor trained on "
+                "your training data. To extract non-pretrained entities, add one of "
+                f"{types_to_str(TRAINABLE_EXTRACTORS)} to your pipeline."
```
wochinge, line 176: can we add a docs link to the NLU components?

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+                f"You have defined training data with regexes, but "
+                f"your NLU pipeline does not include a 'RegexFeaturizer' or a "
+                f"'RegexEntityExtractor'. To use regexes, include either a "
+                f"'{RegexFeaturizerGraphComponent.__name__}' or a "
+                f"'{RegexEntityExtractorGraphComponent.__name__}' in your pipeline."
+            )
+
+        # NOTE: bugs
+        # 1. removed the check that complains if neither "CRFEntityExtractor",
+        #   nor "DIETClassifier" are contained in case a lookup table is given...
```
wochinge, line 207: Does DIET ignore all features by the `RegexFeaturizer` or only the lookup ones?

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+        #    b) we definitely should warn if "patterns" is defined but there
+        #       is no regex featurizer
+        #       (this is a new warning added under the if "CRFEntityExtractor...")
+
+        if training_data.lookup_tables and self._component_types.isdisjoint(
+            [RegexFeaturizerGraphComponent, RegexEntityExtractorGraphComponent],
+        ):
+            rasa.shared.utils.io.raise_warning(
+                f"You have defined training data consisting of lookup tables, but "
+                f"your NLU pipeline does not include a "
```
wochinge, line 225: maybe let's not use pipeline as this would outdate as the configuration format changes, what do you think? 
```suggestion
                f"your NLU configuration does not include a "
```

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+
+        Raises:
+            `InvalidConfigException`
+        """
+        featurizers: List[SchemaNode] = [
+            node
+            for node in self._graph_schema.nodes.values()
+            if issubclass(node.uses, Featurizer2)
+        ]
+        Featurizer2.raise_if_featurizer_configs_are_not_compatible(
```
wochinge, line 389: how about using protocols for this? That would avoid that users / we have to implement subclasses

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+        self._warn_if_no_rule_policy_is_contained()
+        self._raise_if_domain_contains_form_names_but_no_rule_policy_given(domain)
+        self._raise_if_a_rule_policy_is_incompatible_with_domain(domain)
+        self._warn_if_priorities_are_not_unique()
+        self._warn_if_rule_based_data_is_unused_or_missing(story_graph=story_graph)
+
+    def _warn_if_no_rule_policy_is_contained(self) -> None:
+        """Warns if there is no rule policy among the given policies.
+
+        NOTE: core/policies/_ensemble/check_for_important_policies
```
wochinge, line 416: Can we drop the notes as they'll be invalid once we drop the old things?

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+                )
+
+    def _warn_if_priorities_are_not_unique(self) -> None:
+        """Warns if the priorities of the policies are not unique.
+
+        NOTE: _ensemble/check_priorities
+        """
+        priority_dict = defaultdict(list)
+        for schema_node in self._policy_schema_nodes:
+            if "priority" in schema_node.config:
```
wochinge, line 477: don't we need to get the default priority for the policy otherwise?

rasa/core/policies/rule_policy.py
```diff
@@ -772,7 +787,7 @@ def train(
         Returns:
             The resource which can be used to load the trained policy.
         """
-        self._validate_against_domain(domain)
+        self.raise_if_incompatible_with_domain(self.config, domain)
```
wochinge, line 790: we can drop this one then as we already have this in the `ConfigValidator`, right?

tests/core/policies/test_rule_policy.py
```diff
@@ -3044,3 +3044,9 @@ def test_invalid_fallback_action_name(policy_with_config: Callable[..., RulePoli
 
     with pytest.raises(InvalidDomain):
         policy.train([], Domain.empty())
+
+
+def test_raise_if_incompatible_with_domain():
```
wochinge, line 3049: 🙌🏻

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+from rasa.shared.importers.importer import TrainingDataImporter
+from rasa.shared.nlu.training_data.training_data import TrainingData
+import rasa.shared.utils.io
+
+
+# FIXME: check right paths, not just existence on any path (e.g. end2end featurization
+# might be separate from NLU featurization and only one of them might have a
+# tokenizer)
+
+TRAINABLE_EXTRACTORS = {
```
wochinge, line 56: You can get exactly the same information from the `DefaultV1Recipe` [here](https://github.com/RasaHQ/rasa/blob/8f1469a1fd913ae165e536a91c06e35cfd0e112f/rasa/engine/recipes/default_recipe.py#L200)

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+        domain = importer.get_domain()
+        self._validate_core(story_graph, domain)
+
+    def _validate_nlu(self, training_data: TrainingData) -> None:
+        """Validates whether the configuration matches the training data.
+
+        Args:
+           training_data: training_data
+        """
+        # TODO: raise if training data is empty?
```
wochinge, line 133: That's fine, right? e.g. if we train a Core only model

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+        # TODO: where do we check whether NLU training data is consistent with
```
wochinge, line 134: I think nowhere as NLU historically didn't use a `Domain`. As Core/NLU more and more merge I think such as check would definitely make sense to implement at some point 👍🏻

wochinge (commented):
Nice one! 

When reviewing I first had a couple of suggestions about moving things to the Recipe for the validation but the more I think about it I think it's fine. As the `Recipe` is responsible for adding / not adding the validator to the schema it's okay if the validator is tied to the implementation details of the recipe itself. The one thing I'd change is that I'd emphasize the coupling of Recipe / ConfigValidator by renaming `ConfigValidator` to `DefaultV1ConfigValidator`. What do you think?

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,532 @@
...
+        """
+        self._graph_schema = graph_schema
+        self._component_types = set(node.uses for node in graph_schema.nodes.values())
+        self._policy_schema_nodes: List[SchemaNode] = [
+            node
+            for node in self._graph_schema.nodes.values()
+            if issubclass(node.uses, PolicyGraphComponent)
+        ]
+
+    def validate(self, importer: TrainingDataImporter) -> None:
```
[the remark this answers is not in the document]
ka-bu: Argh, yes, should've noticed that one :)

rasa/graph_components/validators/config_validator.py
ka-bu: DIET is not using the `training_data.lookup_table` anywhere (otherwise, I'm making some mistake here - currently refactoring that data pipeline and testing it 😅 ). Features added to the messages via the `RegexFeaturizer` would be used - iff it exists. So if a DIETClassifier is in the pipeline, the lookup table might still be un-used.

rasa/graph_components/validators/config_validator.py
ka-bu: Wasn't sure what to use. In some places, I called it "graph". Will  (try to) switch to "configuration" (everywhere). Seems the most generic.

rasa/graph_components/validators/config_validator.py
ka-bu: This is a class method that tests if the names of the featurizers overlap which would cause trouble downstream. I'm not sure where protocols would be useful here?

ka-bu: And in general I think it would be cool if we could refactor the base classes such that if we use e.g. a Featurizer we know / it is guaranteed that <insert some feature logic here> is done to a message. That would avoid confusion around what featurizer need to do to make them compatible with everything else (which is why we have the strange lexical syntactic featurizer issue now).

rasa/graph_components/validators/config_validator.py
ka-bu: yep sure - this was just for making coding / reviewing easier

rasa/graph_components/validators/config_validator.py
ka-bu: ouch, yep

rasa/core/policies/rule_policy.py
ka-bu: I'd keep it because it's a quick check and someone might want to play around with this component as standalone and then it might be handy, wdyt?

rasa/graph_components/validators/config_validator.py
ka-bu: ah, perfect :)

rasa/core/policies/rule_policy.py
wochinge: I'd say they're on their own risk 😁  Both options are fine for me 👍🏻

rasa/graph_components/validators/config_validator.py
wochinge: Thanks for the explanation! So the actual warning should be: "You need `RegexFeaturizer + patterns enabled and a trainable intent classifier" to make use of lookup tables?

rasa/graph_components/validators/config_validator.py
wochinge: I personally find it very hard to deal with old vs new names. E.g. "NLU components" just perfectly describes which part in the graph I mean but it technically doesn't make a lot of sense anymore 🙈

rasa/graph_components/validators/config_validator.py
wochinge: > I'm not sure where protocols would be useful here?

Yeah, sorry. Completely misread that one 🤦🏻 Is it weird that all the checks are part of the ConfigValidator but this one check is part of the `Featurizer` class? Should we move it into `ConfigValidator`?

I think I need some more explanations around the last point 🙈 Which lexical featurizer issues are you referring to?

rasa/graph_components/validators/config_validator.py
ka-bu: Oh, ok, now I see what you mean / why these checks where like this. 

The check should be " RegexFeaturizer + something consuming those features " and in NLU the CRF entity extractor with "patterns" and the DIETClassifier are really the only consumers 😅 . SklearnIntentClassifier could/should also be but that only makes use of dense sentence features for some reason. In Core, TED should pick those features up as well... but I guess we ignore this for now with v1.  I'll just add these checks back in then.

rasa/graph_components/validators/config_validator.py
ka-bu: I think if we use "NLU components" for just the components and "NLU configuration" for the part of the graph that is applied per message (and "Core's NLU configuration" for the part of the "Core configuration using NLU components", then we'd have an unambiguous naming convention 🤔  - Not saying it's a good one :D

rasa/graph_components/validators/config_validator.py
ka-bu: > one check is part of the Featurizer class?
Because that check requires understanding why/when featurizers are incompatible, I thought it might be better to keep that logic closer to the Featurizers. Do we need to make that visible here?
>  lexical featurizer issues
that was the one not producing sentence features and the only reason why the "add features to message" function had mandatory sequence but optional sentence features :D - though I'm not sure there's a good way to enforce that through interface design and whether there is much code to be shared - maybe protocols + tests are the way to make those assumptions explicit 🤔

rasa/graph_components/validators/config_validator.py
ka-bu: It's not in main yet - shall we just leave this PR open until that recipe is merged so I can reuse that information?

wochinge: How merging it like it is and then I/you put up another PR which integrate them on the integration branch?

rasa/graph_components/validators/config_validator.py
wochinge: > Do we need to make that visible here?

Both is fine - it's rather a nit pick as the implementation currently is very high level and only check whether we have some duplicate aliases 🤷🏻 

I think the code sharing through the interfaces makes us quite inflexible because although a lot of stuff is similar, some components are always different.

rasa/graph_components/validators/config_validator.py
ka-bu: ... btw, I just mixed up regexes and lookup tables --  the same applies (should apply) to the regex part of the training data 🙈 . If training data has regexes, and a regex extractor (even with `use_regexes == False`) and we have no `DIETClassifier` then we will not get a warning at the moment. I'll revert to the old logic for now, put in some `TODO`s, and (commented out) tests to illustrate what's missing.

Think we should keep this in mind when revisiting the "required components". We could add something like "possible/providing/<insert better name here> components". Something to reflect that e.g. SklearnIntentClassifier does not only need dense featurizers but won't use the sparse features produced by the RegexFeaturizer.

wochinge: Can we skip the Todos and create issues instead? We'll never do them otherwise 😄

ka-bu: 👍

ka-bu: https://github.com/RasaHQ/rasa/issues/9705

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
+from __future__ import annotations
```
wochinge, line 1: Can we rename the module `default_recipe_validator` then? (or something in this direction)

wochinge: (the test module as well)

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+        self, training_data: TrainingData
+    ) -> None:
+        """Validates that all training data will be consumed by some component.
+
+        For example, if you specify response examples in your training data, but there
+        is no `ResponseSelector` component in your configuration, then this method
+        issues a warning.
+
+        Args:
+            training_data: training data
```
wochinge, line 152: There is not much to say about `training_data` but just repeating the param name is not ideal in my opinion. How about something like this?
```suggestion
            training_data: The training data for the NLU components.
```

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+    # TODO: add rest / replace
+}
+POLICY_CLASSSES = {  # Note: this is used in tests only, should belong elsewhere anyway
+    TEDPolicyGraphComponent,
+    MemoizationPolicyGraphComponent,
+    RulePolicyGraphComponent,
+}
+
+
+def types_to_str(types: Iterable[Type]) -> Text:
```
wochinge, line 71:
```suggestion
def _types_to_str(types: Iterable[Type]) -> Text:
```

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+                crf_schema_nodes = [
+                    schema_node
+                    for schema_node in self._graph_schema.nodes.values()
+                    if schema_node.uses == CRFEntityExtractorGraphComponent
+                ]
+
+                has_pattern_feature = False
+                for crf in crf_schema_nodes:
+                    crf_features = crf.config.get("features", [])
+                    has_pattern_feature = "pattern" in itertools.chain(*crf_features)
```
wochinge, line 252: optional: Would it be worth introducing a constant for `pattern`?

wochinge: _theoratically_ there could be multiple CRFEntityExtractors and then it's enough if only one of them uses `pattern`, right? In that case we should probably do something like this?
```suggestion
                    has_pattern_feature = has_pattern_feature or "pattern" in itertools.chain(*crf_features)
```

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+
+        if len(tokenizer_schema_nodes) > 1:
+            names = [
+                schema_node.uses.__name__ for schema_node in tokenizer_schema_nodes
+            ]
+            raise InvalidConfigException(
+                f"The configuration configuration contains more than one tokenizer, "
+                f"which is not possible at this time. You can only use one tokenizer. "
+                f"The configuration contains the following tokenizers: "
+                f"{names}. "
```
wochinge, line 304: should we use `types_to_str`?

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+                f"statistical entity extractors "
+                f"{types_to_str(present_general_extractors)} "
+                f"in your configuration. Specifically both types of extractors will "
+                f"attempt to extract entities of the types "
+                f"{', '.join(overlap_between_types)}. "
+                f"This can lead to multiple "
+                f"extraction of entities. Please read "
+                f"{RegexEntityExtractorGraphComponent.__name__}'s "
+                f"documentation section to make sure you understand the "
+                f"implications: {DOCS_URL_COMPONENTS}#regexentityextractor"
```
wochinge, line 369:
```suggestion
                f"'{RegexEntityExtractorGraphComponent.__name__}' and the "
                f"statistical entity extractors "
                f"{types_to_str(present_general_extractors)} "
                f"in your configuration. Specifically both types of extractors will "
                f"attempt to extract entities of the types "
                f"{', '.join(overlap_between_types)}. "
                f"This can lead to multiple "
                f"extraction of entities. Please read "
                f"'{RegexEntityExtractorGraphComponent.__name__}'s' "
                f"documentation section to make sure you understand the "
                f"implications:", docs=f"{DOCS_URL_COMPONENTS}#regexentityextractor"
```

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+        ] = self._component_types.intersection(TRAINABLE_EXTRACTORS)
+        if len(extractors_in_configuration) > 1:
+            rasa.shared.utils.io.raise_warning(
+                f"You have defined multiple entity extractors that do the same job "
+                f"in your configuration: "
+                f"{types_to_str(extractors_in_configuration)}. "
+                f"This can lead to the same entity getting "
+                f"extracted multiple times. Please read the documentation section "
+                f"on entity extractors to make sure you understand the implications: "
+                f"{DOCS_URL_COMPONENTS}#entity-extractors"
```
wochinge, line 328:
```suggestion
                f"on entity extractors to make sure you understand the implications. ",
                docs=f"{DOCS_URL_COMPONENTS}#entity-extractors"
```

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+                f"{RegexEntityExtractorGraphComponent.__name__}'s "
+                f"documentation section to make sure you understand the "
+                f"implications: {DOCS_URL_COMPONENTS}#regexentityextractor"
+            )
+
+    def _raise_if_featurizers_are_not_compatible(self,) -> None:
+        """Raises or warns if there are problems regarding the featurizers.
+
+        Raises:
+            `InvalidConfigException`
```
wochinge, line 376:
```suggestion
            `InvalidConfigException` in case the featurizers are not compatible
```

rasa/graph_components/validators/config_validator.py
```diff
@@ -0,0 +1,514 @@
...
+        for schema_node in self._policy_schema_nodes:
+            default_priority = schema_node.uses.get_default_config()["priority"]
+            priority = schema_node.config.get("priority", default_priority)
+            name = schema_node.uses.__name__
+            priority_dict[priority].append(name)
+
+        for k, v in priority_dict.items():
+            if len(v) > 1:
+                rasa.shared.utils.io.raise_warning(
+                    f"Found policies {v} with same priority {k} "
```
wochinge, line 469: `type_to_str`?

tests/graph_components/validators/test_config_validator.py
```diff
@@ -0,0 +1,805 @@
...
+from rasa.shared.importers.importer import TrainingDataImporter
+
+
+class DummyImporter(TrainingDataImporter):
+    def __init__(
+        self,
+        training_data: Optional[TrainingData] = None,
+        config: Optional[Dict[Text, Any]] = None,
+        domain: Optional[Domain] = None,
+    ):
```
wochinge, line 53:
```suggestion
    ) -> None:
```

tests/graph_components/validators/test_config_validator.py
```diff
@@ -0,0 +1,805 @@
...
+    validator = DefaultV1ConfigValidator(graph_schema)
+    with pytest.warns(UserWarning, match=match):
+        validator.validate(importer)
+
+    # with 'pattern'
+    crf_schema_node.config = {"features": [["suffix1", "pattern"], ["pos"]]}
+    validator = DefaultV1ConfigValidator(graph_schema)
+    with pytest.warns(None) as records:
+        validator.validate(importer)
+        assert len(records) == 0
```
wochinge, line 328: This tests a different case - should we make it a separate unit test?

tests/graph_components/validators/test_config_validator.py
```diff
@@ -0,0 +1,805 @@
...
+        ],
+    )
+
+    nodes = {"tokenizer": SchemaNode({}, WhitespaceTokenizerGraphComponent, "", "", {})}
+    if policy_type is not None:
+        nodes["some-policy"] = SchemaNode({}, policy_type, "", "", {})
+    graph_schema = GraphSchema(nodes)
+
+    validator = DefaultV1ConfigValidator(graph_schema)
+    monkeypatch.setattr(validator, "_validate_nlu", lambda _: None)
```
wochinge, line 545: why do we need the monkey patching? Can we do without it?

tests/graph_components/validators/test_config_validator.py
```diff
@@ -0,0 +1,805 @@
+from typing import Dict, List, Optional, Set, Text, Any, Tuple, Type
```
wochinge, line 1: I think the following cases are currently not covered:
- missing `EntitySynonymMapper`
- entity roles / groups without DIET / CRF

rasa/graph_components/validators/config_validator.py
ka-bu: > optional: Would it be worth introducing a constant for pattern?
agree :) same for "features"

ka-bu: > theoratically there could be multiple CRFEntityExtractors and then it's enough if only one of them uses pattern, right? In that case we should probably do something like this?

nice observation :) will add that

rasa/graph_components/validators/config_validator.py
ka-bu: saves one line 👍

tests/graph_components/validators/test_config_validator.py
ka-bu: ... for tests?

tests/graph_components/validators/test_config_validator.py
ka-bu: Hm, I've left it in the same test_* because it's testing the same "if and only if". I can parametrize the test, though. Then we get the best of both worlds

tests/graph_components/validators/test_config_validator.py
ka-bu: Some of the monkeypatching is because some of the test ymls have no matching domain file. Shall I relax the checks and just assert a warning is / is not contained in any of the returned user warnings? 
(Current checks assert that the warning is the only one - without the need to check the other warnings explicitly)

ka-bu: (This specific patch can be avoided though - just added a DIETClassifier for that - in case that's what you meant)

tests/graph_components/validators/test_config_validator.py
ka-bu: fun fact: entity roles / groups without DIET/CRF was supposed to be handled by the parametrization with "{DIET ,... }.difference(TRAINABLE_EXTRACTORS)" but the order there is obviously wrong 🤦‍♀️  

(and completely missed EntitySynonymMapper)

tests/graph_components/validators/test_config_validator.py
ka-bu: oh, didn't see that this wasn't a test in the preview, my bad

rasa/graph_components/validators/default_recipe_validator.py
```diff
@@ -0,0 +1,505 @@
...
+        text containing all type names
+    """
+    return ", ".join([type.__name__ for type in types])
+
+
+class DefaultV1RecipeValidator(GraphComponent):
+    """Validates the current graph schema against the training data and domain.
+
+    This validator is just a replacement for the configuration checks done in Rasa 2.x
+    and should only be used for "default v1" configurations.
```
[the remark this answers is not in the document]
ka-bu: I'll remove the 2.8.x reference and just keep the "default v1" 👍

rasa/graph_components/validators/default_recipe_validator.py
```diff
@@ -0,0 +1,505 @@
...
+        )
+
+    def _validate_core(self, story_graph: StoryGraph, domain: Domain) -> None:
+        """Validates whether the configuration matches the training data.
+
+        Args:
+           story_graph: a story graph (core training data)
+           domain: the domain
+        """
+        if not self._policy_schema_nodes and story_graph.story_steps:
```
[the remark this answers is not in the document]
ka-bu: I think that if it is necessary that we do that validation in the default recipe validation, then we can keep it there -- but it looks like it isn't necessary and so I'd rather keep it here because I'd expect the (complete) validation logic inside the validation component and would be surprised later if part of the cases is handled elsewhere. 
Also this validator is unaware of the training type which (I'm guessing here) is no replacement for the "what training data do we have" but for the CLI mode? Wdyt? Shall I merge as is for now? (or add training type flag to the validation here?)

PR #9689, opened 2021-09-21
fix origin check in Features.combine; add tests

ka-bu:
**Proposed changes**:
- remove origin check in `Features.combine` if `expected_features` is `None` because there is nothing we can check - the origins must differ as it is should be the only thing that differs (attribute, type, sparseness are identical)
- added a unit test case 
- adapted tests for `Features.reduce` which was misusing that "feature"

### DIFF
--- a/rasa/shared/nlu/training_data/features.py
+++ b/rasa/shared/nlu/training_data/features.py
@@ -280,22 +280,12 @@ class Features:
                 ):
                     if expected != actual:
                         raise ValueError(
-                            f"Expected {expected} to be the origin of the {idx}-th "
+                            f"Expected '{expected}' to be the origin of the {idx}-th "
                             f"feature (because of `origin_of_combination`) but found a "
-                            f"feature from {actual}."
+                            f"feature from '{actual}'."
                         )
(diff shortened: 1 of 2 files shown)


wochinge (approved):
💯 Thanks for the quick fix!

jupyterjazz (approved):
Looks good 🚀 I just don't understand why we had this check in the first place (I saw your comment on slack but I still couldn't understand it 😄)

ka-bu:
> Looks good 🚀 I just don't understand why we had this check in the first place (I saw your comment on slack but I still couldn't understand it 😄)

There was an earlier version that processed a `List[List[Features]]` (i.e. combined each inner `List[Features]`, where we'd expect that each `List[Features]` contains the same sequence of origin information, which we could do without any `expected_origin` information by e.g. checking how many different combinations of origins those `List[Features]`s have (or by just comparing all `List[Features]` with the origins from the first `List[Features]` contained in the  `List[List[Features]]`). Then one `List` was dropped but that idea of checking (despite no expected origins being given) was kept :D.

PR #9696, opened 2021-09-22
fine tuning validator (changing configs to be validated)

ka-bu:
**Proposed changes**:
- validating configs contained in graph schema from execution context instead of config from importer

### DIFF
--- a/rasa/graph_components/validators/finetuning_validator.py
+++ b/rasa/graph_components/validators/finetuning_validator.py
@@ -145,7 +145,7 @@ class FineTuningValidator(GraphComponent):
                 )
         self._fingerprints[FINGERPRINT_VERSION] = rasa_version
 
-        fingerprint_config = self._get_fingerprint_of_configs_without_epochs()
+        fingerprint_config = self._get_fingerprint_of_schema_without_epochs_config()
         self._compare_or_memorize(
             fingerprint_key=FINGERPRINT_CONFIG_WITHOUT_EPOCHS_KEY,
             new_fingerprint=fingerprint_config,
@@ -231,19 +231,18 @@ class FineTuningValidator(GraphComponent):
         domain.responses = {}
         return domain.fingerprint()
 
-    def _get_fingerprint_of_configs_without_epochs(self,) -> Text:
+    def _get_fingerprint_of_schema_without_epochs_config(self,) -> Text:
         """Returns a fingerprint of the given configuration with "epoch" keys removed.
 
         Returns:
             fingerprint
         """
-        config_copies = [
-            copy.deepcopy(schema_node.config)
-            for schema_node in self._execution_context.graph_schema.nodes.values()
-        ]
-        for config in config_copies:
-            config.pop("epochs", None)
-        return rasa.shared.utils.io.deep_container_fingerprint(config_copies)
+        schema = self._execution_context.graph_schema
+        for schema_node in schema.nodes.values():
+            config_copy = copy.deepcopy(schema_node.config)
+            config_copy.pop("epochs", None)
+            schema_node.config = config_copy
+        return rasa.shared.utils.io.deep_container_fingerprint(schema.as_dict())
 
     @classmethod
     def create(
--- a/rasa/graph_components/validators/finetuning_validator.py
+++ b/rasa/graph_components/validators/finetuning_validator.py
@@ -237,12 +237,13 @@ class FineTuningValidator(GraphComponent):
         Returns:
             fingerprint
         """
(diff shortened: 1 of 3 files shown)


rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -230,32 +231,19 @@ def _get_fingerprint_of_domain_without_responses(domain: Domain) -> Text:
...
-
-        for key in ["pipeline", "policies"]:
-            if copied_config.get(key):
-                for p in copied_config[key]:
-                    if "epochs" in p:
-                        del p["epochs"]
-
-        return rasa.shared.utils.io.deep_container_fingerprint(copied_config)
+        config_copies = [
+            copy.deepcopy(schema_node.config)
```
wochinge, line 241: thanks for the `deepcopy` 😆 💯

wochinge: I think we also need to evaluate the overall structure of the graph, don't we? E.g. things like changed order. How about also including the `needs` section in the fingerprint?

wochinge: And I think we need a change [here](https://github.com/RasaHQ/rasa/blob/f5b2ff4d208f5d376bd65c38eb33f8deb3a67b6a/rasa/engine/training/graph_trainer.py#L110) as well to put the `train_schema` in the execution context instead of the fingerprint schema (the validation will run during the fingerprint run - currently it would get the wrong schema)

wochinge (commented):
💯 Thanks for being so flexible with your time and jumping on this!

rasa/graph_components/validators/finetuning_validator.py
ka-bu: >I think we also need to evaluate the overall structure of the graph, don't we? 

how could I overlook this 😱  - think if we want to catch everything, we also need the "uses" part

ka-bu: >  to put the train_schema in the execution context instead of the fingerprint schema (the validation will run during the fingerprint run - currently it would get the wrong schema)

will that break something else if we just replace it? 🤔  will look into it later

wochinge: It shouldn't as nothing is really using the graph schema in the execution context 🤔

tests/graph_components/validators/test_finetuning_validator.py
```diff
@@ -82,6 +93,22 @@ def inner(
     return inner
 
 
+class DummyNluDataImporter(NluDataImporter):
```
wochinge, line 96: nit
```suggestion
class DummyNLUDataImporter(NluDataImporter):
```

tests/graph_components/validators/test_finetuning_validator.py
```diff
@@ -162,147 +189,136 @@ def test_validate_after_adding_action_to_domain(
...
+    schema1 = _get_example_schema()
+    validate = get_validation_method(
+        finetuning=False, load=False, nlu=nlu, core=core, graph_schema=schema1
+    )
+    validate(importer=EmptyDataImporter())
+
+    # change schema
+    schema2 = copy.deepcopy(schema1)
+    schema_node = schema2.nodes["node-0"]
+    if what == "uses":
```
wochinge, line 305: what can only be `uses` at the moment?

tests/graph_components/validators/test_finetuning_validator.py
```diff
@@ -162,147 +189,136 @@ def test_validate_after_adding_action_to_domain(
...
-        self.training_data = TrainingData(training_examples=messages)
-        self.config = config
-
-    def get_config(self) -> Dict:
-        return self.config
+@pytest.mark.parametrize(
+    "nlu, core, what",
+    [
+        (nlu, core, what)
+        for what in ["uses"]  # , 'needs', 'fn', 'config' ]
```
wochinge, line 285: do you want to comment in the other parts as well?

wochinge (commented):
Two tiny comments. Looks great otherwise!

tests/graph_components/validators/test_finetuning_validator.py
ka-bu: oh, sorry

tests/graph_components/validators/test_finetuning_validator.py
ka-bu: yep

wochinge (approved):
💯

PR #9716, opened 2021-09-24
Implement new default action to set slots after each user turn

### LINKED ISSUE #9364: 3.0 slot mappings - rasa-oss implementation pII: add new default action + adapt `forms.py` + add logic in `processor.py`

ancalita:
**Proposed changes**:
- Closes #9364 

### DIFF
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -1081,7 +1081,7 @@ class ActionExtractSlots(Action):
                 elif should_fill_text_slot:
                     value = [tracker.latest_message.text]
                 elif should_fill_custom_slot:
-                    custom_evts = self._execute_custom_action(
+                    custom_evts = await self._execute_custom_action(
                         mapping, output_channel, nlg, tracker, domain
                     )
                     slot_events.extend(custom_evts)
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -83,7 +83,7 @@ def default_actions(action_endpoint: Optional[EndpointConfig] = None) -> List["A
         TwoStageFallbackAction(action_endpoint),
         ActionUnlikelyIntent(),
         ActionBack(),
-        ActionExtractSlots(),
+        ActionExtractSlots(action_endpoint),
     ]
 
 
@@ -958,6 +958,10 @@ class ActionExtractSlots(Action):
     according to assigned slot mappings.
     """
 
+    def __init__(self, action_endpoint: Optional[EndpointConfig]) -> None:
+        """Initializes default action extract slots."""
+        self.action_endpoint = action_endpoint
+
     def name(self) -> Text:
         """Returns action_extract_slots name."""
         return ACTION_EXTRACT_SLOTS
@@ -973,13 +977,13 @@ class ActionExtractSlots(Action):
         slot_events: List[Event] = []
         extracted_entities = tracker.latest_message.entities
 
-        default_slots = [
+        default_slots = {
             rasa.shared.core.constants.REQUESTED_SLOT,
             rasa.shared.core.constants.SESSION_START_METADATA_SLOT,
             rasa.shared.core.constants.SLOT_LISTED_ITEMS,
             rasa.shared.core.constants.SLOT_LAST_OBJECT,
             rasa.shared.core.constants.SLOT_LAST_OBJECT_TYPE,
-        ]
+        }
         user_slots = [slot for slot in domain.slots if slot.name not in default_slots]
 
         for slot in user_slots:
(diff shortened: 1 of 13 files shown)


rasa/core/actions/action.py
```diff
@@ -940,3 +949,141 @@ def name(self) -> Text:
...
+                            mapping.get(ENTITY_ATTRIBUTE_ROLE),
+                            mapping.get(ENTITY_ATTRIBUTE_GROUP),
+                        )
+                    )
+                elif should_fill_intent_slot:
+                    value = [mapping.get("value")]
+                elif should_fill_text_slot:
+                    value = [tracker.latest_message.text]
+                elif should_fill_custom_slot:
+                    custom_action = mapping.get("action")
```
alwx, line 1041: I would consider moving the execution of a custom action to a separate function — just because it can be treated as an individual unit and because the `run` function is quite huge already.

rasa/core/actions/action.py
```diff
@@ -940,3 +949,141 @@ def name(self) -> Text:
...
+                )
+
+                should_fill_intent_slot = (
+                    mapping["type"] == str(SlotMapping.FROM_INTENT)
+                    and intent_is_desired
+                )
+
+                should_fill_text_slot = (
+                    mapping["type"] == str(SlotMapping.FROM_TEXT) and intent_is_desired
+                )
```
alwx, line 1025: It doesn't look like we really need to store all that (`should_fill_text_slot`, `should_fill_intent_slot`, etc) and have separate variables for them.

rasa/core/actions/action.py
```diff
@@ -940,3 +949,141 @@ def name(self) -> Text:
...
+        """Runs action. Please see parent class for the full docstring."""
+        slot_events: List[Event] = []
+
+        default_slots = {
+            rasa.shared.core.constants.REQUESTED_SLOT,
+            rasa.shared.core.constants.SESSION_START_METADATA_SLOT,
+            rasa.shared.core.constants.SLOT_LISTED_ITEMS,
+            rasa.shared.core.constants.SLOT_LAST_OBJECT,
+            rasa.shared.core.constants.SLOT_LAST_OBJECT_TYPE,
+        }
```
alwx, line 985: `rasa.shared.core.constants` already contains `DEFAULT_ACTION_NAMES` — can we do the same with slot names? Or you think it's better to keep them here?

Asking because my opinion is that this method (`run`) is quite a long one, and I would be happy if we restructure it a bit.

alwx (changes requested):

alwx:
It looks good, just wrote a couple of comments.

rasa/core/actions/action.py
ancalita: I actually thought it's more readable like this and also consistent to how slots get extracted in `FormAction` too 🤔

rasa/core/actions/action.py
```diff
@@ -940,3 +950,147 @@ def name(self) -> Text:
...
+                            mapping.get(ENTITY_ATTRIBUTE_ROLE),
+                            mapping.get(ENTITY_ATTRIBUTE_GROUP),
+                        )
+                    )
+                elif should_fill_intent_slot:
+                    value = [mapping.get("value")]
+                elif should_fill_text_slot:
+                    value = [tracker.latest_message.text]
+                elif should_fill_custom_slot:
+                    custom_evts = await self._execute_custom_action(
```
alwx, line 1084: 🚀

rasa/core/actions/action.py
alwx: Well, that might be true. Let's skip it then.

alwx (approved):
Looks good, thanks a lot!

PR #9727, opened 2021-09-27
Codeclimate integration

### LINKED ISSUE #9253: Migrate rasa repo to codeclimate
**Description of Problem**:

Now that Rasa X is fully migrated to codeclimate, we should consider doing the same for the rasa repo. Would allow us more coverage control than coveralls.

**Overview of the Solution**:

Migrate `rasa` to codeclimate, and remove coveralls integration.
Check with the team regarding exclusion patterns & other settings.
Disable all code smells (already done via CI)

**Definition of Done**:

mprazz:
**Proposed changes**:
- Fixes #9253 
- Move from Coveralls to CodeClimate!
- Repo here --> [codeclimate link](https://codeclimate.com/github/RasaHQ/rasa)
- Coverage percentage will now show up until this code runs on `main` , FYI!
- [The full report](https://codeclimate.com/repos/5d668ca8f6a09f019e010459/settings/test_reports/6153122be2dc0b2d39008455) has been successfully uploaded in the latest test run (not sure if you can see it without being a repo admin though) -- basically, it works lol

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -319,29 +319,57 @@ jobs:
         DD_ARGS: --ddtrace
       run: |
         make ${{ matrix.test }}
+        poetry run coverage xml -o ${{ github.workspace }}/tests/${{ matrix.test }}-coverage.xml --include="rasa/*"
 
-    - name: Send Coverage Report 📊
-      if: needs.changes.outputs.backend == 'true' && matrix.python-version == 3.7 && matrix.os != 'windows-latest'
-      continue-on-error: true
-      env:
-        COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        COVERALLS_SERVICE_NAME: github
-        COVERALLS_PARALLEL: true
-      run: poetry run coveralls
+    - name: Store coverage reports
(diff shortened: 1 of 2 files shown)


.github/workflows/continous-integration.yml
```diff
@@ -319,29 +319,57 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          path: ${{ github.workspace }}/tests_coverage
+
+      - name: Move all xml files to parent dir
+        if: needs.changes.outputs.backend == 'true'
+        run: |
+          dir1="${{ github.workspace }}/tests_coverage"
+          subs=`ls ${{ github.workspace }}/tests_coverage`
+          dir2="${{ github.workspace }}/tests_coverage/final"
```
alwx, line 356: A very very minor and not something you necessarily need to agree with: I would prefer something like `sourceDir` and `destinationDir` instead of `dir1` and `dir2` :)

alwx (approved):
One small comment but it looks very good in general. Thanks!

.github/workflows/continous-integration.yml
mprazz: ah you're totally right, forgot to clean out the code on here, good catch!

PR #9728, opened 2021-09-27
add tmate session to debug runner

HotThoughts:
**Proposed changes**:
Add a step right after backend's test. If the test fails, setup a tmate session, which allows us ssh into it and debug directly from there.

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -302,6 +302,10 @@ jobs:
       run: |
         make ${{ matrix.test }}
 
+    - name: Setup tmate session
+      if: failure()
+      uses: mxschmitt/action-tmate@v3
+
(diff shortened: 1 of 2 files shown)


HotThoughts:
@koernerfelicia Instead of setting up a new instance, I find a way to ssh into GitHub's runners. Would you interested to try it? I tested it and successfully got into the runner. 

**How to get in?**
Check out the "Setup tmate session" step, you should a ssh command printed in the log like this `SSH: ssh tptgAEuESMKnjqdv6rFXatmCs@nyc1.tmate.io`. Then ssh into it by `ssh tptgAEuESMKnjqdv6rFXatmCs@nyc1.tmate.io` (the username changes each time).

koernerfelicia (approved):

Issue #9664, opened 2021-09-17
Success markers: read config


aeshky:
**Overview of solution:**
Implement a function to read the markers config file and produce an object containing the conditions. 

**Input:** yaml config file
**Output:** object containing conditions. 

See [implementation proposal](https://www.notion.so/rasa/Success-Markers-Implementation-Proposal-79cb6f2b96584387aef6ab5c60db4b4c) for project overview. 

**Definition of done:** 
- [x] function implemented 
- [x] function unit tested

m-vdb:
💡 @aeshky can you make sure to close issues instead of moving them to the "Done" column manually? The automation works this way: if you close the issue, it's automatically moved to "Done".

PR #9729, opened 2021-09-27
Success markers: read config

aeshky:
**Proposed changes**:
- implemented "success markers: read config" according to #9664
- followed a similar way of reading the yaml file to [Domain.py](https://github.com/RasaHQ/rasa/blob/main/rasa/shared/core/domain.py)


rasa/shared/constants.py
```diff
@@ -86,6 +86,10 @@
 DEFAULT_CORE_SUBDIRECTORY_NAME = "core"
 DEFAULT_NLU_SUBDIRECTORY_NAME = "nlu"
 DEFAULT_CONVERSATION_TEST_PATH = "tests"
+DEFAULT_MARKERS_PATH = "markers"
+DEFAULT_MARKERS_CONFIG_PATH = "markers/config"
+DEFAULT_MARKERS_OUTPUT_PATH = "markers/output"
+DEFAULT_MARKERS_STATS_PATH = "markers/stats"
```
aeshky, line 92: This is how I propose to structure the folder containing all the markers data (input and output). These constants can be used as defaults (but are not currently used in this issue).

rasa/core/evaluation/markers.py
```diff
@@ -0,0 +1,120 @@
...
+        except YamlException as e:
+            e.filename = original_filename
+            raise e
+
+    @classmethod
+    def from_directory(cls, path: Text) -> Dict:
+        """Loads and appends multiple configs from a directory tree."""
+        config = cls.empty_config()
+        for root, _, files in os.walk(path, followlinks=True):
+            for file in files:
```
aeshky, line 72: Code climate says this is too cognitively complex: https://codeclimate.com/github/RasaHQ/rasa/pull/9729
But I think this is a pretty standard way of generating the filenames in a directory. Do I need to simplify it? 🤔

aeshky:
@alwx and @usc-m, @ancalita was automatically tagged to review, but would either of you also like to take a look since you'll be involved in the success markers implementation? 🙂

usc-m (commented):
Looks good! Have put some comments on some things I think might be worth addressing - still trying to figure out how the read trackers ticket merges with this (separate class in the same file?)

rasa/core/evaluation/markers.py
```diff
@@ -0,0 +1,120 @@
...
+                cannot be read / parsed.
+        """
+        from rasa.shared.data import is_likely_yaml_file
+
+        if not is_likely_yaml_file(filename):
+            return False
+
+        try:
+            # check if the file can be read as a yaml file.
+            rasa.shared.utils.io.read_yaml_file(filename)
```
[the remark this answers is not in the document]
aeshky: Yes, I agree with you. I based this on Domain.py to try an be consistent with the way we process yaml input, but I too didn't like that the yaml file was being read twice. I think the original author wanted to separated checking that the file is a valid yaml from the process of reading it. I can fold them into a single function if consistency with other parts of the codebase is not a huge concern. 
I yet don't know if the function will be used elsewhere. Did you want me to prepend an underscore to the name?

rasa/core/evaluation/markers.py
aeshky: The code climate thing doesn't look like a compulsory test, so I could ignore it in this instance since this is a standard piece of code.

usc-m: Yeah, it's caused me bother before - I think there's plans to switch to a different tool for coverage etc

aeshky:
@usc-m Thanks for the review!

> still trying to figure out how the read trackers ticket merges with this (separate class in the same file?)

Yes, separate class in the same file sounds good! 
Are you wondering how to merge our work? You can create a new branch from this one, then once this is merged you can either rebase your branch onto main, or move your commits to a new branch from main (I always end up with a mess with attempting to `git rebase`, and then decide to instead move my commits to a new branch using `git cherry-pick` 😅).

rasa/core/evaluation/markers.py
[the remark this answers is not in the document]
usc-m: So looking into the Domain one, they have tests to ensure invalid domains are rejected, and otherwise the only place it's really used is for MultiProjectImporter. I'm not sure if that's something that would need extended by this feature? 

I think the tests are a good idea but the multiple reads bothers me a little bit - I suppose it's not critical in this case since it's dealing with a fairly small file, and we can leave it consistent with the rest of the codebase. Might be a possible optimisation thing in future if it comes to that (since it probably wouldn't be hard to attempt a read then return file content if successful else warn)

aeshky:
@usc-m regarding this: 
> still trying to figure out how the read trackers ticket merges with this (separate class in the same file?)

Another option is to create a new file in the folder ` rasa/core/evaluation` and place the tracker reader there.

usc-m:
> Are you wondering how to merge our work?

Nope, haven't started but was going to later today - so was just trying to think if reading trackers belongs to whatever "core" Markers class we have (which means I'd need to create that core class), or if it's the responsibility of some sort of MarkerTrackerLoader (ideally with a better name 😄) that's separate from it but in the same file

aeshky:
> was just trying to think if reading trackers belongs to whatever "core" Markers class we have (which means I'd need to create that core class), or if it's the responsibility of some sort of MarkerTrackerLoader (ideally with a better name 😄) that's separate from it but in the same file

Yeah that's a good question. Both options seem reasonable. Maybe start with a "MarkerTrackerLoader", and then we can decide whether to move it to the core Markers class?

usc-m:
I think I'm actually preferring the TrackerLoader now that I think about it - if we want to expand to workloads that involve accessing trackers that aren't in the tracker store (the ETL stuff we spoke about previously), having a swappable implementation might make that easier to do

aeshky:
Yeah that makes sense. Worth checking what tracker loading functionality we have now (you're probably already doing that) and seeing if we need to expand it, or just call it.

rasa/core/evaluation/markers.py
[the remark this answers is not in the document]
aeshky: Yeah no worries! I moved stuff around, so now I check if the file has the correct extension, and if so, I process it, raising an exception inside read_yaml if it can't be parsed. This way we get the same exception and error message whether we read multiple configs or just one.

usc-m (approved):

aeshky:
🚀

rasa/core/evaluation/markers.py
```diff
@@ -0,0 +1,95 @@
...
+import rasa.shared.utils.common
+from rasa.shared.data import is_likely_yaml_file
+from ruamel.yaml.parser import ParserError
+
+
+class InvalidMarkersConfig(RasaException):
+    """Exception that can be raised when markers config is not valid."""
+
+
+class MarkerConfig:
```
wochinge, line 23: Should this rather be a bunch of functions? As it doesn't even instantiate the class I don't see the purpose - it might even confuse readers as they expect an object but then it's just a dictionary

aeshky: Do classes in python always need to be instantiatable? Or is it a coding convention that we have? I guess I'm using classes to conceptually organise things. Maybe that's not desirable. 

For now we're representing the configs as dicts. If this remains the case, I'm happy to switch these into functions if that's preferred. If we end up needing a more elaborate representation then we might need the class. How does that sound?

wochinge: It's for sure not wrong but I'd very _uncommon_ (in lack of a better word I'd say it's not _pythonic_, e.g. comment [here](https://stackoverflow.com/questions/42757961/a-class-with-all-static-methods#comment72631648_42757961) or comment [here](https://stackoverflow.com/questions/11788195/module-function-vs-staticmethod-vs-classmethod-vs-no-decorators-which-idiom-is)). In cases like these you'd usually rather use modules to organize things than static methods.

In this particular case I'd say that the class is slightly confusing to the reader as they'd expect a class instance to be returned (`@classmethod` and things like `from_yaml` at Rasa are almost always used as constructors). I think the functions would make it more obvious that we are dealing with primitive data objects here.

aeshky: >In this particular case I'd say that the class is slightly confusing to the reader as they'd expect a class instance to be returned (@classmethod and things like from_yaml at Rasa are almost always used as constructors). I think the functions would make it more obvious that we are dealing with primitive data objects here.

Great point. Will refactor 👍

PR #9731, opened 2021-09-28
Update documentation workflow to only publish one version per major

### LINKED ISSUE #9711: update documentation workflow to only publish one version per major
**Goal**

We want to have only one version of the docs per major release. That means that there will be a `2.x` docs and a `3.x` docs. The `2.x` docs will come from the current 2.8 docs, as @m-vdb has already worked on getting the various other 2.x docs merged into this latest branch of the documentation

**Updates to older branches** 
- The 2.8.x branch is the source of truth for the version 2 docs -- any changes that go into that branch can have documentation changes. The workflow should build the documentation whenever there is a new patch release on 2.8.x (not for every push to the .x branch, @m-vdb could clarify this to make sure I understood correctly). 
- Older 2.x branches, like 2.7.x etc, should have the build documentation workflow disabled. 
- Whenever we're ready, we can build the existing docs from that 2.8.x branch -- we don't need to wait for 3.0 to come out to get this in play. 

**Updates for 3.0 docs** 
- For 3.x docs, we are planning to have only one set of documentation, marking changes over time in the docs directly with callouts (e.g. "added in 3.1" in the docs pages themselves, as well as the regular changelog). That means that we only want to build the docs from releases on the latest minor. 
  - I.e. when we are on 3.0.x, we build the docs off of the 3.0.x branch every time we release a new micro. Upon release of 3.1.0, the workflow should build the docs only from the 3.1.x branch, and shouldn't build docs on any micros added to 3.0.x (ideally people are not merging to this, but if they are, it is due to specific backported fixes that those using them are aware of -- docs for backported fixes create confusion as to which later versions of the code the fix is in, so we'll avoid that as much as possible)

tczekajlo:
**Proposed changes**:
- Resolves #9711 

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -88,7 +88,7 @@ jobs:
     name: Build Docs
     runs-on: ubuntu-latest
     needs: [ evaluate_release_tag ]
-    if: github.repository == 'RasaHQ/rasa' && needs.evaluate_release_tag.outputs.check_tag == 'true'   # don't run this for main branches of forks, would fail anyways
+    if: github.repository == 'RasaHQ/rasa' && needs.evaluate_release_tag.outputs.build_docs == 'true'   # don't run this for main branches of forks, would fail anyways
 
     steps:
     - name: Checkout git repository 🕝
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -17,6 +17,7 @@ env:
   DOCS_BRANCH: documentation
   TAG_NAME: ${GITHUB_REF#refs/tags/}
   IS_TAG_BUILD: ${{ startsWith(github.event.ref, 'refs/tags') }}
+  IS_MASTER_BRANCH: ${{ github.ref == 'refs/heads/main' }}
 
 jobs:
   evaluate_release_tag:
@@ -67,7 +68,8 @@ jobs:
         IS_LATEST_VERSION=${{ steps.rasa_get_version.outputs.is_latest_version }}
 
         # the latest major (when 3.0 is out, the latest major is 3.0)
-        if [[ "${IS_LATEST_VERSION}" == "true" ]]; then
+        # build docs on push to the main branch
+        if [[ "${IS_LATEST_VERSION}" == "true" || "${IS_MASTER_BRANCH}" == "true" ]]; then
           echo "::set-output name=build_docs::true"
           exit 0
         fi
@@ -91,7 +93,7 @@ jobs:
     name: Build Docs
     runs-on: ubuntu-latest
     needs: [ evaluate_release_tag ]
-    if: github.repository == 'RasaHQ/rasa' && (needs.evaluate_release_tag.outputs.build_docs == 'true' || github.ref == 'refs/heads/main' )  # don't run this for main branches of forks, would fail anyways
+    if: github.repository == 'RasaHQ/rasa' && needs.evaluate_release_tag.outputs.build_docs == 'true'  # don't run this for main branches of forks, would fail anyways
 
     steps:
     - name: Checkout git repository 🕝
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -2,6 +2,8 @@ name: Publish Documentation
 
 on:
   push:
+    branches:
+    - 'main'
     tags:
     - '**'
 
@@ -14,6 +16,7 @@ env:
   DOCS_FOLDER: docs
   DOCS_BRANCH: documentation
   TAG_NAME: ${GITHUB_REF#refs/tags/}
+  IS_TAG_BUILD: ${{ startsWith(github.event.ref, 'refs/tags') }}
 
 jobs:
   evaluate_release_tag:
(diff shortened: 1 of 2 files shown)


tczekajlo:
@m-vdb Can you 👀  if you have a minute? I hope I got it right 😄

.github/workflows/documentation.yml
```diff
@@ -2,8 +2,6 @@ name: Publish Documentation
 
 on:
   push:
-    branches:
-    - 'main'
```
m-vdb, line 6: I think we still want to build the docs on main (currently available at this URL: https://rasa.com/docs/rasa/next/)

.github/workflows/documentation.yml
```diff
@@ -15,12 +13,82 @@ on:
...
+      run: |
+        # Get latest tagged Rasa version
+        git fetch --depth=1 origin "+refs/tags/*:refs/tags/*"
+        # Fetch branch history
+        git fetch --prune --unshallow
+        LATEST_TAGGED_NON_ALPHA_RASA_VERSION=$(git tag | sort -r -V | grep -E "^[0-9.]+$" | head -n1)
+
+        echo "LATEST_TAGGED_NON_ALPHA_RASA_VERSION=${LATEST_TAGGED_NON_ALPHA_RASA_VERSION}" >> $GITHUB_ENV
+
+        CURRENT_TAG=${GITHUB_REF#refs/tags/}
```
m-vdb, line 41: this seems to be exactly `TAG_NAME`, maybe we can use on one of the 2 env vars?

m-vdb (changes requested):
Nice 🎉 thanks a lot @tczekajlo ! I think there is one part missing, about running the docs script itself [here](https://github.com/RasaHQ/rasa/blob/0795099f5b237e648aed2e46397643f891da0378/scripts/push_docs_to_branch.sh#L17-L18), which is creating a new version of the docs for new majors only, not minors. I think we just need to update the two patterns and we should be good to go (at first sight 🤔 )

.github/workflows/documentation.yml
```diff
@@ -15,12 +15,83 @@ on:
...
+
+        # the latest major (when 3.0 is out, the latest major is 3.0)
+        if [[ "${IS_LATEST_VERSION}" == "true" ]]; then
+          echo "::set-output name=build_docs::true"
+          exit 0
+        fi
+
+        # the previous major (when 3.0 is out, the previous major is 2.0, the latest minor on this version being 2.8)
+        CURRENT_MAJOR_VERSION=$(echo ${LATEST_TAGGED_NON_ALPHA_RASA_VERSION} | awk -F\. '{print $1}')
+        PREVIOUS_MAJOR_LATEST_VERSION=$(git tag | sort -r -V | grep -E "^[0-9.]+$" | grep -vE "^${CURRENT_MAJOR_VERSION}" | head -n1)
```
m-vdb, line 77: 🤔 WDYT about adding a bit of debugging in the CI?

.github/workflows/documentation.yml
```diff
@@ -15,12 +15,83 @@ on:
...
+        fi
+
+        echo "::set-output name=build_docs::false"
+
   docs:
     name: Build Docs
     runs-on: ubuntu-latest
-    if: github.repository == 'RasaHQ/rasa'  # don't run this for main branches of forks, would fail anyways
+    needs: [ evaluate_release_tag ]
+    if: github.repository == 'RasaHQ/rasa' && (needs.evaluate_release_tag.outputs.build_docs == 'true' || github.ref == 'refs/heads/main' )  # don't run this for main branches of forks, would fail anyways
```
m-vdb, line 94: nitpicking: for consistency I'd add another step in the workflow above that sets `build_docs` to `true` if the ref is `main`. But that's a detail / style choice IMO

scripts/push_docs_to_branch.sh
```diff
@@ -33,10 +33,10 @@ NEW_VERSION=
 EXISTING_VERSION=
 if [[ "$GITHUB_REF" =~ $PATTERN_FOR_NEW_VERSION ]]
```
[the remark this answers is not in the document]
tczekajlo: OH, ok. Now I see it 🙈

scripts/push_docs_to_branch.sh
```diff
@@ -14,7 +14,7 @@ set -Eeuo pipefail
 
 TODAY=`date "+%Y%m%d"`
 # we build new versions only for minors and majors
-PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.0$"
+PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.0\\.0$"
 PATTERN_FOR_MICRO_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9][0-9]*$"
```
m-vdb, line 18:
```suggestion
PATTERN_FOR_EXISTING_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[0-9]+$"
```

🤕 sorry for not spotting this one earlier... We need to change this pattern to match new minors too, now (e.g. 3.1.0, etc...). I think it's the last change required, I've doubled check the full logic and it seems in order ✅

m-vdb (changes requested):

m-vdb:
@tczekajlo before merging this PR, I will need to wrap-up [this one](https://github.com/RasaHQ/rasa/pull/9271) (remaining step is to merge the 2.8 docs manually into 2.x). I'll put this up today and ask for a review from CSE. Another thing we'll need to do around the same time is disable the docs workflows on at least 2.7, 2.6 (mostly a safety measure, in case people push new tags there. I don't expect it, but....). I can take care of this and I'll ask a review from you if that's alright

scripts/push_docs_to_branch.sh
```diff
@@ -15,7 +15,7 @@ set -Eeuo pipefail
 TODAY=`date "+%Y%m%d"`
 # we build new versions only for minors and majors
 PATTERN_FOR_NEW_VERSION="^refs/tags/[0-9]+\\.0\\.0$"
-PATTERN_FOR_MICRO_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[1-9][0-9]*$"
+PATTERN_FOR_EXISTING_VERSION="^refs/tags/[0-9]+\\.[0-9]+\\.[0-9]+$"
```
m-vdb, line 18: also needs a change line 38 🤔

tczekajlo: Are you sure? 

for example 

```
GITHUB_REF=/refs/tags/3.0.1
```
So in such a case, we want to update the docs for `3.x`, right?

it will match with `elif [[ "$GITHUB_REF" =~ $PATTERN_FOR_MICRO_VERSION ]]` and `EXISTING_VERSION=$(echo $GITHUB_REF | sed -E "s/^refs\/tags\/([0-9]+)\.([0-9]+)\.[0-9]+$/\1.x/")` will return `3.x`

looks ok to me, or it's Monday morning and I miss something? 😄

m-vdb: I think it's monday morning + not enough precise comment on my end. I meant renaming `PATTERN_FOR_MICRO_VERSION` to `PATTERN_FOR_EXISTING_VERSION` at line 38 😄

tczekajlo: I see 🙈

m-vdb:
awesome 💯 @tczekajlo I will approve the PR and merge it once the `documentation` branch is ready 👍🏻

PR #9733, opened 2021-09-28
Limit pytest jobs to 1 on CI for Windows and test-nlu-predictors

koernerfelicia:
As per discussion [here](https://rasa-hq.slack.com/archives/C01H2UCNP2T/p1632832706102800), I'm limiting the pytest jobs to 1 on CI for problematic configurations (Windows + `test-nlu-predictors`). 

This is a *temporary* fix in order to get TF 2.6 micro out as quickly as possible. I've created a followup issue to investigate and address the memory issues so that the fix can be reverted.

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -294,7 +294,12 @@ jobs:
       # (tests will still run in other envs, they will just not create annotations)
       run: pip install pytest-github-actions-annotate-failures
 
-    - name: Test Code 🔍
+    - name: Test Code 🔍 (monothread)
+      # This is a temporary workaround for OOM on Windows caused by increased memory
+      # consumption associated with the upgrade of our tensorflow dependency
+      # see https://github.com/RasaHQ/rasa/issues/9734 for more information.
+      # Once the memory issues have been addressed, all configurations should be run on
+      # multiple threads
       if: needs.changes.outputs.backend == 'true' && !(matrix.python-version == 3.8 && matrix.os == 'windows-latest') && (matrix.os == 'windows-latest' && matrix.test == 'test-nlu-predictors')
       env:
         JOBS: 1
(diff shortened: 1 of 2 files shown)


.github/workflows/continous-integration.yml
```diff
@@ -295,7 +295,15 @@ jobs:
       run: pip install pytest-github-actions-annotate-failures
 
     - name: Test Code 🔍
-      if: needs.changes.outputs.backend == 'true' && !(matrix.python-version == 3.8 && matrix.os == 'windows-latest')
+      if: needs.changes.outputs.backend == 'true' && !(matrix.python-version == 3.8 && matrix.os == 'windows-latest') && (matrix.os == 'windows-latest' && matrix.test == 'test-nlu-predictors')
+      env:
+        JOBS: 2
```
mprazz, line 300:
```suggestion
        JOBS: 1
```
should be 1 right?

.github/workflows/continous-integration.yml
```diff
@@ -295,7 +295,15 @@ jobs:
...
+        PYTHONIOENCODING: "utf-8"
+      run: |
+        make ${{ matrix.test }}
+
+    - name: Test Code 🔍
+      if: needs.changes.outputs.backend == 'true' && !(matrix.python-version == 3.8 && matrix.os == 'windows-latest') && !(matrix.os == 'windows-latest' && matrix.test == 'test-nlu-predictors')
```
mprazz, line 306:
```suggestion
      if: needs.changes.outputs.backend == 'true' && (matrix.python-version == 3.8 && matrix.os == 'windows-latest') && !(matrix.os == 'windows-latest' && matrix.test == 'test-nlu-predictors')
```
I believe you want the opposite condition here no?

mprazz (commented):
Please rename the steps for clarity, like you said on slack , something like `Test Code (monothread)` vs `Test Code (multithread)` should suffice.

I'd also add a comment on the new step, saying it should be re-unified once memory issues are addressed.

.github/workflows/continous-integration.yml
koernerfelicia: I'm unsure, I was trying to keep the original behaviour: I copied this bit over from the original [step](https://github.com/RasaHQ/rasa/blob/2.8.x/.github/workflows/continous-integration.yml#L298) and put it into both conditions because I thought we don't want this config ever to run? Or is that not the case anymore?

`!(matrix.python-version == 3.8 && matrix.os == 'windows-latest')`

mprazz: oh sorry, I missed that -- seems we're already not running win + py3.8 at all, indeed. Nevermind me, my bad ahah

mprazz (approved):

Issue #9665, opened 2021-09-17
Success markers: read trackers


aeshky:
**Overview of solution:**
Implement a function that loads the tracker files into tracker objects. We already have this loader, so we might just need to call it inside a new markers class. 

Add number of trackers to load. Sorted by ID? Or sampling top N?

See [implementation proposal](https://www.notion.so/rasa/Success-Markers-Implementation-Proposal-79cb6f2b96584387aef6ab5c60db4b4c) for project overview. 

**Definition of done:** 
- [ ] function implemented 
- [ ] function unit tested

PR #9750, opened 2021-09-29
Success markers: read trackers

usc-m:
**Proposed changes**:
- Address #9665 


rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,72 @@
...
+            tracker_store: The underlying tracker store to access.
+            strategy: The strategy to use for selecting trackers, can be 'all', 'sample', or 'first_n'.
+            count: Number of trackers to return, can only be None if strategy is 'all'.
+            seed: Optional seed to set up random number generator, only useful if strategy is 'sample'.
+        """
+        self.tracker_store = tracker_store
+        self.count = count
+
+        if strategy not in MarkerTrackerLoader._STRATEGY_MAP:
+            raise  # Not sure what best exception type to use here is, subclass of RasaException?
```
usc-m, line 50: As it says in the comment - not sure what best exception type to use here is (@ancalita warned against a particular thing to avoid a Sentry deluge and I'm not sure what that was - any advice?)

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self.strategy = MarkerTrackerLoader._STRATEGY_MAP[strategy]
+
+        if not count:
+            if strategy != "all":
+                raise  # Exception
+            else:
+                pass  # Raise warning - redundant parameter
```
usc-m, line 58: Not sure what the best way to surface a warning is here, just log it? Or do we want this to be user visible?

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -0,0 +1,59 @@
+from rasa.core.evaluation.marker_tracker_loader import MarkerTrackerLoader
+from rasa.core.tracker_store import TrackerStore
+import pytest
+
+
+# Assuming a TrackerStore with 5 trackers in it
```
usc-m, line 6: Not sure the best way of setting this up - is it going to be easiest to create a fixture from scratch with a minimal domain or does something already exist?

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,72 @@
...
+        if seed:
+            if strategy == "sample":
+                random.seed(seed)
+            else:
+                pass  # Warning - redundant parameter
+
+    # Not super happy with this - is it sufficient to pull out to optionals or do we unwrap
+    # If we unwrap, do we need to make up the numbers
+    # Do we want a generator version that yields one at a time to reduce memory demands?
+    def load(self) -> List[Optional[DialogueStateTracker]]:
```
usc-m, line 69: So there's a question here about if we want to move this to using a generator - so then we fetch one tracker at a time which constrains memory usage (and if done right maybe gives us some better options for parallelism). Also want to know if we want these unwrapped from the `Optional` when returning, and if so do we want to make up the numbers if some aren't found (they should in theory always be found but we should handle this case if we're unwrapping) or just let the returned number vary from the desired number (what we manage to unwrap is what we get)

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if not count:
+            if strategy != "all":
+                raise  # Exception
+            else:
+                pass  # Raise warning - redundant parameter
+
+        if seed:
+            if strategy == "sample":
+                random.seed(seed)
```
usc-m, line 62: I was thinking about this and I'm not sure this can guarantee us determinism, and I'm not 100% sure we can guarantee it depending on the backing store (basically, if the keys are returned from the backing store without consistent order then deterministically generating indices into that list does nothing) - but this might help a user make it deterministic if they account for that

rasa/core/evaluation/marker_tracker_loader.py
usc-m: I should say I think this should be a fairly trivial change, it's just which way do we prefer? After sleeping on it I think I prefer the generator approach since it's a bit more flexible

rasa/core/evaluation/marker_tracker_loader.py
aeshky: I think we can at least guarantee determinism if we're loading trackers from a folder (do we ever do that, or is it always from a tracker store?). 

I was under the impression that seeds should be set globally rather than inside the class, but can see that `RasaModel` [implements it the way you do](https://github.com/RasaHQ/rasa/blob/22751d7c60b8603c7d052d64060dc8eb5ff9bbab/rasa/utils/tensorflow/models.py#L90), so this will probably work. 

I'm not 100% sure how the seed interacts with the parallelisation you mentioned below. Do you know if it will work?

rasa/core/evaluation/marker_tracker_loader.py
aeshky: >and if so do we want to make up the numbers if some aren't found

Can you clarify what you mean by the numbers? 

I think it's a good idea to parallelise loading the trackers and extracting the markers. We can start with a non-parallelised version for simplicity and add this if we have time, unless it's trivial to do. What do you think?

tests/core/evaluation/test_marker_tracker_loader.py
aeshky: I saw that you asked this question on slack. Makes sense to find existing trackers and use them to create fixtures for your tests.

aeshky (changes requested):

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,72 @@
...
+        count: int = None,
+        seed: Any = None,
+    ) -> None:
+        """Create a MarkerTrackerLoader.
+        
+        Args:
+            tracker_store: The underlying tracker store to access.
+            strategy: The strategy to use for selecting trackers, can be 'all', 'sample', or 'first_n'.
+            count: Number of trackers to return, can only be None if strategy is 'all'.
+            seed: Optional seed to set up random number generator, only useful if strategy is 'sample'.
```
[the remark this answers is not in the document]
usc-m: Annoyingly I did that but it seems that it's not surfacing all the issues locally, going to need to look into this

rasa/core/evaluation/marker_tracker_loader.py
usc-m: As far as I'm aware we have to use a tracker store, and we can't just read files on disk - not seen any native capability to export trackers (except over the event broker which is a different kettle of fish).

That does set a global seed which is why I'm a bit skittish about it (especially when it concerns things that might not happen in a fixed order e.g. parallel execution). It might be slightly better to set the seed immediately before sampling but I think since we're parallelising _after_ sampling it's probably fine? Definitely something we'd need to test if we ever actually added a way of processing them in parallel

rasa/core/evaluation/marker_tracker_loader.py
usc-m: > Can you clarify what you mean by the numbers?
By make up the numbers I mean if we sample 100 Optionals, unwrap them all and we have 10 Nones, we only have 90 samples - do we want to just log that we've lost 10 (implies some weird storage issue, probably easiest) or we could attempt to oversample and then just take 100 (more faff, not guaranteed to get 100 either but probably more chance of it), or do we want to fail out and exit?

So making it a generator should be trivial, parallelising it from the generator I think is straightforward too but would happen elsewhere/be a later addition. I was thinking I could change it now and then the type signature should remain the same (barring any weird things appearing), which would make those modifications down the line hopefully a little easier

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_load_sample(marker_trackerstore: TrackerStore):
+    """Test loading trackers using 'sample' strategy."""
+    loader = MarkerTrackerLoader(marker_trackerstore, "sample", 3)
+    result = list(loader.load())
+
+    assert len(result) == 3
+
+    for item in result:
+        assert marker_trackerstore.exists(item.sender_id)
```
aeshky, line 31: Worth testing invalid sample sizes and also the seed. 
To test the seed, you can vary the items in your tracker, then check that a certain seed gvies the same results. You can hardcode both the seed and the expected resutls.

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,92 @@
...
+    def load(self) -> Iterator[Optional[DialogueStateTracker]]:
+        """Load trackers according to strategy."""
+        stored_keys = list(self.tracker_store.keys())
+        if self.count is not None and self.count > len(stored_keys):
+            rasa.shared.utils.io.raise_warning(
+                "'count' exceeds number of trackers in the store"
+            )
+            self.count = len(stored_keys)
+
+        keys = self.strategy(self.count, stored_keys)
```
aeshky, line 90: I wonder if it's worth swapping the two arguments around (here and the strategies above): 
```
self.strategy(stored_keys, self.count)
```
That way the interface is consistent with the sampling function `random.sample` and also the unused argument `count` in `strategy_all` comes after the main argument. What do you think?

rasa/core/evaluation/marker_tracker_loader.py
aeshky: > That does set a global seed which is why I'm a bit skittish about it (especially when it concerns things that might not happen in a fixed order e.g. parallel execution)

We can look into this a little bit more if we parallelise the execution. I'm sure we can control the behaviour of the seed.

aeshky (changes requested):
Looks good. I requested a few changes 👍

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def test_load_all(marker_trackerstore: TrackerStore):
+    """Test loading trackers using 'all' strategy."""
+    loader = MarkerTrackerLoader(marker_trackerstore, "all")
+    result = list(loader.load())
+
+    assert len(result) == len(marker_trackerstore.keys())
```
aeshky, line 50: Pycharm is complaining about `marker_trackerstore.keys()`: `Expected type 'Sized', got 'Iterable[str]' instead`. Any way to address this warning?

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    def __init__(
+        self,
+        tracker_store: TrackerStore,
+        strategy: str,
+        count: int = None,
+        seed: Any = None,
+    ) -> None:
+        """Create a MarkerTrackerLoader.
```
aeshky, line 40:
```suggestion
        """Creates a MarkerTrackerLoader.
```

aeshky, line 40: I can't find the notion page which says this, but I think this is the tone we prefer for docstrings.

aeshky: I can't find the notion page which says this, but I think this is the tone we prefer for docstrings.

aeshky: Found it! 😄 https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#7670c37267fb433dbafce1cd9c8feade

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def strategy_first_n(count: int, keys: List[Text]) -> Iterable[Text]:
+    """Takes the first N keys from the set of keys."""
+    return keys[:count]
+
+
+def strategy_sample(count: int, keys: List[Text]) -> Iterable[Text]:
+    """Takes a sample of N keys from the set of keys."""
+    return random.choices(keys, k=count)
```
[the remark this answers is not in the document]
usc-m: Don't think we need to - doesn't seem like `k=0` breaks anything, and `k > len(keys)` is impossible due to line 88. I have swapped `choices` for `sample` though, good spot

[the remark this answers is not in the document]
usc-m: I'll add a check for `k < 0` in the constructor since that's a problem for `first_n` too

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -0,0 +1,104 @@
...
+
+    assert len(result) == len(list(marker_trackerstore.keys()))
+
+    for item in result:
+        assert marker_trackerstore.exists(item.sender_id)
+
+
+def test_exception_invalid_strategy(marker_trackerstore: TrackerStore):
+    """Tests an exception is thrown when an invalid strategy is used."""
+    with pytest.raises(RasaException):  # Make this more specific
```
aeshky, line 71:
```suggestion
    with pytest.raises(RasaException): 
```

aeshky: Remove these comments before merging to main.

aeshky (changes requested):

tests/core/evaluation/test_marker_tracker_loader.py
aeshky: I couldn't find an existing execption that better matches this case, so I think `RasaException` is good.

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,102 @@
...
+            if not count:
+                raise RasaException(
+                    "Desired tracker count must be given for strategy '{strategy}'"
+                )
+
+            if count < 0:
+                # Prefer exception here because if count is ever passed less than 0
+                # something has definitely gone wrong, but if count exceeds population size
+                # that might be someone overestimating how much data they have so prefer
+                # to warn for that case and round down
```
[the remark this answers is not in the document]
usc-m: So I thought this comment specifically was useful to provide context for us as developers if we ever revisit this code, on why we chose to warn in some places and error in others. Is there any particular rephrasing you had in mind?

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.strategy = MarkerTrackerLoader._STRATEGY_MAP[strategy]
+
+        if strategy != "all":
+            if not count:
+                raise RasaException(
+                    "Desired tracker count must be given for strategy '{strategy}'"
+                )
+
+            if count < 0:
```
aeshky, line 66:
```suggestion
            if count < 1:
            # At least one tracker required.
```

rasa/core/evaluation/marker_tracker_loader.py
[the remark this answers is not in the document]
aeshky: It's currently written as a "note to self". Happy to suggest a rephrasing if you want to keep it. However, if you implement the check that count is less than the number of trackers, then the code should be fairly self-explanatory.

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,102 @@
...
+                # Prefer exception here because if count is ever passed less than 0
+                # something has definitely gone wrong, but if count exceeds population size
+                # that might be someone overestimating how much data they have so prefer
+                # to warn for that case and round down
+                raise RasaException(
+                    "Parameter 'count' must be greater than or equal to 0"
+                )
+
```
aeshky, line 74:
```suggestion
            if count > len(list(self.tracker_store.keys())):
                # if count exceeds population size, user may have over-estimated
                # the size of their data. Issue a warning and round down.
                rasa.shared.utils.io.raise_warning(
                    "Supplied count exceeds number of available trackers. "
                    "Count rounded down to total number of trackers."
                )
                self.count = len(list(self.tracker_store.keys()))
```

aeshky: Maybe something like this comment, and remove the longer comment starting on line 67?

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,98 @@
...
+                    "Parameter 'seed' is ignored by strategy '{strategy}'"
+                )
+
+    def load(self) -> Iterator[Optional[DialogueStateTracker]]:
+        """Loads trackers according to strategy."""
+        stored_keys = list(self.tracker_store.keys())
+        if self.count is not None and self.count > len(stored_keys):
+            # Warn here as user may have overestimated size of data set
+            rasa.shared.utils.io.raise_warning(
+                "'count' exceeds number of trackers in the store"
```
aeshky, line 92:
```suggestion
                "Parameter 'count' exceeds number of trackers in the store. "
                "All trackers will be processed."
```

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if strategy != "all":
+            if not count:
+                raise RasaException(
+                    "Desired tracker count must be given for strategy '{strategy}'"
+                )
+
+            if count < 1:
+                # If count is ever < 1, user has an error, so issue exception
+                raise RasaException("Parameter 'count' must be set to at least 1")
```
aeshky, line 68:
```suggestion
                raise RasaException("Parameter 'count' must be greater than 0.")
```

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,98 @@
...
+            count: Number of trackers to return, can only be None if strategy is 'all'.
+            seed: Optional seed to set up random number generator,
+                  only useful if strategy is 'sample'.
+        """
+        self.tracker_store = tracker_store
+
+        if strategy not in MarkerTrackerLoader._STRATEGY_MAP:
+            raise RasaException(
+                "Invalid strategy for loading markers - '{strategy}' was given, \
+                options 'all', 'sample', or 'first_n' exist"
```
aeshky, line 55:
```suggestion
                options 'all', 'sample', or 'first_n' exist."
```

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,98 @@
...
+            )
+
+        self.strategy = MarkerTrackerLoader._STRATEGY_MAP[strategy]
+
+        if strategy != "all":
+            if not count:
+                raise RasaException(
+                    "Desired tracker count must be given for strategy '{strategy}'"
```
aeshky, line 63:
```suggestion
                    "Desired tracker count must be given for strategy '{strategy}'."
```

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,98 @@
...
+
+            if count < 1:
+                # If count is ever < 1, user has an error, so issue exception
+                raise RasaException("Parameter 'count' must be set to at least 1")
+
+        self.count = count
+
+        if count and strategy == "all":
+            rasa.shared.utils.io.raise_warning(
+                "Parameter 'count' is ignored by strategy 'all'"
```
aeshky, line 74:
```suggestion
                "Parameter 'count' is ignored by strategy 'all'."
```

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -0,0 +1,98 @@
...
+            )
+            self.count = None
+
+        if seed:
+            if strategy == "sample":
+                random.seed(seed)
+            else:
+                rasa.shared.utils.io.raise_warning(
+                    "Parameter 'seed' is ignored by strategy '{strategy}'"
```
aeshky, line 83:
```suggestion
                    "Parameter 'seed' is ignored by strategy '{strategy}'."
```

aeshky (changes requested):
Looks good. Thanks for addressing the comments. 
I noticed a convention in the codebase of having a full stop at the end of warning and error messages, so I've added these. I also rephrased a couple of warnings, and one comment in `strategy_sample`. Otherwise happy to approve 🚀

aeshky (approved):
I think it's good to explain to the user the consequences of exceeding number of trackers.
```
"Parameter 'count' exceeds number of trackers in the store. "
"All trackers will be processed."
```
Otherwise looks good!

aeshky:
🚀

PR #9756, opened 2021-09-30
Tweak codeclimate exclusion patterns

mprazz:
**Proposed changes**:
- ...


.codeclimate.yml
```diff
@@ -1,6 +1,32 @@
...
+- Makefile
+- NOTICE
+- PRONCIPLES.md
+- README.md
+- bandit.yml
+- binder/
+- changelog/
+- data/         # double-check this
+- docs/         # double-check this
+- examples/     # double-check this
```
joejuzl, line 20: This is an interesting one because it's actual example bots so will have things like custom actions. I would say it's good to check

PR #9785, opened 2021-10-04
Read TF version from poetry.lock

HotThoughts:
**The problem:**

We were using `sed`-style to read TF version from `pyproject.toml`, which leads us to problems when the version constraints aren't cover in the sed command. We need to find a way to read TensorFlow version from `poetry.lock` directly.

**Proposed changes**:

Use python's toml library to parse and read TensorFlow version from `poetry.lock`. 
It now uses the new `read_tenforflow_version.sh` script to retrieve tf version (e.g. [`2.6`](https://github.com/RasaHQ/rasa/pull/9785/checks?check_run_id=3790279267#step:4:17)) and CUDA/cuDNN libraries [successfully load](https://github.com/RasaHQ/rasa/pull/9785/checks?check_run_id=3790295792).

### DIFF
--- a/.github/workflows/ci-model-regression-on-schedule.yml
+++ b/.github/workflows/ci-model-regression-on-schedule.yml
@@ -59,10 +59,9 @@ jobs:
 
       - name: Get TensorFlow version
         run: |-
-          # Read TF version, trim special characters ^~><+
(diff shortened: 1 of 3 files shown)


HotThoughts:
/modeltest
```yml
include:
 - dataset: ["Carbon Bot"]

```

tczekajlo (approved):
LGTM

PR #9787, opened 2021-10-04
fix cache usage in tests

wochinge:
**Proposed changes**:
- fix caching fixture for tests
- fix some test errors

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -743,9 +743,8 @@ def temp_cache_for_fixtures(tmp_path_factory: TempPathFactory) -> None:
         f"cache-{uuid.uuid4()}"
     )
 
-    yield
-
-    LocalTrainingCache._get_cache_location = old_get_cache_location
+    # We can omit reverting the monkeypatch as this fixture is torn down after all the
+    # tests ran
 
 
 @pytest.fixture(autouse=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 import copy
 import os
 import random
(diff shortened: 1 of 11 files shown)


tests/conftest.py
```diff
@@ -731,18 +712,37 @@ def default_execution_context() -> ExecutionContext:
...
+    # cache. If you want to use the cache in a session scoped fixture, then please
+    # consider using the `enable_cache` context manager.
+    old_get_cache_location = LocalTrainingCache._get_cache_location
+    LocalTrainingCache._get_cache_location = lambda: tmp_path_factory.mktemp(
+        f"cache-{uuid.uuid4()}"
+    )
+
+    yield
+
+    LocalTrainingCache._get_cache_location = old_get_cache_location
```
joejuzl, line 728: Won't this happen after all the tests have run?

tests/conftest.py
```diff
@@ -731,18 +712,37 @@ def default_execution_context() -> ExecutionContext:
...
+) -> None:
+    # This fixture makes sure that a single test function has a constant cache
+    # cache.
     cache_dir = tmp_path_factory.mktemp(uuid.uuid4().hex)
-    monkeypatch.setattr(
-        LocalTrainingCache,
-        LocalTrainingCache._get_cache_location.__name__,
-        Mock(return_value=cache_dir),
-    )
+    monkeypatch.setattr(LocalTrainingCache, "_get_cache_location", lambda: cache_dir)
```
joejuzl, line 738: I assume this has no effect if the cache has already been created in a fixture within `enable_cache`?

joejuzl (changes requested):

tests/conftest.py
wochinge: yep - `enable_cache` would just override it. 

With the two fixtures we basically get the best out of two worlds: No global cache for `session_wide` things + a working function scoped cache

joejuzl (approved):

PR #9799, opened 2021-10-06
Include training type in metadata

joejuzl:
**Proposed changes**:
Include training type in metadata

### DIFF
--- a/rasa/engine/training/graph_trainer.py
+++ b/rasa/engine/training/graph_trainer.py
@@ -14,6 +14,7 @@ from rasa.engine.training.components import (
     FingerprintStatus,
 )
 from rasa.engine.training.hooks import TrainingHook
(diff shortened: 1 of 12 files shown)


rasa/engine/training/graph_trainer.py
```diff
@@ -48,6 +49,7 @@ def train(
         output_filename: Path,
         force_retraining: bool = False,
         is_finetuning: bool = False,
+        training_type: TrainingType = TrainingType.BOTH,
```
wochinge, line 52: wondering if we have to many params here - e.g. `is_finetuning` is also related to the `ModelMetadata`. The alternative would be that the `Recipe` returns schemas _and_ a `ModelMetadata` object, what do you think?

wochinge (approved):

PR #9801, opened 2021-10-06
Add slot mapping stats to telemetry

### LINKED ISSUE #9656: 3.0 slot mappings: add telemetry tracking
**Description of Problem**:
Background: This is provided in Taking Apart Forms and Slot Mappings Notion docs.

Summary: The approach of filling slots in 2.0 includes an implicit mechanism (via auto-fill) and an explicit mechanism (via forms). This makes it challenging to keep information up-to-date throughout the course of a conversation, as well as requires slot mapping duplication across forms if multiple forms should fill the same slot.

**Blockers**:

- 

**Definition of Done**:
- [ ] add telemetry tracking, few suggestions can be found in the implementation proposal.

usc-m:
**Proposed changes**:
- Add slot mapping statistics to telemetry (closes #9656)

### DIFF
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -1767,6 +1767,23 @@ class Domain:
 
         return []
 
+    def count_slot_mapping_statistics(self) -> Tuple[int, int]:
+        """Counts the total number of slot mappings and custom slot mappings.
+
+        Returns:
+            A pair of integers where the first entry is the total number of mappings
+            and the second entry is the total number of custom mappings.
+        """
+        total_mappings = 0
+        custom_mappings = 0
+        for slot in self.slots:
+            for mapping in slot.mappings:
+                total_mappings += 1
+                if mapping.get("type") == SlotMapping.CUSTOM:
+                    custom_mappings += 1
+
+        return (total_mappings, custom_mappings)
+
 
 class SlotMapping(Enum):
     """Defines the available slot mappings."""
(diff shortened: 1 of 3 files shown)


rasa/shared/core/domain.py
```diff
@@ -1767,6 +1767,23 @@ def required_slots_for_form(self, form_name: Text) -> List[Text]:
...
+        Returns:
+            A pair of integers where the first entry is the total number of mappings
+            and the second entry is the total number of custom mappings.
+        """
+        total_mappings = 0
+        custom_mappings = 0
+        for slot in self.slots:
+            for mapping in slot.mappings:
+                total_mappings += 1
+                if mapping.get("type") == SlotMapping.CUSTOM:
```
ancalita, line 1782:
```suggestion
                if mapping.get("type") == str(SlotMapping.CUSTOM):
```

rasa/shared/core/domain.py
```diff
@@ -1767,6 +1767,23 @@ def required_slots_for_form(self, form_name: Text) -> List[Text]:
...
+                    custom_mappings += 1
+
```
ancalita, line 1784: I think we should also have a 3rd metric to count the number of slot mappings that include `conditions` key.

ancalita (commented):
Looks good ⚡ 
Left a few more suggestions + please also update the json example in `telemetry.mdx` with the new properties 🙌🏼

Issue #9663, opened 2021-09-17
Success markers: write summary statistics


aeshky:
[todo - description]

joejuzl:
effort: 1

Issue #9662, opened 2021-09-17
Success markers: compute summary statistics


aeshky:
**Overview of solution:**
A function which reads in the output for extract markers and calculates summary statistics.

We will report summary statistics with respect to the number of user turns preceding each marker. This gives user an indication of how long it took before hitting a critical point in the dialog. 
We should report: the size, min, max, arithmetic mean, and median. 

(Any other stats?)

See [implementation proposal](https://www.notion.so/rasa/Success-Markers-Implementation-Proposal-79cb6f2b96584387aef6ab5c60db4b4c) for project overview. 

**Definition of done:** 
- [x] function implemented 
- [x] function unit tested
- [x] ~integrated with [#9660](https://github.com/RasaHQ/rasa/issues/9660).~ Create new issue for integration.

PR #9802, opened 2021-10-06
Success markers: compute and write summary stats

aeshky:
**Proposed changes**:
- implements #9662 and #9663


rasa/core/evaluation/stats.py
```diff
@@ -0,0 +1,93 @@
...
+    data = {"marker_stats": stats, "tracker_stats": per_tracker_stats}
+    with open(path, "w") as outfile:
+        json_str = json.dumps(data, default=np_encoder, indent=2)
+        outfile.write(json_str)
+
+
+def np_encoder(obj: Any) -> Any:
+    """Encodes numpy array values to make them JSON serializable."""
+    if isinstance(obj, np.generic):
+        return obj.item()
```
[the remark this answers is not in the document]
usc-m: I'm not sure what convention is but IMO a reference would be good - I think you're right on the types, since this is going to be called on objects that the JSON encoder isn't able to encode to try and handle them.
It's going to implicitly return `None` for anything that gets passed to it that isn't an instance of `np.generic` which in this case is probably fine (since we're we're not passing anything else in that should trip it), but formally it should raise a `TypeError` if it doesn't know what to do.

tests/core/evaluation/test_stats.py
```diff
@@ -0,0 +1,255 @@
...
+    assert stats == {
+        "n": 0,
+        "mean": np.nan,
+        "median": np.nan,
+        "min": np.nan,
+        "max": np.nan,
+    }
+
+
+def test_compute_single_tracker_stats(extracted_markers_json):
```
usc-m: There's a convention that we should type test parameters too. You'll probably need to add `from typing import Text` (I think that's correct) to pull the type in as well
```suggestion
def test_compute_single_tracker_stats(extracted_markers_json: Text):
```

tests/core/evaluation/test_stats.py
```diff
@@ -0,0 +1,255 @@
...
+            "min": 5,
+            "max": 5,
+        },
+    }
+    extracted_markers = load_extracted_markers_json_file(extracted_markers_json)
+    stats = compute_single_tracker_stats(extracted_markers[0])
+    assert stats == expected_output
+
+
+def test_compute_single_tracker_stats_some_empty(extracted_markers_json):
```
usc-m:
```suggestion
def test_compute_single_tracker_stats_some_empty(extracted_markers_json: Text):
```

tests/core/evaluation/test_stats.py
```diff
@@ -0,0 +1,255 @@
...
+            "min": 7,
+            "max": 7,
+        },
+    }
+    extracted_markers = load_extracted_markers_json_file(extracted_markers_json)
+    stats = compute_single_tracker_stats(extracted_markers[1])
+    assert stats == expected_output
+
+
+def test_compute_multi_tracker_stats(extracted_markers_json):
```
usc-m:
```suggestion
def test_compute_multi_tracker_stats(extracted_markers_json: Text):
```

tests/core/evaluation/test_stats.py
```diff
@@ -0,0 +1,255 @@
...
+            },
+        },
+    }
+    extracted_markers = load_extracted_markers_json_file(extracted_markers_json)
+    stats, per_tracker_stats = compute_multi_tracker_stats(extracted_markers)
+    assert stats == expected_overall_stats
+    assert per_tracker_stats == expected_per_tracker_stats
+
+
+def test_write_stats(extracted_markers_json, marker_stats_output_json):
```
usc-m, line 208:
```suggestion
def test_write_stats(extracted_markers_json: Text, marker_stats_output_json: Text):
```

rasa/core/evaluation/stats.py
```diff
@@ -0,0 +1,94 @@
...
+    If size is == 0 returns np.nan for mean, median.
+
+    Args:
+        data_points: can be a numpy array or a list of numbers.
+    """
+    summary_stats = dict()
+    summary_stats["n"] = np.size(data_points)
+
+    if np.size(data_points) > 0:
+        summary_stats["mean"] = np.mean(data_points)
```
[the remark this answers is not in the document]
aeshky: Cool!

rasa/core/evaluation/stats.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    if np.size(data_points) > 0:
+        stats: Stats = {
+            "n": int(np.size(data_points)),
+            "mean": float(np.mean(data_points)),
+            "median": float(np.median(data_points)),
+            "min": int(np.min(data_points)),
+            "max": int(np.max(data_points)),
+        }
+        return stats
```
aeshky, line 47: @usc-m let me know if I've used the TypedDict correctly.

usc-m: Only thing I'd say is the name is a bit generic and we could end up with namespace collisions or some confusion if this structure is referenced elsewhere

rasa/core/evaluation/stats.py
```diff
@@ -0,0 +1,111 @@
+import os
+from typing import Dict, Text, Union, Any, List, Tuple, TypedDict
+from pathlib import Path
+import json
+import numpy as np
+
+
+class Stats(TypedDict):
+    """A TypedDict for stats."""
```
usc-m, line 9: I'd maybe say "Represents statistics generated over success markers" or something that makes its purpose clearer, since I don't think we're intending for this to be generic

rasa/core/evaluation/stats.py
```diff
@@ -0,0 +1,110 @@
...
+    for marker in single_tracker_markers["markers"]:
+        tracker_stats[marker["marker"]] = compute_summary_stats(
+            marker["num_preceding_user_turns"]
+        )
+    return tracker_stats
+
+
+def compute_multi_tracker_stats(
+    multi_tracker_markers: list,
+) -> Tuple[Dict, Dict[Any, Dict[str, Stats]]]:
```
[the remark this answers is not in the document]
usc-m: Do you have the error it's throwing up? I have some guesses why it might be doing it (and some of them are resolved by just not touching anything), but it would be good to know what it's doing to see if we can resolve it

Issue #9661, opened 2021-09-17
Success markers: write marker file


aeshky:
[todo - description]

Issue #9660, opened 2021-09-17
Success markers: traverse trackers to extract markers


aeshky:
[todo - description]

aeshky:
@alwx I added a requirement to count the number of user turns preceding the marker. This will be used for the summary stats.

Let me know if you have any questions about this so far. @usc-m is [implementing the tracker reader](https://github.com/RasaHQ/rasa/pull/9750), and my [config validation is in review](https://github.com/RasaHQ/rasa/pull/9746), so this should be unblocked very soon!

aeshky:
Hey @usc-m, thanks for picking up this issue while I'm away! I created a draft PR: https://github.com/RasaHQ/rasa/pull/9890

- I implemented `OR` and `AND` but haven't yet implemented `SEQ`. I added a bunch of tests which should help you understand the code and extend it.
- During the [feedback exercise](https://www.notion.so/rasa/Blank-Success-Markers-Input-Feedback-Exercise-25-mins-dfb96b1269324af58d7adb575fb82903), I decided to add a new variable `scope` which determines whether a marker gets checked at every turn or only at the end of the dialogue. This is particularly helpful for negative atomic conditions (e.g., `action_not_executed`, `intent_not_detected`, `slot_not_set`) which would otherwise fire at every turn. Setting the `scope` to dialogue allows us to only check the marker at the end (if that's the desired behaviour).
- Since I didn't manage to finish this issue, one possibility is for you to implement `SEQ` and `scope`. Another option is for me to do it when I return, while you focus on intergrating what's there. A third option is to remove them from the first version of success markes (after consulting with Ty and Maxime). See what you think. If you have time to implement them then go for it!

Here are the remaining TODOs for this issue. Please tick what you manage to finish.
- [ ] implement `SEQ`. Kedz implemented it for `Weizenbot` [here](https://github.com/RasaHQ/weizenbot/pull/28). Might be helpful to look at his implementation. 
- [ ] implement the `scope` variable (and add it to schema). By default, markers are checked at every turn. 
- [ ] `AND` doesn't currently reset. We could potentially reset it as soon as the condition for the marker is satisfied. This allows us to detect if the condition is met multiple times. What do you think?
[...]

PR #9824, opened 2021-10-07
Success Markers CLI

### LINKED ISSUE #9658: Success markers: read markers config

usc-m:
**Proposed changes**:
- Adds a CLI for the Success Markers feature (closes #9658)

### DIFF
--- a/rasa/utils/io.py
+++ b/rasa/utils/io.py
@@ -17,8 +17,8 @@ from typing import (
     Callable,
     TYPE_CHECKING,
     Pattern,
-    Protocol,
 )
+from typing_extensions import Protocol
 
 import rasa.shared.constants
 import rasa.shared.utils.io


usc-m:
So I've reached a bit of an impasse on a few fronts here:
1. I'm not sure if this is the CLI design we want
2. I don't have an interface to go against for writing out marker files or computing them from the trackers that I can go against to complete implementation

I've made this PR a draft to facilitate discussion on these points (@TyDunn I've tagged you because you have opinions on CLI design/might be aware of others who do as well) before I proceed further with implementing them. At the moment the sketch of an implementation I have here is pretty much exactly what we have in the implementation proposal, however, there was some discussion of unifying them into a single command, and also discussion around where these should live.

Here's some ideas I've seen floated or had:

**Current**
This splits `extract` and `stats` into two subcommands under `markers`, `extract` pulls trackers in from a tracker store specified by the endpoint config (which has its argument handled by the existing default argument stuff for that) according to a strategy (`--by` - would need a default if it were optional), up to a number `--num-trackers` (useless if strategy is `all`). If a seed is set that is used to set up the random number generator (useless if strategy is not `sample`).
Optionally, stats can be extracted to a file at this stage.

`stats` takes a file produced by `extract` and produces a stats file.

Unique properties this has (can be interpreted positively or negatively depending on what our priorities are):
- Markers has it's own subcommand, quite a prominent feature on its own
-- Could be useful if we were to expand what could be done with markers
- Extract and stats are separated

Unanswered questions for this one that I can think of:
- Are these the correct optional/positional parameters?
- For the optionals, what should their defaults be?
- Does this model the most common way of using the feature? (i.e. are people always going to want to be working with files or are they looking for quick stats printed to the terminal?)
- Should we also print stats to the terminal?

```
rasa markers extract [--stats <FILENAME>] [--num-trackers <NUM>] [--by <first-n|sample|all>] [--seed <SEED>] [--endpoints <ENDPOINT_CONFIG>] <OUTPUT_FILENAME>

rasa markers stats <INPUT_FILENAME> <OUTPUT_FILENAME>
```

**Variation**
One idea I had was to try and roll both commands into one and sit it under an `evaluate` namespace, as I feel like that's more likely to be useful as a top-level subcommand later. This version assumes we want to print stats by default and moves file exporting to an option for both extraction and summary statistics. Some problems I've encountered with this though are:

- Not fully clear what the default invocation would be
- Still no clarity on default values of parameters
- Not sure if this models the intended workflow better, but feels a bit cleaner to me

```
rasa evaluate markers [--extract <FILENAME>] [--stats-file <FILENAME>] [--endpoints <ENDPOINT_CONFIG>]  [--num-trackers <NUM>] [--by <first-n|sample|all>] [--no-stats]
```

As for the issue of interfaces, we don't yet have a central markers class/data structure which I can use to implement functions to write out/read in markers from disk, or write out stats. This is proving to be a bit of a blocker for this ticket and the other ticket I have at the moment (#9661), since I'm not sure how to access marker data in memory. We have samples which I can derive a schema of the on-disk representation from the PR closing #9662, but I don't want to duplicate effort which would otherwise be done in #9660. If we can define some of these interfaces now that would unblock me and we can somewhat parallelise the work (though the merge order will probably be weird, and it might be difficult to test this PR until #9660 is in).

aeshky (approved):

aeshky:
🎉

Issue #9691, opened 2021-09-21
Fix & improve docs related to story augmentation


samsucik:
### What problem are you trying to solve?

The `augmentation` CLI arg is quite mysterious even though it impacts Core training a lot. We should add more info to the docs, as discussed in #6550.

### What's your suggested solution?

Changes we should make:
- Make it explicit that `augmentation` is a crucial parameter when trying to speed up Core training, but less augmentation can make TED's performance worse
- Correct the docs to state the correct default augmentation factor (they say it's 20 but it's 50).
- Make it explicit that augmentation affects only TED (out of the policies we provide)

Optionally, we might add these recommendations based on our observations:
- If your config contains `TEDPolicy` and no memoization policies, start with story augmentation turned on (e.g. the default value of `50`) 
- Otherwise (if a memoization policy is used too), try out also with story augmentation turned off -- to potentially speed up training

### Definition of Done

- Discussed and decided whether or not to add the recommendations to the docs
- Updated the docs

PR #9825, opened 2021-10-07
Update data augmentation docs.

kedz:
This PR closes this issue #9691 and updates the Data Augmentation section of the policy docs page. Additionally it adds a note to the TEDPolicy section pointing to the Data Augmentation section.

**Proposed changes**:
- Make it explicit that augmentation is a crucial parameter when trying to speed up Core training, but less augmentation can make TED's performance worse
- Correct the docs to state the correct default augmentation factor (they say it's 20 but it's 50).
- Make it explicit that augmentation affects only TED (out of the policies we provide)
- Add a note to TEDPolicy docs point to Data Augmentation

### DIFF
--- a/CHANGELOG.mdx
+++ b/CHANGELOG.mdx
@@ -16,6 +16,14 @@ https://github.com/RasaHQ/rasa/tree/main/changelog/ . -->
 
 <!-- TOWNCRIER -->
 
+## [2.8.11] - 2021-10-20
+### Bugfixes
+- [#9858](https://github.com/rasahq/rasa/issues/9858): Updates dependency on `sanic-jwt` (1.5.0 -> ">=1.6.0, <1.7.0")
+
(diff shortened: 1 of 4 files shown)


kedz:
@samsucik I created this draft PR to discuss the issue since we can see the proposed edits side-by-side with the original. 

Regarding these optional changes:

>  - If your config contains TEDPolicy and no memoization policies,  start with story augmentation turned on (e.g. the default value of 50)
>  - Otherwise (if a memoization policy is used too), try out also with story augmentation turned off -- to potentially speed up training

For the first item, I think we don't need to say this because by default augmentation is on at 50. 
The second point I've folded into the tradeoff between speed and prediction performance in final paragraph:
> We recommend using a memoization based policy along with `TEDPolicy` when reducing the amount of data augmentation to compensate.

WDYT?

Issue #9485, opened 2021-08-31
Change how we name models in 3.0


TyDunn:
What if we changed how we name models in 3.0? So instead of model_<timestamp> we do something like <timestamp>_<randomname> (e.g. https://pypi.org/project/randomname/)? 

The reason for bringing this up is (1) model names today are not easily distinguishable for humans and (2) some customer reasons we discussed in Notion [here](https://www.notion.so/rasa/Rasa-Open-Source-3-0-Architecture-Implementation-Proposal-51ab90b05c41435ca98189a101676a1e#55ad924dfa674dad91cabcf325ef236d)

PR #9904, opened 2021-10-18
Embed random model name

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9485

### DIFF
--- a/docs/docs/components.mdx
+++ b/docs/docs/components.mdx
@@ -2844,7 +2844,103 @@ Also be sure to read the section on the [Component Lifecycle](./tuning-your-mode
 To get started, you can use this skeleton that contains the most important
 methods that you should implement:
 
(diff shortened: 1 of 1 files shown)


wochinge:
@TyDunn This gives a name like the following: `20211018-094821_composite-pita.tar.gz` (you can also play around with it using `rasa train`). Currently our underscore / dash usage seems a bit inconsistent, what do you think? 

When we have a core / NLU only model then we name it like this btw: `nlu-20211018-094821_composite-pita.tar.gz`

wochinge:
@TyDunn I see the motivation behind your rename. I find it helpful though that the `core` prefix relates to the command which was used to train it `rasa train core`. I'm all for getting rid of "core" but if so then I think it should be a coordinated approach across all commands (btw, are you fine with the `nlu-` prefix).

PR #9920, opened 2021-10-19
Add recipe to example bot configs.

### LINKED ISSUE #9886: Update bot examples in repository to use the `DefaultV1.Recipe`
**Overview**
We use the `DefaultV1Recipe` to convert our current config format to the graph architecture. We choose this recipe by default but print a warning if the `recipe` key is not defined in the config file. 

**Definition of Done**
- [ ] update all configs in example projects to have the `recipe` key specified in their config
- [ ] make sure all examples bots work as expected

joejuzl:
Closes #9886.


wochinge:
yep, that would be great. Separate PR or in here?

PR #9935, opened 2021-10-20
improve graph validation error messages

wochinge:
**Proposed changes**:
- improve the error messages when the graph schema validation fails
- add handling of case if the specified parent node doesn't exist in the graph


wochinge:
I didn't adapt warnings for cycles or invalid targets as these should be purely internal

koaning:
I suppose a more general question: when are these validations triggered? Do we validate on calling `rasa train` or `rasa run`? Or can we also trigger it manually from the command line? 

I can imagine that for the novice who is trying to port components it can be very useful to manually ask Rasa to validate a single component without having to parse through all the train logs.

PR #10011, opened 2021-10-29
Slot mappings documentation update

ancalita:
**Proposed changes**:
- [x] Updated documentation on domain, forms, default actions and new CLI command.
- [x] Created migration guide entry.
- [x] double-check if we need to rewrite/ redesign `business-logic.mdx` and `playground.mdx`
- [x] double-check if we need to update any rasa sdk docs (i.e, excerpts in knowledge base slots etc)


docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+
+You can also create a custom validation action extending the `ValidationAction` in the Rasa SDK to either
+define validation methods for slots with predefined mappings or custom extraction and validation methods for slots
+with custom mappings. In either cases, you do not need to specify the `action` key in the mapping, however you will
+need to list `action_validate_slot_mappings` in the domain `actions`.
+You can continue using `FormValidationAction` as in 2.0, however please note this will only run if the form is activated.
+
+Because slot mappings are now relocated from `forms` to `slots`, the `required_slots` of a form has been converted in a
+list of slot names.
+
```
rctatman, line 166: I would add an extra header here to called something like #### Automatic slot migration from 2.0 to 3.0 to make it easier to skim

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+This mechanism replaces the implicit slot setting via auto-fill of slots with entities of the same name.
+The `auto_fill` key in the domain is no longer available, as well as the `auto_fill` parameter in the constructor of
+the `Slot` class.
+
+While forms continue to request the next slot, the extraction of slots is now delegated exclusively to the new default
+action `action_extract_slots`. This action runs in the background after each user turn without it being required to be
+triggered manually.
+
+This approach allows slots to be kept up to date over the course of a conversation, and removes duplicated effort in
```
rctatman, line 142: Paragraph starting on line 142 can probably go in the first paragraph (makes it easier to find out what changed and why if you're only skimming/reading part way)

rctatman (commented):
Added a couple structure-related comments in the migration section.

docs/docs/forms.mdx
```diff
@@ -142,41 +140,12 @@ write stories or rules for this case.
...
-If `intent_name` is `None`, the slot will be filled regardless of intent name.
-Otherwise, the slot will only be filled if the user's intent is `intent_name`.
-
-If `role_name` and/or `group_name` are provided, the role/group
-label of the entity also needs to match the given values. The slot mapping will not
-apply if the intent of the message is `excluded_intent`. Note that you can
-also define lists of intents for the parameters `intent` and `not_intent`.
+:::caution Changed in 3.0
+As of 3.0, [slot mappings](./domain.mdx#slot-mappings) can only be defined in the slots section of the domain.
+Please follow the migration guide to update your assistant.
```
melindaloubser1, line 145:
```suggestion
As of 3.0, [slot mappings](./domain.mdx#slot-mappings) are defined in the `slots` section of the domain.
Please follow the migration guide to update your assistant.
```

Since there wasn't any alternative before it's not a limitation, just a move.

docs/docs/forms.mdx
```diff
@@ -142,41 +140,12 @@ write stories or rules for this case.
...
-      not_intent: excluded_intent
-forms:
-  your_form:
-    required_slots:
-        - slot_name
-```
+#### Unique entity mapping matching
 
 In `from_entity` mapping, when an extracted entity uniquely maps onto a slot,
 the slot will be filled even if this slot wasn't requested by the form.
```
melindaloubser1, line 151: I know you didn't write this, but could you add a short explanation/example of what "uniquely maps" means here?

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+in the `slots` section of the domain file. This approach allows slots to be kept up to date over the course of a conversation, and removes duplicated effort in
+mapping the same slots in multiple forms.
+
+This new mechanism replaces the implicit slot setting via auto-fill of slots with entities of the same name.
+The `auto_fill` key in the domain is no longer available, as well as the `auto_fill` parameter in the constructor of
+the `Slot` class.
+
+While forms continue to request the next slot, the extraction of slots is now delegated exclusively to the new default
+action `action_extract_slots`. This action runs in the background after each user turn without it being required to be
+triggered manually.
```
melindaloubser1, line 141:
```suggestion
While forms continue to request the next slot, slot extraction is now delegated to the default
action `action_extract_slots`. This action runs in the background automatically after each user turn. Like `action_listen`, it should not be included in stories.
```

melindaloubser1: Or some other phrase to indicate that neither needs to nor should ever be run manually

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
 However, the previous behavior can still be enforced by setting
 `ranking_length=10` and `renormalize_confidences=True`.
 
+### Slot Mappings
+
+As of version 3.0, there is a single explicit mechanism for slot filling enabled by defining slot mappings for each slot
+in the `slots` section of the domain file. This approach allows slots to be kept up to date over the course of a conversation, and removes duplicated effort in
+mapping the same slots in multiple forms.
```
melindaloubser1, line 133:
```suggestion
in the `slots` section of the domain file. This approach keeps slots up to date over the course of a conversation, and removes duplicated effort in
mapping the same slots in multiple forms.
```

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+The `auto_fill` key in the domain is no longer available, as well as the `auto_fill` parameter in the constructor of
+the `Slot` class.
+
+While forms continue to request the next slot, the extraction of slots is now delegated exclusively to the new default
+action `action_extract_slots`. This action runs in the background after each user turn without it being required to be
+triggered manually.
+
+You are now required to add the `mappings` key for each slot in the 3.0 domain format, including any slots with custom
+mappings. The keys used to define predefined mappings in 2.0 were maintained. Additionally, you can define slots with
+custom mappings implemented in a custom action, like so:
```
melindaloubser1, line 145: What about slots that really don't have mappings? e.g. they are just there to be filled in the case that a custom action wants to set them. 


```suggestion
Each slot in the `slots` section must include the `mappings` key. The same keys used for predefined mappings in 2.0 are available in 3.0. Additionally, you can define slots with
custom mappings implemented in a custom action, for example:
```

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+    type: bool
+    mappings:
+    - type: custom
+      action: action_verify_customer_status
+```
+
+You can also create a custom validation action extending the `ValidationAction` in the Rasa SDK to either
+define validation methods for slots with predefined mappings or custom extraction and validation methods for slots
+with custom mappings. In either cases, you do not need to specify the `action` key in the mapping, however you will
+need to list `action_validate_slot_mappings` in the domain `actions`.
```
melindaloubser1, line 159: This info should rather go in the doc on slot mappings, and the migration guide can refer to it

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+```yaml
+slots:
+  is_existing_customer:
+    type: bool
+    mappings:
+    - type: custom
+      action: action_verify_customer_status
+```
+
+You can also create a custom validation action extending the `ValidationAction` in the Rasa SDK to either
```
melindaloubser1, line 156: Suggest adding a sub-heading under slot mappings for `slot validation methods` or something similar where this is covered.

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+define validation methods for slots with predefined mappings or custom extraction and validation methods for slots
+with custom mappings. In either cases, you do not need to specify the `action` key in the mapping, however you will
+need to list `action_validate_slot_mappings` in the domain `actions`.
+You can continue using `FormValidationAction` as in 2.0, however please note this will only run if the form is activated.
+
+Because slot mappings are now relocated from `forms` to `slots`, the `required_slots` of a form has been converted in a
+list of slot names.
```
melindaloubser1, line 163: This point is very N.B. from an end user perspective. It should rather go first, then the explanation of where mappings have moved to.

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+
+#### Automatic migration from 2.0 domain format to the 3.0 format
+
+To migrate automatically to the 3.0 domain format, you can run the following command:
```
melindaloubser1, line 167: Mentioned in the glossary too, but a blurb that only the domain file changes in format would be good

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+```bash
+rasa data migrate -d DOMAIN --out OUT_PATH
+```
+
+In addition to creating a a valid 3.0 domain in the indicated out path, this command will automatically backup your
+original domain file(s) in a file labeled `original_domain.yml` or `original_domain` directory if a directory was
+provided instead.
+
+Please note that the migrated slot mappings from forms will include mapping conditions for each form.
+To learn more on mapping conditions, you can read more in this [section](./domain.mdx#mapping-conditions).
```
melindaloubser1, line 178:
```suggestion
To maintain the behavior of forms in the 2.0 format, all migrated slot mappings will include mapping conditions for each form. This can be changed manually according to your use case.
See the docs on [mapping conditions](./domain.mdx#mapping-conditions) for more information.
```

Since all migrated slot mappings will be coming from forms

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+
+#### Manually migrating from 2.0 domain format to the 3.0 format
+
+Let's consider the following 2.0 domain file:
```
melindaloubser1, line 182: First briefly list the steps for migration, then the worked example

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+       intent: deny
+       value: false
+       conditions:
+       - active_loop: restaurant_form
+         requested_slot: outdoor_seating
+```
+
+Please note it is not required to use mapping conditions for custom slot mappings defined in a custom action inheriting
+from `FormValidationAction`. However if you wish certain custom slot mappings to only be set within the context of a form,
+make sure your custom action inherits from `FormValidationAction` rather than from `ValidationAction`.
```
melindaloubser1, line 272: Keep explanation mostly in the main doc and link to it since the point here is domain format, not validation behaviour.

docs/docs/reaching-out-to-user.mdx
```diff
@@ -111,7 +111,7 @@ easier to see which intents are expected to come from external devices when work
 ### 4. Update the Domain
 
 To tell the assistant which plant needs watering, you can define an entity that you'll post along with the intent.
-To be able to use the entity value directly in a response, define a slot by the same name:
+To be able to use the entity value directly in a response, define a `from_entity` slot mappings for the `plant` slot:
```
melindaloubser1, line 114:
```suggestion
To be able to use the entity value directly in a response, define a `from_entity` slot mapping for the `plant` slot:
```

melindaloubser1 (changes requested):
What a boatload of work @ancalita  😅 Nice one.

I left a lot of comments, mostly about restructuring paragraphs. I realized in retrospect that many of the paragraphs in question probably came directly from the previous docs and were just moved. 
 I think the most important is to clarify:
* custom slot mappings: requirements vs. possibilities
* validation actions: This should get its own section at the same level as `Custom Slot Mappings`. Questions to address under it:
  * briefly what a validation action is for 
  * when a validation action will be run (inside a form vs. globally)
  *  how to define a validation action (inheritance etc.)

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
 
 ### Slot Auto-fill
 
-The slot will be set automatically if your NLU model picks up an entity, and your domain
- contains a slot with the same name, provided the following conditions are met:
- 
-1. `store_entities_as_slots` is set to true
-2. the slot's `auto_fill` property is set to true
+:::caution Removed in 3.0
```
m-vdb, line 473: We don't have a clear guideline for this at the moment. The only guideline we have now is that we remove callouts when we remove deprecated features / behaviours in order to start fresh in a new major and reducing the amount of callouts in a doc page (i.e. it doesn't scale otherwise). I would assume the same for this case (even if nothing was deprecated before).

How about:
1. remove this section altogether as it doesn't apply to 3.x
2. in the callout that announces the new feature (i.e. ```:::info New in 3.0```), use some of the text that you have here to briefly explain the behaviour change and link at the migration guide? (like @melindaloubser1  suggested)

?

docs/docs/forms.mdx
m-vdb: @ancalita how about a short sentence that says that now it's deterministic, as opposed to automatic/guessing in 2.x + link to migration guide?

docs/docs/domain.mdx
m-vdb: I would add such callout in [the next section below](https://github.com/RasaHQ/rasa/pull/10011/files#diff-cb86102a1484b02650a25eceae48c41c59f43802ab70061e6dd1d120327cd0e6R483-R484), as well as in `docs/docs/forms.mdx`. I think it's good to add duplicated callouts for the same feature wherever we need them (i.e. across different docs pages), because it's more explicit to the user!

🤔 I think I should add this as a documentation guideline, WDYT?

melindaloubser1:
@ancalita Sorry I just thought of this now while reviewing what else in the doc needs to change: I think the right place for describing how to implement custom slot mappings is really under the `Actions` heading - there should be another page under here called `Slot Extraction Actions` or something similar. It can then explain about which classes are valid base classes, and the limitations on what should be returned by this kind of action. The `custom slot mappings` section in the domain page then only has to describe how to reference a custom slot mapping.

image

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
...
 
+Slot Auto-fill functionality was removed in 3.0 and replaced with a new explicit mechanism of
+setting slots after every user message.
+You will need to explicitly define slot mappings for each slot in the `slots` section of `domain.yml`.
+In addition, stories must include the `slot_was_set` step.
+Please read through the migration guide to update your assistant.
```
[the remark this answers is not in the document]
ancalita: Actually L478 (on `stories must include the slot_was_set step`) won't be a requirement until a later minor, considering this long thread [here](https://rasa-hq.slack.com/archives/C01HHMR4X8S/p1635861549213800). I will remove it and add the link to migration guide as you suggested.

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
...
 
-For example:
+:::
+
+### Slot Mappings
+
+Rasa Open Source comes with four predefined mappings to fill the slots of a form
+based on the latest user message.
```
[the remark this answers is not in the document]
ancalita: Good catch 😄

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
...
 ```
 
+#### from_intent
 
-:::note Auto-filled slots & `influence_conversation`
-An auto-filled slot that is defined with `influence_conversation: true` will influence the conversation the same way any other slot does.
+The `from_intent` mapping will fill slot `slot_name` with value `my_value` if
+user intent is `intent_name` or `None`. The slot mapping will not
+apply if the intent of the message is `excluded_intent`. Note that you can
+also define lists of intents for the parameters `intent` and `not_intent`.
```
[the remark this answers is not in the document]
ancalita: Hmm, I am not the original author of this section, but based on the (translated) code in [here](https://github.com/RasaHQ/rasa/blob/1ab2908e9e8dca8cf140220f3cbd9c4fb0acbc6e/rasa/shared/core/slot_mappings.py#L103), `intent` can be `None` if it's missed and that gets translated to an empty list of `mapping_intents` which later means `intent_not_blocked` is `True` as long as the `not_intent` doesn't contain the latest user message intent. In a nutshell, it means that if the `intent` parameter is not used, then the slot can be filled with the `value` for any given intent as long as it's not listed as `not_intent`. I can rewrite this as you suggested.

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
...
 
-:::
+The `from_trigger_intent` mapping will fill slot `slot_name` with value `my_value`
+if a form is activated by a user message with intent `intent_name`.
+The slot mapping will not apply if the intent of the message is
+`excluded_intent`. Note that you can
+also define lists of intents for the parameters `intent` and `not_intent`.
+
+A slot with `from_trigger_intent` mapping must always include mapping conditions
+and be listed in the form's `required_slots`.
```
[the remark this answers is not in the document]
ancalita: I'm actually going to change my mind about the first half of the sentence after rechecking the code, doesn't need to include mapping conditions, only needs to be in the form's required slots. This can also be checked with `rasa data validate`, should I link to that info instead?

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
...
+      conditions:
+      - active_loop: your_form
+        requested_slot: slot_name
+      - active_loop: another_form
+```
+
+:::note
+If conditions are not included in a slot mapping, the slot is eligible to be set universally
+outside the context of any forms. However if a slot is also listed in a form's `required_slots`,
+the slot can also be requested by the form if unset at that moment.
```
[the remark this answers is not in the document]
ancalita: I wanted to emphasise the behaviour without `conditions` so that there are no surprises if this key is omitted. And also to point out that forms cannot update slots, they can only set slots if they are unset at that point in the conversation.

docs/docs/domain.mdx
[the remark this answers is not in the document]
melindaloubser1: Yes, that makes sense!

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
...
+:::
+
+### Custom Slot Mappings
+
+A custom slot mapping is useful when none of the predefined mappings fit your use case.
+For example, if you use custom policies, custom actions or the custom validation action to extract and validate slots,
+you will have to define the slot mapping in the domain as `custom`.
```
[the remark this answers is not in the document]
ancalita: >  not sure how a policy could be setting slots using this method?

I don't know either, but I heard feedback from some Rasa superheroes that some custom components they use can set slots too.

> it reads as: if the user already uses custom slot mappings, then they need to specify custom slot mappings.

Mhm, yes, I think I was subconsciously writing it as a migration guide for an audience of previous Rasa developers vs. new developers. I'll rephrase as suggested.

docs/docs/domain.mdx
```diff
@@ -465,64 +470,180 @@ stories:
...
+
+In order to write your own extraction and / or validation code, we recommend you to create a custom action inheriting
+from the `ValidationAction` interface from the Rasa SDK and defining the `extract_<slot_name>` and / or
+`validate_<slot_name>` methods. The name of the action is already implemented and
+should not be overridden: `action_validate_slot_mappings`. Please make sure to add this action name to the list of
+actions in the domain.
```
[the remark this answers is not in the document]
ancalita: It's only an option, the alternative is the custom action example. I'll attempt to rephrase this more clearly.

docs/docs/migration-guide.mdx
ancalita: Every slot that gets listed in the domain needs to have the `mappings` parameter defined, as well as everytime a `Slot` class gets instantiated (although I don't have real-life examples for the latter part). There is a little hack I used in the case of default slots internally, (e.g, `session_started_metadata`) meaning I assigned an empty list to mappings argument 😄 But that's not something I advise for domain slots. Thinking about this now, I should actually add a test case for this, when a domain slot has no mapping in `mappings`. I assume it would never get set when `action_extract_slots` runs.

docs/docs/domain.mdx
ancalita: Sounds good, thank you 👍🏼

docs/docs/forms.mdx
```diff
@@ -142,44 +140,16 @@ write stories or rules for this case.
...
-      not_intent: excluded_intent
-forms:
-  your_form:
-    required_slots:
-        - slot_name
-```
-
-In `from_entity` mapping, when an extracted entity uniquely maps onto a slot,
-the slot will be filled even if this slot wasn't requested by the form.
+In `from_entity` mapping, when the attributes of an extracted entity (i.e, entity name, role, group) uniquely map
```
melindaloubser1, line 151:
```suggestion
In the `from_entity` slot mapping, when the attributes of an extracted entity (i.e, entity name, role, group) uniquely map
```

docs/docs/forms.mdx
```diff
@@ -142,44 +140,16 @@ write stories or rules for this case.
...
+onto a slot, the slot will be filled even if this slot wasn't requested by the form.
```
melindaloubser1, line 152: isn't this the case for all slot mappings? They will apply even if the slot was not requested excepting conditions 🤔
This concept was kinda confusing in 2.x too, but now it's more so because slot filling happens somewhere else.

Is this only applicable when `store_entities_as_slots` is on? I think I need to understand this behaviour before suggesting anything 😅

docs/docs/forms.mdx
melindaloubser1: Also: What is "uniquely" - the slot has only this mapping from this entity, or the entity is only referenced by one slot mapping?

docs/docs/forms.mdx
```diff
@@ -446,13 +346,19 @@ to extract customs slots:
 1. Define a method `extract_<slot_name>` for every slot that should be mapped in a
    custom way.
 2. Make sure that in the domain file you list for your form only those slots that use
```
melindaloubser1, line 348:
```suggestion
2. In your domain file, for your form's `requested_slots`, list only those slots that use
```

Is this still true? Can it not list slots with custom mappings under `required_slots` now?

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+As of version 3.0, there is a single explicit mechanism for slot filling enabled by defining slot mappings for each slot
+in the `slots` section of the domain file. This approach keeps slots up to date over the course of a conversation, and
+removes duplicated effort in mapping the same slots in multiple forms.
+
+This new mechanism replaces the implicit slot setting via auto-fill of slots with entities of the same name.
+The `auto_fill` key in the domain is no longer available, as well as the `auto_fill` parameter in the constructor of
+the `Slot` class.
+
+While forms continue to request the next slot, slot extraction is now delegated to the default
+action `action_extract_slots`. This action runs in the background automatically after each user turn.
```
melindaloubser1, line 140: link to default action docs
```suggestion
action [`action_extract_slots`](link). This action runs in the background automatically after each user turn.
```

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+slots:
+  is_existing_customer:
+    type: bool
+    mappings:
+    - type: custom
+      action: action_verify_customer_status
+```
+
+You can also use [slot validation actions](./slot-validation-actions.mdx) to define custom validation of slots with
+predefined mappings or custom extraction and validation of slots with custom mappings.
```
melindaloubser1, line 156:
```suggestion
You can also use [slot validation actions](./slot-validation-actions.mdx) to either validate slots with predefined mappings, or to both extract and validate slots with custom mappings.
```
If I'm reading it right, the point is that a slot validation action can be used for only validation, BUT only for slots with predefined mappings. To validate a slot with a custom mapping, both mapping and validation must sit in the same slot validation action. Suggestion is based on that assumption

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+
+:::note
+The `required_slots` of a form has been converted in a list of slot names because slot mappings are now relocated
+from `forms` to `slots` in the domain.
```
melindaloubser1, line 160:
```suggestion
The `required_slots` of a form used to be a list of slot mappings.
Since slot mappings are relocated to the `slots` section of the domain, `required_slots` has been converted to a list of slot names only.```

File: docs/docs/migration-guide.mdx [hunk 30]
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+       value: true
+     - type: from_intent
+       intent: deny
+       value: false
+ forms:
+   restaurant_form:
+     required_slots:
+	 - cuisine
+     - num_people
+     - outdoor_seating
```
melindaloubser1, line 247:
```suggestion
     required_slots:
     - cuisine
     - num_people
     - outdoor_seating
```

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+         outdoor_seating:
+           - type: from_intent
+             intent: affirm
+             value: true
+           - type: from_intent
+             intent: deny
+             value: false
+```
+
+The initial result of this migration could look like this:
```
melindaloubser1, line 216:
```suggestion
The initial result of migrating this domain to 3.0 format would look like this:
```

docs/docs/migration-guide.mdx
```diff
@@ -126,6 +126,152 @@ confidences. This change can affect the maximum confidence predicted by the
...
+       value: false
+ forms:
+   restaurant_form:
+     required_slots:
+	 - cuisine
+     - num_people
+     - outdoor_seating
+```
+
+In case you'd prefer certain slots to be filled only within the context of a form, you can add mapping conditions like so:
```
melindaloubser1, line 250:
```suggestion
For slots that should be filled only in the context of a form, add mapping conditions to specify which form(s) should be active. Adding `conditions` is required preserve the behavior of slot mappings from 2.0, since without them the mappings will be applied on each user turn regardless of whether a form is active or not.
```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
+---
+id: slot-validation-actions
+sidebar_label: Slot Validation Actions
+title: Slot Validation Actions
+abstract: A slot validation action is a special type of custom action, designed to handle custom extraction and/or validation of slot values. This can be used to validate slots with predefined mappings or extract slots with custom mappings.
+---
+
+Rasa SDK provides three base classes that you can use to inherit from in order to define your custom slot extraction and
+validation code:
```
melindaloubser1, line 9:
```suggestion
Rasa SDK provides three base classes that you can inherit from in order to implement custom slot extraction and
validation logic:
```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+## `Action` class
+
+You can use this class to build a [custom action](./custom-actions.mdx) that is limited to returning only
+`SlotSet` and `BotUttered` events. Any other event type will be filtered out by the default action
```
melindaloubser1, line 14:
```suggestion
You can use this class to create a regular [custom action](./custom-actions.mdx). The base class will not prevent you from returning arbitrary events, however a slot validation action should only return
`SlotSet` and `BotUttered` events. Any other event type will be filtered out by the default action
```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
...
+[`action_extract_slots`](./default-actions.mdx#action_extract_slots).
+The name of the custom action will have to be specified in the `action` key of the
```
melindaloubser1, line 16:
```suggestion
The name of the custom action must be specified in the `action` key of the relevant
```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
...
+[custom slot mapping](./domain.mdx#custom-slot-mappings) in the domain.
+Note that the action name must be listed in the domain `actions` section too.
+
+## `ValidationAction` class
+
+You can extend the `ValidationAction` in the Rasa SDK to define custom extraction and / or validation of slots that
```
melindaloubser1, line 22:
```suggestion
You can extend the `ValidationAction` class in the Rasa SDK to define custom extraction and / or validation of slots that
```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
...
+can be set or updated outside of a form context.
+With this option, you do not need to specify the `action` key in the [custom slot mapping](./domain.mdx#custom-slot-mappings).
+
+Note that you do not need to implement the `name` method in the custom action extending `ValidationAction`, since this
+is already provided: `action_validate_slot_mappings`. However this name must be added to the domain `actions` list.
```
melindaloubser1, line 27: What happens if a user overrides the name? It never gets called right? If so worth mentioning here

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+:::note
+Slots mapped to a form in the mapping `conditions` will be ignored by `ValidationAction` unless you override
+the `required_slots` method to include the slots you want to extract and / or validate.
+:::
+
+### Validation of Slots with Predefined Mappings
+
+You must write functions named `validate_<slot_name>` for every predefined slot that you need validated.
```
melindaloubser1, line 36:
```suggestion
To validate slots with a predefined mapping, you must write functions named `validate_<slot_name>`.
```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
...
+            return {"location": slot_value.capitalize()}
+        else:
+            # validation failed, set this slot to None
+            return {"location": None}
+```
+
+### Extraction of Custom Slot Mappings
+
+In order to define custom extraction code, you must write an `extract_<slot_name>` method for every slot with a
+custom slot mapping.
```
melindaloubser1, line 68: What happens if the user specifies a custom action in the `action` key of a custom slot mapping, and then provides a validation method here? Does that work?

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,101 @@
...
+        current_count_of_insults = tracker.get_slot("count_of_insults")
+        if intent_of_last_user_message == "insult":
+           current_count_of_insults += 1
+
+        return {"count_of_insults": current_count_of_insults}
+```
+
+## `FormValidationAction` class
+
+If you prefer that certain custom slot mappings should only be set within the context of a form,
```
melindaloubser1, line 95: Bit confused by this condition phrasing because `conditions` already allow you to limit applicability of any mapping to a form. 
Will all validation actions run regardless of conditions? If so, this could read:
```suggestion
If certain custom slot mappings should only be validated within the context of a form,
```

melindaloubser1 (changes requested):
Some phrasing suggestions and a few questions that might need phrasing changes, but I don't know what to suggest yet 😅

docs/docs/default-actions.mdx
```diff
@@ -230,3 +228,28 @@ takes its presence into account so that [conversation behaviour is customizable]
...
 
 :::
+
+## action_extract_slots
+
+This action runs after each user turn, before the next assistant action prediction and execution.
+`action_extract_slots` loops through the [slot mappings](./domain.mdx#slot-mappings) of each domain slot in order to set or update
+slots throughout the conversation with information extracted from the latest user message.
+
+If `action_extract_slots` finds a [custom slot mapping](./domain.mdx#custom-slot-mappings), it will check first if a custom action was defined in the
```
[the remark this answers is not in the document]
ancalita: Should I replace with `initially`? If there's no custom action defined, an empty list is returned (or in other words no SlotSet events are added to the tracker because there is no custom action to execute). Then the default action just moves on to check if it needs to run a custom validation action which is explained in the next paragraph.

docs/docs/forms.mdx
ancalita: The slot can have multiple mappings, and some can be shared by other slots. But if the slot contains a mapping that is not found in other slots' mappings, then we consider it unique.  Or the entity can be referenced in multiple mappings, but if the role/group is only specific to a slot, then the mapping is unique to that slot.

docs/docs/forms.mdx
ancalita: This is actually separate from auto-fill (which required `store_entities_as_slots`). 

To maintain regression of this unique entity mapping feature in forms, the new slot mappings maintains it during an active_loop with slot mappings that have conditions. So it will not extract the slot if the user utterance contained an entity which is mapped onto a non-unique `from_entity` mapping and the slot was not requested by the form.

docs/docs/forms.mdx
ancalita: Both ways should work actually but decided to keep the old docs in since it's not necessary to migrate 2.x `FormValidationAction` to the new way - I've made a note to test this today actually to make sure both cases work and let you know.

docs/docs/slot-validation-actions.mdx
ancalita: The mapping `conditions` actually have no effect if a custom extraction was defined in an action extending `ValidationAction` instead of `FormValidationAction`. That's why I tried to emphasise that if there are any custom slot mappings meant to be both extracted and validated only within a form, better use `FormValidationAction`.

m-vdb:
@ancalita do you want another review from Melinda here mostly?

ancalita:
> @ancalita do you want another review from Melinda here mostly?

@m-vdb Yup, although not ready yet to re-request review, testing some things she asked about.
It would be great if you could check if the info notes (`New in` , `Changed in `) make sense.

docs/docs/forms.mdx
ancalita: Yes, you can list slots with custom mappings under form `required_slots`, meaning you don't have to override the `required_slots` anymore, and it actually gives the extra benefit that you can use `utter_ask_<custom_slot>` for these slots now. However, you can continue implementing the 2.x way currently described in the docs too. How should we write that both ways work?

docs/docs/slot-validation-actions.mdx
ancalita: Yes, that's correct, the custom validation action will never run, I'll add a comment.

docs/docs/slot-validation-actions.mdx
ancalita: Yup, this _should_ work, however there's a reported bug blocking this from working fully (by that I mean that the ValidationAction fails to run for the extracted slot value by the custom action from the `action` key), but should be fixed before the release 🤞🏼

docs/docs/migration-guide.mdx
ancalita: To follow-up on this, if a slot doesn't specify mappings (via the domain hack of `mappings: []`), that slot never gets set via a form or `action_extract_slots`.

docs/docs/domain.mdx
```diff
@@ -463,66 +468,177 @@ stories:
...
-The slot will be set automatically if your NLU model picks up an entity, and your domain
- contains a slot with the same name, provided the following conditions are met:
- 
-1. `store_entities_as_slots` is set to true
-2. the slot's `auto_fill` property is set to true
+:::info New in 3.0
 
+As of 3.0, slot mappings are defined in the `slots` section of the domain.
+This change removes the implicit mechanism of setting slots via auto-fill which is replaced with a new explicit
+mechanism of setting slots after every user message.
```
m-vdb, line 477:
```suggestion
This change removes the implicit mechanism of setting slots via auto-fill and replaces it with a new explicit
mechanism of setting slots after every user message.
```

m-vdb (approved):
LGTM (I've only reviewed the callouts)

docs/docs/command-line-interface.mdx
```diff
@@ -355,6 +356,45 @@ You can specify the input file or directory, output file or directory, and the o
...
+If no arguments are specified, the default domain path (`domain.yml`) will be used for both input and output files.
+
+This command will also back-up your 2.0 domain file(s) into a different `original_domain.yml` file or
+directory labeled `original_domain`.
+
+Note that the slots in the migrated domain will contain [mapping conditions](./domain.mdx#mapping-conditions) if these
+slots are part of a form's `required_slots`.
+
+:::caution
+Exceptions will be raised and the migration process terminated if invalid domain files are provided, slots or forms
```
samsucik, line 385: As per #10243, I'd change this to:
```suggestion
Exceptions will be raised and the migration process terminated if invalid domain files are provided or if they are already in the 3.0 format, if slots or forms
```

docs/docs/default-actions.mdx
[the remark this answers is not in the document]
melindaloubser1: Gotcha, reading it again now it makes more sense so I think it's fine this way

docs/docs/domain.mdx
```diff
@@ -463,66 +468,177 @@ stories:
...
+You can define custom slot mappings using [slot validation actions](./slot-validation-actions.mdx) when none of the
+predefined mappings fit your use case. You must define this slot mapping to be of type `custom`, for example:
+
+```yaml
+slots:
+  day_of_week:
+    type: text
+    mappings:
+    - type: custom
+      action: action_calculate_day_of_week
```
melindaloubser1, line 641:
```suggestion

You can define custom slot mappings using [slot validation actions](./slot-validation-actions.mdx) when none of the
predefined mappings fit your use case. You must define this slot mapping to be of type `custom`, for example:

```yaml
slots:
  day_of_week:
    type: text
    mappings:
    - type: custom
      action: action_calculate_day_of_week
\```

You can also use the `custom` slot mapping to list slots that will be filled by arbitrary custom actions in the course of a conversation, by listing the type and no specific. For example:

```yaml
slots:
  handoff_completed:
    type: boolean
    mappings:
    - type: custom
\```

This slot will not be updated on every user turn, but only once a custom action that returns a `SlotSet` event for it is predicted. 

```

Can't get the backticks to play along; Is this true though? To maintain the previous behaviour of a slot that is created without being filled on each turn user's will need to know this

docs/docs/domain.mdx
```diff
@@ -616,20 +732,17 @@ call, or to start the conversation with a bot message. The docs on
...
 The `config` key in the domain file maintains the `store_entities_as_slots` parameter.
-When an entity is recognized by the NLU model and the entity name matches a slot name,
-`store_entities_as_slots` defines whether the entity value should be placed in that slot.
-By default, entities will auto-fill slots of the same name.
+When an entity is recognized by the NLU model and it matches the `from_entity` slot mapping, `store_entities_as_slots`
+defines whether the entity value should be placed in that slot.
+By default, this behaviour is switched on.
 
-You can turn off all slot auto-filling by setting the `store_entities_as_slots` parameter to `false`:
+You can turn off this functionality by setting the `store_entities_as_slots` parameter to `false`:
```
melindaloubser1, line 739: with `auto_fill` the turning off made sense, but in the new scenario, what's the point of controlling this here? Doesn't it implicitly nullify the `from_entity` mappings, so one could just as well delete them?
I'm sure there's a good reason, just think we should tell users what it is

docs/docs/forms.mdx
melindaloubser1: Let me see if I've got this right:
1. The definition of "unique entity mapping":

There is only one slot mapping in the entire domain which is `from_entity`  for a given (entity + attributes)  combo . 

2. Applicability of `from_entity` mappings when a form is active

`from_entity` mappings will get applied while a form is active iff either 
(1) the form requested the slot OR
(2) There is only one slot in the entire domain with a slot mapping for the specific (entity +attributes) combo listed.
 
Is this correct?

melindaloubser1: I remember the point of this now (also based on example below), which was to prevent filling a slot with an entity intended to fill only the requested slot e.g. given a form that collects both `amount` (from `number` entity) and `account_number` (also from `number` entity), the form shouldn't go filling both slots with the first `number` entity extracted, it should leave the not-yet-requested one empty and wait for the user's response to the specific prompt.

If my understanding above is correct, I'll draft a short explanation of what "unique" means, because I'm sure someone besides me is gonna wonder about it

docs/docs/forms.mdx
melindaloubser1: Great that both ways do work! 
Since this is not the migration guide, my instinct is it should rather push users to the most sensible usage in 3.0, which is listing all required slots, custom and otherwise, in the domain, and only indicate the possibility of overwriting the method in the custom action when actual dynamic behaviour is required i.e. in the `dynamic form behaviour` below.

docs/docs/migration-guide.mdx
```diff
@@ -1254,7 +1254,157 @@ consult the following migration guides for the individual policies:
...
 components were deprecated in Rasa 2.x and have been removed in Rasa 3.0. See the
-[migration guide for Rasa 2.x](#deprecations-2) for replacing these components in your pipeline.  
+[migration guide for Rasa 2.x](#deprecations-2) for replacing these components in your pipeline.
+
+### Slot Mappings
+
+As of version 3.0, there is a single explicit mechanism for slot filling enabled by defining slot mappings for each slot
+in the `slots` section of the domain file. This approach keeps slots up to date over the course of a conversation, and
+removes duplicated effort in mapping the same slots in multiple forms.
+
```
melindaloubser1, line 1264: Similar to the comment above about `custom` mappings being allowed to be empty: The phrasing here could make a user think they can't define "empty" slots that are filled only at specific points in the conversation anymore. But they can - they can just do more than that. 
```suggestion
As of version 3.0, there is a single explicit mechanism for slot filling enabled by defining slot mappings for each slot
in the `slots` section of the domain file. This approach keeps slots up to date over the course of a conversation, and
removes duplicated effort in mapping the same slots in multiple forms. It is still possible to fill slots from arbitrary custom actions and not update them on every turn of the conversation if that behavior is desired.

```

docs/docs/migration-guide.mdx
```diff
@@ -1254,7 +1254,157 @@ consult the following migration guides for the individual policies:
...
+slots:
+  is_existing_customer:
+    type: bool
+    mappings:
+    - type: custom
+      action: action_verify_customer_status
+```
+
+You can also use [slot validation actions](./slot-validation-actions.mdx) to either validate slots with predefined
+mappings, or to both extract and validate slots with custom mappings.
```
melindaloubser1, line 1286:
```suggestion
available in 3.0. Additionally, you can define slots with custom mappings implemented in a custom action which will be run on every user turn, for example:

```yaml
slots:
  is_existing_customer:
    type: bool
    mappings:
    - type: custom
      action: action_verify_customer_status
\```

You can use [slot validation actions](./slot-validation-actions.mdx) to either validate slots with predefined
mappings, or to both extract and validate slots with custom mappings.

Slots which will be filled by arbitrary custom actions in the course of the conversation, and which should not be updated on every user turn, should be listed with mappings of type `custom` and no action. For example:


```yaml
slots:
  handoff_completed:
    type: bool
    initial_value: false
    mappings:
    - type: custom
\```

 This slots value will only change when a custom action is predicted that sets it. This mapping maintains the behavior from 2.x for a slot which was not filled by an entity or by slot mappings in a form.

```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## `ValidationAction` class
+
+You can extend the `ValidationAction` class in the Rasa SDK to define custom extraction and / or validation of slots that
+can be set or updated outside of a form context.
+With this option, you do not need to specify the `action` key in the [custom slot mapping](./domain.mdx#custom-slot-mappings).
+
+Note that you do not need to implement the `name` method in the custom action extending `ValidationAction`, since this
+is already provided: `action_validate_slot_mappings`. However this name must be added to the domain `actions` list.
+If you override the `name` method, the custom validation action will not run because the original name has been hardcoded
```
melindaloubser1, line 28:
```suggestion
If you override the `name` method, the custom validation action will not run because the original name is hardcoded
```

docs/docs/slot-validation-actions.mdx
melindaloubser1: Got it, are we sure enough about that to update this section accordingly? Else that bug should get an additional "Change docs when you fix this" note

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## `FormValidationAction` class
+
+If certain custom slot mappings should only be extracted and / or validated within the context of a form,
+make sure your custom action inherits from `FormValidationAction` rather than from `ValidationAction`.
+It is not required to use mapping conditions for custom slot mappings defined in a custom action extending
+`FormValidationAction`.
+Note that this custom action will only run if the form is activated.
+
+To learn more about how to implement this class, see Forms [Advanced Usage](./forms.mdx#advanced-usage).
```
melindaloubser1, line 103:
```suggestion
A `FormValidationAction` custom action will only run if the form it specifies in its name is activated.
If certain custom slot mappings should only be extracted and / or validated within the context of a form,
the custom action should inherit from `FormValidationAction` rather than from `ValidationAction`.
Since `FormValidationAction` runs on every user turn only as long as the form is active, it is not required to use mapping conditions for custom slot mappings defined in a custom action extending
`FormValidationAction`.

To learn more about how to implement this class, see Forms [Advanced Usage](./forms.mdx#advanced-usage).
```

melindaloubser1 (changes requested):
We are getting there!
A good test of the docs from an end user's take would be that a QAer of RC3 who hasn't been looking at them already refers to them when migrating a bot's slots & forms. 
We want to merge these before then, just noting it for now.

docs/docs/slot-validation-actions.mdx
ancalita: I actually fixed the bug that I thought was causing the problem yesterday, but the problem persists so I created this issue in `rasa-sdk` repo: https://github.com/RasaHQ/rasa-sdk/issues/593
To be earnest, I don't know if this will be prioritised before the release - mainly because I think it should not be a recommended route of extraction and validation (I think it's better design for `actions.py` to have extraction and validation for one slot in one place, either the custom action or the `ValidationAction`; if the custom action is supposed to return a `SlotSet` event it should have the validated value already). I can add the note to change docs to its definition of done.

docs/docs/forms.mdx
ancalita: > Applicability of from_entity mappings when a form is active

There is a nuance here when the form is active:
- either a slot is requested -> any of its mappings (unique or non-unique) become eligible for filling that slot.
- a slot was not requested but the user utterance has an entity that could fill that slot -> that entity can fill that slot only according to a unique entity mapping; otherwise a non-unique entity mapping won't apply and the slot remains unfilled.

docs/docs/domain.mdx
ancalita: Yup, that's correct, I'll add that 👍🏼

docs/docs/domain.mdx
ancalita: Yes, you're right, this doesn't make sense that much anymore; i only kept it because this attribute (`store_entities_as_slots`) conditions the running of a Domain method (`slots_for_entities()`) that gets called in `StoryStep.explicit_events`. So it helps with a story such as the one debated in a thread a while ago:
```
- story: example story
  steps:
  - or:
     - intent: greet
     - intent: greet
        entities:
        - name: Molly
 - action: utter_greet
```
Probably an idea to mention that if this parameter gets turned off the stories won't work accordingly 🤔

docs/docs/forms.mdx
melindaloubser1: Right, I think we're on the same page now 😅 Let me make a phrasing suggestion then.

docs/docs/domain.mdx
melindaloubser1: Ah right, I remember that thread now. 

Not sure this is clarified yet though: Does `store_entities_as_slots: false` == removing all `from_entity` mappings?
Or, is there some way the `from_entity` mappings get applied, even when `store_entities_as_slots: false`?

 Does this parameter control how the _syntax_ in stories is interpreted at training time, and is this any different from how user messages are interpreted & used to set slots at inference time?
e.g.
if `store_entities_as_slots: true`, then this story and the following one are equivalent:
```
- story: example story
  steps:
  - or:
     - intent: greet
     - intent: greet
        entities:
        - name: Molly
 - action: utter_greet
```
==
```
- story: example story
  steps:
  - or:
     - intent: greet
     - intent: greet
        entities:
        - name: Molly
     - slot_was_set:
        - name: Molly
 - action: utter_greet
```

If `store_entities_as_slots: false`, then the first story is rather equivalent to this other one:

```
- story: example story
  steps:
  - or:
     - intent: greet
     - intent: greet
        entities:
        - name: Molly
     - slot_was_set:
        - name: null
 - action: utter_greet
```

melindaloubser1: An example of the impact of this parameter is definitely needed here, I've ran into A Lot of user confusion on this behaviour.

ancalita: > Does store_entities_as_slots: false == removing all from_entity mappings?
> Or, is there some way the from_entity mappings get applied, even when store_entities_as_slots: false?

No, this will not affect how `from_entity` slots are filled during the conversation (meaning the default action will carry on running just fine).

> Does this parameter control how the syntax in stories is interpreted at training time, and is this any different from how user messages are interpreted & used to set slots at inference time?

Yes, this parameter will only control to how stories are interpreted at training time. It doesn't influence how slots are set at inference time after every user turn.

I don't know for sure if that exact story syntax you listed for `store_entities_as_slots: false` works, but  it's in the right direction: if this parameter gets turned off, then the tracker won't have a `SlotSet` event for the entity in the user message listed in the story.

melindaloubser1: Ok! Then that's what I think we should explain in this section,  that this parameter is only relevant to reading stories and turning them into trackers, and making slot set events out or entities.

docs/docs/forms.mdx
ancalita: Hmm, it's intended to allow other slots that are not in that form's required_slots list to be filled, if those slots don't have mapping with conditions (we assumed that any slot without conditions is meant to be filled by any user utterance outside forms too).

> it will still fill slots required by other forms with an extracted entity even if a form is active and the mapping is not unique across all slots in the domain.

Let's say you have slot_2 that should only be set by form_2, slot_2 should have its mapping with conditions to prevent this behaviour you mention in form_1 that doesn't require slot_2.

Maybe something else I need to clarify in the docs in `mapping conditions` section?

melindaloubser1: I think this is the right place to explain it, will include in update I'm pushing

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,103 @@
...
+The name of the custom action must be specified in the `action` key of the relevant
+[custom slot mapping](./domain.mdx#custom-slot-mappings) in the domain.
+Note that the action name must be listed in the domain `actions` section too.
+
+## `ValidationAction` class
+
+You can extend the `ValidationAction` class in the Rasa SDK to define custom extraction and / or validation of slots that
+can be set or updated outside of a form context.
+With this option, you do not need to specify the `action` key in the [custom slot mapping](./domain.mdx#custom-slot-mappings).
+
```
melindaloubser1, line 25: There needs to be a N.B. note here that says, if you use this class to to do custom extraction, mapping conditions are ignored, in addition this suggestion since this will run even if a form is active:

```suggestion
You can extend the `ValidationAction` class in the Rasa SDK to define custom extraction and / or validation of slots that
can be set or updated whether a form is active or not.
With this option, you do not need to specify the `action` key in the [custom slot mapping](./domain.mdx#custom-slot-mappings), since the default action `action_extract_slots` runs `action_validate_slot_mappings` automatically. 

```

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,103 @@
...
+Note that you do not need to implement the `name` method in the custom action extending `ValidationAction`, since this
+is already provided: `action_validate_slot_mappings`. However this name must be added to the domain `actions` list.
+If you override the `name` method, the custom validation action will not run because the original name is hardcoded
+in the call that the default action `action_extract_slots` makes to the action server.
+
```
melindaloubser1, line 30: What happens with non-rasa-sdk action servers here? Can the users implement their own `action_validate_slot_mappings` ?

melindaloubser1: This pertains to the last point in the PR about updating action server docs - Some of this might need to be moved over there entirely, so that people with non-sdk action servers also can refer to and understand how to use custom actions for slot mappings.

ancalita: > What happens with non-rasa-sdk action servers here?

Completely forgot this is also possible 🙈 I guess they'll have to implement their own `ValidationAction` mirroring the Python sdk. Should I then open a new issue for action server docs update? What kind of issues need to be updated, I made a note about knowledge base slots code that might need to be updated with  mapping of `type: custom`?

melindaloubser1: >Completely forgot this is also possible

And I completely neglected to consider it in this review up till now, so equally guilty 🙈

Yes, let's make a new issue for sdk docs updates. 
That issue can include 
- deciding how much of the slot validation action docs should move to the action server docs
- describing the behaviour of rasa-sdk's `ValidationAction` in broad strokes to allow other language implementations of it.

docs/docs/slot-validation-actions.mdx
ancalita: Yup, will apply this suggestion. Do you mean bringing the existing N.B note further up after the first sentence?

docs/docs/slot-validation-actions.mdx
ancalita: In our inbox: https://github.com/RasaHQ/rasa-sdk/issues/597

docs/docs/slot-validation-actions.mdx
melindaloubser1: Ja, bring it up and phrase it stronger (I can't find the comment I may or may not have left on it sorry!)

ancalita: Done!

docs/docs/slot-validation-actions.mdx
```diff
@@ -0,0 +1,107 @@
...
+## `ValidationAction` class
+
+You can extend the `ValidationAction` class in the Rasa SDK to define custom extraction and / or validation of slots that
+can be set or updated outside of a form context.
+
+:::note
+`ValidationAction` will ignore any slots mapped to a form added to the slot mapping's `conditions`.
+Please extend the [`FormValidationAction`](./slot-validation-actions.mdx#formvalidationaction-class) class for applying
+custom slot mappings only within the context of a form.
+:::
```
melindaloubser1, line 29:
```suggestion
:::note
`ValidationAction` is intended for extracting slots **outside**  the context of a form. It will ignore extraction and validation methods for any slots that have a form specified under the slot mapping's `conditions`. It will not run these methods when the specified form is active nor when no form is active.
Please extend the [`FormValidationAction`](./slot-validation-actions.mdx#formvalidationaction-class) class for applying
custom slot mappings only within the context of a form.
:::
```

melindaloubser1 (approved):
One suggestion to clean up a warning note, otherwise let's get this merged and create follow up issues for any usability issues that come out of the next RC 🚀

docs/docs/slot-validation-actions.mdx
ancalita: Whoop whoop so excited about merging this, yipeee 😄

docs/docs/migration-guide.mdx
melindaloubser1: No idea why I only saw this now, but cool! What is the difference between doing `mappings: []` and `mappings: - type: custom` without any custom action to extract the slot?

There should indeed be a test case for empty mappings.

ancalita: There's no difference really, other than that `type: custom` is an explicit way of indicating this slot can be set in other ways than the predefined ways (could be custom component or validation actions etc).

PR #10012, opened 2021-10-29
markers / validation

### LINKED ISSUE #9964: Success Markers: validate config
**Definition of Done**
- [ ] Add checks to all `from_config` functions
- [ ] Add check for marker name uniqueness
- [ ] Add unit tests
- [ ] Add checks against domain -- also needs to be added to CLI
- [ ] handle TODO from [this commit](https://github.com/RasaHQ/rasa/pull/10003/commits/8557f1d8e679392930321ae10fec48ad45258416)

ka-bu:
**Proposed changes**:
- closes #9964

**Overview**
- "atomic/compound" vs "condition/operator"
    * renamed since I guess we will call it the latter in docs and the code was only half-way consistent anyway 😅 
- tag vs name:
    * often, "marker_name" was used to describe a "tag" that must be of a predefined set, while there is also a "name" parameter that can actually set to something arbitrary --> renamed "marker_name" to "tag" where appropriate
    * since "tag" can now be a positive or negative one --> renamed "tag()" to "positive_tag()" since it returns just that
- expected number of sub-markers:
    * I've split the not marker into it's own class since otherwise the expected number of sub-markers would've depended on whether the marker was negated or not .... which I found a bit too confusing
    * tests are adapted accordingly already  + a test is added to check whether we raise an error if a certain number of markers is expected but one more is passed 
 - config_dict vs from_path
    * the "from_config_dict" is gone because we would not have been able to raise errors that point you to the right file where you have to search for the error -> two step approach now that first reads the yamls, checks the top level (custom marker names) and retains the links to the files
 -  config format:
     * no listing of text parameters under a single condition tag anymore
     * no implicit and marker under a top-level custom marker name anymore

**TODO**
- [ ] tests and hopefully not many fixes :D 
- [ ] check against domain


tests/core/evaluation/test_marker.py
```diff
@@ -517,14 +513,145 @@ def test_markers_cli_results_save_correctly(tmp_path: Text):
         assert len(senders) == 5
 
 
-def test_all_operators_in_schema():
-    operators_in_schema = rasa.shared.utils.schemas.markers.OPERATOR_SCHEMA["enum"]
-    operators_in_schema = {tag.lower() for tag in operators_in_schema}
+def collect_slots(marker: Marker) -> Set[Text]:
```
ka-bu, line 516: we can merge these 3 functions into one :)

rasa/cli/evaluate.py
```diff
@@ -128,8 +133,16 @@ def _run_markers(
             "A file with the stats filename already exists"
         )
 
-    tracker_loader = _create_tracker_loader(endpoint_config, strategy, count, seed)
+    domain = Domain.load(domain_path)
     markers = Marker.from_path(config)
+
+    if not markers.validate_against_domain(domain):
```
ka-bu, line 139: nice idea to just "bubble up" the warnings and then decide to break it here to gather all warnings first

usc-m: I was hoping to do it more like a compiler where you can track line number and also collect warnings a bit more elegantly, but it seemed like adding that would involve a lot of complicated stuff around adding information about where markers were parsed from to the markers and I didn't want to do too much in the time we have left

rasa/core/evaluation/marker.py
```diff
@@ -131,12 +149,27 @@ def negated_tag() -> Optional[Text]:
...
+
+        Args:
+            domain: The domain to check against
+        """
+        valid = self.text in domain.action_names_or_texts
+
+        if not valid:
+            rasa.shared.utils.io.raise_warning(
+                f"Referenced action '{self.text}' does not exist in the domain"
+            )
```
aeshky, line 163: @usc-m and @ka-bu I was creating the QA tests and was curious why this (and the same for intent and slot) is a warning and not an error. Does this mean that the extraction process will continue if there is a mismatch with the domain? What's the advantage of allowing this?

usc-m: The extraction process isn't managed here - this goes over each marker and decides if they are valid, and if not emits a message, and then passes back True/False if it's valid or not. In the CLI we catch that boolean and error and exit based on it, but in these I wanted a way to collect _all_ of the issues without stopping first, and then stop later (since reporting one at a time would be frustrating for users). The reason it's a warning is mostly because that was my first pass on it. Ideally it would go out as an error without terminating (and then get caught in the CLI to wrap it up and actually terminate) but we don't have a helper function for that so it's a little more involved - definitely something I want to fix before this merges.

usc-m: The CLI also terminates before the extraction process begins. It loads the marker definitions and the domain, does the validation first, and terminates if issues are found. It will only establish a connection to the tracker store (and by extension perform extraction) if there are no issues

ka-bu: or in short: https://github.com/RasaHQ/rasa/pull/10012#discussion_r740836126 :D

usc-m: Looked at how DialogueStateTracker manages this and have replicated

ka-bu: I like that it's consistent but that approach is also a bit confusing, isn't it? I would've expected a "raise..." if there's a logger error (and usually, I would not explicitly log the error but just have a decorator to pass on the error message)

ka-bu: ... on the other hand, getting an error as a result of "only warnings" is also a bit unexpected maybe 😅

usc-m: Yeah, I figured it was consistent from the perspective of the user, but it might be worth re-evaluating how we do error management across the codebase at some point

usc-m (approved):

Issue #8974, opened 2021-06-28
Update Intent ID Type from Integer to String


b-quachtran:
**Description of Problem**:

Intent ID's are currently stored as integers, making it possible in certain cases for the integers to be cast as scientific notation and precision lost as a result. Instead, ID's should be stored as a string to avoid this issue. 

**Overview of the Solution**:

Update Rasa NLU schema and classifier implementations so intent ID's are stored as strings. 

**Definition of Done:**
- Rasa NLU schema is updated
- Classifier implementations are updated
- All the required docs changes are also done

PR #10013, opened 2021-10-29
Remove intent's syntetic ID construct and use name

tayfun:
Fixes https://github.com/RasaHQ/rasa/issues/8974

Intent IDs were basically hashed from intent's labels (aka names?) and
were in number format. If we change this to a string to make sure
scientific notation does not lose precision, we essentially can use the
label in place. Intent labels are unique and much more human readable.
Does the job in a simpler way.

**Proposed changes**:
- Use intent name instead of artificially creating an intent ID by hashing its name.


tayfun:
I don't think a changelog is needed here because it is not a user facing change.

joejuzl (changes requested):
I think it makes sense having a changelog entry as the data that is sent to the action server for custom actions will now be different. And probably worth explaining that they can get the equivalent value by hashing the name too.

Also can you check we don't still have the id in any test events in unit tests?  e.g.  `tests/core/test_actions.py`

tayfun:
> I think it makes sense having a changelog entry as the data that is sent to the action server for custom actions will now be different. And probably worth explaining that they can get the equivalent value by hashing the name too.
> 
> Also can you check we don't still have the id in any test events in unit tests? e.g. `tests/core/test_actions.py`

Thanks, good catch about other tests with ID in it :) Somehow it is not easy to grep them all. I have fixed all I've found and added the changelog now.

tttthomasssss (commented):
LGTM in general 🤖!

One small thing, wasn't the intent `id` exposed to users via the Events API in the rasa-sdk (i.e. inside `UserUttered`)? And if so, this could potentially cause unexpected crashes for any user who - for one reason or another - relies on the intent id for some processing, right?

tayfun:
> LGTM in general 🤖!
> 
> One small thing, wasn't the intent `id` exposed to users via the Events API in the rasa-sdk (i.e. inside `UserUttered`)? And if so, this could potentially cause unexpected crashes for any user who - for one reason or another - relies on the intent id for some processing, right?

@tttthomasssss , yes you're essentially right. This will be a backwards incompatible change for version 3. Rasa X may have to change some stuff including this change and as you have said, other users which rely on Intent ID too. 

Recommendation is to use intent name and if they can't, they can simply hash the name to get same number as before.

changelog/8974.bugfix.md
```diff
@@ -0,0 +1,5 @@
+Intent IDs sent with events (to kafka and elsewhere) have been removed, intent
+names can be used instead (or if numerical values are needed for backwards
+compatibility, one can also hash the names to get previous ID values, ie.
+`hash(intent_name)` is the old ID values). Intent IDs have been removed because
+they are too big and can be turned to scientific notation causing confusion.
```
joejuzl, line 5: "Intent IDs have been removed because as they were providing no extra value and integers that large were problematic for some event broker implementations." ?

joejuzl (changes requested):

joejuzl (approved):
Looks good! 🚀

tttthomasssss (approved):

PR #10055, opened 2021-11-02
Remove `Domain.slots_for_entities` usage in `DialogueStateTracker.update()`

### LINKED ISSUE #9744: 3.0 slot mappings: revisit `Domain.slots_for_entities()` method and its usage
**Overview of the Problem**:
- In rasa 2.0, `Domain.slots_for_entities()` method used to return a list of `SlotSet` events for slots with the same entity name which had auto-fill enabled. 
- In 3.0, auto-fill has been removed and the method was adapted slightly to set any slots if matching entities are extracted. 
- This method is being used in 2 places: `DialogueStateTracker.update()` and `StoryStep.explicit_events()`:

1. `DialogueStateTracker.update()` has 74 usages in the codebase (including tests) that might need to be checked/adapted if this method is removed. Alternatively, if the method should be maintained, the implementation of `DialogueStateTracker.update()` should be checked if we want a certain section concerning `UserUttered` events to be kept.
This might be a duplicate of what the default `action_extract_slots` now does.

- A side effect of this method is that although it generally allows for slots with `FROM_ENTITY` mappings to be updated throughout the conversation, it is not owed to running the `action_extract_slots` default action, and therefore confusingly not logged.

**Definition of Done**:
- [x] investigate the usage ramifications of this method to determine if it makes sense to remove or adapt/ refactor now that the default action_extract_slots runs after every user turn
- [x] investigate adaptation of `DialogueStateTracker.update()` and complete if feasible
- [x] create new issue for deprecating the usage of entities in stories (in `StoryStep.explicit_events`) and carrying exploratory work that the replacement story format with `slot_was_set` maintains consistent behaviour
- [x] handle `EntitiesAdded` case: If an action can also add entities, then we also need to run the `ActionExtractSlots` after these actions
- [x] adapt any unit tests

ancalita:
**Proposed changes**:
- Closes #9744 


rasa/core/processor.py
```diff
@@ -742,6 +745,9 @@ async def _run_prediction_loop(
                     self.on_circuit_break(tracker, output_channel, self.nlg)
                 break
 
+            if prediction.is_end_to_end_prediction:
+                tracker = await self._run_action_extract_slots(output_channel, tracker)
```
wochinge, line 749: can we please add a log statement to make it clear that we're running this a second time?

rasa/core/processor.py
```diff
@@ -742,6 +745,9 @@ async def _run_prediction_loop(
                     self.on_circuit_break(tracker, output_channel, self.nlg)
                 break
 
+            if prediction.is_end_to_end_prediction:
```
wochinge, line 748: do we have a test which makes sure that we run this / to avoid regression?

wochinge (commented):
💯 

The title is slightly weird as we're keeping `Domain.slots_for_entities` for now, isn't it?

rasa/core/processor.py
ancalita: You're right, I need to add a test for this part: I'm amending the `e2e_bot` domain file to contain slots, hope that's fine 🤞🏼 The test I've added passes, we'll see what the CI says 😅

rasa/core/processor.py
ancalita: Should I also add a changelog entry for this - or do you think the global slot mappings initial changelog encapsulates this too?

rasa/core/processor.py
ancalita: @wochinge Adding slots to the `e2e_bot` domain causes this test `test_e2e_with_entity_evaluation` to [fail](https://github.com/RasaHQ/rasa/runs/4167179517?check_suite_focus=true), because `report` variable is empty. Do you think it's better for me to create a different e2e fixture?

wochinge: I'd rather investigate why the report is empty to be honest.

rasa/core/processor.py
wochinge: As we haven't released yet the initial changelog encapsulates this 👍🏻

rasa/core/processor.py
ancalita: Ok, I deep dived in the area related to testing stories and this is what i found:

1. the result of [`evaluate_entities`](https://github.com/RasaHQ/rasa/blob/4620db7607d61baa050d9779b1c1264568ccef61/rasa/core/test.py#L1119) is an empty dict `{ }` for `report` [here](https://github.com/RasaHQ/rasa/blob/4620db7607d61baa050d9779b1c1264568ccef61/rasa/nlu/test.py#L882) because this [if statement](https://github.com/RasaHQ/rasa/blob/4620db7607d61baa050d9779b1c1264568ccef61/rasa/model_testing.py#L359) in `model_testing.py` gets executed.

2. Looking more into this: `entity_results` output [here](https://github.com/RasaHQ/rasa/blob/4620db7607d61baa050d9779b1c1264568ccef61/rasa/core/test.py#L1079) is an empty list -> because [`tracker_entity_results`](https://github.com/RasaHQ/rasa/blob/4620db7607d61baa050d9779b1c1264568ccef61/rasa/core/test.py#L955) is empty list -> because every `entity_result` [here](https://github.com/RasaHQ/rasa/blob/4620db7607d61baa050d9779b1c1264568ccef61/rasa/core/test.py#L838) is predicted as `None`.  

And this is because [this block](https://github.com/RasaHQ/rasa/blob/4620db7607d61baa050d9779b1c1264568ccef61/rasa/core/test.py#L604) never gets executed and I'm not sure why?

ancalita: I've looked at what's in `tracker.events` and based on the fact that a `SlotSet` event gets added now to the tracker after every `UserUttered`, that block in `_get_e2e_entity_evaluation_result` never gets executed because every last event will be a `SlotSet`, should I modify this method to actually get the penultimate event if last event is a `SlotSet`, test passes when I check that 🤔 :

```
['action_listen', 'UserUttered(text: my name is joe, intent: None, entities: joe (Type: name, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: name, value: joe)']
['action_listen', 'UserUttered(text: my name is joe, intent: None, entities: joe (Type: name, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: name, value: joe)', 'hi joe!']
['action_listen', 'UserUttered(text: my name is joe, intent: None, entities: joe (Type: name, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: name, value: joe)', 'hi joe!', 'action_listen', 'UserUttered(text: I am feeling happy, intent: None, entities: happy (Type: mood, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: mood, value: happy)']
['action_listen', 'UserUttered(text: my name is joe, intent: None, entities: joe (Type: name, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: name, value: joe)', 'hi joe!', 'action_listen', 'UserUttered(text: I am feeling happy, intent: None, entities: happy (Type: mood, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: mood, value: happy)', "I'm glad"]
['action_listen', 'UserUttered(text: today I was very cranky, intent: None, entities: cranky (Type: mood, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: mood, value: cranky)']
['action_listen', 'UserUttered(text: today I was very cranky, intent: None, entities: cranky (Type: mood, Role: None, Group: None), use_text_for_featurization: True)', 'SlotSet(key: mood, value: cranky)', "I'm sorry!"]
```

ancalita: @wochinge that seemed to have fixed the test 😁 Let me know what you think, ready for a new round of review.

rasa/core/test.py
```diff
@@ -601,6 +601,12 @@ async def _get_e2e_entity_evaluation_result(
     prediction: PolicyPrediction,
 ) -> Optional[EntityEvaluationResult]:
     previous_event = tracker.events[-1]
+
+    if isinstance(previous_event, SlotSet):
+        # UserUttered events with entities can be followed by SlotSet events
+        # if slots are defined in the domain
+        previous_event = tracker.get_last_event_for(UserUttered)
```
wochinge, line 608: If there was an ActionExecuted in between then this condition should be wrong (only if there was nothing between user uttered and the current action except slots and bot uttered stuff)
```suggestion
    if isinstance(previous_event, SlotSet):
        # UserUttered events with entities can be followed by SlotSet events
        # if slots are defined in the domain
        previous_event = tracker.get_last_event_for((UserUttered, ActionExecuted))
        if isinstance(previous_event, UserUttered):
             ...
```

alwx (approved):
Looks good!

PR #10065, opened 2021-11-02
Success markers telemetry

### LINKED ISSUE #9830: Success makers: telemetry
**Definition of done**
- [ ] telemetry decided
- [ ] telemetry implemented

alwx:
**Proposed changes**:
- fix #9830


alwx:
This PR adds events specified in this comment: https://github.com/RasaHQ/rasa/issues/9830#issuecomment-946483085

docs/docs/telemetry/events.json
```diff
@@ -422,6 +423,63 @@
...
+          "description": "Number of processed trackers."
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "trackers_count"
+      ]
+    },
+    "Markers Stats Computed": {
+      "description": "Triggered when marker stats has been computed.",
```
usc-m, line 470:
```suggestion
      "description": "Triggered when marker stats have been computed.",
```

docs/docs/telemetry/events.json
```diff
@@ -422,6 +423,63 @@
...
+          "description": "Number of trackers to extract from (for any strategy except 'all')."
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "count"
+      ]
+    },
+    "Markers Extracted": {
+      "description": "Triggered when markers has been extracted.",
```
usc-m, line 455:
```suggestion
      "description": "Triggered when markers have been extracted.",
```

rasa/cli/evaluate.py
```diff
@@ -123,6 +124,10 @@ def _run_markers(
         stats_file: (Optional) Path to write out statistics about the extracted
                     markers.
     """
+    telemetry.track_markers_evaluation_initiated(
```
usc-m, line 127: @ka-bu if this merges before your stats/wrap-up one we'll need to change this

usc-m (commented):
Not 100% sure if we should be referring to as "Markers Evaluation" - maybe "Evaluation - Success Markers" (@aeshky @ka-bu any opinions?). Otherwise looks good though!

aeshky (changes requested):
Thanks for the PR @alwx. 
Are we not tracking the number of markers defined? (I can see that it's not in the final list [here](https://github.com/RasaHQ/rasa/issues/9830#issuecomment-946483085)). If not, why did we decide to drop it? It's informative because it tells us whether users have 1-2 KPIs or are abusing the feature by defining dozens of things to track.

aeshky:
@usc-m sorry I didn't read your comment until after I finished my review:
> Not 100% sure if we should be referring to as "Markers Evaluation" - maybe "Evaluation - Success Markers"

I made a similar observation but suggested a different name. I like your suggestion: "Evaluation - Success Markers"

docs/docs/telemetry/events.json
```diff
@@ -422,6 +423,63 @@
...
+        },
+        "count": {
+          "type": ["string", "null"],
+          "description": "Number of trackers to extract from (for any strategy except 'all')."
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "count"
+      ]
```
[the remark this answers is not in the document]
alwx: Yep, that's a mistake, fixed it in the latest commit.

rasa/cli/evaluate.py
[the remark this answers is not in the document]
alwx: the event is now called "Markers Extraction Initiated" (based on one of your suggestions) so I renamed this function to have a similar name

alwx:
Re-requesting your review because I changed all the names and fixed the issue with `required` in `events.json`. 

> Are we not tracking the number of markers defined?

I think we can do this if it makes sense. When does this even needs to be tracked?

aeshky:
> > Are we not tracking the number of markers defined?

> I think we can do this if it makes sense. When does this event need to be tracked? 

What I mean is the number of markers in the config file. So as soon as the config file is processed we would know the number of markers. I can dig into the code to see where this happens. 

@usc-m anything else you want to track? You [suggested](https://github.com/RasaHQ/rasa/issues/9830#issuecomment-938680028) a measure of marker complexity. It's a good idea, but maybe not as trivial to implement so maybe we can add it later. What do you think?

usc-m:
> What I mean is the number of markers in the config file. So as soon as the config file is processed we would know the number of markers. I can dig into the code to see where this happens.

Number of markers in the config file should be a matter of counting the number of sub-markers of the marker returned from `Marker.from_config` in the CLI code I believe. (so `len(markers.sub_markers)` in `_run_markers` in rasa\cli\evaluate.py)

> @usc-m anything else you want to track? You suggested a measure of marker complexity. It's a good idea, but maybe not as trivial to implement so maybe we can add it later. What do you think?

Not strictly necessary but something like maximum depth of nested markers, or how wide they get (maximum number of sub-markers under a marker that isn't top-level). Don't think these would be hard to add but also aren't as necessary - I think we'd get some useful info out about how the feature is used but it's something we could add later perhaps?

aeshky:
> Not strictly necessary but something like maximum depth of nested markers, or how wide they get (maximum number of sub-markers under a marker that isn't top-level). Don't think these would be hard to add but also aren't as necessary

It tells us how complex the conditions get, and thus how people are using (or potentially abusing?) Markers. 

> I think we'd get some useful info out about how the feature is used but it's something we could add later perhaps?

Definitely useful but not urgent. Let's just count how many Markers are defined for now 👍

docs/docs/telemetry/events.json
```diff
@@ -422,6 +423,66 @@
...
+          "type": ["string", "null"],
+          "description": "Number of trackers to extract from (for any strategy except 'all')."
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "strategy",
+        "only_extract",
+        "seed",
+        "count"
```
aeshky, line 454: @usc-m are all of these required? 🤔 or just the top two?

docs/docs/telemetry/events.json
```diff
@@ -422,6 +423,66 @@
...
+          "description": "Number of processed trackers."
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "trackers_count"
+      ]
+    },
+    "Markers Stats Computed": {
+      "description": "Triggered when marker stats has been computed.",
```
aeshky, line 473:
```suggestion
      "description": "Triggered when marker statistics have been computed.",
```

aeshky: Let's use the full word in text, and reseve "stats" for code :)

docs/docs/telemetry/events.json
usc-m: From the perspective of telemetry I think it makes sense - we'd want to know if it's being used or not. The types of those are `string` or `null` so if it's not present we'd see an explicit null. Would be good to check with someone who understands the telemetry schema here better (though I think this is only used in the docs to explain what data we send back to users and isn't used to actually validate anything internally)

usc-m:
OK, I think I've added the extra telemetry. Is there any other changes we need to make? Anything more to collect, names all fine etc.?

rasa/telemetry.py
```diff
@@ -1005,3 +1009,64 @@ def track_nlu_model_test(test_data: "TrainingData") -> None:
...
+        seed: (Optional) The seed used if strategy is 'sample' and the user selects
+              one
+        count: (Optional) The number of trackers the user is trying to select.
+    """
+    _track(
+        TELEMETRY_MARKERS_EXTRACTION_INITIATED_EVENT,
+        {
+            "strategy": strategy,
+            "only_extract": only_extract,
+            "seed": seed,
```
usc-m, line 1033: This will report the actual seed used by the user - do we want the actual seed or do we want to just know if they used a seed? Is this something that's worth changing or are we considering the actual seed value not important enough to be careful about not collecting?

aeshky: We only want to know if the user set a seed. I don't think it counts as private info so maybe it doesn't matter if we collect it. But if we can do true/false that might be better.

rasa/cli/evaluate.py
```diff
@@ -139,6 +147,10 @@ def _run_markers(
             "Please see errors listed above and fix before running again."
         )
 
+    # Subtract one to remove the virtual OR over all markers
+    num_markers = len(markers) - 1
```
ka-bu, line 151: this is the total number of all conditions and operators used in all marker configurations that were evaluated -- to get the number of user-defined markers we should
1. get rid of the special case [here](https://github.com/RasaHQ/rasa/blob/main/rasa/core/evaluation/marker_base.py#L450) and always add an "ANY_MARKER"
2. `len(markers.sub_markers)` instead of `len(markers)`

rasa/core/evaluation/marker_base.py
```diff
@@ -608,6 +609,9 @@ def evaluate_trackers(
             if tracker:
                 tracker_result = self.evaluate_events(tracker.events)
                 processed_trackers[tracker.sender_id] = tracker_result
+
+        processed_trackers_count = len(processed_trackers)
```
ka-bu, line 613: do we really want the number of processed trackers or processed sessions (or both)?

rasa/cli/evaluate.py
ka-bu: @usc-m , I could help and add that while you work on other comments if you like?

usc-m: Yeah, I suggested that too, also the branching factor (maximum number of children under any marker)

rasa/core/evaluation/marker_base.py
usc-m: I think in this case processed trackers is actually useful because we get it before, as a command line argument (the users intent) and after (what they actually got). Might help to highlight issues in our models of how tracker stores work. Right now we don't collect session info anywhere so I'm not sure what info we could get from it, but we could also just collect it as well if you think it would be useful

rasa/cli/evaluate.py
```diff
@@ -147,9 +147,14 @@ def _run_markers(
             "Please see errors listed above and fix before running again."
         )
 
+    # Calculate telemetry
     # Subtract one to remove the virtual OR over all markers
     num_markers = len(markers) - 1
-    telemetry.track_markers_parsed_count(num_markers)
+    max_depth = markers.depth() - 1
+    # Find maximum branching of marker
+    branching_factor = max(len(marker) - 1 for marker in markers)
```
ka-bu, line 155: `branching_factor = max(len(sub_marker) - 1  for marker in markers.sub_markers for for sub_marker in marker)` to exclude the artificial Or marker?

usc-m: Oh yep, otherwise we can end up with cases where the number of markers (which we already get) is returned twice - good spot

rasa/core/evaluation/marker_base.py
ka-bu: I have no clue how people use sender_ids and trackers usually, so no idea if that is better - but I guess it won't hurt if we don't add this now (and maybe later)

ka-bu (approved):
LGTM

aeshky (approved):
Looks good. There are still a couple of unresolved typos. Did you miss them? (one spotted by you even 😄)

PR #10075, opened 2021-11-03
Add Datadog to regression tests 

markus-hinsche:
**Proposed changes**:
- Send `test_run_time`, `train_run_time`, and `total_run_time` to Datadog via datadog API
- Visualize on a [Datadog dashboard](https://app.datadoghq.eu/dashboard/xsh-dvk-jca/mhinsches-dashboard-tue-nov-9-101519-am?from_ts=1636462443929&to_ts=1636466043929&live=true)
- enable for CPU and GPU

**Questions for reviewers:**
- I would like to know a good location in the codebase where to put the function (could be a new module) `transform_to_seconds`
- After this I would like to add unit tests for at least this function
- Do you suggest to add documentation for the feature? Where can I put this?


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]
```

.github/scripts/mr_publish_results.py
```diff
@@ -22,6 +25,66 @@
...
+    options = {
+        "statsd_host": "127.0.0.1",
+        "statsd_port": 8125,
+        "statsd_constant_tags": tags_list,
+    }
+    initialize(**options)
+
+    # Send  metrics
+    metrics = {
+        "test_run_time": os.environ["TEST_RUN_TIME"],
```
tczekajlo, line 79: Please use prefix for metrics, e.g. `rasa.perf.benchmark.test_run_time`

.github/workflows/ci-model-regression.yml
```diff
@@ -604,6 +622,10 @@ jobs:
           DATASET_REPOSITORY_BRANCH: ${{ matrix.dataset_branch }}
           TYPE: ${{ matrix.type }}
           DATASET_COMMIT: ${{ steps.set-dataset-commit.outputs.dataset_commit }}
+          BRANCH: ${{ steps.extract_branch.outputs.branch }}
+          DD_ENV: regression-tests # ${{ matrix.test }}
```
koernerfelicia, line 626: How come matrix.test is commented?

koernerfelicia:
> I would like to know a good location in the codebase where to put the function 

I think it could go in `rasa.utils.common`, but double-checking on this one (might be we don't want it in the codebase proper), I will get back to you! 

> Do you suggest to add documentation for the feature? Where can I put this?

I would say no. If it does go into the codebase it will need a docstring.

.github/scripts/mr_publish_results.py
```diff
@@ -22,6 +26,66 @@
...
+        "accelerator_type": os.environ["ACCELERATOR_TYPE"],
+        "github_run_id": os.environ["GITHUB_RUN_ID"],
+        "github_sha": os.environ["GITHUB_SHA"],
+        "github_event": os.environ["GITHUB_EVENT_NAME"],
+        "type": os.environ["TYPE"],
+        "branch": os.environ["BRANCH"],
+    }
+    tags_list = [f"{k}:{v}" for k, v in tags.items()]
+    options = {
+        "statsd_host": "127.0.0.1",
```
tczekajlo, line 72: why do you need it for? I assume you're using a DD agent to send metrics, it'll be better if you send the metrics via API. https://docs.datadoghq.com/api/latest/metrics/#submit-metrics

.github/workflows/ci-model-regression.yml
tczekajlo: It looks like it's copy & pastes from the other workflow where `${{ matrix.text }}` makes sense. Here a constant name is ok.

I'd change the name to `rasa-regression-tests`.

.github/workflows/ci-model-regression.yml
```diff
@@ -544,12 +550,19 @@ jobs:
...
+
       - name: Run test
         id: run_test
         if: steps.set_dataset_config_vars.outputs.is_dataset_exists == 'true' && steps.set_dataset_config_vars.outputs.is_config_exists == 'true'
         env:
           TFHUB_CACHE_DIR: ~/.tfhub_cache/
           OMP_NUM_THREADS: 1
+          DD_ENV: regression-tests
+          DD_SERVICE: rasa
+          DD_ARGS: --ddtrace
```
tczekajlo, line 565: It's not needed, it works only for pytests.

.github/workflows/ci-model-regression.yml
```diff
@@ -561,21 +574,21 @@ jobs:
           fi
 
           if [[ "${{ matrix.type }}" == "nlu" ]]; then
-            poetry run rasa train nlu --quiet -u ${DATASET_DIR}/${DATASET}/${TRAIN_DIR} -c dataset/configs/${CONFIG} --out ${DATASET_DIR}/models/${DATASET}/${CONFIG}
+            poetry run ddtrace-run rasa train nlu --quiet -u ${DATASET_DIR}/${DATASET}/${TRAIN_DIR} -c dataset/configs/${CONFIG} --out ${DATASET_DIR}/models/${DATASET}/${CONFIG}
```
tczekajlo, line 577: Looking at the data that we can get from APM, there are no benefits of using trace + APM in this particular case. I'd remove ddtrace.

.github/workflows/ci-model-regression.yml
```diff
@@ -418,6 +418,12 @@ jobs:
     if: "!contains(github.event.pull_request.labels.*.name, 'runner:gpu') && github.repository == 'RasaHQ/rasa' && contains(github.event.pull_request.labels.*.name, 'status:model-regression-tests') && needs.read_test_configuration.outputs.configuration_id != ''"
 
     steps:
+      - name: Run DataDog Agent
+        run: |
+          docker run -d --name dd_agent -p 8125:8125/udp -p 8126:8126 -v /var/run/docker.sock:/var/run/docker.sock:ro -v /proc/:/host/proc/:ro -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro -e DD_API_KEY=${{ secrets.DD_API_KEY }} -e DD_SITE="datadoghq.eu" -e DD_DOGSTATSD_NON_LOCAL_TRAFFIC=true -e "DD_INSIDE_CI=true" -e "DD_HOSTNAME=none" -e GITHUB_ACTIONS=true -e CI=true gcr.io/datadoghq/agent:7
```
tczekajlo, line 423: If we send metrics via API/DataDog client, and we removed ddtrace, using the agent is not needed then.

tczekajlo (changes requested):

.github/workflows/ci-model-regression.yml
tczekajlo: Or at least I wouldn't enable it by default, rather by an extra-label.

koernerfelicia:
> I would like to know a good location in the codebase where to put the function

Okay, so I think we would not put this into the codebase, but rather leave it where it is. You could maybe separate it out into its own file for other GHA to use, but I'd leave that decision up to infra (@tczekajlo ?)

tczekajlo:
> > I would like to know a good location in the codebase where to put the function
> 
> Okay, so I think we would not put this into the codebase, but rather leave it where it is. You could maybe separate it out into its own file for other GHA to use, but I'd leave that decision up to infra (@tczekajlo ?)

Let's leave where it is 😄

markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot"]
```

tczekajlo (approved):
LGTM ⭐

markus-hinsche:
@tczekajlo Sorry for adding commits after your LGTM. I just had the idea to enable the functionality also for GPU, this could work now since we don't need the docker dd_agent container anymore

koernerfelicia (approved):

Issue #9989, opened 2021-10-27
Replace deprecated `oauth2client` in favour of `google-auth`


joejuzl:
### Description

`oauth2client` has been [deprecated](https://google-auth.readthedocs.io/en/latest/oauth2client-deprecation.html).
It is now causing an error when `pyparsing>=3.0.0`. 
```
module 'pyparsing' has no attribute 'downcaseTokens'
```
See [sentry](https://sentry.io/organizations/rasahq/issues/2746125829/?project=2801673&referrer=slack).

The recommended library to change to is `google-auth`. 

We currently only use `oauth2client` in `hangouts.py` (to verify the bot token).

### Definition of Done:
- [ ] We no longer use `oauth2client` in Rasa OSS.
- [ ] We maintain current functionality for the hangouts channel.

Issue #9998, opened 2021-10-28
importing `DefaultV1Recipe` leads to AttributeError: module 'pyparsing' has no attribute 'downcaseTokens'


koaning:
On the main branch, importing this line of code in a `try_this.py` script file causes an error. 

```python
from rasa.engine.recipes.default_recipe import DefaultV1Recipe
```

This is the traceback. 

```
Traceback (most recent call last):
  File "components/test_nlu_tok.py", line 7, in <module>
    from rasa.engine.recipes.default_recipe import DefaultV1Recipe
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/rasa/engine/recipes/default_recipe.py", line 24, in <module>
    from rasa.engine.constants import (
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/rasa/engine/constants.py", line 3, in <module>
    from rasa.core.channels import UserMessage
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/rasa/core/channels/__init__.py", line 26, in <module>
    from rasa.core.channels.hangouts import HangoutsInput  # noqa: F401
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/rasa/core/channels/hangouts.py", line 9, in <module>
    from oauth2client import client
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/oauth2client/client.py", line 39, in <module>
    from oauth2client import transport
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/oauth2client/transport.py", line 17, in <module>
    import httplib2
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/httplib2/__init__.py", line 52, in <module>
    from . import auth
  File "/Users/vincent/Development/rasa-examples/venv/lib/python3.7/site-packages/httplib2/auth.py", line 20, in <module>
    auth_param_name = token.copy().setName("auth-param-name").addParseAction(pp.downcaseTokens)
AttributeError: module 'pyparsing' has no attribute 'downcaseTokens'
```

Here's the version info. 

```
Rasa Version      :         2.8.12
Minimum Compatible Version: 2.8.9
Rasa SDK Version  :         2.8.2
Rasa X Version    :         None
Python Version    :         3.7.7
Operating System  :         Darwin-19.3.0-x86_64-i386-64bit
[...]

PR #10078, opened 2021-11-03
Replace deprecated oauthclient2 with google-auth

tayfun:
**Proposed changes**:

- Fixes https://github.com/RasaHQ/rasa/issues/9989
- Replace oauthclient2 with google-auth
- This allows us to upgrade pyparsing and get rid of various [incompatibility](https://github.com/RasaHQ/rasa/issues/9998) bugs
- Add caching so Google certificates are not downloaded on each bot event


tayfun:
I think changelog is needed as it is not a user facing change.

rasa/core/channels/hangouts.py
```diff
@@ -254,21 +264,22 @@ def _extract_input_channel(self) -> Text:
...
-        # see https://developers.google.com/hangouts/chat/how-tos/bots-develop#verifying_bot_authenticity # noqa: E501, W505
+        # see https://developers.google.com/chat/how-tos/bots-develop#verifying_bot_authenticity # noqa: E501, W505
+        # and https://google-auth.readthedocs.io/en/latest/user-guide.html#identity-tokens # noqa: E501, W505
         try:
-            token = client.verify_id_token(
-                bot_token, self.project_id, cert_uri=CERT_URI
+            decoded_token = id_token.verify_token(
+                bot_token, self.google_request, audience=self.project_id
             )
+        except ValueError:
```
ancalita, line 273: Is this the only exception that could be raised?

rasa/core/channels/hangouts.py
```diff
@@ -254,21 +264,22 @@ def _extract_input_channel(self) -> Text:
         return self.name()
 
     def _check_token(self, bot_token: Text) -> None:
-        # see https://developers.google.com/hangouts/chat/how-tos/bots-develop#verifying_bot_authenticity # noqa: E501, W505
+        # see https://developers.google.com/chat/how-tos/bots-develop#verifying_bot_authenticity # noqa: E501, W505
+        # and https://google-auth.readthedocs.io/en/latest/user-guide.html#identity-tokens # noqa: E501, W505
         try:
-            token = client.verify_id_token(
-                bot_token, self.project_id, cert_uri=CERT_URI
+            decoded_token = id_token.verify_token(
```
ancalita, line 270: I only had a brief look at the docs and GH code, I see there's also a [`verify_oauth2_token` function](https://github.com/googleapis/google-auth-library-python/blob/e6278a815895e050e57fc516f086b4bc89a0864a/google/oauth2/id_token.py#L143), what's the reason for selecting `verify_token`?

rasa/core/channels/hangouts.py
```diff
@@ -207,8 +206,19 @@ def __init__(
         self.hangouts_room_added_intent_name = hangouts_room_added_intent_name
         self.hangouts_user_added_intent_name = hangouts_removed_intent_name
 
+        # Google's Request obj (this is used to make HTTP requests) and cached
```
ancalita, line 209: I would rewrite this comment a bit since it was confusing at first.
```suggestion
        # Google's Request obj (this is used to make HTTP requests) is stored as an attribute with the cached
```

ancalita (commented):
Great stuff 💯 
Left some minor questions, and also wanted to double-check there's already a unit test covering this change? If it's not, could you please add it?

ancalita:
> I think changelog is needed as it is not a user facing change.

You still need a changelog entry for this - it could be of type `misc`.

ancalita:
@tayfun One more thing 😄 I would also double-check if any update is required in the channel connector docs: https://rasa.com/docs/rasa/connectors/hangouts

tayfun:
> Great stuff 💯 Left some minor questions, and also wanted to double-check there's already a unit test covering this change? If it's not, could you please add it?

There are 3 tests already for hangouts channels only, but none for testing identity verification. Additions would be hard to test though, it's not doing much other than calling external libraries.

rasa/core/channels/hangouts.py
tayfun: `verify_token` is the more general version of the functionality, it can verify service requests such as those coming from Google Chat API. verify_oauth2_token and verify_firebase_token are more specialized functions, and you can't supply certification URLs as they are hardcoded (in earlier version of the code I had mistakenly removed the cert URL). It's a bit confusing :)

rasa/core/channels/hangouts.py
tayfun: This and `TransportError` but [TransportError](https://github.com/googleapis/google-auth-library-python/blob/8f1e9cfd56dbaae0dff64499e1d0cf55abc5b97e/google/oauth2/id_token.py#L102) we can't really do much about. If there's a network problem, we can't fetch certs and hence we can't verify request is coming from a Google service we can't really return 401. 500 default seems like a good response in that case.

changelog/README.md
```diff
@@ -1,11 +1,10 @@
-This directory contains "newsfragments" which are short files that contain a small 
+This directory contains "newsfragments" which are short files that contain a small
 **Markdown**-formatted text that will be added to the next `CHANGELOG`.
 
-The `CHANGELOG` will be read by **users**, so this description should be aimed 
-to Rasa OSS users instead of describing internal changes which are only relevant
```
tayfun, line 5: I've removed reference to `internal changes` here as I thought it was misleading. We do add internal change log within misc as you mentioned. @ancalita

tayfun:
> @tayfun One more thing 😄 I would also double-check if any update is required in the channel connector docs: https://rasa.com/docs/rasa/connectors/hangouts

Normally no update should be needed as it is just changing a dependency but I added a few clarifications :)

changelog/9989.misc.md
```diff
@@ -0,0 +1,2 @@
+Remove deprecated oauth2client and use google-auth instead. Also cache cert
```
joejuzl, line 1: Would be good to mention this is just for the hangouts channel.

joejuzl (changes requested):
looking good 👍

joejuzl (approved):
💯

ancalita (approved):
🎉

PR #10081, opened 2021-11-04
success markers / cleanup

ka-bu:
**Proposed changes**:
- change tags to tags used in rules (as far as possible)
- changed __iter__ -> flatten()
- removed ambiguous __len__ (can be replaced with explicit `len(marker.sub_markers)` or `len(marker.flatten()`)
- upper-case operators (that was the choice for slot mappings until that got removed - but think it's nice to make it easier to differentiate)


rasa/core/evaluation/marker.py
```diff
@@ -221,12 +216,12 @@ class SlotSetMarker(ConditionMarker):
...
     def positive_tag() -> Text:
         """Returns the tag to be used in a config file."""
-        return "slot_is_set"
+        return "slot_was_set"
 
     @staticmethod
     def negated_tag() -> Optional[Text]:
         """Returns the tag to be used in a config file for the negated version."""
-        return "slot_is_not_set"
+        return "slot_was_not_set"
```
aeshky, line 224: @ka-bu why did you decide to change them? After yesterday's discussion I felt that 'is' more accurately describes the behaviour (since slots must explicitly get unset for the condition to be false).

ka-bu: because 
- it's used in rules and it is unnecessary extra work to remember to switch to those slightly diffferent terms
- that mismatch was also brought in the feedback excercise
- the "slot_is_set" was not the only thing -> "intent_detected" is also just "intent" in rules
- the meaning is *the exact same* as in rules

rasa/core/evaluation/marker.py
```diff
@@ -80,15 +88,24 @@ def _to_str_with(self, tag: Text) -> Text:
...
+        for sequence_idx, sub_marker_idx in enumerate(self._progress):
+            next_sub_marker_idx = sub_marker_idx + 1
+            if self.sub_markers[next_sub_marker_idx].history[-1]:
+                self._progress[sequence_idx] = next_sub_marker_idx
+        # (2) If the first marker applies to the current time step, start a new
+        # sequence:
+        if self.sub_markers[0].history[-1]:
+            self._progress.append(0)
+        # (3) Check if any sequence is complete (where we exploit that, because of the
+        # "i_{j} < i_{j+1}"  condition, only one sequence in progress ccn be
```
usc-m, line 103: Shouldn't we complete all sequences at once?

If we have Seq(1,2,3) and we have a sequences of messages like 121212333, do we want that to match on the first 3 or all of the 3s (which I think is what will happen at the moment)? 

My understanding is that in that example, we start a sequence A at timestep 1, then add to A at 2, then a new one B at 3, then add to B at 4, then a new one C at 5, add to C at 6, then resolve A at 7, resolve B at 8, resolve C at 9?

usc-m: Or at least maybe this needs documented because I'm not sure that's the behaviour I'd expect (though it does have a certain logic to it)

ka-bu: two minds, one thought (https://rasa-hq.slack.com/archives/C01HHMR4X8S/p1636040258342600?thread_ts=1636036199.333900&cid=C01HHMR4X8S) :D

ka-bu: yeah think I should just make the progress and int and then just output a True whenever we see the "3" in the example above

usc-m: This makes me wonder if we could attach metadata to each marker match - it would be good to know when each stage of the sequence triggered (if they're close together it's more likely to be a logical flow rather than just incidentally the user hit those events over a long period of time) Not in scope here though

ka-bu: based on discussions: resetting after completed sequence now (and *not* interrupting sequences if a sub-marker triggers again)

rasa/core/evaluation/marker.py
```diff
@@ -18,12 +19,12 @@ class AndMarker(OperatorMarker):
     @staticmethod
     def positive_tag() -> Text:
         """Returns the tag to be used in a config file."""
-        return "and"
+        return "AND"
```
aeshky, line 22: Does this mean we're capitalising the operators in the config file?

ka-bu: yes, "upper-case operators (that was the choice for slot mappings until that got removed - but think it's nice to make it easier to differentiate)"

rasa/core/evaluation/marker.py
```diff
@@ -182,12 +189,12 @@ class IntentDetectedMarker(ConditionMarker):
...
     def positive_tag() -> Text:
         """Returns the tag to be used in a config file."""
-        return "intent_detected"
+        return "intent"
 
     @staticmethod
     def negated_tag() -> Optional[Text]:
         """Returns the tag to be used in a config file for the negated version."""
-        return "intent_not_detected"
+        return "not_intent"
```
aeshky, line 197: "not_intent" doesn't give the same meaning as "intent_not_detected".  It sounds like "tag all intents that are not this intent" rather than "tag dialogues when the intent never occurs".

ka-bu: "not_intent: bla" is true whenever "intent: bla" is not true - it is exactly what it sounds like

ka-bu: for "never" and "occurs" we have the operators

ka-bu: .... ok, then we'll switch back to the old ones

rasa/core/evaluation/marker.py
ka-bu: So, do you think this is easier to read/write or not?

ka-bu:
... whatever we do with the tags, as long as can please change that __len__ and __iter__, I'm happy cause I think that is a bit too confusing really :D 

Again: There was some feedback from those feedback exercises (or was it Vincent? 🤔 ) pointing out that it is strange that it's a new set of tags one has to learn (that could be replaced by those from rules), which is why I suggested these changes. If people think this is too confusing, then let's switch back 🤷‍♀️

rasa/core/evaluation/marker.py
ka-bu: but to avoid confusion: "intent_not_detected" does not mean "tag dialogues when the intent never occurs" ... the marker "intent_not_detected : bla " applies to every event in a conversation where the current action is *not* a user uttered event with "bla" as intent. We're not tagging dialogues per se, but we output statistics on per-session level and the overall statistics over all sessions. So if we want to "tag dialogues when..." , we could load the per-session results and look up for which sessions there is at least one relevant event logged for this marker. (Or, if we want to make sure we log this only once, we add an "at_least_once" marker, for which only one relevant event will be logged, namely the first event).

rasa/core/evaluation/marker.py
aeshky: I got used to the lowercase one but I think they are both fine. No strong preference.

rasa/core/evaluation/marker.py
aeshky: Okay I see. Then the meaning of the original label changed. It's fine as long as we're clear about what it means. As you say, people can use "NEVER" for the other meaning.

rasa/core/evaluation/marker.py
aeshky: 👍

aeshky:
> Again: There was some feedback from those feedback exercises (or was it Vincent? 🤔 ) pointing out that it is strange that it's a new set of tags one has to learn (that could be replaced by those from rules)

I think it was both the feedback exercises and then Vincent. I recall Greg and Ben also mentioning something similar.

aeshky (approved):

rasa/core/evaluation/marker.py
ka-bu: By original you mean the previous implementation, right? Yes, the "intent_detected" is now defined exactly as in weizenbot (https://www.notion.so/rasa/Weizenbot-Conditions-Docs-c268cdaa34cd4ec9a68d067620c4f4d7#8739eae05a684e58a73e7d5f3d5f856c) and afaik "intent_not_detected"  does not exist there (We added those "not" versions only because they were in the original proposal). In the previous implementation, there was an implicit "at_least_once" in front of every atomic condition (i.e. a negated "never" in weizenbot).

rasa/core/evaluation/marker.py
ka-bu: @usc-m how about you?

usc-m: I prefer the lowercase one as well since it's less effort to type and reads visually a bit nicer to me, but it's not a strong preference either

ka-bu: the typing issue is a good point ... looks like all in all lower case wins this, will revert :)

usc-m (approved):
LGTM

Issue #10041, opened 2021-11-01
Scheduled Model Regression Test Failed


rasabot:
*This PR is automatically created by the Scheduled Model Regression Test workflow. Checkout the Github Action Run [here](https://github.com/RasaHQ/rasa/actions/runs/1405766870).* <br> --- <br> **Description of Problem:** <br> Scheduled Model Regression Test failed. <br> **Configuration**: `Rules + AugMemo` <br> **Dataset**: `retail-demo`

tttthomasssss:
Affected datasets:
* financial-demo ([PR](https://github.com/RasaHQ/financial-demo/pull/172))
* insurance-demo ([PR](https://github.com/RasaHQ/insurance-demo/pull/19))
* retail-demo ([PR](https://github.com/RasaHQ/retail-demo/pull/35))
* helpdesk-assistant ([PR](https://github.com/RasaHQ/helpdesk-assistant/pull/58))
* Sara ([Branch](https://github.com/RasaHQ/training-data/tree/fix-model-regression-tests) with fixed `domain.yml`)

tttthomasssss:
Currently the new `domain.yml` file is not compatible with `rasa<3.0` (see Melinda's comment [here](https://github.com/RasaHQ/financial-demo/pull/172#issuecomment-958776250)). That means that we need to run the scheduled model regression tests from a different branch for `insurance-demo`, `financial-demo`, `retail-demo` and `helpdesk-assistant` until `rasa 3.0` is out. 

Specifying a different branch for the scheduled test for a single dataset from the `training-data` repo is slightly more involved (see [Slack thread](https://rasa-hq.slack.com/archives/C01H2UCNP2T/p1635947670068600)). 

Given the situation with the incompatible domain file is only temporary (until Rasa 3.0 is out), and that the last remaining dataset affected by the change is Sara from the `training-data` repository, the simplest solution is to have the old and new domain file in the Sara dataset and to remove the old version once Rasa 3.0 is released.

tttthomasssss:
fixed with test runs #10085 and #10088

PR #10085, opened 2021-11-04
dummy change to trigger model regression tests

tttthomasssss:
dummy change to trigger model regression tests (addresses #10041 et al.)


tttthomasssss:
/modeltest

```yaml
include:
  - dataset: ["all"]
    config: ["all"]
```

PR #10086, opened 2021-11-05
conversation_granularity for slack - per user | per user per channel | per user per thread

srinivasupadhya:
allow conversation_id to be per user | per user per channel | per user per thread

**Proposed changes**:
- Allow multiple active conversations for a user on slack with bot


ka-bu:
Thanks for raising the issue and implementing the fix 🚀

What will happen next:
- [x] I'll enable the tests and will get some feedback from more people cause it looks to me like the `conversation_id` that you constructed might be a more reasonable default value for the `sender_id` than the current choice
- [ ] we'll address the test issues / feedback / review items
- [ ] we'll need to update the rasa docs (https://rasa.com/docs/rasa/connectors/slack/) and add a changelog entry 

And then we can merge this into the new rasa 3.0.0 :)

rasa/core/channels/slack.py
```diff
@@ -182,7 +184,10 @@ def __init__(
             slack_signing_secret: Slack creates a unique string for your app and
                 shares it with you. This allows us to verify requests from Slack
                 with confidence by verifying signatures using your signing secret.
-
+            conversation_id_config: conversation id config for slack conversations.
```
tayfun, line 187: Config has the connotation of a more general configuration, sometimes a file or a larger amount of options. Would another name such as `conversation_level` or `conversation_granularity` (with values being `sender`, `channel`, `thread`) make more sense here?

Also it makes sense to set and talk about the default here in my opinion.

srinivasupadhya:
@ka-bu added docs for the config. Changing default would be backward incompatible (new rasa with old conversations in tracker store) & hence haven't changed this.

ka-bu:
Thanks for the quick reply and addressing the feedback!

> @ka-bu added docs for the config. Changing default would be backward incompatible (new rasa with old conversations in tracker store) & hence haven't changed this.

Great that you thought of this :) . I've been trying to get some feedback in the meantime on whether it is ok to make that breaking change for 3.0. For now, there is no feedback so your non-breaking version is what we proceed with, it seems.
 
> Also, is there anyway we can backport this fix to 2.8.x release?

If you like, you can open another PR and apply the same changes there to get this merged into 2.8.x. as well (by choosing that version as a target for your merge then). Otherwise, we can open an issue and someone from our team will pick that up.

ka-bu:
I like the new name for that setting and the names for the options 👍  -- Just one tiny linting issue so far from the checks.

srinivasupadhya:
@ka-bu addressed the formatting issue. thanks for the inputs, will raise a separate PR for 2.8.x.

tayfun:
> @ka-bu addressed the formatting issue. thanks for the inputs, will raise a separate PR for 2.8.x.

Thanks for your contributions! Once this is merged you can cherry-pick your commit and create a new PR for 2.8.x and we should be quick to get that one in. 

By the way, you can use pre-commit hooks to run checks locally or just use `black` manually to check formatting issues on your own machine.

ka-bu:
Reminder: We also have to add a changelog entry (i.e. a `<pr number>.improvement` file with a description under the changelog folder)

srinivasupadhya:
@ka-bu @tayfun everything seems to be green now. let me know if there is anymore work here.

changelog/10086.misc.md
```diff
@@ -0,0 +1 @@
+`conversation_granularity` config for slack channel, supports `sender`, `channel`, `thread`.
```
ka-bu, line 1:
```suggestion
A new additional slack options `conversation_granularity` is now available  that allows to choose whether interactions across all channels ("sender"), within a channel ("channel") or only within a thread ("thread") will be treated as a single conversation.
```

docs/docs/connectors/slack.mdx
```diff
@@ -164,6 +164,7 @@ slack:
   slack_retry_number_header: "x-slack-retry-num"  # Slack HTTP header name indicating the attempt number. This configuration is optional.
   errors_ignore_retry: None  # Any error codes given by Slack included in this list will be ignored. Error codes are listed [here](https://api.slack.com/events-api#errors).
   use_threads: False  # If set to True, bot responses will appear as a threaded message in Slack. This configuration is optional and set to False by default.
+  conversation_granularity: "sender" # sender allows 1 conversation per user (across channels), channel allows 1 conversation per user per channel, thread allows 1 conversation per user per thread. This configuration is optional and set to sender by default.
```
ka-bu, line 167: @tayfun @srinivasupadhya  ,  Shall we add something like "In case a retrieved message does not specify the necessary information for the required level of granularity, then Rasa will fall back to use the next possible level of granularity and use the respective information as 'sender_id' (e.g. if `thread_id` is set to `None` and a `channel` is specified, while `conversation_granularity` is set to `thread`, then the `sender_id` will be a combination of the `user` and `channel` information only)."   -- It is kind of obvious that we can't get/use that info if it isn't there. But just to raise awareness that this might happen?

rasa/core/channels/slack.py
```diff
@@ -506,6 +512,11 @@ async def webhook(request: Request) -> HTTPResponse:
                 user_message = event.get("text", "")
                 sender_id = event.get("user", "")
                 metadata = self.get_metadata(request)
+                channel_id = metadata.get("out_channel")
+                thread_id = metadata.get("thread_id")
+                conversation_id = self._get_conversation_id(
+                    sender_id, channel_id, thread_id
```
ka-bu, line 518: Think to avoid confusion, it might be good to replace `sender_id` with `user_id` and `conversation_id` with `sender_id`. Wdyt?

ka-bu (commented):
Some tiny suggestions. Otherwise looks good to me 🚀

docs/docs/connectors/slack.mdx
tayfun: Might be a good idea, makes it more clear for sure.

srinivasupadhya: This doesn't seem to happen for a slack message, but happened during a webhook validation on slack app setup page (Event Subscriptions). If channel / thread data is missing, responding to the message will not be possible & hence it is unlikely to happen. I think we can skip this.

rasa/core/channels/slack.py
srinivasupadhya: i wanted to make sure people understand we are trying to compute a `conversation_id` instead of `sender_id`. I can replace `sender_id` if you feel that would be better.

tayfun: I see what you mean. We use this below when we call `process_message` and we currently do 
```
sender_id=conversation_id,
```
I think it makes sense to rename the variables as you recommend. `sender_id` is Rasa's own name, whereas `user_id` we get from Slack (along with channel and thread IDs).

ka-bu: yeah, that is a good point -- I agree that conversation_id is a much nicer name -- the confusion would just be caused on our side because of that historic name 😅  . Would be great if we could replace that `sender_id` by `user` or `user_id` or so 👍

ka-bu (approved):
Feel free to squash merge when ready 🚀

tayfun (approved):

srinivasupadhya:
@ka-bu seems like coverage report upload job failed. Not sure how to proceed. Also, #10146 is ready as well.

PR #10090, opened 2021-11-05
Add og:image to the docs

maggo:
**Proposed changes**:
- Adds `og:image` file and property in the custom docs theme


lunelson (changes requested):
Nice find @maggo ! I think this is an elegant solution. Can you verify whether the og-image related tags are now complete, and add any that are missing? See [metatags.io](https://metatags.io/) as a reference

@m-vdb if you're cool with this method, Marco or I would repeat the same on the other repos

maggo:
> Can you verify whether the og-image related tags are now complete, and add any that are missing? See [metatags.io](https://metatags.io/) as a reference

Had a quick check, all required fields title, description, url, image will be there after this merge

lunelson (approved):

m-vdb:
@lunelson good for me, but to be honest I'm missing context and I don't think I can give a proper review. If you're fine with this change, I'm fine.

lunelson:
@m-vdb yeah this is the way to do this, AFAICT for the `alpha.63` version of Docusaurus that we're on here, and we have the custom theme anyway; I've made a [PR against `rasa-x`](https://github.com/RasaHQ/rasa-x/pull/6751) which uses the newer config options however

lunelson:
BTW @m-vdb is `main` the right merge target here?

m-vdb:
yes I think so

maggo:
Rebased onto the latest master.
Will also port this change to https://github.com/RasaHQ/rasa-sdk

PR #10097, opened 2021-11-07
unhandle echo messages from messenger

aristidednd:
According to Facebook Messenger API documentation https://developers.facebook.com/docs/messenger-platform/reference/webhook-events/message-echoes/?locale=en_En, an echo message is also triggered for image, video, file, and audio attachments. 
These changes ignore those echo messages.

**Proposed changes**:
- Ignore echo messages related to file, audio, video, and image attachments from Facebook

### DIFF
--- a/rasa/core/channels/facebook.py
+++ b/rasa/core/channels/facebook.py
@@ -45,6 +45,7 @@ class Messenger:
             "message" in message
             and "attachments" in message["message"]
             and message["message"]["attachments"][0]["type"] == "audio"
+            and not message["message"].get("is_echo")
         )
 
     @staticmethod
@@ -54,6 +55,7 @@ class Messenger:
             "message" in message
             and "attachments" in message["message"]
             and message["message"]["attachments"][0]["type"] == "image"
+            and not message["message"].get("is_echo")
         )
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


rasa/core/channels/facebook.py
```diff
@@ -45,6 +45,7 @@ def _is_audio_message(message: Dict[Text, Any]) -> bool:
             "message" in message
             and "attachments" in message["message"]
             and message["message"]["attachments"][0]["type"] == "audio"
+            and not message["message"].get("is_echo")
```
ancalita, line 48: I think adding this in the different `_is_<type>_message` checks leads to these methods having more than 1 purpose. I suggest instead extracting this boolean variable in a separate function `_is_echo_message` and then adding this check once only at the top of the `message` method before all the other checks:
```
    async def message(
        self, message: Dict[Text, Any], metadata: Optional[Dict[Text, Any]]
    ) -> None:
        """Handle an incoming event from the fb webhook."""
        if self._is_echo_message(message):
                logger.warning("Received a message echo from Facebook which will be ignored.")
                return None
        # quick reply and user message both share 'text' attribute
        # so quick reply should be checked first
        if self._is_quick_reply_message(message):
              ...
```

ancalita (commented):
@aristidednd Thank you for your contribution! 💯 
Could you please add a changelog entry specifying the why and how of your improvement.
Also if possible please add unit test coverage.

Tung8787:
Tui chả hiểu hì :)))

PR #10159, opened 2021-11-10
3.0 remove user warnings

### LINKED ISSUE #9777: Final check to remove any comments or `UserWarnings` that reference 3.0
**Definition of Done**
 - [ ] search for any comments or UserWarnings referencing 3.0  (For example, [this](https://github.com/RasaHQ/rasa/blob/abe75cd7fcf89b92ec60fa9f7031eb8059454042/rasa/core/processor.py#L199) and [this](https://github.com/RasaHQ/rasa/blob/abe75cd7fcf89b92ec60fa9f7031eb8059454042/rasa/utils/train_utils.py#L521))
- [ ] adapt any outstanding items and remove comments/warnings

carlad:
**Proposed changes**:
Closes #9777

Removes and/or updates `UserWarnings` pertaining to 3.0 changes.

### DIFF
--- a/rasa/utils/train_utils.py
+++ b/rasa/utils/train_utils.py
@@ -556,9 +556,7 @@ def _check_loss_setting(component_config: Dict[Text, Any]) -> None:
     ):
         rasa.shared.utils.io.raise_warning(
             f"{CONSTRAIN_SIMILARITIES} is set to `False`. It is recommended "
-            f"to set it to `True` when using cross-entropy loss. It will be set to "
-            f"`True` by default, "
-            f"Rasa Open Source 3.0.0 onwards.",
+            f"to set it to `True` when using cross-entropy loss.",
             category=UserWarning,
         )
 
--- a/README.md
+++ b/README.md
@@ -18,12 +18,11 @@
 
 <hr />
 
-💡 **Rasa Open Source 3.0 is coming up!** 💡
+💡 **Rasa Open Source 3.0 is here!** 💡
 
-[2.8](https://github.com/RasaHQ/rasa/milestone/39) will be the last minor in the 2.x series, as we need to pause releasing minors while we work on architectural changes in 3.0. You can still contribute new features and improvements which we plan to release together with 3.0. Read more
-about [our contributor guidelines](#how-to-contribute).
-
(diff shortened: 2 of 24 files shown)


rasa/utils/train_utils.py
```diff
@@ -556,9 +556,7 @@ def _check_loss_setting(component_config: Dict[Text, Any]) -> None:
     ):
         rasa.shared.utils.io.raise_warning(
             f"{CONSTRAIN_SIMILARITIES} is set to `False`. It is recommended "
-            f"to set it to `True` when using cross-entropy loss. It will be set to "
-            f"`True` by default, "
-            f"Rasa Open Source 3.0.0 onwards.",
+            f"to set it to `True` when using cross-entropy loss.",
```
usc-m, line 559: So this one seems a bit weird to me, since I figured if it's forced to `True` we wouldn't really need to warn at all. I looked into it, and this function is only used it one place (`validate_configuration_settings`), and ultimately is only called from `DIETClassifier` (which has a default value of `False` for `CONSTRAIN_SIMILARITIES`). I see when I try `rasa train` I don't get a warning, and it looks like the default pipeline has it set to `True`, but if we're going to warn when people set it to `False` is there a reason we let them? @samsucik do you have some more context on this?

alwx (approved):
Please, take a look at the comment written by @usc-m. The rest looks good for me! 🚀

carlad:
Thanks @alwx I saw the comment but was waiting for @samsucik to advise.

rasa/utils/train_utils.py
samsucik: @usc-m @carlad honestly, I'm confused by the original warning. Like, what does it mean `It will be set to True by default, Rasa Open Source 3.0.0 onwards.`? Does that mean that in 3.0.0 we'll override whatever user-set value with `True`, or that the default pipelines will set it to `True` but the user can still override it? @dakshvar22 maybe you know 🙂 If my first interpretation above is correct, then I'd change & rename `_check_loss_setting` such that it sets `constrain_similarities=True` and perhaps informs the user. If the 2nd interpretation is correct, I'd go with the warning as updated by @carlad.

dakshvar22: > or that the default pipelines will set it to True but the user can still override it?

This

Issue #10101, opened 2021-11-08
Model file not recognized in `rasa shell` for 3.0 RC1


tayfun:
### Rasa Open Source version

3.0.0rc1

### Python version

3.8

### What operating system are you using?

OSX

### What happened?

1. Init a new project - either in current directory or a new one, it doesn't matter
2. If you speak to the assistant during `rasa init`, it works; but if you run `rasa shell` afterwards, you get this error:

```
$ rasa shell
Traceback (most recent call last):
...
IsADirectoryError: [Errno 21] Is a directory: 'models/20211108-105331.tar.gz'
```
If you check directory structure it is really a directory:
```
$ tree models/
models/
└── 20211108-105331.tar.gz
    └── 20211108-105334-inclusive-dam.tar.gz

1 directory, 1 file
```
Removing `models` directory and running `rasa train` fixes the problem:
```
 $ tree models/
models/
├── 20211108-104603.tar.gz
└── 20211108-104824-spicy-teaser.tar.gz

1 directory, 1 file

```

PR #10164, opened 2021-11-10
Fix model dir

tayfun:
Fixes https://github.com/RasaHQ/rasa/issues/10101

### DIFF
--- a/rasa/cli/scaffold.py
+++ b/rasa/cli/scaffold.py
@@ -6,13 +6,13 @@ from typing import List, Text
 from rasa import telemetry
 from rasa.cli import SubParsersAction
 from rasa.cli.shell import shell
-from rasa.cli.utils import create_output_path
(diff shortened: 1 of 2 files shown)


carlad (approved):

tayfun:
> LGTM 👍 Re: the check boxes in the description. Are we supposed to check them even if we haven't updated documentation, updated the changelog, or added tests?

Haha I think so :D 

I mean I check boxes for items done or not applicable. It's just a system to check you have considered all in my opinion. 

Docs need not changed here as it is a bugfix, changelog is not needed because it is not a bug that's been released (only in RC yet), and I think again tests are not needed as I'm only removing buggy code.

joejuzl:
>  Are we supposed to check them even if we haven't updated documentation, updated the changelog, or added tests?
I normally remove the whole line if it's not applicable.

joejuzl (approved):
💯

PR #10167, opened 2021-11-10
Validations for unused intents and utterances

### LINKED ISSUE #10079: Data validation with --fail-on-warnings fails but there are no warnings printed
### Rasa Open Source version

2.8.6

### Rasa SDK version

2.8.2

### Rasa X version

None

### Python version

3.8

OSX

### What happened?

Output for `rasa data validate`:

```
2021-11-03 16:22:58 INFO     rasa.validator  - Validating intents...
2021-11-03 16:22:58 INFO     rasa.validator  - Validating uniqueness of intents and stories...
2021-11-03 16:22:58 INFO     rasa.validator  - Validating utterances...
2021-11-03 16:22:58 INFO     rasa.validator  - Story structure validation...
Processed story blocks: 100%|█████████████████████████████████████████████████| 38/38 [00:00<00:00, 365.67it/s, # trackers=1]
2021-11-03 16:22:58 INFO     rasa.core.training.story_conflict  - Considering all preceding turns for conflict analysis.
2021-11-03 16:23:12 INFO     rasa.validator  - No story structure conflicts found.
```

Output for `rasa data validate --fail-on-warnings`:

```
2021-11-03 15:35:55 INFO     rasa.validator  - Validating intents...                                                                                                                                                                                        
2021-11-03 15:35:55 INFO     rasa.validator  - Validating uniqueness of intents and stories...                                                                                                                                                              
2021-11-03 15:35:56 INFO     rasa.validator  - Validating utterances...                                                                                                                                                                                     
Project validation completed with errors.
```

There are no warnings printed for both commands but the one with `--fail-on-warnings` fails.

When running with `-vv` option I can see `DEBUG` level logs (but still no warnings) like:
```
2021-11-03 17:36:21 DEBUG    rasa.validator  - The intent 'back' is listed in the domain file, but is not found in the NLU training data.
```
and
```
2021-11-03 17:36:21 DEBUG    rasa.validator  - The utterance 'utter_default' is not used in any story.
```

Possibly related source code:
* https://github.com/RasaHQ/rasa/blob/2.8.6/rasa/validator.py#L137
* https://github.com/RasaHQ/rasa/blob/2.8.6/rasa/validator.py#L196

### Command / Request

```shell
rasa data validate --fail-on-warnings
```

### Relevant log output

```shell
2021-11-03 15:35:55 INFO     rasa.validator  - Validating intents...                                                                                                                                                                                        
2021-11-03 15:35:55 INFO     rasa.validator  - Validating uniqueness of intents and stories...                                                                                                                                                              
2021-11-03 15:35:56 INFO     rasa.validator  - Validating utterances...                                                                                                                                                                                     
Project validation completed with errors.
```

**Definition of done**
- [ ] output for `rasa data validate --fail-on-warnings` correct

losterloh:
Fixes #10079.

The following changes were made:
- we exclude the default / implicit intents from generating warnings if they're unused; otherwise, any empty / newly initialized rasa project would throw warnings for unused intents, which would be confusing to the user
- if an intent is defined, but not used in the NLU data at all, we generate a warning and consequently abort if `--fail-on-warnings`.
- if an intent is defined, but not used either in a story or rule we generate a warning and consequently abort if `--fail-on-warnings`. Note here that the debug message that was printed before was also inexact since it mentioned only stories, this was clarified to also include rules.
- The same as in the bullet point before was also implemented for utterances

### DIFF
--- a/data/test_config/config_defaults.yml
+++ b/data/test_config/config_defaults.yml
@@ -24,3 +24,16 @@ pipeline: []
 #     ambiguity_threshold: 0.1
 
 data:
+policies:
+# # No configuration for policies was provided. The following default policies were used to train your model.
+# # If you'd like to customize them, uncomment and adjust the policies.
+# # See https://rasa.com/docs/rasa/policies for more information.
+#   - name: MemoizationPolicy
+#   - name: RulePolicy
+#   - name: UnexpecTEDIntentPolicy
+#     max_history: 5
+#     epochs: 100
+#   - name: TEDPolicy
+#     max_history: 5
+#     epochs: 100
+#     constrain_similarities: true
--- a/data/test_validation/data/stories.yml
+++ b/data/test_validation/data/stories.yml
@@ -1,0 +1,6 @@
+version: "2.0"
+stories:
+- story: happy path
+  steps:
+  - intent: greet
+  - action: utter_greet
--- a/rasa/validator.py
+++ b/rasa/validator.py
@@ -1,7 +1,7 @@
 import itertools
 import logging
 from collections import defaultdict
-from typing import Set, Text, Optional, Dict, Any
+from typing import Set, Text, Optional, Dict, Any, List
 
 import rasa.core.training.story_conflict
 import rasa.shared.nlu.constants
@@ -12,6 +12,7 @@ from rasa.shared.constants import (
     DOCS_URL_ACTIONS,
     REQUIRED_SLOTS_KEY,
 )
+from rasa.shared.core import constants
 from rasa.shared.core.constants import MAPPING_CONDITIONS, ACTIVE_LOOP
 from rasa.shared.core.domain import Domain
 from rasa.shared.core.events import ActionExecuted, ActiveLoop
@@ -58,15 +59,22 @@ class Validator:
 
         return cls(domain, intents, story_graph, config)
 
+    def _non_default_intents(self) -> List[Text]:
+        return [
+            item
+            for item in self.domain.intents
+            if item not in constants.DEFAULT_INTENTS
+        ]
+
     def verify_intents(self, ignore_warnings: bool = True) -> bool:
         """Compares list of intents in domain with intents in NLU training data."""
         everything_is_alright = True
 
         nlu_data_intents = {e.data["intent"] for e in self.intents.intent_examples}
 
-        for intent in self.domain.intents:
+        for intent in self._non_default_intents():
             if intent not in nlu_data_intents:
-                logger.debug(
+                rasa.shared.utils.io.raise_warning(
                     f"The intent '{intent}' is listed in the domain file, but "
                     f"is not found in the NLU training data."
                 )
@@ -134,9 +142,11 @@ class Validator:
                 )
                 everything_is_alright = False
 
-        for intent in self.domain.intents:
+        for intent in self._non_default_intents():
             if intent not in stories_intents:
-                logger.debug(f"The intent '{intent}' is not used in any story.")
+                rasa.shared.utils.io.raise_warning(
+                    f"The intent '{intent}' is not used in any story or rule."
+                )
                 everything_is_alright = ignore_warnings and everything_is_alright
 
         return everything_is_alright
@@ -195,7 +205,9 @@ class Validator:
 
         for utterance in utterance_actions:
             if utterance not in stories_utterances:
-                logger.debug(f"The utterance '{utterance}' is not used in any story.")
+                rasa.shared.utils.io.raise_warning(
+                    f"The utterance '{utterance}' is not used in any story or rule."
+                )
                 everything_is_alright = ignore_warnings and everything_is_alright
 
         return everything_is_alright
--- a/tests/test_validator.py
+++ b/tests/test_validator.py
@@ -1,11 +1,25 @@
 from typing import Text
 
 import pytest
+from _pytest.logging import LogCaptureFixture
 
 from rasa.validator import Validator
 from rasa.shared.importers.rasa import RasaFileImporter
 from rasa.shared.importers.autoconfig import TrainingType
 from pathlib import Path
+
+
+@pytest.fixture(scope="class")
+def validator_under_test() -> Validator:
+    importer = RasaFileImporter(
+        domain_path="data/test_validation/domain.yml",
+        training_data_paths=[
+            "data/test_validation/data/nlu.yml",
+            "data/test_validation/data/stories.yml",
+        ],
+    )
+    validator = Validator.from_importer(importer)
+    return validator
 
 
 def test_verify_nlu_with_e2e_story(tmp_path: Path, nlu_data_path: Path):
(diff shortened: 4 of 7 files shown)


rasa/validator.py
```diff
@@ -64,9 +65,14 @@ def verify_intents(self, ignore_warnings: bool = True) -> bool:
 
         nlu_data_intents = {e.data["intent"] for e in self.intents.intent_examples}
 
-        for intent in self.domain.intents:
+        # Exclude the default intents that are always implicitly present
+        for intent in [
+            item
+            for item in self.domain.intents
+            if item not in constants.DEFAULT_INTENTS
+        ]:
```
wochinge, line 73: optional: We could also make this a property on `Domain` to avoid duplicating the logic

data/test_validation/data/stories.yml
```diff
@@ -0,0 +1,7 @@
+version: "2.0"
+stories:
+- story: happy path
+  steps:
+  - intent: greet
+  - action: utter_greet
+
```
wochinge, line 7:
```suggestion
```

tests/test_validator.py
```diff
@@ -213,6 +226,44 @@ def test_verify_there_is_example_repetition_in_intents(nlu_data_path: Text):
     assert not validator.verify_example_repetition_in_intents(False)
 
 
+def test_verify_logging_message_for_intent_not_used_in_nlu(
+    caplog, validator_under_test
```
wochinge, line 230: can we please add type annotations (same for the others)?

rasa/validator.py
```diff
@@ -64,9 +65,14 @@ def verify_intents(self, ignore_warnings: bool = True) -> bool:
...
             if intent not in nlu_data_intents:
-                logger.debug(
+                rasa.shared.utils.io.raise_warning(
```
wochinge, line 75: I've noticed that there are a couple of others with `debug` messages. In my opinion we should be consistent and have them all as warnings, what do you think?

wochinge (commented):
Welcome to Rasa Open Source @losterloh 🚀

tests/test_validator.py
losterloh: Yes; however, if that's what we want, and everywhere, I suggest we should enforce this with a linter.

rasa/validator.py
losterloh: Yes, I've noticed the same thing, but I feel like not all of them are clear cut - which actually  should be more of a warning, and which are really purely for debug purposes? Happy to go through with somebody who's in deeper already than me.

rasa/validator.py
losterloh: Yup can definitely do! Do we have a general policy for code repetitions and when to de-duplicate them? My usual approach is to tackle those when code would otherwise be triplicated.

data/test_validation/data/stories.yml
losterloh: Whoops! You're right of course.

tests/test_validator.py
losterloh: For instance, in this file, the `caplog` that is used to capture log output is not annotated anywhere, the type would even need to be imported into the file first from a private package. I'll do this now.

rasa/validator.py
wochinge: Thanks for changing the to warning as well 👍🏻 Makes way more sense now 🎉

tests/test_validator.py
```diff
@@ -213,6 +227,46 @@ def test_verify_there_is_example_repetition_in_intents(nlu_data_path: Text):
...
+
+def test_verify_logging_message_for_unused_utterance(
+    caplog: LogCaptureFixture, validator_under_test: Text
+):
+    caplog.clear()
+    with pytest.warns(UserWarning) as record:
+        validator_under_test.verify_utterances_in_stories(False)
+
+    for m in record:
+        print(m)
```
wochinge, line 264:
```suggestion
```

data/test_config/config_defaults.yml
```diff
@@ -24,3 +24,16 @@ pipeline: []
...
+# # See https://rasa.com/docs/rasa/policies for more information.
+#   - name: MemoizationPolicy
+#   - name: RulePolicy
+#   - name: UnexpecTEDIntentPolicy
+#     max_history: 5
+#     epochs: 100
+#   - name: TEDPolicy
+#     max_history: 5
+#     epochs: 100
+#     constrain_similarities: true
```
wochinge, line 39: is this change actually needed?

wochinge (commented):
> Yes; however, if that's what we want, and everywhere, I suggest we should enforce this with a linter.

Good point 👍🏻 We do this for the production code but are less strict in the test code 😬

tests/test_validator.py
losterloh: 🤦  thanks. Another thing where we could think about involving the linter 😓

wochinge: There is actually a https://pypi.org/project/flake8-print/ 👍🏻 Will add create an issue for it 👍🏻

losterloh: Perfect! 🎉

wochinge: https://github.com/RasaHQ/rasa/issues/10205

rasa/validator.py
wochinge: > My usual approach is to tackle those when code would otherwise be triplicated.

I think that's a good rule of thumb. In this case I just felt like we'd need that in several places sooner or later and it felt quite "natural" to the Domain. I like the solution which you chose 💯

data/test_config/config_defaults.yml
losterloh: I think it happens around here:
https://github.com/RasaHQ/rasa/blob/d13b153574acb1a70dd4c170a8bc23b514e48a63/rasa/shared/importers/autoconfig.py#L67

wochinge: Mhm, I see your point. We should probably change the tests to not write there instead 🙈 Fine to keep it as it is 👍🏻

wochinge: yep, but in my opinion the tests should e.g. copy the config before doing this.

tests/test_validator.py
```diff
@@ -213,6 +227,44 @@ def test_verify_there_is_example_repetition_in_intents(nlu_data_path: Text):
...
+    with pytest.warns(UserWarning) as record:
+        validator_under_test.verify_intents_in_stories(False)
+
+    assert "The intent 'goodbye' is not used in any story or rule." in (
+        m.message.args[0] for m in record
+    )
+
+
+def test_verify_logging_message_for_unused_utterance(
+    caplog: LogCaptureFixture, validator_under_test: Text
```
wochinge, line 257:
```suggestion
    caplog: LogCaptureFixture, validator_under_test: Validator
```

tests/test_validator.py
```diff
@@ -213,6 +227,44 @@ def test_verify_there_is_example_repetition_in_intents(nlu_data_path: Text):
...
+
+    assert (
+        "The intent 'goodbye' is listed in the domain file, "
+        "but is not found in the NLU training data."
+        in (m.message.args[0] for m in record)
+    )
+
+
+def test_verify_logging_message_for_intent_not_used_in_story(
+    caplog: LogCaptureFixture, validator_under_test: Text
```
wochinge, line 245:
```suggestion
    caplog: LogCaptureFixture, validator_under_test: Validator
```

tests/test_validator.py
```diff
@@ -213,6 +227,44 @@ def test_verify_there_is_example_repetition_in_intents(nlu_data_path: Text):
     assert not validator.verify_example_repetition_in_intents(False)
 
 
+def test_verify_logging_message_for_intent_not_used_in_nlu(
+    caplog: LogCaptureFixture, validator_under_test: Text
```
wochinge, line 231:
```suggestion
    caplog: LogCaptureFixture, validator_under_test: Validator
```

wochinge (approved):
💯 Congrats to your first contribution to Rasa Open Source! For context: I'd usually target 2.8.x for this change, but as we are very close to the 3.0 release and this issue is not very urgent it's in my opinion perfectly fine to merge to `main`.

data/test_config/config_defaults.yml
losterloh: Yes, that would be another way to do it; in general, I wonder if just writing over an existing file is really what we want. I'm assuming that in most cases - although not here - this would happen in a more interactive context anyways, where it might also make sense, for instance, to ask the user if they want these default comments to be added.

wochinge: In my opinion fine in _this_ case as we're just adding a comment and not actually changing something 🤔 Agree with you in general though 👍🏻

tests/test_validator.py
losterloh: Thanks! Fixed.

losterloh:
thanks @wochinge! I'll merge this then once CI runs through again after the latest small change (& rebase against `main`).

wochinge:
Congrats on beating the CI! 😁

PR #10168, opened 2021-11-10
couple model meta data with version check

### LINKED ISSUE #10108: 2.X -> 3.0 QA: Loading a 2.X model with 3.0 produces an unexpected file not found error.
Creating a model with rasa 2.8.X using `rasa init`, training it, and then attempting to load it with `rasa shell` after switching to a rasa 3.0 produces a file not found error:

`FileNotFoundException: Failed to read file, '/var/folders/4x/yjfd_jb15c55fy63yj53cjgm0000gn/T/tmphdo98uql/metadata.json' does not exist.`

**Steps to Reproduce**
1. Create a virtual environment and install rasa 2.8
2. Create another virtual environment and install rasa 3.0rc1
3. Using the rasa 2.8 venv, create a new project and train it
4. Using the rasa 3.0rc1 venv, attempt to run `rasa shell` on the project

ka-bu:
**Proposed changes**:
- closes #10108 


ka-bu:
I've changed the solution back and forth and ended up with this version that tightly couples the version check to the meta data because ...
- for now it seems we really do not want to do anything with old models and 
- models always have that version attached in their meta data (so that seems to be the place to handle this) and 
- hence it will always be checked automatically (like we can't forget an explicit call / have to create no wrapper function that makes sure this check is called for all storages) 

... and if we ever want to be able to load models with old meta data we can still easily make the check explicit. Wdyt?

rasa/engine/storage/storage.py
```diff
@@ -128,6 +139,25 @@ class ModelMetadata:
...
+            than the minimum compatible version
+        """
+        minimum_version = version.parse(MINIMUM_COMPATIBLE_VERSION)
+        model_version = version.parse(self.rasa_open_source_version)
+        if model_version < minimum_version:
+            raise UnsupportedModelError(
+                f"The model version is trained using Rasa Open Source {model_version} "
+                f"and is not compatible with your current installation "
+                f"({minimum_version}). "
+                f"This means that you either need to retrain your model "
```
wochinge, line 156: ... with Rasa Open Source {minimum_version} or higher ...?

wochinge (commented):
What if the model is a Rasa Open Source 2 model? In this case the unarchiving itself will fail as the metadata file isn't there.

ka-bu:
> What if the model is a Rasa Open Source 2 model? In this case the unarchiving itself will fail as the metadata file isn't there.

Thanks for the reminder 🤦‍♀️ . Hopefully added a fix for the actual bug now (to the 3.x version checks which I left as is).

rasa/exceptions.py
```diff
@@ -1,6 +1,33 @@
...
+    """Raised when a model is too old to be loaded."""
+
+    def __init__(self, model_version: Text) -> None:
+        """Instantiates a new exception indicating an unsupported model version.
+
+        Args:
+          model_version: the used model version that is not supported and triggered
+             this exception
+        """
+        self.model_version = model_version
```
wochinge, line 18: optional: should we use a `@dataclass`?

tests/engine/storage/test_local_model_storage.py
```diff
@@ -136,6 +165,66 @@ def test_create_model_package(
...
+    old_version = "0.0.1"
+    outdated_model_meta_data.rasa_open_source_version = old_version
+
+    # Package model - and inject the outdated model meta data
+    monkeypatch.setattr(
+        LocalModelStorage,
+        "_create_model_metadata",
+        lambda *args, **kwargs: outdated_model_meta_data,
+    )
+    with freezegun.freeze_time(trained_at):
```
wochinge, line 206: why do we need the freezegun?

rasa/engine/storage/local_model_storage.py
```diff
@@ -88,6 +89,18 @@ def _extract_archive_to_directory(
     ) -> None:
         with TarSafe.open(model_archive_path, mode="r:gz") as tar:
             tar.extractall(temporary_directory)
+        LocalModelStorage._assert_not_rasa2_archive(temporary_directory)
+
+    @staticmethod
+    def _assert_not_rasa2_archive(temporary_directory: Union[Text, Path],) -> None:
```
wochinge, line 95: I checked and it's actually always a `Path` 
```suggestion
    def _assert_not_rasa2_archive(temporary_directory: Path) -> None:
```

wochinge (approved):
💯 Nice solution!

tests/engine/storage/test_local_model_storage.py
ka-bu: nice observation 👍  has become redundant with the monkey patch

rasa/exceptions.py
ka-bu: I like 👍  done

PR #10170, opened 2021-11-10
changed the generator to load full tracker including all sessions + added markers changelog

### LINKED ISSUE #10113: Rasa 3.0.0rc1: No changelog for Success Markers
There should be a changelog entry for the new Success Markers feature

### Definition of Done ###
- [x] Changelog entry.

### LINKED ISSUE #10124: The result of markers evaluations doesn't seem to detect/include markers in all cases
### Rasa Open Source version

3.0rc1

### Rasa SDK version

_No response_

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

I was testing that a marker defined for each condition (`action_executed`, `intent_detected`, `slot_set`) is correctly extracted and accidentally found out that the result doesn't seem to contain all the markers that actually should be extracted.

I used the default bot (the one created with `rasa init`), and here is my `markers.yml`:
```
marker_mood:
 or:
   - intent_detected: mood_unhappy
   - intent_detected: mood_great
```
I had several conversations but the `extracted_markers.csv` contains only two markers:
```
sender_id,session_idx,marker_name,event_id,num_preceding_user_turns
345edc3430004f33ace142d33e327d43,0,marker_mood,7,1
5139e06d573a46f9825f19ec163b4960,0,marker_mood,7,1
```

The thing is my tracker store actually contains more than just two `mood_unhappy` or `mood_great` intents. It might also be that my understanding is wrong, but shouldn't all of them be extracted?
You can find my bot that also includes a database (`simple_tracker_store.db`) with all conversations in attachements [here](https://alwx.me/files/bot-success-markers-3.zip)

### Command / Request

_No response_

### Relevant log output

_No response_

aeshky:
**Proposed changes**:
- Changed the `MarkerTrackerLoader` to load the full tracker using `retrieve_full_tracker` instead of `retrieve`. 
Fixes #10124
- Added Changelog for markers project. Closes #10113

### DIFF
--- a/rasa/core/evaluation/marker_tracker_loader.py
+++ b/rasa/core/evaluation/marker_tracker_loader.py
@@ -96,4 +96,4 @@ class MarkerTrackerLoader:
 
         keys = self.strategy(stored_keys, self.count)
         for sender in keys:
-            yield self.tracker_store.retrieve(sender)
+            yield self.tracker_store.retrieve_full_tracker(sender)
--- a/rasa/core/evaluation/marker_tracker_loader.py
+++ b/rasa/core/evaluation/marker_tracker_loader.py
@@ -5,6 +5,10 @@ from typing import Any, Iterable, Iterator, List, Text, Optional
 from rasa.core.tracker_store import TrackerStore
 import rasa.shared.utils.io
 
+STRATEGY_ALL = "all"
+STRATEGY_FIRST_N = "first_n"
+STRATEGY_SAMPLE = "sample"
+
 
 def strategy_all(keys: List[Text], count: int) -> Iterable[Text]:
     """Selects all keys from the set of keys."""
--- a/tests/core/evaluation/test_marker_tracker_loader.py
+++ b/tests/core/evaluation/test_marker_tracker_loader.py
@@ -1,10 +1,11 @@
-from rasa.shared.core.events import UserUttered
+import pytest
+import os
+from rasa.shared.core.events import UserUttered, SessionStarted
 from rasa.shared.core.trackers import DialogueStateTracker
 from rasa.shared.core.domain import Domain
 from rasa.shared.exceptions import RasaException
 from rasa.core.evaluation.marker_tracker_loader import MarkerTrackerLoader
-from rasa.core.tracker_store import InMemoryTrackerStore, TrackerStore
-import pytest
+from rasa.core.tracker_store import InMemoryTrackerStore, TrackerStore, SQLTrackerStore
 
 
 @pytest.fixture
@@ -18,6 +19,38 @@ def marker_trackerstore() -> TrackerStore:
         store.save(tracker)
 
     return store
+
+
+@pytest.fixture
+def multi_session_tracker() -> [TrackerStore, str]:
+    """Sets up a TrackerStore with 1 tracker with 2 sessions."""
+    temp_db = "temp.db"
+    if os.path.isfile(temp_db):
+        os.remove(temp_db)
+    domain = Domain.empty()
+    store = SQLTrackerStore(domain, db=temp_db)
+    tracker = DialogueStateTracker("test123", None)
+    tracker.update_with_events(
+        [
+            UserUttered("0"),
+            UserUttered("1"),
+            SessionStarted(),
+            UserUttered("2"),
+            UserUttered("3"),
+        ],
+        domain,
+    )
+    store.save(tracker)
+    return store, temp_db
+
+
+def test_load_sessions(multi_session_tracker):
+    """Tests loading a tracker with multiple sessions."""
+    loader = MarkerTrackerLoader(multi_session_tracker[0], "all")
+    result = list(loader.load())
+    assert len(result) == 1
+    assert len(result[0].events) == 5
+    os.remove(multi_session_tracker[1])
 
 
 def test_load_sample(marker_trackerstore: TrackerStore):
--- a/tests/core/evaluation/test_marker_tracker_loader.py
+++ b/tests/core/evaluation/test_marker_tracker_loader.py
@@ -22,11 +22,9 @@ def marker_trackerstore() -> TrackerStore:
 
 
 @pytest.fixture
-def multi_session_tracker() -> [TrackerStore, str]:
+def multi_session_tracker() -> TrackerStore:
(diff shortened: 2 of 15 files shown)


tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,36 @@ def marker_trackerstore() -> TrackerStore:
...
+            UserUttered("1"),
+            SessionStarted(),
+            UserUttered("2"),
+            UserUttered("3"),
+        ],
+        domain,
+    )
+    store.save(tracker)
+    yield store
+    os.remove(temp_db)
```
ka-bu, line 43: If this test is interrupted, the `temp.db` which is created in the current directory won't be removed. For this reason, other tests use the `tmp_path` fixture which will take care that the temporary file does not appear in the current directory  and that it is removed later

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,36 @@ def marker_trackerstore() -> TrackerStore:
     return store
 
 
+@pytest.fixture
+def multi_session_tracker() -> TrackerStore:
```
ka-bu, line 25: There's no need to put this in a fixture since it's only used in one function

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,36 @@ def marker_trackerstore() -> TrackerStore:
...
+            UserUttered("0"),
+            UserUttered("1"),
+            SessionStarted(),
+            UserUttered("2"),
+            UserUttered("3"),
+        ],
+        domain,
+    )
+    store.save(tracker)
+    yield store
```
ka-bu, line 42: The "yield" here means that this function is returning a generator which would mean we should change the output type here if we kept this approach. But the approach with the temp path fixture is much cleaner and so there should be no need for a yield anymore anyway :)

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,36 @@ def marker_trackerstore() -> TrackerStore:
...
+    os.remove(temp_db)
+
+
+def test_load_sessions(multi_session_tracker):
```
ka-bu, line 46: (Type was missing here -- but we should move the stuff that is in the fixture into this function anyway :)

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,36 @@ def marker_trackerstore() -> TrackerStore:
...
+    """Tests loading a tracker with multiple sessions."""
+    loader = MarkerTrackerLoader(multi_session_tracker, "all")
+    result = list(loader.load())
+    assert len(result) == 1
+    assert len(result[0].events) == 5
```
ka-bu, line 51: When we move that store creation code into this function, we can also pull out the event creation so that we can use something like `len(events)` here instead of hard coding the 5

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,36 @@ def marker_trackerstore() -> TrackerStore:
...
+        domain,
+    )
+    store.save(tracker)
+    yield store
+    os.remove(temp_db)
+
+
+def test_load_sessions(multi_session_tracker):
+    """Tests loading a tracker with multiple sessions."""
+    loader = MarkerTrackerLoader(multi_session_tracker, "all")
```
ka-bu, line 48: Do we have a constant for the "all"? If not, we should add one

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -96,4 +96,4 @@ def load(self) -> Iterator[Optional[DialogueStateTracker]]:
 
         keys = self.strategy(stored_keys, self.count)
         for sender in keys:
-            yield self.tracker_store.retrieve(sender)
+            yield self.tracker_store.retrieve_full_tracker(sender)
```
ka-bu, line 99: Ah, this is the only thing from the first version which we didn't check / adapt to work with multiple sessions. Looks like feature complete now 🥳

ka-bu (commented):
🚀

tests/core/evaluation/test_marker_tracker_loader.py
aeshky: Do you have a style preference for constants? I know we usually place them in a separate file, but since we only have three I put them in `rasa/core/evaluation/marker_tracker_loader.py`.

rasa/core/evaluation/marker_tracker_loader.py
```diff
@@ -96,4 +100,4 @@ def load(self) -> Iterator[Optional[DialogueStateTracker]]:
...
```
usc-m, line 103: Good spot 🎉

usc-m (approved):
LGTM

tests/core/evaluation/test_marker_tracker_loader.py
ka-bu: that's fine 👍

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,29 @@ def marker_trackerstore() -> TrackerStore:
     return store
 
 
+def test_load_sessions(tmp_path):
```
ka-bu, line 24:
```suggestion
def test_load_sessions(tmp_path : Path):
```

aeshky:
Thanks both for a speedy review 🚀

tests/core/evaluation/test_marker_tracker_loader.py
```diff
@@ -20,6 +21,29 @@ def marker_trackerstore() -> TrackerStore:
...
+    """Tests loading a tracker with multiple sessions."""
+    domain = Domain.empty()
+    store = SQLTrackerStore(domain, db=os.path.join(tmp_path, "temp.db"))
```
ka-bu, line 27: btw, with paths, you can also do things like 
```suggestion
    store = SQLTrackerStore(domain, db= tmp_path / "temp.db")
```
(https://docs.python.org/3/library/pathlib.html#basic-use)

PR #10177, opened 2021-11-11
Parallelize regression test jobs

markus-hinsche:
**Motivation:** Running regression tests takes a lot of time if it runs for many configs and datasets, e.g. 13 hours for [on-schedule](https://github.com/RasaHQ/rasa/actions/workflows/ci-model-regression-on-schedule.yml). We can parallelize the sequential jobs which are fairly independent. 

**Proposed changes**:
- Use github actions strategy: `max-parallel: 3` by default
- Combine reports of all job into a single report.json
- Send metrics to Segment and Datadog while running the jobs in parallel
- Refactor: DRY the `add_comment_results_cpu` and `add_comment_results_gpu` jobs


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["Carbon Bot", "financial-demo"]
   config: ["Sparse + DIET(bow) + ResponseSelector(bow)", "BERT + DIET(seq) + ResponseSelector(t2t)", "Spacy + DIET(seq) + ResponseSelector(t2t)", "Sparse + BERT + DIET(seq) + ResponseSelector(t2t)", "Sparse + DIET(seq) + ResponseSelector(t2t)", "Sparse + Spacy + DIET(seq) + ResponseSelector(t2t)"]
```

.github/workflows/ci-model-regression-on-schedule.yml
```diff
@@ -335,15 +335,14 @@ jobs:
         run: |-
           poetry run pip install analytics-python
           poetry run python .github/scripts/mr_publish_results.py
-          poetry run python .github/scripts/mr_generate_summary.py
           cat $SUMMARY_FILE
 
       - name: Upload an artifact with the report
```
tczekajlo, line 340: It seems that the `Download artifact` step is not needed anymore (in both workflows)

.github/workflows/ci-model-regression.yml
```diff
@@ -649,102 +648,68 @@ jobs:
...
-  add_comment_results_gpu:
-    name: Add a comment with the results
+  combine_reports:
+    name: Combine reports
     runs-on: ubuntu-latest
     needs:
+      - model_regression_test_cpu
       - model_regression_test_gpu
-
+    if: always() && (needs.model_regression_test_cpu.result == 'success' || needs.model_regression_test_gpu.result == 'success')
```
tczekajlo, line 671: I think `always()` is not needed. if a result for the `model_regression_test_cpu` is not successful then it will be skipped anyway.

tczekajlo (dismissed):

.github/workflows/ci-model-regression.yml
markus-hinsche: I think it is needed, because:

I would like to run `combine_reports` if either `model_regression_test_cpu` or `model_regression_test_gpu` is successful.

I added `model_regression_test_cpu` to the `needs` section. We expect either `model_regression_test_cpu` or `model_regression_test_gpu` to run. 
By default, I think `needs` works in a way that both would have to be successful (https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idneeds), but in our case only one will be successful and the other will be skipped. 

Similar discussion: https://github.com/actions/runner/issues/491#issuecomment-850884422

RASADSA (approved):
LGT 🚀

RASADSA (approved):

PR #10181, opened 2021-11-11
Fix mapping from CLI arg to config

### LINKED ISSUE #10132: `--augmentation` parameter overwritten with default value
### Rasa Open Source version

3.0.0rc1

### Rasa SDK version

_No response_

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

The `--augmentation` parameter is not making it into the `TrainingDataGenerator` component but instead seems to be overwritten with the default value.

Here in `model_training.py` it is correctly passed to `train`:
Screenshot 2021-11-10 at 09 47 10

Here in `TrainingTrackerProvider.provide` it has been overwritten with the default value of 50:
Screenshot 2021-11-10 at 09 50 00

### Command / Request

```shell
rasa train --augmentation 200
```

### Relevant log output

_No response_

usc-m:
**Proposed changes**:
- Closes #10132 
  - Problem was CLI arguments being mapped to a key not used by the `TrainingTrackerProvider` (`augmentation` instead of `augmentation_factor`), so when the component config (generated from the CLI args) was merged with the default config, the default value for `augmentation_factor` was never replaced


rasa/engine/recipes/default_recipe.py
```diff
@@ -477,7 +477,7 @@ def _add_core_train_nodes(
                 config_key: cli_parameters[param]
                 for param, config_key in {
```
joejuzl, line 478: Can this just be a list now as the param and config_key are always the same?

joejuzl (changes requested):

rasa/engine/recipes/default_recipe.py
usc-m: Yep, was unsure if I should change it or not in case we needed to do something later - will do so now

joejuzl (approved):
🚀

joejuzl (changes requested):
Actually just thought... we should probably have a test for this. Especially as we missed it before.

usc-m:
So me and Tobi had a think about this yesterday and it's a bit of a nightmare to test (since IMO the problem is that CLI params are packed into a dict and we have no way of validating that dicts structure through all the layers, so we can test any one layer but that doesn't guarantee we actually respect the structure at the other layers). I could do a unit test for just this bit of the recipe to make sure that if the dict is constructed the way the CLI constructs it now, it maps correctly, but I'm not sure how we'd test to make sure the mapping isn't actually broken in future?

joejuzl:
Is it possible to add a test to `tests/cli/test_rasa_train.py` where we actually pass the `--augmentation` flag and then spy on the `TrainingTrackerProvider` to see it get's the value passed to its `__init__`?

If it's too time consuming don't worry though.

usc-m:
I think that would be difficult because most of the tests there will invoke the CLI on the shell, where we can't mock things very easily (had a similar problem with markers recently) - so unless the value of the parameter was logged somewhere we would probably have a hard time intercepting it. I could add a logging line to facilitate this though if that sounds good?

koernerfelicia:
Was thinking about testing as well, this seems like such a sneaky bug.. only reason I found it is because I was stepping through the code. I wonder if there are other instances like these where it might not be clear if the arguments aren't making it to the component? Not an urgent thing atm, but in general would be nice to have tests for all the components

joejuzl:
Maybe a compromise for now is a unit test for the recipe like you suggested @usc-m - just to stop a regression for this specific bug.

usc-m:
I think long run we might want to think about refactoring this somehow to get the type checker to make sure things are mapped properly, and then also maybe think about a better testing strategy for the CLI overall (like some harness that invokes `main` directly with a parameter string and then we can mock things and check the internals - alongside some shell execution tests)

usc-m:
So when I was looking for where to add a new test, I noticed `test_tracker_generator_parameter_interpolation` actually tested the old functionality (with the incorrect mapping - which as a result was failing on this branch), so I've updated it and I think that should catch regressions of this kind now

joejuzl (approved):
💯

Issue #10142, opened 2021-11-10
Training is broken for windows with py38


joejuzl:
When training with the default `init` project on windows with python38 we get this error:
```
[...]

PR #10182, opened 2021-11-11
fix windows 3.8 issue

wochinge:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/10142
- check origin type instead of type annotation (as we are ignoring the contents of the list)

### DIFF
--- a/rasa/engine/validation.py
+++ b/rasa/engine/validation.py
@@ -340,7 +340,7 @@ def _validate_run_fn_return_type(
         )
 
     # TODO: Handle forward references here
-    if typing_utils.issubtype(return_type, typing.List):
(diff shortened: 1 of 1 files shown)


joejuzl:
Do we know why this was only broken on windows 3.8.0?

wochinge:
I've checked out the issue tracker a bit but couldn't find anything. In my opinion could be some changes to how `isinstance` works with `TypeVars`. The issue is that `isinstance(typing.List[TrackerWithCachedStates], typing.List)` was wrong on older 3.8 versions, while the check was `True` on newer versions.

joejuzl (approved):
What a pesky bug! 
🚀

Issue #10183, opened 2021-11-11
Add model name to Tensorboard log dir


cep-ter:
### What problem are you trying to solve?

Previously Tensorboad log dir would automatically create a separate directory for each model we trained using ‍‍Rasa. (The image below shows that) source ([https://rasa.com/blog/tensorboard-in-rasa/](https://rasa.com/blog/tensorboard-in-rasa/)) [<span class="image-wrap" style="">![](https://user-images.githubusercontent.com/55178902/141369796-fa9df46d-846e-4bf5-a92a-2b6bab5b0012.png)</span>](https://user-images.githubusercontent.com/55178902/141369796-fa9df46d-846e-4bf5-a92a-2b6bab5b0012.png)

But currently, all models are saved in the same directory and we can't compare models. (Show all results from multiple models in the same chart which is wrong and Tensorboard also throws a warning about it).

### What's your suggested solution?

As Rasa was previously doing, we can add model-name (class_name+current_time) as a subdirectory to the Tensorboard log dir. This way we can compare models and Tensorboard logs without the need to change the Tensorboard log dir before each train manually. ![](/images/icons/emoticons/smile.png)

### Examples (if relevant)

 _No response_ 

### Is anything blocking this from being implemented? (if relevant)

 _No response_ 

### Definition of Done

 _No response_

PR #10184, opened 2021-11-11
Add class name and current time to tensorboard log dir

cep-ter:
**Proposed changes**:
As I explained here #10183, I add class name and current time to tensorboard log dir.
- ...


samsucik:
@cep-ter thanks for proposing this! Sorry for the delay, I've now looked properly at what you're proposing and also why (and since when) we don't log TB information separately for each model. I'll have a think about the way you're solving the issue and will push this PR a step further 🙂

samsucik:
Right, so it looks like we removed the model- and run-specific grouping of TensorBoard logs in [this PR](https://github.com/RasaHQ/rasa/pull/7438/files?authenticity_token=pdIfT9BqnirUluRLJatjJ1mji%2Bdy%2FHvmHQ%2FfMNhC4lPCv9eaxqvuQdmUZa3qDEErWl5LOdzI34mdGpFP8DwqYQ%3D%3D&file-filters%5B%5D=.md&file-filters%5B%5D=.mdx&file-filters%5B%5D=.py#diff-1b0d53e90a7f1dcca1113cda22c58c98685004f8dfc7ba3d4e5bf9cee78e6a19L122) and dump all logs into one directory. If I'm honest, I don't see why. I tried out the current approach and @cep-ter is right that it makes it hard to compare models or even just know which logs come from where. I think the reason for these problems is that, for a long time after the big e2e refactor, TensorBoard didn't really work with `rasa`, so we couldn't test things and make sure that the logging was done in the most sensible way (@dakshvar22 correct me if I remember something incorrectly).

In any case, I agree that we should separate TB logs by model name and timestamp like we did previously and like @cep-ter is proposing. However, I think this should be implemented in only one place -- `rasa.utils.train_utils.create_common_callbacks` -- and not in individual models like DIET or TED. @cep-ter could you make this change?

I've also dropped this PR into our Research inbox so that the effort is visible, someone takes this on fully and together with @cep-ter drives this to completion 🙂

dakshvar22:
>  I think the reason for these problems is that, for a long time after the big e2e refactor, TensorBoard didn't really work with rasa, so we couldn't test things and make sure that the logging was done in the most sensible way

The graph logging didn't work and might not be working even now but the metric logging should have worked always. Anyways, I agree that segregating the logs with timestamp is much better.

rasa/core/policies/ted_policy.py
```diff
@@ -664,9 +665,15 @@ def run_training(
             self.config[EVAL_NUM_EXAMPLES],
             self.config[RANDOM_SEED],
         )
+
+        current_time = time.strftime("%Y%m%d-%H%M%S")
+        class_name = self.__class__.__name__
+        tensorboard_log_dir = (
+            f"{self.component_config[TENSORBOARD_LOG_DIR]}/{class_name}/{current_time}/"
+        )
```
dakshvar22, line 673: Since the same logic is used across multiple components, lets pull out this logic into a separate function inside `rasa/utils/train_utils.py` which would accept the path to log directory  specified inside the component and the name of the class invoking that function and return the complete path to be used, i.e. `self.component_config[TENSORBOARD_LOG_DIR]}/{class_name}/{current_time}`.

This function can then be invoked by respective components inside their `train` functions to get the correct path that should be passed to `create_common_callback` function.

Also, please use `os.path.join` to construct the complete path. 🙏

dakshvar22 (changes requested):
Thanks for opening up this enhancement request ✨  Requested a change to reduce code duplication. Also, some rework of tests will be needed:
1. Add a unit test for testing the new function you'll create. You can add this test to `tests/utils/test_train_utils.py`.
2. Fix the test `test_train_tensorboard_logging` inside tests/nlu/classifiers/test_diet_classifier.py`. Add a similar test for response selector in `tests/nlu/selectors/test_response_selector.py`, for ted policy in `tests/core/policies/test_ted_policy.py` and for unexpected_intent_policy in `tests/core/policies/test_unexpected_intent_policy.py`. This would really help in testing out this enhancement properly for all components where its involved 🙏

m-vdb:
Hi there! Thanks for your contribution. As the PR has been stale for a while, I'm closing it. Please feel free to reopen an issue / PR if the problem still occurs on latest rasa versions.

PR #10189, opened 2021-11-12
update `/status` response and docs

### LINKED ISSUE #10130: API QA: `/status` endpoint returns a response body that's not consistent with the docs
**Description of the problem**:
Sending a GET request to `/status` endpoint results in this response body that doesn't match the [docs](https://rasa.com/docs/rasa/pages/http-api#operation/getStatus):
```
{
    "model_file": "models",
    "fingerprint": "5e0a7a1433b84d53843518203cef922c",
    "num_active_training_jobs": 0
}
```
- Used a rasa init project for this test case.

carlad:
Closes #10130 

**Proposed changes**:
- The ensures the response payload has the correct key/values.
- Updates the docs to conform to this.

- [x] updated the changelog


alwx (approved):
Looks good, thanks for fixing this!

joejuzl (changes requested):
looks good! 
I think we should add a changelog too though as it may affect users.

carlad:
I've made a change to how the filename is being retrieved that differs from your suggestion @joejuzl .
The `model_tar` returns the full path to the file, not just the filename, so even though the tests were passing, locally the DEFAULT_MODELS_PATH, `models` was being returned in the payload.

Instead I added a `model_filename` attribute to the `Processor`, which uses the `model_path` to `get_latest_model` and then derive the `os.path.basename` from that.

The tests still passes, and I'm hoping that fixes things locally...

tayfun (approved):

Issue #10136, opened 2021-11-10
`rasa data migrate` doesn't clean up if migration fails


wochinge:
### Rasa Open Source version

3.0.0rc1

### Python version

3.8

### What operating system are you using?

OSX

### What happened?

If run `rasa data migrate`  and the migration fails for some reason then you still have an `original_domain` and an empty `new_domain` directory. This is especially problematic if you fix your domain and try to re-run `rasa data migrate`  as this leads to this error:

```
File "C:\Users\User\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.8_qbz5n2kfra8p0\LocalCache\local-packages\Python38\site-packages\rasa\core\migrate.py", line 194, in migrate_domain_format
    backup_dir.mkdir()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.2800.0_x64__qbz5n2kfra8p0\lib\pathlib.py", line 1288, in mkdir
    self._accessor.mkdir(self, mode)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'original_domain
```

### Command / Request

```shell
`rasa data migrate`
```

PR #10191, opened 2021-11-12
`rasa data migrate` doesn't clean up if migration fails

alwx:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/10136


tests/core/test_migrate.py
```diff
@@ -764,3 +764,48 @@ def test_migrate_domain_raises_for_non_domain_files(tmp_path: Path):
...
+        """,
+        "slots_two.yml",
+    )
+
+    with pytest.raises(
+        RasaException,
+        match="Domain files with multiple 'slots' sections were provided.",
+    ):
+        rasa.core.migrate.migrate_domain_format(domain_dir, domain_dir)
+
```
ancalita, line 806: I think it's sufficient to only have one `pytest.raises` and then assert that the backup directory doesn't exist and that any migrated (eligible) domain files were cleaned up from the provided `out` path.

rasa/core/migrate.py
```diff
@@ -207,48 +250,11 @@ def migrate_domain_format(
...
-                f"successful migration."
-            )
-
-        original_domain.update(
-            {KEY_SLOTS: slots, KEY_FORMS: forms, KEY_ENTITIES: entities}
-        )
+        try:
+            original_domain = _migrate_domain_files(domain_file, backup_dir, out_file)
+        except Exception as e:
+            shutil.rmtree(backup_dir)
```
ancalita, line 256: If the out_file was the created directory `new_domain`, this should be removed too, as well as any migrated domain files (for example, a `responses.yml` file).

ancalita (commented):
Nice one ⚡ 
I added 2 suggestions based on the issue description that any partially migrated files as well as back-up or `new_domain` directory should be removed if this command fails.

tests/core/test_migrate.py
ancalita: Oh I just realised why you're attempting the migration twice, to make sure that it doesn't fail with `FileExistsError` - I sometimes process things with a delay 😅 But it would be good to add those additional asserts just for the sake of clarity.

tests/core/test_migrate.py
```diff
@@ -764,3 +765,53 @@ def test_migrate_domain_raises_for_non_domain_files(tmp_path: Path):
...
+        rasa.core.migrate.migrate_domain_format(domain_dir, migrated_domain_dir)
+
+    with pytest.raises(
+        RasaException,
+        match="Domain files with multiple 'slots' sections were provided.",
+    ):
+        rasa.core.migrate.migrate_domain_format(domain_dir, migrated_domain_dir)
+
+    current_dir = domain_dir.parent
+    assert not os.path.exists(current_dir / "original_domain")
```
ancalita, line 816: Only one tiny comment here: I'd use the pathlib `Path.exists` here rather than `os` library, because I think there's an open issue for replacing `os` with `pathlib` which some community members are working on.

tests/core/test_migrate.py
```diff
@@ -764,3 +765,53 @@ def test_migrate_domain_raises_for_non_domain_files(tmp_path: Path):
...
+    assert not os.path.exists(current_dir / "new_domain")
```
ancalita, line 817: Since you provided `domain2` dir as out_path, I think `new_domain` never gets created. One thought here should we get also get rid of any migrated files in out_path dir if it's provided in case migration fails?

ancalita (commented):
Left one suggestion and another idea that we might also have to support during clean-up.

tests/core/test_migrate.py
alwx: Updated the test to check for multiple scenarios!

ancalita: Nice one! It's much clearer, but I have two follow-up q's:

- in the case of provided `migrated_domain_dir`, does the whole folder get removed instead of only its contents/files?
Removing the whole directory might be annoying for a user who might have created a specific location for this.

- in the case the default `domain_dir` is provided as `out_path`, what persists in this folder finally, only the original files I guess, meaning nothing changes there? 🤔

alwx: 1. Updated it and also updated the test to check if the dir was not removed (`assert Path.exists(migrated_domain_dir)` after the migration)
2. I guess so. I didn't touch this logic tbh, don't think it makes sense to do it as part of this issue.

tests/core/test_migrate.py
```diff
@@ -764,3 +764,75 @@ def test_migrate_domain_raises_for_non_domain_files(tmp_path: Path):
...
+
+    with pytest.raises(
+        RasaException,
+        match="Domain files with multiple 'slots' sections were provided.",
+    ):
+        rasa.core.migrate.migrate_domain_format(domain_dir, migrated_domain_dir)
+
+    # check if the cleanup was done successfully after
+    # the `migrated_domain_dir` was created
+    assert Path.exists(migrated_domain_dir)
```
ancalita, line 836: Is there a simple way also to assert that `migrated_domain_dir` is empty after a failed migration?

alwx: Yep, done that!

ancalita: It could be worth testing this by having a 3rd file in `domain_dir` that doesn't contain slots or forms (maybe just intents, or just responses) to see if that file gets migrated and removed after failure.

ancalita (approved):
Yipeee 💯

Issue #10192, opened 2021-11-12
Markers: the first session in trackers sometimes contains only one event `action: action_session_start` which produces noisy marker statistics.


aeshky:
**Problem:**
After fixing the `MarkerTrackerLoader` to load full trackers including all sessions, I noticed that the first session usually only contains a single event `action: action_session_start`, followed by `session_started: null` which we use to create new sessions. 

This produces noisy statistics. For example, I created three dialogues using `rasa shell`. The statistics show 6 sessions (which also messes up the percentages), and the per session statistics contain events for these noisy sessions with lots of nans. 

**Example:**
[trackerstore.csv](https://github.com/RasaHQ/rasa/files/7528467/trackerstore.csv)
[stats-per-session.csv](https://github.com/RasaHQ/rasa/files/7528451/stats-per-session.csv)
[stats-overall.csv](https://github.com/RasaHQ/rasa/files/7528453/stats-overall.csv)
[extracted.csv](https://github.com/RasaHQ/rasa/files/7528454/extracted.csv)

**Solution:**
When splitting sessions, consider both `action: action_session_start` and `session_started: null` as part of the same session. 
Otherwise, discard sessions containing only a single event equal to `action: action_session_start`.

aeshky:
**Example:**
[trackerstore.csv](https://github.com/RasaHQ/rasa/files/7528467/trackerstore.csv)
[stats-per-session.csv](https://github.com/RasaHQ/rasa/files/7528451/stats-per-session.csv)
[stats-overall.csv](https://github.com/RasaHQ/rasa/files/7528453/stats-overall.csv)
[extracted.csv](https://github.com/RasaHQ/rasa/files/7528454/extracted.csv)

aeshky:
@usc-m @ka-bu 

I can look into the this and submit a PR/fix. I can also address other small Marker issues at the same time (renaming a column and chaning CLI). WDYT?

ka-bu:
Just to avoid confusion: The “nan”s in the per session results are not noisy per se. If there are no events, then “nan” is only reasonable default (same as used in cases where events exist but no relevant event for the respective marker exist). -- The requirement that had been extracted from the initial implementation was to output "summary statistics" for every session, even if no relevant event was extracted (this case will still exist and will still produce some "nans"). 

The statistic that suffering from the noise caused by empty sessions is the percentage of sessions where a marker applies at least once, since the empty sessions count towards the total number of sessions.

PR #10195, opened 2021-11-12
Marker fixes

aeshky:
**Proposed changes**:
- bug fix. closes https://github.com/RasaHQ/rasa/issues/10192
- changed the column names in marker output file
- changed the stat names to past tense
- referenced the Marker docs in the error messages

### DIFF
--- a/rasa/core/evaluation/marker_base.py
+++ b/rasa/core/evaluation/marker_base.py
@@ -342,9 +342,9 @@ class Marker(ABC):
         if len(session_start_indices) == 0:
             return [(events, 0)]
         sessions_and_start_indices: List[Tuple[List[Event], int]] = []
-        for session_idx in range(len(session_start_indices)-1):
+        for session_idx in range(len(session_start_indices) - 1):
             start_idx = session_start_indices[session_idx]
-            end_idx = session_start_indices[session_idx+1]
+            end_idx = session_start_indices[session_idx + 1]
             session = [events[idx] for idx in range(start_idx, end_idx)]
             sessions_and_start_indices.append((session, start_idx))
         last_session = [
--- a/rasa/core/evaluation/marker_base.py
+++ b/rasa/core/evaluation/marker_base.py
@@ -653,8 +653,8 @@ class Marker(ABC):
                 [
                     "sender_id",
                     "session_idx",
-                    "marker_name",
-                    "event_id",
+                    "marker",
+                    "event_idx",
                     "num_preceding_user_turns",
                 ]
             )
--- a/rasa/core/evaluation/marker_base.py
+++ b/rasa/core/evaluation/marker_base.py
@@ -309,7 +309,7 @@ class Marker(ABC):
 
         extracted_markers: List[Dict[Text, List[EventMetaData]]] = []
         for session, start_idx in sessions_and_start_indices:
-            if len(session) == 0:  # skip trailing empty sessions
+            if len(session) == 0:  # skip empty sessions
                 continue
             # track all events and collect meta data per time step
             meta_data = self._track_all_and_collect_meta_data(
--- a/rasa/core/evaluation/marker_base.py
+++ b/rasa/core/evaluation/marker_base.py
@@ -308,6 +308,8 @@ class Marker(ABC):
 
         extracted_markers: List[Dict[Text, List[EventMetaData]]] = []
         for session, start_idx in sessions_and_start_indices:
+            if len(session) == 0:  # skip trailing empty sessions
+                continue
             # track all events and collect meta data per time step
             meta_data = self._track_all_and_collect_meta_data(
                 events=session, event_idx_offset=start_idx
--- a/rasa/core/evaluation/marker_base.py
+++ b/rasa/core/evaluation/marker_base.py
@@ -342,14 +342,11 @@ class Marker(ABC):
         if len(session_start_indices) == 0:
             return [(events, 0)]
         sessions_and_start_indices: List[Tuple[List[Event], int]] = []
-        for session_idx in range(len(session_start_indices)):
-            start_idx = (
-                session_start_indices[session_idx - 1] if (session_idx > 0) else 0
-            )
-            end_idx = session_start_indices[session_idx]
+        for session_idx in range(len(session_start_indices)-1):
+            start_idx = session_start_indices[session_idx]
+            end_idx = session_start_indices[session_idx+1]
             session = [events[idx] for idx in range(start_idx, end_idx)]
-            if len(session) > 0:
-                sessions_and_start_indices.append((session, start_idx))
+            sessions_and_start_indices.append((session, start_idx))
         last_session = [
             events[idx] for idx in range(session_start_indices[-1], len(events))
         ]
--- a/rasa/core/evaluation/marker_base.py
+++ b/rasa/core/evaluation/marker_base.py
@@ -309,8 +309,6 @@ class Marker(ABC):
 
         extracted_markers: List[Dict[Text, List[EventMetaData]]] = []
         for session, start_idx in sessions_and_start_indices:
-            if len(session) == 0:  # skip empty sessions
-                continue
             # track all events and collect meta data per time step
             meta_data = self._track_all_and_collect_meta_data(
                 events=session, event_idx_offset=start_idx
@@ -350,7 +348,8 @@ class Marker(ABC):
             )
             end_idx = session_start_indices[session_idx]
             session = [events[idx] for idx in range(start_idx, end_idx)]
-            sessions_and_start_indices.append((session, start_idx))
+            if len(session) > 0:
+                sessions_and_start_indices.append((session, start_idx))
         last_session = [
             events[idx] for idx in range(session_start_indices[-1], len(events))
         ]
--- a/rasa/core/evaluation/marker_base.py
+++ b/rasa/core/evaluation/marker_base.py
@@ -30,6 +30,7 @@ from rasa import telemetry
 from rasa.shared.core.domain import Domain
 from rasa.shared.core.trackers import DialogueStateTracker
 from rasa.utils.io import WriteRow
+from rasa.shared.constants import DOCS_URL_MARKERS
 
 import logging
 import csv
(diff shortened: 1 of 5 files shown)


rasa/core/evaluation/marker_base.py
```diff
@@ -308,6 +309,8 @@ def evaluate_events(
 
         extracted_markers: List[Dict[Text, List[EventMetaData]]] = []
         for session, start_idx in sessions_and_start_indices:
+            if len(session) == 0:  # skip trailing empty sessions
+                continue
```
ka-bu, line 313:
```suggestion
            if len(session) <= 1:  # skip trailing empty sessions
                continue
```

The check with `len(session)==0` does not solve the problem described in the issue: The `_split_sessions` method partitions the given list of events such that each partition starts with an `ActionExecuted` event with the name `ACTION_SESSION_START_NAME`. 

Hence, the `session` list here will be empty if and only if the tracker contains no events at all -- and it contains exactly 1 element if and only if the session only consist of the `ActionExecuted` described before.

By checking for `<=1` we'll cover both cases, i.e. that the tracker is completely empty (which probably cannot even happen, but doesn't hurt) and that the session is effectively empty.

ka-bu: Guess it would be good to also add a test for this :)

ka-bu (changes requested):
The session length check doesn't solve the problem :/ -- It will only have an effect if there are no events at all in the tracker but not if a session is practically empty.

rasa/core/evaluation/marker_base.py
aeshky: Sure I'll add a test 👍 
I tested it manually and it works. For some reason that session with the one event that I described becomes a session with zero events. I'll look into it to find out why + add a test.

ka-bu: 1. No, the above cannot work -- Please read the comment above carefully and let me know if you have any questions
2.  I overlooked the "followed by session_started: null."  => Means even the approach with "<=1" won't work properly since sessions containing only the `ActionExecuted` with `ACTION_SESSION_START_NAME` name and that session restart event will contain those 2 events. => We'll have to check what is in the list of events called `session` in the code snippet above.

tests/core/evaluation/test_marker.py
```diff
@@ -728,3 +728,17 @@ def test_marker_from_path_raises(
...
+
+    events = [
+        ActionExecuted("action_session_start"),
+        SessionStarted(),
+        UserUttered(intent={"name": "this-intent"}),
+    ]
+    sessions = Marker._split_sessions(events)
+    assert len(sessions) == 2
+    assert len(sessions[0]) == 0
+    assert len(sessions[1]) == len(events)
```
aeshky, line 744: I was in the process of testing the solution, but realised that in the original issue I didn't describe it correctly. It looks like `_split_session` creates two sessions, one empty, and one with all the other events including `action exectued: action_start_session`. 
Will look into `_split_session` now to see why this happens, and maybe the fix should be inside this function.

rasa/core/evaluation/marker_base.py
aeshky: > No, the above cannot work

Sorry, let me clarify. In the original issue I linked an example input and output. For that example, adding `if len(session) == 0:` gave me the output that I expect.

My original description of the issue was a guess about the problem. But after debugging further I realised that `split_sessions` is returning one empty session, then another session with remaining events. I added a test to illustrate the problem. The PR is not ready to re-review since I haven't fixed the problem, but the test should help illustrate the issue. I'll keep looking into it. 

Sorry this won't be fixed for RC2, but I'll fix it for RC3.

tests/core/evaluation/test_marker.py
aeshky: To clarify `sessions` should contain only one session of `len(events)`. So the current assertion is just a placeholder while I fix the problem.

ka-bu: > Will look into _split_session now to see why this happens

As already explained above:   The `_split_sessions` method partitions the given list of events such that each partition starts with an ActionExecuted event with the name `ACTION_SESSION_START_NAME`.

The motivation for this approach was the way sessions are split by boteval [(slack)](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1635243635080800).  -- So I don't understand why the above example does not start with sessions start event. Can you explain why you constructed the example like this?

Oh, but I think I found a severe bug now: [this](https://rasa.com/docs/rasa/default-actions/#action_session_start) indicates that the slots should be carried over from previous sessions. At the moment, we ignore those slots - while the bot would actually remember those. Will open an issue for that in a bit. @tttthomasssss , FYI

aeshky: >So I don't understand why the above example does not start with sessions start event. Can you explain why you constructed the example like this?

I didn't manually construct the examples, used `rasa shell` and had a normal dialogue with the bot. This resulted in: 
```
ActionExecuted("action_session_start")
SessionStarted()
ActionExecuted("action_listen")
...
```

>As already explained above: The _split_sessions method partitions the given list of events such that each partition starts with an ActionExecuted event with the name ACTION_SESSION_START_NAME.

Sure, but it creates one empty session first. Why is that?

rasa/core/evaluation/marker_base.py
```diff
@@ -347,7 +348,8 @@ def _split_sessions(events: List[Event]) -> List[Tuple[List[Event], int]]:
             )
             end_idx = session_start_indices[session_idx]
             session = [events[idx] for idx in range(start_idx, end_idx)]
-            sessions_and_start_indices.append((session, start_idx))
+            if len(session) > 0:
+                sessions_and_start_indices.append((session, start_idx))
```
aeshky, line 352: @ka-bu Are you happy with this fix or do you think there's a better way to do it?

tests/core/evaluation/test_marker.py
ka-bu: Ah, ok, I overlooked that the first action has an "action_session_start" - Please use constants instead of strings (`ACTION_SESSION_START_NAME`). That makes examples easier to parse and ensures consistency / bugs due to typos in the tests.

So the initial issue was just this and not really empty sessions, yes? 

The issue  here is with the first index being interpreted as a session end index. We just have to adapt the parameter in the range there.

rasa/core/evaluation/marker_base.py
ka-bu: It would be better to change the parameter of the range and refactor variable names a bit to reflect - and prepend session start events to the tests. The current implementation assumes that a session start is issued when a new conversation starts but doesn't take into account the first one starts with that as well. I can just add that to this branch if that's ok?

aeshky: Yes no problem, if you want to fix it differently then go for it.

tests/core/evaluation/test_marker.py
aeshky: Yes, I initially thought it was a different problem when I wrote the issue. Sounds like we're on the same page now.

ka-bu:
@usc-m  Could you double-check / review please?

tests/core/evaluation/test_marker.py
ka-bu: Opened issue #10214  for the issues with the slots that I stumbled over above

ka-bu: Actually this issue is related to the whole area of "applied event" logic which was basically ignored so far / where we were just told to not use applied events. I'll open another issue just so we have a reminder on that as well.

rasa/core/evaluation/marker_base.py
```diff
@@ -341,11 +342,9 @@ def _split_sessions(events: List[Event]) -> List[Tuple[List[Event], int]]:
         if len(session_start_indices) == 0:
             return [(events, 0)]
         sessions_and_start_indices: List[Tuple[List[Event], int]] = []
-        for session_idx in range(len(session_start_indices)):
-            start_idx = (
-                session_start_indices[session_idx - 1] if (session_idx > 0) else 0
-            )
-            end_idx = session_start_indices[session_idx]
+        for session_idx in range(len(session_start_indices)-1):
```
usc-m, line 345:
```suggestion
        for session_idx in range(len(session_start_indices) - 1):
```

rasa/core/evaluation/marker_base.py
```diff
@@ -341,11 +342,9 @@ def _split_sessions(events: List[Event]) -> List[Tuple[List[Event], int]]:
...
+            start_idx = session_start_indices[session_idx]
+            end_idx = session_start_indices[session_idx+1]
```
usc-m, line 347:
```suggestion
            end_idx = session_start_indices[session_idx + 1]
```

usc-m (approved):
Looks good but think we've still got an unresolved problem around empty sessions/session retstarts - e.g. a user starts a session and immediately leaves, or a session is immediately restarted. We'd need to think about how to handle those, and if they also add noise to the data (or maybe there's value in them and it would be good to see how often that's happening?)

tests/core/evaluation/test_marker.py
ka-bu: done: #10215

ka-bu:
> a user starts a session and immediately leaves, or a session is immediately restarted.

Yeah that is not so easily determined just by the count of the session events. E.g  there should be at least an initial action listen in addition to the start events, no? 🤔   

> We'd need to think about how to handle those, and if they also add noise to the data (or maybe there's value in them and it would be good to see how often that's happening?).

Agree, that is probably a whole separate thing people would be interested in monitoring ("How often do people just ignore the bot and give up before they even started talking to it" :D) and that when analysing the data one would want to ignore those cases / treat them separately. But since we don't really compute metrics over all sessions / we already agreed that the extracted meta data on relevant events is more relevant anyway, think we're ok / don't need to open an issue on that now.

aeshky:
Thank you both. @ka-bu Looks like you need to approve the PR too since you requested some changes. I've enabled auto merge.

ka-bu (approved):

PR #10207, opened 2021-11-15
3.0 mitie nlp error message

### LINKED ISSUE #10140: Unclear error message when mitie model does not exist on path
When a model is provided for `MitieNLP` but doesn't exist on disk
e.g.
```yml
pipeline:
  - name: MitieNLP
    model: "data/total_word_feature_extractor.dat"
```
but no `"data/total_word_feature_extractor.dat"`

This is the error:
```
InvalidConfigException: The MITIE component 'MitieNLP' needs the configuration value for 'model'.Please take a look at the documentation in the pipeline section to get more info about this parameter.
```
 However this makes it seem like the parameter is not provided which is confusing. We should have a different message if the file does not exist.

carlad:
closes #10140

Adds a separate error message to clarify when configuration for MITIE NLU component is incorrect due to a missing file.

### DIFF
--- a/rasa/nlu/utils/mitie_utils.py
+++ b/rasa/nlu/utils/mitie_utils.py
@@ -99,7 +99,7 @@ class MitieNLP(GraphComponent):
             raise InvalidConfigException(
                 "The model file configured in the MITIE"
                 "component cannot be found. "
-                "Please ensure the directory path and "
+                "Please ensure the directory path and/or "
                 "filename, '{}', are correct.".format(Path(model_file))
             )
         extractor = mitie.total_word_feature_extractor(str(model_file))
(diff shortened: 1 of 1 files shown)


rasa/nlu/utils/mitie_utils.py
```diff
@@ -95,6 +95,13 @@ def create(
                 "to get more info about this "
                 "parameter."
             )
+        if not Path(model_file).is_file():
+            raise InvalidConfigException(
+                "The model file configured in the MITIE "
+                "component cannot be found. "
+                "Please ensure the directory path and/or "
+                "filename, '{}', are correct.".format(Path(model_file))
```
tayfun, line 103: Is it better to use `model_file` for error message instead of a Path object?

tayfun (changes requested):
Minor comment

alwx (approved):
Approving it but agree with @tayfun that it might be better to just use `model_file` instead of `Path(model_file)` in this case.

tayfun (approved):

Issue #10227, opened 2021-11-16
Order of commands broken in rc2 for succes markers


koaning:
I'm trying out success markers, [per docs](https://github.com/RasaHQ/rasa/blob/main/docs/docs/markers.mdx):

```bash
rasa evaluate markers all --config marker.yml extracted_markers.csv
```

This results in an error in RC2 though. 

```
> python -m rasa evaluate markers all --config marker.yml extracted_markers.csv 

usage: rasa evaluate markers [-h] [-v] [-vv] [--quiet] [--config CONFIG]
                             [--no-stats | --stats-file-prefix [STATS_FILE_PREFIX]]
                             [--endpoints ENDPOINTS] [-d DOMAIN]
                             output_filename {first_n,sample,all} ...
rasa evaluate markers: error: argument strategy: invalid choice: 'extracted_markers.csv' (choose from 'first_n', 'sample', 'all')
```

If you flip the order around, then it does work. 

```
python -m rasa evaluate markers --config markers.yml extracted_markers.csv all
```

It seems like either the docs are faulty or the implementation needs to be changed.

aeshky:
Thanks for submitting this @koaning 
Based on some feedback we received in the docs, we decided to reorder the command to make it more intuitive for users. However, we haven't impelmetned the changes yet. We discussed them here: https://github.com/RasaHQ/rasa/issues/10174
We'll create a PR to close both this issue and the one I linked. :)

Issue #10174, opened 2021-11-11
Marker CLI changes


aeshky:

usc-m:
2/3 and are trivial and good ideas so that sounds good.

1 is a bit fiddly because with the current design we have a tradeoff between exclusivity (making sure users can't input invalid commands) and prettiness here. The way `argparse` works, we can't have mutually exclusive groups of arguments contingent on the `strategy` without using sub-parsers. The way sub-parsers work they will show their own help independently when you call `--help` on them directly. One way around this is to add all the arguments to the parser a level up (in this case `markers`), but then we lose the exclusivity property and have to implement validation to make sure commands are still valid - this would achieve what you're describing.

However, I think a better idea is to actually push the common arguments out to the individual strategies, so that `--help` on `evaluate markers` would show you the list of strategies, and then you would pick one and then call `--help` on that to see *all* of the arguments. This doesn't create any extra validation hassle, means help only shows exactly what it needs to, and makes the command flow a bit better (`rasa evaluate markers first_n 100 output.csv --config blah.yaml`). The main extra hassle is making sure all of the duplicated arguments don't interfere etc (which is more QA related than writing extra validation code)

Another suggestion I saw was to change `--config` to `--markers` by @koaning since that clarifies that we're not talking about the model config, but the markers config - we could also do that here too.

aeshky:
`rasa evaluate markers first_n 100 output.csv --config blah.yaml` looks good 👍 

I first attempted `rasa evaluate markers first_n --help`, then realised I had to insert the file, so I prefer the design you're suggesting.

As for `--config`, I prefer it over `--markers`. I worry we might call too many things `markers` (e.g., the marker configurations, vs marker output).

koaning:
There's a separate issue for `--help` commands in general here that may be relevant to be aware of: https://github.com/RasaHQ/rasa/issues/10103.

PR #10234, opened 2021-11-16
Markers cli changes

aeshky:
**Proposed changes**:
- Closes https://github.com/RasaHQ/rasa/issues/10174
- Closes https://github.com/RasaHQ/rasa/issues/10227

### DIFF
--- a/rasa/cli/evaluate.py
+++ b/rasa/cli/evaluate.py
@@ -43,8 +43,6 @@ def add_subparser(
         help="Applies marker conditions to existing trackers.",
     )
 
-    arguments.set_markers_arguments(marker_parser)
-
     markers_subparser = marker_parser.add_subparsers(dest="strategy")
 
     markers_first_n_subparser = markers_subparser.add_parser(
@@ -56,8 +54,10 @@ def add_subparser(
     )
     arguments.set_markers_first_n_arguments(markers_first_n_subparser)
 
+    arguments.set_markers_arguments(markers_first_n_subparser)
+
     markers_sample_subparser = markers_subparser.add_parser(
-        "sample",
+        "sample_n",
         parents=parents,
         conflict_handler="resolve",
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
@@ -65,13 +65,18 @@ def add_subparser(
     )
     arguments.set_markers_sample_arguments(markers_sample_subparser)
 
-    markers_subparser.add_parser(
+    arguments.set_markers_arguments(markers_sample_subparser)
+
+    # markers_subparser.add_parser(
+    markers_all_subparser = markers_subparser.add_parser(
         "all",
         parents=parents,
         conflict_handler="resolve",
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
         help="Select all trackers.",
     )
+
+    arguments.set_markers_arguments(markers_all_subparser)
 
     marker_parser.set_defaults(func=_run_markers_cli)
 
@@ -83,7 +88,7 @@ def _run_markers_cli(args: argparse.Namespace) -> None:
         args: The arguments passed in from the CLI.
     """
     seed = args.seed if "seed" in args else None
-    count = args.count if "count" in args else None
+    count = args.count if "N" in args else None
 
     stats_file_prefix = args.stats_file_prefix if args.stats else None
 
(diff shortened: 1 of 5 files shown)


rasa/cli/evaluate.py
```diff
@@ -83,7 +87,7 @@ def _run_markers_cli(args: argparse.Namespace) -> None:
         args: The arguments passed in from the CLI.
     """
     seed = args.seed if "seed" in args else None
-    count = args.count if "count" in args else None
+    count = args.count if "N" in args else None
```
usc-m, line 90: This won't work because `count` doesn't exist in `args` any more. I would add `dest='count'` to the argument definition and then leave it as it was before - since I think having `N` as an entry in the argument namespace will be a lot less clear

usc-m (changes requested):

rasa/cli/evaluate.py
aeshky: Ok will fix it.

usc-m: What I meant by this was to add the `dest` to the `add_argument` calls for `N` since then you can redirect where the value is stored (since in the code imo `count` is more understandable) - this also means that none of the rest of the code needs changed because the value the user gives for `N` would be saved to `args.count`

aeshky: I have reverted back from `N` to `count` since it's clearer what it means. 

>I would add dest='count' to the argument definition

Where did you want to add that? `count` is already the dest [here](https://github.com/RasaHQ/rasa/blob/30026d2658f2dd079c3d76e93eeee1cc24e8f633/rasa/cli/arguments/evaluate.py#L54) and [here](https://github.com/RasaHQ/rasa/blob/30026d2658f2dd079c3d76e93eeee1cc24e8f633/rasa/cli/arguments/evaluate.py#L64) (first positional argument is [dest](https://docs.python.org/3/library/argparse.html#dest)).

usc-m: First postional argument is `name` (see [here](https://docs.python.org/3/library/argparse.html#the-add-argument-method) - `dest` is set as `name` by default unless `dest` is overwritten as it says in the docs you've linked. So what I was suggesting was change `name` to `N`, but then set `dest` as `count` (as an additional named argument) to avoid breaking the usage of `args.count` (since `count` will be where the value the user gives for `N` is saved)

aeshky: Ok thanks for clarifying. Are you happy to keep it as `count` everywhere? Or do you think it's better to implement the change you suggested? What I mean is: what do you think will be clearer for users, `count` or `N`?

usc-m: I'm happy to keep it as `count`, I thought you were keen to change it to `n` for clarity (I personally am not sure there's a difference between the two in terms of clarity) - if you want to change it to `n` I'm happy with that too

aeshky: Thanks Matt. I'm happy to keep it as `count`. 
Let me know if you're happy with the rest of the PR.

tests/cli/test_rasa_evaluate_markers.py
```diff
@@ -47,58 +47,62 @@ def write_markers_config_to_yaml(
...
-def test_evaluate_markers_sample_help(run: Callable[..., RunResult]):
+def test_evaluate_markers_sample_n_help(run: Callable[..., RunResult]):
     # We need to specify an output_filename as that's the first positional parameter
-    output = run("evaluate", "markers", "test.csv", "sample", "--help")
+    output = run("evaluate", "markers", "sample_n", "--help")
 
-    help_text = """usage: rasa evaluate markers output_filename sample [-h] [-v] [-vv] [--quiet]
-                                                    [--seed SEED]
-                                                    count"""
+    help_text = """usage: rasa evaluate markers sample_n [-h] [-v] [-vv] [--quiet] [--seed SEED]
```
usc-m, line 81: Not entirely sure why these have become un-indented but tests seem to pass?

usc-m (approved):
Rest of the PR looks fine - hopefully that makes the CLI a bit clearer to use :)

tests/cli/test_rasa_evaluate_markers.py
aeshky: They don't have to be un-indented. I just had to change the content and ended up formatting differently. I am stripping leading and trailing whitepace in the test so that the indentation wouldn't matter, just the content.

PR #10236, opened 2021-11-16
Fix confusing warning message when using custom mapping action

### LINKED ISSUE #10144: Warning when using custom slot mapping action
### Rasa Open Source version

3.0.0rc1

### Rasa SDK version

_No response_

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

I ran into this warning when using a custom slot mapping action:

```python
2021-11-09 15:28:38 INFO     rasa.core.actions.action  - Running custom action 'action_validate_slot_mappings' has resulted in an event of type 'slot'. This is disallowed and the tracker will not be updated with this event.
```

I guess this happens because of the 

```
if (
    isinstance(event, SlotSet)
    # this triggers if the action extracted the same slot again
    and tracker.get_slot(event.key) != event.value
)
```

```python
class BetterConditionalResponseValidator(ValidationAction):

    async def extract_better(
        self,
        tracker: Tracker,
        domain: Dict[Text, Any],
    ) -> Dict[Text, Any]:

        return {"better": bool(len(tracker.events) % 2)}
```

### Command / Request

_No response_

### Relevant log output

_No response_

ancalita:
**Proposed changes**:
- Fixes #10144 

### DIFF
--- a/rasa/core/actions/action.py
+++ b/rasa/core/actions/action.py
@@ -1026,11 +1026,9 @@ class ActionExtractSlots(Action):
                 output_channel, nlg, tracker, domain
             )
             for event in custom_events:
-                if (
-                    isinstance(event, SlotSet)
-                    and tracker.get_slot(event.key) != event.value
-                ):
-                    slot_events.append(event)
+                if isinstance(event, SlotSet):
+                    if tracker.get_slot(event.key) != event.value:
+                        slot_events.append(event)
                 elif isinstance(event, BotUttered):
                     slot_events.append(event)
                 else:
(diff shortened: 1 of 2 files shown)


rasa/core/actions/action.py
```diff
@@ -1026,11 +1026,9 @@ async def _run_custom_action(
...
             )
             for event in custom_events:
-                if (
-                    isinstance(event, SlotSet)
-                    and tracker.get_slot(event.key) != event.value
-                ):
-                    slot_events.append(event)
+                if isinstance(event, SlotSet):
+                    if tracker.get_slot(event.key) != event.value:
+                        slot_events.append(event)
```
alwx, line 1031: Nite catch! Do we also need a test for this?

alwx (approved):

rasa/core/actions/action.py
ancalita: It would probably be a good idea rather than relying on manual testing 😄 I'll create one now.

alwx (approved):

PR #10263, opened 2021-11-17
some typos; add one sentence

ka-bu:
**Proposed changes**:
- found some missing commas while testing the custom component migration
- the rendering of the `python (...file ) {...lines...}` doesn't seem to work at one point -- I've switched the order of the lines and file reference but I'm not sure that solves it (cannot test this locally atm :/) 
- added one sentence to remind people that they cannot expect that the getting started examples solve everything
- also: should we add a remark that the parallelisation described in the introduction will be a feature build on top in future versions (i.e. imho the parallelisation of count vector featurizers can't work at the moment but is suggested by the first picture -- the only parallelisation that we (maybe, can't remember how we translate that default recipe to graph) have is the policies, no?)


joejuzl (approved):
💯

PR #10265, opened 2021-11-17
Improve success message of data migration

### LINKED ISSUE #10242: `rasa data migrate` should give more informative message
Upon successfully running `rasa data migrate`, I am told:
```
Your domain was successfully migrated! It's location is domain.yml.
```

I think this isn't clear enough and we should change it to something like:
```
Your domain file (<original domain name file/path>) was successfully migrated! The migrated version is now <new file/path>. The original domain file is backed-up as <back-up file/path>.
```
Note: The message is this brief even if I also had some file named `original_domain.yml` (by accident) which got overwritten by the migration command.

carlad:
closes #10242

**Proposed changes**:
This improves the success message when domain data is successfully migrated from 2.0 to 3.0

### DIFF
--- a/rasa/core/migrate.py
+++ b/rasa/core/migrate.py
@@ -245,6 +245,7 @@ def migrate_domain_format(
     if domain_file.is_dir():
         backup_dir = current_dir / "original_domain"
         backup_dir.mkdir()
+        backup_path = backup_dir / "original_domain.yml"
 
         if out_file.is_file() or not out_file.exists():
             out_file = current_dir / "new_domain"
@@ -274,8 +275,8 @@ def migrate_domain_format(
                 f"domain file. Only domain yaml files can be migrated. "
             )
 
-        backup = current_dir / "original_domain.yml"
-        original_domain = _create_back_up(domain_file, backup)
+        backup_path = current_dir / "original_domain.yml"
+        original_domain = _create_back_up(domain_file, backup_path)
 
     new_forms, updated_slots = _migrate_form_slots(original_domain)
     new_slots = _migrate_auto_fill_and_custom_slots(original_domain, updated_slots)
@@ -285,5 +286,5 @@ def migrate_domain_format(
     rasa.shared.utils.cli.print_success(
         f"Your domain file '{str(domain_file)}' was successfully migrated! "
         f"The migrated version is now '{str(out_file)}'. "
-        f"The original domain file is backed-up as 'original_domain.yml'."
+        f"The original domain file is backed-up as '{str(backup_path)}'."
     )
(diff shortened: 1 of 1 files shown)


rasa/core/migrate.py
```diff
@@ -283,5 +283,7 @@ def migrate_domain_format(
     _write_final_domain(domain_file, new_forms, new_slots, out_file)
 
     rasa.shared.utils.cli.print_success(
-        f"Your domain was successfully migrated! It's location is {out_file}."
+        f"Your domain file '{str(domain_file)}' was successfully migrated! "
+        f"The migrated version is now '{str(out_file)}'. "
+        f"The original domain file is backed-up as 'original_domain.yml'."
```
ancalita, line 288: The back-up location shouldn't be hardcoded, because if a directory path was provided as domain, then the back-up location will be the `original_domain` directory, otherwise if a `domain.yml` file was provided as domain, then the back-up location will be `original_domain.yml`. 

Probably best to align the two variables in this function in the above if statement to have same name (`backup_dir` and `backup`) and then provide this as something along the lines of  `The original domain is backed-up in '{str(backup_path)}'`

ancalita (changes requested):
⚡ Left one suggestion to replace the hardcoded back-up location.

rasa/core/migrate.py
carlad: I needed to define `backup_path` in more than one if statement. Does this change work better?

rasa/core/migrate.py
```diff
@@ -245,6 +245,7 @@ def migrate_domain_format(
     if domain_file.is_dir():
         backup_dir = current_dir / "original_domain"
         backup_dir.mkdir()
+        backup_path = backup_dir / "original_domain.yml"
```
ancalita, line 248: Oh this line is not necessary, you can rename `backup_dir` to `backup_path`, that's what i meant.

carlad: Okay - so the message will contain either a path to a directory, or a path to a file.

ancalita: Yup, exactly!

rasa/core/migrate.py
```diff
@@ -274,14 +274,16 @@ def migrate_domain_format(
                 f"domain file. Only domain yaml files can be migrated. "
             )
 
-        backup = current_dir / "original_domain.yml"
-        original_domain = _create_back_up(domain_file, backup)
+        backup_dir = current_dir / "original_domain.yml"
```
ancalita, line 277: Almost there! 💯 
This could be nitpicking, but I'd rather rename all `occurences` of `backup_dir` to a more general name (`backup_location` or any other), otherwise having this assignment of file path to a name that contains `dir` is misleading and confusing for somebody looking in this file for first time.

carlad: yes I was thinking the same 👍

ancalita (approved):
🎉

samsucik (approved):
Thanks @carlad, I tested this locally and it works 🙂

Issue #10148, opened 2021-11-10
Model API QA: Errors met when testing model training and replacing loaded model


ancalita:
**Problem description**:

1. Testing `/model/train` endpoint with the example in the docs leads to a YAML validation error (400 Bad Request error:
```
"The request body does not contain valid YAML. Error: Failed to validate YAML. Please make sure the file is correct and all mandatory parameters are specified. Here are the errors found during validation:\n  in Line 1:\n      Value 'b'- intent: greet\\\\n examples: |\\\\n - hey\\\\n - hello\\\\n - hi'' is not a list. Value path: '/nlu'"
```

2. Testing PUT `/model` endpoint with this request body (existing file on disk) leads to 400 Bad Request:

```  
{
"model_file": "/models/20211110-091805-impulsive-birth.tar.gz"
}
```
It seems not to be able to identify the path given, this is the response error message:
`"Agent with name '/models/20211110-091805-impulsive-birth.tar.gz' could not be loaded."`
The debug log shows this user warning:
`UserWarning: No valid configuration given to load agent. Agent loaded with no model!`

melindaloubser1:
Found the same - @tayfun I see you're assigned so maybe you already saw this, but it expects literal `yaml`, so a documentation issue - this request works:
```
curl --request POST \
  --url http://localhost:5005/model/train \
  --header 'Content-Type: text/yaml' \
  --data 'version: '\''3.0'\''
config:
  store_entities_as_slots: true
session_config:
  session_expiration_time: 60
  carry_over_slots_to_new_session: true
intents:
- greet:
    use_entities: true
- goodbye:
    use_entities: true
- affirm:
    use_entities: true
- deny:
    use_entities: true
- mood_great:
    use_entities: true
- mood_unhappy:
    use_entities: true
- bot_challenge:
    use_entities: true
entities:
- name
slots:
  name:
    type: rasa.shared.core.slots.AnySlot
    initial_value: null
    influence_conversation: false
    mappings:
    - type: from_entity
      entity: name
  yes:
    type: rasa.shared.core.slots.AnySlot
    initial_value: null
    influence_conversation: false
    mappings:
    - type: from_intent
      intent: affirm
      value: yes
    - type: from_intent
      not_intent: affirm
      value: no
responses:
  utter_greet:
  - text: Hey {name}! How are you?
  utter_cheer_up:
  - text: '\''Here is something to cheer you up:'\''
    image: https://i.imgur.com/nGF1K8f.jpg
[...]

PR #10268, opened 2021-11-17
Fix API docs

tayfun:
**Proposed changes**:

Fixes https://github.com/RasaHQ/rasa/issues/10148

- Updates `/model/train` endpoint in docs to only accept YAML input
- Updates `/model` endpoint to clarify model path

### DIFF
--- a/docs/static/spec/rasa.yml
+++ b/docs/static/spec/rasa.yml
@@ -482,12 +482,12 @@ paths:
         description: >-
           The training data should be in YAML format.
         content:
-          application/json:
+          application/yaml:
             schema:
-              $ref: '#/components/schemas/JSONTrainingRequest'
-          application/x-yaml:
-            schema:
-              type: string
+              $ref: '#/components/schemas/YAMLTrainingRequest'
+            example:
+              $ref: '#/components/examples/YAMLTrainingRequest'
+
 
       responses:
         200:
@@ -1407,19 +1407,40 @@ components:
           type: integer
           description: Time to wait between pulls from model server
 
-    JSONTrainingRequest:
+    YAMLTrainingRequest:
       type: object
       properties:
-        domain:
-          $ref: '#/components/schemas/DomainFile'
-        config:
-          $ref: '#/components/schemas/ConfigFile'
+        intents:
+          type: array
+          description: All intent names and properties
+          items:
+            $ref: '#/components/schemas/IntentDescription'
+        responses:
+          description: Bot response templates
+          type: object
+          additionalProperties:
+            $ref: '#/components/schemas/TemplateDescription'
+        session_config:
+          description: Session configuration options
+          type: object
+          properties:
+            session_expiration_time:
+              type: integer
+            carry_over_slots_to_new_session:
+              type: boolean
+        pipeline:
+          description: Pipeline list
+          type: array
+        policies:
+          description: Policies list
         nlu:
           $ref: '#/components/schemas/NLUTrainingData'
-        responses:
-          $ref: '#/components/schemas/RetrievalIntentsTrainingData'
+        rules:
+          description: Rule list
+          type: array
         stories:
           $ref: '#/components/schemas/StoriesTrainingData'
+
         force:
           type: boolean
           description: >-
@@ -1432,7 +1453,6 @@ components:
             If `true` (default) the trained model will be saved in the default model
             directory, if `false` it will be saved in a temporary directory
           deprecated: True
-      required: ["config"]
 
     NLUTrainingData:
       type: string
@@ -1860,3 +1880,50 @@ components:
           type: string
           description: Text of the message
           example: "Hello!"
+  examples:
+    YAMLTrainingRequest: |
+      intents:
+        - greet
+        - goodbye
+
+      responses:
+        utter_greet:
+        - text: "Hey! How are you?"
+
+        utter_goodbye:
+        - text: "Bye"
+
+      session_config:
+        session_expiration_time: 60
+        carry_over_slots_to_new_session: true
+
+      pipeline:
+
+      policies:
+
+      nlu:
+      - intent: greet
+        examples: |
+          - hey
+          - hello
+
+      - intent: goodbye
+        examples: |
+          - bye
+          - goodbye
+
+      rules:
+
+      - rule: Say goodbye anytime the user says goodbye
+        steps:
+        - intent: goodbye
+        - action: utter_goodbye
+
+      stories:
+
+      - story: happy path
+        steps:
+        - intent: greet
+        - action: utter_greet
+        - intent: goodbye
+        - action: utter_goodbye
--- a/docs/static/spec/rasa.yml
+++ b/docs/static/spec/rasa.yml
@@ -718,7 +718,8 @@ paths:
       summary: Replace the currently loaded model
       description: >-
         Updates the currently loaded model.
-        First, tries to load the model from the local storage system.
+        First, tries to load the model from the local (note: local to Rasa server)
+        storage system.
         Secondly, tries to load the model from the provided model server configuration.
         Last, tries to load the model from the provided remote storage.
       requestBody:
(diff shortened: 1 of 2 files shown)


docs/static/spec/rasa.yml
```diff
@@ -1052,7 +1053,7 @@ components:
         model_file:
           type: string
           description: Path to model file
-          example: "/models/20190512.tar.gz"
+          example: "/CHANGE-TO-DIR-ON-REMOTE-SERVER/models/20190512.tar.gz"
```
melindaloubser1, line 1056:
```suggestion
          example: "/absolute-path-to-models-directory/models/20190512.tar.gz"
```

rasa/server.py
```diff
@@ -1422,34 +1422,6 @@ def _test_data_file_from_payload(request: Request, temporary_directory: Path) ->
...
-        )
-
-    if "nlu" not in rjs and "stories" not in rjs:
-        raise ErrorResponse(
-            HTTPStatus.BAD_REQUEST,
-            "BadRequest",
-            "To train a Rasa model you need to specify at least one type of "
-            "training data. Add `nlu` and/or `stories` to the request.",
-            {"parameters": ["nlu", "stories"], "in": "body"},
-        )
```
melindaloubser1, line 1441: This is a functionality change, not just docs change - Can you tag an additional reviewer because I am not caught up on why this should or shouldn't happen

docs/static/spec/rasa.yml
```diff
@@ -1407,19 +1408,40 @@ components:
...
 
-    JSONTrainingRequest:
+    YAMLTrainingRequest:
       type: object
       properties:
-        domain:
-          $ref: '#/components/schemas/DomainFile'
-        config:
-          $ref: '#/components/schemas/ConfigFile'
+        intents:
```
melindaloubser1, line 1414: This section should include all properties (missing ones from the domain include actions, entities, e2e_actions, config) and be ordered (IMO) by config.yml, then domain.yml, then the data files

docs/static/spec/rasa.yml
```diff
@@ -1860,3 +1881,50 @@ components:
...
           example: "Hello!"
+  examples:
+    YAMLTrainingRequest: |
+      intents:
+        - greet
+        - goodbye
+
+      responses:
+        utter_greet:
+        - text: "Hey! How are you?"
```
melindaloubser1, line 1892: Same on this section: reorder by config.yml then domain.yml then data, and add all keys even if they're empty e.g. `actions: []`

melindaloubser1 (changes requested):
Mostly concerned about the function removal - not sure that should be happening here?

rasa/server.py
melindaloubser1: Ope I see it auto tagged someone, so that's good

tayfun: I'm removing code that's not used anymore. JSON was not supported, this function is not used currently. People have said docs had been stale for some time now.

docs/static/spec/rasa.yml
tayfun: > This section should include all properties (missing ones from the domain include actions, entities, e2e_actions, config) and be ordered (IMO) by config.yml, then domain.yml, then the data files

@melindaloubser1 are you sure actions etc. would be handled here? I don't see it in stale docs either: https://rasa.com/docs/rasa/pages/http-api#operation/trainModel I've taken the example and schema from unit tests.

docs/static/spec/rasa.yml
tayfun: > Same on this section: reorder by config.yml then domain.yml then data, and add all keys even if they're empty e.g. `actions: []`

@melindaloubser1 I don't think actions are used here. I don't understand the ordering, why would it `session_config` come before others? Or do you mean `pipeline` comes first, then `policies`?

melindaloubser1: For this particular example, there are no custom actions to pass, so you are correct. However, in a bot with custom actions, or slots, or entities, these sections would need to be included.

Re. the ordering: Sorry for the shorthand, I meant the contents of config.yml, so yes, pipeline + policies, then domain content (intents, entities, actions, e2e_actions, forms, slots, responses, config<- this is a key in the domain in 3.0, which is quite confusing) then content of nlu, rules and stories. It's not required to do it this way for the example to work, but will make it much easier for readers to understand the example

tayfun: @melindaloubser1 Yeah I meant do you know custom actions are used if provided in request body like here? I don't see that in code, and I don't see them being tested in unit tests. I wouldn't be surprised if not all CLI options were supported here.

docs/static/spec/rasa.yml
joejuzl: Isn't all of this covered by `#/components/schemas/DomainFile`?

docs/static/spec/rasa.yml
```diff
@@ -1407,19 +1451,40 @@ components:
...
+            $ref: '#/components/schemas/TemplateDescription'
+        session_config:
+          description: Session configuration options
+          type: object
+          properties:
+            session_expiration_time:
+              type: integer
+            carry_over_slots_to_new_session:
+              type: boolean
+        pipeline:
```
joejuzl, line 1475: Why the change from `#/components/schemas/ConfigFile`?

joejuzl (changes requested):

docs/static/spec/rasa.yml
tayfun: @joejuzl It's not under `domain` anymore, request body is flattened. This is why I've newly added this schema. Also, examples inside schemas do not work for any content type except for JSON in redocly, so `example` keys were removed and a separate example on the same level as schema was added. The other thing is I've simplified the example, previously there were 6 intents in the example request, I've added 2 to show how a list can be sent, I think people can extrapolate that and add more intents if needed.

docs/static/spec/rasa.yml
tayfun: @joejuzl Similar to domain, request body is flattened, `config` key doesn't exist anymore.

joejuzl: is there no way to include the contents of `#/components/schemas/ConfigFile` at the top level, i.e. without it being under it's own key?

docs/static/spec/rasa.yml
melindaloubser1: They would have to be supported to train on stories which include custom actions I believe

docs/static/spec/rasa.yml
tayfun: I've tried using `allOf` but couldn't get it to work. To test this, I've run `make livedocs` in top level and then added:
```
    TrainingSchema:
      allOf:
        - $ref: '#/components/schemas/DomainFile'
        - $ref: '#/components/schemas/ConfigFile'
```
to the bottom of the file and referenced it in requestBody as:
```
          application/yaml:
            schema:
              - $ref: '#/components/schemas/TrainingSchema'
```
I don't see schema and only see `any` in the docs output in that case. @joejuzl @melindaloubser1 do you know a better way?

tayfun: Well that second part's schema should not be a list:
```
         application/yaml:
            schema:
              $ref: '#/components/schemas/TrainingSchema'
```
Notice the dash is missing. This is still not showing the schema, but instead of `any`, I get `string (TrainingSchema)`

melindaloubser1: Unfortunately I haven't found another way either - Only experimented briefly but ran into the same issues you did @tayfun

rasa/server.py
melindaloubser1: Ah right, I see now there's no references left to this function anywhere 👍

joejuzl (approved):
Bit annoying we can't inline the other schemas... but at least it's correct now! 😁

melindaloubser1 (approved):

Issue #10246, opened 2021-11-16
Wrong error message during finetuning


dakshvar22:
### Rasa Open Source version

3.0.0rc2

### Python version

3.8

### What operating system are you using?

OSX

### What happened?

Started with a `rasa init` project which trained an initial model. I made changes to the NLU training data by adding a new intent:

```
- intent: new_intent
  examples: |
    - yes i am new
    - can i break you
    - i don't think i want to do this
    - but its exciting
```

Ran `rasa train --finetune`. The process exits with:

```
InvalidConfigException: Cannot finetune because more than just the 'epoch' keys have been changed in the configuration. Please revert your configuration and only change the 'epoch' settings where needed.
```

whereas according to 2.8.x, the log should have been:

```
2021-11-16 15:44:23 INFO     rasa.model  - Data (nlu_labels) for finetune section changed.
NLU model can not be finetuned.
```

which is the correct info as we only changed the data and added a new intent label which should disallow finetuning.

### Command / Request

```shell
rasa train --finetune
```

dakshvar22:
It is correct that if new intents, entities, entity groups, entity roles, and action names have been added compared to the initial training phase, finetuning is not possible and it should return an appropriate message that new labels have been added (you can try `2.8.x` branch to test this).

This does not apply to adding a new training example to **existing** intents, entities, etc. 

To answer your questions:

> Are changes to the nlu.yml counted as "configuration" changes?

No, they are not configuration changes. Configuration changes are only the ones that are made to `config.yml` file.

> did you make any configuration changes?

Nope 

> then perhaps we should re-arrange the order in which we check the cases listed above?

The checks are fine but I suspect that an addition of a new training example to existing intent is being treated as an addition of a new intent which is wrong. 😅

carlad:
Thanks @dakshvar22 😄 

I did have a question:

> The checks are fine but I suspect that an addition of a new training example to existing intent is being treated as an addition of a new intent which is wrong

If I understand your original comment, what you did was add an entirely new intent:
```
- intent: new_intent
  examples: |
    - yes i am new
    - can i break you
    - i don't think i want to do this
    - but its exciting
 ```

And that this should trigger the 3rd check (the nlu check), and show the message:
```
"Cannot finetune because NLU training data contains new labels "
"or does not contain any examples for some known labels. " ...
```

Did you add anything to an existing intent?

[...]

PR #10282, opened 2021-11-18
exlcude defaults from validation

### LINKED ISSUE #10247: Unable to finetune on rc2
### Rasa Open Source version

3.0.0rc2

### Rasa SDK version

_No response_

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

Trained an initial model with `rasa init`, added a new training example to one of the existing intents and ran `rasa train --finetune`. Fails with:

```
InvalidConfigException: Cannot finetune because more than just the 'epoch' keys have been changed in the configuration. Please revert your configuration and only change the 'epoch' settings where needed.
```

Might be related to https://github.com/RasaHQ/rasa/issues/10246

### Command / Request

_No response_

### Relevant log output

_No response_

ka-bu:
**Proposed changes**:
- closes #10247 and #10246

### DIFF
--- a/rasa/cli/arguments/train.py
+++ b/rasa/cli/arguments/train.py
@@ -151,7 +151,6 @@ def add_augmentation_param(
     parser.add_argument(
         "--augmentation",
         type=int,
-        default=50,
         help="How much data augmentation to use during training.",
     )
 
@@ -162,7 +161,6 @@ def add_debug_plots_param(
     """Specifies if conversation flow should be visualized."""
     parser.add_argument(
         "--debug-plots",
-        default=False,
         action="store_true",
         help="If enabled, will create plots showing checkpoints "
         "and their connections between story blocks in a  "
--- a/rasa/cli/arguments/train.py
+++ b/rasa/cli/arguments/train.py
@@ -8,8 +8,9 @@ from rasa.cli.arguments.default_arguments import (
     add_out_param,
     add_domain_param,
 )
-from rasa.graph_components.providers.training_tracker_provider import \
-    TrainingTrackerProvider
+from rasa.graph_components.providers.training_tracker_provider import (
+    TrainingTrackerProvider,
+)
 from rasa.shared.constants import DEFAULT_CONFIG_PATH, DEFAULT_DATA_PATH
 
 USE_LATEST_MODEL_FOR_FINE_TUNING = True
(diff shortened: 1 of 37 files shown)


carlad:
Do we also want to drop the relevant CLI defaults when training?

carlad (approved):

rasa/cli/arguments/train.py
```diff
@@ -151,7 +153,7 @@ def add_augmentation_param(
     parser.add_argument(
         "--augmentation",
         type=int,
-        default=50,
+        default=TrainingTrackerProvider.get_default_config()['augmentation_factor'],
```
usc-m, line 156: Oh this isn't actually as bad as I thought now that I see it written out - actually happier with this now. Still think it's a bit uncomfortable but it's not the worst

rasa/engine/graph.py
```diff
@@ -22,40 +22,44 @@
...
+        uses: The class which models the behavior of this specific graph node.
+        constructor_name: The name of the constructor which should be used to
+            instantiate the component. If `eager==False` then the `constructor` can
+            also specify parameters which are filled by parent nodes. This is e.g.
+            useful if a parent node returns a `Resource` and this node wants to
+            directly load itself from this resource.
+        fn: The name of the function which should be called on the instantiated
+            component when the graph is executed. The parameters from `needs` are
+            filled from the parent nodes.
+        config: The configuration for this graph node.
```
usc-m, line 39: I wonder if we shouldn't change this to mention that this is the config specified by the user, which will then be merged with the default config?

rasa/cli/arguments/train.py
ka-bu: 🥳

usc-m (approved):
🎉

rasa/engine/graph.py
```diff
@@ -36,7 +36,9 @@ class SchemaNode:
         fn: The name of the function which should be called on the instantiated
             component when the graph is executed. The parameters from `needs` are
             filled from the parent nodes.
-        config: The configuration for this graph node.
+        config: The user's configuration for this graph node. This configuration
+            does not need to be specify all possible parameters; the default values
```
carlad, line 40: A small typo here: `does not need to be specify` should be `does not need to specify`

PR #10283, opened 2021-11-18
updated-example-components

koaning:
I've updated these components. I've done a pass with flake8/black, re-ordered the methods to be friendlier to users and ensured that each component has a docstring attached.


joejuzl (approved):
💯

Issue #10235, opened 2021-11-16
`/restart` in `rasa shell` results in "zombie" slots


samsucik:
Normally, you expect `/restart` to reset all slots (dump their values). However, I found a way to bring such "dead" slot values back to life 👻  which occurs in 3.0rc2 but not in 2.8.

Example set-up is similar to [this](https://www.notion.so/rasa/Test-unique-entity-mapping-matching-behaviour-within-an-active-loop-68a4e2cc487548f8ba25aef86be394ea): To a `rasa init` project add these domain parts:
```yml
intents:
  - inform

responses:
  utter_ask_slot_0:
  - text: "Let's start this form. Say something!"

  utter_ask_slot_A:
  - text: "Could you, possibly, please, give me a city."

  utter_ask_slot_B:
  - text: "Yo pal gimme a city!"

entities:
- city

slots:
  slot_0:
    type: any
    mappings:
    - type: from_text
      not_intent: greet
  slot_A:
    type: any
    mappings:
    - type: from_entity
      entity: city
      conditions:
      - active_loop: my_form
  slot_B:
    type: any
    mappings:
    - type: from_entity
      entity: city
      conditions:
      - active_loop: my_form
forms:
  my_form:
    required_slots:
        - slot_0
        - slot_A
        - slot_B
```
some NLU data for the new intent & entity:
```yml
- intent: inform
  examples: |
    - [Bratislava](city)
    - [Berlin](city)
    - [Prague](city)
    - [Amsterdam](city)
    - [Edinburgh](city)
```
the complete rules will look like this:
```yml
version: "2.0"

rules:

- rule: Say goodbye anytime the user says goodbye
  steps:
  - intent: goodbye
  - action: utter_goodbye

- rule: Say 'I am a bot' anytime the user challenges
  steps:
  - intent: bot_challenge
  - action: utter_iamabot

- rule: Activate form
  steps:
  - intent: greet
  - action: my_form
  - active_loop: my_form

- rule: Submit form
  condition:
  - active_loop: my_form
  steps:
  - action: my_form
  - active_loop: null
  - slot_was_set:
    - requested_slot: null
  - action: utter_happy
```
and you'll have to comment out all stories.

Then, `rasa train` the bot and `rasa shell --debug` talk to it. Below is the sequence of user messages and the interesting slots after the message has been processed, as reported by `rasa shell`:
```
message     slot_0      slot_A      slot_B
hi          None        None        None
Berlin      Berlin      None        None
[...]

Issue #9945, opened 2021-10-21
Fix already valid `SlotSet` events being duplicated in the tracker by the `FormAction`


ancalita:
**Description of the problem**:
[This PR](https://github.com/RasaHQ/rasa/pull/9889) removed slot extraction methods from `FormAction` so that slot extraction is only executed by `ActionExtractSlots`.  The existing `FormAction.validate_slots` method adds already valid events to the tracker in addition to validated slots. This creates duplicate events in the tracker since the already valid events were added to the tracker initially by `ActionExtractSlots`. 

**Solution**: Adapt `FormAction.validate_slots` to only return the validated events.

**Definition of Done**:
- [x] already valid `SlotSet` events are not added to the tracker by `FormAction`
- [x] `test_forms.py` is adapted with the new expected output.

ancalita:
Closed via #10479

PR #10295, opened 2021-11-19
Fix slot extraction if duplicate user utterance

tayfun:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/10235
- This PR fixes 2 issues:

1. Initial issue of zombie slots after `/restart`: FormAction was checking all tracker events from the beginning. Coupled with the fact that UserUtterance has `__eq__` defined so that two messages with same content (such as `hi` - entities etc. are also equal) are taken to be equal, this meant that all events after the first message were taken into consideration, and `restart` events are also ignored. This bug only shows itself when there are duplicate messages and a `restart` event, and this is because slots are linked to unique forms and that once a form is filled in (slots filled) you cannot activate it again. Solution to the problem was searching in tracker store from the end, and also stopping at `RestartedEvent`. Slots are filled from then on.

2. After implementing fix above, `test_formbot_example` was failing at this line: https://github.com/RasaHQ/rasa/runs/4289441802?check_suite_focus=true

  - The `AssertionError` was caused because all form slots were filled so there was nothing to set afterwards, and that resulted in form rejection and the next action prediction was `action_default_fallback`.
  - Upon further inspection of `tracker.events`, we realised that slot `preferences` got set after the `/restart`  user utterance because the original slot mapping conditions were satisfied - the `tracker.active_loop` was the same and the `requested_slot` was `preferences` while intent `restart` was allowed. This was fixed, by adding `restart` to the `not_intent` [list](https://github.com/RasaHQ/rasa/blob/747a7b8e137fc9eee32248d6e638dd127a239ac2/examples/formbot/domain.yml#L64) in the slot mapping.
  - We also realised that `outdoor_seating` got set immediately after [another unhappy path](https://github.com/RasaHQ/rasa/blob/747a7b8e137fc9eee32248d6e638dd127a239ac2/tests/core/test_examples.py#L121), i.e, the sequence of user utterance `stop` -> bot action `utter_ask_continue`  -> user utterance `affirm` , because the form was still active, `requested_slot`  was `outdoor_seating` , intent was `affirm` and therefore all conditions in the slot [mapping](https://github.com/RasaHQ/rasa/blob/747a7b8e137fc9eee32248d6e638dd127a239ac2/examples/formbot/domain.yml#L41) were satisfied. In order to prevent this slot to be set during an unhappy path, we added the 2nd fix in `ActionExtractSlots.run()` - this approach for checking for `LOOP_REJECTED` was required because the event `LoopInterrupted` got added to `tracker.events` after `SlotSet(outdoor_seating, True)` event. We couldn't quite figure out why `LoopInterrupted` doesn't get immediately added to `tracker.events` after the `stop` user utterance and instead `MemoizationPolicy` selects `utter_ask_continue`.

Actions for step 2 above have taken while pair programming with @ancalita 


rasa/core/actions/forms.py
```diff
@@ -308,7 +308,10 @@ def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
     def _get_events_since_last_user_uttered(
         tracker: "DialogueStateTracker",
     ) -> List[SlotSet]:
+        # Why would latest message not be in events?
```
ancalita, line 311: `tracker.latest_message` is of type `Optional[UserUttered]` so in some tests there can be trackers without `UserUttered` events that could fail if this check is not in place.

rasa/core/actions/forms.py
```diff
@@ -308,7 +308,10 @@ def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
...
         if tracker.latest_message in tracker.events:
+            # This will find first event with same text (since entities and intents
```
ancalita, line 313: Good spot with `UserUttered` equality! I did not foresee the problem of duplicate messages with same text. Probably this needs to be fixed also to find the `latest` one, not the first one. But check first if removing the problem with duplicate `SlotSet` events fixes this, it would be awesome if there would be a unit test for `/restart` too, this way the CI would have caught this issue early.

rasa/cli/scaffold.py
```diff
@@ -128,7 +128,7 @@ def init_project(args: argparse.Namespace, path: Text) -> None:
     """Inits project."""
     os.chdir(path)
     create_initial_project(".")
-    print("Created project directory at '{}'.".format(os.path.abspath(path)))
+    print("Created project directory at '{}'.".format(os.path.abspath(os.getcwd())))
```
tayfun, line 131: (This is a fix while logging model directory.)

rasa/core/actions/forms.py
```diff
@@ -308,14 +312,22 @@ def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
     def _get_events_since_last_user_uttered(
         tracker: "DialogueStateTracker",
     ) -> List[SlotSet]:
-        if tracker.latest_message in tracker.events:
-            index = tracker.events.index(tracker.latest_message)
-        else:
+        # TODO: Better way to get this latest_message index is through an instance
+        # variable, eg. tracker.latest_message_index
```
joejuzl, line 316: I think I was wrong when I said we couldn't do this because of the tracker store. We could probably use a similar method to `latest_message`, i.e. setting it in `UserUttered.apply_to`. When we deserialize a tracker from a store we re-apply each event in order, so it should work

rasa/core/actions/action.py
```diff
@@ -1153,6 +1154,12 @@ async def run(
         ]
 
         for slot in user_slots:
+            if (
+                tracker.active_loop
+                and tracker.active_loop.get(LOOP_REJECTED)
+                and tracker.get_slot(REQUESTED_SLOT) == slot.name
+            ):
+                continue
```
ka-bu, line 1162: Looks to me like this check belongs to `_matches_mapping_conditions`. Does that make sense or would we skip some case this way?

rasa/core/actions/forms.py
```diff
@@ -348,6 +360,10 @@ def _add_dynamic_slots_requested_by_dynamic_forms(
     def _get_slot_extractions(
         self, tracker: "DialogueStateTracker", domain: Domain,
     ) -> Dict[Text, Any]:
+        if not tracker.active_loop:
+            # pre-filled slots were already validated at form activation
+            return {}
+
```
ka-bu, line 366: Hm, `_get_slot_extractions` is only called by `validate` which is invoked by `_validate_if_required` if there's no active loop set (i.e. we're in that activation case) or we're in a loop -- If I interpret this correctly, then we might want to pull this check into `_validate_if_required` and explicitly *not* call any validate there in that second case?

ka-bu (commented):
Thanks for the detailed explanation for the fixes! 🚀 

I had a bit trouble following the part about the /restart - because it looks to me like the loop interrupt related fix in the last part might fix that as well. And so I just checked and removed the additional "restart" from the domain and it still seems to work fine for me. (i.e. the slot is extracted from that restart but it doesn't matter because the restart action happens afterwards. If we don't even want that or if I messed something up here :), then we should find a fix that doesn't require users to specify that restart imho (like by just not extracting anything from those special actions))

rasa/core/actions/action.py
```diff
@@ -1153,6 +1154,12 @@ async def run(
         ]
 
         for slot in user_slots:
+            if (
+                tracker.active_loop
+                and tracker.active_loop.get(LOOP_REJECTED)
```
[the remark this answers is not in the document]
ancalita: I considered it, but then I thought this should skip the slot entirely if the active_loop is in an unhappy path and the slot is `requested_slot`. It also doesn't really have anything to do with the mapping conditions check, it would make sense though to extract this code in its own method (`_active_loop_in_unhappy_path` or something like that) 🤔

examples/formbot/domain.yml
```diff
@@ -61,7 +61,7 @@ slots:
       - active_loop: restaurant_form
         requested_slot: preferences
     - type: from_text
-      not_intent: affirm
+      not_intent: [affirm, restart]
```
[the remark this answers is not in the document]
ancalita: Yeah, when we used the `formbot` with `rasa shell --debug`, it seems that the `/restart` text was classified with intent `restart`.

rasa/core/actions/action.py
ancalita: i think it would clutter `_matches_mapping_conditions` with a 2nd use-case that has more to do with the tracker state than with what was defined as mapping conditions in the domain. See also my comment [here](https://github.com/RasaHQ/rasa/pull/10295/files#r756768236).

ancalita: @ka-bu Do you know though why `LoopInterrupted` event would not get added immediately after the user utters `stop`? It's been bugging me for a few days now 😄

rasa/core/actions/forms.py
tayfun: @joejuzl Yes, this is something I discussed with @ancalita as well. The naming would be a little odd, we not only want latest message index but also `restart` event. I can create another PR to check this alternative if that makes sense?

ancalita:
> we should find a fix that doesn't require users to specify that restart imho (like by just not extracting anything from those special actions)

@ka-bu Yes, I agree, but I don't even know how did the bot know to classify `/restart` as `restart` intent?
I guess we could also add these default intents (once we know their provenance) to `not_intent` mapping parameter for all slots maybe during Domain initialisation?

ancalita:
>  the slot is extracted from that restart but it doesn't matter because the restart action happens afterwards.

Aahh I think I just understood this, once the dialogue manager runs the action `restart` it resets all slots anyways? Let me know if I got the order right:
1. `preferences` is first extracted by `action_extract_slots`, 
2. dialogue manager continues with running and executing the predicted action (restart)
3. all slots get reset as a side-effect of action `restart`

If the bot and the `test_formbot` work fine without this change in the slot mapping, then yes, we don't need to action anything else 👍🏼

rasa/core/actions/action.py
ka-bu: Hmmm, in my brain it would unclutter things if this was part of checking whether the loop conditions is *really* met 😅  (Couldn't we have un-intended side effects from skipping all slot mappings? 🤔 )

>  why LoopInterrupted event would not get added immediately after the user utters stop?

Think that's because the "/<command>" stuff first becomes a special intent and then the rule policy takes care of inserting the `LoopInterrupted` and that next action prediction stuff happens as last step. Does that make sense?

ka-bu:
> > we should find a fix that doesn't require users to specify that restart imho (like by just not extracting anything from those special actions)
> 
> @ka-bu Yes, I agree, but I don't even know how did the bot know to classify `/restart` as `restart` intent? I guess we could also add these default intents (once we know their provenance) to `not_intent` mapping parameter for all slots maybe during Domain initialisation?

that sounds good :) - `regex_message_handler` converts the special "/" messages (which uses the yaml extractor thing (which checks the extracted str against the domain)) and `rasa.shared.core.constants.DEFAULT_INTENTS` should be the ones we could exclude then I guess

rasa/core/actions/forms.py
```diff
@@ -308,14 +312,22 @@ def _user_rejected_manually(self, validation_events: List[Event]) -> bool:
...
+        # variable, eg. tracker.latest_message_index
+        try:
+            index_from_end = next(
+                i
+                for i, event in enumerate(reversed(tracker.events))
+                if event == Restarted() or event == tracker.latest_message
+            )
+            index = len(tracker.events) - index_from_end - 1
+        except StopIteration:
             index = 0
```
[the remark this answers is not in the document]
tayfun: Ah that's great, thanks @ka-bu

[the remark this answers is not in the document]
tayfun: @ka-bu I was happy with one less indentation from `try..except` but then we need another parenthesis for `next` when there's a default 🙃  https://stackoverflow.com/a/14250244/73750 this is still better though, thanks!

rasa/core/actions/forms.py
```diff
@@ -256,6 +258,7 @@ async def validate_slots(
         validate_name = f"validate_{self.name()}"
 
         if validate_name not in domain.action_names_or_texts:
+            # TODO: should return empty list
```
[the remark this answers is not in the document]
tayfun: @joejuzl no, it was for this ticket which felt related: https://github.com/RasaHQ/rasa/issues/9945 I'll remove these lines and we can work on this in another pull request.

rasa/core/actions/forms.py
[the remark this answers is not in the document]
ka-bu: oh right, I always miss those extra "()" :D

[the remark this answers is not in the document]
ka-bu: .... and I always forget that there's no "default=". For some reason my brain does not want to remember that :D

rasa/core/actions/action.py
ancalita: > Couldn't we have un-intended side effects from skipping all slot mappings?

Oh you're right! We'll aim to move this into `_matches_mapping_conditions` 👍🏼 

> the "/" stuff first becomes a special intent 

This is the part I don't understand how it happens 🙈 

> the rule policy takes care of inserting the LoopInterrupted

I should have said `LoopInterrupted` is added only after the 2nd user uttered in this sequence, so it wasn't clear to me why it wouldn't get added before `utter_ask_continue` 🤔 :
`stop` -> `utter_ask_continue`-> `affirm` -> `SlotSet` -> `LoopInterrupted`

ka-bu: >  is added only after the 2nd user uttered in this sequence

ahhh, ok, that looks weird - will try to find out 🕵️‍♀️

rasa/core/actions/forms.py
ancalita: Not sure I understood this part, do you mean breaking [`needs_validation`](https://github.com/RasaHQ/rasa/blob/2c17cd24003e2cc9e23276e02b32757445867774/rasa/core/actions/forms.py#L527) into 2 checks, one for `not tracker.active_loop` in which case we return `[ ]` and the other half of the `or` check which can be used in the `if/else` [here](https://github.com/RasaHQ/rasa/blob/2c17cd24003e2cc9e23276e02b32757445867774/rasa/core/actions/forms.py#L532)?

rasa/core/actions/action.py
ancalita: Yeah if the `LoopInterrupted` would be added before the `utter_ask_continue`, it would have simplified our check in the `tracker` for the unhappy path, rather than trying to access the `LOOP_REJECTED` in `tracker.active_loop` which felt a bit like mental gymnastics 🤪

ancalita: Also, just read your comment on `RegexMessageHandler` for handling special intents, going to study that a bit 👍🏼

ka-bu: Sooo, seems like...
- a rule policy will only cause the `LoopInterrupted` to be added if it makes a prediction where `returning_from_unhappy_path = True` but that is not set when it predicts a (i.e. any) default action ( https://github.com/RasaHQ/rasa/blob/main/rasa/core/policies/rule_policy.py#L1123) --> so the stop on it's own has no implications / there is no "default rule" being added for any intent
- but during training, the augmented memoization policy should learn from the stories which includes this:
```
  - story: stop form + stop
    steps:
      - intent: request_restaurant
      - action: restaurant_form
      - active_loop: restaurant_form
      - intent: stop. <-- **stop here**
      - action: utter_ask_continue
      - intent: deny
      - action: action_deactivate_loop
      - active_loop: null <-- **deactivate much later than we think**
```
Does that make sense? 🤔

ancalita: I see, so we stick with the current check for `LOOP_REJECTED`

ka-bu: doh, it doesn't make sense because we affirm there instead of denying ... wishful thinking 😅

rasa/core/actions/forms.py
ka-bu: yep

changelog/10235.bugfix.md
```diff
@@ -0,0 +1,3 @@
+Fix ghost slot issues after a restart. Previously events were searched from
```
joejuzl, line 1: I would use more boring, descriptive language here as "ghost slots" might not mean anything to most users. e.g. "Duplicate slots after restart".

rasa/core/actions/forms.py
joejuzl: Yeah - fine to look at it in a separate issue/PR I think.

rasa/core/actions/forms.py
```diff
@@ -524,7 +529,10 @@ async def _validate_if_required(
            - form validation was not cancelled
         """
         # no active_loop means that it is called during activation
```
joejuzl, line 531: Is this comment still valid?

rasa/shared/core/events.py
```diff
@@ -495,6 +495,7 @@ def full_retrieval_intent_name(self) -> Optional[Text]:
         """Returns full retrieval intent name or `None` if no retrieval intent."""
         return self.intent.get(FULL_RETRIEVAL_INTENT_NAME_KEY)
 
+    # Note that this means two UserUttered events in different times will be equal.
```
joejuzl, line 498:
```suggestion
    # Note that this means two UserUttered events with the same text, intent and entities but _different_ timestamps will be considered equal.
```

tests/core/actions/test_forms.py
```diff
@@ -1625,3 +1634,60 @@ async def test_action_extract_slots_custom_mapping_with_condition():
...
+            Restarted(),
+            ActiveLoop("some_form"),
+            SlotSet(REQUESTED_SLOT, "name"),
+            UserUttered("hi", intent={"name": "greet", "confidence": 1.0}),
+            ActionExecuted(ACTION_LISTEN_NAME),
+        ],
+    )
+    action = FormAction("some_form", None)
+
+    with pytest.raises(ActionExecutionRejection):
```
joejuzl, line 1687: Could we have a comment explaining why we expect this? I guess it's because we restart so it resets the slots so there are still required slots to be filled?

rasa/core/actions/forms.py
tayfun: I think so, if there's no active loop => form is not active => no events from form returned? @ancalita can you confirm?

ancalita: Maybe we can improve the comment: `no active_loop means there are no form filled slots to validate yet`?

joejuzl: Yeah I was a bit confused as the behaviour changed but the comment stayed the same. I think Anca's suggestion of comment sounds good.

tayfun: Makes sense 👍

joejuzl (approved):
💯

rasa/core/actions/forms.py
```diff
@@ -523,8 +528,11 @@ async def _validate_if_required(
            - the form is called after `action_listen`
            - form validation was not cancelled
         """
-        # no active_loop means that it is called during activation
-        needs_validation = not tracker.active_loop or (
+        # No active_loop means there are no form filled slots to validate yet
+        if not tracker.active_loop:
+            return []
+
+        needs_validation = (
```
melindaloubser1, line 535: @tayfun I have a retrospective question about this change.
I _think_ it's the thing that is causing the following behaviour between 2.x and 3.x to change: In 2.x, it was possible to add "form entry" logic to `FormValidationAction` by doing a check on whether `requested_slot` was `None`. Now that the form validation action is guaranteed not to run until the loop is active, there's nowhere to "hook into" the form before it starts.
Do you know if this was a conscious decision or is it an unintended side-effect?

tayfun: I think it was a side effect, though I'm not 100% sure about slots feature in general. This does feel like a hack for "form entry" logic though, I wonder if there is another way so that form validation action can distinguish form entry or if we should add a new explicit way. @ancalita thoughts?

ancalita: Oh this was to remove the code between L361-363 further up  in original `forms.py` + I also think it's confusing to allow a `FormValidationAction` to run when the form is not active yet / just being activated

ancalita: I remember adding that L361 code to prevent some unit tests from breaking.

melindaloubser1: It's true that there's no slots to validate yet, but since a `FormValidationAction` is there to validate form input in general, it does make sense that it could also "prepare the way" for form input when it has just been activated. What was the specific unwanted effect of letting the validation action run here? Maybe there's a way to mitigate that without completely disabling the form validation from running when it's just starting.

melindaloubser1: In general though I would advocate having a separate method like `run_on_activation` instead of having logic added directly to `run` with checks on `requested_slot` to see if it's the start of a form. That would need to be a bit more thought out though since it would be new behaviour.

melindaloubser1: Request opened [here](https://github.com/RasaHQ/rasa/issues/10913)

PR #10352, opened 2021-11-22
add warning log asking user to update training data version

### LINKED ISSUE #10288: Add warning or error for incorrect training data version
Context [here](https://rasa-hq.slack.com/archives/C01HHMR4X8S/p1637313176350300)
Further Slack discussion [here](https://rasa-hq.slack.com/archives/C01HHMR4X8S/p1637333261368900)

For users who have not used `rasa data migrate` and have manually created or updated a domain that follows the 3.0 format (slot mappings defined in slots, form required_slots is a list) **but** who have not updated the version key in the `domain.yml`, everything will run smoothly without any prompt/ warnings for them to update version key too.

We want a warning when the user runs `rasa data validate` and / or `rasa train` with Rasa Open Source 3.0 and the version in training data files is not at least as high as the LATEST_TRAINING_DATA_FORMAT_VERSION.

carlad:
**Proposed changes**:
Closes #10288
- This adds a log warning in the case where a user has manually migrated their domain and data files to the latest Rasa Open Source version format, but is still specifying a lower version in these files. eg: using a rasa 3.0 format, but still specifying `version: "2.0"`. 
- This warning will be logged during the validation step, asking the user to update the version in their domain or data files to the LATEST_TRAINING_DATA_FORMAT_VERSION.

Screenshot 2021-11-22 at 12 55 36

### DIFF
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -247,6 +247,17 @@ def validate_training_data_format_version(
             raise TypeError
 
         if version.parse(LATEST_TRAINING_DATA_FORMAT_VERSION) >= parsed_version:
+            if version_value < LATEST_TRAINING_DATA_FORMAT_VERSION:
+                # not raising here since it's not critical
+                logger.warning(
+                    f"Training data file {filename} has a lower "
+                    f"format version than your Rasa Open Source installation: "
+                    f"{version_value} < {LATEST_TRAINING_DATA_FORMAT_VERSION}. "
+                    f"Rasa Open Source will read the file as a version "
+                    f"{LATEST_TRAINING_DATA_FORMAT_VERSION} file but "
+                    f"please update your version key to 3.0. "
+                    f"See {DOCS_URL_TRAINING_DATA}."
+                )
             return True
 
     except TypeError:
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -255,8 +255,8 @@ def validate_training_data_format_version(
                     f"{version_value} < {LATEST_TRAINING_DATA_FORMAT_VERSION}. "
                     f"Rasa Open Source will read the file as a version "
                     f"{LATEST_TRAINING_DATA_FORMAT_VERSION} file but "
-                    f"please update your version key to 3.0. "
-                    f"See {DOCS_URL_TRAINING_DATA}."
+                    f"please update your version key to "
+                    f"{LATEST_TRAINING_DATA_FORMAT_VERSION}. See {DOCS_URL_TRAINING_DATA}."
                 )
             return True
 
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -256,7 +256,8 @@ def validate_training_data_format_version(
                     f"Rasa Open Source will read the file as a version "
                     f"{LATEST_TRAINING_DATA_FORMAT_VERSION} file but "
                     f"please update your version key to "
-                    f"{LATEST_TRAINING_DATA_FORMAT_VERSION}. See {DOCS_URL_TRAINING_DATA}."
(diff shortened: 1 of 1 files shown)


tayfun (approved):
There's a line length check failing, other than that 👍

rasa/shared/utils/validation.py
```diff
@@ -247,6 +247,18 @@ def validate_training_data_format_version(
             raise TypeError
 
         if version.parse(LATEST_TRAINING_DATA_FORMAT_VERSION) >= parsed_version:
+            if version_value < LATEST_TRAINING_DATA_FORMAT_VERSION:
+                # not raising here since it's not critical
+                logger.warning(
```
ancalita, line 252: I would suggest raising a `UserWarning` using  `rasa.shared.utils.io.raise_warning` (according to the convention [here](https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#6d4bea523bef4b18b200f47988087d4e)).

rasa/shared/utils/validation.py
```diff
@@ -247,6 +247,18 @@ def validate_training_data_format_version(
...
+                    f"Training data file {filename} has a lower "
+                    f"format version than your Rasa Open Source installation: "
+                    f"{version_value} < {LATEST_TRAINING_DATA_FORMAT_VERSION}. "
+                    f"Rasa Open Source will read the file as a version "
+                    f"{LATEST_TRAINING_DATA_FORMAT_VERSION} file but "
```
ancalita, line 257:
```suggestion
                    f"{LATEST_TRAINING_DATA_FORMAT_VERSION} file. "
```

rasa/shared/utils/validation.py
```diff
@@ -247,6 +247,18 @@ def validate_training_data_format_version(
             raise TypeError
 
         if version.parse(LATEST_TRAINING_DATA_FORMAT_VERSION) >= parsed_version:
+            if version_value < LATEST_TRAINING_DATA_FORMAT_VERSION:
```
ancalita, line 250: I think this should be a standalone if statement, not a nested one, so should be placed just beneath this one and return `False`.

ancalita (commented):
Left a few suggestions, mostly good to go 💯 
Have you tested with `rasa train` too?

ancalita:
Ohh, didn't realise you've merged this PR, moved a bit slow I guess. Nevertheless could you have a look at the comments I left? @carlad

rasa/shared/utils/validation.py
carlad: I considered that, but then the output wasn't coloured. So I used the convention used on this line:
https://github.com/RasaHQ/rasa/blob/main/rasa/shared/utils/validation.py#L233
but am happy to update.

carlad:
> Have you tested with rasa train too?
yes :)

rasa/shared/utils/validation.py
carlad: If it returns False, what then happens? Won't the validation fail, and the training not run?

carlad: Shouldn't it simply not return anything? If we return False, won't the validation fail and the training not run?

carlad: Yes, that is what happens:
Screenshot 2021-11-22 at 15 08 37

I'll just have it return nothing.
This is what the output now looks like:
Screenshot 2021-11-22 at 15 10 03

Screenshot 2021-11-22 at 15 10 45

rasa/shared/utils/validation.py
ancalita: I guess it depends on the shell you're using, I'm using `zsh` and UserWarnings appear bright yellow or orange 😄

ancalita: Sorry, meant to say this: https://ohmyz.sh/

PR #10356, opened 2021-11-22
10288 training data version warning

carlad:
**Proposed changes**:
Improvements based on @ancalita's comments on https://github.com/RasaHQ/rasa/pull/10352

### DIFF
--- a/rasa/shared/utils/validation.py
+++ b/rasa/shared/utils/validation.py
@@ -241,24 +241,25 @@ def validate_training_data_format_version(
         )
         return True
 
+    if str(version_value) < LATEST_TRAINING_DATA_FORMAT_VERSION:
+        rasa.shared.utils.io.raise_warning(
+            f"Training data file {filename} has a lower "
+            f"format version than your Rasa Open Source installation: "
+            f"{version_value} < {LATEST_TRAINING_DATA_FORMAT_VERSION}. "
+            f"Rasa Open Source will read the file as a version "
+            f"{LATEST_TRAINING_DATA_FORMAT_VERSION} file. "
+            f"Please update your version key to "
+            f"{LATEST_TRAINING_DATA_FORMAT_VERSION}. "
+            f"See {DOCS_URL_TRAINING_DATA}."
+        )
+
     try:
         parsed_version = version.parse(version_value)
         if isinstance(parsed_version, LegacyVersion):
             raise TypeError
 
         if version.parse(LATEST_TRAINING_DATA_FORMAT_VERSION) >= parsed_version:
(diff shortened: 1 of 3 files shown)


rasa/shared/utils/validation.py
```diff
@@ -241,24 +241,25 @@ def validate_training_data_format_version(
         )
         return True
 
+    if str(version_value) < LATEST_TRAINING_DATA_FORMAT_VERSION:
```
ancalita, line 244: I think this should actually come after L260, just before L261 `if`, to prioritise raising the `TypeError` if a legacy version is being.

ancalita (commented):
Thank you for looking into the suggestions 🙏🏼

rasa/shared/utils/validation.py
```diff
@@ -246,19 +246,20 @@ def validate_training_data_format_version(
...
+        if str(version_value) < LATEST_TRAINING_DATA_FORMAT_VERSION:
+            rasa.shared.utils.io.raise_warning(
+                f"Training data file {filename} has a lower "
+                f"format version than your Rasa Open Source installation: "
+                f"{version_value} < {LATEST_TRAINING_DATA_FORMAT_VERSION}. "
+                f"Rasa Open Source will read the file as a version "
+                f"{LATEST_TRAINING_DATA_FORMAT_VERSION} file. "
+                f"Please update your version key to "
+                f"{LATEST_TRAINING_DATA_FORMAT_VERSION}. "
+                f"See {DOCS_URL_TRAINING_DATA}."
```
ancalita, line 258: Just noticed this docs link reference, could you have a look to see if we need to insert a comment that the training data version in `3.0` should be updated to `3.0`?

ancalita: in that docs page, I mean, sorry keep pressing enter without finishing my comment 😅

ancalita (approved):
🎉

rasa/shared/utils/validation.py
```diff
@@ -246,19 +246,20 @@ def validate_training_data_format_version(
         if isinstance(parsed_version, LegacyVersion):
             raise TypeError
 
+        if str(version_value) < LATEST_TRAINING_DATA_FORMAT_VERSION:
```
joejuzl, line 249: Shouldn't we be using `version.parse` for this check?

carlad: we can actually use `parsed_version` (which is `version.parse(version_value)`)

carlad: PR: https://github.com/RasaHQ/rasa/pull/10366

PR #10360, opened 2021-11-23
Log when loading model.

joejuzl:
**Proposed changes**:
- Log when loading a model like we did in 2.8.x

### DIFF
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -93,18 +93,22 @@ class MessageProcessor:
         self.max_number_of_predictions = max_number_of_predictions
         self.on_circuit_break = on_circuit_break
         self.action_endpoint = action_endpoint
+        self.model_filename = ""
         self.model_metadata, self.graph_runner = self._load_model(model_path)
         self.model_path = Path(model_path)
-        self.model_filename = os.path.basename(get_latest_model(Path(model_path)) or "")
         self.domain = self.model_metadata.domain
         self.http_interpreter = http_interpreter
 
-    @staticmethod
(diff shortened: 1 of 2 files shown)


rasa/core/processor.py
```diff
@@ -93,19 +93,23 @@ def __init__(
...
+    def _load_model(
+        self, model_path: Union[Text, Path]
+    ) -> Tuple[ModelMetadata, GraphRunner]:
         """Unpacks a model from a given path using the graph model loader."""
         model_tar = get_latest_model(model_path)
         if not model_tar:
             raise ModelNotFound(f"No model found at path '{model_path}'.")
 
+        logger.info(f"Loading model {model_tar}...")
+        self.model_filename = os.path.basename(model_tar)
```
joejuzl, line 111: I moved this here so we don't have to do `get_latest_model` twice which globs the whole directory.

rasa/core/processor.py
```diff
@@ -93,19 +93,23 @@ def __init__(
         self.max_number_of_predictions = max_number_of_predictions
         self.on_circuit_break = on_circuit_break
         self.action_endpoint = action_endpoint
+        self.model_filename = ""
         self.model_metadata, self.graph_runner = self._load_model(model_path)
```
ka-bu, line 97: nit: how about we let the load return the model name just like the other of the loaded assets? (then setting the filename is not so much an implicit side effect)
```suggestion
        self.model_filename, self.model_metadata, self.graph_runner = self._load_model(model_path)
```

ka-bu (approved):

rasa/core/processor.py
joejuzl: Done 👍

PR #10362, opened 2021-11-23
Load model by full path correctly.

joejuzl:
**Proposed changes**:
We were always loading the latest model in a directory even if given the full path to the model archive. This fixes this so now we will load directly in the path is a file.

### DIFF
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -93,27 +93,32 @@ class MessageProcessor:
         self.max_number_of_predictions = max_number_of_predictions
         self.on_circuit_break = on_circuit_break
         self.action_endpoint = action_endpoint
-        self.model_metadata, self.graph_runner = self._load_model(model_path)
+        self.model_filename, self.model_metadata, self.graph_runner = self._load_model(
+            model_path
+        )
         self.model_path = Path(model_path)
-        self.model_filename = os.path.basename(get_latest_model(Path(model_path)) or "")
         self.domain = self.model_metadata.domain
         self.http_interpreter = http_interpreter
 
     @staticmethod
-    def _load_model(model_path: Union[Text, Path]) -> Tuple[ModelMetadata, GraphRunner]:
+    def _load_model(
+        model_path: Union[Text, Path]
+    ) -> Tuple[Text, ModelMetadata, GraphRunner]:
         """Unpacks a model from a given path using the graph model loader."""
         model_tar = get_latest_model(model_path)
         if not model_tar:
             raise ModelNotFound(f"No model found at path '{model_path}'.")
 
+        logger.info(f"Loading model {model_tar}...")
         with tempfile.TemporaryDirectory() as temporary_directory:
             try:
-                return loader.load_predict_graph_runner(
+                metadata, runner = loader.load_predict_graph_runner(
                     Path(temporary_directory),
                     Path(model_tar),
                     LocalModelStorage,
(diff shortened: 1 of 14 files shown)


tayfun (approved):

rasa/core/processor.py
```diff
@@ -105,9 +105,12 @@ def _load_model(
...
         """Unpacks a model from a given path using the graph model loader."""
-        model_tar = get_latest_model(model_path)
-        if not model_tar:
-            raise ModelNotFound(f"No model found at path '{model_path}'.")
+        if os.path.isfile(model_path):
+            model_tar = model_path
+        else:
+            model_tar = get_latest_model(model_path)
+            if not model_tar:
+                raise ModelNotFound(f"No model found at path '{model_path}'.")
```
wochinge, line 113: should this be a separate method? In my opinion we should avoid that we have a similar bug in other modules using `get_latest_model`

joejuzl: I was going to change `get_latest_model` but looking at it, its pretty clear in its function - I think I just miss-used it here.

ka-bu: Could we just raise in `get_latest_model` if what is passed is already a file? I don't find that behaviour of falling back to the dirname in that case is that obvious

joejuzl: > I don't find that behaviour of falling back to the dirname in that case is that obvious

I also think the behaviour is weird. However I didn't want to affect other uses of this method in the codebase this close to the release. So I opted to just fixing how we load in the `Processor`.

melindaloubser1:
I think this is also what caused the replace model endpoint to reload the latest model instead of the one at the provided path - sound right? https://github.com/RasaHQ/rasa/issues/10148#issuecomment-969350009

joejuzl:
> I think this is also what caused the replace model endpoint to reload the latest model instead of the one at the provided path - sound right? #10148 (comment)

Yes that makes sense 👍

tests/core/test_processor.py
```diff
@@ -1496,3 +1498,26 @@ async def test_model_name_is_available(trained_moodbot_path: Text):
     processor = Agent.load(model_path=trained_moodbot_path).processor
     assert len(processor.model_filename) > 0
     assert "/" not in processor.model_filename
+
+
+async def test_loads_correct_model_from_path(
+    trained_core_model: Text, trained_nlu_model: Text, tmp_path
```
wochinge, line 1504:
```suggestion
    trained_core_model: Text, trained_nlu_model: Text, tmp_path: Path
```

rasa/core/processor.py
```diff
@@ -105,9 +105,15 @@ def _load_model(
         model_path: Union[Text, Path]
     ) -> Tuple[Text, ModelMetadata, GraphRunner]:
         """Unpacks a model from a given path using the graph model loader."""
-        model_tar = get_latest_model(model_path)
-        if not model_tar:
-            raise ModelNotFound(f"No model found at path '{model_path}'.")
+        try:
+            if os.path.isfile(model_path):
+                model_tar = model_path
```
wochinge, line 110: Does this change require adding / updating a test?

wochinge (approved):

rasa/core/processor.py
joejuzl: the new test `test_loads_correct_model_from_path` breaks before this is added.

PR #6960, opened 2020-10-07
Annotation test

tmbo:
**Proposed changes**:
- use github annotations to create annotations for failed pytest tests and flake8 errors / warnings
- makes better use of the github UI -> annotations will be shown in the files changed view (reduces the need to dig through the logs)

Example:
Bildschirmfoto 2021-03-24 um 15 36 06

PR #10376, opened 2021-11-24
add regex validation to pattern utils

carlad:
**Proposed changes**:
addresses the OSS side of #6960

This adds regex validation to the `extract_pattern` method of `rasa/nlu/utils/pattern_utils.py` and a corresponding test.
I'm not totally sure about the error message, eg:
```
 "Model training failed. '*[0-9]{5}' is not a valid regex. Please update your nlu training data configuration at {'name': 'zipcode', 'pattern': '*[0-9]{5}'}."
```

Discussion concluded that training should fail and a readable error message displayed, however I'm not sure if this message gives the user enough info about where this config is.

### DIFF
--- a/rasa/nlu/utils/pattern_utils.py
+++ b/rasa/nlu/utils/pattern_utils.py
@@ -3,6 +3,7 @@ from typing import Dict, List, Text, Union
 
 import rasa.shared.utils.io
 from rasa.shared.nlu.training_data.training_data import TrainingData
+from rasa.shared.exceptions import InvalidConfigException
 
 
 def _convert_lookup_tables_to_regex(
@@ -152,5 +153,16 @@ def extract_patterns(
                 training_data, use_only_entities, use_word_boundaries
             )
         )
+
+    # validate regexes, raise Error when invalid
+    for pattern in patterns:
+        try:
+            re.compile(pattern["pattern"])
+        except re.error:
+            raise InvalidConfigException(
+                f"Model training failed. '{pattern['pattern']}' "
+                "is not a valid regex. Please update your nlu "
+                f"training data configuration at {pattern}."
+            )
 
     return patterns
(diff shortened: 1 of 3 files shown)


rasa/nlu/utils/pattern_utils.py
```diff
@@ -153,4 +154,15 @@ def extract_patterns(
             )
         )
 
+    # validate regexes, raise Error when invalid
+    for pattern in patterns:
+        try:
+            re.compile(pattern["pattern"])
```
tayfun, line 160: Compiling regex could be slowish and [Python recommends reusing regex objects](https://docs.python.org/3/library/re.html#re.compile). Would it work to catch the exception during later compilation while training and raise exception? I mean in regex_featurizer is compiling anyway, catch exception there and raise. Not sure if we lose pattern name etc. over there? Otherwise this PR is fine and the function here would validate albeit run a little slower.

tests/nlu/utils/test_pattern_utils.py
```diff
@@ -181,3 +181,40 @@ def test_extract_patterns_use_only_lookup_tables_or_regex_features(
...
+    with pytest.raises(Exception) as e:
+        pattern_utils.extract_patterns(
+            training_data,
+            use_lookup_tables=use_lookup_tables,
+            use_regexes=use_regex_features,
+        )
+
+    assert "Model training failed." in str(e.value)
+    assert "not a valid regex." in str(e.value)
+    assert "Please update your nlu training data configuration" in str(e.value)
```
tayfun, line 220: 👍 thinking out loud, e is still in scope here so this works

rasa/nlu/utils/pattern_utils.py
carlad: The regex_featurizer compiles when calling `re.finditer`. This method takes a few parameters which messes with the error returned.
I agree that compiling more than once is less efficient, but I feel like it is more logical to validate on the utility side.

PR #10380, opened 2021-11-24
Add Datadog system agent

markus-hinsche:
**Motivation:** For the benchmarking project, we want to collect CPU utilization and memory usage. Tracking these metrics manually introduces a lot of custom code (e.g. `mprof`, `nvidia-smi`, `top`) cluttering the test. To avoid this, we can have a datadog agent running in the background which reports the metrics every so many seconds. 

**Proposed changes**:
- add Datadog agent to send system (e.g. CPU+memory) metrics to Datadog
- introduce a bash script that can be called from Github Actions yaml (avoid code duplication)
- add NVML integration (if `ACCELERATOR_TYPE=GPU`)


.github/scripts/start_dd_agent.sh
```diff
@@ -0,0 +1,28 @@
...
+DD_API_KEY=$1
+dataset=$2
+config=$3
+
+echo "Dataset: ${dataset}"
+echo "Config: ${config}"
+
+# Install Datadog system agent
+DD_AGENT_MAJOR_VERSION=7 DD_API_KEY=$DD_API_KEY DD_SITE="datadoghq.eu" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
+DATADOG_YAML_PATH=/etc/datadog-agent/datadog.yaml
```
tczekajlo, line 12: Be sure that APM and Process agent are off:

```
process_config:
    enabled: false
apm_config:
    enabled: false
```

.github/scripts/start_dd_agent.sh
```diff
@@ -0,0 +1,28 @@
...
+{
+    echo "env: rasa-regression-tests"
+    echo "tags:"
+    echo "- service:rasa"
+    echo "- dataset:$dataset"
+    echo "- config:$config"
+} >> $DATADOG_YAML_PATH
+
+# Enable system_core integration
+sudo mv /etc/datadog-agent/conf.d/system_core.d/conf.yaml.example /etc/datadog-agent/conf.d/system_core.d/conf.yaml
```
tczekajlo, line 25: it looks like you don't enable the integration for GPU, I guess it's something that we want to use, right?

.github/workflows/ci-model-regression-on-schedule.yml
```diff
@@ -217,6 +217,11 @@ jobs:
           echo $DATASET_COMMIT
           echo "::set-output name=dataset_commit::$DATASET_COMMIT"
 
+      - name: Run Datadog Agent
+        if: steps.set_dataset_config_vars.outputs.is_dataset_exists == 'true' && steps.set_dataset_config_vars.outputs.is_config_exists == 'true'
+        run: |
+          .github/scripts/start_dd_agent.sh ${{ secrets.DD_API_KEY }} '${{ matrix.dataset }}' '${{ matrix.config }}'
```
tczekajlo, line 223: Be sure that at the end of workflow we stop the Agent (`service datadog-agent stop`)

tczekajlo (changes requested):

.github/scripts/start_dd_agent.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Install Datadog system agent
+DD_AGENT_MAJOR_VERSION=7 DD_API_KEY=$DD_API_KEY DD_SITE="datadoghq.eu" bash -c "$(curl -L https://s3.amazonaws.com/dd-agent/scripts/install_script.sh)"
+DATADOG_YAML_PATH=/etc/datadog-agent/datadog.yaml
+sudo chmod 666 $DATADOG_YAML_PATH
+
+# Associate metrics with tags and env
+{
+    echo "env: rasa-regression-tests"
+    echo "tags:"
```
tczekajlo, line 18: I think it'd be useful to add a PR number, git sha, and branch name as tags.

markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

.github/scripts/start_dd_agent.sh
virtualroot: You can try with `${{ github.event.number }}` for getting the PR ID.

virtualroot (approved):

.github/scripts/start_dd_agent.sh
markus-hinsche: Thanks for the suggestion. I will keep this in mind for the future, but at the moment I am not sure about all the implications:
The challenge with the PR number is to make ensure this also works when the MR test haven't been triggered by a PR, but some other event (push, or schedule). The datadog tag can't be empty, so maybe in this case we need a default value.

virtualroot: Good catch. In that scenario, we could try to assign "schedule" as default when `$PR_ID` is unset or empty. Could that work?

```
    echo "- pr_id:${PR_ID:-schedule}"
```

markus-hinsche: That should work, I added the PR_ID tag

markus-hinsche:
This is ready to be merged from my side, but I can't merge yet because @tczekajlo requested changes

PR #10382, opened 2021-11-24
Merge 2.8.x into main

rasabot:
:auto_rickshaw: This PR should be merged automatically once it has been approved. If it doesn't happen:
- [ ] Handle merge conflicts
- [ ] Fix build errors

:bulb: It has been opened automatically after changes were merged in a feature branch.


pyproject.toml
```diff
@@ -108,7 +108,7 @@ SQLAlchemy = ">=1.3.3,<1.5.0"
 sklearn-crfsuite = "~0.3"
 psycopg2-binary = ">=2.8.2,<2.10.0"
 python-dateutil = "~2.8"
-tensorflow = ">=2.6,<2.7"
+tensorflow = "~2.6.2"
```
tttthomasssss, line 111: The `tensorflow` and `google-auth` dependencies have been set to the respective versions b/c of a micro release after a customer reported problems.

This is just to double check whether we want to merge these back into `main` (b/c idk)?

cc @hsm207 @tayfun

tttthomasssss (commented):
Would be great to have sb double-check whether the version changes for `tensorflow` and `google-auth` should be merged back into main.

Otherwise LGTM.

pyproject.toml
hsm207: In my opinion, this should be merge to main because if we don't do it, customers will report the same tensorflow vulnerabilities when they upgrade to 3.x.

PR #10393, opened 2021-11-25
Pin GitHub Actions to Commit Hash

rasa-jmac:
**Proposed changes**:
- Pin all third-party GitHub Actions to the commit hash that corresponds to the versions we use.
- Updated the action `marocchino/sticky-pull-request-comment` from 2.1.1 to 2.2.0 to pin to a signed version.
- This is a best practice for GitHub Actions and ensures that if one of the third party actions is compromised in the future and retagged, we won't execute compromised code during our build process.


.github/workflows/ci-model-regression.yml
```diff
@@ -692,7 +692,7 @@ jobs:
           echo "::set-output name=result::$OUTPUT"
 
       - name: Publish results as a PR comment
-        uses: marocchino/sticky-pull-request-comment@v2.1.1
+        uses: marocchino/sticky-pull-request-comment@39c5b5dc7717447d0cba270cd115037d32d28443 # v2.2.0
```
erohmensing, line 695: Intentional upgrade? (also other uses of this action have been upgraded)

erohmensing (approved):
LGTM, unfortunately I don't think I can approve! 

Is there a standard to which libraries we pin to which specificity? e.g. `#v7` `#v2.0`, `#v1.6.0` -- I think that dependabot uses these as signals of what it can automatically open PRs to upgrade (though i've not confirmed this)

.github/workflows/ci-model-regression.yml
rasa-jmac: Yeah I've intentionally upgraded this one - dependabot updated it recently and it pinned it to a version tag that didn't have a signed commit, where other tags did. Just seemed a bit weird so I bumped it to the next signed released tag.

rasa-jmac:
Hey @RASADSA could you take a look at this when you get a chance please?

rasa-jmac:
@RASADSA could you check if you're happy with this please? I know it's super old but I do still want it merged

RASADSA (approved):
LGTM

Issue #6712, opened 2020-09-18
Entity detection in End to End testing fails when multiple entity extractors both extract the same entity


Plurno:
**Rasa version**: 1.10.9 **Rasa SDK version** (if used & relevant):

**Rasa X version** (if used & relevant):

**Python version**: 3.7.7 **Operating system** (windows, osx, ...): Debian GNU/Linux 10 (buster) **Issue**: Here is an example of an End to End test that I put in my tests/conversation_tests.md file.

1.  1.  Foreign language story

*   greet: hi

*   action_greet
*   rewind

*   ask_book: Are there any books in <span class="error">[french]</span>(foreign_language)?

*   utter_ask_book

When I run rasa test, I get the following result in results/failed_stores.md:

1.  1.  Foreign language story

*   greet: hi

*   action_greet
*   rewind

*   ask_book: Are there any books in <span class="error">[french]</span>(foreign_language)? \

*   utter_ask_book

I think what’s happening is that I have two entity extractors (DIETClassifier and CRFEntityExtractor) that are both independently finding the ‘french’ foreign_language entity, so it gets duplicated when the test is run. It's not a huge deal, but it's impacting my CI/CD pipeline a bit.

Thanks

**Error (including full traceback)**:

 No runtime error, just a failed line in a failed_stories.md

**Command or request that led to error**:

 rasa test

**Content of configuration file (config.yml)** (if relevant):

**Content of domain file (domain.yml)** (if relevant):

sara-tagger:
Thanks for the issue, @erohmensing will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

samsucik:
Hi @Plurno, could you share your config file (or at least the relevant parts of it)? If you suspect that this is due to 2 extractors extracting the same entity, have you tried disabling one of them (let's say Duckling)? Do you intentionally use both for the `foreign_language` entity? I hope I'll be better able to help once I know more :-)

carlosmeds:
Hi there! I had the same issue here and I kind solved it disabling CRFEntityExtractor. I think it has something to do with @samsucik comment.

PR #10394, opened 2021-11-25
Allow multiple entities for one token

raoulvm:
**Proposed changes**:
Allow multiple entities to be annotated for the same word/tokens.

When using entity extractors that support generating multiple entities from a single expression, the test stories fail as there is no way to annotate multiple entity_types and entity_values

New annotation option is

```YAML
stories:
  - story: Some story
    steps:
      - user: |
          I would like work for [Rasa][{"entity":"chatbot_software", "value":"rasa"},{"entity":"software_company", "value":"rasa"},{"entity":"employer", "value":"rasa"}]
        intent: apply_job
```
As the extractors as pipeline components neither have access to _domain_ during training, nor to the dialog _tracker_ during inference, they can't judge which _characteristic_ (or _attribute_) of the word "Rasa" is important for answering a question. If the entity extractor could access the _domain_ knowledge, it would probably be able to filter the possible entity types to those defined with `use_entities` in the domain (e.g. employer).

If it could access the tracker, and the preceding question was a closed question, the possible answers could also limit the applicable entity types. 

In this PR I have changed the training data parser to be able to cope with the format shown above, and also changed the readerwriter.py to be able to generate such training data messages if the entities found have exact same start and end positions. 


raoulvm:
@ancalita Do you know why I get linting errors in "old" code? And if I change that, some other old lines are complained about.
Thanks
Raoul

raoulvm:
Hi @samsucik I was told by @hsm207 that you are the dev on duty today.
Could you explain why `extractors` is `None` (and thus it [returns True here](https://github.com/RasaHQ/rasa/pull/10394/files#diff-8d9591f26ceb08200d3364a40b1b25b9403f13501efd9dd8922ff1c2e47901c3L1092)) if it was `[CRFEntityExtractor.name]` [provided here](https://github.com/RasaHQ/rasa/blob/2.8.x/tests/nlu/test_evaluation.py#L203)? 

I do not understand what is happening here.

Thanks
Raoul

ancalita:
> After changing that globally I could run `poetry run flake8 rasa test --extend-ignore D` without getting old code complained about.

@raoulvm Great! In addition, you can also use `make lint` [command](https://github.com/RasaHQ/rasa/blob/2c17cd24003e2cc9e23276e02b32757445867774/Makefile#L82) in your Terminal for a more comprehensive check.

samsucik:
@raoulvm I checked your changes now. I'm glad you found a workaround for the failing test. And 'm excited about the proposed improvement, thanks a lot for that!

As for going ahead with this, I have to think about it a bit more and discuss with others because it'd likely have a broader impact. Some points that come to my mind are:
- When an entity can be "multi-annotated", the validity of this annotation is strongly coupled with the models used. I.e. it'd be invalid if we only have DIET in the pipeline (or a different extractor that assigns only a single entity label to a token), but would be valid if we've got a RegexExtractor.
- Once an entity can be "multi-annotated" in test stories, a natural follow-up is "why not in training stories?". And that leads us to the thick line drawn between extractors like DIET vs ones like RegexExtractor -- namely, the latter one not requiring annotations in training data. Additionally, we'd be also drawing another line, this time between training and test story format.
- The exact annotation format needs to work with all special cases, i.e. we'd need to triple-check the regexes (one immediate candidate edge case to check is when a language doesn't have words separated by spaces and you've got 2 entities coming one right after another). But this is ultimately just a technicality.
- If we decide to go with the change, then it'll have to become part of 3.x sooner or later, which means we want to be certain about it.

raoulvm:
HI @ancalita , you should consider cleaning "old stuff" anyhow. If you run the lint against the whole repo (no diff) you get 3000+ doc-string complaints.
```shell
(rasa2source) #:~/PyProjects/rasa$ poetry run flake8 rasa tests --select D | wc -l
3720
```

raoulvm:
Hi @samsucik 

> it'd be invalid if we only have DIET in the pipeline. 

I just ran a test with the mood bot including new _multi-annotated_ entities in the NLU Training Data , with DIET as the entity extractor. DIET is training on the last item in the list only.

> Once an entity can be "multi-annotated" in test stories, a natural follow-up is "why not in training stories?"  

Good question. It should.

> And that leads us to the thick line drawn between extractors like DIET vs ones like RegexExtractor -- namely, the latter one not requiring annotations in training data.  

Actually that line is already there. There are intent training data _containing_ entities, and there are Regex Expressions and Lookups, which are separated. If you include other ontological tools like mine it is separated even more, as those data do not fit into the nlu files at all.

For the other details, you are perfectly right: I do not speak any non-whitespace language, and I am no fan of regex for literal lookups (regards to @koaning for his splendid flashtext extractor idea which I used as a basis for the hierarchies) so I wasn't very deep into those and did not make test cases for them beyond what is already in the tests.

samsucik:
Thanks for digging into the questions so quickly. As you can see, there are some rough edges to overcome if this was to become part of 2.8 (and later 3.x) and, more importantly, if these changes were to be maintained and further built upon in future iterations.

Regarding the line between DIET and RegexExtractor: Indeed, that line already exists (sorry for not being 100% clear in my previous reply). But, in my opinion, we should try to not make these arbitrary lines even worse.

All in all, I think it's important that you're exploring how far you can go with 2.8. The other part is then taking a step back and seeing what's the high-level problem that you're trying to solve, as we discussed previously 🙂

raoulvm:
Test with 3.0 see here #10404

JEM-Mosig:
We should use this also to resolve [this issue](https://github.com/RasaHQ/rasa/issues/6712).

Issue #10138, opened 2021-11-10
Verify that Rasa OSS github project doesn't list py3.6 when 3.0 is out on pypi


joejuzl:
We dropped support for py36 so we shouldn't have it in the repo badge.

<img width="850" alt="Screenshot 2021-11-10 at 10 20 58" src="https://user-images.githubusercontent.com/1702538/141086016-558fadcf-3ad8-486b-a99e-b0467f02d55f.png">

However it may be the case that we can't change this because of old release branches - so needs to be looked into.

m-vdb:
@joejuzl as far as I can tell [the badge](https://github.com/RasaHQ/rasa/blame/main/README.md#L7) is going to be updated automatically when there is a newer version of `rasa` published on Pypi. I think the reason we're not seeing the badge updated right now is that the `3.0.0rc1` is a prerelease version, hence it's not treated as a stable latest version. **I will keep this ticket open, in the backlog for now and double check that it's updated when 3.0 is out.**

TyDunn:
@m-vdb Is there anything we can do here? I don't see it updated on the GitHub repo

m-vdb:
🤦🏻 apparently the Python 3.6 classifier was defined independently of the required python dependency.... I've put up a PR here: https://github.com/RasaHQ/rasa/pull/10405

m-vdb:
<img width="844" alt="Screenshot 2021-12-03 at 16 19 34" src="https://user-images.githubusercontent.com/1931615/144627049-0e379b9c-4652-4aec-88ef-66e57e1a4d49.png">

It is fixed ✅  (will clean things up on the SDK though)

PR #10405, opened 2021-11-26
Remove 3.6 classifier in pyproject.toml

m-vdb:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa/issues/10138


pyproject.toml
```diff
@@ -16,7 +16,7 @@ maintainers = [ "Tom Bocklisch <tom@rasa.com>",]
 homepage = "https://rasa.com"
 repository = "https://github.com/rasahq/rasa"
 documentation = "https://rasa.com/docs"
-classifiers = [ "Development Status :: 4 - Beta", "Intended Audience :: Developers", "License :: OSI Approved :: Apache Software License", "Programming Language :: Python", "Programming Language :: Python :: 3.6", "Programming Language :: Python :: 3.7", "Programming Language :: Python :: 3.8", "Topic :: Software Development :: Libraries",]
+classifiers = [ "Development Status :: 4 - Beta", "Intended Audience :: Developers", "License :: OSI Approved :: Apache Software License", "Programming Language :: Python", "Programming Language :: Python :: 3.7", "Programming Language :: Python :: 3.8", "Topic :: Software Development :: Libraries",]
```
m-vdb, line 19: Actually [poetry doc](https://python-poetry.org/docs/pyproject/#classifiers) says:
> Note that Python classifiers are still automatically added for you and are determined by your python requirement.

So maybe we can just remove the python classifier. Still need to test this though 🤔

TyDunn (approved):
Looks good. I added doing this to the def of done for 3.9 support

pyproject.toml
```diff
@@ -16,7 +16,7 @@ maintainers = [ "Tom Bocklisch <tom@rasa.com>",]
 homepage = "https://rasa.com"
 repository = "https://github.com/rasahq/rasa"
 documentation = "https://rasa.com/docs"
-classifiers = [ "Development Status :: 4 - Beta", "Intended Audience :: Developers", "License :: OSI Approved :: Apache Software License", "Programming Language :: Python", "Programming Language :: Python :: 3.6", "Programming Language :: Python :: 3.7", "Programming Language :: Python :: 3.8", "Topic :: Software Development :: Libraries",]
```
m-vdb, line 19: the python classifiers are auto generated by [poetry](https://python-poetry.org/docs/pyproject/#classifiers):
> Note that Python classifiers are still automatically added for you and are determined by your python requirement.

pyproject.toml
```diff
@@ -16,7 +16,7 @@ maintainers = [ "Tom Bocklisch <tom@rasa.com>",]
...
+classifiers = [ "Development Status :: 4 - Beta", "Intended Audience :: Developers", "License :: OSI Approved :: Apache Software License", "Topic :: Software Development :: Libraries",]
```
m-vdb, line 19: @TyDunn I noticed that the development status of Rasa Open Source is still in "Beta". Should it rather be 5 - Production/Stable at some point? ([reference here](https://pypi.org/classifiers/))

m-vdb:
actually @TyDunn we don't need to do anything for Python 3.9, it is handled automatically by poetry (see my comment in the code)

pyproject.toml
m-vdb: couldn't find a way to test this locally though, there is no debug command in poetry to do that 🤔

joejuzl: so it will be added in again after a `poetry update`?

pyproject.toml
TyDunn: Yes. I think it should be 5 now

pyproject.toml
m-vdb: poetry doesn't update the `pyproject.toml`, my understanding is that when we build a new version of rasa to publish on pypi, poetry reads the [python requirement here](https://github.com/RasaHQ/rasa/blob/4fd90e7b55a6e354a7dfd16286e5fe7f11b53f52/pyproject.toml#L66) and updates the classifiers on the fly, so that they end up in pypi

joejuzl: aha ok - makes sense

joejuzl (approved):
💯

PR #10413, opened 2021-11-29
Make `RASA_ENVIRONMENT` available in events logged by Kafka producer

melindaloubser1:
**Proposed changes**:
- Add a `RASA_ENVIRONMENT` header to the Kafka event producer to make it available to consumers [as is done in the `app_id` for Pika](https://github.com/RasaHQ/rasa/blob/4fd90e7b55a6e354a7dfd16286e5fe7f11b53f52/rasa/core/brokers/pika.py#L297)
- Closes https://github.com/RasaHQ/rasa-x/issues/7010 in conjunction with https://github.com/RasaHQ/rasa-x/pull/7013

### DIFF
--- a/rasa/core/brokers/kafka.py
+++ b/rasa/core/brokers/kafka.py
@@ -1,3 +1,4 @@
+import os
 import json
 import logging
 from asyncio import AbstractEventLoop
@@ -8,6 +9,7 @@ from rasa.core.brokers.broker import EventBroker
 from rasa.shared.utils.io import DEFAULT_ENCODING
 from rasa.utils.endpoints import EndpointConfig
 from rasa.shared.exceptions import RasaException
+import rasa.shared.utils.common
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


rasa/core/brokers/kafka.py
```diff
@@ -198,7 +200,16 @@ def _publish(self, event: Dict[Text, Any]) -> None:
         logger.debug(
             f"Calling kafka send({self.topic}, value={event}, key={partition_key!s})"
         )
-        self.producer.send(self.topic, value=event, key=partition_key)
+
+        headers = [("RASA_ENVIRONMENT", bytes(self.rasa_environment, encoding=DEFAULT_ENCODING))]
```
joejuzl, line 204: Can we have a unit test for this?

melindaloubser1: Trying to figure out how to check this because it's not a property like it is for Pika - it only shows up in the headers when the event is sent. Is it necessary to mock a consumer to check what is sent, or do you think it can be mocked higher up?

joejuzl: If it requires everything to be mocked then probably not worth it. We really need some integration tests using an actual kafka.

melindaloubser1: Gotcha. So no unit tests on this PR then? All we can assert without mocks is that the environment var is fetched but that seems a little ¯\_(ツ)_/¯

joejuzl (approved):

PR #10424, opened 2021-11-30
add condition for additional entities

### LINKED ISSUE #10137: use_entities is broken when domain is split
### Rasa Open Source version

2.8.12

### Rasa SDK version

2.8.2

### Rasa X version

_No response_

### Python version

3.8

Linux

### What happened?

Create a domain.yml with the following content:

```yaml
intents:
  - greet:
      use_entities: [country]
  - goodbye:
      use_entities: []
  - affirm
  - deny
  - mood_great
  - mood_unhappy
  - bot_challenge
 
entities:
  - city
  - country
```

Create another domain yaml with the following content:

```yaml
intents:
  - inform:
      use_entities: [city]
```

Reading the merged domain files using `Domain.load` shows  that the intent `inform` is still using the entities `city` and `country`.

The workaround is to repeat the list of entities in the first domain file in the second domain file i.e.:

```
intents:
  - inform:
      use_entities: [city]

entities:
  - city
  - country
```

Here is the [code](https://github.com/hsm207/rasa_moodbot/tree/bug_split_domain) to reproduce this behavior.

### Command / Request

_No response_

### Relevant log output

_No response_

**Definition of done**
- [ ] Fix bug in 2.8 and 3.0
- [ ] Create integration test to ensure this works as expected
- [ ] Fixed merged into 2.8 and 3.0

carlad:
**Proposed changes**:
closes #10137

When loading a domain, the condition for when an entity was being used, but not explicitly defined in a domain file was ignored. This adds to and refactors the `_transform_intents_for_file` method to include a check for this condition to ensure the entity is loaded correctly.

**Status**:
- [x] added a changelog entry


rasa/shared/core/domain.py
```diff
@@ -1476,9 +1476,11 @@ def _transform_intents_for_file(self) -> List[Union[Text, Dict[Text, Any]]]:
                 if rasa.shared.core.constants.ENTITY_LABEL_SEPARATOR not in entity
             )
             ignore_entities = set(self.entities) - use_entities
-            if len(use_entities) == len(self.entities):
+            use_length = len(use_entities)
+            self_length = len(self.entities)
+            if use_length == self_length:
                 intent_props[USE_ENTITIES_KEY] = True
-            elif len(use_entities) <= len(self.entities) / 2:
+            elif use_length > self_length or use_length <= self_length / 2:
```
tayfun, line 1483: I might be wrong here but I think what we need to do here is to merge entities from all domain files to be able to use `intent_props[IGNORE_ENTITIES_KEY]`. Otherwise whenever we fall into the `else` 2 lines below, we'll be using a subset of ignored entities and it will be wrong, ie. other entities defined in the other domain files will be missing and hence not in ignore_entities and they will not be ignored. Does that make sense? Can we add a test where second domain adds 5 entities, uses 4 of them for an intent, and hence it will fall into `else` below but it will still have a separate intent defined in the first domain in ignored list? This is hard to explain but let me give examples:

Domain1:
```
intents:
  - greet:
      use_entities: [city]
 
entities:
  - city
```
Domain2:
```
intents:
  - goodbye:
      use_entities: [time, price, gender, date]
 
entities: [time, price, gender, date, not_used_entity]
```
The test should check that ignored entities for `goodbye` intent includes `city` even though it is not used in domain2 (I think it will correctly include `not_used_entity`, but would be good to assert that's there too).

tayfun: I'm actually not sure about this method here, as it seems to be used only for display or writing to a file. It looks as though `_transform_intent_properties_for_internal_use` is the method which transforms user domain files to internal format. For display/file output, we may not need to know about other domain files, but while using intents internally we do (because config files are merged and used all together).

raoulvm: @tayfun You are perfectly right. We must not make a Domain object before merging all dicts. See issue #10137 comments.

carlad: > The test should check that ignored entities for goodbye intent includes city even though it is not used in domain2 (I think it will correctly include not_used_entity, but would be good to assert that's there too).

@tayfun @raoulvm Does that mean that ignored entities for `greet` should include `time`, `price`, `gender`, `date`, and `not_used_entity`?

And what if `greet` does not specify any entities:
```
intents:
  - greet:
 
entities:
  - city
 ```
Usually this would mean all defined entities are used. So would it use _all_ entities (`city`, `greet` should include `time`, `price`, `gender`, `date`, and `not_used_entity) or just the entity defined within the same file, `city`?

Comment (raoulvm) [reply rasa/shared/core/domain.py]: Hi @carlad from my point of view you are right in both cases.

Comment (carlad) [reply rasa/shared/core/domain.py]: thanks @raoulvm so the behaviour should be that _any_ `entities`, no matter which domain file they are specified in, should be assigned to _any_ `intent`, unless explicitly `use`d or `ignore`d. Further, any explicit use automatically ignores remaining entities, and vice versa, where any explicit ignore, automatically uses remaining entities.

Comment (carlad) [reply rasa/shared/core/domain.py]: Okay - so I think I know what's happening that caused the original issue.

In a domain file, when an intent is set, with a `use_entity` configured, and that same entity is the _only_ entity specified in the file, eg:
```
intents:
  - greet:
      use_entities: [city]
 
entities:
  - city
```
This is interpreted as `'intents': [{'greet': {'use_entities': True}}]` because it is using all the entities as specified at that point. 

When the subsequent domain files are merged, the value of `greet` stays the same, i.e. ` {'use_entities': True}}` and it uses all the entities.

I want to look more into the existing `Domain.merge` method to see if we can implement a solution to this there.

File: rasa/shared/core/domain.py [hunk 2]
```diff
@@ -225,15 +225,17 @@ def _get_session_config(session_config: Dict) -> SessionConfig:
...
         """Loads and merges multiple domain files recursively from a directory tree."""
-
-        domain = Domain.empty()
+        domain_dict = {}
         for root, _, files in os.walk(path, followlinks=True):
             for file in files:
                 full_path = os.path.join(root, file)
                 if Domain.is_domain_file(full_path):
-                    other = Domain.from_file(full_path)
-                    domain = other.merge(domain)
```
tayfun, line 235: What's the difference between `Domain.merge` vs `Domain.merge_domain_dicts`?

rasa/shared/core/domain.py
raoulvm: No you can't because for that you need to create a Domain object from the previous file, which then looses the information on which entities to use. "used_entities" is a positive only list. In @tayfun example above `greet` uses one entity out of one. That is, concerning the Domain.to_dict(), all of them. But given that the to-be-merged file also contains entities, that information "use exactly that one entity" is lost. Flipside, an intent using all entities would get the same "one of one" entity attribute in "used_entities". So both would have either all or only `city` entity.

rasa/shared/core/domain.py
raoulvm: Domain.merge() is a method that merges an already parsed Domain object onto another already parsed domain object.

Domain.merge_domain_dicts() is a _static_ method that merges to "raw" domain dicts _before_ they get converted into a Domain object (in which process the entity details are lost: One-out-of-one == all)

raoulvm: The culprit is `Domain.from_dict()` and similar here. In there the use_entities is replaced by "used_entities" which lacks some of the original information.

raoulvm: So we do that as last action, after merging the original unchanged data. See line 238 below.

tayfun: I see what you mean @raoulvm thanks. I'm trying to think of an elegant way to fix this issue. I feel there's code duplication with two `merge` methods but in the end I guess we can remove the old `merge` method and just keep the new `merge_domain_dicts`. It makes sense to create Domain after dictionaries are merged.

raoulvm: >  I feel there's code duplication with two merge methods

Like "a lot" because that is the way I created it...

carlad:
> Did you mean to use `sorted(...list...)` - list.sort() does not return an object (except None)

Thanks for spotting that @raoulvm. I've had to debug some issues in the tests relating to the ordering of intents and intent properties, but believe they should be passing now.

rasa/shared/core/domain.py
```diff
@@ -309,8 +311,93 @@ def merge_lists_of_dicts(
...
+        This is a fix for #10137
+        """
+        if not domain2:  # or domain2.is_empty():
+            return domain1
+
+        if domain1 == {}:  # .is_empty():
+            return domain2
+
+        domain_dict = domain2  # .as_dict(mergeflag=True)
+        combined = domain1  # .as_dict(mergeflag=True)
```
aeshky, line 335:
```suggestion
        if not domain2:
            return domain1

        if domain1 == {}:
            return domain2

        domain_dict = domain2
        combined = domain1
```

aeshky: I think these were left here by accident? Maybe you copied the old merge function to edit it and forgot to delete them?

rasa/shared/core/domain.py
```diff
@@ -309,8 +311,93 @@ def merge_lists_of_dicts(
...
+    ) -> Dict[Text, Any]:
+        """Merge this domain dict with another one, combining their attributes.
+
+        Used when multiple domain files are configured (in a single directory).
+
+        List attributes like ``intents`` and ``actions`` will be deduped
+        and merged. Single attributes will be taken from `domain1` unless
+        override is `True`, in which case they are taken from `domain2`.
+
+        This is a fix for #10137
```
aeshky, line 326:
```suggestion
```

aeshky: I don't think we normally mention bug fixes in the comments / docstrings (at least I've not seen it before). Usually we just mention it in the changelog.

rasa/shared/core/domain.py
```diff
@@ -309,8 +311,93 @@ def merge_lists_of_dicts(
...
+                merge_dicts, b = tempDict1.copy(), tempDict2.copy()
+
+            else:
+                merge_dicts, b = tempDict2.copy(), tempDict1.copy()
+
+            merge_dicts.update(b)
+            return merge_dicts
+
+        def merge_lists(list1: List[Any], list2: List[Any]) -> List[Any]:
+            return sorted(list(set(list1 + list2)))
```
aeshky, line 353: There's currently a lot of code duplication. For example the highlighted code is repeated in `merge`. Should we reduce code duplication? Maybe these nested functions can be moved outside of the merge functions and reused?

rasa/shared/core/domain.py
```diff
@@ -229,15 +229,17 @@ def _get_session_config(session_config: Dict) -> SessionConfig:
...
-
+                    _ = Domain.from_file(full_path)  # does the validation here only
+                    other_dict = rasa.shared.utils.io.read_yaml(
+                        rasa.shared.utils.io.read_file(full_path)
+                    )
+                    domain_dict = Domain.merge_domain_dicts(domain_dict, other_dict)
+        domain = Domain.from_dict(domain_dict)
         return domain
 
     def merge(self, domain: Optional["Domain"], override: bool = False) -> "Domain":
```
aeshky, line 245: I can see that you kept the old `merge` function and that it's used elsewhere. I'm wondering if that's deliberate, especially since the two functions duplicate a lot of code. Maybe the docstrings could clarify the difference between the two?

rasa/shared/core/domain.py
```diff
@@ -309,8 +311,93 @@ def merge_lists_of_dicts(
...
+
+        for key in [KEY_ENTITIES, KEY_ACTIONS, KEY_E2E_ACTIONS]:
+            combined[key] = merge_lists(combined.get(key, []), domain_dict.get(key, []))
+
+        for key in [KEY_FORMS, KEY_RESPONSES, KEY_SLOTS]:
+            combined[key] = merge_dicts(
+                combined.get(key, {}), domain_dict.get(key, {}), override
+            )
+
+        return combined  # self.__class__.from_dict(combined)
```
aeshky, line 396:
```suggestion
        return combined
```

aeshky: Also left by accident?

rasa/shared/core/domain.py
```diff
@@ -309,8 +311,93 @@ def merge_lists_of_dicts(
 
         return self.__class__.from_dict(combined)
 
+    @staticmethod
+    def merge_domain_dicts(
+        domain1: Dict, domain2: Dict, override: bool = False
+    ) -> Dict[Text, Any]:
+        """Merge this domain dict with another one, combining their attributes.
```
aeshky, line 318:
```suggestion
        """Merges this domain dict with another one, combining their attributes.
```

aeshky: I changed it to the style that we prefer in the docs (can't seem to find the reference for it on Notion...)

rasa/shared/core/domain.py
```diff
@@ -309,8 +311,93 @@ def merge_lists_of_dicts(
...
+        for key in [KEY_FORMS, KEY_RESPONSES, KEY_SLOTS]:
+            combined[key] = merge_dicts(
+                combined.get(key, {}), domain_dict.get(key, {}), override
+            )
+
+        return combined  # self.__class__.from_dict(combined)
+
     @staticmethod
     def collect_slots(slot_dict: Dict[Text, Any]) -> List[Slot]:
+        """Collect the slots."""
```
aeshky, line 400:
```suggestion
        """Collects the slots."""
```

data/test_domains/test_domain_from_directory_for_entities/game_1.yml
```diff
@@ -0,0 +1,13 @@
...
+  - play:
+      use_entities:
+        - ball
+        - chess
+  - stow_away:
+      use_entities: true
+
+entities:
+  - ball
+  - chess
```
aeshky, line 13: Very minor comment: I have previously been asked to add a new line at the end of yaml files in our data directory. Probably worth adding them to be consistent with existing files.

aeshky (dismissed):
Thanks for the bug fix @carlad! From a research perspective, nothing stands out to me. 
I left a few comments. Main thing is about code duplication and clarifying the difference between the two merge functions. Also left a few stylistic recommendations. I think we should also add a changelog.
I can see that this is targetting the `2.8.x` branch. Does this need fixing on `main` too?

rasa/shared/core/domain.py
carlad: @aeshky 
> Does this need fixing on main too?
We're going to confine this fix to 2.8 for now as we are considering an overhaul to the Domain class.

rasa/shared/core/domain.py
carlad: Yes. Unfortunately because of how the `Domain` is loaded, the `merge` function is not aware of all the `entities` it needs to be aware of when there are multiple domain files. I've kept `merge` for now because we're planning an overhaul of the Domain class and fixing this bug has some urgency. I'll clarify in the docstrings.

rasa/shared/core/domain.py
```diff
@@ -362,7 +365,7 @@ def merge_lists(list1: List[Any], list2: List[Any]) -> List[Any]:
 
     @staticmethod
     def merge_lists_of_dicts(
-        self,
+        self: Type[D],
```
raoulvm, line 368: As a `staticmethod`- there shouldn't be any `self`or `cls`, right?

carlad: That's what I thought, but as this method calls `merge_dicts` it needs to call`self.merge_dicts`. And then the mypy type check failed with an `untyped-defs` error.

raoulvm: In that case `merge_dicts`should be declared as `@staticmethod` too, because as far as I can see it doesn't need a `self` instance. 

`mypy` complains as `self` here does not refer to the Domain instance but is a independent parameter that has to be passed to `merge_list_of_dicts` as first parameter :-(

rasa/shared/core/domain.py
```diff
@@ -256,61 +263,128 @@ def merge(self, domain: Optional["Domain"], override: bool = False) -> "Domain":
...
+        """Merges 2 dict lists."""
+        dict1 = {
+            (sorted(list(i.keys()))[0] if isinstance(i, dict) else i): i
+            for i in dict_list1
+        }
+        dict2 = {
+            (sorted(list(i.keys()))[0] if isinstance(i, dict) else i): i
+            for i in dict_list2
+        }
+        merged_dicts = self.merge_dicts(dict1, dict2, override_existing_values)
```
raoulvm, line 382: As a static method it can be called as `Domain.merge_dicts()` - sorry forgot to mention that.

carlad: Thanks for the tips @raoulvm. Much appreciated 😄 
Are you satisfied that the new tests I added for domain directories with multiple files are strong enough? Can you think of any ways you might configure the domain across multiple files that isn't reflected?

rasa/shared/core/domain.py
```diff
@@ -256,61 +259,127 @@ def merge(self, domain: Optional["Domain"], override: bool = False) -> "Domain":
...
+        ``actions``) are known to the Domain when the object is created.
+
+        List attributes like ``intents`` and ``actions`` are deduped
+        and merged. Single attributes are taken from `domain1` unless
+        override is `True`, in which case they are taken from `domain2`.
+        """
+        if not domain2:
+            return domain1
+
+        if domain1 == {}:
```
alwx, line 306: Since empty dictionaries evaluate to `False` then maybe it's better to do it as it's done above? 
I mean `if not domain1` instead of `if domain == {}`

data/test_domains/test_domain_from_directory_tree/skill_2_domain/domain_pt3.yml
```diff
@@ -0,0 +1,4 @@
+version: "2.0"
+
+intents:
+  - utter_skill_2
```
alwx, line 4: Needs an empty line.

data/test_domains/test_domain_from_directory_tree/skill_2_domain/skill_2_subdirectory/domain_pt4.yaml
```diff
@@ -0,0 +1,9 @@
+version: "2.0"
+
+intents:
+  - utter_subskill
+  - utter_subroot
+
+entities:
+  - cluedo
+  - monopoly
```
alwx, line 9: Needs an empty line.

data/test_domains/test_domain_from_directory_for_entities/game_3.yml
```diff
@@ -0,0 +1,4 @@
+version: "2.0"
+
+intents:
+  - question
```
alwx, line 4: Needs an empty line.

rasa/shared/core/domain.py
```diff
@@ -256,61 +259,127 @@ def merge(self, domain: Optional["Domain"], override: bool = False) -> "Domain":
...
+        dict1 = {
+            (sorted(list(i.keys()))[0] if isinstance(i, dict) else i): i
+            for i in dict_list1
+        }
+        dict2 = {
+            (sorted(list(i.keys()))[0] if isinstance(i, dict) else i): i
+            for i in dict_list2
+        }
+        merged_dicts = Domain.merge_dicts(dict1, dict2, override_existing_values)
+        return list(merged_dicts.values())
```
alwx, line 378: I don't see any duplicates extraction when needs to be done when merging the domain.
Here is where it currently is: https://github.com/RasaHQ/rasa/blob/main/rasa/shared/core/domain.py#L291

alwx (changes requested):

rasa/shared/core/domain.py
carlad: Hey @alwx the method you've linked to is on main, but this PR is merging into 2.8.x, which doesn't actually have the `extract_duplicates` method that you've linked to.
We're planning to eventually merge this into `main` (I've raised a [ticket](https://github.com/RasaHQ/rasa/issues/10570) for this), so will add a note there about being careful to observe any changes in domain logic between 2.8.x and 3.0.x.

alwx: Oh, that‘s my bad then, haven’t noticed that. Thanks for explanation!

alwx (approved):

PR #10426, opened 2021-11-30
Some docs improvements

### LINKED ISSUE #10291: `required_slots` deserve attention in migration guide
We've learned in https://github.com/RasaHQ/rasa/issues/10261#issuecomment-973999029 that users need to be aware of modest differences between 2.x and 3.x if they have forms that use`required_slots` in a `FormValidationAction`. 

This issue serves as a reminder while we think of a proper way to describe what to pay attention to.

ancalita:
**Proposed changes**:
- Fixes yaml indentation in slot mappings migration guide + replaces `next` links with the released docs.
- Closes #10291 

### DIFF
--- a/docs/docs/migration-guide.mdx
+++ b/docs/docs/migration-guide.mdx
@@ -1424,6 +1424,13 @@ If you have used `FormValidationAction` to define custom extraction and validati
 `required_slots` method, note that `slots_mapped_in_domain` argument has been replaced by the `domain_slots` argument.
 You will need to make this replacement to continue using your custom code.
 
+:::note
+If you have custom validation actions extending `FormValidationAction` which override `required_slots` method, it is highly
+recommended you double-check the dynamic form behavior of your migrated assistant, in case slots set by the default action
+[`action_extract_slots`](./default-actions.mdx#action_extract_slots) might need to be reset within the context of your form by the custom validation actions.
+
+:::
+
 ## Rasa 2.7 to 2.8
 
 :::caution
--- a/docs/docs/migration-guide.mdx
+++ b/docs/docs/migration-guide.mdx
@@ -1425,9 +1425,9 @@ If you have used `FormValidationAction` to define custom extraction and validati
 You will need to make this replacement to continue using your custom code.
 
 :::note
-If you have custom validation actions extending `FormValidationAction` which override `required_slots` method, it is highly
-recommended you double-check the dynamic form behavior of your migrated assistant, in case slots set by the default action
-[`action_extract_slots`](./default-actions.mdx#action_extract_slots) might need to be reset within the context of your form by the custom validation actions.
+If you have custom validation actions extending `FormValidationAction` which override `required_slots` method, you should
+double-check the dynamic form behavior of your migrated assistant. Slots set by the default action
+[`action_extract_slots`](./default-actions.mdx#action_extract_slots) may need to be reset within the context of your form by the custom validation actions.
 
 :::
 
--- a/docs/docs/migration-guide.mdx
+++ b/docs/docs/migration-guide.mdx
@@ -1337,14 +1337,14 @@ Let's consider the following 2.0 domain file:
 entities:
  - cuisine
  - number
- slots:
+slots:
    cuisine:
      type: text
    num_people:
      type: float
    outdoor_seating:
      type: bool
- forms:
+forms:
    restaurant_form:
      required_slots:
          cuisine:
@@ -1368,7 +1368,7 @@ The initial result of migrating this domain to 3.0 format would look like this:
 entities:
  - cuisine
  - number
- slots:
+slots:
    cuisine:
      type: text
      mappings:
@@ -1388,7 +1388,7 @@ entities:
      - type: from_intent
        intent: deny
        value: false
- forms:
+forms:
    restaurant_form:
      required_slots:
      - cuisine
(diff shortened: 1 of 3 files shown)


docs/docs/migration-guide.mdx
```diff
@@ -1424,6 +1424,13 @@ If you have used `FormValidationAction` to define custom extraction and validati
 `required_slots` method, note that `slots_mapped_in_domain` argument has been replaced by the `domain_slots` argument.
 You will need to make this replacement to continue using your custom code.
 
+:::note
+If you have custom validation actions extending `FormValidationAction` which override `required_slots` method, it is highly
+recommended you double-check the dynamic form behavior of your migrated assistant, in case slots set by the default action
+[`action_extract_slots`](./default-actions.mdx#action_extract_slots) might need to be reset within the context of your form by the custom validation actions.
```
melindaloubser1, line 1430:
```suggestion
If you have custom validation actions extending `FormValidationAction` which override `required_slots` method, you should
double-check the dynamic form behavior of your migrated assistant. Slots set by the default action
[`action_extract_slots`](./default-actions.mdx#action_extract_slots) may need to be reset within the context of your form by the custom validation actions.
```

While sometimes there's no other way "it is recommended" should be avoided, especially when the recommendation is more like a global instruction

melindaloubser1: >Slots set by the default action
[`action_extract_slots`](./default-actions.mdx#action_extract_slots) may need to be reset within the context of your form by the custom validation actions.

How does this work? does the `FormValidationAction` do the resetting?

Does adding a short example make sense?

melindaloubser1 (changes requested):

docs/docs/migration-guide.mdx
ancalita: Yup, the best example is the one created by Vincent that lead to this request in docs improvements.
For example, we have this domain:
```
forms:
  name_form:
    required_slots:
      - first_name
      - last_name
      
slots:
  first_name:
    type: text
    influence_conversation: true
    mappings:
      - type: from_text
        conditions:
         - active_loop: name_form
           requested_slot: first_name
  last_name:
    type: text
    influence_conversation: true
    mappings:
      - type: from_text
        conditions:
         - active_loop: name_form
           requested_slot: last_name
  name_spelled_correctly:
    type: bool
    influence_conversation: true
    mappings:
    - type: from_intent
      value: true
      intent: affirm
    - type: from_intent
      value: false
      intent: deny
```
`name_spelled_correctly` is a dynamic slot added via `FormValidationAction` and there's also a `utter_ask_name_spelled_correctly` with buttons.
But considering it uses a global slot mapping, when the user utters a message with the first name, that message gets misclassified as either `affirm` or `deny` and `action_extract_slots` will set the slot `name_spelled_correctly`. So what we want to do is that when `first_name` gets validated we want to reset the slot `name_spelled_correctly` to `None`, then this slot can be requested next.

```
    def validate_first_name(
        self,
        slot_value: Any,
        tracker: Tracker,
        domain: DomainDict,
    ) -> Dict[Text, Any]:
        """Validate `first_name` value."""
        # If the name is super short, it might be wrong.
        print(f"First name given = {slot_value} length = {len(slot_value)}")
        if len(slot_value) <= 1:
            dispatcher.utter_message(text=f"That's a very short name. I'm assuming you mis-spelled.")
            return {"first_name": None, "name_spelled_correctly": None}
        else:
            return {"first_name": slot_value, "name_spelled_correctly": None}
```
How best to add this example?

melindaloubser1: Tbh this example is still confusing to me because it's showing the reset in the context of validating a _different_ slot  - would need the dynamic addinig method too to make a complete example, but maybe that gets too long for this section. 
Is there a concise way to show in which method the resetting should happen, or is this entirely case dependant?

ancalita: > Is there a concise way to show in which method the resetting should happen, or is this entirely case dependant?

Depends case by case I'm afraid, as to when the dynamic slot is expected to be requested, is it after the first slot, 2nd slot in the form `required_slots` etc.

melindaloubser1: Gotcha. This is my attempt at a verbal example:
```
Slots set by the default action
[`action_extract_slots`](./default-actions.mdx#action_extract_slots) may need to be reset within the context of your form by custom validation actions for the form's required slots. For example, if your form dynamically adds a required slot after the first slot is filled, you may want to reset the potential required slot  as part of the first required slot's validation method to ensure it will be empty when added. 
```
Pretty confusing still but maybe better than nothing

melindaloubser1: Isn't it possible to overwrite a different method of the FormValidationAction to do this resetting based on the value of `requested_slot` itself?

ancalita: If you mean allow the developer overwrite a different method, there's nothing eligible from what's available in the interface. Even more, from what I can see [here](https://github.com/RasaHQ/rasa-sdk/blob/231e6a683cba9f7bc4e81f4972c46647f10a1f13/rasa_sdk/forms.py#L326), `next_requested_slot` method requires the dynamic slot to be set to `None` in order for it to be requested. But that's an interesting proposal, I'll submit a short issue in the sdk repo and drop it in our inbox, then prioritise according to community feedback.

melindaloubser1: That sounds promising! Let's merge the warning here as is then (generic warning to check behaviour), and then if this is a good solution we add it later.

melindaloubser1 (approved):

PR #10428, opened 2021-11-30
Overriding default config in case of nested dictionaries

### LINKED ISSUE #10391: GraphComponentException if `hidden_layers_size` used with only one (text) overwritten key
### Rasa Open Source version

3.0.0

### Rasa SDK version

3.0.0

### Rasa X version

_No response_

### Python version

3.8

Windows

### What happened?

I wanted to train a migrated AI with the following config for `DIETClassifier`:

```
- name: DIETClassifier
  random_seed: 42
  loss_type: cross_entropy
  model_confidence: softmax
  constrain_similarities: True
  intent_classification: True
  entity_recognition: False
  use_masked_language_model: True
  epochs: 80
  number_of_transformer_layers: 4
  transformer_size: 256
  drop_rate: 0.2
  batch_size: [64, 256]
  embedding_dimension: 50
  hidden_layers_sizes:
    text: [512, 128]
```
The result was the following error:

```
  File "...rasa\nlu\classifiers\diet_classifier.py", line 1277, in __init__
    self._prepare_layers()
  File "...rasa\nlu\classifiers\diet_classifier.py", line 1437, in _prepare_layers
    self.config[HIDDEN_LAYERS_SIZES][self.label_name],
  File "...wrapt\wrappers.py", line 394, in __getitem__
    return self.__wrapped__[key]
KeyError: 'label'
```

followed by 

`rasa.engine.exceptions.GraphComponentException: Error running graph component for node train_DIETClassifier5.`

After digging into it, it seems to me, that even if the documentation sais:

```
hidden_layers_sizes: This parameter allows you to define the number of feed forward layers and their output dimensions for user messages and intents (default: text: [], label: []). 
```
it is not enough to specify only one of the keys. After changing the config to:

```
  hidden_layers_sizes:
    text: [512, 128]
    label: []
```

everything worked fine. I am not absolutely sure but this worked in Rasa `2.8.x`.

### Command / Request

_No response_

### Relevant log output

```shell
2021-11-25 14:36:48 INFO     rasa.engine.training.hooks  - Restored component 'LexicalSyntacticFeaturizer' from cache.
2021-11-25 14:36:48 INFO     rasa.engine.training.hooks  - Restored component 'CountVectorsFeaturizer' from cache.
2021-11-25 14:36:48 INFO     rasa.engine.training.hooks  - Restored component 'CountVectorsFeaturizer' from cache.
2021-11-25 14:36:48 INFO     rasa.engine.training.hooks  - Restored component 'MemoizationPolicy' from cache.
2021-11-25 14:36:48 INFO     rasa.engine.training.hooks  - Restored component 'RulePolicy' from cache.
2021-11-25 14:36:49 INFO     rasa.engine.training.hooks  - Starting to train component 'DIETClassifier'.
...
  File "...rasa\nlu\classifiers\diet_classifier.py", line 1277, in __init__
    self._prepare_layers()
  File "...rasa\nlu\classifiers\diet_classifier.py", line 1437, in _prepare_layers
    self.config[HIDDEN_LAYERS_SIZES][self.label_name],
  File "...wrapt\wrappers.py", line 394, in __getitem__
    return self.__wrapped__[key]
KeyError: 'label'
```

Definition of Done:
- [x] Discovered the problem
- [x] Fixed the problem
- [x] Made sure the same problem does not exist for other parameters of the component and also other ML components like TEDPolicy, ResponseSelector and UnexpecTEDIntentPolicy
- [x] Model regression tests run on the PR
- [x] PR is merged to `3.0.x`
- [ ] New micro version released after the bugfix is merged.

mleimeister:
Fixes #10391 

The error reported in the above ticket seemingly comes from the default configuration being overridden without taking care of nested dictionaries. This as far as I can tell happens in the `GraphNode` implementation in `rasa/engine/graph.py`, where the default and custom configs of a component get merged via [dictionary unpacking](https://github.com/RasaHQ/rasa/blob/3.0.x/rasa/engine/graph.py#L362). Since by definition of the dictionary creation via unpacking, the [last value for a given key wins](https://stackoverflow.com/a/39678945/17496163), this will not keep the unspecified keys of the nested dict in the default config.

**Proposed changes**:
- In Rasa 2.x, default and custom config parameters were combined using the function [override_defaults](https://github.com/RasaHQ/rasa/blob/2.8.x/rasa/utils/train_utils.py#L379) that keeps the entries in the default config nested dictionary if their keys are not specified in the custom config. A possible solution would therefore be to use a similar function in the initialization of `GraphNode` to merge default and custom configs.
- However, a potential issue w.r.t. to testing the solution is that components are initialized as a `GraphNode` via [from_schema_node](https://github.com/RasaHQ/rasa/blob/3.0.x/rasa/engine/graph.py#L522) when they are run in the engine, but unit tested by calling the component's `create` function directly, e.g. [here](https://github.com/RasaHQ/rasa/blob/3.0.x/tests/nlu/classifiers/test_diet_classifier.py#L75). The unit tests use the same dictionary unpacking as the graph node initialization and therefore this currently leads to the same error. However, a fix in `graph.py` does not have any effect on the unit tests of the individual components.
- The current state of the code is such that a fix is implemented for `GraphNode` and tested generically in `/tests/engine/test_graph_node.py`, as well as a unit test for `DIETClassifier` with the observed setting that failed from the bug report, which needs the same fix in the unit tests fixture.
- The output of the unit tests before the fix got added can be seen [here](https://github.com/RasaHQ/rasa/runs/4366488870?check_suite_focus=true#step:16:109) for `GraphNode` and [here](https://github.com/RasaHQ/rasa/runs/4366489403?check_suite_focus=true#step:16:131) for `DIETClassifier`. NOTE: They are from the previous PR that has been closed since it branched off from `main` rather than `3.0.x`, but the changes were taken over into this PR.

**Open questions**
- Due to the second bullet point described above, it seems that the components unit tests do not fully reproduce how the components are run in the graph engine, at least w.r.t. initializing the components and consuming the custom config. Is this a general issue that needs to be looked at further?

### DIFF
--- a/tests/nlu/classifiers/test_diet_classifier.py
+++ b/tests/nlu/classifiers/test_diet_classifier.py
@@ -5,6 +5,7 @@ import numpy as np
 import pytest
 from typing import Callable, List, Optional, Text, Dict, Any, Tuple
 
+import rasa.utils.common
 from rasa.engine.graph import ExecutionContext, GraphComponent
 from rasa.engine.storage.resource import Resource
 from rasa.engine.storage.storage import ModelStorage
@@ -35,6 +36,7 @@ from rasa.utils.tensorflow.constants import (
     ENTITY_RECOGNITION,
     INTENT_CLASSIFICATION,
     MODEL_CONFIDENCE,
+    HIDDEN_LAYERS_SIZES,
 )
 from rasa.nlu.tokenizers.whitespace_tokenizer import WhitespaceTokenizer
 from rasa.nlu.classifiers.diet_classifier import DIETClassifier
@@ -73,7 +75,9 @@ def create_diet(
 
         default_execution_context.is_finetuning = finetune
         return constructor(
-            config={**DIETClassifier.get_default_config(), **config},
+            config=rasa.utils.common.override_defaults(
+                DIETClassifier.get_default_config(), config
+            ),
             model_storage=default_model_storage,
             execution_context=default_execution_context,
             resource=default_diet_resource,
@@ -303,6 +307,16 @@ async def test_train_persist_load_with_different_settings(
     create_diet: Callable[..., DIETClassifier],
 ):
     config = {LOSS_TYPE: "margin", EPOCHS: 1}
+    create_train_load_and_process_diet(config)
+    create_diet(config, load=True, finetune=True)
+
+
+@pytest.mark.timeout(240, func_only=True)
+async def test_train_persist_load_with_nested_dict_config(
+    create_train_load_and_process_diet: Callable[..., Message],
+    create_diet: Callable[..., DIETClassifier],
+):
+    config = {HIDDEN_LAYERS_SIZES: {"text": [256, 512]}}
     create_train_load_and_process_diet(config)
     create_diet(config, load=True, finetune=True)
 
--- a/tests/nlu/classifiers/test_diet_classifier.py
+++ b/tests/nlu/classifiers/test_diet_classifier.py
@@ -316,7 +316,7 @@ async def test_train_persist_load_with_nested_dict_config(
     create_train_load_and_process_diet: Callable[..., Message],
     create_diet: Callable[..., DIETClassifier],
 ):
-    config = {HIDDEN_LAYERS_SIZES: {"text": [256, 512]}}
+    config = {HIDDEN_LAYERS_SIZES: {"text": [256, 512]}, ENTITY_RECOGNITION: False}
     create_train_load_and_process_diet(config)
     create_diet(config, load=True, finetune=True)
 
--- a/changelog/10391.bugfix.md
+++ b/changelog/10391.bugfix.md
@@ -1,0 +1,3 @@
+Fix overriding of default config with custom config containing nested dictionaries. Before,
+ the keys of a nested dictionary in the default config that were not specified in the 
+custom config got lost.
--- a/rasa/core/policies/unexpected_intent_policy.py
+++ b/rasa/core/policies/unexpected_intent_policy.py
@@ -5,6 +5,7 @@ import tensorflow as tf
 from pathlib import Path
 from typing import Any, List, Optional, Text, Dict, Type, TYPE_CHECKING
 
+import rasa.utils.common
 from rasa.engine.graph import ExecutionContext
 from rasa.engine.recipes.default_recipe import DefaultV1Recipe
 from rasa.engine.storage.resource import Resource
@@ -889,7 +890,7 @@ class UnexpecTEDIntentPolicy(TEDPolicy):
 
     @classmethod
     def _update_loaded_params(cls, meta: Dict[Text, Any]) -> Dict[Text, Any]:
-        meta = train_utils.override_defaults(cls.get_default_config(), meta)
+        meta = rasa.utils.common.override_defaults(cls.get_default_config(), meta)
         return meta
 
     @classmethod
--- a/rasa/engine/graph.py
+++ b/rasa/engine/graph.py
@@ -11,6 +11,7 @@ from rasa.engine.exceptions import (
     GraphSchemaException,
 )
 import rasa.shared.utils.common
+import rasa.utils.common
 from rasa.engine.storage.resource import Resource
 
 from rasa.engine.storage.storage import ModelStorage
@@ -359,10 +360,9 @@ class GraphNode:
         self._constructor_fn: Callable = getattr(
             self._component_class, self._constructor_name
         )
-        self._component_config: Dict[Text, Any] = {
-            **self._component_class.get_default_config(),
-            **component_config,
-        }
+        self._component_config: Dict[Text, Any] = rasa.utils.common.override_defaults(
+            self._component_class.get_default_config(), component_config,
+        )
         self._fn_name: Text = fn_name
         self._fn: Callable = getattr(self._component_class, self._fn_name)
         self._inputs: Dict[Text, Text] = inputs
--- a/rasa/utils/common.py
+++ b/rasa/utils/common.py
@@ -325,17 +325,16 @@ def override_defaults(
     Returns:
         updated config
     """
-    if defaults:
-        config = copy.deepcopy(defaults)
-    else:
-        config = {}
+    config = copy.deepcopy(defaults) if defaults else {}
 
-    if custom:
-        for key in custom.keys():
-            if isinstance(config.get(key), dict):
-                config[key].update(custom[key])
-            else:
-                config[key] = custom[key]
+    if not custom:
+        return config
+
+    for key in custom.keys():
+        if isinstance(config.get(key), dict):
+            config[key].update(custom[key])
+            continue
+        config[key] = custom[key]
 
     return config
 
--- a/rasa/utils/common.py
+++ b/rasa/utils/common.py
@@ -1,4 +1,5 @@
 import asyncio
+import copy
 import logging
 import os
 import shutil
@@ -308,6 +309,35 @@ def update_existing_keys(
         if k in updated:
             updated[k] = v
     return updated
+
+
+def override_defaults(
+    defaults: Optional[Dict[Text, Any]], custom: Optional[Dict[Text, Any]]
+) -> Dict[Text, Any]:
+    """Override default config with the given config.
+
+    We cannot use `dict.update` method because configs contain nested dicts.
+
+    Args:
+        defaults: default config
+        custom: user config containing new parameters
+
+    Returns:
+        updated config
+    """
+    if defaults:
+        config = copy.deepcopy(defaults)
+    else:
+        config = {}
+
+    if custom:
+        for key in custom.keys():
+            if isinstance(config.get(key), dict):
+                config[key].update(custom[key])
+            else:
+                config[key] = custom[key]
+
+    return config
 
 
 class RepeatedLogFilter(logging.Filter):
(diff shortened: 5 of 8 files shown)


tests/nlu/classifiers/test_diet_classifier.py
```diff
@@ -307,6 +311,16 @@ async def test_train_persist_load_with_different_settings(
     create_diet(config, load=True, finetune=True)
 
 
+@pytest.mark.timeout(240, func_only=True)
+async def test_train_persist_load_with_nested_dict_config(
+    create_train_load_and_process_diet: Callable[..., Message],
+    create_diet: Callable[..., DIETClassifier],
+):
+    config = {HIDDEN_LAYERS_SIZES: {"text": [256, 512]}}
```
koernerfelicia, line 319: Detail: you can bring the test time of this down by specifying `ENTITY_RECOGNITION: False` (because TF upgrade slowed DIET specifically with entity recognition down)

koernerfelicia:
Wow, this looks great! Thank you in particular for the thorough analysis, the very clear explanation, and making sure to add tests to cover the generic and specific case 🚀 

> Due to the second bullet point described above, it seems that the components unit tests do not fully reproduce how the components are run in the graph engine, at least w.r.t. initializing the components and consuming the custom config. Is this a general issue that needs to be looked at further?

In the interest of limiting the methods tested in each unit test, I think it makes sense to use `create`, but clearly this can be dangerous as we've missed this particular bug. I'm not sure about the more general question whether there are other places we've assumed `create` is equivalent to the graph node creation. I wasn't involved in the migration of these components, so not sure what the testing strategy was. I'd pull in someone from in enable for this. @ka-bu do you have any thoughts?

I think in general an approach like below could work:
- unit test the components using `create` (this stays the same)
- unit test the graph node creation as well as the override method (you've done this!)
- change the diet fixture to use the helper method/mirror the graph node component initialisation
- integration tests for testing how the config gets propagated

mleimeister:
Thanks so much for the kind feedback and outline of general next steps!

Regarding further places where `create` is used directly, it looks like all NLU component tests follow the same pattern: [test_mitie_entity_extractor.py](https://github.com/RasaHQ/rasa/blob/main/tests/nlu/extractors/test_mitie_entity_extractor.py#L59), [test_regex_entity_extractor.py](https://github.com/RasaHQ/rasa/blob/main/tests/nlu/featurizers/test_convert_featurizer.py#L38), [test_count_vectors_featurizer.py](https://github.com/RasaHQ/rasa/blob/main/tests/nlu/featurizers/test_count_vectors_featurizer.py#L30), [test_lm_featurizer.py](https://github.com/RasaHQ/rasa/blob/main/tests/nlu/featurizers/test_lm_featurizer.py#L38), etc. Those don't have nested dict parameters though, so it should be ok that way? (It seems the `additional_vocabulary_size` setting of `CountVectorsFeaturizer` has been deprecated in 3.0)

For policies, the components are initialized directly in the tests, but the config is overridden in the same way in a base class in [test_policies.py](https://github.com/RasaHQ/rasa/blob/3.0.x/tests/core/test_policies.py#L99) and in several individual tests e.g. [here](https://github.com/RasaHQ/rasa/blob/3.0.x/tests/core/policies/test_ted_policy.py#L645). However, since no nested dict config is tested in any of them, the issue wouldn't show up. If we'd like to test e.g. the `dense_dimension` setting for `TEDPolicy`, the `override_defaults` helper function could be used there as well.

ka-bu:
Think this is a good example for why we should have a common `__init__` for graph components (see other issue linked above). If we had that, that would get rid of the issue with adapting any create / tests :D

Apart from that: The only cases I can think of where we might need to be careful with this approach is when ...
- we have components that do not break because they query config sub-items via something like .get(bla,None) and by overwriting defaults now we basically replace that with .get(bla, <default-value>). Does that happen somewhere? (Should be enough to check the .get_default_config functions and check where those contain dicts)
- the dicts do not contain a fixed set of parameters (e.g. {custom1 : bla} should not be extended with {default-choice-1: blub, default-choice-2 : blubber}) -- but not sure that happens anywhere  🤔 

@joejuzl  wdyt ?

joejuzl:
@ka-bu I agree that moving this into a common `__init__` or into the `GraphComponent.load/create` would solve this issue, especially problem with unit tests differing from reality. 
Although this would end up being potentially quite a big change (and still has some open questions I think) - so maybe doesn't make sense for this PR? And this PR already improves things from how they are now I think. WDYT?

koernerfelicia:
> Although this would end up being potentially quite a big change (and still has some open questions I think) - so maybe doesn't make sense for this PR

So an action item for Matthias could be to create a followup issue around this? Unless you already have one, and I missed it

mleimeister:
Hi all, I explored the option of adding an `__init__` to `GraphComponent` for the example of `DIETClassifier`. I think it would work well for the components `create`. However, there is a potential problem with `DIETClassifier.load` since it overwrites `GraphComponent.load`, so putting the merge of the default and custom configs in there would not work. The component's constructor is called at the end of `DIETClassifier.load` and therefore would run `GraphComponent.__init__`, but it needs to access the merged config before that in [_load](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/classifiers/diet_classifier.py#L1087) and [_load_model](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/classifiers/diet_classifier.py#L1181). I'm not sure how to solve this best via the base class.

As you suggested, my main question would be whether to follow this further, potentially triggering some larger refactoring. Or whether to merge this PR to fix the reported issue relevant for migrating configs to 3.0.0 and create a follow-up ticket?

koernerfelicia:
> main question would be whether to follow this further, potentially triggering some larger refactoring. Or whether to merge this PR to fix the reported issue relevant for migrating configs to 3.0.0 and create a follow-up ticket?

I defer to @ka-bu and @joejuzl on that part

ka-bu:
I'm fine with just doing the necessary changes here and leaving that solution as is for now  👍 

Regarding the problem with the load: I think a clean solution would be to separate the instantiation from model loading, i.e. let __init__ only instantiate an fresh/untrained model and consistently add a "load/set/whatever" method that takes whatever is needed to set the state of an untrained model so that it equals the trained one. This would help in at least two ways:
- There was at least 1 edge case already where things were tricky because __init__ implements those two things at once (can't remember where, have to look that up) and, depending on whether we're initialising or loading, stuff had to be treated differently (which was not that obvious because everything ended up in the config again)
- Conceptually, we either want to pass all the parameters needed for loading (when loading/restoring a component) or none of them (when initialising a fresh one), but currently from the types it looks like (and should work) to only pass half of them. I know this is no issue really but just my preference.

If we do that consistently, then that would be some more changes... Can't think of more gaps now (but I also overlooked the above in the first pass :D). 

@joejuzl, do you think it makes sense to bundle the changes to __init__ with changes to load and add that as one issue to the backlog (or do you think that separation of loading not helping / making things worse / there's a better way? )

joejuzl:
> @joejuzl, do you think it makes sense to bundle the changes to init with changes to load and add that as one issue to the backlog (or do you think that separation of loading not helping / making things worse / there's a better way? )

New issue sounds good 👍  - we can discuss the details further there

mleimeister:
@koernerfelicia Regarding regression tests, as those are mentioned in "definition of done" of the original ticket #10391 , which of them would be best to run for this PR? Using `- dataset: ["all"] -config: ["all"]` might take very long I assume?

koernerfelicia:
@mleimeister, that's a good question! We tend to use "all" for these types of things, although that's not exactly efficient. 

You could easily just run all over the weekend, but if you'd like to do a subset:

If you look at the configs [here](https://github.com/RasaHQ/training-data/tree/main/configs/en/nlu), I think all of them have some sort of nested dict, so maybe it would be a good idea to run one where `DIET` and one where `ResponseSelector` have specified parameters in a nested dict.

I'm not 100% sure what we're looking for here -- we tend to (ab)use these tests as a sanity check/integration tests, and I think if something had broken it would be pretty obvious. Maybe just a mid-size dataset (`Hermit` is very large, the others are mid-size). I would not run Private 2 and 3 since they have no entities. You can see which dataset can be found where [here](https://github.com/RasaHQ/training-data/blob/main/dataset_config_mapping.json) if you'd like to take a look.

mleimeister:
/modeltest

```yml
include:
 - dataset: ["Sara"]
   config: ["Sparse + BERT + DIET(bow) + ResponseSelector(bow)", "BERT + DIET(seq) + ResponseSelector(t2t)"]
```

koernerfelicia (commented):
LGTM, and the regression tests look good, too!  Can you add a changelog entry? Read more about this [here](https://github.com/RasaHQ/rasa/tree/main/changelog)

mleimeister:
Great, thanks! I added a changelog paragraph here: [10391.bugfix.md](https://github.com/RasaHQ/rasa/blob/cba292650acde009ed41f3971df2c6127c3bedcf/changelog/10391.bugfix.md). Let me know if it should be more detailed/any further steps have to be applied to the file.

koernerfelicia (approved):
@mleimeister changelog looks good, I think this is ready to merge 🚀 
If you haven't already, could you also create the followup issue?

mleimeister:
@koernerfelicia @ka-bu @joejuzl Fyi: I created a follow-up ticket for the shared init/model loading question and assigned it to Enable here: https://github.com/RasaHQ/rasa/issues/10464

PR #10437, opened 2021-12-01
domain_for_core_training_provider

### LINKED ISSUE #10145: Changing the required slots of a form retrains the model
### Rasa Open Source version

3.0.0rc1

### Rasa SDK version

_No response_

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

When you change the required slots of a form in the domain then the policies retrain. This is not necessary. 

### Command / Request

_No response_

### Relevant log output

_No response_

**Definition of done**
- [ ] List what changes to domain would need to cause a retrain
- [ ] Change the` domain without responses` component (so only the things that need to cause a retrain actually cause a retrain)
- [ ] Add integration test that ensures this works as expected

ka-bu:
**Proposed changes**:
- closes #10145


ka-bu:
I'm not really sure how to add a useful integration test here. What I could test is something like: If I change things in the domain (in some valid way -- i.e. I can't even directly reuse the "prune_domain" thing, I think) and re-train the whole pipeline (including featurizers because without them the core parts won't be affected as much), then does the core part look just like as if I trained without those changes to the domain -- but that whole test would work without this component :D. With this component included, I'm not sure how to approach that 🤔 . Any ideas?

rasa/shared/core/slot_mappings.py
```diff
@@ -7,25 +6,27 @@
...
-    FROM_INTENT = 1
-    FROM_TRIGGER_INTENT = 2
-    FROM_TEXT = 3
-    CUSTOM = 4
-
-    def __str__(self) -> Text:
-        """Returns a string representation of the object."""
-        return self.name.lower()
+class SlotMapping:
+    """Defines functionality for the available slot mappings."""
```
joejuzl, line 29: I think it was a conscious (and recent) decision to encapsulate these constants in this enum class. So I wouldn't move it without good reason.

joejuzl:
> I'm not really sure how to add a useful integration test here. What I could test is something like: If I change things in the domain (in some valid way -- i.e. I can't even directly reuse the "prune_domain" thing, I think) and re-train the whole pipeline (including featurizers because without them the core parts won't be affected as much), then does the core part look just like as if I trained without those changes to the domain -- but that whole test would work without this component :D. With this component included, I'm not sure how to approach that 🤔 . Any ideas?

I would do something similar to `test_models_not_retrained_if_only_new_responses` in `test_model_training.py`. i.e. change all the things which are pruned and check that it doesn't cause core to retrain.

rasa/graph_components/providers/domain_for_core_training_provider.py
```diff
@@ -0,0 +1,92 @@
...
+    KEY_FORMS,
+)
+from rasa.shared.core.constants import SLOT_MAPPINGS
+
+
+class DomainForCoreTrainingProvider(GraphComponent):
+    """Provides domain without information that is irrelevant for core training.
+
+    The information that we retain includes:
+    - intents and their "used" and "ignored" entities because intents influence the
```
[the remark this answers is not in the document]
ka-bu: So do you think I should rephrase that or keep the keys that users see / we have in the config.yml in the description?

rasa/graph_components/providers/domain_for_core_training_provider.py
```diff
@@ -0,0 +1,92 @@
...
+    def provide(self, domain: Domain) -> Domain:
+        """Recreates the given domain but drops information that is irrelevant for core.
+
+        Args:
+            domain: A domain.
+
+        Returns:
+            A similar domain without information that is irrelevant for core training.
+        """
+        return self.create_pruned_version(domain)
```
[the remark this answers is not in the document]
ka-bu: If we grab the domain in the init from the execution context or so -- Do we then we also remove "domain" from the input here? And why do we have a domain provider component then? 🤔

rasa/graph_components/providers/domain_for_core_training_provider.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Args:
+            domain: A domain.
+
+        Returns:
+             A similar domain without information that is irrelevant for core training.
+        """
+        serialized_domain = copy.deepcopy(domain.as_dict())
+
+        serialized_domain.pop("config", None)  # `store_entities_as_slots`
```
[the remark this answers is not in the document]
ka-bu: it is a reminder for what is in the "config" key

rasa/graph_components/validators/finetuning_validator.py
```diff
@@ -208,17 +211,16 @@ def _compare_or_memorize(
...
 
         Args:
             domain: a domain
         Returns:
             fingerprint
         """
-        domain = copy.copy(domain)
-        domain.responses = {}
-        return domain.fingerprint()
+        pruned_domain = DomainForCoreTrainingProvider.create_pruned_version()
```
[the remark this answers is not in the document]
ka-bu: The finetuning validation should happen before the core parts of the graph (i.e. before the domain without ... provider)

rasa/shared/core/slot_mappings.py
ka-bu: And I really like enums but I had trouble with circular imports and couldn't properly replace some of the str with constants otherwise :/ 

We could create an enum inside the constants instead (and we could inherit from str and Enum to get rid of some str() conversions). Is that bettter / Do you have an idea for how to call that enum vs. the class here?

Or shall I revert the str/const replacements and other changes?

ka-bu:
> > I'm not really sure how to add a useful integration test here. What I could test is something like: If I change things in the domain (in some valid way -- i.e. I can't even directly reuse the "prune_domain" thing, I think) and re-train the whole pipeline (including featurizers because without them the core parts won't be affected as much), then does the core part look just like as if I trained without those changes to the domain -- but that whole test would work without this component :D. With this component included, I'm not sure how to approach that 🤔 . Any ideas?
> 
> I would do something similar to `test_models_not_retrained_if_only_new_responses` in `test_model_training.py`. i.e. change all the things which are pruned and check that it doesn't cause core to retrain.

ok yeah, I can include that as well -- had thought if I only test that the domain items get removed in the expected way, then I don't need to test that fingerprints will be the same and hence retraining won't happen, but it doesn't hurt 👍 

What I had described above was another kind of test that checks correctness of the approach -- ie.. whether actually removing that information from the domain is ok and leads to the same result as if we did not remove it and retrain. But I think there's no good way to do that, no? 🤔

joejuzl:
> What I had described above was another kind of test that checks correctness of the approach -- ie.. whether actually removing that information from the domain is ok and leads to the same result as if we did not remove it and retrain. But I think there's no good way to do that, no? 🤔

Ah yeah I see - I think you're right that would be hard to do. I can't think of any way apart from the approach you mentioned above, and yeah seems a bit overkill/heavyweight.

rasa/shared/core/slot_mappings.py
ka-bu: All in all, we didn't really make use of the enum so far (we don't exploit that it is a collection - the only `is in` check related to these constants was for the `PREDEFINED_MAPPINGS` which was a `Set[str]`) . Or did I miss something? 🤔  

So I just pushed a **suggestion** on how to use the enum in/from the `constants.py` instead, that tries to make use of the enum class a bit in the `actions.py` (by grabbing the enum item and comparing that with other items rather than casting to str -- and the `is_predefined` instead of a set of str obtained by casting the enum items) and still has the enum defined in the `constants.py` to avoid issues with circular imports. Wdyt?

joejuzl: cc: @ancalita

ancalita: Originally `SlotMapping` class already extended `Enum` in `2.x` versions, and included those constants + `__str__` + `validate` methods, this is how I found the class originally. I must admit, I might have bloated the original class with a lot of static methods during the 3.0 work. So I'm not against refactoring. Although I would prefer keeping all `SlotMapping` related objects in this module, rather than spreading into `constants.py`.

ka-bu: - enum items -> `constants.py`: all other constants are expected to be in `constants.py`s which also helps to avoid circular imports like those that happen in places if we leave it in the `slot_mappings` module (I actually like what you say better and would like to see that in general, but it seems inconsistent with what we have atm)
- enum items -> `class SlotMapping`: Think the naming isn't ideal then because **all** functionality is staticmethods and an `SlotMapping` instance is nothing more than an object describing the **type** of the mapping (I know, no current change, still something we could fix along the way now). And without the constants we could drop the whole class as well, because there's no much point in keeping that collection of static functions if there is only 1 more function in that whole module 😅  (good that we talked about this - think I left that in there as is 😬 ).

Does that (`SlotMappingType` in `constants.py` + make `SlotMapping` static methods module level methods) make sense? Or do you think we could i.e. add something to the `SlotMapping` instances? 🤔

ancalita: This might sound 🤪 How about having `SlotMappingType` as a nested class in `SlotMapping`, then have a constructor for `SlotMapping` that accepts `mapping` as parameter in order to have an extracted `SlotMappingType` from the `mapping` as an attribute of `SlotMapping`? Would that work?

rasa/graph_components/providers/domain_for_core_training_provider.py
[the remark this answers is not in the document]
joejuzl: I think it's actually fine as it is - was just being picky :)

joejuzl (commented):
I think it's still worth having a test in `test_model_training` that shows we can change these things in the domain without retraining policies.

rasa/shared/core/slot_mappings.py
ka-bu: > constructor for SlotMapping that accepts mapping as parameter in order to have an extracted SlotMappingType from the mapping as an attribute of SlotMapping

No worries :D That is not 🤪   at all, that is what the enum class enable :) So you can do `SlotMappingType(<string representation from value here>)` to receive the slot mapping type instance :)  (and then a `SlotMapping` instance would still not do more than wrapping/remembering it's type :/ and there is not really functionality that is bound/specific to the slot mapping type to justify why this should be a class now).

If you like we can have a chat after holidays and talk through that stuff again -- because I think in general I agree with you that classes could be nice to use here but that would require much more rework (which would focus on the `mapping`s as well). For now, let's move the slot types to the `constants.py` just like other constants and exploit that we can replace some strings easily then :)

rasa/graph_components/providers/domain_for_core_training_provider.py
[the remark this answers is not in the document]
ka-bu: no problem ^^

ka-bu:
> I think it's still worth having a test in `test_model_training` that shows we can change these things in the domain without retraining policies.

Hmm, I thought I was doing that in `test_train_core_with_original_or_provided_domain_and_compare`. I'm basically using the provider to modify the domain instead of removing keys from the domain manually -- and then pass in that modified domain for re-training and check that the policy would be loaded from cache.  Does that make sense or am I skipping something then? 
I could add the same changes to the domain manually again, but the other test should already take care of asserting that the respective keys are really removed/replaced, no? 🤔

rasa/shared/core/slot_mappings.py
ancalita: Sounds good 👍🏼

tests/graph_components/providers/test_domain_for_core_training_provider.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # Let the provider create a modified domain
+    original_domain = Domain.from_file(original_domain_file)
+    component = DomainForCoreTrainingProvider.create(
+        {"arbitrary-unused": 234},
+        default_model_storage,
+        Resource("xy"),
+        default_execution_context,
+    )
+    modified_domain = component.provide(domain=original_domain)
```
joejuzl, line 147: Isn't this already going to happen as it's part of the default recipe? Or am I missing something?

joejuzl:
> Hmm, I thought I was doing that in test_train_core_with_original_or_provided_domain_and_compare. I'm basically using the provider to modify the domain instead of removing keys from the domain manually -- and then pass in that modified domain for re-training and check that the policy would be loaded from cache. Does that make sense or am I skipping something then?
I could add the same changes to the domain manually again, but the other test should already take care of asserting that the respective keys are really removed/replaced, no? 🤔

Missed that you'd added that there (was expecting it in a different file) - the test looks good 👍

tests/graph_components/providers/test_domain_for_core_training_provider.py
ka-bu: If I don't modify the domain, then this test is pointless, so...
>  I'm basically using the provider to modify the domain instead of removing keys from the domain manually -- and then pass in that modified domain for re-training and check that the policy would be loaded from cache. [...] I could add the same changes to the domain manually again, but the other test should already take care of asserting that the respective keys are really removed/replaced, no? 🤔

joejuzl (approved):
LGTM 💯

rasa/shared/core/slot_mappings.py
ka-bu: Just re-read your comment, to see if I still have something todo here: Think your idea goes into a good direction :) Apart from the representation of the type, what could go into the slot mapping class is, for example:
- all the static methods could become regular methods (saving us from passing the `mapping` every time), 
- validation of the "mapping" could go in to the constructor (enabling more elaborate checks than just via the schema - if needed), 
- all the .get(<key-I-need-to-remember>) could be replaced by e.g. attribute access via dataclasses (saving us from remembering / making transparent / handling all those constants, so we can just `mapping.type` instead of `mapping.get(MAPPING_TYPE)` (which is also a bit confusing because that type is not optional, so "[]" access would suffice) 
- similar to the static methods in slot mappings, some stuff from the ActionExtractSlots might become SlotMapping methods (e.g. it might be easier to remember that we have `mapping.can_be_applied_to(tracker)` instead of searching for `ActionExtractSlots._matches_mapping_conditions(mapping: ..., tracker: ...)`)
- (( plus small things like removing type from ` SlotMapping.check_mapping_validity` which we could just as well get from the mapping then ))
... and I would love to add some of those changes, but I think it will be considered way out of scope 😅

ancalita: Sounds great, do you want to place these refactorings in an issue of itself? Or continue in this PR?

ka-bu: @joejuzl just suggested that we should make it separate so that the domain stuff can be merged first (and then increase the scope of that slot mappings refactor a bit maybe) -> https://github.com/RasaHQ/rasa/issues/10654 :)

ka-bu:
@joejuzl @ancalita , any idea why e.g. `pytest tests/core/test_processor.py -k test_restart_triggers_session_start` now fails because because an `action_default_fallback` is predicted -- and only on Ubuntu? 

We now get:
```
E       assert {'recipient_i...rephrase it?'} == {'recipient_i...there name1!'}
E         Omitting 1 identical items, use -vv to show
E         Differing items:
E         {'text': "sorry, I didn't get that, can you rephrase it?"} != {'text': 'hey there name1!'}
E         Use -v to get the full diff
```
while the events after in the `handle_message` call (before running prediction and returning the tracker that causes the error) look fine and end, as expected, with a slots set: `{'event': 'slot', 'timestamp': 1641829919.124695, 'metadata': {'model_id': '0a92dbf884b7409b9b4b37acf3db4fc8'}, 'name': 'name', 'value': 'name1'}` .

joejuzl:
> @joejuzl @ancalita , any idea why e.g. pytest tests/core/test_processor.py -k test_restart_triggers_session_start now fails because because an action_default_fallback is predicted -- and only on Ubuntu?

hmm that's really weird... `RegexMessageHandler` should return a confidence of `1.0` (unless overwritten).
Is the component definitely running?

ka-bu:
> > @joejuzl @ancalita , any idea why e.g. pytest tests/core/test_processor.py -k test_restart_triggers_session_start now fails because because an action_default_fallback is predicted -- and only on Ubuntu?
> 
> hmm that's really weird... `RegexMessageHandler` should return a confidence of `1.0` (unless overwritten). Is the component definitely running?

The `RegexMessageHandler` works fine, the event sequence produced by handle message is
```
(Pdb) tracker.events
deque([ActionExecuted(action: action_session_start, policy: None, confidence: 1.0), <rasa.shared.core.events.SessionStarted object at 0x179a8fac0>, ActionExecuted(action: action_listen, policy: None, confidence: None), <rasa.shared.core.events.UserUttered object at 0x179aadf70>, <rasa.shared.core.events.SlotSet object at 0x179aadd90>, <rasa.shared.core.events.DefinePrevUserUtteredFeaturization object at 0x179aa19d0>, ActionExecuted(action: action_default_fallback, policy: RulePolicy, confidence: 0.3), BotUttered('sorry, I didn't get that, can you rephrase it?', {"elements": null, "quick_replies": null, "buttons": null, "attachment": null, "image": null, "custom": null}, {"utter_action": "utter_default", "model_id": "a2f6bc2d229a4beeb0e7c305fb79b711"}, 1641830914.705086), <rasa.shared.core.events.UserUtteranceReverted object at 0x179a89b20>, ActionExecuted(action: action_listen, policy: followup_action, confidence: 1.0)])
```
with the `UserUttered` containing
```
(Pdb) tracker.events[3].intent
{'name': 'greet', 'confidence': 1.0}
```

ka-bu:
@joejuzl  @ancalita  Found it: When parsing the stories and turn them into event sequences, we automatically insert explicit `SlotSet` events based on the `from_entity` mapping. The `domain_for_core_training` provider removed those mappings :). (References are in the docstr of the provider component, if you want to have a look). 

After the fix, the only addition that is left is the "required slots" which was what Tobi asked for initially. Moreover, I couldn't find any references to forms in the story to event logic so, I'll enable merge if this passes now.

PR #10443, opened 2021-12-02
Warn that rasa 3.0 is not compatible with `rasa-x` for `rasa x` CLI command

### LINKED ISSUE #10401: RuntimeError: There is no current event loop in thread 'MainThread'.
Sentry Issue: [RASA-OPEN-SOURCE-HVR](https://sentry.io/organizations/rasahq/issues/2813460037/?referrer=github_integration)

```
RuntimeError: There is no current event loop in thread 'MainThread'.
(9 additional frame(s) were not displayed)
...
  File "rasa/cli/x.py", line 414, in run_in_production
    _rasa_service(args, endpoints, None, credentials_path)
  File "rasa/cli/x.py", line 89, in _rasa_service
    serve_application(
  File "rasa/core/run.py", line 178, in serve_application
    app = configure_app(
  File "rasa/core/run.py", line 103, in configure_app
    app = server.create_app(
  File "rasa/server.py", line 650, in create_app
    Initialize(
```

**Definition of done**
- [x] Investigate cause (could be caused by upgrade of `sanic-jwt` to version `1.7.0`), some initial ideas [here](https://rasa-hq.slack.com/archives/C01M5TAHDHA/p1637925358000800?thread_ts=1637785366.000100&cid=C01M5TAHDHA)
- [x] Raise `UserWarning` when rasa version is >=3.0 as this is not compatible yet with `rasa-x`.

ancalita:
**Proposed changes**:
- Closes #10401 


rasa/cli/x.py
```diff
@@ -351,6 +353,16 @@ def rasa_x(args: argparse.Namespace) -> None:
 
     _configure_logging(args)
 
+    if version.parse(rasa.version.__version__) >= version.parse(
```
ancalita, line 356: I've tried testing this in a fresh venv where ideally I would have first wanted to `pip install` rasa-x, then editable rasa, however the first installation step of rasa-x failed, please see this [thread](https://rasa-hq.slack.com/archives/C36SS4N8M/p1638441134454200). Let me know if there are other ways to test this?

joejuzl: Isn't this `if` always going to be true?

joejuzl: You can unit test this by monkeypatching versions/constants.
For manual test, it must be possible to install rasa-x 😁 I would just suggest a fresh venv and installing rasa-x first.

ancalita: >  Isn't this if always going to be true?

I was wondering the same thing until i read this comment [here](https://github.com/RasaHQ/rasa/blob/666eac66993682fa907d2aa5fb6f8f984a8f4be6/rasa/version.py#L1) - I assume the version gets overwritten with the one installed by pip? If I misread that comment, how else can I get the user's version?

ancalita: > suggest a fresh venv and installing rasa-x first.

Yeah, I tried that 3 times with fresh venv and the same `psycopg` error came up, Nick also tested on his end and ran in the same error. 
I'll try with monkeypatching in a unit test.

ancalita: Done, ready for review 😄 @joejuzl

joejuzl: > I was wondering the same thing until i read this comment here - I assume the version gets overwritten with the one installed by pip? If I misread that comment, how else can I get the user's version?

This version is overridden when we release, and is hardcoded in the code. So if a user installs via pip e.g. `rasa==3.0.1` then the code in the package they install with have `rasa.version.__version__` as `3.0.1`. And that version should always be higher than `MINIMUM_COMPATIBLE_VERSION` in the same code version.

rasa/cli/x.py
```diff
@@ -351,6 +353,17 @@ def rasa_x(args: argparse.Namespace) -> None:
...
+        MINIMUM_COMPATIBLE_VERSION
+    ):
+        rasa.shared.utils.io.raise_warning(
+            f"Your version of rasa '{rasa.version.__version__}' is currently "
```
joejuzl, line 360: How do we make sure we remember to fix this once we have Rasa X support? 🤔

rasa/cli/x.py
ancalita: Oh I see, thanks for the explanation 🙌🏼

rasa/cli/x.py
ancalita: I could just open a wrap-up issue for the future and have it in our inbox already, potentially also ask Ty what labels he wants to use for the rasa-x support issues.

rasa/cli/x.py
joejuzl: So I think we can just check if `rasa.version.__version__` is at least `3.00`

rasa/cli/x.py
```diff
@@ -351,6 +353,15 @@ def rasa_x(args: argparse.Namespace) -> None:
 
     _configure_logging(args)
 
+    if version.parse(rasa.version.__version__) >= version.parse("3.0.0"):
+        rasa.shared.utils.io.raise_warning(
+            f"Your version of rasa '{rasa.version.__version__}' is currently "
+            f"not supported by Rasa X. Running `rasa x` CLI command with rasa "
+            f"version higher or equal to {MINIMUM_COMPATIBLE_VERSION} "
```
joejuzl, line 360:
```suggestion
            f"version higher or equal to 3.0.0 "
```

ancalita: Good catch 😄

joejuzl (approved):
💯

Issue #9555, opened 2021-09-06
rasa data validate should fail if conflicting slots are created in different domain files


hsm207:
### What problem are you trying to solve?

Users may split the `domain.yml` into multiple files and have different teams work on different domain.

I want `rasa data validate` to check and fail if there are duplicate intents, conflicting slots, etc across the split domain files.

For example:

1. Run `rasa init`.
2. Copy `domain.yml` to `domain/domain1.yml`.
3. Delete `domian.yml`
4. Add the following slots in `domain/domain1.yml`:
![image](https://user-images.githubusercontent.com/2398765/132243210-3b9036bd-780b-4b60-8ba0-1066ee863ad1.png)
5. Create an empty file at `domain/domain2.yml` and write:
![image](https://user-images.githubusercontent.com/2398765/132243327-04c472b7-0302-4a49-bd7e-87062dc86a21.png)
6. Run  `rasa data validate -d domain/`.

The check will complete without errors

### What's your suggested solution?

`rasa data validate` consolidates all the domain files and then runs the yaml validation check.

**Definition of Done**

- [ ] `rasa data validate -d domain/` does not fail when conflicting/duplicate slots/intents/etc are found
- [ ] Add an integration test to ensure this works as expected
- [ ] Fixed merged into 2.8 and 3.0

TyDunn:
The fix for this issue will need to be backported to the last minor of the previous major (i.e. Rasa Open Source 2.8)

PR #10444, opened 2021-12-02
Make `rasa data validate` check for duplicated intents, forms, responses and slots when using domains split between multiple files

alwx:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/9555


data/test_domains/test_domain_with_duplicates/domain2.yml
```diff
@@ -0,0 +1,22 @@
...
+
+  utter_happy:
+    - text: "Great, carry on!"
+
+slots:
+  mood:
+    type: bool
+    mappings:
+      - type: from_entity
+        entity: some_slot
```
ancalita, line 22: New line required 😄

rasa/shared/core/domain.py
```diff
@@ -185,11 +185,15 @@ def from_yaml(cls, yaml: Text, original_filename: Text = "") -> "Domain":
...
     @classmethod
-    def from_dict(cls, data: Dict) -> "Domain":
+    def from_dict(
+        cls, data: Dict, duplicates: Optional[Dict[Text, List[Text]]] = None
+    ) -> "Domain":
         """Deserializes and creates domain.
 
         Args:
             data: The serialized domain.
+            duplicates: The dictionary of duplicated intents, slots, forms, etc when the
```
ancalita, line 195: I'd prefer having a more explicit docstring defining the key-value pairs of `duplicates` dict, I didn't quite know what to expect until looking further in the code in `merge`

rasa/shared/core/domain.py
```diff
@@ -298,9 +298,12 @@ def merge_lists_of_dicts(
         if override or self.session_config == SessionConfig.default():
             combined[SESSION_CONFIG_KEY] = domain_dict[SESSION_CONFIG_KEY]
 
-        combined[KEY_INTENTS] = merge_lists_of_dicts(
-            combined[KEY_INTENTS], domain_dict[KEY_INTENTS], override
-        )
+        duplicates: Dict[Text, List[Text]] = {}
```
ancalita, line 301: Because you removed a method and introduced the duplicate extraction, I'd also add a brief comment about the steps to clarify why these are done.

tests/shared/core/test_domain.py
```diff
@@ -1611,6 +1611,26 @@ def test_domain_invalid_yml_in_folder():
...
+    Check if a domain with duplicated slots, responses and intents contains
+    a correct information in `duplicates` field.
+    """
+    domain = Domain.from_directory("data/test_domains/test_domain_with_duplicates/")
+    assert domain.duplicates["slots"] == ["mood"]
+    assert domain.duplicates["responses"] == ["utter_greet", "utter_did_that_help"]
+    assert domain.duplicates["intents"] == ["greet"]
+
+
+def test_domain_duplicates_when_one_domain_file():
```
ancalita, line 1625: Perhaps it would be good to also add a test for a valid domain dir, not only domain file?

rasa/shared/core/domain.py
```diff
@@ -588,6 +594,8 @@ def __init__(
                 events for entities if there are slots with the same name as the entity.
             session_config: Configuration for conversation sessions. Conversations are
                 restarted at the end of a session.
+            duplicates: The dictionary of duplicated intents, slots, forms, etc when the
```
ancalita, line 597: The update to docstring in `from_dict` should be copied here too 😄

ancalita (commented):
Looks great 👍🏼 
I left mainly some documentation related suggestions.
I was also wondering if we want to raise `InvalidDomain` when duplicates are not None in `from_dict`? What happens now if you train with a Domain with conflicting slots for example?
This could be purely out of scope for the issue which refers only to `rasa data validate`, but it would be just an addition of a few lines of code 🤔

ancalita (approved):
Looks good 🎉 
I'd still explore with Ty and the rest of the squad looking at the other `Domain.merge` related issues whether a holistic approach is required for this entire umbrella + if to raise `InvalidDomain` whenever duplicates are found?

rasa/shared/core/domain.py
```diff
@@ -298,9 +299,13 @@ def merge_lists_of_dicts(
...
 
-        combined[KEY_INTENTS] = merge_lists_of_dicts(
-            combined[KEY_INTENTS], domain_dict[KEY_INTENTS], override
-        )
+        duplicates: Dict[Text, List[Text]] = {}
+
+        # this code merges lists of dicts of intents
+        dict1 = {list(i.keys())[0]: i for i in combined[KEY_INTENTS]}
+        dict2 = {list(i.keys())[0]: i for i in domain_dict[KEY_INTENTS]}
+        duplicates[KEY_INTENTS] = extract_duplicates(dict1, dict2)
```
joejuzl, line 307: Did we consider just raising an error/warning here so that we don't need to store the duplicates on the class?

alwx: The idea was to show this information only when running `rasa data validate`, as it's written in the description of this issue. We can update it to show the warning anyway but in this case I don't see the point in doing this sort of check in `rasa data validate` which will mean the whole task will become different.

alwx: Maybe @TyDunn can take a look at it and say which approach we prefer.

alwx: If we do it the way @joejuzl proposes then the warnings/exceptions will be raised every time the domain is getting merged (e.g. when just running `rasa train` or literally any other command that uses domain), not only when doing `rasa data validate`

alwx: @TyDunn If we do it the way @joejuzl proposes then the warnings/exceptions will be raised every time the domain is getting merged (e.g. when just running rasa train or literally any other command that uses domain), not only when doing rasa data validate

TyDunn: @alwx I believe Joe is out for the rest of the year, and I am not going to be able to gain enough context atm. Can you make a judgement call here yourself?

alwx: @TyDunn I think it makes sense to keep this one as it as and maybe create a new issue because what Joe was talking about makes the scope much bigger

alwx:
@TyDunn please approve to merge this one.

TyDunn (approved):

Issue #8762, opened 2021-05-28
Add support to Python 3.9


m-vdb:
[Python 3.9](https://www.python.org/downloads/release/python-390/) was released on October 5, 2020.

To be able to add support for this version, we'll need rasa's dependencies to also support it. [Tensorflow seems to now support it](https://github.com/tensorflow/tensorflow/issues/44485), but we need to check that the various tensorflow extensions and other dependencies support it.

The draft PR: https://github.com/RasaHQ/rasa/pull/10002

**Definition of Done**
- [x] Update rasa-sdk to support Python 3.9
- [x] https://github.com/RasaHQ/rasa/issues/10119 closed
- [x] Make sure our dependencies support Python 3.9
- [x] Update the CI
- [x] Update developer environments
- [ ] Merge https://github.com/RasaHQ/rasa/pull/10002
- [ ] Close https://github.com/RasaHQ/rasa/issues/10115
- [ ] Close https://github.com/RasaHQ/rasa/issues/10110
- [ ] ~Add 3.9 classifier to pyproject.toml like https://github.com/RasaHQ/rasa/pull/10405~

alwx:
Closing it because of this one: https://github.com/RasaHQ/rasa/issues/10119

m-vdb:
@alwx should we treat the Python 3.9 separated to upgrading `transformers`? There's an obvious dependency between the two, but upgrading to Python 3.9 needs work on its own (which you already started doing). If you agree, I'd keep this issue open to track it in our backlog

alwx:
The draft PR: https://github.com/RasaHQ/rasa/pull/10002

PR #10459, opened 2021-12-06
10378/tf27

### LINKED ISSUE #10379: Tensorflow/Transformers dependency updates
- [ ] https://github.com/RasaHQ/rasa/issues/10119
- [ ] https://github.com/RasaHQ/rasa/issues/10378
- [ ] https://github.com/RasaHQ/rasa/issues/10358 (consider this for TF2.7 instead)

**Status**:
- cf. TL;DR section in https://github.com/RasaHQ/rasa/pull/10459#issue-1071875401

**Updates**:
- Nov 11th: https://github.com/RasaHQ/rasa/issues/10119#issuecomment-966279173
- Dec 6th: https://github.com/RasaHQ/rasa/pull/10459#issue-1071875401

ka-bu:
**Proposed changes**:
- closes #10379 
- updated `temp_keras_model` + our sequence generator with changes from tf 2.7
    - replaced keras dependencies with `tf.keras`/ `keras`
- removed seemingly unused dependencies (e.g. tf probability)

## Progress

- [x] need fix/workaround for [this issue](https://github.com/huggingface/transformers/issues/14346)
    - passing a tensor with the attention mask instead of numpy array did the trick
- [x] investigate differences in embeddings  
   - [x] decrease tolerance successively to check how much differences there are (just as a test)
   - [x] investigate which configurations are affected and now much (compared to tf 2.6)
   - [x] check for api changes / issues that could explain differences
   - [x] run regression tests afterwards to check impact
- [x] ~~switch to 3.9 for tests (cf. PR by Alex)~~ (see #8762)
- [x] fix sentencepiece dependency 
   - [x] figure out how to properly mark it as optional / transformer related (seems that must be added as main dependency)
- [x] inspect timings and memory consumption in comparison to tf 2.6 
   - [x] created datadog notebook (`10378/tf27`)
   - [x] investigate why the memory measure measurements are broken (see below)
   - [x] try to workaround via datadog
   - [x] workaround via nvidia-smi/mprof on cloud (see `research experiments`, branch `kb/tf27`)
   - [x] ~~spike on tb / keras callbacks~~
- [x] realise `tf.python.keras` code exists but is obsolete
    - [x]  check for bits of tf.python left and find replacements in `keras` or `tf.keras` if accessible from there
    - [x] re-run regression tests 
- [x] revisit sentencepiece / transformer dependency -> clean approach `transformer[sentencepiece]`
- [x] fix unexpected updated dependencies 

### `tf.python` / `TempKerasModel`
- *Note: `tf.python` is to be used for private stuff and not public use. Ideally, we should remove those references if possible (e.g. those also appear in modules not touched in this PR such as layers).*
- When we updated the `TempKerasModel` according to the changes in the reference implementation, this resulted in some problems:
  - ~~for data generators, the keras export does not work well (the adapters `ALL_ADAPTER_CLS` in tf's `data_adapter` won't recognise it, leading to `ValueError: Failed to find data adapter that can handle input: <class 'rasa.utils.tensorflow.data_generator.RasaBatchDataGenerator'>,`)~~
  - ~~the `tf_util` needed in`temp_keras_model` only exists in `tf.python...` (same function as in non-temp tf model :)~~
- however, this reference implementation that we had used was the one from `tf.python.keras` ... 
   - ... but the implementation of the `tf.keras`Model can actually be found in the keras repository
   - => The above problems vanished and we replaced the functionality with that from `keras` directly

### `transformers`

- recommended way to install transformer's slightly changed and to get the supported pre-trained models working, we now need to make the sentencepiece import dependency explicit (see [this discussion with more details](https://github.com/RasaHQ/rasa/pull/10459#discussion_r795895693))
- sounds like we should be able to use non-dev transformer version with 3.9 e.g. [here]. 
  (https://github.com/huggingface/transformers/issues/9410#issuecomment-761616082) 

#### Overview: `transformers` versions
  * `4.x` versions tested caused trouble because of  `'tuple' object doesn't have attribute as_list` in 4.12.2 -> [issue](https://github.com/huggingface/transformers/issues/14346) and potential fix already exists
     *  issue description sounds like `4.6.1` isn't affected but it is :/ 
     *  -> was resolved by the workaround with the tf.tensor
     * we end up with the same issue as with 3.5, i.e. differences in logits (more details below)
  * `3.5`: tests fail cause of slighty larger differences in tests now ( more details below ) 
   * `>=2.4,<2.12`:  is what we used before 
        * depends on `tokenizers 0.7.0` for which there is no python 3.9 build of
        
### DataDog monitoring

- turned out monitoring intervals are not good enough to capture GPU utilisation/memory consumption properly
- References: see [Matt's explanation on the granularity of measurements](https://github.com/RasaHQ/rasa/pull/10459#issuecomment-1017607706) and our DataDog notebook  (`10378/tf27`)

### General Problem: Dependency Updates

- resolving lock file conflicts lead to updated dependencies (because not all dependencies are up to date...)
- some of those are breaking changes:
    - e.g. [Sanic caused problems in the very end](https://github.com/RasaHQ/rasa/pull/10459#issuecomment-1026733011) which is why we need to pin it

  
### Differences in Embeddings 

1. curent vs updated version
  - details: https://github.com/RasaHQ/rasa/tree/10378/tf/local/test-tf27 
  - compared (locally):
      1. `transformers 4.13`, `tf 2.7.0` and `python 3.9.9`
      2. current version `tf 2.6.1`, `transformers 2.11`, `python 3.8.10`
  - computed:
   -  maximum over all element-wise absolute differences of the (first 5 dimensions of the) actual and expected embeddings (computed as in `test_lm_featurizer_shapes_in_process_training_data`)
   - computed the differences between these deviations using 1. and 2., respectively (per text)
  - result: 
      -  for all configurations != {gpt, xlnet} the results stay roughly the same, i.e. the maximum deviation between expected and actual outcome increases by only  < 1e-7
      - for gpt, the max. deviation increases by 1e-3 
      - for xlnet, the max. deviation increases by 1e-2
  - Note: the differences between expected and actual embeddings are also > (precision of the hard coded expected embeddings) in the current version (1.)
2. same as 1. without attention mask
   - tested the same as above, with attention_mask commented out in the rasa versions of both environments to check whether it has something to do with that (cause there are some issues related to attention on hugginface repo) but ...  
   - that yields larger differences for even more models :) 
3. other
 -  tokenizers (and our inputs) seem to be ok (i.e. verified that test case for gpt yield the same token ids in both versions), as to be expected 😅


usc-m:
/modeltest

```yaml
include:
  - dataset: ["all"]
     config: ["all"]
 ```

**Comment by ka-bu** [comment]
More comparisons can be found in the corresponding datadog *notebook* (not dashboard) named "1037-tf27".  -- We're continuing to look into (fixing) the funny GPU mem/utilisation measurements, but from the time comparison (also in datadog, see notes above last two tables) and the scores (comment above), I expect no surprises (let me know if you do) and so I'll mark the PR ready for review to not delay this more later.

**Comment by koernerfelicia** [comment]
> for data generators, the keras export does not work well (the adapters ALL_ADAPTER_CLS in tf's data_adapter won't recognise it, leading to ValueError: Failed to find data adapter that can handle input: <class 'rasa.utils.tensorflow.data_generator.RasaBatchDataGenerator'>,)

This caught my eye because it reminded me of a discussion from last upgrade (to 2.6), so noting here for mine and Daksh's benefit. We thought we had to make this change during upgrade to TF 2.6, but then we [weren't able to reproduce](https://github.com/RasaHQ/rasa/pull/9649#discussion_r720050805) the original error so reverted the change. *This time* I was able to reproduce:

```
ValueError: Failed to find data adapter that can handle input: <class 'rasa.utils.tensorflow.data_generator.RasaBatchDataGenerator'>, <class 'NoneType'>
```

**Comment by usc-m** [comment]
> We're continuing to look into (fixing) the funny GPU mem/utilisation measurements

So we were looking into if we could increase the sampling rate on the Datadog agent to try and capture more granular changes if they existed. From looking at the docs it is possible for us to control the sampling rate, but it's unclear what our limits are (most of the actual hard rate limits are defined in terms of events/hour/organisation - so not clear what's the maximum we could push any particular agent to). So we could just start guessing values but it might take an arbitrary amount of time to actually find anything, or we may find nothing at all - so we're not trying this approach any longer.

**Comment by ka-bu** [comment]
@koernerfelicia @dakshvar22 , as expected, I couldn't reproduce any unexpected differences between the two versions via GCP (some results are [here](https://github.com/RasaHQ/research-experiments/tree/kb/tf27/results/2.6vs2.7/results/carbon-bot)). Only the inference times seemed a bit odd -- in both versions -- but that should get better when [this](https://github.com/RasaHQ/rasa/issues/10728) is fixed. Let me know if something is still missing

**Comment by dakshvar22** [comment]
> some results are here). Only the inference times seemed a bit odd

@ka-bu Inference time results aren't in that folder or [the notebook](https://app.datadoghq.eu/notebook/23125/10378-tf27). Can I see what kind of odd behaviour you are seeing? I mean is it odd between TF 2.6 and TF 2.7? Asking because [this issue](https://github.com/RasaHQ/rasa/issues/10728) will fix both TF 2.6 and TF 2.7 and not just one of them.

File: rasa/utils/tensorflow/data_generator.py [hunk 1]
```diff
@@ -3,15 +3,15 @@
 import logging
 import scipy.sparse
 import numpy as np
-import tensorflow as tf
+from tensorflow.python.keras.utils.data_utils import Sequence
```
dakshvar22, line 6: I thought we don't want to use `tf.python` as its private. Can't we leave it to `tf.keras` as that seems to be the public API from [TF 2.6 onwards](https://github.com/tensorflow/tensorflow/releases?page=2)

rasa/utils/tensorflow/temp_keras_modules.py
```diff
@@ -13,16 +13,18 @@
...
-from keras.callbacks import Callback, History
-from keras.engine import training, training_utils, data_adapter, base_layer
-from keras.engine.data_adapter import DataHandler
+from tensorflow.keras.callbacks import Callback, CallbackList, History
+from tensorflow.keras.models import Model
+
+from tensorflow.python.keras.engine import training, training_utils, data_adapter
 from tensorflow.python.eager import context
-from keras.utils import tf_utils, version_utils
+from tensorflow.python.keras.utils import tf_utils, version_utils
```
dakshvar22, line 21: We are making a trade-off here by using a private API of TF, i.e. `tf.python` in favour of keeping `keras` as a dependency. It kind of goes against the recommendation of [TF changelog for version 2.6](https://github.com/tensorflow/tensorflow/releases?page=2)

Are we okay relying on a private API? It might be prone to too many breaking changes between versions? 😬

dakshvar22 (commented):
Haven't looked through the complete PR yet, but some questions on using `tf.python` v/s keeping `keras` as dependency

rasa/utils/tensorflow/data_generator.py
ka-bu: That's needed because of the 
> for data generators, the keras export does not work well (the adapters ALL_ADAPTER_CLS in tf's data_adapter won't recognise it, leading to ValueError: Failed to find data adapter that can handle input: <class 'rasa.utils.tensorflow.data_generator.RasaBatchDataGenerator'>,)

rasa/utils/tensorflow/temp_keras_modules.py
ka-bu: Thanks for pointing this out again! My bad. Not sure what I was thinking. I'll test something and then ping you again.

ka-bu: So, to some up what was done before and why:
- what I had wanted to avoid was mixing stuff from `tf.keras` and `keras` and to have a separate keras dependency, which we also had in the poetry file at some point (for I don't know what reason). 
- what I had used as reference was [this](https://github.com/tensorflow/tensorflow/blob/r2.7/tensorflow/python/keras/engine/training.py#L882) (i.e. just looking at updates of the [keras model reference implementation we had used before](https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/keras/engine/training.py#L824-L1146), as was also stated [here](https://github.com/RasaHQ/rasa/blob/main/rasa/utils/tensorflow/temp_keras_modules.py#L25)).

Now we have:
- actually the [tf docs for the tf.keras.Model](https://www.tensorflow.org/api_docs/python/tf/keras/Model) actually link [this](https://github.com/keras-team/keras/blob/v2.7.0/keras/engine/training.py#L66-L2947) 
-  which means the mismatch isn't a potential issue as I thought and we can use `keras` imports directly
- (still don't know why there are those two model implementations, but the `tf.keras` one seems to be the "right" thing)

@dakshvar22 , thanks again!

@usc-m  , could you please double-check the last three changes and the above make sense?

usc-m: So if I'm understanding correctly it looks like `tf.keras` is `keras` everywhere we care about, so there is no mismatch issue and we can drop using the private imports? If that understanding agrees with yours then yes, that all makes sense.

ka-bu: .... apart from `tf.keras.Model` which does link to `tf.python.keras` and not to `keras.models.Model`. Would not have expected that from the link in the doc. Am I mixing something up here?

rasa/utils/tensorflow/temp_keras_modules.py
```diff
@@ -179,28 +372,40 @@ def fit(
...
+                        "issue/bug to `tf.keras`."
+                    )
                 epoch_logs = copy.copy(logs)
 
                 # Run validation.
                 if validation_data and self._should_eval(epoch, validation_freq):
                     # Create data_handler for evaluation and cache it.
                     if getattr(self, "_eval_data_handler", None) is None:
-                        self._eval_data_handler = CustomDataHandler(
+                        self._eval_data_handler = data_adapter.get_data_handler(
```
ka-bu, line 408: This was kind of a copy and paste bug on my side but -- now that I see it I think we should leave it as is. If we ever make use of this, we wouldn't use the custom data handler with the increasing batch size, no?

rasa/utils/tensorflow/temp_keras_modules.py
usc-m: If you do it from the REPL by importing the class and `inspect` and doing `inspect.getfile` on the class, it doesn't go to `tf.python` - I guess pycharm is caching the lookups?

ka-bu: Yep, just the __file__ attribute also points in the right direction, matching what's in the docs 👍

ka-bu:
> > some results are here). Only the inference times seemed a bit odd
> 
> @ka-bu Inference time results aren't in that folder or [the notebook](https://app.datadoghq.eu/notebook/23125/10378-tf27). Can I see what kind of odd behaviour you are seeing? I mean is it odd between TF 2.6 and TF 2.7? Asking because [this issue](https://github.com/RasaHQ/rasa/issues/10728) will fix both TF 2.6 and TF 2.7 and not just one of them.

Sorry for the late reply -- Yep the strange behaviour between tf 2.7 and tf 2.6 --- and also between different regression test runs for the same tf2.6 version -- can be seen in the notebook: very last table focusing on carbon-bot data only shows strange drops of maximum gpu usage

Regarding inference: For memory consumption, I can generate corresponding plots for inference (running with GPU) tomorrow. Didn't want to create spam but doesn't hurt to check that as well.  For the timings, there's the regression results above + measurements in datadog that are consistent across runs imho.

rasa/utils/tensorflow/temp_keras_modules.py
ka-bu: @dakshvar22 and @koernerfelicia , to sum up:  as confirmed by @usc-m (who just figured out the funny tf export story which looks like tf.python.keras should  be unused), the current version (i.e. using `keras` model not the `tf.python.keras` model as reference) should be the way to go. (Thanks again @dakshvar22  and @usc-m :))

ka-bu:
Two tests failed because of connection timeouts / installation problems. The rest looks fine. Any other comments/questions, or can we merge? :)

dakshvar22:
I hope to take a final look in an hour 😬

rasa/utils/tensorflow/layers.py
```diff
@@ -2,9 +2,12 @@
 from typing import List, Optional, Text, Tuple, Callable, Union, Any
 import tensorflow as tf
 import tensorflow_addons as tfa
+
+# TODO: The following is not (yet) available via tf.keras
+from keras.utils.control_flow_util import smart_cond
```
dakshvar22, line 7: But I see that `keras` is removed as a dependency in the poetry file. Does it come with the tensorflow package anyway?

rasa/utils/tensorflow/temp_keras_modules.py
```diff
@@ -2,28 +2,50 @@
...
     # TODO
-    #  we don't need this anymore once
+    #  we don't need this anymore once the fix from
     #  https://github.com/tensorflow/tensorflow/pull/45338
-    #  is merged and released
+    #  has been ported over to keras and merged there
 
     # This code is adapted from
-    # https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/keras/engine/training.py#L824-L1146
+    # https://github.com/keras-team/keras/blob/v2.7.0/keras/engine/training.py#L65
```
dakshvar22, line 46: Since its really just the `fit` function, this line directly takes you to the function 😬 

```suggestion
    # https://github.com/keras-team/keras/blob/v2.7.0/keras/engine/training.py#L902
```

rasa/utils/tensorflow/temp_keras_modules.py
dakshvar22: Yeah, this seems correct

dakshvar22 (changes requested):
Looks great! Just one question about the `keras` dependency to be sure.
Other than that, please add a changelog and migration guide entry.

If this is a model breaking change (are models trained with a previous version of Rasa Open Source loadable with this new version, without re-training of course?) then that should be mentioned in the migration guide and changelog.

Also, since the embedding values have marginally changed, we can't guarantee the same model performance / output if:

1. Config uses `LanguageModelFeaturizer`
and
2. A model is re-trained with the new rasa open source version without changing the config / data / random seeds **or** a model trained with previous version of rasa open source is loaded with this new version and inference is made on it.

These details can go in the changelog and migration guide IMO :)

pyproject.toml
```diff
@@ -171,8 +168,8 @@ datadog = "^0.43.0"
 [tool.poetry.extras]
 spacy = [ "spacy",]
 jieba = [ "jieba",]
-transformers = [ "transformers",]
-full = [ "spacy", "transformers", "jieba",]
+transformers = [ "transformers", "sentencepiece"]
```
dakshvar22, line 171: So `sentencepiece` is an optional dependency on `transformers` side and hence we also need to add it as a dependency?

rasa/utils/tensorflow/layers.py
ka-bu: yep - https://github.com/keras-team/keras#installation

dakshvar22: Hahah, and then there are modules that are not in `tf.keras` 🤦

pyproject.toml
ka-bu: yep - I can double check one last time, but I think we had trouble when we didn't include it explicitly this way

ka-bu: @usc-m could you double check this was a required dep by transformer that won't be installed otherwise? For me it looks fine here, can't get it to complain

usc-m: It's a required dep for the models we use, for the previous version of `transformers` it was a required dependency, but for the newer one it's an extra, so we can install it with `transformers[sentencepiece]` (which I just discovered confirming this, and would respect their constraints on the version of `sentencepiece` better) or add it as an explicit dependency like this. (They do their dependencies in a really strange way so it's very non-obvious this is the case)

https://github.com/huggingface/transformers/blob/master/setup.py#L269 (current)
vs
https://github.com/huggingface/transformers/blob/v2.11.0/setup.py#L125 (previous)

usc-m: As a side note that's maybe more useful for research to know, this means that `transformers` is now (in new versions) installed in a 'lightweight' way where the dependencies for some models are not installed by default. So you may encounter runtime failures due to missing dependencies because of this

rasa/utils/tensorflow/temp_keras_modules.py
```diff
@@ -225,35 +444,35 @@ def fit(
                 if self.stop_training:
                     break
 
-            # If _eval_data_handler exists, delete it after all epochs are done.
+            # If eval data_hanlder exists, delete it after all epochs are done.
```
koernerfelicia, line 447:
```suggestion
            # If eval_data_handler exists, delete it after all epochs are done.
```
actually this typo is also in the original TF code.. so maybe no point in fixing here 🤷‍♀️

koernerfelicia (approved):
Looks good to me! I made a small typo suggestion, but not that important since the typo came from TF itself.

Thank you so much for adding the "Adaption" comments, and all the explanations to the imports 🚀

ka-bu:
> If this is a model breaking change (are models trained with a previous version of Rasa Open Source loadable with this new version, without re-training of course?) then that should be mentioned in the migration guide and changelog.

Thanks for the reminders and sorry for nearly skipping this 🤦‍♀️  

Loading and `rasa shell`-ing a model trained with the version from `main` works without problem (also with `LanguageModelFeaturizer` in the pipeline). 

But `rasa train`-ing that model again (with the version from this branch) triggers retraining. @usc-m , can you confirm this? No idea why we would not load from the cache then...

pyproject.toml
ka-bu: We should put this comment in the code, I think. Maybe above the list of the pretrained models? - Or a notion page on transformer related things?

dakshvar22:
> No idea why we would not load from the cache then…

Because the features from the featurizers would change right?

pyproject.toml
dakshvar22: > installed in a 'lightweight' way where the dependencies for some models are not installed by default. So you may encounter runtime failures due to missing dependencies because of this

Would it work for at least the pre-trained models that we support officially and mention in the[ docs](https://rasa.com/docs/rasa/components/#languagemodelfeaturizer)?

dakshvar22:
> Because the features from the featurizers would change right?

Ohh but the data wouldn't have changed, so it should actually skip training. Iteresting edge case for the caching feature 😂

ka-bu:
>>Because the features from the featurizers would change right?

> Ohh but the data wouldn't have changed, so it should actually skip training. Iteresting edge case for the caching feature 😂


No you're right I think, because the input for e.g. DIET specifically changed and because we: https://github.com/RasaHQ/rasa/blob/413ac0551721709ca78dc1e890e075a75fa35533/rasa/shared/nlu/training_data/message.py#L140

pyproject.toml
koernerfelicia: @dakshvar22 I don't think so, [looks like](https://huggingface.co/docs/transformers/tokenizer_summary#sentencepiece) we need it for XLNet?

changelog/10378.improvement.md
```diff
@@ -0,0 +1,8 @@
+Bump TensorFlow version to 2.7.
+
+:::caution
+We can't guarantee the exact same output and hence model performance if your 
+configuration uses `LanguageModelFeaturizer`. This applies to the case where the 
+model is re-trained with the new rasa open source version without changing the 
+configuration, random seeds, and data as well as to the case where a model trained with 
+a previous version of rasa open source is loaded with this new version for inference.
```
dakshvar22, line 8:
```suggestion
a previous version of rasa open source is loaded with this new version for inference.

We suggest training a new model if you are upgrading to this version of Rasa Open Source.
```

docs/docs/migration-guide.mdx
```diff
@@ -12,6 +12,22 @@ import TabItem from '@theme/TabItem';
...
 how you can migrate from one version to another.
 
+
+## Rasa 3.0 to 3.1
+
+### Machine Learning Components
+
+#### TensorFlow Upgrade
+
+Due to the TensorFlow upgrade, we ce can't guarantee the exact same output and hence
```
dakshvar22, line 22:
```suggestion
Due to the TensorFlow upgrade, we can't guarantee the exact same output and hence
```

docs/docs/migration-guide.mdx
```diff
@@ -12,6 +12,22 @@ import TabItem from '@theme/TabItem';
...
+model performance if your configuration uses `LanguageModelFeaturizer`.
+This applies to the case where the model is re-trained with the new rasa open source
+version without changing the configuration, random seeds, and data as well as to the
+case where a model trained with a previous version of rasa open source is loaded with
+this new version for inference.
+
+Please check whether your trained still performs as expected and retrain if needed.
```
dakshvar22, line 29:
```suggestion
Please check whether your trained model still performs as expected and retrain if needed.
```

dakshvar22 (approved):

pyproject.toml
```diff
@@ -191,16 +188,17 @@ version = "^3.1"
...
 
 [tool.poetry.dependencies."github3.py"]
 version = "~1.3.0"
 optional = true
 
 [tool.poetry.dependencies.transformers]
-version = ">=2.4,<2.12"
+version = "~4.13.0"
 optional = true
+extras = [ "sentencepiece",]
```
dakshvar22, line 201: Does this mean that if a user did `pip install rasa[transformers]`, it wouldn't bring in `sentencepiece` automatically?

usc-m: No, it means that it will bring in `transformers`, and we've specified transformers should be brought in with it's `sentencepiece` extra - so basically what we've done is outsourced installing `sentencepiece` to `transformers` since it isn't an explicit dependency for us but one we have because of `transformers` in the first place. Doesn't make sense for us to play catch up and mirror their dependency stuff into our project every time we need to update `transformers`

dakshvar22: Makes perfect sense, thanks for clarifying 🙏

usc-m:
Looks like Poetry bumped us to `sanic` 21.12.* when I was resolving the last lock file issues which appears to have a lot of breaking changes for us - investigating if we can pin this to the last working version to get this merged

PR #10470, opened 2021-12-06
Merge 2.8.x into main

rasabot:
:auto_rickshaw: This PR should be merged automatically once it has been approved. If it doesn't happen:
- [ ] Handle merge conflicts
- [ ] Fix build errors

:bulb: It has been opened automatically after changes were merged in a feature branch.


m-vdb:
@usc-m @samsucik would you mind taking a look at the changes in this PR? It's about merging 2.8.x back into main. Thanks

m-vdb (approved):

samsucik (commented):
@m-vdb two things:
1. Just ensuring I understand the role of `changelog/10413.improvement.md`: That file describes an improvement which has been released in a 2.8 micro and now will be also released in a 3.0 micro (and at that point the file's content will be turned into an entry in `CHANGELOG`)?
2. I see a changelog entry being added that says were bumping TF to 2.6.2, but in `pyproject.toml` on this PR I see this line: `tensorflow = ">=2.6,<2.6.2"`. That's not expected, is it?

m-vdb:
> I see a changelog entry being added that says were bumping TF to 2.6.2, but in pyproject.toml on this PR I see this line: tensorflow = ">=2.6,<2.6.2". That's not expected, is it?

yeah this is deliberate because we have a ticket that will take care of upgrading tensorflow on `main`. But maybe I'm wrong. @usc-m WDYT?

> Just ensuring I understand the role of changelog/10413.improvement.md: That file describes an improvement which has been released in a 2.8 micro and now will be also released in a 3.0 micro (and at that point the file's content will be turned into an entry in CHANGELOG)?

So I assumed it was in the diff because 2.8.x was not released yet, but maybe I'm wrong. Let me see why this file is still here 🤔

m-vdb:
@samsucik the reason why the changelog fragment for #10413 is still here is because it hasn't been released yet. So it's all good :) Can you approve the PR if everything is good on your end? (I'll wait for @usc-m approval to merge of course)

samsucik (approved):

rasa/core/brokers/kafka.py
```diff
@@ -194,10 +196,24 @@ def _publish(self, event: Dict[Text, Any]) -> None:
...
-        self.producer.send(self.topic, value=event, key=partition_key)
+
+        self.producer.send(self.topic, value=event, key=partition_key, headers=headers)
 
     def _close(self) -> None:
         self.producer.close()
+
+    @rasa.shared.utils.common.lazy_property
+    def rasa_environment(self) -> Optional[Text]:
+        """Get value of the `RASA_ENVIRONMENT` environment variable."""
```
samsucik, line 218: @m-vdb in case you're after extremely nit-picky comments 😂 
```suggestion
        """Gets the value of the `RASA_ENVIRONMENT` environment variable."""
```

samsucik (approved):

rasa/core/brokers/kafka.py
m-vdb: we don't really adjust the code in these PRs, it should have been done in the original PR 😬  (this is documented on `main` [here](https://github.com/RasaHQ/rasa/blob/main/.github/PULL_REQUEST_AUTOMATIC_TEMPLATE.md))

m-vdb: but thanks for noticing!

usc-m:
>yeah this is deliberate because we have a ticket that will take care of upgrading tensorflow on main. But maybe I'm wrong. @usc-m WDYT?

Yeah, that's the plan - we're in the middle of updating to 2.7 on `main` right now which will probably alter a bunch of the dependencies changed here. Technically the changelog will be wrong though as @samsucik says, so we might want to consider fixing that - but code wise it seems fine.

m-vdb:
🤔 yeah for completeness I would add a note in the changelog item that this only applies to 2.8 at the moment and we'll upgrade Tensorflow on 3.x independently.

m-vdb:
@usc-m I've added a note in the changelog, please review when you have the time!

usc-m (approved):

PR #10479, opened 2021-12-07
Remove the duplication of valid SlotSet events in the tracker by FormAction

### LINKED ISSUE #9945: Fix already valid `SlotSet` events being duplicated in the tracker by the `FormAction`

ancalita:
**Proposed changes**:
- Fixes #9945 

### DIFF
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -398,11 +398,10 @@ class FormAction(LoopAction):
         )
 
         some_slots_were_validated = any(
-            isinstance(event, SlotSet)
+            isinstance(event, SlotSet) and not event.key == REQUESTED_SLOT
             for event in validation_events
             # Ignore `SlotSet`s  for `REQUESTED_SLOT` as that's not a slot which needs
             # to be filled by the user.
-            if isinstance(event, SlotSet) and not event.key == REQUESTED_SLOT
         )
 
         # extract requested slot
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -379,18 +379,10 @@ class FormAction(LoopAction):
         If nothing was extracted reject execution of the form action.
         Subclass this method to add custom validation and rejection logic
         """
-        slot_values = self._get_slot_extractions(tracker, domain)
+        extracted_slot_values = self._get_slot_extractions(tracker, domain)
 
         validation_events = await self.validate_slots(
-            slot_values, tracker, domain, output_channel, nlg
-        )
-
-        some_slots_were_validated = any(
-            isinstance(event, SlotSet)
-            for event in validation_events
-            # Ignore `SlotSet`s  for `REQUESTED_SLOT` as that's not a slot which needs
-            # to be filled by the user.
-            if isinstance(event, SlotSet) and not event.key == REQUESTED_SLOT
+            extracted_slot_values, tracker, domain, output_channel, nlg
         )
 
         # extract requested slot
@@ -398,7 +390,7 @@ class FormAction(LoopAction):
 
         if (
             slot_to_fill
-            and not some_slots_were_validated
+            and not extracted_slot_values
             and not self._user_rejected_manually(validation_events)
         ):
             # reject to execute the form action
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -258,22 +258,15 @@ class FormAction(LoopAction):
         validate_name = f"validate_{self.name()}"
 
         if validate_name not in domain.action_names_or_texts:
-            return events
+            return []
 
         _tracker = self._temporary_tracker(tracker, events, domain)
         _action = RemoteAction(validate_name, self.action_endpoint)
         validate_events = await _action.run(output_channel, nlg, _tracker, domain)
 
-        validated_slot_names = [
-            event.key for event in validate_events if isinstance(event, SlotSet)
-        ]
-
-        # If the custom action doesn't return a SlotSet event for an extracted slot
-        # candidate we assume that it was valid. The custom action has to return a
-        # SlotSet(slot_name, None) event to mark a Slot as invalid.
-        return validate_events + [
-            event for event in events if event.key not in validated_slot_names
-        ]
+        # Only return the validated SlotSet events by the slot validation action to
+        # avoid adding duplicate SlotSet events for slots that are already valid.
+        return validate_events
 
     def _temporary_tracker(
         self,
--- a/rasa/core/actions/forms.py
+++ b/rasa/core/actions/forms.py
@@ -248,7 +248,10 @@ class FormAction(LoopAction):
 
         Returns:
             The validation events including potential bot messages and `SlotSet` events
-            for the validated slots.
+            for the validated slots, if the custom form validation action is present in
+            domain actions.
+            Otherwise, returns empty list since the extracted slots already have
+            corresponding `SlotSet` events in the tracker.
         """
         logger.debug(f"Validating extracted slots: {slot_candidates}")
         events: List[Union[SlotSet, Event]] = [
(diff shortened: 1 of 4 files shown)


rasa/core/actions/forms.py
```diff
@@ -386,26 +379,18 @@ async def validate(
...
+            extracted_slot_values, tracker, domain, output_channel, nlg
         )
 
         # extract requested slot
         slot_to_fill = self.get_slot_to_fill(tracker)
 
         if (
             slot_to_fill
-            and not some_slots_were_validated
+            and not extracted_slot_values
```
ka-bu, line 393: What if the validation adds events that un-set the slots? -- Think this is might have been the reason why the `validate_slots` returned those (validated) slot set events and why there was this search for the validated slots. 

=> Seems to me that we can keep the changes in `validation_events` (looks so much cleaner now) and should then search through the `validation_events` first (from the end) to figure out whether the `slot_to_fill` was un-set in the `validation_events` (even though it might have been set before)

Does that make sense? 🤔

ka-bu: Think we should also update the docstring here to make clear that we're only returning the new/added events that were created via the validation and that we raise if the action gets rejected.

Actually, this is strange. Why do we raise here? Thought the event list represents what really happens but what could happen is that we validate + then reject but the validation events won't end up in a tracker then, no?

rasa/core/actions/forms.py
```diff
@@ -258,22 +258,15 @@ async def validate_slots(
         validate_name = f"validate_{self.name()}"
 
         if validate_name not in domain.action_names_or_texts:
-            return events
+            return []
```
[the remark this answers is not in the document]
ka-bu: would also be good to add some docstr to the temporary tracker (or maybe there's also a better name? 🤔 ) -- wasn't  really aware that the validation action would not receive the original tracker  but just the slot set events plus some very selectively chosen events :D -- Think we even might need to add that to some docs. Wdyt?

ka-bu (commented):
The fix looks really nice and straightforward :) However, I'm not sure that we're not missing some edge cases now 🤔 . Some questions in the comments.

rasa/core/actions/forms.py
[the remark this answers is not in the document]
ancalita: Yup, sure, will see what I can do about the docstring and if anything in Forms section of the docs should be updated too. I guess the idea with the temporary tracker is that we only want to look at SlotSet events from last user turn.

rasa/core/actions/forms.py
ancalita: My personal interpretation was that if the form requests a slot but nothing got extracted from the user message (either to set requested_slot or other slots), then the form has to be rejected.
Maybe better check is:
```
if (
    slot_to_fill
    and (not extracted_slot_values or not some_slots_were_validated)
    ....
```

ancalita: Actually scrap that - probably it's better to link by AND, not OR, always get caught out by truth tables 🙈

ka-bu: Hm... I think it is very ok to be confused... like couldn't / wouldn't it be ok if the validation adds a slot set event even though nothing was extracted? 🙈 

Also shouldn't the `some_slots_were_validated` also check for the *last* event per slot? So that if a slot eventually un-set we catch that (just in case for some reason something funny happens that adds multiple set and unset events)?

ancalita: > wouldn't it be ok if the validation adds a slot set event even though nothing was extracted?

That would be completely fine, that would actually be expected for custom slot mappings requested by a form.
The reason why I just thought I have to check that both (pre)`extracted_slot_values` and `some_slots_were_validated` are empty is to mimic the previous code - `some_slots_were_validated` used to be a contained for both predefined slots that do not get validated, predefined slots that do get validated, and slots with custom mappings. So to maintain this, I have to implement this longer check 😄

rasa/core/actions/forms.py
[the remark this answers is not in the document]
ka-bu: Think that's a good guess and if that is the only reason, then I'd vote for removing that implicit restriction in the long run. Short term we should definitely explicitly state that in the docs because it could be very confusing  for people who write custom actions and expect the tracker to contain the same things they see when they implement a custom rule policy :/.

rasa/core/actions/forms.py
ancalita: > shouldn't the some_slots_were_validated also check for the last event per slot?

I don't immediately see why we'd need this extra filtering, also to optimize the tracker size?

ka-bu: What I mean is, that the validation could return `SlotSet("slot1", None)` while the predefined mappings had extracted `SlotSet("slot1", <invalid choice>)`, and then the "or" would evaluate to false even though we should abort since that slot1 is effectively not filled. -- Plus the custom validation action could return a events like `[SlotSet('slot1', <maybe this value instead?>), SlotSet("slot1", None)]". 
Am I missing something here? Doesn't that need to be resolved somewhere?

Btw, we could also cleanup this bit of the old code: 
```
some_slots_were_validated = any(
            isinstance(event, SlotSet)
            for event in validation_events
            # Ignore `SlotSet`s  for `REQUESTED_SLOT` as that's not a slot which needs
            # to be filled by the user.
            if isinstance(event, SlotSet) and not event.key == REQUESTED_SLOT
        )
```
to this: 
```
some_slots_were_validated = any(
            isinstance(event, SlotSet) and not event.key == REQUESTED_SLOT
            for event in validation_events
            # Ignore `SlotSet`s  for `REQUESTED_SLOT` as that's not a slot which needs
            # to be filled by the user.
        )
```
(this is so weird, but that extra isinstance check is really pointless, isn't it? 👀 )

rasa/core/actions/forms.py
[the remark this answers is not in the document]
ancalita: I've extended a sentence in the `Validating Form Input` docs section, that seemed the only appropriate place to clarify that only recent `SlotSet` events are validated. The style of the docs is not to go too heavy on internal implementation details, so I hope this is good enough, a paragraph on some temporary tracker used when running custom action could be a bit confusing for new users.

rasa/core/actions/forms.py
ancalita: Yeah, I'll apply that code clean-up suggestion 👍🏼 

Regarding validation reseting slots to `None`, that's an eligible validation event and it should not abort/ reject the form; I've actually changed OR to AND in the changes I pushed. Also, even if the custom form validation action returns the double events `[SlotSet('slot1', <maybe this value instead?>), SlotSet("slot1", None)]`, finally the value of 'slot1' will be `None` in the tracker.

ancalita: Oh I realised why this code looks a bit funky, we can't actually clean it up, because `validation_events` could contain other events than `SlotSet` and those events don't have `key` attribute (such as `BotUttered`).

ka-bu: > Regarding validation reseting slots to None, that's an eligible validation event and it should not abort/ reject the form; ...

Oh ok, I see. That old comment on setting slots to None to mark them to invalidate got me thinking that we should reject if a slot is effectively removed via the setting to None. (And the more I think about it this looks like inconsistent behaviour because we have no way of using the validation events to correct the slot extraction such that it looks like no slot was extracted -- We'll always have a SlotSet to None event. But it is consistent with the old behaviour then now so... )

> Oh I realised why this code looks a bit funky, we can't actually clean it up, 

I'm not sure I know what you mean, but maybe this helps: Python evaluates this expression lazily, meaning that the `not event.key == REQUESTED_SLOT` will only be evaluated if `isinstance(event, SlotSet)` evaluates to `True` and hence the simplified expression which avoids the double isinstance check will work fine (this is also why the the expression in the "if" part of the original expression works)

ka-bu (approved):
Thanks for addressing the comments and answering questions! Would still rephrase that one funky expression :D but otherwise looks good to me  :)

rasa/core/actions/forms.py
ancalita: Ahhh, I just assumed mypy could complain (without first testing this), thanks for the explanation 👍🏼

PR #10486, opened 2021-12-08
rasa data migrate 3.0

### LINKED ISSUE #10243: `rasa data migrate` fails with a cryptic error if run on a 3.0 domain file
The proposed docs for `rasa data migrate` mention:
> Exceptions will be raised and the migration process terminated if invalid domain files are provided, slots or forms
are missing from your original files[...]

I (accidentally) ran the command on an already migrated domain file (a case not mentioned above) and got an error:
```bash
  File "/Users/sams/.miniconda/envs/qaenv/bin/rasa", line 8, in <module>
    sys.exit(main())
  File "/Users/sams/.miniconda/envs/qaenv/lib/python3.8/site-packages/rasa/__main__.py", line 121, in main
  File "/Users/sams/.miniconda/envs/qaenv/lib/python3.8/site-packages/rasa/cli/data.py", line 259, in _migrate_domain
    rasa.core.migrate.migrate_domain_format(args.domain, args.out)
  File "/Users/sams/.miniconda/envs/qaenv/lib/python3.8/site-packages/rasa/core/migrate.py", line 262, in migrate_domain_format
    new_forms, updated_slots = _migrate_form_slots(original_domain)
  File "/Users/sams/.miniconda/envs/qaenv/lib/python3.8/site-packages/rasa/core/migrate.py", line 89, in _migrate_form_slots
    for slot_name, mappings in form_data.items():
AttributeError: 'list' object has no attribute 'items'
```
I think it'd be nice to first check if the file looks like a valid 3.0 domain and if it does, inform the user instead of trying to migrate the file.

**Definition of done**
- [ ] Stop and warn user that they are trying to migrate a 3.0 domain file with the 2.0 --> 3.0 migration tool

ka-bu:
**Proposed changes**:
- closes #10243
- added assertion that files that we want to migrate do not contain the 3.0 version tag already 
- prevent users from overwriting existing output or backup location (which prevents us from removing user data when cleaning up the backup location in case migration fails)
- added check so that `assemble_new_domain` only works on domain files
- remove backup of original domain in case a domain file is given (we only removed the backup directory previously)

**Questions:**
-  this is still missing cases where the original files did not contain a version tag and then were migrated - should we drop the "key == version" checks and force a "version: 3.0" to be added if it doesn't exist?


rasa/cli/arguments/data.py
```diff
@@ -74,6 +74,8 @@ def set_migrate_arguments(parser: argparse.ArgumentParser) -> None:
 
     add_out_param(
         parser,
-        default=DEFAULT_DOMAIN_PATH,
-        help_text="Path (for `yaml`) where to save migrated domain in Rasa 3.0 format.",
+        default=None,
+        help_text="Path (for `yaml`) where to save migrated domain in Rasa 3.0 format."
+        "If none is specified, a generic file/folder will be created in the folder "
```
ancalita, line 79: This can even be more specific regarding the names of the generated `out_path`:
```suggestion
        "If none is specified, either a `new_domain.yml` file or `new_domain` folder will be created in the folder "
```

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
-            out_file.mkdir()
-            created_out_dir = True
-            rasa.shared.utils.io.raise_warning(
-                f"The out path provided is not a directory, "
-                f"creating a new directory '{str(out_file)}' "
-                f"for migrated domain files."
-            )
+    # Choose a default output location if nothing was specified
+    if out_path is None:
+        suffix = DEFAULT_DOMAIN_PATH if migrate_file_only else "new_domain"
```
ancalita, line 290: I'm actually debating now if we shouldn't actually use a new suffix for the new domain file, instead of overwriting `DEFAULT_DOMAIN_PATH`, it would be more consistent this way. WDYT?
```suggestion
        suffix = "new_domain.yml" if migrate_file_only else "new_domain"
```

rasa/core/migrate.py
```diff
@@ -178,6 +182,8 @@ def _write_final_domain(
 ) -> None:
     if domain_file.is_dir():
         for file in domain_file.iterdir():
+            if not Domain.is_domain_file(file):
+                continue
```
ancalita, line 186: This case needs a unit test 😄

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+
+    # Ensure the backup location does not exist yet
+    # Note: We demand that file as well as folder with this name gets deleted before
+    # the command is run to avoid confusion afterwards.
+    suffix = "original_domain"
+    suffix = f"{suffix}.yml" if migrate_file_only else suffix
+    backup_location = domain_parent_dir / suffix
+    if backup_location.exists():
+        backup_location_str = "directory" if backup_location.isdir() else "file"
+        raise RasaException(
```
ancalita, line 282: These lines 281-282 need a unit test too.

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+            f"The domain from '{domain_path}' could not be migrated since the "
+            f"a {backup_location_str} {backup_location} already exists."
```
ancalita, line 284:
```suggestion
            f"{backup_location_str} '{backup_location}' already exists."
```

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+            f"Please remove that there is no file or folder at {backup_location}."
```
ancalita, line 285:
```suggestion
            f"Please make sure that there is no {back_up_location_str} at '{backup_location}'."
```

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+        suffix = DEFAULT_DOMAIN_PATH if migrate_file_only else "new_domain"
+        out_path = domain_parent_dir / suffix
 
-        try:
-            original_domain = _migrate_domain_files(
-                domain_file, backup_location, out_file
+    # Ensure the output location is not already in-use
+    if not migrate_file_only:
+        if out_path.is_dir() and any(out_path.iterdir()):
+            raise RasaException(
```
ancalita, line 296: Unit test please 😄

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+                f"The domain from '{domain_path}' could not be migrated to "
+                f"{out_path} because that folder is not empty."
```
ancalita, line 298:
```suggestion
                f"'{out_path}' because that folder is not empty."
```

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+                "Please remove the folder and try again."
```
ancalita, line 299:
```suggestion
                "Please remove the contents of the folder and try again."
```

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
-                    out_file.mkdir()
-            raise e
     else:
-        if not Domain.is_domain_file(domain_file):
+        if out_path.is_file():
             raise RasaException(
-                f"The file '{domain_file}' could not be validated as a "
-                f"domain file. Only domain yaml files can be migrated. "
+                f"The domain from '{domain_path}' could not be migrated to "
+                f"{out_path} because a file already exists."
```
ancalita, line 305:
```suggestion
                f"'{out_path}' because a file already exists."
```
Plus I think this might need a unit test too?

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+        if domain_path.is_dir()
+        else [domain_path]
+    )
+    migrated_files = [
+        file
+        for file in original_files
+        if rasa.shared.utils.io.read_yaml_file(file).get("version") == "3.0"
+    ]
+    if migrated_files:
+        raise RasaException(
```
ancalita, line 323: Unit test please

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
 
-    _write_final_domain(domain_file, new_forms, new_slots, out_file)
+    # Validate given domain file(s) and migrate them
+    try:
+        created_out_dir = False
+        if not migrate_file_only:
+            if not out_path.is_dir():
+                rasa.shared.utils.io.raise_warning(
+                    f"The out path provided did not exist yet. Created directory "
+                    f"{out_path}."
```
ancalita, line 336:
```suggestion
                    f"The out path provided does not exist. Created directory "
                    f"'{out_path}'."
```

rasa/core/migrate.py
```diff
@@ -238,59 +262,114 @@ def _migrate_domain_files(
...
+        if backup_location.is_dir():
+            shutil.rmtree(backup_location)
+        if out_path.is_dir():
+            if created_out_dir:
+                shutil.rmtree(out_path)
+            else:  # just remove contained files so we do not mess with access rights
+                for f in out_path.glob("*"):
+                    f.unlink()
+        if backup_location.is_file():
+            backup_location.unlink()
```
ancalita, line 374: Needs to be unit tested.

tests/core/test_migrate.py
```diff
@@ -916,3 +941,22 @@ def test_migration_cleanup(tmp_path: Path,):
...
+        domain_dir,
+        """
+        version: "3.0"
+        """,
+        "domain.yml",
+    )
+
+    with pytest.raises(
+        RasaException,
+        match=f"The domain directory '{domain_dir}' does not contain any domain "
```
ancalita, line 959: A bit confused here why this exception message gets raised? Shouldn't the message be the one that says that `some of the given files were already migrated to 3.0`?

ancalita (commented):
Great work 🥇 This module looks much better, left a few suggestions on exception messages and requested some more unit tests to be added.

Re. the forcing of version `3.0`? In which scenarios does the version doesn't exist in a domain file, that's quite strange? Is it because the key is not a required key in training data files?

rasa/core/migrate.py
ka-bu: yesss, I like that :)

tests/core/test_migrate.py
ka-bu: Ohhh, yep, forgot to fix this. Files not containing any of the indicator keys (like "intents" and so on) are not considered domain files. So the above is no domain file actually -- Need to add some key and change the message here

ka-bu:
> Re. the forcing of version 3.0? In which scenarios does the version doesn't exist in a domain file, that's quite strange? Is it because the key is not a required key in training data files?

Guess that should've been my question: Are there cases where the version key doesn't exist? 🤔  (Doesn't hurt to cover that case though, no?)

rasa/core/migrate.py
ka-bu: done

rasa/core/migrate.py
ka-bu: done

rasa/core/migrate.py
ka-bu: done

rasa/core/migrate.py
ka-bu: done

tests/core/test_migrate.py
ka-bu: done

rasa/shared/utils/validation.py
```diff
@@ -242,6 +242,7 @@ def validate_training_data_format_version(
         return True
 
     try:
+        version_value = version_value.strip("\"'")
```
ancalita, line 245: Why is this required? Also causes a unit test to fail 🤔

ancalita (approved):
Looks good to go 🎉 Only needs some code quality fixes and then some clarification on the extra line in the validation which makes and existing test to fail.

rasa/core/migrate.py
ka-bu: done :)

rasa/core/migrate.py
ka-bu: done

rasa/shared/utils/validation.py
ka-bu: because the version tags in the configs will look like `version: '3.0'` and the yaml read function will convert that to `"'3.0'"` (if we just write `3.0`, then we get an int and a type check will fail instead)

PR #10492, opened 2021-12-09
Unset the default values for `num_threads` and `finetuning_epoch_fraction` to `None` in order  to fix cases when CLI defaults override the data from config.

### LINKED ISSUE #10409: `num_threads` and `finetuning_epoch_fraction` CLI defaults overwrite user config
### Rasa Open Source version

3.0.0

### Rasa SDK version

_No response_

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

In `default_recipe.py`,  `_extra_config_from_cli` will always return something for the `num_threads` (and for the finetuning fraction, if we're in finetuning mode) (because otherwise the [`cli_parameter[param]` call would fail](https://github.com/RasaHQ/rasa/blob/main/rasa/engine/recipes/default_recipe.py#L364)). If the user does not specify any CLI parameters then that something are the default (CLI) parameters which will then [overwrite the user settings](https://github.com/RasaHQ/rasa/blob/main/rasa/engine/recipes/default_recipe.py#L340):

To reproduce this, e.g. create a config with MITIE components and set the num_threads
```
pipeline:
- name: "MitieNLP"
  model: "<add-your-path-here>/total_word_feature_extractor.dat"
- name: "MitieTokenizer"
- name: "MitieFeaturizer"
- name: "MitieEntityExtractor"
- name: "MitieIntentClassifier"
  num_threads: 200000
```
and add a breakpoint to `_add_nlu_train_mode`

### Command / Request

_No response_

### Relevant log output

_No response_

**Definition of done**
- [ ] Ensure that `num_threads` can be set
- [ ] Determine a way to test this properly
- [ ] Add a test

alwx:
**Proposed changes**:
- fix #10409 

I unset the default values to fix this because in this case:
1) the default config is being used anyway
2) it makes it possible to still use CLI `--num-threads` and `--epoch-fraction` to override the values


ka-bu:
Sorry if my description wasn't clear there. What I mean is then when I **use the CLI**, the `cli_parameters` dict is not empty but populated with the defaults even though I don't pass any arguments in the command line. The new tests do not cover that case - we'd need to test `rasa train` there

alwx:
@ka-bu could you check it again?

I did another fix, and now I basically updated the default values and made sure `--num-threads` and `--epoch-fraction` can still be used to override the values coming from the config.
We don't have any tests for CLI arguments extraction, so in this case I extended the previously created test to make sure that in case nothing is extracted (= no extra arguments for `rasa train` specified) the default config is being used.

rasa/engine/recipes/default_recipe.py
```diff
@@ -374,9 +374,11 @@ def _extra_config_from_cli(
             old_number_epochs = component_config.get(
                 EPOCHS, component.get_default_config()[EPOCHS]
             )
-            epoch_fraction = float(cli_parameters["finetuning_epoch_fraction"])
-
-            config_from_cli[EPOCHS] = math.ceil(old_number_epochs * epoch_fraction)
+            epoch_fraction = cli_parameters["finetuning_epoch_fraction"]
```
ka-bu, line 377: how about we 
 ```suggestion
            epoch_fraction = cli_parameters.get("finetuning_epoch_fraction", 1.0)
```
and drop the "if" to make it more explicit that the default is 1.0 (in case we need to find that again one day ^^). Ideally, we should grab the default from some `constants.py`, but seems like there is none for that? 👀

File: rasa/engine/recipes/default_recipe.py [hunk 2]
```diff
@@ -374,9 +374,11 @@ def _extra_config_from_cli(
...
+            if epoch_fraction is not None:
+                config_from_cli[EPOCHS] = math.ceil(
```
ka-bu, line 379: ... why is this even computed here? Doesn't whatever receives the fraction parameter and the number of epochs downstream compute this (again)?

jupyterjazz: Agree with @ka-bu,  I think `config_from_cli["finetuning_epoch_fraction"] = epoch_fraction` should be enough here

rasa/engine/recipes/default_recipe.py
jupyterjazz: +1 to adding constants for `1.0` and `finetuning_epoch_fraction`

jupyterjazz (changes requested):

rasa/engine/recipes/default_recipe.py
alwx: In this case all this block doesn't have any sense — `config_from_cli["finetuning_epoch_fraction"]` will be equal to `cli_parameters["finetuning_epoch_fraction"]` after the config from CLI is extracted (line 363)
so I guess the reason to do this calculation is to fill `config_from_cli[EPOCHS]`, isn't it?

alwx: Updated the code, not sure if `config_from_cli[EPOCHS]` needs to be removed, can you confirm?

rasa/engine/recipes/default_recipe.py
```diff
@@ -363,7 +363,7 @@ def _extra_config_from_cli(
         config_from_cli = {
             param: cli_parameters[param]
             for param in cli_args_mapping.get(component, [])
-            if param in cli_parameters
+            if param in cli_parameters and cli_parameters[param] is not None
         }
 
         if (
```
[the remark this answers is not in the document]
alwx: Done!

rasa/engine/recipes/default_recipe.py
jupyterjazz: Sure, makes sense

jupyterjazz (approved):
Looks good to me! :rocket:
Since we use `"finetuning_epoch_fraction"` so frequently now, should we add a constant for it? (also for its default value - `1.0`)

tests/engine/recipes/test_default_recipe.py
```diff
@@ -161,6 +161,53 @@ def test_generate_graphs(
...
+        (
+            {"num_threads": 1, "finetuning_epoch_fraction": 0.5},
+            "train_MitieIntentClassifier6",
+            # there is no `epochs` value specified so it doesn't get overridden
+            {"num_threads": 1, "finetuning_epoch_fraction": 0.75},
+        ),
+        (
+            {"finetuning_epoch_fraction": 0.5},
+            "train_DIETClassifier7",
+            {"epochs": 150, "num_threads": 200000, "finetuning_epoch_fraction": 0.5},
```
tayfun, line 191: What's the difference between DIET and Mitie classifiers? I mean both don't have `epochs` value specified in parameters so why Mitie has not been overwritten while DIET has been?

jupyterjazz: @tayfun, it's because `MitieIntentClassifier` doesn't need `epochs` (more specifically, its default config doesn't contain `epochs`)

PR #10495, opened 2021-12-09
prepared release of version 3.0.2

tayfun:
**Proposed changes**:
- Micro release


tayfun:
Not sure why but CI check is picking up a changelog on `main` and failing. I'll restart `make release`.

tayfun:
It probably happened because I initially had the PR against `main` and changing it didn't work.

Issue #7645, opened 2020-12-29
Incorrect warning: Action 'utter_chitchat' is listed as a response action in the domain file, but there is no matching response defined.


boryslawskamaria:
**Rasa version**: 2.2.0

**Rasa SDK version** (if used & relevant):

**Rasa X version** (if used & relevant):

**Python version**: Python 3.6.9

**Operating system** (windows, osx, ...): ubuntu

**Issue**:
I keep getting the above error even though I've included responses in my domain file. 

**Error (including full traceback)**:
```
..projects/voicebot-about-baltazar/.env/lib/python3.6/site-packages/rasa/shared/utils/io.py:93: UserWarning: Action 'utter_about_baltazar' is listed as a response action in the domain file, but there is no matching response defined. Please check your domain.
  More info at https://rasa.com/docs/rasa/responses
../projects/voicebot-about-tbaltazar/.env/lib/python3.6/site-packages/rasa/shared/utils/io.py:93: UserWarning: Action 'utter_chitchat' is listed as a response action in the domain file, but there is no matching response defined. Please check your domain.
  More info at https://rasa.com/docs/rasa/responses


```

**Command or request that led to error**:
```
rasa train
```

**Content of configuration file (config.yml)** (if relevant):
```yml
language: en

pipeline:
   - name: WhitespaceTokenizer
   - name: RegexFeaturizer
   - name: LexicalSyntacticFeaturizer
   - name: CountVectorsFeaturizer
   - name: CountVectorsFeaturizer
     analyzer: char_wb
     min_ngram: 1
     max_ngram: 4
   - name: DIETClassifier
     epochs: 100
   - name: EntitySynonymMapper
   - name: ResponseSelector
     epochs: 100
     retrieval_intent: chitchat
   - name: ResponseSelector
     epochs: 100
     retrieval_intent: about_baltazar
   - name: DIETClassifier
     epochs: 100
   - name: FallbackClassifier
     threshold: 0.3
     ambiguity_threshold: 0.1

policies:
   - name: MemoizationPolicy
   - name: TEDPolicy
     max_history: 5
     epochs: 100
   - name: RulePolicy

```

**Content of domain file (domain.yml)** (if relevant):
```yml
version: "2.0"

intents:
  - chitchat
  - greet
  - goodbye
  - affirm
  - deny
  - thanks
  - mood_great
  - mood_unhappy
  - bot_challenge
  - what_bot_does

responses:

  utter_default:
  - text: No one could understand that input.
[...]

PR #10487, opened 2021-12-08
Fix warning when action for subintent has response

tayfun:
Fixes https://github.com/RasaHQ/rasa/issues/7645

**Proposed changes**:
- Take subintent action utterances into consideration when checking for missing responses

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] ~updated the documentation~ - NO NEED
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #10511, opened 2021-12-10
Fix warning when action for subintent has response (#10487)

tayfun:
Fixes https://github.com/RasaHQ/rasa/issues/7645

**Proposed changes**:
- Backport fixes at https://github.com/RasaHQ/rasa/pull/10487

- [ ] ~updated the documentation~ - NO NEED

### DIFF
--- a/rasa/core/agent.py
+++ b/rasa/core/agent.py
@@ -863,7 +863,7 @@ class Agent:
 
         if isinstance(domain, str):
             domain = Domain.load(domain)
-            domain.check_missing_responses()
+            domain.check_missing_templates()
             return domain
         elif isinstance(domain, Domain):
             return domain
--- a/rasa/core/agent.py
+++ b/rasa/core/agent.py
@@ -863,7 +863,7 @@ class Agent:
 
         if isinstance(domain, str):
             domain = Domain.load(domain)
-            domain.check_missing_templates()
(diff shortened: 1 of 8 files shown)


rasa/core/agent.py
```diff
@@ -863,7 +863,7 @@ def _create_domain(domain: Union[Domain, Text, None]) -> Domain:
 
         if isinstance(domain, str):
             domain = Domain.load(domain)
-            domain.check_missing_responses()
```
tayfun, line 866: This was a bug not noticed before, API v3 used in v2 🤔  Wonder when it would be run into.

rasa/shared/importers/importer.py
```diff
@@ -329,7 +329,7 @@ async def get_domain(self) -> Domain:
         )
 
         existing_domain = existing_domain.merge(domain_with_retrieval_intents)
-        existing_domain.check_missing_responses()
```
tayfun, line 332: This was a bug not noticed before, API v3 used in v2 🤔 Wonder when it would be run into.

rasa/core/agent.py
tayfun: Actually I'm totally wrong 🤦  let me fix this.

usc-m (approved):
LGTM

PR #10533, opened 2021-12-15
Introduce unit tests for GH actions scripts

markus-hinsche:
**Proposed changes**:
- Setup target `test-gh-actions` in Makefile, specify category
- Add poetry `dev-dependencies` and update `poetry.lock`
- Prepare modules under test: Drop strict dependence of environment variables so the module is testable without
- Add two test modules

Potential caveats (Feedback from reviewers encouraged):
- I currently use `sys.path.append` because `.github/scripts/` is not yet in the `PYTHONPATH`. Any ideas?
- The test coverage (reported after running unit tests in CI) only for `rasa/`, but not `.github/scripts`


virtualroot (approved):

.github/scripts/mr_publish_results.py
```diff
@@ -18,17 +18,17 @@
 DD_SERVICE = "rasa"
 METRIC_RUNTIME_PREFIX = "rasa.perf.benchmark."
 METRIC_ML_PREFIX = "rasa.perf.ml."
-IS_EXTERNAL = os.environ["IS_EXTERNAL"]
-DATASET_REPOSITORY_BRANCH = os.environ["DATASET_REPOSITORY_BRANCH"]
+IS_EXTERNAL = os.getenv("IS_EXTERNAL")
```
wochinge, line 21: isn't `getenv` a bit risky as it hides the case when the env variable for some reason isn't filled? Or are they all optional? The old approach had the advantage that it would fail early in this case

tests/gha/test_mr_publish_results.py
```diff
@@ -0,0 +1,77 @@
+from pathlib import Path
+import sys
+
+REPO_DIR = Path(__file__).parents[2]
```
wochinge, line 4: this seems fragile - can we do this in a more robust way?

tests/gha/test_mr_generate_summary.py
```diff
@@ -0,0 +1,94 @@
+from pathlib import Path
```
wochinge, line 1: Awesome that we start testing this code as well! 🚀 

Wondering if it's worth separating these "infrastructure" tests from the production code tests. In my opinion it should be a separate job / workflow in the GitHub actions to clearly separate things 🤔 Treating this as the regular production tests also makes these infrastructure tests a dependency for CI steps which target the production code which seems off to me 🤔

markus-hinsche: We could move the test files from `tests/gha` to `.github/tests`. Would this be what you have in mind?

I fear that when splitting `.github/workflows/continous-integration.yml` s.t. `test-gh-actions` ran in a new separate workflow, we would duplicate a lot of CI code. But maybe this is worth to break the dependency?

tests/gha/test_mr_publish_results.py
markus-hinsche: I agree this looks ugly. 

Since `.github/` is not a package (has no `__init__.py`), I am not sure how to best do an import. 

If we decided to move the tests into `.github/` (see other comment), a relative import might work. 

Happy to hear more ideas

markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
   config: ["Sparse + BERT + DIET(seq) + ResponseSelector(t2t)", "BERT + DIET(seq) + ResponseSelector(t2t)"]

File: tests/gha/test_data/intent_report.json [hunk 4]
```diff
@@ -0,0 +1,120 @@
...
+		"f1-score": 1.0,
+		"support": 28
+	},
+	"weighted avg": {
+		"precision": 1.0,
+		"recall": 1.0,
+		"f1-score": 1.0,
+		"support": 28
+	}
+}
```
wochinge, line 120:
```suggestion
}

```

tests/gha/test_mr_generate_summary.py
wochinge: > We could move the test files from tests/gha to .github/tests. Would this be what you have in mind?

That sounds good 👍🏻 

> we would duplicate a lot of CI code.

Is it actually? These tests are pretty lightweight so all we'd need is 
1. Check out source code
2. Install dependencies
3. Test with a single Python version

Isn't it? 
Curious to hear @joejuzl 's opinion on it

joejuzl: Agree it makes sense to separate from production code tests 👍

markus-hinsche: Tests are now separated

.github/scripts/mr_publish_results.py
```diff
@@ -238,6 +224,17 @@ def send_to_segment(context: Dict[str, Any]):
     )
 
 
+def _get_is_external_and_dataset_repository_branch():
+    is_external = os.environ["IS_EXTERNAL"]
+    dataset_repository_branch = os.environ["DATASET_REPOSITORY_BRANCH"]
+    if str(is_external).lower() in ("yes", "true", "t", "1"):
```
wochinge, line 230: environment variables are always strings, aren't they?
```suggestion
    if is_external.lower() in ("yes", "true", "t", "1"):
```

.github/scripts/mr_publish_results.py
```diff
@@ -254,10 +251,10 @@ def read_results(file: str) -> Dict[str, Any]:
     return result
 
 
-def push_results(file_name: str, file: str):
+def _push_results(file_name: str, file: str):
```
wochinge, line 254:
```suggestion
def _push_results(file_name: str, file: str) -> None:
```

.github/workflows/ci-github-actions.yml
```diff
@@ -0,0 +1,60 @@
...
+    - '*'
+  pull_request:
+
+jobs:
+  test:
+    name: Run Tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
```
wochinge, line 17: do we even need the matrix? There will be always just one set to test right?

.github/workflows/ci-github-actions.yml
```diff
@@ -0,0 +1,60 @@
...
+      run: |
+        echo "POETRY_VERSION=$(scripts/poetry-version.sh)" >> $GITHUB_ENV
+      shell: bash
+
+    - name: Install poetry 🦄
+      uses: Gr1N/setup-poetry@v7
+      with:
+        poetry-version: ${{ env.POETRY_VERSION }}
+
+    - name: Load Poetry Cached Libraries ⬇
```
wochinge, line 42: don't we need an extra step to also install the dependencies? Fully relying on the cache seems risky

markus-hinsche: 🙈, good catch

.github/workflows/ci-github-actions.yml
markus-hinsche: I removed the matrix

wochinge (approved):
Looks good to me. Do we want these tests to run on each PR or only when the scripts where changed?

markus-hinsche:
I thought about running it on every PR in the beginning. If it turns out to be annoying for others, we can restrict it of course

markus-hinsche:
@wochinge Sure, making it required would be great. I am not sure if I have enough permissions to configure this in GitHub though. Could you set the check as required?

PR #10560, opened 2021-12-21
Fix bug: MR tests runs (on self-hosted runners) can impede each other

markus-hinsche:
**Bug:** I encountered a bug: If appears when using self-hosted runners, and running multiple CI regression tests (e.g. 2 PRs) in parallel. In this case, each CI run creates a self-hosted runner. Each test job will then execute on any runner they find (`runs-on: [self-hosted, gpu]`). At first sight, this seems good because a CI run can also use the runner and can therefore parallelize. However, when one CI run finishes, it destroys the runner, while the other CI run might have a job running on this runner, which causes this job to fail.

**Proposed changes**:
- Only let a job run on the runner created by it's own CI run.
- also for on-schedule workflow

**For reviewers**:
- There might be better ways to solve this issue. Let me know about your ideas.


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

virtualroot (approved):

PR #10562, opened 2021-12-21
Update `rasa data validate` to not fail when `active_loop` is `null`

### LINKED ISSUE #10447: `rasa data validate` fails on example formbot
### Rasa Open Source version

3.0.1

### Rasa SDK version

3.0.0rc3

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

`cd` into `rasa/examples/formbot`, run `rasa data validate` and get 
```bash
UserWarning: The form 'None' is used in the 'submit form' block, but it is not listed in the domain file. You should add it to your domain file!
```
This seems to break validation of all forms of all bots. Training works, though.

### Command / Request

```shell
`rasa data validate`
```

### Relevant log output

```shell
(rasa-3.0.0) johannes:formbot/ (b055f901954) $ rasa data validate                                                               [18:28:57]
/Users/johannes/rasa/rasa/rasa/shared/utils/io.py:99: UserWarning: The form 'None' is used in the 'submit form' block, but it is not listed in the domain file. You should add it to your domain file!
  More info at https://rasa.com/docs/rasa/forms
Project validation completed with errors.
```

**Definition of done**
- [ ] Fix bug

alwx:
**Proposed changes**:
- fix #10447 


tests/cli/test_rasa_data.py
```diff
@@ -202,6 +202,30 @@ def test_validate_files_form_not_found_invalid_domain(
         data.validate_files(namedtuple("Args", args.keys())(*args.values()))
 
 
+def test_validate_files_with_active_loop_null(tmp_path: Path):
```
jupyterjazz, line 205: I tried running the test locally and it failed. 
It fails because of this line: https://github.com/RasaHQ/rasa/blob/02fa04c358e7c8d19a970c6c5d4fbbc0c2e7d2ee/rasa/validator.py#L436
Returns false which results in an invalid domain and thus an error.
However, I see that the test is passed on CI. Any ideas about this?

alwx: Hm I was able to run it locally without issues. The only reason why it can be is when `data/test_restaurantbot/domain.yml`, you sure it's not the case?

jupyterjazz (approved):
:+1:

PR #10564, opened 2021-12-21
Use multiple MR test runners to parallelize

markus-hinsche:
**Motivation**: Currently when multiple MR tests (e.g. multiple configs and/or datasets) are executed on a self-hosted runner, the runner doesn't parallelize this. This is because the `deploy_runner_gpu` job spins up only 1 node.

**Proposed changes**:
- In Kubernetes deployment, increase `replicas` to (up to) 3 ([Kub docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment))
- If less jobs are in a CI run, then less replicas will be used
- Make the same changes for on-schedule workflow

**Checked:**
- If only one test should be run, then use only one replica
    - DONE check for 2 jobs: 2 runners, run in parallel
    - DONE check for 4 jobs: 3 runners, 3 out of 4 in parallel
- No changes necessary for CPU (since this is still running on a Github runner, and runs in parallel already)


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
   config: ["Sparse + DIET(bow) + ResponseSelector(bow)", "BERT + DIET(bow) + ResponseSelector(bow)"]
```

virtualroot (approved):

PR #10574, opened 2021-12-22
Fix issue of job continuing although 'Run Tests' step failed

markus-hinsche:
**Bug:** Even if jobs fail (e.g. because of 'Run Tests' step), the CI run will continue and mark the run as successful. Example: https://github.com/RasaHQ/rasa/runs/4550410412?check_suite_focus=true

**Proposed changes**:
- Don't `continue-on-error`, but set `fail-fast=False` instead: This will still let all the `model_regression_test_cpu` jobs run even if there is a failing job. However, it will declare `needs.model_regression_test_cpu.result` as `failed` (previously with `continue-on-error` this was `success` confusing the jobs down the line.
- always try to stop the datadog agent
- Don't run `combine_reports` and `add_comment_results` in case of failure (this could be changed though)
- Make `add_comment_results` only dependent on `combine_results`
- How to proceed with on-schedule workflow?: `analyse_performance` job should run `always()`, therefore this change might not be desired for the on-schedule workflow. Therefore proposing to keep it as is.

**Checked:**
- [x] when failure on CPU: https://github.com/RasaHQ/rasa/pull/10542
- [x] when failure on GPU: https://github.com/RasaHQ/rasa/pull/10542 -> still deletes runner [example run](https://github.com/RasaHQ/rasa/actions/runs/1612153928)
- [x] success CPU: new PR: https://github.com/RasaHQ/rasa/pull/10574
- [x] success GPU: new PR: https://github.com/RasaHQ/rasa/pull/10574
- [x] when we have many jobs, some get skipped on purpose -> `needs.model_regression_test_cpu.result` should still return 'success'


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

virtualroot (approved):

Issue #10587, opened 2021-12-27
`test_nlu` breaks when optional value isn't given


koaning:
### Rasa Open Source version

3.0.4

### Rasa SDK version

 _No response_ 

### Rasa X version

 _No response_ 

### Python version

3.7

### What operating system are you using?

Linux

### What happened?

I am building smoke tests for rasa nlu examples and it seems that in the 3.0 transition some of the internal testing methods break if optional parameters are not passed.

Here's the script in particular.

<pre>import asyncio
 from rich.traceback import install
 from rasa.model_training import train_nlu
 from rasa.model_testing import test_nlu

install()

if __name__ == "__main__": 
 fp = "sparse-naive-bayes-intent-classifier-config.yml"
 mod = train_nlu(
 nlu_data="tests/data/nlu/en/nlu.yml",
 config=f"tests/configs/

 {fp}",

 output="models",

 )

 asyncio.run(test_nlu(model=mod, nlu_data="tests/data/nlu/en/nlu.yml", additional_arguments={}))</pre>

 The final line is the important one. If `additional_arguments` are not passed I get this traceback.

 <pre>╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮

 │ /home/vincent/Development/rasa-nlu-examples/scripts/train-nlu.py:17 in 

 │

 │ │

 │ 14 │ │ config=f"tests/configs/{fp} 

", │
 │ 15 │ │ output="models", │
 │ 16 │ ) │
 │ ❱ 17 │ asyncio.run(test_nlu(model=mod, nlu_data="tests/data/nlu/en/nlu.yml")) │
 │ 18 │
 │ │
 │ /usr/lib/python3.7/asyncio/runners.py:43 in run │
 │ │
 │ 40 │ try: │
 │ 41 │ │ events.set_event_loop(loop) │
 │ 42 │ │ loop.set_debug(debug) │
 │ ❱ 43 │ │ return loop.run_until_complete(main) │
 │ 44 │ finally: │
 │ 45 │ │ try: │
 │ 46 │ │ │ _cancel_all_tasks(loop) │
 │ │
 │ /home/vincent/Development/rasa-nlu-examples/uvloop/loop.pyx:1456 in │
 │ uvloop.loop.Loop.run_until_complete │
 │ │
 │ <span class="error">[Errno 2]</span> No such file or directory: │
 │ '/home/vincent/Development/rasa-nlu-examples/uvloop/loop.pyx' │
 │ │
[...]

PR #10588, opened 2021-12-27
Update model_testing.py

koaning:
Fixes https://github.com/RasaHQ/rasa/issues/10587.

### DIFF
--- a/rasa/model_testing.py
+++ b/rasa/model_testing.py
@@ -128,7 +128,7 @@ async def test_core(
     model: Optional[Text] = None,
     stories: Optional[Text] = None,
     output: Text = DEFAULT_RESULTS_PATH,
-    additional_arguments: Optional[Dict] = None,
+    additional_arguments: Dict = {},
     use_conversation_test_files: bool = False,
 ) -> None:
     """Tests a trained Core model against a set of test stories."""


rasa/model_testing.py
```diff
@@ -128,7 +128,7 @@ async def test_core(
     model: Optional[Text] = None,
     stories: Optional[Text] = None,
     output: Text = DEFAULT_RESULTS_PATH,
-    additional_arguments: Optional[Dict] = None,
+    additional_arguments: Dict = {},
```
wochinge, line 131: We can't have mutable defaults in method params https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

wochinge (changes requested):

rasa/model_testing.py
koaning: Ah right! Good catch. 

What would be an alternative solution to https://github.com/RasaHQ/rasa/issues/10587 though? I can internally make a change so that internally the function goes: 

```python
if not additional_arguments:
    additional_arguments = {}
```

Would that suffice? The main gripe I have is an optional type that isn't optional.

wochinge: isn't that what's done [here](https://github.com/RasaHQ/rasa/blob/909374144524f16ec99025c151ce4e0d3eff662c/rasa/model_testing.py#L159)?

koaning: Your link is from `test_core`, I'm concerned about `test_nlu` [here](https://github.com/RasaHQ/rasa/blob/909374144524f16ec99025c151ce4e0d3eff662c/rasa/model_testing.py#L188-L222).

wochinge: sorry, I was confused as you've changed the param in `test_core` 🙈 

You're right, we'd need to do what you said in `test_nlu` as well:

```python
if not additional_arguments:
    additional_arguments = {}
```

koaning: Done!

rasa/model_testing.py
```diff
@@ -196,6 +196,9 @@ async def test_nlu(
 
     rasa.shared.utils.io.create_directory(output_directory)
 
+    if not additional_arguments:
```
wochinge, line 199: can we please also add a test to avoid regression?

wochinge (commented):
💎

rasa/model_testing.py
koaning: While writing the test I realized something else. 

I'm pretty sure that the `model` and `nlu_data` parameters aren't optional as-is either. Can I remove the `Optional` type here?

wochinge: good catch 👍🏻 Yeah, but please check the usages of this function then

m-vdb (approved):

Issue #10510, opened 2021-12-10
Improve Domain Loading


carlad:
### What problem are you trying to solve?

This issue comes out of an [initial bug](https://github.com/RasaHQ/rasa/issues/10137) reported by @raoulvm , where `use_entities` was not being set correctly when domain config is split over several domain `.yml` files.

Investigating this bug led to the realisation of a problem with the current `Domain` loading implementation in the case of multiple domain files.

Currently, with a split domain, the `Domain` object is constructed iteratively: with separate `Domain`s  are created from each domain file and merged together sequentially.  As a step in creating a `Domain` includes linking the `intents` in a file with the `entities` they will use:  either `True` (to use all entities in the _entire_ domain), `use_entities` (to use specific entities, when these entitites are less than half of _all_ entities), or `ignore_entities` (when these entities are less than half of _all_ entities), a problem arises when entities are configured in different domain files.  Each file does not know about entities in other files, and consequently, as each `Domain` object is created and merged sequentially, the method to link `entities` to `intents`, `_transform_intents_for_file`, can make mistakes.

### What's your suggested solution?

@raoulvm proposed a solution which has been implemented in this [PR](https://github.com/RasaHQ/rasa/pull/10424), but it duplicates code somewhat in a process that couples validation with attribute setting.

I explored the possibility of decoupling the load process by _first_ validating and merging all domain `yml` files recursively into a single `defaultdict(list)`. Locally I used split domains from a couple of customers to create single dict objects that appeared legit (see below for links to the output). 

Passing this dict to the `from_dict` method, created correct entity/intent association in the `Domain`, but broke other steps. A PR reflecting these changes, and the breakages, can be found [here](https://github.com/RasaHQ/rasa/pull/10509/files).

### Examples (if relevant)

[Example Merged Domain 1](https://www.notion.so/rasa/Customer-1-merged-Domain-6b243266e5bb422c8e8f066494dacccc) 

[Example Merged Domain 2](https://www.notion.so/rasa/Customer-2-merged-Domain-9b94d2fd6cbf4fad9e88ef5b095bdc84)

(these docs are only visible to Rasa team members)

### Definition of Done

This issue was identified using rasa 2.8.x and that's the version I've been debugging against. 
I would confine any changes to 2.8.x for this issue, then create another for porting the changes to 3.0.x

- [ ] write down rules that constrain how domains are able to be configured (including how code works and how customers split domains)
- [ ] get feedback from and align with CSE and Dev Rel on these rules
[...]

Issue #9351, opened 2021-08-13
rasa test nlu should not care if intents in stories are not the domain file


hsm207:
**Description of Problem**:

Running `rasa test nlu` on a bot that has split domain e.g:

```
|-- domain
|   |-- domain1.yml
|   |-- domain2.yml
```

will print warnings about an intent being in a story but not part of the domain e.g.:

> /opt/venv/lib/python3.8/site-packages/rasa/shared/utils/io.py:97: UserWarning: Issue found in 'data/rules.yml': 
> Found intent 'greet' in stories which is not part of the domain.
>   More info at https://rasa.com/docs/rasa/stories
> /opt/venv/lib/python3.8/site-packages/rasa/shared/utils/io.py:97: UserWarning: Issue found in 'data/stories.yml': 
> Found intent 'greet' in stories which is not part of the domain.
>   More info at https://rasa.com/docs/rasa/stories
> /opt/venv/lib/python3.8/site-packages/rasa/shared/utils/io.py:97: UserWarning: Issue found in 'data/stories.yml': 
> Found intent 'mood_great' in stories which is not part of the domain.
>   More info at https://rasa.com/docs/rasa/stories
> /opt/venv/lib/python3.8/site-packages/rasa/shared/utils/io.py:97: UserWarning: Issue found in 'data/stories.yml': 
> Found intent 'mood_unhappy' in stories which is not part of the domain.
>   More info at https://rasa.com/docs/rasa/stories
> /opt/venv/lib/python3.8/site-packages/rasa/shared/utils/io.py:97: UserWarning: Issue found in 'data/stories.yml': 
> Found intent 'affirm' in stories which is not part of the domain.

The source of these warnings is the test data importer assuming that the domain is in a file named domain.yml in the root folder (see [here](https://github.com/RasaHQ/rasa/blob/819c339f3f5cdf32593cc725ff7d8f4a8253ea3a/rasa/cli/test.py#L184-L186)).

The nlu evaluation runs fine despite these warnings so these warnings might confuse users.

**Overview of the Solution**:

Add an option to `rasa test nlu` to specify the path to a domain file or folder. For example

`rasa test nlu --domain /path/to/domain/folder`

or suppress the user warnings when loading the data. For example:

```
    warnings.filterwarnings("ignore", category=UserWarning)
    test_data_importer = TrainingDataImporter.load_from_dict(
        training_data_paths=[data_path], domain_path=DEFAULT_DOMAIN_PATH,
    )
```

[...]

PR #10589, opened 2021-12-27
add  #10137 changes to domain merging

### LINKED ISSUE #10570: Apply #10137 domain merging changes to `main`
### Rasa Open Source version

3.0

### Rasa SDK version

_No response_

### Rasa X version

_No response_

### Python version

3.8

OSX

### What happened?

We need to apply the changes that fixed #10137 in this [PR](https://github.com/RasaHQ/rasa/pull/10424) for `2.8.x` to `main`.

We need to be aware that Domain logic in `main` likely differs to that of `2.8.x`

One example is the additional method in the `merge` function on `main`:
https://github.com/RasaHQ/rasa/blob/main/rasa/shared/core/domain.py#L291

### Command / Request

_No response_

### Relevant log output

_No response_

carlad:
**Proposed changes**:
closes #10570 
- ports changes on `2.8.x` in #10137 into `main` 
- see https://github.com/RasaHQ/rasa/pull/10424 for the `2.8.x` PR


rasa/shared/core/domain.py
```diff
@@ -270,58 +275,124 @@ def merge(self, domain: Optional["Domain"], override: bool = False) -> "Domain":
...
+
+    @staticmethod
+    def merge_dicts(
+        tempDict1: Dict[Text, Any],
+        tempDict2: Dict[Text, Any],
+        override_existing_values: bool = False,
+    ) -> Dict[Text, Any]:
+        """Merges two dicts."""
+        if override_existing_values:
+            merged_dicts, b = tempDict1.copy(), tempDict2.copy()
```
ePreda, line 367: Line 281 suggest Domain-style dicts are nested, specifically `config` key. It might be worth seeing about using the `copy.deepcopy()` method from the `copy` module, to apply it down the dictionary layers.

ePreda: Right, it gets resolved earlier in the stream into a single tier dict. Got it.

ePreda (approved):

carlad:
@ePreda I realised that there was some logic in the Domain class to detect duplication that I had accidentally removed. It was added to address [#9351](https://github.com/RasaHQ/rasa/issues/9351) which I wasn't aware of.

rasa/shared/core/domain.py
```diff
@@ -270,58 +273,172 @@ def merge(self, domain: Optional["Domain"], override: bool = False) -> "Domain":
...
-            return merge_dicts
+    def merge_domain_dicts(
+        self, domain1: Dict, domain2: Dict, override: bool = False
+    ) -> Dict[Text, Any]:
+        """Merges this domain dict with another one, combining their attributes.
 
-        def merge_lists(list1: List[Any], list2: List[Any]) -> List[Any]:
-            return sorted(list(set(list1 + list2)))
+        This is used when multiple domain yml files are configured in a single
+        directory. Unlike the merge method above, which merges Domain objects by
```
joejuzl, line 307: Is there a reason we can't use this implementation for both?

carlad: @joejuzl So this PR is to port the changes from `2.8.x` into `main`, which we decided to implement in this way as we were planning to [rethink how we handle the domain](https://github.com/RasaHQ/rasa/issues/10510). 

But as we wanted to ensure that `main` also handled multiple domain files, we thought it would be better in the short-term to implement this solution...

tayfun (approved):
Approving as this is already in 2.8.x. We should come back to this and find a more elegant solution and remove duplication as well like @joejuzl has mentioned in the comment, specifically around similar merge domain methods.

joejuzl (approved):

PR #10598, opened 2021-12-28
fix domain dir merge when no intents in first file

carlad:
**Proposed changes**:
- Fixes a bug where mergeing domains from multiple files failed when the first domain file did not contain any - intents

- [x] changed some test data for the functionality

### DIFF
--- a/data/test_domains/test_domain_from_directory_tree/skill_2_domain/domain_pt3.yml
+++ b/data/test_domains/test_domain_from_directory_tree/skill_2_domain/domain_pt3.yml
@@ -1,4 +1,5 @@
 version: "2.0"
 
 intents:
+  - utter_skill_1
   - utter_skill_2
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -317,7 +317,11 @@ class Domain:
         if override or domain2.get("session_config"):
             combined[SESSION_CONFIG_KEY] = domain_dict[SESSION_CONFIG_KEY]
 
-        if combined[KEY_INTENTS] or domain_dict[KEY_INTENTS]:
+        if combined.get(KEY_INTENTS) or domain_dict.get(KEY_INTENTS):
+            if not combined.get(KEY_INTENTS):
+                combined[KEY_INTENTS] = []
+            if not domain_dict.get(KEY_INTENTS):
+                domain_dict[KEY_INTENTS] = []
             combined[KEY_INTENTS] = self.merge_lists_of_dicts(
                 combined[KEY_INTENTS], domain_dict[KEY_INTENTS], override
             )
--- a/rasa/shared/core/domain.py
+++ b/rasa/shared/core/domain.py
@@ -318,10 +318,8 @@ class Domain:
             combined[SESSION_CONFIG_KEY] = domain_dict[SESSION_CONFIG_KEY]
 
         if combined.get(KEY_INTENTS) or domain_dict.get(KEY_INTENTS):
-            if not combined.get(KEY_INTENTS):
-                combined[KEY_INTENTS] = []
-            if not domain_dict.get(KEY_INTENTS):
-                domain_dict[KEY_INTENTS] = []
+            combined[KEY_INTENTS] = combined.get(KEY_INTENTS, [])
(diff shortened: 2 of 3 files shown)


data/test_domains/test_domain_from_directory_tree/skill_2_domain/domain_pt3.yml
```diff
@@ -1,4 +1,5 @@
 version: "2.0"
 
 intents:
+  - utter_skill_1
```
jupyterjazz, line 4: What was the purpose of changing these domain files? Is there already a test for this scenario? If not, should we add it?

jupyterjazz (commented):
Do you think 3.0 has this bug as well?

rasa/shared/core/domain.py
```diff
@@ -317,7 +317,11 @@ def merge_domain_dicts(
         if override or domain2.get("session_config"):
             combined[SESSION_CONFIG_KEY] = domain_dict[SESSION_CONFIG_KEY]
 
-        if combined[KEY_INTENTS] or domain_dict[KEY_INTENTS]:
+        if combined.get(KEY_INTENTS) or domain_dict.get(KEY_INTENTS):
+            if not combined.get(KEY_INTENTS):
+                combined[KEY_INTENTS] = []
```
jupyterjazz, line 322:
```suggestion
            combined[KEY_INTENTS] = combined.get(KEY_INTENTS, [])
```

data/test_domains/test_domain_from_directory_tree/skill_2_domain/domain_pt3.yml
carlad: These files were added to my previous work updating the way the Domain is created when there are multiple domain files. Previously the test created the files in memory, so here I simply moved them into a directory of their own as most other domain test data is contained in files. I've moved the intent from the first domain file to a subsequent file because I wanted to test that a domain would be successfully merged when the first domain file contained no intents. Before, if this was the case, merging would fail with a `KeyError` (the code would look for an `'intents'` key, which didn't exist. My fix addresses that scenario, hence I wanted to have no intents in the first domain file, and moved them from there, to this file.

rasa/shared/core/domain.py
```diff
@@ -317,7 +317,11 @@ def merge_domain_dicts(
...
+            if not domain_dict.get(KEY_INTENTS):
+                domain_dict[KEY_INTENTS] = []
             combined[KEY_INTENTS] = self.merge_lists_of_dicts(
                 combined[KEY_INTENTS], domain_dict[KEY_INTENTS], override
```
ePreda, line 326: Is there a reason why we shouldn't be setting a default here (ex: 'combined.get('KEY_INTENTS', [])') instead of setting them inside conditionals?

rasa/shared/core/domain.py
carlad: thanks, I knew there was a more elegant way of writing this...

jupyterjazz (approved):
:rocket:

rasa/shared/core/domain.py
carlad: So, `combined` or `domain_dict` may or may not have any `'intents'`. This depends on the domain yml file that they are derived from. Is there a neater way to assign a default value if no key exists (other than what @jupyterjazz has suggested above?)

data/test_domains/test_domain_from_directory_tree/skill_2_domain/domain_pt3.yml
ePreda: Just for my erudition, since it's the KeyError happens at the 'OR' statement, wouldn't missing intents in either domain cause a KeyError when checked?

carlad:
@jupyterjazz 
> Do you think 3.0 has this bug as well?

I discovered the bug while porting the changes from `2.8.x` into `main`.
`3.0` will have the original merging bug (to do with `entities`), but not this one. And I've addressed it in my other PR for `main`

rasa/shared/core/domain.py
ePreda: Ah, that's fine then. That would be trick then, stylistic brevity aside.

ePreda (approved):

PR #10599, opened 2021-12-28
prepared release of version 2.8.19

carlad:


CHANGELOG.mdx
```diff
@@ -16,6 +16,9 @@ https://github.com/RasaHQ/rasa/tree/main/changelog/ . -->
 
 <!-- TOWNCRIER -->
 
+## [2.8.19] - 2021-12-28 No significant changes.
```
jupyterjazz, line 19: Does this mean that code-wise `2.8.19` is the same as `2.8.18`?

carlad: I'm not sure - this was auto-generated, then I added the line below. Perhaps I should add a changelog file for the bugfix?

jupyterjazz: We didn't have a changelog file in #10598 and that's why it wasn't shown here. I checked the `prepare-release-2.8.19` branch and it contains the changes so all good.
One thing though - I stumbled upon the same piece of code that you fixed in the last PR in `domain.py`.
https://github.com/RasaHQ/rasa/blob/cf8d6d50c35cac72dd93cc75b5ad66f37b773566/rasa/shared/core/domain.py#L270
Should we also fix this? maybe for the next release

jupyterjazz (approved):

PR #10601, opened 2021-12-28
Fix CI workflow runs using manually created virtual environments

m-vdb:
**Proposed changes**:
- Create virtual environments manually in all workflows that need one (mimic'ing what we do for unit tests)
- Update `change_filters.yml` to also listen to changes in workflow files
- Introduce PR labels to invalidate poetry cache if needed

### DIFF
--- a/.github/workflows/ci-docs-tests.yml
+++ b/.github/workflows/ci-docs-tests.yml
@@ -60,11 +60,24 @@ jobs:
         poetry-version: ${{ env.POETRY_VERSION }}
 
     - name: Load Poetry Cached Libraries ⬇
+      id: cache-poetry
       uses: actions/cache@v2
       with:
-        path: ~/.cache/pypoetry/virtualenvs
+        path: .venv
         key: ${{ runner.os }}-poetry-${{ env.POETRY_VERSION }}-${{ env.DEFAULT_PYTHON_VERSION }}-${{ hashFiles('**/poetry.lock') }}-${{ secrets.POETRY_CACHE_VERSION }}
         restore-keys: ${{ runner.os }}-poetry-${{ env.DEFAULT_PYTHON_VERSION }}
+
+    - name: Clear Poetry cache
+      if: steps.cache-poetry.outputs.cache-hit == 'true' && contains(github.event.pull_request.labels.*.name, 'tools:clear-poetry-cache-docs-tests')
+      run: rm -r .venv
+
+    - name: Create virtual environment
+      if: (steps.cache-poetry.outputs.cache-hit != 'true' || contains(github.event.pull_request.labels.*.name, 'tools:clear-poetry-cache-docs-tests'))
+      run: python -m venv create .venv
+
+    - name: Set up virtual environment
+      if: needs.changes.outputs.docs == 'true'
+      run: poetry config virtualenvs.in-project true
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-docs-tests.yml
```diff
@@ -60,12 +60,25 @@ jobs:
...
+      id: cache-poetry
       uses: actions/cache@v2
       with:
-        path: ~/.cache/pypoetry/virtualenvs
+        path: .venv
         key: ${{ runner.os }}-poetry-${{ env.POETRY_VERSION }}-${{ env.DEFAULT_PYTHON_VERSION }}-${{ hashFiles('**/poetry.lock') }}-${{ secrets.POETRY_CACHE_VERSION }}
         restore-keys: ${{ runner.os }}-poetry-${{ env.DEFAULT_PYTHON_VERSION }}
 
+    - name: Clear Poetry cache
+      if: steps.cache-poetry.outputs.cache-hit == 'true' && contains(github.event.pull_request.labels.*.name, 'tools:clear-poetry-cache-docs-tests')
```
m-vdb, line 71: do we document PR labels like `tools:clear-poetry-cache-docs-tests` somewhere? I haven't created the label yet on github (same for other new labels introduced in this PR)

.github/workflows/ci-docs-tests.yml
```diff
@@ -60,12 +60,25 @@ jobs:
...
+      run: rm -r .venv
+
+    - name: Create virtual environment
+      if: (steps.cache-poetry.outputs.cache-hit != 'true' || contains(github.event.pull_request.labels.*.name, 'tools:clear-poetry-cache-docs-tests'))
+      run: python -m venv create .venv
+
+    - name: Set up virtual environment
+      if: needs.changes.outputs.docs == 'true'
+      run: poetry config virtualenvs.in-project true
```
tayfun, line 80: Is it correct to set this `in-project` option to True? As far as I can see, [poetry will create a new virtualenv](https://python-poetry.org/docs/configuration/#virtualenvsin-project) itself with this option and it will be overwritten by `python -v venv create .venv` later on I'm guessing? If not set -which is the default- it will use existing venv if available (which will exist after manual creation and if it had not existed I think poetry would be creating a new venv by default and this is handled by `virtualenvs.create` option).

m-vdb: I think this is bypassing the creation of the venv in poetry somehow, because this is what failed before
> If set to true, the virtualenv wil be created and expected in a folder named .venv within the root directory of the project.

In the case the `.venv` already exists (which we now enforce), I'm guessing poetry doesn't do anything (`expected in a folder named .venv`)

.github/workflows/ci-docs-tests.yml
virtualroot: This [notion page](https://www.notion.so/rasa/PR-deployments-f95cc7feee70453388e7d306dc75ee79) is the closest I can find

virtualroot (approved):

tayfun (approved):

.github/workflows/ci-docs-tests.yml
m-vdb: 🤔 I guess I'll just create the labels in GitHub for now, and describe them here

PR #10608, opened 2021-12-30
Regression test Profiler

markus-hinsche:
**Proposed changes**:
- Enable profiler (part of Datadog's APM)
- Use `service datadog-agent restart` (instead `start-stop-daemon` introduced in [here](https://github.com/RasaHQ/rasa/pull/10380/commits/db4c52a26972ecd921bf400484683fbcc285e693) which didn't work for starting the APM agent)
- Provide flag for enabling (**not** enable ddtrace and profiling by default)
- Add more scripting to restart APM agent


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

.github/workflows/ci-model-regression.yml
```diff
@@ -17,6 +17,7 @@ concurrency:
 env:
   GKE_ZONE: us-central1
   GCLOUD_VERSION: "318.0.0"
+  DD_PROFILING_ENABLED: true
```
tczekajlo, line 20: The profiler will be enabled for all model regressions workflows, whenever someone runs the workflow. I'd prefer to disable it by default and give an option to explicitly use it e.g. by adding an extra label in a PR.

.github/scripts/start_dd_agent.sh
```diff
@@ -49,12 +49,4 @@ fi
...
-
-INSTALL_DIR="/opt/datadog-agent"
-AGENTPATH="$INSTALL_DIR/bin/agent/agent"
-PIDFILE="$INSTALL_DIR/run/agent.pid"
-AGENT_ARGS="run -p $PIDFILE"
-AGENT_USER="dd-agent"
-LD_LIBRARY_PATH="/usr/local/cuda/extras/CUPTI/lib64:/usr/local/cuda/lib64:/usr/local/nvidia/lib:/usr/local/nvidia/lib64"
-
-sudo -E start-stop-daemon --start --background --quiet --chuid $AGENT_USER --pidfile $PIDFILE --user $AGENT_USER --startas /bin/bash -- -c "LD_LIBRARY_PATH=$LD_LIBRARY_PATH $AGENTPATH $AGENT_ARGS"
+sudo service datadog-agent restart
```
tczekajlo, line 52: `start-stop-daemon` was used directly to pass the `LD_LIBRARY_PATH` to the agent process, otherwise, GPU metrics won't be collected.

tczekajlo (changes requested):

.github/workflows/ci-model-regression.yml
markus-hinsche: Agreed, as discussed, always running the profiler is wasteful, because the profiler cost additional money.
Also, a developer would know beforehand if they want to profile, in which case he/she can enable it.

Therefore, I will set this to false for now, and create a follow-up PR to introduce the usability feature of enabling via a PR-label.

markus-hinsche:
@tczekajlo I updated the `start-stop-daemon` logic so it also restarts APM agent. 
I re-ran regression tests on both GPU and CPU. 
Lastly, I disable profiling so it is turned off by default now.

tczekajlo (approved):
LGTM

Issue #10203, opened 2021-11-15
Make logging verbosity configurable per package for verbose packages.


melindaloubser1:
### What problem are you trying to solve?

Logs from pika (rabbitmq) and kafka libraries can be very verbose, cluttering debug logs.
Usually turning on `--debug` mode is done to debug model prediction.

### What's your suggested solution?

Make logging configurable for the more verbose libraries separately from overall log level.
These could be env vars e.g. `PIKA_DEBUG_LEVEL`, `KAFKA_DEBUG_LEVEL`. 

Some investigation required to see which other libraries are most verbose; these are the most obvious ones.

### Is anything blocking this from being implemented? (if relevant)

None known

### Definition of Done

- [ ] logging made configurable for specific overly verbose packages [Kafka and Pika for starters]

b-quachtran:
Would be great to get an option for this in Rasa X as well

tayfun:
I've been working on this ticket. There are a few problems that I can see:

1. First is that we already have LOG_LEVEL_LIBRARIES which is not documented, and not used for matplotlib correctly (eg. loaded modules are printed by matplotlib because the parent logger level is not set with LOG_LEVEL_LIBRARIES). I can fix that easily and also add a specific setting just for matplotlib/kafka/pika etc.

2. Current default log level is INFO. So when we run rasa shell we see INFO logs (and other higher level logs ie. WARNING, ERROR, CRITICAL). Docs say default is None which is actually correct but None for verbosity actually means INFO in practice :) This also means that -v doesn’t do anything as it should be setting log level to INFO and this is default as I’ve said. What do we want default log level to be? I feel INFO level is what everyone really uses so that we can see what’s going on, so this should stay? I can update docs to say INFO is default but then -v doesn’t do anything and we’d be missing ERROR  (though we do have --quiet for WARNING which is good enough I think - it’s also normally the default for Python programs as in root logger comes with WARNING level).

Note we had already worked on a similar problem with https://github.com/rasahq/rasa/issues/6900

I'm planning on adding specific log levels for 3 of the libraries (rabbitmq, matplotlib and kafka). I'm renaming pika to rabbitmq since it is really about connecting to RabbitMQ and also because the loggers to disable are aio_pika and aiormq.

tayfun:
[...]

PR #10582, opened 2021-12-23
add TypeError fix and testing

ePreda:
Resolves #10567.  Simply adds a default empty string when requesting an environmental var to avoid passing in a NoneType.

Added some, perhaps over-engineered, testing to catch future encoding errors. 

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #10614, opened 2021-12-31
Configurable logging for libraries

tayfun:
**Proposed changes**:

- Fixes https://github.com/RasaHQ/rasa/issues/10203

- Add environment variables to configure logging at library level. This is enabled for matplotlib, rabbitmq (pika etc.) and kafka at the moment.


rasa/core/brokers/kafka.py
```diff
@@ -216,4 +212,4 @@ def _close(self) -> None:
     @rasa.shared.utils.common.lazy_property
     def rasa_environment(self) -> Optional[Text]:
         """Get value of the `RASA_ENVIRONMENT` environment variable."""
-        return os.environ.get("RASA_ENVIRONMENT")
+        return os.environ.get("RASA_ENVIRONMENT", "RASA_ENVIRONMENT_NOT_SET")
```
tayfun, line 215: @melindaloubser1 I've added this default because otherwise kafka publish fails with
```
2021-12-31 16:26:18 ERROR    rasa.core.brokers.kafka  - Could not publish message to kafka url 'localhost'. Failed with error: encoding without a string argument
```

I think we also need to document this?

tayfun:
Does this solution look good? I'll add tests and docs once we agree on a path to take.

rasa/core/brokers/kafka.py
joejuzl: noticed a similar change [here](https://github.com/RasaHQ/rasa/pull/10582/files)

joejuzl (approved):
Nice 👍

docs/docs/command-line-interface.mdx
```diff
@@ -6,6 +6,31 @@ description: Command line interface for open source chatbot framework Rasa. Lear
...
+
+In addition to CLI arguments, several environment variables allow you to control log output in a more granular way. With these environment variables, you can configure log levels for messages created by external libraries such as matplotlib, pika, and kafka. Currently following environment variables are supported:
+
+1. LOG_LEVEL_LIBRARIES: This is the general environment variable to configure log level for all libraries.
+2. LOG_LEVEL_MATPLOTLIB: This is the specialized environment variable to configure log level only for Matplotlib.
+3. LOG_LEVEL_RABBITMQ: This is the specialized environment variable to configure log level only for AMQP libraries, at the moment it handles log levels from `aio_pika` and `aiormq`.
+4. LOG_LEVEL_KAFKA: This is the specialized environment variable to configure log level only for kafka.
+
+General configuration (`LOG_LEVEL_LIBRARIES`) has less priority than library level specific configuration (`LOG_LEVEL_MATPLOTLIB`, `LOG_LEVEL_RABBITMQ` etc); and CLI parameter has the highest precedence. This means variables can be used together with a predictable result. As an example:
```
joejuzl, line 20: Does CLI parameter have the highest precedence really? In the example below the cli is set to `DEBUG` but does't override the `WARNING` level of matplotlib... or am I understanding the example incorrectly?
Or maybe it only overrides to a _less_ verbose level?

tayfun: Good catch, CLI parameter sets the level at the root level (which also has the important handler - coloredlogs handler). This means CLI level will be used for every log which is not explicitly set (matplotlib, rabbitmq, or rasa and any other library) but it also means CLI level overrides all other log levels if it is a less verbose level, like you said.

I'll change the wording in the docs.

rasa/core/brokers/kafka.py
melindaloubser1: Sorry I was not online so just seeing this now - is it resolved/documented already?

tayfun: @melindaloubser1 I've added the default `RASA_ENVIRONMENT_NOT_SET` but I haven't documented it as I'm not sure what the use case for this is. Seeing you added this you may be better suited to document it? We can also create a ticket for someone else. What do you think?

melindaloubser1: Ohh wait a sec I remember this now! The default here should be `production`. This is about multiple deployment environments.

melindaloubser1: I thought the `RASA_ENVIRONMENT` already defaulted to `production` though 🤔? Under which circumstances did you see the error above?

tayfun: Oh I see, is there a document around this? I just got above error while running it locally so no env variables set by me.

melindaloubser1: Context-document is https://rasa.com/docs/rasa-x/enterprise/deployment-environments, but nothing for OSS because the parameter is meaningless without Rasa Enterprise attached. 
The default is set up in Rasa Enterprise, but not Rasa OSS without Enterprise attached, so I was wrong. 
In https://github.com/RasaHQ/rasa/pull/10413, I should have accounted for Kafka running without Enterprise setting the `RASA_ENVIRONMENT` var. I'm wondering if it's more reasonable to disable adding the header if the env var is not set.

melindaloubser1: I'll open a PR to fix it and tag you

melindaloubser1: https://github.com/RasaHQ/rasa/pull/10634

melindaloubser1: and a backport since the original was also backported: https://github.com/RasaHQ/rasa/pull/10635

tayfun: @melindaloubser1 I think there's a typo in  https://rasa.com/docs/rasa-x/enterprise/deployment-environments where environments are listed after `By default, two environments are defined:`. Worker should be `development`. 

I've approved your PRs.

melindaloubser1: The image shows `development` in addition to `worker`, but only `worker` is default, so no typo as far as I can see.

tayfun: I mean there are two environments, production and development but the list shows `worker` as an environment:
```
By default, two environments are defined:

production
worker
```

melindaloubser1: It's referring to the right hand side box:
image

`development` is not a default environment. It must be added manually.

tayfun: @melindaloubser1 Hmm, I see. I think I'm just confused then :D I thought prod and dev are default environments whereas worker would be separate instance just for training. Thanks for the explanation.

PR #10618, opened 2022-01-03
Fix AugmentedMemoizationPolicy max history bug. (3.0.x)

kedz:
**Proposed changes**:

Fixes three bugs with `AugmentedMemoizationPolicy`:

1. When more than max_history actions occurred in a row, information about the last user uttered event was incorrectly being discarded.
2. When a training story started with a user utterance AugmentedMemoizationPolicy would not be able to match it at test time if there were previous events before the initial user utterance event.
3. In the tests of `AugmentedMemoizationPolicy` in 3.0.x, the policy was not being constructed with the correct `max_history`, though its featurizer was, causing the policy not to truncate the tracker according to `max_history` actions in the unit tests.

For more discussion of the first two points see [this PR](https://github.com/RasaHQ/rasa/pull/10605) for the 2.8.x branch.


kedz:
/modeltest

```yml
include:
 - dataset: ["all-core"]
   config: ["all-core"]

```

kedz:
/modeltest

```yaml
include:
  - dataset: ["all-core"]
    config: ["all-core"]
```

tests/core/test_policies.py
```diff
@@ -663,6 +664,89 @@ def test_augmented_prediction(
             model_storage=model_storage,
             resource=resource,
             execution_context=execution_context,
+            config={"max_history": max_history},
```
samsucik, line 667: Here and elsewhere we could use the `POLICY_MAX_HISTORY` constant instead of the `max_history` string.

samsucik (commented):
I left only one minor comment but my comments from the 2.8 review apply here as well.

JEM-Mosig (approved):
Looks good to me :)

PR #10624, opened 2022-01-05
Use HF Auto classes in LanguageModelFeaturizer

mleimeister:
In order to allow users to use arbitrary HuggingFace models in `LanguageModelFeaturizer`, this PR gets rid of the hard-coded mapping between model architecture and model/tokenizer classes. Instead the model classes are inferred from the specified weights using `AutoTokenizer` and `TFAutoModel`. The implementation aims to provide the following:

- Enable all HF models that are compatible with our current version of `transformers`
- Keep default weights for the currently available models (for backward compatibility)
- Reproduce output for existing model/weights combinations in the unit tests

TODO:
- [x] One part that so far was tricky to move to the Auto classes is the delimiter token cleaning (e.g. `##` in BERT). The current version has a fixed list of known delimiter tokens (from BERT, GPT2, XLNET). Should investigate how the `Tokenizer.convert_tokens_to_string` function can be used that implements this cleaning step in the child classes. -> The 3 existing delimiter tokens are the ones currently listed in the HF documentation/tokenizer course.
- [x] Validate role of different special tokens in HF tokenizers (especially, should only `CLS`, `BOS`, `EOS` be filtered?) -> Testing various tokenizers, removing the `UNK` token seems the correct way. Nothing would prevent though someone writing a custom tokenizer that breaks this. The latest HF version contains a base class function [SpecialTokensMixin.get_special_tokens_mask](https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_utils_base.py#L3277) that however includes the `UNK` token in the mask and is therefore not useful for our purpose.
- [x] Remove `nlu.utils.huggingface` subdirectory
- [x] Add unit tests for "unknown" model architectures and missing default weights
- [x] Run model regression tests
- [x] Update docs

**Proposed changes**:
- Use the HuggingFace Auto classes to load arbitrary weights from the HF model hub

- [ ] Check on HF issue board if exposing the delimiter prefixes could be implemented without too much work
- [x] Go through existing tokenizers and check if `convert_tokens_to_string` behaves as expected with our cleanup function
- [ ] Unit tests for tokenizers that differ considerably to the currently supported


mleimeister:
/modeltest

```
include:
 - dataset: ["Sara"]
   config: ["Sparse + BERT + DIET(bow) + ResponseSelector(bow)", "BERT + DIET(seq) + ResponseSelector(t2t)"]
```

changelog/6307.feature.md
```diff
@@ -0,0 +1,4 @@
+Load HuggingFace models in `LanguageModelFeaturizer` using `AutoTokenizer` and `TFAutoModel` classes. 
+This enables to use arbitrary models from the [Huggingface model hub](https://huggingface.co/models) instead
```
koernerfelicia, line 2:
```suggestion
This enables the use of arbitrary models from the [Huggingface model hub](https://huggingface.co/models) instead 
```

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -28,14 +31,22 @@
...
+    "xlnet": "xlnet-base-cased",
+    "distilbert": "distilbert-base-uncased",
+    "roberta": "roberta-base",
 }
+CLS_TOKEN = "[CLS]"
+
+# The following are the prefix characters added to sub-tokens by the currently
+# used WordPiece, SentencePiece and Byte Pair tokenizers. See
+# https://huggingface.co/course/chapter6 for a description of each.
+# In case a new tokenizer uses different characters to split sub-tokens, this
```
koernerfelicia, line 47: For clarification -- does this cover all the existing tokenizers? Or just the ones corresponding to the list we had above? I am not sure if our version of `tokenizers` (fairly old) has an additional tokenizer we need to worry about
Screenshot 2022-01-19 at 10 38 13

We should clarify "currently used" to make sure it's clear for users (currently used by us? currently existing?)

Is there a way to programmatically check the prefix characters? Maybe it is an attribute of the tokenizer, or something? I am not sure there is (seems like the `BaseTokenizer` class does not have an attribute), and SentencePiece calls these `replacement`, whereas WordPiece calls this `wordpieces_prefix`.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -215,33 +220,54 @@ def _lm_specific_token_cleanup(
 
         Returns: Cleaned up token ids and token strings.
         """
-        from rasa.nlu.utils.hugging_face.registry import model_tokens_cleaners
+        token_ids_string = [
+            (id_, re.sub(TOKENIZER_PREFIX_CHARS, "", token))
```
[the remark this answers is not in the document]
koernerfelicia: Actually, I wonder if we can just do:
```
split_token_strings = self.tokenizer.convert_ids_to_tokens(split_token_ids, skip_special_tokens=True)
```
in `_lm_tokenize`

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
koernerfelicia: Ah, I think maybe I was looking for `self.all_special_tokens` [here](https://huggingface.co/docs/transformers/v4.15.0/en/main_classes/tokenizer#transformers.PreTrainedTokenizer)?

Old docs won't load

koernerfelicia (commented):
This looks like it will be a very nice simplification! I think that we can maybe try to make the prefix character part safer. Would be nice to cover the case where there is some other token used (not sure how likely this is, maybe not), but also make the prefix removal less aggressive, in case people want to use one architecture, and not treat the prefixes of other architectures as special.

Either we may be able to use some attribute of PretrainedTokenizer:
`PretrainedTokenizer.all_special_tokens`

or maybe we can even just prevent them from being added to the str in the first place in `_lm_tokenize`.

Left comments to that effect.

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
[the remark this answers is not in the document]
mleimeister: I think this will only remove the special tokens (`[CLS], [UNK], etc.`), not the prefixes from the split sub tokens. To test I tried the following:

```
tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
text = "awesome word embeddings"
token_ids = tokenizer.encode(text, add_special_tokens=True)
print(token_ids)
print(tokenizer.convert_ids_to_tokens(token_ids, skip_special_tokens=False))
print(tokenizer.convert_ids_to_tokens(token_ids, skip_special_tokens=True))
```

resulting in

```
[101, 12476, 2773, 7861, 8270, 4667, 2015, 102]
['[CLS]', 'awesome', 'word', 'em', '##bed', '##ding', '##s', '[SEP]']
['awesome', 'word', 'em', '##bed', '##ding', '##s']
```

still leaving the `##` in the output strings.

mleimeister:
Thanks a lot for the feedback 👍 ! I looked a bit more into the tokenizer classes. The problem is I think that the notion of special tokens applies to the reserved tokens like `[CLS], [UNK], [PAD]`, etc., but not the prefixes that are added by the tokenizers.

I did not find a coherent way to extract those from the various implementations of the `PretrainedTokenizer` class, since some do set them as input parameter (as the `wordpieces_prefix` in case of `BertTokenizerFast`), while for others like `BertTokenizer` it is hard-coded in the splitting function itself (e.g. [here](https://github.com/huggingface/transformers/blob/master/src/transformers/models/bert/tokenization_bert.py#L539)).

One function that looks promising though is `PretrainedTokenizer.convert_tokens_to_string`. This forms a string from a list of tokens and re-uses the prefixes in the various implementations. Apart from the BERT tokenizers, this seems to give what we want. For BERT however, the implementation [here](https://github.com/huggingface/transformers/blob/master/src/transformers/models/bert/tokenization_bert.py#L243) assumes a list of tokens, whereas we input only single tokens. I'll shortly commit an attempt using this inside a helper function, which would then get rid of the hard coded list and allow to use the prefixes of other architectures.

docs/docs/components.mdx
```diff
@@ -614,6 +595,28 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
       cache_dir: null
   ```
 
+  Apart from the default pretrained model weights, further models can be used from the
+  [HuggingFace model hub](https://huggingface.co/models) provided that the model has pretrained Tensorflow weights
+  (make sure that the file `tf_model.h5` exists in the "Files and Versions" section). To use such a model, it is enough
+  to specify the `model_weights` parameter as the HuffingFace model identifier. For example, the model found at
```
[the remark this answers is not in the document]
mleimeister: I've made that typo sooo often by now 🤣

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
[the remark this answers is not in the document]
koernerfelicia: That's a shame... Thank you for looking into it!

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -212,33 +209,72 @@ def _lm_specific_token_cleanup(
...
+            sub-token and the builtin `convert_tokens_to_string` only works for a
+            list of tokens, while we apply it to an individual token. For other
+            tokenizers the prefixes signify whitespace or end-of-words.
+            """
+            if issubclass(
+                type(self.tokenizer), transformers.tokenization_bert.BertTokenizer
+            ) or issubclass(
+                type(self.tokenizer), transformers.tokenization_bert.BertTokenizerFast
+            ):
+                return token.replace("##", "")
```
koernerfelicia, line 226: I think we should maybe make this a constant. I know we only need it here in this specific case, but I think it'll help document... really hated that HF doesn't have some sort of constant I could search for :)

koernerfelicia:
@mleimeister thank you for looking into this more! I am definitely happier with this new version. A shame that we can't seem to get these tokens programmatically.. makes my skin crawl a bit. I wonder if it is worth trying a PR with HF to make those special tokens not a hard-coded string? That's just a whole other rabbit hole, so it would be up to you whether you want to pursue that.

> we'd have to monitor if any new upcoming models break that logic

Do you have any estimate on how likely this would be? Is there anyway to protect against that? I haven't looked into the base classes and their implementations in more depth, so forgive me if this is naive, but can we enumerate the base classes that we know of/have tested, and do a check when `LMFeaturizer` is instantiated? Something like the if block we do to check if it's a BERT style tokenizer:
```
if issubclass(type(self.tokenizer), transformers.tokenization_bert.BertTokenizer)
```

Then we can output a warning or something if someone tries to use a tokenizer we don't know?

Another thing I haven't looked into myself, so don't have the context for: does this enable significantly different models from the ones we've tested for? Is it worth adding tests for these to the existing test cases?

mleimeister:
> I wonder if it is worth trying a PR with HF to make those special tokens not a hard-coded string? That's just a whole other rabbit hole, so it would be up to you whether you want to pursue that.

I would assume this is a lot of work, since the tokens are not assigned coherently throughout the different tokenizers (even in between the standard and fast versions). I could raise it on the HF repo to see if they encountered this before and if there might be a good way to solve it.

I just pushed an updated version, where there is no more distinction between BERT/non-BERT tokenizers and `convert_tokens_to_string` is used coherently - at the expense of having to introduce an artificial empty start token (otherwise the `##` wouldn't get filtered since it can only appear at sub-tokens *within* a token, while we pass single sub-tokens to the function). I tried explaining this in this [Note](https://github.com/RasaHQ/rasa/blob/lmfeaturizer-automodel/rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py#L211). This might be the best that can be achieved with this version of HF, until the special tokens are exposed programatically. What do you think about this solution?

> but can we enumerate the base classes that we know of/have tested, and do a check when LMFeaturizer is instantiated? 
Then we can output a warning or something if someone tries to use a tokenizer we don't know?

I think this would be an option. What I could definitely do is go through the 30ish current tokenizer implementations and verify that the `convert_tokens_to_string` function is indeed returning the expected original sub-tokens the way we need them for further processing. Then it would become a recurring task when switching to a new HF version we'd need to watch out for.

> Another thing I haven't looked into myself, so don't have the context for: does this enable significantly different models from the ones we've tested for? Is it worth adding tests for these to the existing test cases?

That's a good point. From all I saw so far, the relevant tokenizers are all based on the same base models (WordPiece, SentencePiece, BytePair). I would double-check that when going through them though and can add tests accordingly. What's not covered however would be transformer models for e.g. multi-modal data, such as `LayoutLM`, `VisionTransformer`, etc. So hopefully nobody would expect those to work within Rasa? 😅 

I'll add open tasks for the points mentioned above.

koernerfelicia:
Awesome, thank you for this, I think we are getting closer! Also, just a heads up that [this PR](https://github.com/RasaHQ/rasa/pull/10459) looks close to merge, and will update the transformers version. Hopefully that is not too annoying 😬 

> I would assume this is a lot of work, since the tokens are not assigned coherently throughout the different tokenizers

Yep, that's fair, I think you're right there.

> I just pushed an updated version, where there is no more distinction between BERT/non-BERT tokenizers [...] What do you think about this solution?

That looks really great to me. I don't think I see a drawback or risk to that empty token, since this is after we've already gotten the features.

> Then it would become a recurring task when switching to a new HF version we'd need to watch out for.

This one concerns me a little bit because I guess it would be a silent failure if the tokens don't come out right. But maybe it is unlikely to change as this would be more like a bug or breaking change from my understanding. Can we do a unit test that runs through the tokenizers in `TOKENIZER_MAPPING` and does a sanity check that the `string in == concatenated tokens out`? I guess it is hard to think of what kind of a string would break the method, since that would probably be dependent on the nature of the tokenizer... If I am being paranoid here and it seems mostly safe from your POV let me know :)

> So hopefully nobody would expect those to work within Rasa? 😅

I think we can just add some notes in the docs to that effect 😄

koernerfelicia:
> just a heads up that this PR looks close to merge, and will update the transformers version. Hopefully that is not too annoying 😬

In particular, note [this](https://github.com/RasaHQ/rasa/pull/10459#discussion_r794386918), which may complicate things here.

mleimeister:
Hi @koernerfelicia, thanks a lot for the review and comments! 

> Can we do a unit test that runs through the tokenizers in TOKENIZER_MAPPING and does a sanity check that the string in == concatenated tokens out? I guess it is hard to think of what kind of a string would break the method, since that would probably be dependent on the nature of the tokenizer... If I am being paranoid here and it seems mostly safe from your POV let me know :)

I thinks that's very fair and we should ensure that any potential failure is transparent, since the errors would likely surface in an obscure way by suddenly having those delimiter chars turn up in e.g. extracted entities. I planned to run such a test to check the currently available tokenizers, but it could as well be a permanent unit test.

I started with this for all subclasses and directly ran into a [known issue](https://github.com/huggingface/transformers/issues/5232) for `BertTokenizerFast` that would block this current solution. It is fixed however in `transformers` v4 and should be gone with the upcoming update of TF and `transformers`. Hence, I would for now definitely wait with this PR and adapt it to the changed versions as soon as that is merged.

> In particular, note this, which may complicate things here.

`sentencepiece` is indeed used for quite a few tokenizers. As far as I understand the [changes](https://github.com/RasaHQ/rasa/pull/10459#discussion_r794356723), we will install it separately though, right? So the same models should still be available. Another change with `transformers` v4 is apparently that `AutoTokenizer` will [load the fast Rust implementation](https://huggingface.co/docs/transformers/migration#1-autotokenizers-and-pipelines-now-use-fast-rust-tokenizers-by-default) from the `tokenizers` library, so the `sentencepiece` dependency would not be strictly necessary, but instead we might need some careful tests for other breaking changes.

koernerfelicia:
>Hence, I would for now definitely wait with this PR and adapt it to the changed versions as soon as they're merged.

Sounds good! 👍 

>As far as I understand the changes, we will install it separately though, right? So the same models should still be available.

Ah yes, sorry I wasn't very specific. I think sentencepiece is fine, what I was unsure about is whether there are other dependencies like it, since Matthew had said "the dependencie**s** for **some** models are not installed by default". I thought we might run into a case where some dependencies other than sentencepiece aren't needed for our fixed set of models, but may be needed if/when we allow for any model with this PR.

mleimeister:
Hi @koernerfelicia, I adapted the PR to the updated version of HF on `main` and there's now a [unit test](https://github.com/RasaHQ/rasa/blob/f92c3fb4157ebe0c0cd588cd2dab5c7e873290da/tests/nlu/featurizers/test_lm_featurizer.py#L823) to check the correct clean up of the delimiter tokens for the tokenizers we supported so far as well as some other frequently requested models from forum posts.

I ran an informal check on all currently available HF tokenizers, by using pretrained models that were either specified in the tokenizer's `pretrained_init_config` variable or - if not set there - by searching for a model on the HuggingFace hub programatically. This showed that out of 68 tokenizers, for 47 the delimiter cleanup works as expected, 3 failed to reproduce the input word, and the others failed due to unrelated exceptions (mostly no suitable model being found, but also different input formats expected, etc.). The notebook output is attached below. Caveat is that this is for one particular input, that gets split by all tokenizers though, and not testing non-English models. While this is not perfect, it suggests a relatively good coverage of the available models. What would you suggest going forward?

[test_transformers_tokenizer_cleanup.pdf](https://github.com/RasaHQ/rasa/files/8033015/test_transformers_tokenizer_cleanup.pdf)

docs/docs/components.mdx
```diff
@@ -615,6 +615,13 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
   end-to-end before trying to optimize this component with other weights/architectures.
   :::
 
+  :::note
+  While we make an effort to support as many HuggingFace models as possible, please note that some architectures will
+  not work due to how Rasa's internal tokenization interacts with the HuggingFace tokenizers. If you encounter an issue
+  with a specific model other than the ones listed in the table above, please let us know on
```
koernerfelicia, line 621: Maybe we can sketch to them how to check whether a tokenizer will work?

koernerfelicia: Could even link to the unit test. I'd still hedge this, meaning I wouldn't promise that if they get good results on one test it will work perfectly

koernerfelicia: SImilarly, I think it would help if we listed known problematic models:
CRTL, BERTweet, PhoBERT?

koernerfelicia:
Hi @mleimeister thanks for testing all those models! I feel pretty good about this now, though I think we can be even more transparent in the docs: basically explain to them how they might be able to figure out whether a tokenizer is problematic, and describe what might happen. Mainly because I think the failure might not be obvious to users: they might expect a crash, and think they are good if everything runs. But it could also happen that the tokens are not cleaned up and I *think* they likely wouldn't notice.

I am also not so confident people will go looking in the docs. Maybe we could issue a user warning if they initialise a tokenizer not covered by the unit tests or in our defaults and link to them to the relevant part of the docs?

mleimeister:
@TyDunn Since this PR is close to being finished and contains a feature update (specifically, expanding the functionality of the current `LanguageModelFeaturizer`), I wanted to check with you regarding where this should go? Would this be appropriate for the upcoming 3.1 minor release? The changes would be as follows:

- More models from HuggingFace are supported, by using the model identifier from the HF model hub as the `model_weights` parameter. This enables most of the  models we've seen in the forum that people wanted to use and couldn't so far.
- The models that are not supported as of `transformers` version 4.13.0 are listed in the documentation. There is also an explicit check when the component loads and an error is being raised in case of an incompatible model, pointing to the documentation for further information.
- I will discuss with QA about how to best run a regular check to ensure that the models we do support are up to date.

Let me know if you have any questions or we should discuss details in a short meeting.

TyDunn:
@mleimeister Since this contains enhancement, then it should go into the next minor release (3.1)

mleimeister:
Hi @koernerfelicia, the latest commits now contain the changes we discussed. Particularly:

- Incompatible models are listed in the [docs](https://10624--rasahq-docs-rasa-v2.netlify.app/docs/rasa/components#languagemodelfeaturizer) and are [checked](https://github.com/RasaHQ/rasa/blob/lmfeaturizer-automodel/rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py#L47) [in the component](https://github.com/RasaHQ/rasa/blob/lmfeaturizer-automodel/rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py#L187)
- A [unit test](https://github.com/RasaHQ/rasa/blob/lmfeaturizer-automodel/tests/nlu/featurizers/test_lm_featurizer.py#L826) checks the token cleanup for the currently supported model architectures and that incompatible models [raise an error](https://github.com/RasaHQ/rasa/blob/lmfeaturizer-automodel/tests/nlu/featurizers/test_lm_featurizer.py#L857)
- A [script](https://github.com/RasaHQ/rasa/blob/lmfeaturizer-automodel/.github/scripts/validate_lmfeaturizer_models.py) is in place to test the token cleanup for all models that are supported from this current `transformers` version. I reached out to QA to discuss how to integrate this in e.g. a cronjob, or the pre-release QA process. The concrete implementation would be covered in this follow-up [ticket](https://github.com/RasaHQ/rasa/issues/10893).
- Product is informed and [confirmed](https://github.com/RasaHQ/rasa/pull/10624#issuecomment-1040330480) it's viable for the upcoming minor release 

Let me know if this looks ok to you 🙂

rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py
```diff
@@ -149,6 +182,16 @@ def _load_model_metadata(self) -> None:
         self.model_weights = get_model_weights(self._config)
         self.cache_dir = self._config["cache_dir"]
         model_config = AutoConfig.from_pretrained(self.model_weights)
+
+        model_name = type(model_config).__name__.replace("Config", "")
+        if model_name in INCOMPATIBLE_MODELS:
+            raise ValueError(
+                f"You tried using a {model_name} model, which is not "
+                f"compatible with `LanguageModelFeaturizer`. Please "
+                f"consult the documentation on which models are "
```
koernerfelicia, line 191: Can we link to the page in the docs here?

docs/docs/components.mdx
```diff
@@ -614,6 +595,45 @@ Note: The `feature-dimension` for sequence and sentence features does not have t
       cache_dir: null
   ```
 
+  Apart from the default pretrained model weights, further models can be used from the
+  [HuggingFace model hub](https://huggingface.co/models) provided that the model has pretrained Tensorflow weights
+  (make sure that the file `tf_model.h5` exists in the "Files and Versions" section). To use such a model, it is enough
+  to specify the `model_weights` parameter as the HuggingFace model identifier. For example, the model found at
+  [https://huggingface.co/allenai/longformer-base-4096](https://huggingface.co/allenai/longformer-base-4096)
+  can be loaded as
```
koernerfelicia, line 603:
```suggestion
  can be loaded as:
```

(to mirror the way we do this in the rest of the page)

koernerfelicia (changes requested):
Some small comments. Also it occurred to me that I think we should have someone else review the docs part (dunno if CSE still does this). I'm not sure anymore if the concept of weights and models and how they relate to each other is confusing since we were not always good about consistent terminology here. Maybe we can have fresh eyes on just that docs section?

.github/scripts/validate_lmfeaturizer_models.py
```diff
@@ -0,0 +1,102 @@
...
+from typing import Text
+
+from rasa.engine.graph import ExecutionContext, GraphSchema
+from rasa.nlu.featurizers.dense_featurizer.lm_featurizer import LanguageModelFeaturizer
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+
+# Those model weights cover all supported model architectures as of transformers 4.13.0
+model_weights_list = [
```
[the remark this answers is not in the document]
koernerfelicia: (Could also deal with that question in the other PR, I guess it is just a note that we have to do something like this thing?

```
def get_subclasses(class_):
subclasses = []
for subclass in class_.__subclasses__():
subclasses.append(subclass)
subclasses.extend(get_subclasses(subclass))
return subclasses

subclasses = get_subclasses(transformers.tokenization_utils.PreTrainedTokenizer)
```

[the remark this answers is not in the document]
koernerfelicia: Okay! I am happy with that

koaning:
There are some thoughts that jumped in my mind. 

1. Can we check some non-Latin alphabets here? It’d be a shame if we work for English, but break for Hebrew, Korean, Chinese, or Arabic. This is probably best served as a “slow test” that we run once a week or so. Running that on every PR would be horrendously slow.
2. It might prove beneficial if the documentation has a small guide on which language models one might want to try out first. It'd be a shame if community members try out all sorts of "heavy" embeddings while a light one might suffice.

mleimeister:
> This is probably best served as a “slow test” that we run once a week or so. Running that on every PR would be horrendously slow.

@koaning That would be great. I'm actually looking for a similar functionality to regularly check if the tokenizer cleanup works as expected. Do we already have a process in place for running such "slow tests"? I guess it could be a cron job (?) and created this follow-up ticket to discuss with QA how to best do this: https://github.com/RasaHQ/rasa/issues/10893

koaning:
It's not super formal, but we have a cronjob for the base Rasa install that might serve as a source of inspiration. That runs daily though, this might be better served as an optional/weekly job.

mleimeister:
> we have a cronjob for the base Rasa install that might serve as a source of inspiration

@koaning  Ah yes, I saw that. I expanded the follow-up [ticket](https://github.com/RasaHQ/rasa/issues/10893) to also contain the testing of non-latin script models. I would set up a meeting with QA to discuss next week how to best implement this, also as part of maybe a general strategy to run such "slow tests". Would you be happy to have this handled there in order to not expand this PR further?

Do you have any concerns regarding the docs in terms of being understandable for bot developers?

koaning:
@mleimeister yeah for sure the topic of "what are good slow tests to run from cron" is a larger topic that's a bit out of scope for us here.

koaning:
I guess the main thing in my mind when I read the docs is "this reads fine, but maybe it's time that we have a benchmarking guide". My main fear is that folks try out the Huggingface models but forget to check the computational overhead.

That's a whole 'nother effort though. Also related to "can we make the benchmarking dev experience better".

koernerfelicia:
@koaning maybe we could have a forum post announcing this change and link to your video where you go through why it's important to focus on other things like data quality over fancy, heavy embeddings?

koaning:
A blog post wouldn't hurt, but the docs might be a more appropriate place. If you're interested in doing a benchmark, will you sooner look for info on the blog or on the docs? 

This [algorithm whiteboard video might be appropriate to link](https://www.youtube.com/watch?v=GTClb8RQSGM0).

losterloh:
@dakshvar22 Do you think we should still follow up on this PR?

Issue #10473, opened 2021-12-06
Support other recipes - starting with graph 


tayfun:
### What problem are you trying to solve?

Currently we only have a single recipe type called `default.v1` which fits most cases but falls short of advanced configuration. For research teams and other users who want customised recipes, we want to add a feature to modularly load different types of recipes. 

This will be released with a minor version (merge to main branch), ie. released as a 3.x version.

### What's your suggested solution?

For the minimum viable product, we want to support graph schemas so research can specify graph components/nodes as they like. 

1. Move `TrainingType` into `engine/recipes/recipe.py`. This file has base recipe class and other classes common to other recipes.
2. Move everything else in the file `shared/importers/autoconfig.py` to default recipe at `engine/recipes/default_recipe.py`. Default recipe itself will know how to validate the recipe and correct it (dumping missing keys etc.) and every new recipe from now on too will hold that information - high cohesion, low coupling.
3. Change `RasaFileImporter.get_config` so that it instantiates correct recipe, ie.

```python
def get_model_config(self, cli_par, training_type=training_type, is_finetuning=is_finetuning):
  # TODO: Need to read yaml config as dict here to be able to get recipe name, ie. default, graph etc
  config = ... # read yaml file as dict
  recipe = Recipe.recipe_for_name(config.get("recipe"))
  model_config = recipe.graph_config_for_recipe(
    config, kwargs, training_type=training_type, is_finetuning=is_finetuning
  )
  return model_config
```

4. Change `model_training._train_graph` so it can get model config directly:

```python
model_config = file_importer.get_model_config(kwargs, training_type, is_finetuning)
```

5. Add new recipe option `graph.v1` by adding new module `engine.recipes.graph_recipe` and in it class `GraphV1Recipe`.  Add this class to the options `recipes`in `Recipe.recipe_for_name` method.

6. For the new `GraphV1Recipe` , don't worry about validation or dumping default values etc. as this is an experimental and advanced feature which will evolve over time. We can add validation in a later step; default policies or pipeline components are not needed because we want more control over the graph with this recipe. Mark as experimental so users know this feature can change.

7. `GraphV1Recipe` will read a single file which has graph schema defined and can be directly used (no need to transform into usable format like in the default recipe). File is simply:

```yaml
recipe: graph.v1

language: en

train_schema:
  nodes:
    train_RegexFeaturizer1:
      needs:
        training_data: run_WhitespaceTokenizer0
[...]

PR #10641, opened 2022-01-06
Add support for different recipes

tayfun:
Fixes https://github.com/RasaHQ/rasa/issues/10473

**Proposed changes**:
- Add support for different recipe files
- See the docs: https://10641--rasahq-docs-rasa-v2.netlify.app/docs/rasa/graph-recipe/


tayfun:
After removing autoconfigure module, following rename was done:
```
get_configuration => auto_configure
_auto_configure => complete_config
```

rasa/shared/importers/importer.py
```diff
@@ -56,6 +58,23 @@ def get_config(self) -> Dict:
...
+        """Retrieves model config object.
+
+        Used both by E2EImporter and NLUDataImporter
+        """
+        config = self.get_config()
+        recipe = Recipe.recipe_for_name(config.get("recipe"))
+        model_config = recipe.graph_config_for_recipe(
+            config, cli_par, training_type=training_type, is_finetuning=is_finetuning,
+        )
+        return model_config
```
tayfun, line 76: This is causing tests failing because shared is not supposed to import anything from outside (ie. recipes).

tests/shared/importers/test_importer.py
```diff
@@ -368,15 +367,6 @@ def test_nlu_data_domain_sync_responses(project: Text):
     assert "utter_rasa" in domain.responses.keys()
 
 
-def test_importer_with_invalid_model_config(tmp_path: Path):
-    invalid = {"version": "2.0", "policies": ["name"]}
-    config_file = tmp_path / "config.yml"
-    rasa.shared.utils.io.write_yaml(invalid, config_file)
-
-    with pytest.raises(YamlValidationException):
-        TrainingDataImporter.load_from_config(str(config_file))
```
tayfun, line 377: This test moved to `tests/graph_components/validators/test_default_recipe_validator.py` as auto-config is dependent on recipe and not importer.

tests/shared/importers/test_autoconfig.py
```diff
@@ -1,223 +0,0 @@
...
-
-
-@pytest.mark.parametrize("config_file", ("non_existent_config.yml", None))
-def test_get_configuration_missing_file(tmp_path: Path, config_file: Text):
-    if config_file:
-        config_file = tmp_path / "non_existent_config.yml"
-
-    config = autoconfig.get_configuration(str(config_file))
-
-    assert config == {}
```
tayfun, line 57: This test has been removed because importers will return empty dict without going through auto-config now on.

tayfun: All other tests in this file have been moved to `tests/engine/recipes/test_default_recipe.py` because auto-config is recipe business.

changelog/10473.feature.md
```diff
@@ -0,0 +1 @@
+Support graph recipes.
```
joejuzl, line 1: Might be worth having a slightly more expansive description here. Also we can link to the docs?

joejuzl, line 1: Might be worth having a slightly more expansive description here. Also we can link to the docs?

rasa/engine/recipes/config_files/graph_config.yml
```diff
@@ -0,0 +1,532 @@
+# The config recipe.
```
joejuzl, line 1: Why do we need this file? Is it only for testing?

data/graph_schemas/graph_config_predict_schema.yml
```diff
@@ -0,0 +1,238 @@
+nodes:
```
joejuzl, line 1: Is this the same as `default_config_predict_schema.yml`?

docs/docs/graph-recipe.mdx
```diff
@@ -0,0 +1,128 @@
+---
+id: graph-recipe
+sidebar_label: Graph Recipe
+title: Graph Recipe
+description: Learn about Graph Recipe for Rasa Open Source.
+abstract: Graph recipes provide a more fine tuned configuration for your executable graphs.
+---
+
+We now support graph recipes in addition to the default recipe. What graph recipes provide is more powerful control over how execution graph schemas are built.
```
joejuzl, line 9:
```suggestion
We now support graph recipes in addition to the default recipe. Graph recipes provide more granular control over how execution graph schemas are built.
```

docs/docs/graph-recipe.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+- Default recipe is named `default.v1` in the config file whereas graph recipes are named `graph.v1`.
+- Default recipes provide an easy to use recipe structure whereas graph recipes are more advanced and powerful.
+- Default recipes are very opinionated and provide various defaults whereas graph recipes are more explicit.
+- Default recipes can auto-configure themselves and dump the defaults used to the file if some sections in `config.yml` are missing, whereas graph recipes do none of this and assume what you see is what you get. There are no surprises with graph recipes.
+- Default recipe divides graph configuration into mainly two parts: `pipeline` and `policies`. These can also be described as NLU and core (dialogue management) parts. For graph recipe on the other hand, the separation is between training (ie. `train_schema`) and prediction (ie. `predict_schema`).
+
+:::tip Starting from scratch?
+
+If you don't know which recipe to choose, use the default recipe to bootstrap your project fast. If later you find that you need more powerful finetuning options, you can always change your recipe to be a graph recipe.
```
joejuzl, line 32: "finetuning" has a double meaning here 😁

rasa/engine/recipes/default_recipe.py
```diff
@@ -816,3 +836,254 @@ def _add_end_to_end_features_for_inference(
...
+        training_type: Optional[TrainingType] = TrainingType.BOTH,
+    ) -> Dict[Text, Any]:
+        """Determine configuration from a configuration file.
+
+        Keys that are provided and have a value in the file are kept. Keys that are not
+        provided are configured automatically.
+
+        Note that this needs to be called explicitly; ie. we cannot
+        auto-configure automatically from importers because importers are not
+        allowed to access stuff outside shared package.
```
joejuzl, line 853:
```suggestion
        allowed to access code outside of `rasa.shared`.
```

rasa/engine/recipes/default_recipe.py
```diff
@@ -816,3 +836,254 @@ def _add_end_to_end_features_for_inference(
...
         )
         return node_with_e2e_features
+
+    @staticmethod
+    def auto_configure(
+        config_file_path: Optional[Text],
+        config: Dict,
+        training_type: Optional[TrainingType] = TrainingType.BOTH,
+    ) -> Dict[Text, Any]:
+        """Determine configuration from a configuration file.
```
joejuzl, line 846: Maybe something more explicit about filling in gaps in a configuration file?

rasa/engine/recipes/graph_recipe.py
```diff
@@ -0,0 +1,52 @@
...
+        """Converts the default config to graphs (see interface for full docstring)."""
+        mark_as_experimental_feature("graph recipe")
+        if cli_parameters or is_finetuning:
+            raise_warning(
+                "Unlike the Default Recipe, Graph Recipe does not utilize CLI "
+                "parameters or finetuning and these configurations will be ignored. "
+                "Add configuration to the recipe itself if you want them to be used.",
+                docs=DOCS_URL_GRAPH_RECIPE,
+            )
+        # TODO: Add core_target and nlu_target to graph configuration options so we
```
joejuzl, line 39: hmm this is interesting - I hadn't considered this...
I reckon this is worth doing now.

rasa/engine/recipes/graph_recipe.py
```diff
@@ -0,0 +1,52 @@
...
+class GraphV1Recipe(Recipe):
+    """Recipe which converts the graph model config to train and predict graph."""
+
+    name = "graph.v1"
+
+    def graph_config_for_recipe(
+        self,
+        config: Dict,
+        cli_parameters: Dict[Text, Any],
+        training_type: TrainingType = TrainingType.BOTH,
```
joejuzl, line 27: hmm - how is this determined? Maybe it could be based on what targets are provided in the config?

rasa/model_training.py
```diff
@@ -199,10 +199,17 @@ def _train_graph(
 
     config = file_importer.get_config()
     recipe = Recipe.recipe_for_name(config.get("recipe"))
+    config = recipe.auto_configure(
```
joejuzl, line 202: ok - so we only autoconfigure when training? I think that makes sense. Does this change the behaviour at all though? i.e. were there times we were autoconfiguring before but not now?

rasa/shared/importers/importer.py
```diff
@@ -276,6 +275,13 @@ def get_nlu_data(self, language: Optional[Text] = "en") -> TrainingData:
             lambda merged, other: merged.merge(other), nlu_data, TrainingData()
         )
 
+    @rasa.shared.utils.common.cached_method
+    def get_config_file_for_auto_config(self) -> Optional[Text]:
+        """Returns config file path for auto-config only if there is a single one."""
+        if len(self._importers) != 1:
+            return None
```
joejuzl, line 282: Why don't we just always return the first?

rasa/shared/importers/rasa.py
```diff
@@ -40,11 +39,16 @@ def __init__(
...
+        self.config_file = config_file
 
     def get_config(self) -> Dict:
         """Retrieves model config (see parent class for full docstring)."""
-        return self.config
+        if not self.config_file or not os.path.exists(self.config_file):
+            logger.debug("No configuration file was provided to the RasaFileImporter.")
+            return {}
+
+        config = rasa.shared.utils.io.read_model_configuration(self.config_file)
```
joejuzl, line 50: Do we risk using the non-autoconfigured config sometimes?

joejuzl (commented):
🚀

rasa/shared/importers/importer.py
tayfun: We return None because we don't want to auto-fill any of the files when there are multiple? Does that make sense? I think auto-fill is valid only for a single file. Multi file auto-fill seems to me more complex and better to leave alone.

rasa/shared/importers/rasa.py
tayfun: Recipes do auto-configure work now, so importers always read raw files. It is recipe's responsibility to auto-fill now (and dump the file after auto-fill), so at this point it is just raw config file read, and we can get what recipe needs to be extracted afterwards.

docs/docs/custom-graph-components.mdx
```diff
@@ -328,16 +328,16 @@ class MyComponent(GraphComponent):
...
 of your model configuration to executable
 [graphs](custom-graph-components.mdx#graph-components).
-Currently, Rasa Open Source only supports the `default.v1` recipe.
-Register your graph component with this recipe by using the `DefaultV1Recipe.register`
+Currently, Rasa Open Source supports the `default.v1` and the experimental `graph.v1` recipes.
+For `default.v1` recipe, you need to register your graph component by using the `DefaultV1Recipe.register`
 decorator:
 
-:::code language="python" source="docs/sources/data/test_classes/registered_component.py"
- highlight="5-9":::
```
tayfun, line 340: This was failing in current documents so I fixed it.

joejuzl:
Can we pull in someone from either Devrel or CSE to review the docs?

rasa/engine/recipes/config_files/graph_config.yml
tayfun: Yes it is used for testing at the moment. We already had default_config.yml here so I've added the graph config next to it. Is it better to move it to data dir?

data/graph_schemas/graph_config_predict_schema.yml
tayfun: Yup, removed the file for that reason.

docs/docs/graph-recipe.mdx
tayfun: Double speak? How? :)

rasa/engine/recipes/graph_recipe.py
tayfun: It is determined from CLI argument, ie. `rasa train core` or `rasa train nlu`. If neither `core` nor `nlu` is provided, in that case it is "inferred", TrainingType.BOTH by default, but if stories or nlu files are missing then it is changed to the other. Because of this, I think training type is independent of targets, and maybe later on we can let users customize targets as well.

rasa/engine/recipes/graph_recipe.py
tayfun: I wasn't sure if this would be of much use so I had left as a comment but should not be hard to change this. It would provide even more flexibility for researchers.

rasa/model_training.py
tayfun: Yes, only when we make use of recipes and only explicitly, by calling `auto_configure`. It used to be that importer would implicitly do this before. I don't see many other places using importers and everything seems fine with this change.

rasa/engine/recipes/default_recipe.py
```diff
@@ -816,3 +836,254 @@ def _add_end_to_end_features_for_inference(
...
+            missing_keys: Keys that need to be added to the config file.
+            auto_configured_keys: Keys for which a commented out auto
+            configuration section needs to be added to the config file.
+            training_type: NLU, CORE or BOTH depending on which is trained.
+        """
+        config_as_expected = DefaultV1Recipe._is_config_file_as_expected(
+            config_file_path, missing_keys, auto_configured_keys, training_type
+        )
+        if not config_as_expected:
+            rasa.shared.utils.cli.print_error(
```
ka-bu, line 967: Uh, this looks weird. Are we sure we don't want to raise here? Like if I for whatever reason managed to move my config back to it belongs (without noticing because I copy stuff back and forth and retrain and then get a different result 🤔

ka-bu (commented):
🚀

rasa/engine/recipes/default_recipe.py
```diff
@@ -816,3 +836,254 @@ def _add_end_to_end_features_for_inference(
...
+        auto-configure automatically from importers because importers are not
+        allowed to access code outside of `rasa.shared`.
+
+        Args:
+            config_file_path: The path to the configuration file.
+            config: Configuration in dictionary format.
+            training_type: Optional training type to auto-configure. By default
+            both core and NLU will be auto-configured.
+        """
+        missing_keys = DefaultV1Recipe._get_missing_config_keys(config, training_type)
```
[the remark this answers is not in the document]
tayfun: @ka-bu yup, I moved this code from importer module to recipe as auto-filling depends on recipes. It's all a bit complex at the moment but I didn't want to touch much here.

rasa/engine/recipes/default_recipe.py
tayfun: I think this is a very edge case, I won't imagine anyone changing config file while config is being read and auto-filled. I guess we decided here to continue training (not fail), be fault tolerant but warn the user too.

rasa/engine/recipes/default_recipe.py
```diff
@@ -816,3 +836,254 @@ def _add_end_to_end_features_for_inference(
...
+        updated_lines = DefaultV1Recipe._get_lines_including_autoconfig(
+            current_config_lines, autoconfig_lines
+        )
+
+        rasa.shared.utils.io.write_text_file("".join(updated_lines), config_file_path)
+
+        auto_configured_keys = transform_collection_to_sentence(auto_configured_keys)
+        rasa.shared.utils.cli.print_info(
+            f"The configuration for {auto_configured_keys} was chosen automatically. "
+            "It was written into the config file at '{config_file_path}'."
```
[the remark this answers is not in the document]
tayfun: Good catch.

rasa/engine/recipes/graph_recipe.py
```diff
@@ -0,0 +1,52 @@
...
+            raise_warning(
+                "Unlike the Default Recipe, Graph Recipe does not utilize CLI "
+                "parameters or finetuning and these configurations will be ignored. "
+                "Add configuration to the recipe itself if you want them to be used.",
+                docs=DOCS_URL_GRAPH_RECIPE,
+            )
+        # TODO: Add core_target and nlu_target to graph configuration options so we
+        # don't force graph node names to be `run_RegexMessageHandler` and
+        # `select_prediction`. These are used in validation and during execution.
+        core_target = None if training_type == TrainingType.NLU else "select_prediction"
```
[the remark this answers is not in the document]
tayfun: No constants but if I remove this here and make it configurable, the only place it is used will be default config which is tolerable I think.

rasa/shared/data.py
```diff
@@ -172,3 +172,21 @@ def _copy_files_to_new_dir(files: Iterable[Text]) -> Text:
...
+    END_TO_END = 4
+
+    @property
+    def model_type(self) -> Text:
+        """Returns the type of model which this training yields."""
+        if self == TrainingType.NLU:
+            return "nlu"
+        if self == TrainingType.CORE:
+            return "core"
+        return "rasa"
```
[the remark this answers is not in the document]
tayfun: Good catch 😅  In this case I moved this Enum class from importer module since recipes also need to know about training type and I didn't change it. `model_type` is not used in many places (mostly for telemetry) and I guess the only relevant values when used are NLU and core at the moment.

rasa/telemetry.py
```diff
@@ -782,6 +782,7 @@ def track_model_training(
             "num_synonyms": len(nlu_data.entity_synonyms),
             "num_regexes": len(nlu_data.regex_features),
             "is_finetuning": is_finetuning,
+            "recipe": config.get("recipe"),
```
wochinge, line 785: we do some key accesses to `policies` and `pipeline` a few lines up - what happens if we a use a recipe different than the default?

rasa/shared/data.py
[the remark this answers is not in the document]
tayfun: Oh forgot the part about file, I checked constants.py but it seemed only to have single line data types such as strings and this being a class with property method (model_type) I thought it made more sense put it here. I don't have strong opinions though, and I can move it there @ka-bu

rasa/telemetry.py
tayfun: They will be both empty. We can send graph recipe as train and predict instead of pipeline/policy but they are much more verbose so I don't think it's worth it.

tests/engine/recipes/test_default_recipe.py
```diff
@@ -453,3 +465,205 @@ def test_train_core_without_nlu_pipeline():
...
+def test_get_configuration(
+    config_path: Path, expected_keys_to_configure: Set[Text], monkeypatch: MonkeyPatch
+):
+    def complete_config(_, keys_to_configure: Set[Text]) -> Set[Text]:
+        return keys_to_configure
+
+    monkeypatch.setattr(DefaultV1Recipe, "_dump_config", Mock())
+    # This means get_configuration below will only return keys to configure,
+    # ie. missing keys
+    monkeypatch.setattr(DefaultV1Recipe, "complete_config", complete_config)
```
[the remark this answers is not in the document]
tayfun: Applied your suggestions and removed monkeypatch, thanks @ka-bu !

rasa/engine/recipes/default_recipe.py
ka-bu: So one could argue, if it happens so rarely, we could as well raise. But I don't have a strong opinion on this one. Think we do this in a lot of places (just warn and continue).

rasa/shared/data.py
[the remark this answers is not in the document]
ka-bu: If it had more functionality attached, then I think this would be a good place but since it's just this, I'd move it to the constants (that don't specify that only string constants are allowed). -- If no one else is confused, then feel free to leave it here

rasa/engine/recipes/config_files/graph_config.yml
joejuzl: I would move this to a test directory as it's just for testing. The other default config has meaning because it's what we autoconfigure too right?

docs/docs/graph-recipe.mdx
joejuzl: We use it to mean incremental training

rasa/shared/importers/importer.py
joejuzl: Can we log a warning when this happens?

docs/docs/graph-recipe.mdx
tayfun: Oh yes, I see what you mean. I can change "finetuning" to "configuration" here?

rasa/telemetry.py
wochinge: mhm, might still be interesting though, no? Just in case

docs/docs/graph-recipe.mdx
```diff
@@ -6,7 +6,13 @@ description: Learn about Graph Recipe for Rasa Open Source.
 abstract: Graph recipes provide a more fine tuned configuration for your executable graphs.
 ---
 
-We now support graph recipes in addition to the default recipe. What graph recipes provide is more powerful control over how execution graph schemas are built.
+:::tip Default Recipe or Graph Recipe?
+
+You will probably only need graph recipes if you're running ML experiments or ablation studies on an existing model. We recommend starting with the default recipe and for many applications that will be all that's needed.
```
joejuzl, line 11: "ablation studies" feels a bit specific? (I had to google it 😁 )

docs/docs/graph-recipe.mdx
```diff
@@ -29,31 +35,31 @@ There are some differences between the default recipe and the new graph recipe.
 
 :::tip Starting from scratch?
 
-If you don't know which recipe to choose, use the default recipe to bootstrap your project fast. If later you find that you need more powerful finetuning options, you can always change your recipe to be a graph recipe.
+If you don't know which recipe to choose, use the default recipe to bootstrap your project fast. If later you find that you need more powerful fine-tuning options, you can always change your recipe to be a graph recipe.
```
joejuzl, line 38:
```suggestion
If you don't know which recipe to choose, use the default recipe to bootstrap your project fast. If later you find that you need more fine-grained control, you can always change your recipe to be a graph recipe.
```

rasa/engine/recipes/graph_recipe.py
```diff
@@ -36,17 +36,23 @@ def graph_config_for_recipe(
...
-        core_target = None if training_type == TrainingType.NLU else "select_prediction"
+        # Note that default recipe has `nlu_target` and `core_target` as
+        # fixed values of `run_RegexMessageHandler` and `select_prediction`
+        # respectively. For graph recipe, target values are customizable. These
+        # can be used in validation (default recipe does this validation check)
+        # and during execution (all recipes use targets during execution).
+        if training_type == TrainingType.NLU:
+            core_target = None
+        else:
+            core_target = config.get("core_target", "select_prediction")
```
joejuzl, line 47: Maybe it's better to fail fast rather than have a default?

docs/docs/graph-recipe.mdx
tayfun: I've used the exact wording from @rctatman actually and I had to google it too :D I think it is good because it has the effect of warning against using graph recipes if the user is not too advanced.

rasa/engine/recipes/graph_recipe.py
tayfun: I'm more inclined towards "fail-safe" rather than "fail-fast" but @ka-bu had a similar opinion on another line where if config file changes between Rasa reading it and auto-configuring that we should fail instead of warning the user. She has a comment on this pull request. I feel defensive programming is better here as this is also the default target for default recipe? We could also add a warning log as a sort of a trade off? Otherwise I can also fail it :) Not too strong on this.

joejuzl: I'm just thinking from a user perspective:
1. User decides to create a new graph recipe config.
2. They forget to add the targets
3. They get an error saying that component `select_prediction` or `RegexMessageHandler` does not exist in the schema.
4. They assume they are missing a required component, when actually the just need to configure the targets to fit their components.

rasa/telemetry.py
tayfun: > mhm, might still be interesting though, no? Just in case

I've added training and predict schemas coming from graph recipe to telemetry now. I don't look at metabase often for this information but could be useful to others.

changelog/10473.feature.md
```diff
@@ -1 +1,7 @@
-Support graph recipes.
+Support other recipe types.
+
+This pull request also adds support for graph recipes, see details at
+https://rasa.com/docs/rasa/model-configuration check Graph Recipes section.
```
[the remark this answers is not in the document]
tayfun: It's a new page not an anchor and I can't link to it because it's not live yet, docs check fails with dead link :D

rasa/engine/recipes/graph_recipe.py
tayfun: Added code to raise with a message if targets are missing.

joejuzl (approved):
🚀 🚀

PR #10643, opened 2022-01-07
add single event schema

kalkbrennerei:
**Proposed changes**:
- add single event schema so that it can be used to validate events in rasa x


changelog/10643.misc.md
```diff
@@ -0,0 +1 @@
+Add a schema for single events.
```
ancalita, line 1: Could you please add the motivation for this refactoring please, either in the changelog or the PR description?

tests/shared/utils/schemas/test_events.py
```diff
@@ -27,4 +89,16 @@ def test_remote_action_validate_all_event_subclasses(event_class: Type[Event]):
         "wrong_action",
         "warning_predicted",
     ]:
-        jsonschema.validate(response, RemoteAction.action_response_format_spec())
+        validate(response, RemoteAction.action_response_format_spec())
+
+
+def test_validate_single_event():
```
ancalita, line 95: Doesn't this [existing test](https://github.com/RasaHQ/rasa/blob/d7456ee04bd55d49468ee82a70a123a73f9fe6f8/tests/core/test_actions.py#L507) already check for valid events?

ancalita (commented):
Looks good, just need some clarification on the refactoring motivation, also one of the tests seems to be a duplicate of an existing one for the same check.

changelog/10643.misc.md
kalkbrennerei: good point!!

changelog/10643.misc.md
```diff
@@ -0,0 +1 @@
+Add a schema for single events so that it can be used for validation.
```
ancalita, line 1: Just needs a little bit more detail, since it's not very clear to my in which context would you need to use `EVENT_SCHEMA`  rather than `EVENTS_SCHEMA`? 😅

kalkbrennerei: haha yeah it seems kind of random, but I want to use `EVENT_SCHEMA` to validate message events in rasa x. That's why I also wanted to test the validation separately from the rasaos endpoint

ancalita (approved):
👍🏼

Issue #10649, opened 2022-01-10
Add LogisticRegression as a replacement for SKlearnIntentClassifier


koaning:
### What problem are you trying to solve?

The current implementation for the `SklearnIntentClassifier` is a bit of an anti-pattern. It does a grid-search internally and it can only use the dense features in the Rasa pipeline. If we want to have an alternative to benchmark DIET, we may want to introduce another method.

### What's your suggested solution?

My proposal here is to replace it with a simple logistic regression classifier. This should train faster and it's also very lightweight.

### Examples (if relevant)

There's an implementation in [rasa nlu examples](https://github.com/RasaHQ/rasa-nlu-examples/blob/main/rasa_nlu_examples/classifiers/logistic_regression_intent_classifier.py) that we could go ahead and port.

### Is anything blocking this from being implemented? (if relevant)

 _No response_ 

### Definition of Done

We should make sure that:

*   [ ] the logistic regression gets a proper review
*   [ ] we also consider how we want to deprecate the original sklearn classifier

dakshvar22:
Do we have any numbers to see how LR models perform compared to DIET on some datasets?

koaning:
It's not hard to generate since we have an implementation. What might be good datasets? Sara?

dakshvar22:
All the datasets on which we run model regression tests

dakshvar22:
The original plan was to deprecate and remove `SklearnIntentClassifier` altogether and just keep DIET. If training times are a concern then features like incremental training can help there. The trade-off that I want to point to is that if we add a new component to Rasa, specifically an intent classifier, users have a new choice to make when deciding on the config and there is additional complexity in maintaining and guiding them on when and how to migrate to a different classifier. For those reasons, I'd like to see a very clear benefit in supporting two intent classifiers in the framework.

koaning:
I can imagine that, while fine-tuning can help, the speed of the sklearn implementation will be very hard to beat. I imagine retraining with sklearn to be faster than training a single epoch for DIET. 

I suppose the main reason to add it is to allow Rasa users to run a benchmark against a bag of words approach. If the difference between DIET and BOW is very small then this indicates that you may have an easy dataset. I could argue that this is useful knowledge. 

Another reason perhaps is that this approach does not require Tensorflow, which might be helpful to some users that experience installation issues.

PR #10650, opened 2022-01-10
Add Logistic Regression to our NLU classifiers.

koaning:
This PR will add a logistic regression. The related issue is https://github.com/RasaHQ/rasa/issues/10649.

- [x] ran the benchmarks to confirm the merits of this component


rasa/nlu/classifiers/logistic_regression_classifier.py
```diff
@@ -0,0 +1,182 @@
...
+    def __init__(
+        self,
+        config: Dict[Text, Any],
+        name: Text,
+        model_storage: ModelStorage,
+        resource: Resource,
+    ) -> None:
+        """Construct a new classifier."""
+        self.name = name
+        self.clf = LogisticRegression(
```
mleimeister, line 53: Would we want to expose more parameters than these? E.g. `random_state` would be useful for reproducibility. `solver` seems a very technical detail, and afaik there's some cross-dependency with the `penalty. Should in this case both be configurable, so users can also chose `l1` or `elasticnet`? Alternatively, could also pass on `**config` filtered for all valid parameter choices?

koaning: Oh totally! The current PR is a bit more work in progress than "ready for review". I'll update the title to reflect this. 

At the moment I'm still running the benchmarks so that we can confirm that it's not an unreasonable alternative. 

In general, I think it's best to be explicit in the parameters that we allow. The **kwargs trick is nice for quick iteration, but requires deep knowledge of sklearn which can get in the way of using the component.

koernerfelicia:
> The current PR is a bit more work in progress than "ready for review"

@koaning I hadn't realised that and came here feeling a bit bad I was late to give you a review 😅 Would you, please, ping me when it *is* ready? Alternatively, @mleimeister can also continue to review when you're ready. Or we both can, if you'd like multiple eyes and Matthias is fine with that. Whatever works, just let me know!

koaning:
@koernerfelicia my bad! This was a miscommunication on my part. I figured I'd make the PR but then Daksh (quite reasonably) asked if I could run a few benchmarks first just to be sure that this is a reasonable addition. The results are currently being discussed on [slack](https://rasa-hq.slack.com/archives/C8LEHMTAM/p1642007499069100).

koernerfelicia:
@koaning no worries! We've discussed before that the auto-tagging for code-owners is not an ideal system. Will definitely be bringing this up in the next research retro. I'll check out the Slack thread!

rasa/nlu/classifiers/logistic_regression_classifier.py
koaning: I've just added the random seed but I'm wondering if it makes sense to either be a bit minimal in the parameters that we allow or if we want to allow the user to fully customize everything. 

@koernerfelicia @mleimeister got any strong opinions on this?

koaning:
I've removed the WIP flag @koernerfelicia @mleimeister. I'm still working on the docs, but I think the code/tests are at a stage where I'm certainly open to suggestions.

rasa/nlu/classifiers/logistic_regression_classifier.py
mleimeister: Apart from allowing full customization (which I think we don't necessarily need), another argument for being explicit about all default parameters is transparency and ensuring reproducibility. If e.g. Sklearn would change a default, we wouldn't notice, but the component suddenly behaves differently. I think we list all defaults like this in `CountVectorsFeaturizer`, but some would rarely be set in the pipeline config.

rasa/nlu/classifiers/logistic_regression_classifier.py
```diff
@@ -0,0 +1,190 @@
...
+        examples = [
+            e
+            for e in training_data.intent_examples
+            if (e.get("intent") and e.get("text"))
+        ]
+
+        for e in examples:
+            # First element is sequence features, second is sentence features
+            # for intent classification we're only interested in sentence features
+            sparse_feats = e.get_sparse_features(attribute=TEXT)[1]
```
mleimeister, line 102: The extraction of `X` from an example looks like the same as in `_create_X`. Could this be merged into a helper function to be re-used in both places?

rasa/nlu/classifiers/logistic_regression_classifier.py
```diff
@@ -0,0 +1,190 @@
+import logging
+from typing import Any, Text, Dict, List, Type
+
+from joblib import dump, load
```
mleimeister, line 4: Since `dump` and `load` could be occurring in other modules as well and lead to shadowing, it might be better to call them by their full module path, as per [here](https://www.notion.so/rasa/Python-Code-Conventions-26835ffd72484dd699f28fa693916d2e#d5fb5c6142394ff99505dc8160255ef5):

```
import joblib
joblib.dump(...)
```

rasa/nlu/classifiers/logistic_regression_classifier.py
```diff
@@ -0,0 +1,190 @@
...
+    @classmethod
+    def load(
+        cls,
+        config: Dict[Text, Any],
+        model_storage: ModelStorage,
+        resource: Resource,
+        execution_context: ExecutionContext,
+    ) -> GraphComponent:
+        """Loads trained component (see parent class for full docstring)."""
+        with model_storage.read_from(resource) as model_dir:
```
mleimeister, line 174: To cover the case where there is no persisted data for the component, this should be covered in an exception block catching `ValueError`.

koaning: Is there an example of this? When would there not be any persisted data?

mleimeister: [This](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/classifiers/diet_classifier.py#L1074) would be an example. I don't think this would ever happen programatically, but I think if e.g. someone deleted the individual cached model folder in `.rasa`, but not the database containing the pointers to the cached components, the lookup would fail.

koaning: This begs the question: do we have a unit test for this behavior?

koaning:
@mleimeister what should I do with the codeclimate warning by the way? Is that something that we should ignore? It'll be hard to remove complexity there.

mleimeister:
> @mleimeister what should I do with the codeclimate warning by the way? Is that something that we should ignore? It'll be hard to remove complexity there.

I think that's fine as it's not a required check. But seems to have lifted itself :)

tests/nlu/classifiers/test_logistic_regression_classifier.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Make predictions.
+    classifier.process(training_data.training_examples)
+
+    # Check that the messages have been processed correctly
+    for msg in training_data.training_examples:
+        _, conf = msg.get("intent")["name"], msg.get("intent")["confidence"]
+        # Confidence should be between 0 and 1.
+        assert 0 < conf
+        assert conf < 1
```
mleimeister, line 84: Can move to single line: `assert 0 < conf < 1`

mleimeister (commented):
For unit tests, in order to cover the full functionality, the following should be added:
- Train -> persist -> load -> predict, as e.g. [here](https://github.com/RasaHQ/rasa/blob/main/tests/nlu/classifiers/test_keyword_classifier.py#L38)
- Test with featurizers using only dense / sparse+dense features (dense could be just some dummy vectors)
- Include `LogisticRegressionClassifier` in `tests.nlu.test_*`
- Empty text messages (if this is expected to happen?)

koaning:
@mleimeister could you clarify what you mean with `tests.nlu.test_*`? I've added an element to the `test_train.py` file. Is that what you meant?

koaning:
@mleimeister I'm getting close. One question on the empty message ... what's the expected behavior? Simply that we don't break?

docs/docs/components.mdx
```diff
@@ -1052,6 +1052,83 @@ Intent classifiers assign one of the intents defined in the domain file to incom
   ```
 
 
+### LogisticRegressionClassifier
+
+* **Short**
+
+  Logistic regression intent classifier, using the [scikit-learn implementation](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html).
```
mleimeister, line 1059: Not sure how pedantic we are about external links, but should this be hard-coded to the currently used version (I think it's `0.24.2`?) in case sklearn changes its implementation in future versions. On the other hand, it's quite likely that this link will not get updated if we upgrade the sklearn version either. So no real preference here 😄

rasa/nlu/classifiers/logistic_regression_classifier.py
```diff
@@ -0,0 +1,200 @@
...
+        probas = self.clf.predict_proba(X)
+        for idx, message in enumerate(messages):
+            intent = {"name": pred[idx], "confidence": probas[idx].max()}
+            intents = self.clf.classes_
+            intent_info = {
+                k: v
+                for i, (k, v) in enumerate(zip(intents, probas[idx]))
+                if i < LABEL_RANKING_LENGTH
+            }
+            intent_ranking = [
```
mleimeister, line 154: One issue I noticed with using this component with Moodbot is that the intent ranking seems to be unsorted. For a user text `hello`, I got the following scores:

```
[{'name': 'affirm', 'confidence': 0.018605408851111678}, {'name': 'bot_challenge', 'confidence': 0.0006403107430143258}, {'name': 'deny', 'confidence': 0.025455037690681255}, {'name': 'goodbye', 'confidence': 0.013747843491355998}, {'name': 'greet', 'confidence': 0.914782164381611}, {'name': 'mood_great', 'confidence': 0.017658700205053904}, {'name': 'mood_unhappy', 'confidence': 0.009110534637171744}]
```

This in turn leads to `FallbackClassifier` being triggered, since it will look at the difference of the first two elements, which is below the default ambiguity threshold. Therefore the probabilities need to be sorted, as e.g. [here](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/classifiers/sklearn_intent_classifier.py#L257).

koaning: Well spotted!

docs/docs/components.mdx
```diff
@@ -1052,6 +1052,83 @@ Intent classifiers assign one of the intents defined in the domain file to incom
...
+        {
+            "confidence": 0.780,
+            "name": "greet"
+        },
+        {
+            "confidence": 0.140,
+            "name": "goodbye"
+        },
+        {
+            "confidence": 0.0810,
```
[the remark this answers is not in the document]
koaning: Ah that shouldn't happen. Lemme fix that!

docs/docs/components.mdx
koaning: I think it's best to leave as-is. Mainly because keeping an eye on the docs is likely going to be more work and it's not unlikely that Rasa will keep sklearn updated.

rasa/nlu/classifiers/logistic_regression_classifier.py
koaning: Also added a test for it.

mleimeister:
> @mleimeister I'm getting close. One question on the empty message ... what's the expected behavior? Simply that we don't break?

Yes I think that's what some of the other components do in terms of checks, e.g. [here](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/classifiers/sklearn_intent_classifier.py#L198). 

However, I was not able to break the component with empty text. It moreover seems that even with empty text, our featurizers do produce feature vectors for the `Message`, so they are actually never empty when they arrive at the classifier (see screenshot below). Therefore I'd say the current behaviour is good as it is and an upcoming rework of the NLU components should rather define the contract between different components more clearly, after which it would be much easier to say what such checks should look like.

In that sense LGMT 👍 

Screenshot 2022-02-09 at 16 45 18

mleimeister (approved):

koaning:
🎉

Issue #10519, opened 2021-12-13
(3.0) Error initializing ResponseSelector in `rasa test nlu` with `use_text_as_label: true`


melindaloubser1:
### Rasa Open Source version

3.0.2

### Python version

3.8

### What operating system are you using?

OSX

### What happened?

Trained a model with the following config without error:

config.yml
```
language: en

pipeline:
  - name: WhitespaceTokenizer
  - name: RegexFeaturizer
  - name: LexicalSyntacticFeaturizer
  - name: CountVectorsFeaturizer
  - name: ResponseSelector
    epochs: 10
    use_text_as_label: true
    number_of_transformer_layers: 1
```
  
But running `rasa test nlu` fails with the error below. I tried a few variations  on transformer configuration with the same result e.g. 
```
  - name: ResponseSelector
    epochs: 10
    use_text_as_label: true
    number_of_transformer_layers: 1
    transformer_size: 256
    hidden_layers_sizes:
      text: []
      label: []
```

### Command / Request

```shell
rasa train nlu
rasa test nlu
```

### Relevant log output

```shell
2021-12-13 16:57:30 INFO     rasa.core.processor  - Loading model models/nlu-20211213-165606-muffled-procedure.tar.gz...
Traceback (most recent call last):
  File "/Users/melinda/.pyenv/versions/3.8.12/lib/python3.8/site-packages/rasa/engine/graph.py", line 393, in _load_component
    self._component: GraphComponent = constructor(  # type: ignore[no-redef]
  File "/Users/melinda/.pyenv/versions/3.8.12/lib/python3.8/site-packages/rasa/nlu/selectors/response_selector.py", line 669, in load
    model: ResponseSelector = super().load(
  File "/Users/melinda/.pyenv/versions/3.8.12/lib/python3.8/site-packages/rasa/nlu/classifiers/diet_classifier.py", line 1076, in load
    return cls._load(
  File "/Users/melinda/.pyenv/versions/3.8.12/lib/python3.8/site-packages/rasa/nlu/classifiers/diet_classifier.py", line 1107, in _load
    model = cls._load_model(
  File "/Users/melinda/.pyenv/versions/3.8.12/lib/python3.8/site-packages/rasa/nlu/classifiers/diet_classifier.py", line 1200, in _load_model
    model = cls._load_model_class(
[...]

PR #10651, opened 2022-01-10
Correct `transformer_size` parameter if needed for `ResponseSelector` during loading

jupyterjazz:
**Proposed changes**:
- Corrects `transformer_size` parameter in `ResponseSelector` which is None by default with a constant integer during model loading
- Otherwise it could lead to errors like #10519
- Credits to @mleimeister for an [informative and clear analysis](https://github.com/RasaHQ/rasa/issues/10519#issuecomment-999599561)


rasa/utils/tensorflow/rasa_layers.py
```diff
@@ -805,6 +806,7 @@ def _get_transformer_dimensions(
         transformer_units = config[TRANSFORMER_SIZE]
         if isinstance(transformer_units, dict):
             transformer_units = transformer_units[attribute]
+        transformer_units = transformer_units or DEFAULT_TRANSFORMER_SIZE
```
ancalita, line 809: Nitpicking: would it be more readable if this would be rewritten using the `if/else` ternary operator?

jupyterjazz: you mean changing to this?
`transformer_units = transformer_units if transformer_units else DEFAULT_TRANSFORMER_SIZE`

mleimeister (commented):
Does the proposed fix in `RasaSequenceLayer` also take effect in the training phase of `ResponseSelector`, not just model loading? In that case the original function [ResponseSelector._warn_and_correct_transfomer_size](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/selectors/response_selector.py#L387) would be redundant and could be removed. 

However, if this gets problematic with the `ResponseSelector` tests (afaik they check that this function is run as expected), it could also be dealt with in a follow-up ticket.

mleimeister:
For transparency, you could add to the docstring of `RasaSequenceLayer._get_transformer_dimensions` to clarify that `DEFAULT_TRANSFORMER_SIZE` is used if none or 0 given.

mleimeister:
The existing [check](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/selectors/response_selector.py#L394) in `ResponseSelector._warn_and_correct_transfomer_size` tests for both `transformer_size == None` and `transformer_size < 1`, which allows to correct configs like this [test case](https://github.com/RasaHQ/rasa/blob/main/tests/nlu/selectors/test_selectors.py#L556), where `transformer_size = -1` is expected to be corrected to the default size. To be consistent the same check should be applied in the fix here.

rasa/utils/tensorflow/rasa_layers.py
mleimeister: The existing [check](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/selectors/response_selector.py#L394) in `ResponseSelector._warn_and_correct_transfomer_size` tests for both `transformer_size is None` and `transformer_size < 1`, which allows to correct configs like this [test case](https://github.com/RasaHQ/rasa/blob/main/tests/nlu/selectors/test_selectors.py#L556), where `transformer_size = -1` is expected to be corrected to the default size. To be consistent the same check should be applied in the fix here.

jupyterjazz:
> Does the proposed fix in `RasaSequenceLayer` also take effect in the training phase of `ResponseSelector`, not just model loading? In that case the original function [ResponseSelector._warn_and_correct_transfomer_size](https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/selectors/response_selector.py#L387) would be redundant and could be removed.
> 
> However, if this gets problematic with the `ResponseSelector` tests (afaik they check that this function is run as expected), it could also be dealt with in a follow-up ticket.

Yes, it does and as you predicted removing the existing correction causes some problems in the tests. I'll create a separate ticket to address that. 
Just to clarify, should we remove the whole [function](https://github.com/RasaHQ/rasa/blob/85ffd7c87ff465430ad5387dd6c000e8ccf4bd5d/rasa/nlu/selectors/response_selector.py#L387) or should we leave the warning and disable the correction?

ancalita (approved):
LGTM 👍🏼

mleimeister:
> Just to clarify, should we remove the whole [function](https://github.com/RasaHQ/rasa/blob/85ffd7c87ff465430ad5387dd6c000e8ccf4bd5d/rasa/nlu/selectors/response_selector.py#L387) or should we leave the warning and disable the correction?

Since the default is then set in `RasaSequenceLayer`, the warning could be moved there as well and the original function be removed completely. That way, the warning would be logged both when training and loading a model?

mleimeister:
LGTM 🚀

mleimeister (approved):

PR #10658, opened 2022-01-11
Added small clarification related to pretrained models

koaning:
Added a note to the docs. Some forum members were a bit confused about pre-trained models in DIET.

### DIFF
--- a/docs/docs/components.mdx
+++ b/docs/docs/components.mdx
@@ -1246,7 +1246,7 @@ Intent classifiers assign one of the intents defined in the domain file to incom
   single semantic vector space. We use the dot-product loss to maximize the similarity with the target label and
   minimize similarities with negative samples.
 
-  DIET does not provide pre-trained word embeddings or language models but it is able to use these features if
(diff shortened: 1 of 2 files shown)


docs/docs/components.mdx
```diff
@@ -1246,7 +1246,8 @@ Intent classifiers assign one of the intents defined in the domain file to incom
   single semantic vector space. We use the dot-product loss to maximize the similarity with the target label and
   minimize similarities with negative samples.
 
-  If you want to learn more about the model, check out the
+  DIET does not provide pre-trained word embeddings or language models but it is able to use these features if
```
samsucik, line 1249: Just to make it 100% clear:
```suggestion
  DIET does not provide pre-trained word embeddings or pre-trained language models but it is able to use these features if
```

samsucik (changes requested):
@koaning all is looking good but let's add a changelog entry as well.

samsucik (approved):
Let's merge this! Thanks, @koaning 🚀

PR #10663, opened 2022-01-11
Sorting lookup elements on bases of length in descending order

azizullah2017:
Sorting lookup elements on bases of length in descending order, so the larger value of lookup should be matched first. The generator matches the first value and return.

## Unsorted lookup

lookup:
- one
- twenty
- twenty one

**if input is "twenty one"**. it matched with **"one"** and **"twenty"**

## After storing in descending order
- twenty one
- one
- twenty

now if if input is "twenty one", it matched with **"twenty one"**


melindaloubser1 (changes requested):
Hi @azizullah2017, this PR is quite old, if you do still want to merge it, could you add or expand a unit test to check that this fixes the behavior you mentioned?

PR #10002, opened 2021-10-28
Add Python 3.9 support

alwx:
**Proposed changes**:
- fix https://github.com/RasaHQ/rasa/issues/8762
- adds Python 3.9 support to rasa

**Status (please check what you already did)**:
- [ ] added some tests for the functionality
- [ ] updated the documentation
- [ ] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [ ] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #10684, opened 2022-01-13
8762/python3.9

### LINKED ISSUE #8762: Add support to Python 3.9

ka-bu:
**Proposed changes**:
- closes #8762 
- based on work from #10002

**Progress**:
- [x] `full model training` fails: breaking with OOM when training TED which is known to be flakey -> `wontfix`
- [x] `test other unit test` fails: reverted `test_server` changes; switched to current `sanic`; tried with(out) asyncio workaround (uvloop issue persists); ....
- [x] rebase changes after (squash) merging tf2.7 update 
- [x] fix black issues -- if there are any left


usc-m:
Looks fine - some style fixes needed (unused imports) + need to investigate if we can upgrade `uvloop` to make the workaround unnecessary

usc-m:
Update on this: We're investigating if we can avoid having a workaround for uvloop issues. It's looking like we can, as the underlying bug in uvloop is fixed in 0.15.0 or newer, but rasa-sdk pins it at < 0.15.0 which prevents us from getting the fix. After some investigation it looks like that pin can be removed (it was pinned for reasons which no longer apply a long time ago), which will remove the constraint on it and permit rasa to pull in the correct version. I'm looking into making the update to rasa-sdk now.

wochinge:
I tagged a 3.0.6 SDK without the uvloop pin

rasa/cli/test.py
```diff
@@ -267,7 +266,7 @@ def run_core_test(args: argparse.Namespace) -> None:
     Args:
         args: the parsed CLI arguments for 'rasa test core'.
     """
-    asyncio.run(run_core_test_async(args))
+    asyncio_run_workaround(run_core_test_async(args))
```
wochinge, line 269: @ka-bu Why do we need this?

ka-bu: we needed it because we had trouble with the version of uvloop then

wochinge: okay, so we can revert that? sweet

wochinge:
@ka-bu Do you wanna give it a review?

rasa/utils/tensorflow/temp_keras_modules.py
```diff
@@ -279,7 +279,6 @@ def fit(
             ValueError: In case of mismatch between the provided input data
                 and what the model expects or when the input data is empty.
         """
-        base_layer.keras_api_gauge.get_cell("fit").set(True)
```
ka-bu, line 282: this change should be reverted because we mimic the keras 2.7 version here and that contains that: https://github.com/keras-team/keras/blob/v2.7.0/keras/engine/training.py#L1131

ka-bu (commented):
Just one tiny thing. Otherwise looks good to me :)

wochinge (approved):

Issue #10683, opened 2022-01-13
Docs list faulty configuration for CRFEntityExtractor


koaning:
On our [documentation](https://rasa.com/docs/rasa/components#crfentityextractor) you can find this configuration example for the CRFEntityExtractor:

```
pipeline:
- name: "CRFEntityExtractor"
  # BILOU_flag determines whether to use BILOU tagging or not.
  "BILOU_flag": True
  # features to extract in the sliding window
  "features": [
    ["low", "title", "upper"],
    [
      "bias",
      "low",
      "prefix5",
      "prefix2",
      "suffix5",
      "suffix3",
      "suffix2",
      "upper",
      "title",
      "digit",
      "pattern",
    ],
    ["low", "title", "upper"],
    ["text_dense_features"]
  ]
  # The maximum number of iterations for optimization algorithms.
  "max_iterations": 50
  # weight of the L1 regularization
  "L1_c": 0.1
  # weight of the L2 regularization
  "L2_c": 0.1
  # Name of dense featurizers to use.
  # If list is empty all available dense features are used.
  "featurizers": []
  # Indicated whether a list of extracted entities should be split into individual entities for a given entity type
  "split_entities_by_comma":
      address: False
      email: True
```

I thought it a bit strange that the `features` key has a list of four items. I figured there needs to be a middle word. But I copied the snippet because I figured that the docs ought to be correct. 

This resulted in an error [when running benchmarks](https://github.com/RasaHQ/rasa/runs/4801616608?check_suite_focus=true). 

```
Traceback (most recent call last):
  File "/home/runner/work/rasa/rasa/rasa/engine/graph.py", line 393, in _load_component
    self._component: GraphComponent = constructor(  # type: ignore[no-redef]
  File "/home/runner/work/rasa/rasa/rasa/nlu/extractors/crf_entity_extractor.py", line 217, in create
    return cls(config, model_storage, resource)
  File "/home/runner/work/rasa/rasa/rasa/nlu/extractors/crf_entity_extractor.py", line 196, in __init__
    self._validate_configuration()
  File "/home/runner/work/rasa/rasa/rasa/nlu/extractors/crf_entity_extractor.py", line 204, in _validate_configuration
    raise ValueError(
ValueError: Need an odd number of crf feature lists to have a center word.
```

My proposal is to change it to: 

```
  "features": [
    ["low", "title", "upper"],
    [
      "bias",
      "low",
      "prefix5",
      "prefix2",
      "suffix5",
[...]

PR #10687, opened 2022-01-14
Fix CRFEntityExtractor docs example. 

koaning:
This PR fixes https://github.com/RasaHQ/rasa/issues/10683.


samsucik (approved):
Thanks, @koaning!

PR #10689, opened 2022-01-16
Remove unfeaturized messages for some NLU components before training and prediction

### LINKED ISSUE #10504: Empty (or some spaces) input text causes Runtime error
### Rasa Open Source version

3.0.0

### Rasa SDK version

3.0.0

### Rasa X version

_No response_

### Python version

3.8

Linux

### What happened?

In NLU shell, when I input text="" or text="     "(some spaces), it gets error as follow (full details in 'Relevant log output') :
`
rasa.engine.exceptions.GraphComponentException: Error running graph component for node run_DIETClassifier4.
`

But in RASA-2.8.6, everything goes well and gets following response:
`
{
  "text": "",
  "intent": {
    "name": null,
    "confidence": 0.0
  },
  "entities": []
}
`

I suppose this is a new bug.

### Command / Request

_No response_

### Relevant log output

```shell
File "~/venv/lib/python3.8/site-packages/rasa/engine/graph.py", line 461, in __call__
output = self._fn(self._component, **run_kwargs)
File "~/venv/lib/python3.8/site-packages/rasa/nlu/classifiers/diet_classifier.py", line 1003, in process
out = self._predict(message)
File "~/venv/lib/python3.8/site-packages/rasa/nlu/classifiers/diet_classifier.py", line 932, in _predict
return self.model.run_inference(model_data)
File "~/venv/lib/python3.8/site-packages/rasa/utils/tensorflow/models.py", line 318, in run_inference
] = self._rasa_predict(batch_in)
File "~/venv/lib/python3.8/site-packages/rasa/utils/tensorflow/models.py", line 280, in _rasa_predict
outputs = tf_utils.sync_to_numpy_or_python_type(self.predict_step(batch_in))
File "~/venv/lib/python3.8/site-packages/rasa/utils/tensorflow/models.py", line 228, in predict_step
return self.batch_predict(batch_in)
File "~/venv/lib/python3.8/site-packages/rasa/nlu/classifiers/diet_classifier.py", line 1721, in batch_predict
tf_batch_data = self.batch_to_model_data_format(
File "~/venv/lib/python3.8/site-packages/rasa/utils/tensorflow/models.py", line 473, in batch_to_model_data_format
tensor, idx = RasaModel._convert_sparse_features(
File "~/venv/lib/python3.8/site-packages/rasa/utils/tensorflow/models.py", line 518, in _convert_sparse_features
shape = [batch[idx + 2][i] for i in range(number_of_dimensions - 1)] + [
File "~/venv/lib/python3.8/site-packages/rasa/utils/tensorflow/models.py", line 518, in <listcomp>
shape = [batch[idx + 2][i] for i in range(number_of_dimensions - 1)] + [
IndexError: index 1 is out of bounds for axis 0 with size 1

The above exception was the direct cause of the following exception:

File "~/venv/bin/rasa", line 8, in <module>
sys.exit(main())
File "~/venv/lib/python3.8/site-packages/rasa/__main__.py", line 121, in main
File "~/venv/lib/python3.8/site-packages/rasa/cli/shell.py", line 91, in shell_nlu
rasa.nlu.run.run_cmdline(model)
File "~/venv/lib/python3.8/site-packages/rasa/nlu/run.py", line 25, in run_cmdline
result = asyncio.run(agent.parse_message(message))
File "/usr/lib/python3.8/asyncio/runners.py", line 44, in run
return loop.run_until_complete(main)
File "~/venv/lib/python3.8/site-packages/rasa/core/agent.py", line 419, in parse_message
return await self.processor.parse_message(message)
File "~/venv/lib/python3.8/site-packages/rasa/core/processor.py", line 635, in parse_message
parse_data = self._parse_message_with_graph(message, only_output_properties)
File "~/venv/lib/python3.8/site-packages/rasa/core/processor.py", line 659, in _parse_message_with_graph
results = self.graph_runner.run(
File "~/venv/lib/python3.8/site-packages/rasa/engine/runner/dask.py", line 106, in run
dask_result = dask.get(run_graph, run_targets)
File "~/venv/lib/python3.8/site-packages/dask/local.py", line 560, in get_sync
return get_async(
File "~/venv/lib/python3.8/site-packages/dask/local.py", line 503, in get_async
for key, res_info, failed in queue_get(queue).result():
File "/usr/lib/python3.8/concurrent/futures/_base.py", line 437, in result
return self.__get_result()
File "/usr/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
raise self._exception
File "~/venv/lib/python3.8/site-packages/dask/local.py", line 545, in submit
fut.set_result(fn(*args, **kwargs))
File "~/venv/lib/python3.8/site-packages/dask/local.py", line 237, in batch_execute_tasks
return [execute_task(*a) for a in it]
File "~/venv/lib/python3.8/site-packages/dask/local.py", line 237, in <listcomp>
return [execute_task(*a) for a in it]
File "~/venv/lib/python3.8/site-packages/dask/local.py", line 228, in execute_task
result = pack_exception(e, dumps)
File "~/venv/lib/python3.8/site-packages/dask/local.py", line 223, in execute_task
result = _execute_task(task, data)
File "~/venv/lib/python3.8/site-packages/dask/core.py", line 121, in _execute_task
return func(*(_execute_task(a, cache) for a in args))
File "~/venv/lib/python3.8/site-packages/rasa/engine/graph.py", line 468, in __call__
raise GraphComponentException(
rasa.engine.exceptions.GraphComponentException: Error running graph component for node run_DIETClassifier4.
```

**Definition of done**
- [ ] Fix bug

jupyterjazz:
**Proposed changes**:
- Fixes a bug (#10504) which was caused by `DIETClassifier` (`ResponseSelector`, `SklearnIntentClassifier` and `CRFEntityExtractor` have the same issue) trying to process message which didn't have required features. Implements removing unfeaturized messages for the above-mentioned components before training and prediction.


jupyterjazz:
/modeltest
```yml
include:
  - dataset: ["all"]
    config: ["all"]
```

jupyterjazz:
/modeltest
```yml
include:
  - dataset: ["all"]
    config: ["all"]
```

jupyterjazz:
Two regression tests are failing. 
@dakshvar22 any idea on how this is related to the changes I've made? If I got it right, the error happens during training which should not be affected by my changes because I only modified `process` methods.
```
  File "/home/github/_work/rasa/rasa/rasa/engine/graph.py", line 461, in __call__
    output = self._fn(self._component, **run_kwargs)
  File "/home/github/_work/rasa/rasa/rasa/nlu/classifiers/diet_classifier.py", line 904, in train
    self.model.fit(
  File "/home/github/_work/rasa/rasa/rasa/utils/tensorflow/temp_keras_modules.py", line 190, in fit
    tmp_logs = train_function(iterator)
  File "/home/github/.cache/pypoetry/virtualenvs/rasa-ip_mm8IB-py3.8/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 885, in __call__
    result = self._call(*args, **kwds)
  File "/home/github/.cache/pypoetry/virtualenvs/rasa-ip_mm8IB-py3.8/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 917, in _call
    return self._stateless_fn(*args, **kwds)  # pylint: disable=not-callable
  File "/home/github/.cache/pypoetry/virtualenvs/rasa-ip_mm8IB-py3.8/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 3039, in __call__
    return graph_function._call_flat(
  File "/home/github/.cache/pypoetry/virtualenvs/rasa-ip_mm8IB-py3.8/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 1963, in _call_flat
    return self._build_call_outputs(self._inference_function.call(
  File "/home/github/.cache/pypoetry/virtualenvs/rasa-ip_mm8IB-py3.8/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 591, in call
    outputs = execute.execute(
  File "/home/github/.cache/pypoetry/virtualenvs/rasa-ip_mm8IB-py3.8/lib/python3.8/site-packages/tensorflow/python/eager/execute.py", line 59, in quick_execute
    tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
tensorflow.python.framework.errors_impl.ResourceExhaustedError:  OOM when allocating tensor with shape[232,217,217] and type float on /job:localhost/replica:0/task:0/device:GPU:0 by allocator GPU_0_bfc
	 [[node zeros_like_41 (defined at /home/github/_work/rasa/rasa/rasa/utils/tensorflow/models.py:158) ]]
Hint: If you want to see a list of allocated tensors when OOM happens, add report_tensor_allocations_upon_oom to RunOptions for current allocation info. This isn't available when running in Eager mode.
 [Op:__inference_train_function_285315]

```

tests/nlu/selectors/test_selectors.py
```diff
@@ -615,6 +615,34 @@ def test_transformer_size_gets_corrected(train_persist_load_with_different_setti
...
+    message = Message(data={TEXT: ""})
+    message = process_message(loaded_pipeline, message)
+
+    classified_message = response_selector.process([message])[0]
+    output = classified_message.get("response_selector").get("default").get("response")
+
+    assert classified_message.get(TEXT) == ""
+    assert not output.get("responses")
+    assert output.get("confidence") == 0.0
+    assert not output.get("intent_response_key")
```
dakshvar22, line 643: Please use existing constants for the strings over here. You'll find them being used in `process` method of `ResponseSelector` class as well.

dakshvar22:
> Two regression tests are failing

Yeah you can ignore them.. It's a known failure.

ka-bu:
What about the "or some spaces" stuff? Do we expect the spaces to be removed by the whitespace tokenizer or so?

jupyterjazz:
> What about the "or some spaces" stuff? Do we expect the spaces to be removed by the whitespace tokenizer or so?

Yes, in the original bug (using `JiebaTokenizer`) "some spaces" were transformed into an empty string which then caused the bug. Same happens with `WhitespaceTokenizer`. I don't think we have any reason to test spaces as well.

jupyterjazz:
@dakshvar22 is the [third failing check](https://github.com/RasaHQ/rasa/runs/4847728641?check_suite_focus=true) also expected?

tests/nlu/selectors/test_selectors.py
jupyterjazz: forgot to change this
wIll do now

jupyterjazz:
Oh, wow. I've been doing this so wrong. 

1. You're right, tokenizers shouldn't and don't manipulate raw text. When I was testing manually I, for some reason, assumed that tokenizers were doing that when in reality it was this line
https://github.com/RasaHQ/rasa/blob/e522be7d989de94f5c2c57f86dd6fc1626941efb/rasa/core/channels/channel.py#L60

2. I zoomed in on the code and `MitieFeaturizer` is not a problem here. It's just some tokenizers - `JiebaTokenizer`, `MitieTokenizer` and `SpacyTokenizer` - don't produce tokens when given an empty string which then caused problems in the featurizers (`MitieFeaturizer` in this case). This means we don't need to modify other components. 

3. Your comment makes sense. Checking the required features would be the correct way to go. Thanks. 

4. I looked at the featurizers and most of them use only tokens for creating features, except `LanguageModelFeaturizer` and `ConveRTFeaturizer` - which look like they're using raw text
https://github.com/RasaHQ/rasa/blob/e522be7d989de94f5c2c57f86dd6fc1626941efb/rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py#L386
But I'm not 100% sure because I can't run these on my laptop. Intent classifiers use these features so it looks like there are some featurizers that produce features on the raw text which are used by intent classifiers.

ka-bu:
> 4. I looked at the featurizers and most of them use only tokens for creating features, except `LanguageModelFeaturizer` and `ConveRTFeaturizer` - which look like they're using raw text
>    https://github.com/RasaHQ/rasa/blob/e522be7d989de94f5c2c57f86dd6fc1626941efb/rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py#L386

Ah, so great that you checked those, I had missed that :)

Hm, and I think the story gets even a bit more confusing:  https://github.com/RasaHQ/rasa/blob/e522be7d989de94f5c2c57f86dd6fc1626941efb/rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py#L418
Seems for the sequence features, convert is using the token it seems. While for the sentence features, we use the raw text. 

There's some notion pages with internal documentation that I've started (as preparation for refactoring / better external docs etc) where we should add all this. I'll send you the link on slack because I think we're not supposed to put notion page links here. 

>    But I'm not 100% sure because I can't run these on my laptop. Intent classifiers use these features so it looks like there are some featurizers that produce features on the raw text which are then used by intent classifiers. (if this is what you asked)

Might be good to get setup on cloud then. I'll send you some links

> 5. Did you mean `CountVectorsFeaturizer` here?
> 
> > (e.g. whitespace tokenizer kind of works with whole text when producing sentence features but that text is actually just the concatenated tokens)

Oh, yes, that's what I meant :) 

> Checking the required features would be the correct way to go. 

Yep, agree, looking at tokens alone seems insufficient (plus the"workaround" with the used features is actually nicer/easier to understand :)  Btw, for the "required" features there's a "Features.filter" function (or something like that).

ka-bu:
Hm, I don't think we should add an artificial "empty token" if there is no token, because this will mean the intent classifiers etc have to work and will work with that empty token and produce some intent. Instead of relying on the intent classifiers to do something meaningful to that (which they probably can't - because they've not really been taught how), it seems clearer to me if the intent classifier would just not output an intent in that case. 
An alternative route that enables that would be: 
- Before we process during inference, we assert that the message contains some of the features requested by the classifier (if it does) or some at all (if it doesn't). If that can't be asserted, then we don't set (or remove) the intent along with the ranking  ranking
- Before we train, we filter out those messages that do not pass that assertion we do during inference.
Does that make sense or did I miss something?

jupyterjazz:
@ka-bu, 
Agree, handling unfeaturized messages seems more reasonable. Can you look at the draft version I created? Does it make sense?

rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -706,6 +706,12 @@ def _create_model_data(
                 example for example in training_data if label_attribute in example.data
             ]
 
+        training_data = [
+            example
+            for example in training_data
+            if example.features_present(attribute=TEXT)
```
ka-bu, line 712: yep good start 👍  -- we should also check that the features that are actually used by diet are checked for here (there should be a list of features in the config under something like `FEATURIZERS`) (and then we need to do the same in the `process` method)

jupyterjazz: This part of the code is called during both training and prediction

ka-bu: Finally realised that this must not be the same in training and prediction because otherwise we return less predictions than we requested. .... And I just realise this change is outdated so probably already fixed 👀

rasa/nlu/utils/data_utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+    Returns:
+        List of messages where unfeaturized messages are removed.
+    """
+    filtered_messages = [
+        message
+        for message in messages
+        if message.features_present(attribute=attribute, featurizers=featurizers)
+    ]
+    return filtered_messages
```
ka-bu, line 27: I appreciate that you tried to remove complexity and added the docstring with the explanation + the unit test 🚀 , but I think in this case the extra layer of abstraction is not really needed: 
- During prediction, we can just call `message.features_present(attribute=attribute, featurizers=featurizers)` directly. 
- During training, the extra lines don't hurt much and it is more transparent/explicit (i.e. easier to notice that we only check whether any of the requested featurizers produced something and not all of them - which is nit-picking but might become relevant for some cases later)
- Moreover, we already have tests for `features_present` and so what this really only tests is that the listcomp is implemented ok (which we get for free via the new `test_process_unfeatrized` ...)
Does that make sense?

jupyterjazz: Yeap, definitely makes sense. Now that I think about it I'm not even sure why I decided to go for this tbh :smile:

ka-bu (approved):
LGTM :)

PR #10700, opened 2022-01-18
Daniel/9789 allow custom request timeout

donodje:
**Proposed changes**:
-  add the `--request-timeout` flag to rasa cli commands allowing a user to set the request timeout for action server calls


rasa/core/constants.py
```diff
@@ -1,3 +1,14 @@
+import sys
+
+try:
+    # sys.argv is a list so if we find "--request-timeout"
+    # the next item on the list is the passed value
+    arg_index = sys.argv.index("--request-timeout")
```
tayfun, line 6: Is there a better way to set this constant after argparse has parsed it? Using `sys` seems to beat the purpose of `argparse`. What do you think? Maybe calling a function to reset a set of arguments such as `*_TIMEOUT` after argparse has finished parsing CLI arguments? Maybe another better way?

usc-m: Yeah, agree with Tayfun here - we probably also don't want to set constants dynamically like this.

tayfun: @usc-m We have request timeout as CLI parameter which if provided should overwrite this default. This way, anywhere else in code which imports this constant will use CLI updated or default value. If we leave the constant as it is here, we will have to pass the CLI variable to any code which may use it and that code should use either that or this one which seems a little cumbersome?

tayfun: @joejuzl Do you have an idea?

joejuzl: I'd say that fact that we're changing it dynamically shows it's not a constant (only the default value is a constant). So while it's cumbersome I feel like we should pass it from the CLI to wherever it's needed. This also means that we are not necessarily restricted to one global timeout, and can potentially handle it differently for different use cases (e.g. remote actions, model pulling etc.)

rasa/core/channels/console.py
```diff
@@ -125,14 +125,18 @@ async def send_message_receive_block(
 
 
 async def _send_message_receive_stream(
-    server_url: Text, auth_token: Text, sender_id: Text, message: Text
+    server_url: Text,
+    auth_token: Text,
+    sender_id: Text,
+    message: Text,
+    request_timeout: Optional[int],
```
usc-m, line 132: Probably should have a default of `None` as well so it's an optional arg

usc-m (commented):
Looks good to me! I spotted one optional parameter that wasn't actually optional, but everything else is good

rasa/core/channels/console.py
```diff
@@ -142,14 +146,21 @@ async def _send_message_receive_stream(
                     yield json.loads(line.decode(DEFAULT_ENCODING))
 
 
-def _get_stream_reading_timeout() -> ClientTimeout:
-    timeout_in_seconds = int(
-        os.environ.get(
-            STREAM_READING_TIMEOUT_ENV, DEFAULT_STREAM_READING_TIMEOUT_IN_SECONDS
```
tayfun, line 148: I hadn't realised there was a `STREAM_READING_TIMEOUT_ENV` env variable which can be used to set console channel timeout 😳 

The problem is I think all these settings such as STREAM_READING_TIMEOUT_ENV, DEFAULT_STREAM_READING_TIMEOUT_IN_SECONDS, and DEFAULT_REQUEST_TIMEOUT are confusing; semantically they mean the same thing though in practice they are not used consistently. What I mean is, stream reading timeout sounds like an even more general setting than request timeout but it is only used for console channel, not for example for action server calls. It's a little confusing as one is on the client side (shell) and the other on the server (Rasa calling action server) but both are controlled by the same CLI and setting semantically applies to requests made from both client and server. 

@joejuzl may have better ideas around this.

rasa/utils/endpoints.py
```diff
@@ -10,7 +10,7 @@
 from rasa.shared.exceptions import FileNotFoundException
 import rasa.shared.utils.io
 import rasa.utils.io
-from rasa.constants import DEFAULT_REQUEST_TIMEOUT
+from rasa.core.constants import DEFAULT_REQUEST_TIMEOUT
```
tayfun, line 13: This parameter won't be effected by the new CLI argument and will continue with the value initially assigned in constants file. Same with `rasa/core/actions/action.py` where requests to action server is made. Is this expected? @joejuzl @usc-m 

We may want to differentiate between these request timeouts? Another option is moving these variables from constants to a settings file and in there overriding the values from command lines? Semantically this would be better than changing inside "constants" file. Also they are not constants if they can be changed?

tayfun (approved):

rasa/core/channels/console.py
usc-m: I think possibly changing them to make them more semantically meaningful might be a good idea, but since it would be a breaking change we'd probably have to plan it out a bit more carefully than can be done in the scope of this ticket

rasa/utils/endpoints.py
usc-m: The defaults are constants, it's just that we didn't have any way of making them non-default until recently it looks like?

Agree with the semantic confusion though - it is very confusing figuring out what should be affected by which setting especially as the names are very unclear

joejuzl: It's not totally clear from the issue which timeouts are important to be able to change. Might be worth clarifying with the author.

Issue #10703, opened 2022-01-19
Rule policy consumes a lot of CPU at inference time


hsm207:
### Rasa Open Source version

2.8.21

### Rasa SDK version

 _No response_ 

### Rasa X version

 _No response_ 

### Python version

3.8

### What operating system are you using?

Linux

### What happened?

Rule Policy consumes a lot of CPU cycles when there are a lot of rules

### Command / Request

Run this [bot](https://github.com/RasaHQ/rule_policy_spike_cpu).

### Relevant log output

 _No response_

tayfun:
We can add unit tests to check no extra computations are made; probably we didn’t test with 4000 rules while developing RulePolicy? `_is_rule_applicable` seems to be the crux of rule policy and we seem to cache `_rule_key_to_state` for 1000 items (which means with 4000 keys it can have cache misses 1 in 4 random cases); we also run `list(reversed(state_list)`  on each run and this can be cached too I'm thinking. These are observations just by reading the code, detailed analysis is needed.

dsjaai:
What do you think of using a decision tree structure to identify rules that can be applied?

For each rule in the training data one would add each condition and requirement (like last intent must be x) as decisions to the tree. 
In addition nodes hold all the rules that can be applied if all conditions from the root until that node are fulfilled or not fulfilled as defined by the path.
As we know by definition, there should be at most one rule, or at least, there should be only non contradicting rules that can be applied at the same time. So, we could simply run through the tree top-down (starting at the root) until a rule is found. If no rule is found until reaching a leave, no rule can be applied.

tayfun:
I've closed the PR linked to this ticket, we can still check it out when we start working on this ticket. It has some guidance on what to look for.

PR #10704, opened 2022-01-19
Increase cache size for RulePolicy

tayfun:
Fixes https://github.com/RasaHQ/rasa/issues/10703

**Proposed changes**:
- Increase cache size for RulePolicy

**Context for the fix**:

This actually came up on the on-point Slack channel after a customer reported higher CPU usage when they upgraded Rasa from 1.5.1 to 2.5.0. Further investigation revealed the source to be rule policy in which they revealed:

> This bot's dialogue is 100% rule based so they have a lot of rules. The RulePolicy's lookup table has 4,082 keys and looping over them to determine if a rule is applicable to a state  consumes a lot of CPU cycles

@JEM-Mosig mentioned

> The OR-statements are in the bot's rule yml files. When the rules are loaded, each rule with an OR is copied - once for each alternative. RulePolicy itself doesn't know anything about OR statements. But it could maybe "compress" the rule dict again in a way that acknowledges this. I'll re-post this on the issue.

I've quickly increased cache size to a huge number to allow for rule fan out if that happens and customer tested this branch. 

What needs to be done: 

- [ ] Test memory usage
- [ ] Check if we can better utilize memory especially with regards to OR statements in rules
- [ ] We also do `list(reversed(state_list)` on each run and this can be cached too, explore this also

### DIFF
--- a/rasa/core/policies/rule_policy.py
+++ b/rasa/core/policies/rule_policy.py
@@ -835,7 +835,7 @@ class RulePolicy(MemoizationPolicy):
     @staticmethod
     # This function is called a lot (e.g. for checking contradictions) so we cache
     # its results.
-    @functools.lru_cache(maxsize=1000)
+    @functools.lru_cache(maxsize=100000)
     def _rule_key_to_state(rule_key: Text) -> List[State]:
         return json.loads(rule_key)
 


hsm207:
Customer reported that the CPU load at inference time with this fix decreased from 18% to 3%.

hsm207:
> > Customer reported that the CPU load at inference time with this fix decreased from 18% to 3%.
> 
> Did the memory usage increase?

Yes, they said it increased slightly and is acceptable. They could not quantify it though

tttthomasssss (changes requested):
I _think_ - pls correct me if I'm wrong - `maxsize` is the number of entries in the underlying lru dict. Have you checked the mem footprint when this gets full?

Can we add a test for this?

Also, can you add a changelog entry?

tayfun:
I'll close this as it is confusing people. This was meant as a quick test for a customer; we definitely need to look at it deeper, wrt. memory usage, rule policy fan out, see if we can increase performance for rule policy by other means and than maybe set a higher cache size or even make it configurable by the user.

PR #10706, opened 2022-01-19
prepare release of version 3.0.5

carlad:


ancalita (approved):

pyproject.toml
```diff
@@ -90,7 +90,7 @@ colorhash = "~1.0.2"
 jsonschema = "~3.2"
 packaging = ">=20.0,<21.0"
 pytz = ">=2019.1,<2022.0"
-rasa-sdk = "^3.0.0"
+rasa-sdk = ">=3.0.3"
```
jupyterjazz, line 93: This might be a stupid question but why are we bumping sdk version? also, how/why was the `poetry.lock` file modified? Asking this in case I'll need to do the same when I cut micro release next time.

carlad: Hey @jupyterjazz, we're bumping it because of my [recent work on aligning dependencies](https://github.com/RasaHQ/rasa-sdk/issues/592) between `rasa` and `rasa-sdk`. Especially as we've updated Black, and there are new formatting rules, it's important to tie these changes to each other so that both project use the same formatting rules moving forward.

To update the `poetry.lock` file I simply deleted it, then ran `poetry install` to create a new one.

PR #10718, opened 2022-01-21
sender_id is now kept

### LINKED ISSUE #10717: Make sender_id accessible from custom NLU components
### What problem are you trying to solve?

I need to access the sender_id from my custom NLU component, because I have an entity extractor that behaves differently depending on the user.

### What's your suggested solution?

nlu_message_converter should keep the sender_id when converting the messages.

### Examples (if relevant)

 _No response_ 

### Is anything blocking this from being implemented? (if relevant)

 _No response_ 

### Definition of Done

sender_id is accessible from a custom NLU component.

Maxinho96:
Resolves #10717

**Proposed changes**:
- nlu_message_converter now keeps the sender_id in the messages that passes to the next components


Maxinho96:
Hi @jupyterjazz , could you please assign a reviewer to this PR 🙏? It's a simple one-line fix, shouldn't take much time.

jupyterjazz:
Hi @Maxinho96, I added it to our backlog so someone should be assigned soon

Maxinho96:
Hi @jupyterjazz any news on this PR?

Maxinho96:
> Hi @Maxinho96 , sorry for the long delay in getting a reviewer. Could you describe the use case for having the `sender_id` inside a custom component? I can imagine a few but I'm wondering if this pertains to also needing other tracker data inside custom components.

Hi @melindaloubser1, thank you for the response.
I need it because I have an entity extractor that behaves differently depending on the user (different users can see different entities). It would be good to have the whole tracker available, so that I can access also the slots and possibly other data. Currently I am computing some data twice (one to populate the slot and one inside the entity extractor).

melindaloubser1:
Got it, in that case it may make more sense to add your feedback on[ this issue ](https://github.com/RasaHQ/rasa/issues/11132) which is about making it possible to create a recipe that passes the tracker to NLU components as well, rather than adding just this piece of the tracker to the message object.

Maxinho96:
> Got it, in that case it may make more sense to add your feedback on[ this issue ](https://github.com/RasaHQ/rasa/issues/11132) which is about making it possible to create a recipe that passes the tracker to NLU components as well, rather than adding just this piece of the tracker to the message object.

Cool, it would be definetely bettery, I will do it thanks.

Issue #10708, opened 2022-01-19
Upgrade Poetry to 1.1.11


tayfun:
### Rasa Open Source version

3

### Python version

3.8

### What operating system are you using?

OSX, Windows

### What happened?

Older poetry seems to fail to install some dependencies sometimes (for some OS/Python versions). @ancalita also run into this issue on Mac OS locally. 

See example CI failure in https://github.com/RasaHQ/rasa/runs/4867661448?check_suite_focus=true

This blocks https://github.com/RasaHQ/rasa/pull/10590

Definition of done:

- [ ] Update `.github/poetry_version.txt` to have 1.1.11

HotThoughts:
FYI, also opened a PR in the rasa-x repo ☝️

PR #10721, opened 2022-01-21
Upgrade poetry to 1.1.11

tayfun:
Fixes https://github.com/RasaHQ/rasa/issues/10708

We noticed some developer environments had problems with earlier Poetry
and also some community pull requests failed. Error such as:

```
ERROR: Could not install packages due to an OSError: [WinError 2] The
system cannot find the file specified:
'd:\\a\\rasa\\rasa\\.venv\\Lib\\site-packages\\nr\\__init__.py'
```

and

`'Link' object has no attribute 'is_absolute'`

**Proposed changes**:
- Upgrade poetry to 1.1.11

### DIFF
--- a/.github/poetry_version.txt
+++ b/.github/poetry_version.txt
@@ -1,2 +1,2 @@
 # The poetry version is stored in a separate file due to the https://github.com/python-poetry/poetry/issues/3316
-poetry-version=1.1.8
+poetry-version=1.1.11
--- a/docker/docker-bake.hcl
+++ b/docker/docker-bake.hcl
@@ -21,7 +21,7 @@ variable "BASE_BUILDER_IMAGE_HASH" {
 # keep this in sync with the version in pyproject.toml
 # the variable is set automatically for builds in CI
 variable "POETRY_VERSION" {
-  default = "1.1.8"
+  default = "1.1.11"
 }
 
 group "base-images" {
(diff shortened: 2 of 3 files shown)


HotThoughts:
All the tests were [skipped](https://github.com/RasaHQ/rasa/runs/4922123536?check_suite_focus=true), so I added a label to trigger them all.

HotThoughts:
Maybe also update Poetry version in [docker](https://github.com/RasaHQ/rasa/blob/36cfeee23bc6b511efd79cd00837f9dc4cd6e787/docker/docker-bake.hcl#L23-L25)? Just to keep the version is the same in docker and in the CI.

A quick question. Do we usually update `poetry.lock` when upgrading poetry? If so, we can run `poetry lock`.

tayfun:
> Maybe also update Poetry version in [docker](https://github.com/RasaHQ/rasa/blob/36cfeee23bc6b511efd79cd00837f9dc4cd6e787/docker/docker-bake.hcl#L23-L25)? Just to keep the same version everywhere.
> 
> A quick question. Do we usually update `poetry.lock` when upgrading poetry? If so, we can run `poetry lock`.

Good catch on docker, updated it now. For lock file it is not strictly needed but done that as well.

HotThoughts (approved):
🚀 Thank you. This PR is ready to go.

PR #10735, opened 2022-01-25
Update rasa-install-cron-check.yml

koaning:
This time around, I figured I'd run the workflow first.

### DIFF
--- a/.github/workflows/rasa-install-cron-check.yml
+++ b/.github/workflows/rasa-install-cron-check.yml
@@ -5,7 +5,7 @@
 name: Cron Test to Check Pip Installation
 
 on:
-  push:
+  pull_request:
     branches:
     - main
   schedule:
--- a/.github/workflows/rasa-install-cron-check.yml
+++ b/.github/workflows/rasa-install-cron-check.yml
@@ -5,6 +5,9 @@
 name: Cron Test to Check Pip Installation
 
 on:
+  push:
+    branches:
+    - main
   schedule:
     - cron: "0 0 * * *"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/rasa-install-cron-check.yml
```diff
@@ -7,6 +7,11 @@ name: Cron Test to Check Pip Installation
 on:
   schedule:
     - cron: "0 0 * * *"
+  pull_request:
+    branches:
+    - main
+    tags:
+    - '*'
```
tczekajlo:
```suggestion
```
You have to remove it, otherwise, it won't run for your PR.

koaning:
@tczekajlo I *think* this ought to be it. Once merged I'll try to manually trigger the job.

.github/workflows/rasa-install-cron-check.yml
koaning: Ah! That might've been it. Will try that now.

koaning: Mhmm. The "cronjob" still does not appear.

tczekajlo: Oh, the workflow was disabled, can you re-push your changes?

tczekajlo:
Isn't it? https://github.com/RasaHQ/rasa/runs/4934659793?check_suite_focus=true

.github/workflows/rasa-install-cron-check.yml
koaning: It seems to run just fine now.

https://github.com/RasaHQ/rasa/runs/4934904869?check_suite_focus=true

I'll reset everything so that it's just a cron job again. This 'ought to do the trick.

tczekajlo (approved):
LGTM

Issue #8531, opened 2021-04-23
`rasa test core` produces inconsistent fields in `story_report.json`


kedz:
**Rasa version**: 2.3.4

**Rasa SDK version** (if used & relevant):

**Rasa X version** (if used & relevant):

**Python version**: 3.7

**Operating system** (windows, osx, ...): osx

**Issue**:
`rasa test core` produces different fields in `story_report.json` depending on the performance of the core policy.
When the core policy is not perfect, there is a "micro avg" field. When it is perfect there is no `micro avg` field and instead there is an `accuracy` field.

**To Reproduce**
```
git clone https://github.com/RasaHQ/training-data
git clone https://github.com/RasaHQ/financial-demo
cd training-data 
git checkout -b regtest #not necessary after regtest merge
cd ../financial-demo
rasa train core -s data -c ../training-data/configs/en/core/rule-augmemo.yml --out model1
rasa test core -s tests --model model1 --out results1
rasa train core -s data -c ../training-data/configs/en/core/rule.yml --out model2
rasa test core -s tests --model model2 --out results2
```

`results1/story_report.json` should not have a `micro avg` field but have an `accuracy` field.
`results2/story_report.json` should have a `micro avg `field, but not have an `accuracy` field.

**Fix**
Report both always?

PR #10745, opened 2022-01-26
Add either accuracy or micro avg to report

markus-hinsche:
**Status Quo:**
- We use https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html to generate a report of metrics, which includes **either** `accuracy` **or** `micro avg`. 

**Problem**
- When using using Datadog, users expect a metric (like `micro avg` `F1`) to always be reported. Otherwise, they can't visualize trends
- https://github.com/RasaHQ/rasa/issues/8531

**Proposed changes**:
- Always include `micro avg` **and** `accuracy` in `classification_report`

Alternative implementation: would be to re-implement sklearn's [classification_report](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/metrics/_classification.py#L2003) with slight tweaks

**Checked:**
- Checked that now accuracy is reported when it was previously not ([notebook](https://app.datadoghq.eu/notebook/23688/m-hinsche-jan-26-2022-09-51-investigate-micro-avg-and-accuracy?start=1643180999670&range=14400000&live=true))


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

carlad (approved):

markus-hinsche:
CC @dakshvar22

rasa/model_testing.py
```diff
@@ -362,9 +363,48 @@ def get_evaluation_metrics(
     f1 = metrics.f1_score(targets, predictions, labels=labels, average="weighted")
     accuracy = metrics.accuracy_score(targets, predictions)
 
+    if output_dict:
+        report = make_classification_report_complete(report, accuracy)
+
     return report, precision, f1, accuracy
 
 
+def make_classification_report_complete(report: dict, accuracy: float) -> dict:
```
wochinge, line 372: can we please unit test this?

rasa/exceptions.py
```diff
@@ -67,3 +67,10 @@ def __str__(self) -> Text:
 
 class ActionLimitReached(RasaException):
     """Raised when predicted action limit is reached."""
+
+
+class ClassificationReportException(RasaException):
```
wochinge, line 72: `RasaException` means it doesn't get raised to sentry - as this is something which is _not_ supposed to happen it in my opinion should not subclass from `RasaException` as we want it to appear in Sentry, what do you think?

tests/test_model_testing.py
```diff
@@ -128,6 +128,189 @@ def test_get_evaluation_metrics(
...
+                    "recall": 0.5,
+                    "f1-score": 0.6666666666666666,
+                    "support": 2,
+                },
+                "accuracy": 0.8,
+            },
+        ),
+    ],
+)
+def test_make_classification_report_complete(report_in, accuracy, report_out):
```
wochinge, line 243: could you please add type annotations to the test functions?

Thanks in any case for adding the unit tests 🙌🏻

wochinge (approved):
🚀 Thank you!

PR #10761, opened 2022-01-27
Use prompt_toolkit ^2.0

carlad:
**Proposed changes**:
- addresses ongoing issue in rasa-sdk [#592](https://github.com/RasaHQ/rasa-sdk/issues/592)
- rasa-sdk has had similar dependency applied and new 3.0.4 version released


ancalita (approved):
💯

Issue #10766, opened 2022-01-28
Can't use multiple domains with entity roles or groups


melindaloubser1:
### Rasa Open Source version

2.8.22 - also happens with 3.0.5

### Python version

3.8

### What operating system are you using?

OSX

### What happened?
You can reproduce by cloning https://github.com/melindaloubser1/moodbot/tree/entity_roles_multidomain and running `rasa train --domain domains`

I have a split domain setup like this:
```
├── domains
│   ├── domain.yml
│   ├── entities.yml
│   └── intents.yml
```

In entities.yml, I have an entity with roles:
```
version: '2.0'
entities:
   - city:
       roles:
       - from
       - to
```
 When I try to train a model with these domains, it fails with the error below i.e. it cannot combine the sections because it is expecting a list, but roles are a dictionary. 
 The same error does not occur when I have no roles for my entity.

### Command / Request

```shell
rasa train --domain domains
```

### Relevant log output

```shell
Traceback (most recent call last):
  File "/Users/melinda/.pyenv/versions/rasa2/bin/rasa", line 8, in <module>
    sys.exit(main())
  File "/Users/melinda/.pyenv/versions/rasa2/lib/python3.8/site-packages/rasa/__main__.py", line 118, in main
    cmdline_arguments.func(cmdline_arguments)
  File "/Users/melinda/.pyenv/versions/rasa2/lib/python3.8/site-packages/rasa/cli/train.py", line 59, in <lambda>
    train_parser.set_defaults(func=lambda args: run_training(args, can_exit=True))
  File "/Users/melinda/.pyenv/versions/rasa2/lib/python3.8/site-packages/rasa/cli/train.py", line 91, in run_training
    training_result = train_all(
  File "/Users/melinda/.pyenv/versions/rasa2/lib/python3.8/site-packages/rasa/api.py", line 109, in train
    return rasa.utils.common.run_in_loop(
  File "/Users/melinda/.pyenv/versions/rasa2/lib/python3.8/site-packages/rasa/utils/common.py", line 296, in run_in_loop
    result = loop.run_until_complete(f)
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/Users/melinda/.pyenv/versions/rasa2/lib/python3.8/site-packages/rasa/model_training.py", line 100, in train_async
    domain = await file_importer.get_domain()
[...]

PR #10767, opened 2022-01-28
Fix multidomain with entities with roles

melindaloubser1:
**Proposed changes**:
- Merge entities in domain as list of dicts not plain list
- fixes https://github.com/RasaHQ/rasa/issues/10766

**Status (please check what you already did)**:
- [x] added some tests for the functionality
- [x] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [x] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #10768, opened 2022-01-28
Fix multidomain with entities with roles  - backport of #10767 for 2.8.x

melindaloubser1:
**Proposed changes**:
- Merge entities in domain as list of dicts not plain list
- fixes https://github.com/RasaHQ/rasa/issues/10766 for 2.8.x
- equivalent to #10767 for 2.8.x

### DIFF
--- a/tests/shared/core/test_domain.py
+++ b/tests/shared/core/test_domain.py
@@ -722,11 +722,11 @@ def test_domain_from_multiple_files():
     ]
 
     expected_entities = [
-        "pistachio_robot",
         "caramel_robot",
-        "rum_and_raisin_robot",
         "chocolate_robot",
         "other_robot",
+        "pistachio_robot",
+        "rum_and_raisin_robot",
         "strawberry_robot",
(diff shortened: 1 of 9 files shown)


tests/shared/core/test_domain.py
```diff
@@ -722,11 +722,11 @@ def test_domain_from_multiple_files():
     ]
 
     expected_entities = [
+        "pistachio_robot",
```
carlad, line 725: So... the previous version was sorted alphabetically because the list of entities returned was sorted.
I wonder if it's possible to keep this list in alphabetical order, and then do something similar to what we do with `domain_slots` below:
```
domain_entities = []
for entity in domain.entities:
        domain_slots.append(entity.name)
```

and then ` assert expected_entities == sorted(domain_entities)`

melindaloubser1: I just sorted in the assert statement since the entities are already just a list of names

carlad (approved):

Issue #2801, opened 2018-12-05
Feature request: Allow passing dynamic timezone to duckling


alexmic:
**Rasa Core version**: 0.11.12

Suppose we have a customer in Berlin (UTC+1) and we allow them to ask the following two questions:

1.  "How much have I spent 2 hours ago?"
2.  "How much have I spent between 15:00 and 16:00?"

If we configure duckling to always use UTC and the customer asks that question at 18:00, question 1 will return the correct time back (15:00 UTC, 16:00 UTC+1) but question 2 will return 15:00 UTC - 16:00 UTC, which is wrong because the customer means UTC+1.

Ducking allows `tz` on `/parse` requests which we can use to get the correct result back so is it a good idea to allow request-level component config to rasa core?

akshay2000:
@wochinge This doesn't work in custom action (unless I am missing something). If the user gives absolute time (case two above) we can blindly replace the timezone of the Duckling with timezone of the user.

However, it doesn't work for relative times. For example, "last 2 hours" will give wrong results if we replace the Duckling timezone with the user timezone.

wochinge:
@akshay2000 

> However, it doesn't work for relative times. For example, "last 2 hours" will give wrong results if we replace the Duckling timezone with the user timezone.

Yes, but that also would not work with Duckling. I mean, how should it now the timezone of the user? Only option is to ask the user I guess.

akshay2000:
> 
> 
> @akshay2000
> 
> > However, it doesn't work for relative times. For example, "last 2 hours" will give wrong results if we replace the Duckling timezone with the user timezone.
> 
> Yes, but that also would not work with Duckling. I mean, how should it now the timezone of the user? Only option is to ask the user I guess.

Hence the option to send timezone with every request. As the bot sits on the client device, it can send timezone along with sender id and message. That is what I ended up doing. Had to override a bunch of methods on a bunch of core and NLU classes though.

wochinge:
> As the bot sits on the client device, it can send timezone along with sender id and message.

That highly depends on the input channel, doesn't it?

mmalhotra:
@wochinge 
I think the statement on wit.ai site is wrong. Duckling has had the timezone feature for a really long time.

I also tried with rasa's docker image:

curl -XPOST http://0.0.0.0:5006/parse --data 'dims=time&lang=en&tz=**Asia/Kolkata**&text=tonight'
[...]

PR #10781, opened 2022-01-31
Allow passing dynamic timezone to duckling #2801

AnthonyDurussel:
**Proposed changes**:
- Allow to dynamically give timezone to the NLU duckling component using message metadata `time_zone` property.

- [ *] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [*] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)


natcohen:
Hey @wochinge, would be awesome if you could merge this! Please let us know what is missing so we can work on it. It's a feature that is much needed for duckling and hasn't been implemented (while the issue is quite old...). Thanks

rasa/server.py
```diff
@@ -1286,7 +1286,9 @@ async def parse(request: Request) -> HTTPResponse:
         try:
             data = emulator.normalise_request_json(request.json)
             try:
-                parsed_data = await app.agent.parse_message(data.get("text"))
+                parsed_data = await app.agent.parse_message(
+                    data.get("text"), data.get("metadata")
+                )
```
melindaloubser1, line 1291: The metadata should already form part of the standard `UserMessage` object that is passed all the way down to Duckling. Can you clarify why it's necessary to pass it in explicitly here?

rasa/nlu/emulators/emulator.py
```diff
@@ -29,6 +29,12 @@ def normalise_request_json(self, data: Dict[Text, Any]) -> Dict[Text, Any]:
                 _data["model"] = data["model"]
 
         _data["time"] = data["time"] if "time" in data else None
+        _metadata = {}
+        for k, v in data.items():
+            if k not in ("text", "model", "time"):
+                _metadata[k] = v
+        if _metadata:
+            _data["metadata"] = _metadata
```
melindaloubser1, line 37: I'm not sure why this is needed?

melindaloubser1 (changes requested):
Thanks for the PR! I left some comments on the PR itself, but in general I'm not clear why the addition of explicit metadata is necessary, since it's already available in  NLU components. Could you clarify your use case and also add tests for the presence and absence of this metadata key?

Issue #10765, opened 2022-01-27
Domain schema overrestricts `slots` options, making custom slots with custom keys invalid


melindaloubser1:
### Rasa Open Source version

3.0.4

### What happened?

In 2.8.x, it was possible to have a custom slot definition that took an additional key e.g. 
```
slots:
  limit:
     type: custom.slots.LimitSlot
     limit: 1000
```
As long as the custom code could process all the keys listed, the domain would be valid.
In 3.0.4, the domain schema restricts the keys allowed per slot to specific ones, and makes this format invalid.

 i.e. running `rasa data validate` with a domain including this section gives the error below
```
slots:
  limit:
    type: custom.slots.LimitSlot
    limit: 1000
    mappings:
    - type: from_entity
      entity: limit
```

### Command / Request

```shell
rasa data validate
```

### Relevant log output

```shell
YamlValidationException: Failed to validate '/Users/melinda/rasa_bots/moodbot/domain.yml'. Please make sure the file is correct and all mandatory parameters are specified. Here are the errors found during validation:
  in /Users/melinda/rasa_bots/moodbot/domain.yml:26:
      Key 'limit' was not defined. Path: '/slots/limit'
```

See the [2.x domain schema](https://github.com/RasaHQ/rasa/blob/a215ae5ef54bed0c62e25331866f97f6f2845df7/rasa/shared/utils/schemas/domain.yml#L43-L45) vs. the [3.x domain schema](https://github.com/RasaHQ/rasa/blob/e6d07d93a689c595efcd13cd03e145fef23c87d2/rasa/shared/utils/schemas/domain.yml#L43-L103)

PR #10782, opened 2022-01-31
Fix domain slot schema

melindaloubser1:
**Proposed changes**:
- Allow arbitrary keys under a slot in the domain as was allowed in 2.x
- closes https://github.com/RasaHQ/rasa/issues/10765

### DIFF
--- a/rasa/shared/utils/schemas/domain.yml
+++ b/rasa/shared/utils/schemas/domain.yml
@@ -47,6 +47,8 @@ mapping:
       regex;([A-Za-z]+):
         type: "map"
         mapping:
+          regex;([A-Za-z]+):
+            type: "any"
           influence_conversation:
             type: "bool"
             required: False
--- a/rasa/shared/utils/schemas/domain.yml
+++ b/rasa/shared/utils/schemas/domain.yml
@@ -46,9 +46,8 @@ mapping:
     mapping:
       regex;([A-Za-z]+):
         type: "map"
+        allowempty: True
         mapping:
-          regex;([A-Za-z]+):
-            type: "any"
           influence_conversation:
             type: "bool"
             required: False
--- a/tests/shared/core/test_domain.py
+++ b/tests/shared/core/test_domain.py
@@ -248,7 +248,7 @@ def test_custom_slot_type_with_custom_key():
 
 
 @pytest.mark.parametrize(
-    "domain_unkown_slot_type",
+    "domain_invalid_type_for_slot_key",
     [
     """slots:
             limit:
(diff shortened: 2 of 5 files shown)


tests/shared/core/test_domain.py
```diff
@@ -859,6 +859,10 @@ def test_load_domain_with_entity_roles_groups():
     assert "origin" in domain.roles["GPE"]
     assert "destination" in domain.roles["GPE"]
 
+def test_load_domain_with_custom_slot():
+    domain = Domain.load("data/test_domains/custom_slot_domain.yml")
+
+    assert domain.slots is not None
```
joejuzl, line 865: Shall we check that the limit param made it to the slot class too?

rasa/shared/utils/schemas/domain.yml
```diff
@@ -47,6 +47,8 @@ mapping:
       regex;([A-Za-z]+):
         type: "map"
         mapping:
+          regex;([A-Za-z]+):
```
joejuzl, line 50: Are we sure that this doesn't short-circuit the other lower one? i.e. do we still enforce that `influence_conversation` is type `bool`?

tests/shared/core/test_domain.py
melindaloubser1: oh yes sorry I never finished this test and forgot 🤦‍♀️

rasa/shared/utils/schemas/domain.yml
melindaloubser1: I'll check & also add a test that the validator yells when type is wrong

melindaloubser1: It is still enforced, added tests 👍

tests/shared/core/test_domain.py
melindaloubser1: updated

joejuzl (approved):
💯

rasa/shared/utils/schemas/domain.yml
wochinge, line 50: can't we just do `allowempty: True`? See docs here https://pykwalify.readthedocs.io/en/unstable/validation-rules.html#allowempty

melindaloubser1: Indeed! I only discovered where these docs were after  the fact so I went off of the rest of the schema, but this makes more sense

rasa/shared/utils/schemas/domain.yml
```diff
@@ -47,6 +47,8 @@ mapping:
...
+            type: "any"
```
melindaloubser1, line 51:
```suggestion
          allowempty: True
```

melindaloubser1, line 51:
```suggestion
        type: "map"
        allowempty: True
        mapping:
```

PR #10783, opened 2022-01-31
Fix `rasa test core` error with story containing entity but no user utterance text

### LINKED ISSUE #10743: Backport fixes to `rasa.core.test._get_e2e_entity_evaluation_result` to `2.8.x`
There's a [customer-affecting](https://rasa-hq.slack.com/archives/C021ZTVSY77/p1643042006084500) bug in `2.8.x` due to which, when one runs `rasa test core`, the trained Core model is also tested on extracting entities from user turns for which the actual user text isn't present, such as this test story:
```yml
- story: happy path
  steps:
  - intent: greet
    entities:
      - name
  - action: utter_greet
```
This bug is fixed in 3.0 and we should back-port the fix.

# How to reproduce the error
With the latest `2.8.x`, create a `rasa init` project and add an entity to a training story by adding
```yml
entities:
  - name
```
to the domain, adding
```yml
    - hey [sara](name)
    - hello [rasa](name)
```
to the NLU examples for intent `greet`, and changing the first training story to:
```yml
- story: happy path
  steps:
  - intent: greet
    entities:
      - name
  - action: utter_greet
```
Now, `rasa train` and `rasa test core --stories ./data`. You get this error:
```shell
  File "/Users/sams/.miniconda/envs/rasa2.8/bin/rasa", line 5, in <module>
    main()
  File "/Users/sams/rasa2.8/rasa/__main__.py", line 118, in main
  File "/Users/sams/rasa2.8/rasa/cli/test.py", line 132, in run_core_test
    test_core(
  File "/Users/sams/rasa2.8/rasa/model_testing.py", line 182, in test_core
    rasa.utils.common.run_in_loop(
  File "/Users/sams/rasa2.8/rasa/utils/common.py", line 296, in run_in_loop
  File "/Users/sams/rasa2.8/rasa/core/test.py", line 1104, in test
    evaluate_entities(
  File "/Users/sams/rasa2.8/rasa/nlu/test.py", line 864, in evaluate_entities
    aligned_predictions = align_all_entity_predictions(entity_results, extractors)
  File "/Users/sams/rasa2.8/rasa/nlu/test.py", line 1224, in align_all_entity_predictions
    aligned_predictions.append(align_entity_predictions(result, extractors))
  File "/Users/sams/rasa2.8/rasa/nlu/test.py", line 1127, in align_entity_predictions
    for t in result.tokens:
TypeError: 'NoneType' object is not iterable
```

# What's the culprit and how it's fixed in 3.0
The issue is that the Core testing code [still expects the trained Core model to be able to extract entities](https://github.com/RasaHQ/rasa/blob/957c6a7a6fd0d2e427c8d6355a62dc447fee1ae4/rasa/core/test.py#L603) even those even if the corresponding user message is not present, i.e. its `text` is `None`. In Rasa 3.0, there's [a fixed version](https://github.com/RasaHQ/rasa/blob/main/rasa/core/test.py#L620) where the Core model is not expected to extract entities if the text of the user message isn't present. It's hard to tell if this fix was put into 3.0 because of a bug report or if it was just Enable engineers who noticed this "imperfection" and improved the code while porting it from 2.8 to 3.0. In any case, this should be back-ported to 2.8 to unblock the customer.

_Note: This issue does not occur if one trains the Core model only (`rasa train core`). However, this should be considered a matter of luck. Why? It just happens that when you run `rasa train core`, the resulting model will end up with a simple `rasa.shared.nlu.interpreter.RegexInterpreter` whose `featurize_message` will return `None` [over here](https://github.com/RasaHQ/rasa/blob/957c6a7a6fd0d2e427c8d6355a62dc447fee1ae4/rasa/core/test.py#L605) and so the Core model won't be expected to predict the entities for the given `UserUttered` event. `rasa train`, on the other hand, produces a proper `rasa.core.interpreter.RasaNLUInterpreter` which on the above line returns an object like `Message(data={'text': None}, output_properties={'text'}, time=None, features=)`, which then breaks things further down the line (as per the above-pasted error message) because this message doesn't have any tokens and entity extraction shouldn't be carried out on it._

**Definition of done**
- [x] Backport this fix to 2.8.x
- [x] Release a Rasa 2.8.x micro

ancalita:
**Proposed changes**:
- Fixes #10743 

### DIFF
--- a/tests/cli/test_rasa_test.py
+++ b/tests/cli/test_rasa_test.py
@@ -277,3 +277,59 @@ def test_test_core_help(run: Callable[..., RunResult]):
     printed_help = set(output.outlines)
     for line in lines:
         assert line in printed_help
+
+
+def test_test_core_with_entities_and_no_user_utterance(
+    run_in_simple_project_with_model: Callable[..., RunResult]
+):
+    write_yaml(
+        {"pipeline": "KeywordIntentClassifier", "policies": [{"name": "TEDPolicy"}],},
+        "config.yml",
+    )
+
+    domain_with_entity_yaml = """
+    version: "2.0"
+    intents:
+    - greet
+    entities:
+    - name
+    responses:
+     utter_greet:
+      - text: "Hey There!"
+    """
+    with open('domain.yml', 'w') as domain:
+        domain.write(domain_with_entity_yaml)
+
+    nlu_yaml = """
+    version: "2.0"
+    nlu:
+    - intent: greet
+      examples: |
+      - hi [sara](name)
+      - hey
+      - hi [rasa](name)
+    """
+    with open('data/nlu.yml', 'w') as nlu:
+        nlu.write(nlu_yaml)
+
+    simple_test_story_yaml = """
+stories:
+- story: happy path
+  steps:
+  - intent: greet
+    entities:
+    - name
+  - action: utter_greet
+"""
+
+    with open("data/stories.yml", "w") as stories:
+        stories.write(simple_test_story_yaml)
+
+    with open("tests/test_story_no_utterance_with_entity.yaml", "w") as f:
+        f.write(simple_test_story_yaml)
+
+    run_in_simple_project_with_model("test", "core", "--fail-on-prediction-errors")
+    assert os.path.exists("results")
+
+    with open("results/failed_test_stories.yml", "r") as failed:
+        assert "# None of the test stories failed - all good!" == failed.read()
--- a/tests/cli/test_rasa_test.py
+++ b/tests/cli/test_rasa_test.py
@@ -312,7 +312,8 @@ def test_test_core_with_entities_and_no_user_utterance(
     with open("data/nlu.yml", "w") as nlu:
         nlu.write(nlu_yaml)
 
-    simple_test_story_yaml = """
+    simple_story_yaml = """
+version: "2.0"
 stories:
 - story: happy path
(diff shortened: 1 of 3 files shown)


tests/cli/test_rasa_test.py
```diff
@@ -277,3 +277,59 @@ def test_test_core_help(run: Callable[..., RunResult]):
...
+- story: happy path
+  steps:
+  - intent: greet
+    entities:
+    - name
+  - action: utter_greet
+"""
+
+    with open("data/stories.yml", "w") as stories:
+        stories.write(simple_test_story_yaml)
```
tayfun, line 326: Isn't this writing stories file under root rasa `data/stories.yml` file? How does this file get use with `run_in_simple_project_with_model`?

tayfun: Ignore me, this works. I always get confused about magic directories in tests.

tayfun (approved):

tests/cli/test_rasa_test.py
ancalita: I was actually following the instructions of the bug in the issue which gets reproduced if you use the training stories as input for `rasa test core --stories`. Maybe I can make this explicit in the test by removing the writing of `tests/` and giving the training data set as a parameter to `--stories`. Let me try.

tayfun (approved):

PR #10795, opened 2022-02-02
Robust check if GPUs are working

markus-hinsche:
**Problem**: We recently ran the regression tests on GPU. The result runtimes had slowed down compared to previous runs. It took a while to understand that this was because GPUs weren't used at all. We want to avoid this situation in the future and rather fail hard instead of benchmarking unintended things.

**Proposed changes**:
- Use `tf.config.experimental.list_physical_devices`  instead `device_lib.list_local_devices`
- Return exit code 1 if no GPU available -> This achieves a hard-fail instead oft a soft-fail
- Validate that CPU run does not use GPUs, otherwise hard-fail

Note: 
- I would like to unit test this. For GPU, I could write a unit test that checks the device checks. But then, these unit tests would fail when run on a GPU (e.g. of a developer owning a GPU). Let me know if you have ideas how to unit test this.


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

virtualroot (approved):

.github/scripts/validate_cpu.py
```diff
@@ -0,0 +1,14 @@
+import sys
+
+import tensorflow as tf
+
+
+def check_gpu_not_available():
+    num_gpus = len(tf.config.experimental.list_physical_devices("GPU"))
```
koernerfelicia, line 7:
```suggestion
    num_gpus = len(tf.config.list_physical_devices("GPU"))
```

This is my bad, I gave you outdated advice. I think according to docs this is no longer experimental?

2.3: https://www.tensorflow.org/versions/r2.3/api_docs/python/tf/config/list_physical_devices
2.7: https://www.tensorflow.org/api_docs/python/tf/config/list_physical_devices

.github/scripts/validate_gpus.py
```diff
@@ -0,0 +1,14 @@
+import sys
+
+import tensorflow as tf
+
+
+def check_gpu_available():
+    num_gpus = len(tf.config.experimental.list_physical_devices("GPU"))
```
koernerfelicia, line 7:
```suggestion
    num_gpus = len(tf.config.list_physical_devices("GPU"))
```
Same here

.github/scripts/validate_cpu.py
koernerfelicia: nb: haven't tested this myself, so probably worth a double-check

.github/workflows/ci-model-regression.yml
```diff
@@ -596,6 +596,11 @@ jobs:
         if: steps.set_dataset_config_vars.outputs.is_dataset_exists == 'true' && steps.set_dataset_config_vars.outputs.is_config_exists == 'true'
         run: poetry run pip install -U datadog-api-client ddtrace
 
+      - name: Validate that GPUs are not working
```
koernerfelicia, line 599:
```suggestion
      - name: CPU run: Validate that no GPUs are available
```
Nitpicking here, but something about GPUs not working seems confusing 😅

koernerfelicia:
>Let me know if you have ideas how to unit test this.

I think you can set: `CUDA_VISIBLE_DEVICES="-1"` to hide the GPU (see [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#env-vars))

koernerfelicia (approved):
Looks good to me! 🎉

PR #10808, opened 2022-02-03
Fix multidomain entity roles 2.8 -  patch

melindaloubser1:
**Proposed changes**:
- Patch the fix in https://github.com/RasaHQ/rasa/pull/10768 - there is a second case of loading that wasn't covered there


melindaloubser1:
@carlad the odd thing about this case is that it only showed up when tested with NLU data that includes retrieval intents - i.e. a plain domain load won't pick this up. 
I'm trying to find the kind of test I should update to account for this case

melindaloubser1:
Found the test! It's an importer test

carlad (approved):

PR #10813, opened 2022-02-04
Overwrite final weights with checkpoint weights before persisting

### LINKED ISSUE #10516: Checkpoint models are never loaded
### Rasa Open Source version

3.0.2

### Rasa SDK version

3.0.2

### Rasa X version

None

### Python version

3.8

OSX

### What happened?

Also relevant for older versions of Rasa. 

**Problem:**
Model checkpointing allows you to save the best model during training. We save this into a [temporary directory](https://github.com/RasaHQ/rasa/blob/main/rasa/utils/tensorflow/callback.py#L58) which is then moved into the model directory as a subdirectory (for example see this behavior [here](https://github.com/RasaHQ/rasa/blob/3a27badc4bfc939db2862f07af4b81211ecb6d4c/rasa/nlu/classifiers/diet_classifier.py#L1036-L1037) for the DIETClassifier, and [here](https://github.com/RasaHQ/rasa/blob/3a27badc4bfc939db2862f07af4b81211ecb6d4c/rasa/core/policies/ted_policy.py#L921-L922) for TEDPolicy). However, because we only load models directly from the model directory, the checkpoint model is never loaded. The expectation from our docs is that we should load it. According to the instructions listed under "More configurable parameters” table for `checkpoint_best_model` in `DIETClassifier` or `TED` which state: 

_"Save the best performing model during training. Models are stored to the location specified by --out. Only the one best model will be saved."_

The fix for this problem should include a unit test to check that we get the correct behavior.

### Command / Request

_No response_

### Relevant log output

_No response_

**Definition of done**
- [ ] Determine what is the current behavior
- [ ] Determine what is the expected behavior
- [ ] Decide if you just fix or create an issue

jupyterjazz:
**Proposed changes**:
- Fixes #10516 
- Implements overwriting the final model weights with the checkpoint weights after training of `DIETClassifier`, `ResponseSelector` and `TEDPolicy`. 

### DIFF
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -1043,10 +1043,10 @@ class DIETClassifier(GraphComponent, IntentClassifier, EntityExtractorMixin):
 
             if self.component_config[CHECKPOINT_MODEL] and self.tmp_checkpoint_dir:
                 self.model.load_weights(self.tmp_checkpoint_dir / "checkpoint.tf_model")
-                # For testing purposes only
-                rasa.utils.io.pickle_dump(
-                    model_path / f"{file_name}.from_checkpoint.pkl", None
-                )
+                # Save an empty file to flag that this model has been
+                # produced using checkpointing
+                checkpoint_marker = model_path / f"{file_name}.from_checkpoint.pkl"
+                checkpoint_marker.touch()
 
             self.model.save(str(tf_model_file))
 
--- a/rasa/nlu/classifiers/diet_classifier.py
+++ b/rasa/nlu/classifiers/diet_classifier.py
@@ -1032,8 +1032,6 @@ class DIETClassifier(GraphComponent, IntentClassifier, EntityExtractorMixin):
 
     def persist(self) -> None:
         """Persist this model into the passed directory."""
-        import shutil
-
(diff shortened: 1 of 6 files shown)


rasa/nlu/classifiers/diet_classifier.py
```diff
@@ -1043,8 +1041,13 @@ def persist(self) -> None:
 
             rasa.shared.utils.io.create_directory_for_file(tf_model_file)
 
-            if self.component_config[CHECKPOINT_MODEL]:
-                shutil.move(self.tmp_checkpoint_dir, model_path / "checkpoints")
+            if self.component_config[CHECKPOINT_MODEL] and self.tmp_checkpoint_dir:
+                self.model.load_weights(self.tmp_checkpoint_dir / "checkpoint.tf_model")
+                # For testing purposes only
```
usc-m, line 1046: Could we change this comment to describe the purpose of the file more, I think it'll be unclear if we revisit this code why we're saving this file. Maybe something like "Save an empty file to flag that this model has been produced using checkpointing".

You could also use something like `touch` to do this (https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch) which might be a bit more elegant than pickling `None`

usc-m (approved):

PR #10834, opened 2022-02-08
Split downloading pretrained models from model runtime via auto classes

markus-hinsche:
**Status quo**: Currently the regression test step `Run Test` downloads pretrained models and trains in the same step. The metrics `train_run_time` and `total_run_time` are therefore heavily dependent for the speed of downloading. 

**Proposed changes**:
- Download the [pretrained Huggingface models](https://github.com/RasaHQ/rasa/blob/main/data/test/hf_transformers_models.txt) for the [LanguageModelFeaturizer](https://rasa.com/docs/rasa/components/#languagemodelfeaturizer) before `rasa train`
- Split rasa train: into downloading in the step before and training without downloading
- Implementation: Use transformers `AutoTokenizer` and `TFAutoModel`

**Future work could be:**
- Disable ability to download during measured `Run Test` step (to ensure downloads happened before)
- Add a metric containing `download_pretrained_time`


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

.github/scripts/download_pretrained.py
```diff
@@ -0,0 +1,83 @@
...
+    print(config)
+    steps = config.get("pipeline", [])
+
+    # Look for LanguageModelFeaturizer
+    steps = list(filter(lambda x: x["name"] == COMP_NAME, steps))
+
+    if len(steps) == 0:
+        print(f"No {COMP_NAME} found")
+        return None, None
+    elif len(steps) > 1:
```
mleimeister, line 27: I haven't seen any example of this in an actual config, but in theory it would be allowed to have two `LanguageModelFeaturizer` components in there I think. In this case, all specified models could be downloaded. Not sure what an actual use case would be though.

.github/scripts/download_pretrained.py
```diff
@@ -0,0 +1,83 @@
...
+    model_name = lmfeat_step["model_name"]
+
+    model_weights = lmfeat_step.get("model_weights", model_weights_defaults[model_name])
+
+    return model_name, model_weights
+
+
+def instantiate_to_download(model_weights: Text) -> None:
+    """Instantiates Auto class instances, but only to download."""
+    _ = AutoTokenizer.from_pretrained(model_weights)
```
mleimeister, line 44: Afaik, this will download to the default cache used by `transformers`. We allow to specify the `cache_dir` in the config though, in which case the model won't be found. To prevent that, the `cache_dir` could be extracted from the config as well and be set in the `from_pretrained` methods, see e.g. [here](https://github.com/huggingface/transformers/blob/v4.16.2/src/transformers/tokenization_utils_base.py#L1555)

.github/scripts/download_pretrained.py
```diff
@@ -0,0 +1,83 @@
...
+    if "model_name" not in lmfeat_step:
+        return DEFAULT_MODEL_NAME, model_weights_defaults[DEFAULT_MODEL_NAME]
+    model_name = lmfeat_step["model_name"]
+
+    model_weights = lmfeat_step.get("model_weights", model_weights_defaults[model_name])
+
+    return model_name, model_weights
+
+
+def instantiate_to_download(model_weights: Text) -> None:
```
mleimeister, line 42: Depending on the `model_weights`, this function could throw Exceptions. Should those be caught so that the script can exit gracefully? The two failures I can think of are

1. The model doesn't exist, e.g. `model_weights = "abc"`, then `AutoTokenizer.from_pretrained` will raise a `requests.exceptions.HTTPError`.
2. The model exists, but has not Tensorflow weights, then `TFAutoModel.from_pretrained` will raise an `OSError`.

markus-hinsche: Good catch 💪

I adjusted the code to be closer to https://github.com/RasaHQ/rasa/blob/main/rasa/nlu/featurizers/dense_featurizer/lm_featurizer.py#L119-L123.
Also adjusted/added test cases.

Some errors, like `OSError` are not handled, because I think the error msg + stack trace leads the developer/user to the mistake.

.github/scripts/download_pretrained.py
markus-hinsche: I adjusted the code to a consider a list of `LanguageModelFeaturizer`s to loop over.

.github/scripts/download_pretrained.py
mleimeister: 👍  Ah yes, very good point. Seeing those errors in the logs actually can help fixing wrong configs much easier.

.github/workflows/ci-github-actions.yml
```diff
@@ -58,6 +58,7 @@ jobs:
 
     - name: Install Dependencies 📦
       run: |
+        poetry install --extras full
         make install
```
mleimeister, line 62: Could those 2 lines be replace by `make install-full`?

mleimeister (approved):

PR #10841, opened 2022-02-08
Improve Domain Loading: Resolve Technical Debt

### LINKED ISSUE #10807: Improve domain loading: Tech debt and associated bugs
# Background
A few bugs were identified around the merging of multiple domain files.
While fixing these bugs relatively quickly we accrued some technical debt. 
After further investigation into how domain loading and merging works some inconsistencies and bugs were found. This ticket encompasses the tech debt and bug fixes, but not the behaviour changes. As the tech debt and bug are very related, it makes sense to do them in one go.
More details: https://www.notion.so/rasa/Improve-domain-loading-b3998eafc5f7406dab0be88613dfd15d

# Overview
## Bugs

- Domain merging fails when there is a dictionary entity.
    - This can be fixed by using the same logic currently used for intents
- If a directory is passed to Domain.load it will use Domain.merge_domain_dicts, if a list of paths is passed it will use Domain.merge(which behaves differently). (And it can end up being a mix of both i.e. if one of the paths is a directory).
    - This should be addressed by the tech debt below.

## Tech debt
- We should combine `Domain.merge_domain_dicts` and `Doman.merge` into one method. (Or at least share the core functionality)
- We call instance method `merge_domain_dicts` on a class. This should be fixed.
- Internal methods such as `merge_dicts` should not be public. We are restricted by and have to support the public interfaces of classes like `Domain`.
    - Can some of these utility methods be moved into a shared utility package?
- We store `self.duplicates` on the `Domain` instance. Can we just warn as we merge/load and then raise `InvalidDomain` without storing this data as a public instance variable?
    - Currently `domain.duplicates` is checked later by `Validator.verify_domain_duplicates` . Maybe we can remove this and just do it in the `Domain` class
- `Domain._check_domain_sanity` appears to duplicate some of the duplication checks - can we remove/combine it?
- Lots of the difficulty of merging domains comes from the fact that we change the internal representation e.g. `_transform_intent_properties_for_internal_use`. This is difficult to revert. Instead can we always keep the original domain dict on the instance, and use that for both domain persistence and merging?
    - We need to confirm that the domain is not mutated in anyway that would render the original dict out-of-date or invalid.

# Definition of Done
- [x] sync with @usc-m to make sure everyone is aware of related ongoing work
- [x] Dictionary entities work with multiple domains
- [x] There is only one domain merging method
- [x] Merging utility methods are not public
- [x] Domain validation happens in one place and time
- [x] Domains dump using their initial dictionary representation (i.e. it is not transformed then transformed back again).

ancalita:
**Proposed changes**:
- Closes #10807 

> Attempting to combine `Domain.merge` and `Domain.merge_domain_dicts` used for domain directories breaks tests. This is because `merge` uses 2 loaded Domain instances that have already done quite a lot of clean-up of attributes, re-using `merge` in `from_directory` loading method doesn't copy entirely the subsequent domain files. Instead I extracted the shared functionality into the `utils` package.
- Edit:  There is now a single `merge` method that returns a dict and the original `data` representation is saved as an attribute on the Domain instance.

- The unit test `test_domain.py::test_domain_from_multiple_files` uses a domain directory that contains dictionary entities, the unit test passes in the context of current changes.
- I opted for raising a `UserWarning` when duplicates are found in different domain files instead of `InvalidDomain`, otherwise this would cause unit tests such as `test_domain.py::test_merge_yaml_domains` to break because it's considered that the `override` parameter will dictate how conflicts are resolved.
- The validation done by `Domain._check_domain_sanity` is different from the check for duplicates before merging, because it checks that the final Domain instance doesn't contain repetitive actions, slots or entities. I've decided to keep these separate.
- I've explored if we can use the original data dict representation for persistence and merging( the `self.data` property) - it can be modified to be used in fingerprinting, however it causes problems down the line when replacing the `as_dict`method in `cleaned_domain` method, lots of tests break because `self.data` format is not quite what `cleaned_domain` expects; and also some other tests such as `test_slots_states_before_user_utterance`, `test_create_train_data_no_history` break too.


raoulvm:
Hi @ancalita 
I really like the clean-up here. As a follow-up I suggest (maybe for 3.1 as it is somewhat breaking) to make the merging _fail_ if override is not True. In our quite large team we encountered duplicates almost *always* to be errors - different people creating the same intents, entities,... you name it, in different files while working on different topics. Failing would give them a hard heads-up to get in contact with their colleagues and sort it out! It's no "I might look later into that warning" any more, then.

rasa/shared/utils/domain_loading.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def as_dict(self) -> Dict:
+        """Return serialized `SessionConfig`."""
+        return {
+            "session_expiration_time": self.session_expiration_time,
+            "carry_over_slots_to_new_session": self.carry_over_slots,
+        }
+
+
+def extract_duplicates(list1: List[Any], list2: List[Any]) -> List[Any]:
```
joejuzl, line 45: I would either keep this stuff in the `Domain` class (as private methods) if it's specific to domains or put it in a general utility file if it can/will be used elsewhere (unit-tested independently). I don't think we need a `domain_loading` utility module.

rasa/shared/core/domain.py
```diff
@@ -320,127 +258,17 @@ def merge_domain_dicts(
...
-        dict2 = {
-            (sorted(list(i.keys()))[0] if isinstance(i, dict) else i): i
-            for i in dict_list2
-        }
-        merged_dicts = Domain.merge_dicts(dict1, dict2, override_existing_values)
-        return list(merged_dicts.values())
+    @property
+    def data(self) -> Dict:
+        """Returns original domain dict representation."""
+        self._data.update(self.as_dict())
```
joejuzl, line 270: Why the `update`?

rasa/shared/core/domain.py
```diff
@@ -251,64 +229,24 @@ def from_directory(cls, path: Text) -> "Domain":
...
-            combined[key] = self.merge_dicts(combined[key], domain_dict[key], override)
-
-        return self.__class__.from_dict(combined)
-
-    def merge_domain_dicts(
-        self, domain1: Dict, domain2: Dict, override: bool = False
+    @classmethod
+    def merge(
+        cls,
+        domain1: Dict,
```
joejuzl, line 240: Is there a reason we decided to use `Dict` instead of `Domain` objects here?

rasa/shared/core/domain.py
```diff
@@ -251,64 +229,24 @@ def from_directory(cls, path: Text) -> "Domain":
...
-            combined[key] = self.merge_lists(combined[key], domain_dict[key])
-
-        for key in [KEY_FORMS, KEY_RESPONSES, KEY_SLOTS]:
-            combined[key] = self.merge_dicts(combined[key], domain_dict[key], override)
-
-        return self.__class__.from_dict(combined)
-
-    def merge_domain_dicts(
-        self, domain1: Dict, domain2: Dict, override: bool = False
+    @classmethod
```
joejuzl, line 237: I like this `classmethod` as the core merging function - I wonder if it would also be useful to have a lightweight `merge` instance method that just calls this. That way we could still do stuff like `domain.merge(other)`?

rasa/shared/core/domain.py
ancalita: The original domain dict could be missing mandatory keys such as `config` and `session_config` that get defaults during initialization - and the current utils function `merge_domain_dicts` expects that the domain that gets merged to the final/combined one has these config keys filled with defaults or otherwise. 

I will scrap using the `as_dict` and instead update only with the config values.

rasa/shared/core/domain.py
ancalita: I thought the steps where `Domain` objects get converted into dicts (L273-274 in the original file) are redundant and instead decided to use dicts directly.

rasa/shared/core/domain.py
ancalita: @joejuzl Seems this change actually breaks tests(see [latest CI run](https://github.com/RasaHQ/rasa/runs/5188385918?check_suite_focus=true)). I'm thinking I should actually remove the `data` property from the Domain since it can't be used in merging either, wdyt?

joejuzl: Just because it breaks tests doesn't mean it necessarily the wrong approach. Unit tests are often coupled somewhat to implementation - is there any specific _behaviour_ we can not replicate by using original data representation rather than converting back during `as_dict`?

ancalita: I've identified the problem that causes the retrieval intent related tests to fail - in the `ResponsesSyncImporter` code, the domain with retrieval intents that gets merged gets initialised with empty `data` dict (my bad!)

I'm trying to modify this `data` parameter to actually include intents and responses, however it's proving quite difficult since it seems the Domain does quite a few transformations in `as_dict` to handle these type of intents and responses. I'll continue working on this for 1-2 more hours 🤞🏼

ancalita: Ah found the trick! Setting override to `True` when merging the retrieval domain makes sure that the combined domain gets the data in the right format.

rasa/shared/utils/domain_loading.py
ancalita: Done 👍🏼 Seems the functions I moved into `common.py` have 100% test coverage, do you think I should still include some unit tests in `test_common.py`?

joejuzl:
So can we now use `self.data` instead of what's in `as_dict` and `cleaned_domain`?

rasa/shared/utils/domain_loading.py
joejuzl: Yeah I still think it's worth unit testing them (at least a bit ) as they live alone now.

ancalita:
> So can we now use `self.data` instead of what's in `as_dict` and `cleaned_domain`?

For writing `Domain` to yaml again only, I can replace [this line ](https://github.com/RasaHQ/rasa/blob/f6f8d0941276e4302a74a5e99305742da3b1b132/rasa/shared/core/domain.py#L1616)with `domain_data.update(self.data)` and no tests would fail. 
However `as_dict` gets used in many different parts of the codebase and replacing that with `self.data` could have unintended consequences? 🤔 
For example, some of the areas it gets used in includes `server.py`, `action.py`, `forms.py`, `storage.py`, `DomainForCoreTrainingProvider`.

joejuzl:
> However as_dict gets used in many different parts of the codebase and replacing that with self.data could have unintended consequences? 🤔
For example, some of the areas it gets used in includes server.py, action.py, forms.py, storage.py, DomainForCoreTrainingProvider.

I think most of those places are just looking for a serialised version of the domain right? (apart from `forms`, however that could probably be changed to use `domain.slots`.)

The main motivation for keeping the original data representation was to avoid all the difficult logic of getting back to that representation that we have to do in `as_dict`. Before we give up on this I would like to know if there is any fundamental reason we can't change to this way of doing things (other than having to fix some unit tests).

ancalita:
> Before we give up on this I would like to know if there is any fundamental reason we can't change to this way of doing things

I will explore a bit more with making this change and seeing what fails and why - my initial impressions were that the test failures are more about expected behaviour (for example, having slots in a certain order), and that this means the required changes go beyond fixing tests.

rasa/engine/storage/storage.py
```diff
@@ -154,7 +154,7 @@ def __post_init__(self) -> None:
     def as_dict(self) -> Dict[Text, Any]:
         """Returns serializable version of the `ModelMetadata`."""
         return {
-            "domain": self.domain.as_dict(),
+            "domain": self.domain.data,
```
joejuzl, line 157: I think we should keep `domain.data` is a private instance variable (i.e. `._data`) and still access it through `as_dict()`

rasa/shared/core/domain.py
```diff
@@ -252,195 +255,191 @@ def from_directory(cls, path: Text) -> "Domain":
...
+        return combined
 
-        else:
-            merged_dicts, b = tempDict2.copy(), tempDict1.copy()
-        merged_dicts.update(b)
-        return merged_dicts
+    @property
+    def data(self) -> Dict:
+        """Returns original domain dict representation."""
+        # add the config, session_config and e2e_actions defaults
```
joejuzl, line 394: Can we do all of this in either the `__init__` or `from_dict` methods so it's clear in the object lifecycle when it happens?

rasa/shared/core/domain.py
```diff
@@ -252,195 +255,213 @@ def from_directory(cls, path: Text) -> "Domain":
...
-
-        if not domain1:
-            return domain2
-
-        domain_dict = domain2
-        combined = domain1
+        if not combined.get(KEY_TRAINING_DATA_FORMAT_VERSION):
+            combined[
+                KEY_TRAINING_DATA_FORMAT_VERSION
+            ] = LATEST_TRAINING_DATA_FORMAT_VERSION
```
joejuzl, line 349: Don't we do this also in `_add_default_keys_to_domain_dict`?

rasa/shared/core/domain.py
```diff
@@ -252,195 +255,213 @@ def from_directory(cls, path: Text) -> "Domain":
...
         if override:
             config = domain_dict["config"]
             for key, val in config.items():
                 combined["config"][key] = val
 
-        if override or domain2.get("session_config"):
+        if (
+            override
+            or (
+                not is_dir
```
joejuzl, line 359: Why do we care about `is_dir` here?

rasa/shared/core/domain.py
```diff
@@ -252,195 +255,213 @@ def from_directory(cls, path: Text) -> "Domain":
...
-                domain_dict[key] = domain_dict.get(key, [])
-                combined[key] = self.merge_lists_of_dicts(
-                    combined[key], domain_dict[key], override
-                )
+        combined, duplicates = Domain._merge_intents_and_entities(
+            combined, domain_dict, duplicates, override
+        )
 
         # remove existing forms from new actions
         for form in combined.get(KEY_FORMS, []):
```
joejuzl, line 374: Do we still need to do this?

rasa/shared/core/domain.py
```diff
@@ -252,195 +255,213 @@ def from_directory(cls, path: Text) -> "Domain":
...
-        return list(merged_dicts.values())
+    def _sanitize_intents_in_domain_dict(data: Dict) -> Dict:
+        for intent in data.get(KEY_INTENTS, []):
+            if isinstance(intent, dict):
+                for intent_name, intent_property in intent.items():
+                    try:
+                        if not intent_property[
+                            USE_ENTITIES_KEY
+                        ]:  # this covers False, None and []
+                            intent_property[USE_ENTITIES_KEY] = []
```
joejuzl, line 456: Why do we have to do this? And what about `IGNORE_ENTITIES_KEY`?

rasa/shared/importers/importer.py
```diff
@@ -376,13 +380,22 @@ def _get_domain_with_retrieval_intents(
...
+
         return Domain(
             retrieval_intent_properties,
             [],
             [],
             responses,
-            ResponsesSyncImporter._construct_retrieval_action_names(retrieval_intents),
+            action_names,
             {},
+            {
```
joejuzl, line 394: again, maybe `from_dict` would work better here.

rasa/shared/core/domain.py
```diff
@@ -1466,84 +1487,6 @@ def get_responses_with_multilines(
...
-            elif entity in self.groups:
-                entities_for_file.append(
-                    {entity: {ENTITY_GROUPS_KEY: self.groups[entity]}}
-                )
-            else:
-                entities_for_file.append(entity)
-
-        return entities_for_file
-
     def cleaned_domain(self) -> Dict[Text, Any]:
```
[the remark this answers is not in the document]
ancalita: Probably not 🤔 It appears useful only insofar as default keys get removed from the domain. I'll look now to remove it.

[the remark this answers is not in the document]
ancalita: Having had a look at this, `cleaned_domain` is useful mainly for removing the defaults that get added, e.g. `e2e_actions: [ ]` etc. I can reduce the scope of this function only to removing keys that have empty values, would that be ok?

rasa/shared/constants.py
```diff
@@ -99,3 +99,34 @@
 
 RESPONSE_CONDITION = "condition"
 CHANNEL = "channel"
+
+CARRY_OVER_SLOTS_KEY = "carry_over_slots_to_new_session"
```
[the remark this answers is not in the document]
ancalita: Oh yes, I placed them here to avoid circular import in the first commits when I created that `domain_loading.py` utils module, I can move them back to `domain.py`.

rasa/shared/core/domain.py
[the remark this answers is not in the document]
joejuzl: Do we need the defaults in `self._data` in the first place?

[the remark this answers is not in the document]
ancalita: Yes, they get added because of those chained importers requiring the default values for certain keys.

rasa/shared/core/domain.py
ancalita: Ah yes, this now becomes redundant. It was in the original code before any PR changes, I will remove it.

rasa/shared/core/domain.py
ancalita: In the original code, the 2 merge functions for loading a file and loading a directory differed in the 2nd part of this check `if override or domain2.get / combined.get` so in order to keep this difference i had to introduce this boolean `is_dir`.

rasa/shared/core/domain.py
ancalita: There are 2 unit tests that fail without this part in `test_domain`: `test_load_intents_from_as_dict_representation` which I've adapted, and `test_clean_domain_for_file`. I could add a similar check for `IGNORE_ENTITIES_KEY`, maybe it got missed from unit testing.

rasa/shared/core/domain.py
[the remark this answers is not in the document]
ancalita: Aha, think I got to the bottom of why I actually had to add all the extra keys, they were added during the merging process and `rasa train` unit tests would fail when comparing the metadata.domain with the loaded domain from file. Making some amendments now to sort this out now 😄

rasa/shared/core/domain.py
ancalita: Since this was part of the original code, I'd say so - the duplicate check is only between the same keys, it won't check for duplicate form names in the other domain's actions.

tests/examples/test_example_bots_training_data.py
```diff
@@ -80,8 +79,7 @@ def test_example_bot_training_on_initial_project(tmp_path: Path):
         importer.get_nlu_data()
         importer.get_stories()
 
-    # 2 for slot auto-fill removal warning
-    assert len(record) == 2
+    assert len(record) == 13
```
ancalita, line 82: @joejuzl For some reason that I don't quite understand, after yesterday's commits, the number of slot auto-fill removal `UserWarnings` have increased, I suppose it's because I'm using `Domain.from_dict()` in `Domain.empty`? Let me know if these test amendments are fine.

rasa/shared/core/domain.py
joejuzl: But what is the behavioural reason? I.e. what does this actually accomplish?

rasa/shared/core/domain.py
```diff
@@ -254,195 +270,183 @@ def from_directory(cls, path: Text) -> "Domain":
...
-        for key in [KEY_FORMS, KEY_RESPONSES, KEY_SLOTS]:
-            duplicates[key] = self.extract_duplicates(
+        for key, merge_func in merge_func_mappings.items():
+            duplicates[key] = rasa.shared.utils.common.extract_duplicates(
                 combined.get(key, []), domain_dict.get(key, [])
             )
-            combined[key] = self.merge_dicts(
-                combined.get(key, {}), domain_dict.get(key, {}), override
+
+            if key in [KEY_FORMS, KEY_RESPONSES, KEY_SLOTS]:
```
joejuzl, line 356:
```suggestion
            if merge_func == rasa.shared.utils.common.merge_dicts:
```
?

tests/examples/test_example_bots_training_data.py
joejuzl: 13 sounds like a lot. Can you run the test and see if the warnings actually make sense?

joejuzl (commented):
Looking good! I think we are close 😀

tests/examples/test_example_bots_training_data.py
ancalita: Yeah, I think they appear because we're instantiating domains with `from_dict` in many more places where we were using the `cls` constructor (for example, in `importers.py`) and that slot auto-fill removal warning appears only because of the slot validation that happens in `from_dict` method.

joejuzl: Ok, well we don't want duplicate warnings, so we need to find a way to avoid that.

ancalita: What about this solution: in `from_dict`, add an if statement to run `validate_slot_mappings` only if there are slots defined in the domain, this way we avoid raising unnecessary warnings?

rasa/shared/core/domain.py
ancalita: Refactored to make it clear what is the desired behaviour for overriding SESSION_CONFIG_KEY with the key from the domain that gets merged.

ancalita:
@joejuzl confirming this works with a domain directory provided to `rasa train`: I split the `moodbot` domain into multiple files in a separate directory and added some random entities too, and it worked fine. `rasa shell` works as expected.

joejuzl (approved):
🚀 🚀 🚀

PR #10899, opened 2022-02-15
Add entity `influence_conversation` flag

### LINKED ISSUE #10852: Allow global `influence_conversation` per entity instead of per intent
### What problem are you trying to solve?

Depending on the use case, it can be that only certain **entities** should influence the conversation ever, as opposed to entities influencing the conversation only for certain **intents**.
In this situation, it is unwieldy to specify entities to ignore/use for every intent. It is also easy to accidentally leave out an ignored entity for some intent, leading to unintended behaviour.

e.g. let's say entities `fox, goat, hare, ibex, jaguar, ` should always be ignored, but the others are used for some intents. That means you'll have to list up to half of all the entities for each intent to get the right result:
```
entities:
- apple
- bear
- canoe
- digit
- eel
- fox
- goat
- hare
- ibex
- jaguar

intents:
- greet: 
   use_entities: 
   - apple
   - bear
   - canoe
   - digit
   - eel
- bye:
   ignore_entities:
  - fox
  - goat
  - hare
  - ibex
  - jaguar
```

In a real use case, there could be dozens of entities and intents, leading to it being very hard to read over the list of intents without getting lost in the lists of entities.

### What's your suggested solution?

Add a global `influence_conversation` flag under each entity (similarly as for slots), which adds the entity to all `ignore_entities` lists for all intents when `false`. 
Let it default to `true` to maintain current behaviour.
The flags `ignore_entities` and `use_entities` continue to work as before, but specifying `use_entity` with an entity that is `influence_conversation: false` should error out. 

E.g.
```
entities:
- apple
- bear
- canoe
- digit
- eel
- fox:
   influence_conversation: false 
- goat
   influence_conversation: false 
- hare
  influence_conversation: false 
- ibex
  influence_conversation: false 
- jaguar
  influence_conversation: false 

intents:
- greet
- bye
```

### Is anything blocking this from being implemented? (if relevant)

No.

### Definition of Done

- [ ] flag added
- [ ] previous default behaviour not changed
- [ ] throw error when illogical combinations are specified
- [ ] add/update tests
- [ ] update documentation

melindaloubser1:
**Proposed changes**:
- Add an `influence_conversation` flag to entities which makes it possible to ignore an entity for all intents without listing it for all intents
- closes #10852 

- [x] bump training data version (references changed to constant throughout codebase in a [separate PR](https://github.com/RasaHQ/rasa/pull/10909))


rasa/shared/core/domain.py
```diff
@@ -564,7 +567,7 @@ def retrieval_intents(self) -> List[Text]:
     @classmethod
     def collect_entity_properties(
         cls, domain_entities: List[Union[Text, Dict[Text, Any]]]
-    ) -> Tuple[List[Text], Dict[Text, List[Text]], Dict[Text, List[Text]]]:
+    ) -> Tuple[List[Text], Dict[Text, List[Text]], Dict[Text, List[Text]], List[Text]]:
```
ancalita, line 570: The docstring is missing info on the 4th returned element (the list of default ignored entities).

rasa/shared/core/domain.py
```diff
@@ -721,11 +741,18 @@ def __init__(
...
                 restarted at the end of a session.
         """
-        self.entities, self.roles, self.groups = self.collect_entity_properties(
-            entities
-        )
+        (
+            self.entities,
+            self.roles,
+            self.groups,
+            self.default_entity_featurization,
```
ancalita, line 748: I would actually label the attribute as `self.default_ignored_entities` the same as you used as parameter in the different method signatures; it's clearer what it contains.

rasa/shared/core/domain.py
```diff
@@ -617,6 +626,7 @@ def collect_intent_properties(
         entities: List[Text],
         roles: Dict[Text, List[Text]],
         groups: Dict[Text, List[Text]],
+        default_ignored_entities: List[Text],
```
ancalita, line 629: I'm wondering if there's a way we can refactor to avoid adding more parameters to existing functions (in case in the future we want to have more keys in entities or other sections of the domain, functions with over 3 args can become easily unmanageable and unreadable).

One suggestion off the top of my head would be to have a dataclass called `EntityProperties` that contains the 4 parameters as attributes:
```
from dataclasses import dataclass

@dataclass
class EntityProperties:
    """Class for keeping track of the defined entities in the domain."""
    entities: List[Text]
    roles: Dict[Text, List[Text]]
    groups: Dict[Text, List[Text]]
    default_ignored_entities: List[Text]
```
Then `collect_entity_properties` would return an instance of this data class which then gets used as parameter everywhere else?

ancalita (commented):
🚀 
Left a few suggestions for refactoring - I'm also wondering if we need a unit test to make sure that stories are not influenced by the unfeaturized entity defined through the new flag?

rasa/shared/core/domain.py
melindaloubser1: Ah ja that makes sense I think this was an accidental leftover from a first attempt

rasa/shared/core/domain.py
melindaloubser1: Makes sense, I'll try it out and see what the impact will be elsewhere

rasa/shared/core/domain.py
melindaloubser1: Resolved as part of making it a dataclass

rasa/shared/core/domain.py
melindaloubser1: Alrighty, works well enough! I notice that `intent_properties` is similar - considered refactoring it in the same way, but it seems overkill since the only preserved property is `used_entities` in the end.

rasa/shared/core/domain.py
```diff
@@ -1524,6 +1542,11 @@ def intents(self) -> List[Text]:
         """Returns sorted list of intents."""
         return sorted(self.intent_properties.keys())
 
+    @rasa.shared.utils.common.lazy_property
+    def entities(self) -> List[Text]:
```
ancalita, line 1546: We also need a setter method for entities, as highlighted by this test failure `test_check_for_unseen_feature`.

rasa/shared/core/domain.py
```diff
@@ -721,11 +738,9 @@ def __init__(
             session_config: Configuration for conversation sessions. Conversations are
                 restarted at the end of a session.
         """
-        self.entities, self.roles, self.groups = self.collect_entity_properties(
-            entities
-        )
+        self.entity_properties = self.collect_entity_properties(entities)
```
ancalita, line 741: Optional (mainly thinking out loud): If this attribute doesn't require access outside `Domain`, should we make it private?

ancalita (commented):
Looks good 💯 Added few minor suggestions, the only one to look into is actually the setter method for `Domain.entities`.

rasa/shared/core/domain.py
melindaloubser1: If it's the only place it's set from outside, I'm wondering if it's not better to change the test - the parallel intents property is not settable from outside and in general seems better not to make a derived property settable

rasa/shared/core/domain.py
melindaloubser1: I'm not that familiar with the benefits of private methods in python and I'm only seeing very limited (mostly representation related) private methods in the code base.  Making this one private would imply many others should be too, which might be, but probably out of scope for this pr.

ancalita: I was actually suggesting to transform this into a private attribute preceded by a single underscore, which only signals to another developer that this shouldn't be accessed or set outside the Domain class.  There are many examples through the codebase of such attributes. If this was preceded by double underscore, then it would be strongly protected from being modified.

ancalita (approved):
Great stuff 🙌🏼

rasa/shared/core/domain.py
melindaloubser1: Ah! Sorry I misread and went looking for private methods instead of attributes. That does make sense.

PR #10906, opened 2022-02-16
Regr Tests: Add GPU NVML Interval

markus-hinsche:
After a [small investigation](https://app.datadoghq.eu/notebook/24825/interval-of-gpu-nvml-metrics-measurement), we see that it is possible to see more fine-grained GPU measurements. The datadog default interval was 15 seconds, in this PR we can set it to 1 second.

**Proposed changes**:
- Add `NVML_INTERVAL_IN_SEC` param to `start_dd_agent.sh`
- The `NVML_INTERVAL_IN_SEC` param can be set as a global env variable in regression tests

**Future work:**
- Consider putting parameter into `/modeltest` comment (to improve easy of use)

**Question for reviewers**
- What should the default interval be?

### DIFF
--- a/.github/workflows/ci-model-regression-on-schedule.yml
+++ b/.github/workflows/ci-model-regression-on-schedule.yml
@@ -12,6 +12,7 @@ env:
   TF_FORCE_GPU_ALLOW_GROWTH: true
   GITHUB_ISSUE_LABELS: '["type:bug :bug:", "tool:model-regression-tests"]'
   PERFORMANCE_DROP_THRESHOLD: -0.05
+  NVML_INTERVAL_IN_SEC: 1
 
 jobs:
   read_test_configuration:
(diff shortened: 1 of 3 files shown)


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

.github/workflows/ci-model-regression-on-schedule.yml
```diff
@@ -12,6 +12,7 @@ env:
   TF_FORCE_GPU_ALLOW_GROWTH: true
   GITHUB_ISSUE_LABELS: '["type:bug :bug:", "tool:model-regression-tests"]'
   PERFORMANCE_DROP_THRESHOLD: -0.05
+  NVML_INTERVAL_IN_SEC: 1
```
tczekajlo, line 15: Do you need to collect it every second?

markus-hinsche: Depends on the use case. One user reported this would be helpful, but other users might not need this granularity.
Do you think a short default interval would have disadvantages?

tczekajlo: Rather I was wondering what is the use case that the default doesn't work, but if it's going to help then I'm ok with it.

tczekajlo (approved):
LGTM

PR #10909, opened 2022-02-16
use constant for training data version everywhere

melindaloubser1:
**Proposed changes**:
- Reference the constant `LATEST_TRAINING_DATA_FORMAT_VERSION` everywhere instead of hardcoding `"3.0"`. This makes it simpler to bump the training data version in the future (will be required soon).

### DIFF
--- a/tests/cli/test_rasa_data.py
+++ b/tests/cli/test_rasa_data.py
@@ -9,6 +9,7 @@ from typing import Callable, Text
 from _pytest.monkeypatch import MonkeyPatch
 from _pytest.pytester import RunResult
 from rasa.cli import data
+from rasa.shared.constants import LATEST_TRAINING_DATA_FORMAT_VERSION
 from rasa.shared.importers.importer import TrainingDataImporter
 from rasa.validator import Validator
 import rasa.shared.utils.io
@@ -156,7 +157,7 @@ def test_validate_files_action_not_found_invalid_domain(
     file_name = tmp_path / f"{file_type}.yml"
     file_name.write_text(
         f"""
-        version: "3.0"
+        version: "{LATEST_TRAINING_DATA_FORMAT_VERSION}"
         {file_type}:
         - {data_type}: test path
           steps:
@@ -183,7 +184,7 @@ def test_validate_files_form_not_found_invalid_domain(
     file_name = tmp_path / f"{file_type}.yml"
     file_name.write_text(
         f"""
-        version: "3.0"
+        version: "{LATEST_TRAINING_DATA_FORMAT_VERSION}"
         {file_type}:
         - {data_type}: test path
           steps:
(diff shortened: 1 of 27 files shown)


tests/cli/test_rasa_data.py
```diff
@@ -156,7 +157,7 @@ def test_validate_files_action_not_found_invalid_domain(
     file_name = tmp_path / f"{file_type}.yml"
     file_name.write_text(
         f"""
-        version: "3.0"
+        version: "{LATEST_TRAINING_DATA_FORMAT_VERSION}"
```
ancalita, line 160: Could we actually use the constant directly without quotation marks around it? That would also prevent the failing unit tests.

ancalita (commented):
Looks great 🎸 Only one tiny comment regarding the removal of quotation marks around the constant when used in yaml.

tests/cli/test_rasa_data.py
melindaloubser1: I think it does want some form of quotes based on this error - but singles, not doubles because of the way we've written the test data. I also noticed that during a migration it ends up writing `'"3.0'"` which then isn't read correctly, but I figured that's a separate issue. I'll see which combo of quotes makes it happy
```        
domain = Domain.from_yaml(test_yaml)
>       assert domain.as_yaml(clean_before_dump=True) == test_yaml
E       assert "version: '3....: from_text\n" == 'version: "3....: from_text\n'
E         Skipping 983 identical trailing characters in diff, use -v to show
E         - version: "3.0"
E         ?          ^   ^
E         + version: '3.0'
```

melindaloubser1: Well, actually maybe I should fix that too because it's going to give an error regardless:
```
>           assert migrated_training_data_version == LATEST_TRAINING_DATA_FORMAT_VERSION
E           assert '"3.0"' == '3.0'
E             - 3.0
E             + "3.0"
```

ancalita: Isn't it because `LATEST_TRAINING_DATA_FORMAT_VERSION` is already assigned a string `"3.0"`? So when added to yaml, you don't need more quotes?

melindaloubser1: I think you're right, lemme check and try the tests again

melindaloubser1:
@ancalita all passing locally now, saw a failure in CI that I'm rerunning for because it looks out of date (?)

tests/core/test_migrate.py
```diff
@@ -80,7 +81,7 @@ def test_migrate_domain_format_with_required_slots(
     migrated_domain = rasa.shared.utils.io.read_yaml_file(domain_out_file)
 
     migrated_training_data_version = migrated_domain.get("version")
-    assert migrated_training_data_version == '"3.0"'
+    assert migrated_training_data_version == LATEST_TRAINING_DATA_FORMAT_VERSION
```
ancalita, line 84: Maybe in order to fix the failing tests in this module, we need to pass in an f-string with the constant? 🤔

ancalita (approved):
Happy to approve to unblock you 💯 Just a few more tests failing need your attention, gave a possible suggestion.

tests/core/test_migrate.py
melindaloubser1: urg there is always one more 😭 I believe it's another `DoubleQuotedScalarString` thing

melindaloubser1: Yep, in migrate.py there were still two hardcoded references.

PR #10910, opened 2022-02-17
Regr Tests: Get VM type tag

markus-hinsche:
**Status quo**: We already some environment information:
- metrics: CPU cores, total memory available
- tags: zone, host name (e.g. kub cluster name/id) for most metrics

**Intention of this PR**: It would be good to also have the VM type (e.g. on GCP) available as a tag in Datadog. This would allow to filter/order the data in Datadog even better.

**Proposed changes**:
- Get VM type
- Add to datadog tags

**Caveats:**
- The current code about finding the GCP VM type doesn't work in our setup
- The current code we rely on the cloud provider API(s)
- The VM type of Azure is hardcode (would be incorrect if GitHub/azure ever changed there implementation)

Ideas for solutions are welcome

### DIFF
--- a/.github/scripts/mr_publish_results.py
+++ b/.github/scripts/mr_publish_results.py
@@ -85,6 +85,9 @@ def get_vm_type() -> Dict[Text, Text]:
         machine_type_full = requests.get(
             metadata_server + "machine-type", headers=metadata_flavor
         ).text
+        print(f'machine_type_full: {machine_type_full}')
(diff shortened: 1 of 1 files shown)


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

markus-hinsche:
Current solution idea PR would be too brittle

PR #10923, opened 2022-02-21
Enable mypy attr defined

m-vdb:
**Proposed changes**:
- Fix https://github.com/RasaHQ/rasa/issues/9098


rasa/shared/core/slots.py
```diff
@@ -282,15 +282,19 @@ def _as_feature(self) -> List[float]:
...
+    @value.setter
     def value(self, value: Any) -> None:
         """Sets the slot's value."""
         if value and not isinstance(value, list):
             # Make sure we always store list items
             value = [value]
 
         # Call property setter of superclass
-        super(ListSlot, self.__class__).value.fset(self, value)
+        super().value.fset(self, value)
```
m-vdb, line 297: this is not working, tests are failing in the CI

m-vdb:
@joejuzl I'm almost done on this PR and I need your help to figure the last type issue:
- mypy doesn't like [this code](https://github.com/RasaHQ/rasa/blob/46b8bdfef2887e3d61fb111ad84e0dad9e12b2a4/rasa/graph_components/validators/default_recipe_validator.py#L479-L483), here is the [CI issue](https://github.com/RasaHQ/rasa/runs/5292103207?check_suite_focus=true)
- [here](https://github.com/RasaHQ/rasa/blob/46b8bdfef2887e3d61fb111ad84e0dad9e12b2a4/rasa/graph_components/validators/default_recipe_validator.py#L86-L90) we define `self._policy_schema_nodes` as a `List[SchemaNode]` but really it is a "list of SchemaNode that represent a Policy"
- I don't want to modify the code more than necessary here. One way to go about it would be to have a `PolicySchemaNode` that inherits `SchemaNode` but modifies `uses` to be `Policy` instead.

Do you have another idea?

joejuzl:
Can you cast ` policy_node.uses` to `Policy`? Or maybe even adding a `isinstance` or `issubtype` check in the generator comprehension would make mypy happy?
Also if there's no easy workaround, we always consider disabling a rule for one line, it's explicit at least.

m-vdb:
will give a try to those 👍🏻

m-vdb:
I went with the `cast(Policy, ...)`, it's simple and explicit and I think it's better than to just disable the check

rasa/nlu/persistor.py
```diff
@@ -162,7 +162,10 @@ class GCSPersistor(Persistor):
     Fetches them when needed, instead of storing them on the local disk."""
 
     def __init__(self, bucket_name: Text) -> None:
-        from google.cloud import storage
+        """Initialise class with client and bucket."""
+        # there are no type hints in this repo for now
```
ancalita, line 166: To double-check I understand this correctly, if there's 3rd party code without type hints without type hints, mypy will complain?

ancalita (approved):
🚀

rasa/nlu/persistor.py
m-vdb: yes exactly. It's enabled by https://github.com/RasaHQ/rasa/blob/348697b7e43d5aa478aa3dc7ccd3fc2140facebc/setup.cfg#L43

ancalita:
@m-vdb I've just merged `main` to my `mypy` branch and getting some unrelated errors to the check I enabled (`union-attr`), full list here:
```
rasa/utils/tensorflow/model_data.py:420: error: unused "type: ignore" comment
rasa/shared/core/domain.py:1499: error: Item "bool" of "Union[bool, List[Any]]" has no attribute "__iter__" (not iterable)  [union-attr]
rasa/shared/core/trackers.py:86: error: Variable "typing_extensions.TypeAlias" is not valid as a type  [valid-type]
rasa/shared/core/trackers.py:86: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
rasa/shared/core/trackers.py:725: error: Variable "rasa.shared.core.trackers.EventTypeAlias" is not valid as a type  [valid-type]
rasa/shared/core/trackers.py:725: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
rasa/shared/core/trackers.py:729: error: Variable "rasa.shared.core.trackers.EventTypeAlias" is not valid as a type  [valid-type]
rasa/shared/core/trackers.py:729: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
rasa/core/lock_store.py:229: error: Library stubs not installed for "redis" (or incompatible with Python 3.8)  [import]
rasa/core/lock_store.py:229: note: Hint: "python3 -m pip install types-redis"
rasa/core/lock_store.py:229: note: (or run "mypy --install-types" to install all missing stub packages)
rasa/core/lock_store.py:229: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
rasa/shared/importers/importer.py:175: error: Too many arguments  [call-arg]
rasa/core/tracker_store.py:319: error: Library stubs not installed for "redis" (or incompatible with Python 3.8)  [import]
rasa/core/actions/action.py:863: error: EventTypeAlias? has no attribute "parse_data"  [attr-defined]
rasa/server.py:1368: error: unused "type: ignore" comment
```

ancalita:
@m-vdb edit: running `make install` cleared these errors - all good now. Just have more `union-attr` errors to fix than before actually 😅

m-vdb:
👍🏻 yeah, sorry I didn't mention that you needed to run `make install`. I'll share a message in #atom-squad

PR #10825, opened 2022-02-07
ci: Upload diagnostic data as artifact

HotThoughts:
This PR made multiple changes on the schedule version of Model Regression Tests workflow:
- Quote path strings
  When passing dataset/config path to `rasa test core/nlu`, paths with whitespaces will break at whitespace which leads to unexpected results, e.g., not being able to locate stories to test on. 
- Upload diagnostic data to artifact if tests that test `rasa test core`. 
  Resolves https://github.com/RasaHQ/rasa/issues/10730
  
- Skip Slack notifications on github events other than `on schedule`
  We don't want to make noise while testing on PR 🙃

TODOs:
- [x] Remove test lines, specifically, revert 2baed07dacd27885fe0ec0956a9323699a111ce9

PR #10938, opened 2022-02-23
Regr Test: Fix missing comment

markus-hinsche:
**Problem description:** Like visible below, after the MR test run, the result comment's table is not rendered 
image

**Proposed changes**:
- Fix jq bug: Use select with `==` instead `=` (Introduced here: https://github.com/RasaHQ/rasa/pull/6699/) 

I am not sure why this worked before: Here is a hypothesis: maybe because `report.json` was always the first alphabetic element, so this was downloaded, but now there is diagnostic data since https://github.com/RasaHQ/rasa/pull/10825 so the order might have changed


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

tczekajlo (approved):
LGTM

PR #10945, opened 2022-02-24
Regr test improve tags

markus-hinsche:
**Proposed changes**:
- Add `host_name` name tag to agent metrics and custom metrics
- Add `INDEX_REPETITION` to datadog agent tags
- minor: Rename DATASET -> DATASET_NAME in start_dd_agent.sh

**What was the issue with `host` tag?**
* We want the same tag for all metrics (agent metrics AND custom metrics), this was not possible with `host`
* On our Kubernetes setup, the agent metrics and the custom metrics show different `host` tags: 
    * Custom metrics uses with the `hostname` command: Example: `github-runner-1892610812-6595c6cf6-4vtrf`
    * DD agent uses: Example: `gke-rasa-ci-gpu-p4-45c90add-q37z.c.rasa-ci-cd.internal`

**Observations after running with this change**
On GPU, when a regression test run contains multiple MR tests jobs (due to many datasets, configs, or repetitions defined), the github runner can be re-used. Example screenshot (with num_repetitions=5) shows that the same hostname was used for different `index_repetition`
image

In contrary, on CPU, we have 1 host_name per repetition:
image

**Check**: both tags can be observed in [this Dashboard](https://app.datadoghq.eu/dashboard/mf4-2hu-x84?tpl_var_branch_baseline=regr-test-improve-tags&from_ts=1645696402006&to_ts=1645700002006&live=false) now: 
- [x] CPU for agent metrics
- [x] CPU for custom agent
- [x] GPU for agent metrics
- [x] GPU for custom agent

https://www.notion.so/rasa/Technical-documentation-3054c4d45ad044da9c22da46b9a7cd2c

- [x] updated the documentation https://www.notion.so/rasa/Technical-documentation-3054c4d45ad044da9c22da46b9a7cd2c#fd3ff74290ce46dbbdb1b3ee6d516c73


markus-hinsche:
/modeltest

```yml
include:
 - dataset: ["financial-demo"]
```

.github/scripts/mr_publish_results.py
```diff
@@ -44,6 +44,7 @@
     "accelerator_type": "ACCELERATOR_TYPE",
     "type": "TYPE",
     "index_repetition": "INDEX_REPETITION",
+    "host_name": "HOST_NAME",
```
tczekajlo, line 47: What is the purpose of using hostname? 

I think a hostname in this case is kinda useless. The host is ephemeral and the name is random.

markus-hinsche: It can tell if 2 jobs run on the same machine one after the other, e.g. when investigating if something is pre-downloaded when the 2nd job starts. 

I started to do some speed/benchmarking investigations myself now, and I was in a situation where I wanted to group by (DD calls it `avg by`) host name.

It also provides all tags (also the custom metrics) with a sign where this was run, (e.g. if on a K8s runner setup, or on an Azure VM).

> The host is ephemeral and the name is random.

Yeah, it can not really be used to go back to this machine and look at it post-hoc, I agree.

tczekajlo (approved):
LGTM

PR #10956, opened 2022-03-01
Fix mypy var annotated

### LINKED ISSUE #9094: Enable `mypy` `var-annotated` error code

Current number of issues with this type: 62

**Definition of Done**:
- [x] enable `mypy` checks for `[var-annotated]`
- [x] all existing  `[var-annotated]` issues are fixed

m-vdb:
**Proposed changes**:
- Fixes #9094


rasa/core/policies/rule_policy.py
```diff
@@ -564,7 +564,11 @@ def _check_prediction(
         gold_action_name: Text,
         prediction_source: Optional[Text],
     ) -> List[Text]:
-        if not predicted_action_name or predicted_action_name == gold_action_name:
+        if (
+            not predicted_action_name
+            or not prediction_source
```
ka-bu, line 569: This is new. Could `None` or `""` be a valid prediction source name?

ka-bu: Looks like `_should_delete` would've thrown errors if that was `None`, no?

rasa/core/training/training.py
```diff
@@ -65,7 +65,7 @@ def create_action_fingerprints(
 
     # take into account only featurized slots
     featurized_slots = {slot.name for slot in domain.slots if slot.has_features()}
-    action_fingerprints = defaultdict(dict)
+    action_fingerprints: Dict[Text, Dict[Text, List[Text]]] = defaultdict(dict)
```
ka-bu, line 68: Can we use `typing.DefaultDict` with the python versions that we support?

rasa/engine/storage/local_model_storage.py
```diff
@@ -112,7 +112,7 @@ def _load_metadata(directory: Path) -> ModelMetadata:
         return ModelMetadata.from_dict(serialized_metadata)
 
     @contextmanager
-    def write_to(self, resource: Resource) -> ContextManager[Path]:
+    def write_to(self, resource: Resource) -> Generator[Path, None, None]:
```
ka-bu, line 115: ouch :)

ka-bu (commented):
🎉  so happy to see more annotations. Just wondering about that `prediction_source` case in `rule_policy` (see comments). Otherwise LGTM

rasa/core/policies/rule_policy.py
m-vdb: yeah exactly, it cannot be `None`. It doesn't seem like it can be the empty string:

https://github.com/RasaHQ/rasa/blob/b8433f14992151dd816028d9473aa58862539bb3/rasa/core/policies/rule_policy.py#L1089

rasa/core/training/training.py
m-vdb: [it's actually deprecated](https://docs.python.org/3/library/typing.html#typing.DefaultDict), using `defaultdict` directly seems to work locally, I pushed 702d3da1108 to see if it works well in the CI

rasa/core/policies/rule_policy.py
ka-bu: Shall we then get rid of the "Optional" instead of catching something that can't happen?

m-vdb: I guess my question is : since it's a prediction, in case we can't predict anything, would prediction_source be `None`?

rasa/core/training/training.py
ka-bu: Oh yes, those additions that get rid of "typing." are nice :) But did we merge 3.9 already? Do we run those checks in other python versions also? 🤔

m-vdb: yes we do, but we have a couple of backwards compatibility modules that get installed, to make sure we can use the latest typing good stuff

rasa/core/policies/rule_policy.py
ka-bu: so, some places where that can become None (but that None won't reach the function we're looking at here):
- [find_action_from_default_actions](https://github.com/RasaHQ/rasa/blob/b8433f14992151dd816028d9473aa58862539bb3/rasa/core/policies/rule_policy.py#L924) sets that to ´None´, and, if the source is set to None, then also the action name is set to None (and so the return type there should better be something like `Optional[Tuple[...`. Which is the reason why this is not ending up in this function (cause that source is only forwarded if the action name is not None).
- similarly for [_find_action_from_rules](https://github.com/RasaHQ/rasa/blob/b8433f14992151dd816028d9473aa58862539bb3/rasa/core/policies/rule_policy.py#L924) where the source can only be `None` if the corresponding action name is None (and hence, optional tuple would be more appropriate imho). Empty string is for initialisation only and also won't happen. 
- ...

So my conclusions are:
1. we could improve typing to express that logic better (i.e. source is specified if and only if name is specified) but that requires more changes
2. It definitely cannot be None because otherwise `should_delete` would fail [here](https://github.com/RasaHQ/rasa/blob/b8433f14992151dd816028d9473aa58862539bb3/rasa/core/policies/rule_policy.py#L554) (well, unless all rules are empty... ?)
3. 1+2 are annoying and hence sorry for the spam + let's keep your workaround + let's add a comment explaining why this was added to avoid confusion in the future :D

m-vdb: thanks for detailing, I'll add a comment

m-vdb: I've quickly tried to changed the return type of the 2 functions you mentioned using `Union[...]` but it does require some refactoring in the code. So the comment is definitely the way to go in this PR :)

ka-bu (approved):

rasa/core/migrate.py
```diff
@@ -210,9 +210,9 @@ def _migrate_domain_files(
         backup_location: where to backup all domain files
         out_path: location where to store the migrated files
     """
-    slots = {}
-    forms = {}
-    entities = []
+    slots: Dict[Text, Any] = {}
+    forms: Dict[Text, Any] = {}
+    entities: List[Any] = []
```
ancalita, line 215: Unrelated to this PR, just spotted something wrong in L264, the default in dict.get should be `[]`, instead of `{}`:
```
entities.extend(original_content.get(KEY_ENTITIES, []))
```

ancalita (approved):
LGTM 💯

PR #10964, opened 2022-03-03
Swap non-deterministic ops with possibly deterministic ones

dakshvar22:
**Proposed changes**:
- ...

### DIFF
--- a/rasa/utils/tensorflow/layers.py
+++ b/rasa/utils/tensorflow/layers.py
@@ -1146,29 +1146,32 @@ class SingleLabelDotProductLoss(DotProductLoss):
                 [softmax_logits, sim_neg_ii, sim_neg_ll], axis=-1
             )
         # create label_ids for softmax
-        label_ids = tf.concat(
-            [
-                tf.ones_like(softmax_logits[..., :1], tf.float32),
-                tf.zeros_like(softmax_logits[..., 1:], tf.float32),
-            ],
-            axis=-1,
-        )
+        # label_ids = tf.concat(
+        #     [
+        #         tf.ones_like(softmax_logits[..., :1], tf.float32),
+        #         tf.zeros_like(softmax_logits[..., 1:], tf.float32),
+        #     ],
+        #     axis=-1,
+        # )
 
         input_logits = tf.nn.softmax(softmax_logits)
 
+        loss = -tf.math.log(tf.math.maximum(input_logits[..., 0], 1e-6))
+
+        return loss
         # rescale to nullify the effect of sigmoid(logits)
         # that `weighted_cross_entropy_with_logits` would apply
-        rescaled_logits = tf.math.log(tf.maximum(input_logits, 1e-6)) - tf.math.log(
-            tf.maximum(1 - input_logits, 1e-6)
-        )
-
-        softmax_loss = tf.nn.weighted_cross_entropy_with_logits(
-            labels=label_ids, logits=rescaled_logits, pos_weight=tf.constant(1.0)
-        )
+        # rescaled_logits = tf.math.log(tf.maximum(input_logits, 1e-6)) - tf.math.log(
+        #     tf.maximum(1 - input_logits, 1e-6)
+        # )
+        #
+        # softmax_loss = tf.nn.weighted_cross_entropy_with_logits(
+        #     labels=label_ids, logits=rescaled_logits, pos_weight=tf.constant(1.0)
+        # )
         # softmax_loss = tf.nn.softmax_cross_entropy_with_logits(
         #     labels=label_ids, logits=softmax_logits
         # )
-        return softmax_loss
+        # return softmax_loss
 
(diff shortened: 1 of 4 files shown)


dakshvar22:
/modeltest

```yml
dataset_branch: "nib-test"
include:
 - dataset: ["Customer 1", "Customer 1", "Customer 1", "Customer 1"]
   config: ["Sparse + DIET(seq) + ResponseSelector(t2t)"]
```

rasa/utils/tensorflow/layers.py
```diff
@@ -1137,11 +1146,36 @@ def _compute_softmax_loss(
...
+        #         tf.ones_like(softmax_logits[..., :1], tf.float32),
+        #         tf.zeros_like(softmax_logits[..., 1:], tf.float32),
+        #     ],
+        #     axis=-1,
+        # )
+
+        # for numerical stability of softmax,
+        # see https://stackoverflow.com/a/49212689/3001665
+        input_logits = tf.nn.softmax(
+            softmax_logits - tf.math.reduce_max(softmax_logits, axis=-1, keepdims=True)
```
ka-bu, line 1160: shifting the logits should only be necessary if you implemented the softmax by hand

rasa/utils/tensorflow/layers.py
```diff
@@ -1137,11 +1146,36 @@ def _compute_softmax_loss(
...
         )
-        return softmax_loss
+
+        loss = -tf.math.log(tf.math.maximum(input_logits[..., 0], 1e-6))
```
ka-bu, line 1163: Not sure what you tried here. We can...
1. plug in the softmax definition into the cross entropy, pull out the numerator, and let the log kill the one exp (nothing more that we can do there)
2. make use of shifting (same as in softmax) to get more stable results

which should yield something like
```suggestion
        logits = softmax_logits # because that name is just confusing
        shifted_logits = logits - tf.reduce_max(logits)
        cross_entropy = - logits[...,0] + tf.reduce_logsumexp(logits)
        loss = tf.reduce_sum( cross_entropy )
```
There's also`tf.nn.log_softmax` but that would give us more summands than we'd need.

ka-bu (changes requested):

rasa/utils/tensorflow/layers.py
dakshvar22: I assume in your suggestion `logits` is non-softmax applied version of logits. In that case, the above is exactly the same as my version.

rasa/utils/tensorflow/layers.py
dakshvar22: This was an additional harmless safeguard because `tf.nn.softmax` is implemented as a C++ function in tensorflow which is not visible in the TF source code right now unless we install it from source using `bazel` (the python version uses it through `gen_nn_ops` module which contains all the compiled C++ code in python). I didn't want to get caught up in doing that right now, can do once we are sure that making these changes is useful.

rasa/utils/tensorflow/layers.py
ka-bu: Nope, cause 1) there's no exp and then log taken from the numerator (which is what I'd think ce from logits would be doing), 2) there's no evaluation of softmax-normalised entires that aren't used, 3) don't think those changes would be optimized away

dakshvar22: > Nope

I meant mathematically its the same and the above version doesn't break on at least our datasets. I wanted to try it to test if our datasets can break it.

> cause 1) there's no exp and then log taken from the numerator

yeah that's reduced but the subraction of `tf.reduce_max(logits)` and `tf.maximum` would act as safeguards.
Surely we should take your version if we merge this.

> there's no evaluation of softmax-normalised entires that aren't used

Not sure what you mean

dakshvar22:
Status: Even with the above changes, the model training is not deterministic. Will need to try a different approach to first determine what's the first point of non-determinism in the model architecture. Clearly its not `sparse.sparse_dense_matmul`.

rasa/utils/tensorflow/layers.py
ka-bu: It's not the same considering that we don't have infinite precision. What I mean by the unused entries is that tf.nn.softmax will return a tensor of the same shape as the input, while we only really need those entries corresponding to the positive examples.

dakshvar22: > It's not the same considering that we don't have infinite precision

Okay, you just mean that we should reduce the number of ops, so that precision is not lost in those ops at least, right?

> What I mean by the unused entries is that tf.nn.softmax will return a tensor of the same shape as the input, while we only really need those entries corresponding to the positive examples.

Sure thing, I didn't mean to have this as the most stable / polished version 👍

twerkmeister:
Closing this as it does not look like this line of work yielded improvements.

PR #10989, opened 2022-03-14
deprecate: deprecate NLU JSON format

wochinge:
**Proposed changes**:
- NLU data in JSON format is deprecated and will be removed in Rasa Open Source 4.0.
- fix ATO-26

### DIFF
--- a/docs/docs/migration-guide.mdx
+++ b/docs/docs/migration-guide.mdx
@@ -28,6 +28,13 @@ this new version for inference.
 
 Please check whether your trained model still performs as expected and retrain if needed.
 
+### NLU JSON Format
+
+[NLU training data](nlu-training-data.mdx) in JSON format is deprecated and will be
+removed in Rasa Open Source 4.0.
+Please use `rasa data convert nlu -f yaml --data <path to NLU data>` to convert your
+NLU JSON data to YAML format before support for NLU JSON data is removed.
+
 ## Rasa 2.x to 3.0
 
 ### Markdown Data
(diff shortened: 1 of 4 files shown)


docs/docs/migration-guide.mdx
```diff
@@ -28,6 +28,13 @@ this new version for inference.
 
 Please check whether your trained model still performs as expected and retrain if needed.
 
+### NLU JSON Format
+
+[NLU training data](nlu-training-data.mdx) in JSON format is deprecated and will be
+removed in Rasa Open Source 4.0.
```
ancalita, line 34: Should we also add that the command will be removed too in 4.0?

rasa/shared/nlu/training_data/formats/rasa.py
```diff
@@ -49,9 +94,20 @@ def read_from_json(self, js: Dict[Text, Any], **_: Any) -> "TrainingData":
...
 class RasaWriter(TrainingDataWriter):
+    """Dumps NLU data as Rasa JSON string."""
+
+    def __init__(self) -> None:
+        """Creates writer."""
+        super().__init__()
+        rasa.shared.utils.io.raise_deprecation_warning(
+            "NLU data in Rasa JSON format is deprecated and will be removed in Rasa "
+            "Open Source 4.0.0. Please convert your JSON NLU data to the "
+            "Rasa YAML format.",
```
ancalita, line 105: Should we also add here or whenever the `rasa data convert nlu` command is used that the command itself will be removed in 4.0?

ancalita (commented):
Looks good 👍🏼 
Left some questions regarding the deprecation of the command itself.

docs/docs/migration-guide.mdx
wochinge: We still support the convert command for users migration from LUIS/Watson/Dialogflow, don't we?

ancalita: I was confused as in the meeting I thought the consensus was towards removing the command, but your point makes sense.

PR #10994, opened 2022-03-15
Avoid silent unit test failures on Windows

markus-hinsche:
**Status quo**: The unit test ([Example job on main](https://github.com/RasaHQ/rasa/runs/5556974140?check_suite_focus=true#step:15:2306)) exhibits 3-4 **silent** Windows failures. 

```
FAILED tests/core/test_broker.py::test_no_pika_logs_if_no_debug_mode  (only seen once)
FAILED tests/core/test_migrate.py::test_migrate_domain_raises_exception_for_non_domain_file
FAILED tests/core/test_migrate.py::test_migrate_domain_raises_for_non_domain_files
FAILED tests/shared/nlu/training_data/test_training_data.py::test_dialogflow_data
```

This can be seen on the main branch as well, looking a different jobs of the same run:
- [windows job](https://github.com/RasaHQ/rasa/runs/5498212446?check_suite_focus=true) fails but looks like success
   - Uses `shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"`
- [linux job](https://github.com/RasaHQ/rasa/runs/5498212216?check_suite_focus=true): fails and looks like failure

**Problem:** These failures stay unnoticed.

**Background information**: 
- We use `fail-fast: false` in GitHub Actions. 
- This has different implementations according to the shell (bash or powershell) used ([GHA docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#custom-shell))
- This appears because there is another command after the failing test command (to handle test coverage): It turns out that when this silences the first error when using powershell.

**Solution**
- Use bash as shell

**Changes:**
- Use bash as shell
- Don't move or upload windows coverage, to avoid a `github.workspace` [path failure](https://github.com/RasaHQ/rasa/runs/5554276130?check_suite_focus=true)
- Fix path issues on Windows: use str instead Path, use `.as_posix()
- Make logic change to `rasa/shared/nlu/training_data/loading.py` to fix dialogflow unit test on Windows


ancalita:
@markus-hinsche I think the tests in `test_migrate.py` fail because of slash mismatch between expected path in error message and actual path in error message. Is it ok to push commits to your PR today so I can try and see how I can fix this? I'll also have a look at the other 2 tests.

markus-hinsche:
@ancalita Great, thanks a lot for the help, yes, feel free to add commits :)

ancalita:
@markus-hinsche I've fixed the failures in `test_migrate.py` + the dialogflow one. The pika one seems not to occur anymore, perhaps that was a flaky one-off? 🤔

markus-hinsche:
Hey @ancalita 
Yes I only saw `test_no_pika_logs_if_no_debug_mode` once or twice. I believe it is flaky. 
Here is a run where I saw it fail: https://github.com/RasaHQ/rasa/runs/5568662175?check_suite_focus=true#step:15:2326 
This still failed silently so it looks successful, but is actually failing.

ancalita:
> Yes I only saw `test_no_pika_logs_if_no_debug_mode` once or twice. I believe it is flaky. Here is a run where I saw it fail: https://github.com/RasaHQ/rasa/runs/5568662175?check_suite_focus=true#step:15:2326 This still failed silently so it looks successful, but is actually failing.

Looking at that run, seems to be because the logs include a WARNING message from Datadog I think: `WARNING  ddtrace.internal.writer:writer.py:378 sent 7.18KB in 2.68800s to http://localhost:8126/v0.4/traces, 17 additional messages skipped`? Maybe we can filter the caplog.records to ignore any `ddtrace` log messages and hopefully this will make the test not flaky?

markus-hinsche:
> Looking at that run, seems to be because the logs include a WARNING message from Datadog I think: WARNING ddtrace.internal.writer:writer.py:378 sent 7.18KB in 2.68800s to http://localhost:8126/v0.4/traces, 17 additional messages skipped? Maybe we can filter the caplog.records to ignore any ddtrace log messages and hopefully this will make the test not flaky?

@ancalita Amazing, great finding! I don't remember exactly, but maybe I saw this error only once. So I would suggest to go with your recommendation if we see that this appears again.

markus-hinsche:
@ancalita 
Ok it actually happened again, see the latest run: https://github.com/RasaHQ/rasa/runs/5641967227?check_suite_focus=true#step:15:1562 where we see a `test_no_pika_logs_if_no_debug_mode` failure. 
Do you want to go forward with your idea of filtering the `caplog.records`? Otherwise I can try to get familiar with the caplog code in order to follow your suggestion

m-vdb:
@markus-hinsche Anca is currently on vacation for the next 2 weeks :)

rasa/shared/nlu/training_data/loading.py
```diff
@@ -60,7 +60,7 @@ def load_data(resource_name: Text, language: Optional[Text] = "en") -> "Training
     elif len(training_data_sets) == 1:
         training_data = training_data_sets[0]
     else:
-        training_data = training_data_sets[0].merge(*data_sets[1:])
+        training_data = training_data_sets[0].merge(*training_data_sets[1:])
```
m-vdb, line 63: do you know why it was only failing on Windows?

.github/workflows/continous-integration.yml
```diff
@@ -337,7 +337,10 @@ jobs:
         DD_ARGS: --ddtrace --ddtrace-patch-all
       run: |
         make ${{ matrix.test }}
-        poetry run coverage xml -o ${{ github.workspace }}/tests/${{ matrix.test }}-coverage.xml --include="rasa/*"
+        if [[ "${{ matrix.os }}" != "windows-2019" ]]; then
+          mv .coverage ${{ github.workspace }}/${{ matrix.test }}-coverage
```
m-vdb, line 341: why not keep the `poetry run coverage ...` command?

.github/workflows/continous-integration.yml
```diff
@@ -337,7 +337,10 @@ jobs:
...
+        fi
+      shell: bash
```
m-vdb, line 343: I think this is worth a comment. For simplicity, wouldn't it be better to use `bash` in all the steps? (maybe that'd require more work though)

m-vdb (changes requested):

.github/workflows/continous-integration.yml
markus-hinsche: On Windows, there was a problem with where `github.workspace` is on a Windows OS file system. Therefore, `mv` failed when using `shell`, I am not sure if it ever worked. 

We run the tests multiple times anyways and I suspect this moving just overwrites the test jobs finishing early (e.g. ubuntu-latest on Py3.8) with the test jobs that finish late (e.g. Windows on Py3.7). The strategy matrix shows all the combinations we run. 

The suggest implementation (`if os!=windows`) therefore collects only the linux coverage.

Which coverage would we like to collect here ideally?

.github/workflows/continous-integration.yml
markus-hinsche: There are some other places (even before this PR) where we use shell for some steps. 
I can comment that the bash shell is a way to make code run for both Linux and Windows (to avoid our code having to be Windows-Powershell-compatible)

rasa/shared/nlu/training_data/loading.py
markus-hinsche: This was one of Anca's fixes, but I try to give my understanding:

For reference: a run where the  error that appeared: https://github.com/RasaHQ/rasa/runs/5556974140?check_suite_focus=true#step:15:843

```
    def test_dialogflow_data():
        td = load_data("data/examples/dialogflow/")
        assert not td.is_empty()
        assert len(td.entity_examples) == 5
        assert len(td.intent_examples) == 24
        assert len(td.training_examples) == 24
        assert len(td.regex_features) == 1
>       assert len(td.lookup_tables) == 2
E       AssertionError: assert 3 == 2
E        +  where 3 = len([{'elements': ['mexican', 'mexico', 'chinese', 'china', 'indian', 'india'], 'name': 'cuisine'}, {'elements': ['mexican...ina', 'indian', 'india'], 'name': 'cuisine'}, {'elements': ['centre', 'west', 'central', 'north'], 'name': 'location'}])
E        +    where [{'elements': ['mexican', 'mexico', 'chinese', 'china', 'indian', 'india'], 'name': 'cuisine'}, {'elements': ['mexican...ina', 'indian', 'india'], 'name': 'cuisine'}, {'elements': ['centre', 'west', 'central', 'north'], 'name': 'location'}] = <rasa.shared.nlu.training_data.training_data.TrainingData object at 0x000001FBC629AB48>.lookup_tables
```

So on Windows, the `load_data` function returned to many elements. 

The difference in the change seems to be: `training_data_sets: List[TrainingData] = [ds for ds in data_sets if ds]`

Therefore, this seems like it was a bug in this unit test, but not completely sure why it only surfaced on Windows

.github/workflows/continous-integration.yml
m-vdb: ok sounds good 👍🏻

rasa/shared/nlu/training_data/loading.py
m-vdb: 🤔 yeah ok, still weird, but oh well

.github/workflows/continous-integration.yml
m-vdb: since the issue you described only affects windows and not linux, and you're now running the command only on linux, why not keep the `poetry` command?

The `coverage xml` command generates an XML report, whereas the `mv` command moves internal coverage information to another location, I think this would break the push of the coverage report 🤔

markus-hinsche: Ohh sorry, I understand your question now (I think 🙈).

https://github.com/RasaHQ/rasa/pull/9913/files changed the behavior and uses `mv` instead `poetry run coverage xml`. The `coverage xml` happens in a later step. 

The current was inconsistent:
- single-process used `poetry run coverage xml` https://github.com/RasaHQ/rasa/blob/main/.github/workflows/continous-integration.yml#L340
- multi-process used `mv`
https://github.com/RasaHQ/rasa/blob/main/.github/workflows/continous-integration.yml#L352

m-vdb: ok that works 👍🏻 then the `Store coverage reports` report step should not execute on windows, right?

markus-hinsche: Correct, this is captured by the `if` condition of the `Store coverage reports`, so this should be good as it is now

m-vdb: oh gosh, I can't read today. Thanks for bearing with me!

Issue #9099, opened 2021-07-12
Enable `mypy` `misc ` error code


wochinge:
[Maxime recently enabled `mypy` support for the first couple of issues](https://github.com/RasaHQ/rasa/pull/6511). In our current setup we ignore certain typing issues as there are currently too many to tackle them all at once.

We should instead tackle them step by step. [Related issue](https://github.com/RasaHQ/rasa/issues/7513). [Parent issue](https://github.com/RasaHQ/rasa/issues/8243).

Current number of issues with this type: 90

**Definition of Done**:

*   [ ] enable `mypy` checks for `<span class="error"><span class="error">[misc]</span></span>`
*   [ ] all existing `<span class="error"><span class="error">[misc]</span></span>` issues are fixed

PR #10696, opened 2022-01-18
Async Tracker Store Support

danielenricocahall:
This is an effort to make tracker stores asynchronous, based off of the work initially done in [this PR](https://github.com/RasaHQ/rasa/pull/8854) and it should address [this issue] (https://github.com/RasaHQ/rasa/issues/8774). I've tried to account for the comments made in the original PR and reduce the surface area/object wrapping from the original implementation, but that comes at a trade-off - specifically making the other tracker store methods async.

Also, should we convert some of the tracker store implementations to use async APIs (i.e; [AIORedis](https://github.com/aio-libs/aioredis-py) for AioRedis, [Motor](https://motor.readthedocs.io/en/stable/) for Mongo)?

. First PR, so entirely open to feedback! 

**Status (please check what you already did)**:
- [X] added some tests for the functionality
- [X] updated the documentation
- [X] updated the changelog (please check [changelog](https://github.com/RasaHQ/rasa/tree/main/changelog) for instructions)
- [X] reformat files using `black` (please check [Readme](https://github.com/RasaHQ/rasa#code-style) for instructions)

PR #11003, opened 2022-03-17
Fix mypy misc

m-vdb:
**Proposed changes**:
- Fix https://github.com/RasaHQ/rasa/issues/9099


rasa/core/actions/action.py
```diff
@@ -406,7 +405,7 @@ def get_full_retrieval_name(
             return None
 
         response_selector_properties = latest_message.parse_data[
-            RESPONSE_SELECTOR_PROPERTY_NAME
+            RESPONSE_SELECTOR_PROPERTY_NAME  # type: ignore[misc]
```
m-vdb, line 408: `parse_data` is defined as a `TypedDict`, but `mypy` doesn't support variables as keys of a `TypedDict` (both in definition and access -- I've tried using `Final` and `Literal` without any luck). I made an attempt to use a dataclass but stopped halfway as I think it requires a proper design phase and understanding of the problem first. I can open a separate issue to document my findings (or non-findings).

rasa/shared/core/trackers.py
```diff
@@ -66,27 +64,23 @@
...
-
     EventTypeAlias = TypeVar("EventTypeAlias", bound=Event)
 
 
+@dataclasses.dataclass
+class TrackerActiveLoop:
+    """Dataclass for `DialogueStateTracker.active_loop`."""
+
+    name: Optional[Text]
+    is_interrupted: bool
```
m-vdb, line 79: a test is failing right now and it's either because the test is not set up properly (and doesn't replicate an actual possible situation) or because of legacy reasons (and in this case the code needs to be updated)

rasa/utils/common.py
```diff
@@ -407,7 +407,7 @@ async def call_potential_coroutine(
     Returns:
         The return value of the function.
     """
-    if asyncio.iscoroutine(coroutine_or_return_value):
+    if inspect.iscoroutine(coroutine_or_return_value):
```
m-vdb, line 410: this makes a difference, but I think that we want `inspect`'s version here

stubs/sanic.pyi
```diff
@@ -0,0 +1,48 @@
+from typing import Any, Dict, Iterable, List, Optional, Text, Union
```
m-vdb, line 1: I had to refactor the stubs and keep all of them in the same file, there were a few issues when loading the stubs

rasa/api.py
```diff
@@ -20,7 +20,7 @@ def run(
     connector: "Text" = None,
     credentials: "Text" = None,
     **kwargs: "Dict[Text, Any]",
-) -> "NoReturn":
```
joejuzl, line 23: Huh - I never knew about `NoReturn`. Seems to indicate a special case where a function only ever exits via Exception i.e. it literally never returns. Maybe we just want this without quotes? Or did the mypy check say it should be `None`?

rasa/core/brokers/pika.py
```diff
@@ -193,6 +193,7 @@ async def _connect(self) -> aio_pika.RobustConnection:
                 )
                 await asyncio.sleep(self._retry_delay_in_seconds)
 
+        last_exception = cast(Exception, last_exception)
```
joejuzl, line 196: Should the `cast` happen in the `except` block as that's when we know that it is no longer `none`?

rasa/core/channels/console.py
```diff
@@ -98,6 +98,11 @@ def _get_user_input(previous_response: None) -> Text:
     ...
 
 
+@overload
+def _get_user_input(previous_response: Dict[str, Any]) -> Optional[Text]:
```
joejuzl, line 102: Why do we need to overload this method?

rasa/core/policies/unexpected_intent_policy.py
```diff
@@ -112,23 +108,24 @@
...
-            THRESHOLD_KEY: Optional[float],
-            SEVERITY_KEY: Optional[float],
-        },
-    )
-
-    UnexpecTEDIntentPolicyMetadata = TypedDict(
-        "UnexpecTEDIntentPolicyMetadata",
-        {QUERY_INTENT_KEY: Text, RANKING_KEY: List["RankingCandidateMetadata"]},
-    )
+@dataclasses.dataclass
```
joejuzl, line 112: `Dataclass`es are much nicer than `TypedDict`s 👍

rasa/utils/common.py
joejuzl: This is also consistent with what will be introduced in https://github.com/RasaHQ/rasa/pull/10696

rasa/api.py
m-vdb: the thing is that the function implicitly returns. There is no `return` statement, but the function isn't raising an exception consistently. So mypy rightfully complains :)

rasa/core/policies/unexpected_intent_policy.py
m-vdb: they're also better supported by mypy!

rasa/core/channels/console.py
m-vdb: it's because there was another overload at line 97, and mypy complained about incomplete overloads

rasa/core/brokers/pika.py
m-vdb: 🤔 I don't think I've tried that, let me see if `mypy` picks it up properly

rasa/shared/core/trackers.py
m-vdb: @joejuzl do you have any idea why [the tests are failing](https://github.com/RasaHQ/rasa/runs/5602642150?check_suite_focus=true) and if my first claim is correct here?

joejuzl (approved):
💯

rasa/shared/core/trackers.py
m-vdb: https://github.com/RasaHQ/rasa/pull/11003/commits/2780efd60adbbc19b9afbf68c38b9f34b6d8407b

PR #11017, opened 2022-03-22
ATO-36 Optionally pass domain path into nlu test

joejuzl:
Optionally pass the domain path into `nlu test`. This fixes user warnings that from when a non default domain location is used.
Fixes: https://rasahq.atlassian.net/browse/ATO-36

### DIFF
--- a/rasa/cli/test.py
+++ b/rasa/cli/test.py
@@ -235,7 +235,7 @@ async def run_nlu_test_async(
             models_path, "model", DEFAULT_MODELS_PATH
         )
 
-        await test_nlu(model_path, data_path, output, all_args)
+        await test_nlu(model_path, data_path, output, all_args, domain_path=domain_path)
 
 
 def run_nlu_test(args: argparse.Namespace) -> None:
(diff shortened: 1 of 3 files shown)


m-vdb (approved):
💯 probably a good idea to automate the testing of this feature. I saw there were a few tests for `rasa test nlu` already

joejuzl:
> 💯 probably a good idea to automate the testing of this feature. I saw there were a few tests for rasa test nlu already

I wanted to add a test - but it's a bit tricky because the domain doesn't actually do anything. I.e. it's not used at all when doing the nlu evaluation. The warnings are just because the wrong domain (default or none) is getting loaded. 
And the nlu test tests are all very heavyweight and mocky....
Really the structure of data passing around the evaluation modules should be changed, this is not actually a 3.1 bug but a known issue in Rasa. However I think that is out-of-scope for now.
I could spend a bit more time getting a test that checks for warnings though if you think it's worth it?

m-vdb:
TBH I think it's ok, thanks for detailing 🙏🏻 

so the domain is not used in any way? not sure why this was in our QA pass

m-vdb:
any idea what's wrong with the CI though?

sanchariGr:
> so the domain is not used in any way? not sure why this was in our QA pass
It was in the QA pass because of this https://github.com/rasahq/rasa/issues/9351 which was in the scope.

joejuzl:
> It was in the QA pass because of this https://github.com/RasaHQ/rasa/issues/9351 which was in the scope.

Ah yes! Got it 👍 
So that fix didn't pass the domain far enough down, and this PR fixes that.

PR #11038, opened 2022-03-28
ci: Use datadog agent-github-action

HotThoughts:
**Proposed changes**:
- ...

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -231,10 +231,15 @@ jobs:
     steps:
     - name: Run DataDog Agent
       if: needs.changes.outputs.backend == 'true' && !((matrix.python-version == 3.8 || matrix.python-version == 3.9) && matrix.os == 'windows-2019')
-      run: |
-        docker run --name dd_agent -p 8126:8126 -d -e "DD_API_KEY=${{ secrets.DD_API_KEY }}" -e "DD_INSIDE_CI=true" -e "DD_HOSTNAME=none" -e "DD_SITE=datadoghq.eu" -e GITHUB_ACTIONS=true -e CI=true datadog/agent:latest
-        docker ps --all --filter name=dd_agent --filter status=running --no-trunc --format "{{.ID}} {{.Status}}"
-        docker port dd_agent
+      uses: datadog/agent-github-action@9e911781db79c5128ea2890f4ab2538123a7e5bd
+      with:
+        api_key: ${{ secrets.DD_API_KEY }}
+        image_name: datadog/agent
+        datadog_site: datadoghq.eu
+        extra_env:
+          - GITHUB_ACTIONS=true
+          - CI=true
+
 
     - name: Checkout git repository 🕝
       if: needs.changes.outputs.backend == 'true'


.github/workflows/continous-integration.yml
```diff
@@ -233,17 +236,19 @@ jobs:
...
-      run: |
-        docker run --name dd_agent -p 8126:8126 -d -e "DD_API_KEY=${{ secrets.DD_API_KEY }}" -e "DD_INSIDE_CI=true" -e "DD_HOSTNAME=none" -e "DD_SITE=datadoghq.eu" -e GITHUB_ACTIONS=true -e CI=true datadog/agent:latest
-        docker ps --all --filter name=dd_agent --filter status=running --no-trunc --format "{{.ID}} {{.Status}}"
-        docker port dd_agent
-
     - name: Checkout git repository 🕝
       if: needs.changes.outputs.backend == 'true'
       uses: actions/checkout@v3
 
+    - name: Run DataDog Agent
```
ancalita, line 243: Does it matter that this step follows the git repo checkout step?

.github/workflows/continous-integration.yml
```diff
@@ -233,17 +236,19 @@ jobs:
...
+      if: needs.changes.outputs.backend == 'true'
+      uses: datadog/agent-github-action@9e31e08e9eebe3eb04401dd3e5a1dedefe989d97
```
ancalita, line 245: Github Actions newbie here 👋🏼 How do I read this line? 🙈 Is there a folder I need to look into?

.github/workflows/continous-integration.yml
```diff
@@ -254,7 +259,6 @@ jobs:
       if: needs.changes.outputs.backend == 'true'
       run: |
         echo "POETRY_VERSION=$(scripts/poetry-version.sh)" >> $GITHUB_ENV
-      shell: bash
```
ancalita, line 257: Why is this removed?

.github/workflows/continous-integration.yml
```diff
@@ -233,17 +236,19 @@ jobs:
         python-version: [ 3.7, 3.8, 3.9 ]
 
     steps:
-    - name: Run DataDog Agent
-      if: needs.changes.outputs.backend == 'true' && !((matrix.python-version == 3.8 || matrix.python-version == 3.9) && matrix.os == 'windows-2019')
```
ancalita, line 237: Because of the limitation we have of not being able to run pytest profiling on Windows, shouldn't we keep running datadog agent on windows even if we get less information?

ancalita (commented):
Amazing work 💯 🎸 
Left some questions but look at them as from the perspective of a beginner 😅

.github/workflows/continous-integration.yml
HotThoughts: [The job-level default shell is bash](https://github.com/RasaHQ/rasa/blob/b0fa94243abda1aa9504b2eacb5044a35a89575e/.github/workflows/continous-integration.yml#L220-L222). No difference really, just a preference of setting shell.

HotThoughts: [The job-level default shell is bash](https://github.com/RasaHQ/rasa/blob/b0fa94243abda1aa9504b2eacb5044a35a89575e/.github/workflows/continous-integration.yml#L220-L222). No difference really, just a preference of setting shell.

.github/workflows/continous-integration.yml
```diff
@@ -300,6 +304,7 @@ jobs:
 
     - name: Install Dependencies (Windows) 📦
       if: needs.changes.outputs.backend == 'true' && matrix.os == 'windows-2019'
+      shell: powershell
```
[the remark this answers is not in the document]
HotThoughts: Since the default shell is now bash and the command in this step is PowerShell-specific, we need to change the shell to PowerShell to execute it.

.github/workflows/continous-integration.yml
HotThoughts: This line tells GitHub Actions to use the repo [`datadog/agent-github-action`](https://github.com/DataDog/agent-github-action) and in particular, use the code at the commit `9e31e08e9eebe3eb04401dd3e5a1dedefe989d97`. Ideally, we want to use a third-party action that is secure and actively maintained. It is an official GitHub Action from DataDog, so all good.

.github/workflows/continous-integration.yml
HotThoughts: Good question. As long as we start the agent before the unit test step, we are good. Plus, the datadog agent works fine. 😜

.github/workflows/continous-integration.yml
HotThoughts: Unfortunately, we are not able to get any profiling information on Windows. If we enable it, the datadog agent will complain in the very beginning and [resulting in a silent failure](https://github.com/RasaHQ/rasa/runs/6453645890?check_suite_focus=true#step:16:35). 🙈

ancalita: I see, and this part that got removed cannot co-exist with the new step that you added?

HotThoughts: [The part](https://github.com/RasaHQ/rasa/blob/129bba0a64f43a0bfafdd41e508e3d4c02648dd2/.github/workflows/continous-integration.yml#L236-L241) that got deleted here is a docker setup for starting datadog agent. By the time we added this step, DataDog didn't provide an official solution for setting up an agent inside GitHub Actions. 

The official GitHub Actions were just released in Jan 2022. It is an elegant solution to properly start and stop an agent. Hence, we [replace](https://github.com/RasaHQ/rasa/blob/74fe422ac29ec9fec8c8479e9921b55487412347/.github/workflows/continous-integration.yml#L243-L250) the old docker commands with the official GitHub Actions.

RASADSA (approved):
LGTM +1
nice work @HotThoughts

HotThoughts:
Would really appreciate if someone has the bandwidth to take care of this PR since I cannot push any commits to the branch.

PR #11052, opened 2022-04-04
Drop python36 and security updates

joejuzl:
Proposed changes:
Fixes https://rasahq.atlassian.net/browse/SB-66


losterloh (approved):
Everything good with the PR in itself, but before merging this we should answer the open questions around how we exactly want to release this.

tmbo:
the pr doesn't include updates for tf / pillow - are they in a separate pr or am i missing something?

m-vdb:
@joejuzl @losterloh are we confident that the redis major version changes is sufficiently tested with our functional / integration tests or do we need to do a manual QA step?

losterloh:
@m-vdb I don't think I can answer that question with my current level of knowledge, I'd need to invest some time to look at the integration test suite for that. (Of course, happy to do that if this is now a high priority).

m-vdb:
sounds good @losterloh -- I don't expect you to get this answer since you just joined Atom last week. I've taken the habit of @-mentioning you and Joe 😅

joejuzl:
@m-vdb the integration tests only test redis with the lock store, so I would say to be safe a QA pass makes sense.

PR #11076, opened 2022-04-12
Update `pytest` dev-dependency

ancalita:
**Proposed changes**:
- Update `pytest` to match [dependabot PR ](https://github.com/RasaHQ/rasa-sdk/pull/738) in rasa-sdk.


tests/core/evaluation/test_marker_stats.py
```diff
@@ -135,8 +135,10 @@ def test_process_results_per_session(seed: int):
                 preceding_user_turn_numbers_used_per_marker[marker][idx]
             )
             for stat_name, stat_value in expected_stats.items():
-                assert pytest.approx(
-                    stats.session_results[marker][stat_name][idx], stat_value
+                assert np.isclose(
```
ancalita, line 138: @ka-bu I requested your review to check that this change in this unit test is valid. The current update to `pytest` breaks the test in its old unit test (see[ CI workflow](https://github.com/RasaHQ/rasa/runs/5993986933?check_suite_focus=true))

kedz: Looking at the error, it looks like the original test was not correctly written:

```
>                   assert pytest.approx(
                        stats.session_results[marker][stat_name][idx], stat_value
                    )
E                   AssertionError: approx() is not supported in a boolean context.
E                   Did you mean: `assert a == approx(b)`?
```

as in 

```python
assert pytest.approx(stats.session_results[marker][stat_name][idx], stat_value)
```
just asserts that the result of `pytest.approx()` is a non `None` object, not whether the expected stat value is equal to the actual stat value.

Why not just rewrite this as 

```python
assert pytest.approx(stats.session_results[marker][stat_name][idx]) == stat_value
```

Or is it the case that we are expecting NaNs and need the NaNs equality check in `np.isclose`?

kedz (commented):
I have a question about the expected test behavior, but otherwise looks like it actually fixes a bug in the original test.

tests/core/evaluation/test_marker_stats.py
ancalita: Oh yes, so the `pytest.approx` assetion fails unless I add the `nan_ok=True` kw arg, which defaults to `False` otherwise. I'll change to this now.

ka-bu: yep, that should've been an np.isclose 👍  And yes, nans can appear because If there are no signals collected, all stats (except the count which will be 0) over those non-existing signals etc will be nan

tests/cli/conftest.py
```diff
@@ -2,12 +2,13 @@
...
 
-from _pytest.tmpdir import TempdirFactory
 from typing import Callable, Text
 import pytest
 import shutil
 import os
-from _pytest.pytester import Testdir, RunResult
+
+from _pytest.legacypath import TempdirFactory, Testdir
+from _pytest.pytester import RunResult
```
ka-bu, line 11: The leading underscore is fishy. Docs look like we should be able to import directly from pytest, i.e. `from pytest import TempdirFactory`

tests/core/evaluation/test_marker_stats.py
ancalita: Thanks for the follow-up @ka-bu 👍🏼 
Should I revert to using `np.isclose` or the current usage of `pytest.approx` is also alright?

ka-bu: Guess the intention had been to use `np.isclose(a,b)` to represent "a aproximately b" but then pytest.approx was called instead of np.isclose, and that is a problem because the notation is different with pytest.approx, i.e. "pytest.approx(a) = b" (There is a comparison of np.iclose and pytest.approx at the end of the docs [here](https://docs.pytest.org/en/6.2.x/reference.html#pytest-approx)). 
Fixing the notation and adding the nan_ok=True as you suggested is ok 👍

ancalita: Great 🙌🏼  let me know if this can be approved 😄

ka-bu (approved):
🎉

PR #11079, opened 2022-04-12
add a footnote about model_confidence cosine

melindaloubser1:
**Proposed changes**:
- Add a note that `model_confidence: cosine` will still be used if `loss_type: margin` is specified, regardless of `model_confidence` set in config.yml.


l2nguyen (approved):
LGTM

PR #11088, opened 2022-04-20
Extract new event streaming for loop into separate `TrackerStore` method

ancalita:
**Proposed changes**:
- Refactor `TrackerStore.stream_events` method by extracting new event streaming for loop into separate `TrackerStore` method

### DIFF
--- a/rasa/core/tracker_store.py
+++ b/rasa/core/tracker_store.py
@@ -299,7 +299,7 @@ class TrackerStore:
         """Publishes new tracker events to a message broker."""
         if self.event_broker is None:
             return None
-        
+
         for event in new_events:
             body = {"sender_id": sender_id}
             body.update(event.as_dict())
--- a/rasa/core/tracker_store.py
+++ b/rasa/core/tracker_store.py
@@ -296,11 +296,13 @@ class TrackerStore:
         events = tracker.events
         new_events = list(itertools.islice(events, offset, len(events)))
 
-        await self._stream_new_events(new_events, tracker.sender_id, self.event_broker)
+        await self._stream_new_events(self.event_broker, new_events, tracker.sender_id)
 
-    @staticmethod
     async def _stream_new_events(
-        new_events: List[Event], sender_id: Text, event_broker: EventBroker
+        self,
+        event_broker: EventBroker,
+        new_events: List[Event],
+        sender_id: Text,
     ) -> None:
         """Publishes new tracker events to a message broker."""
         for event in new_events:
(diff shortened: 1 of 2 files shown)


rasa/core/tracker_store.py
```diff
@@ -289,13 +289,19 @@ async def retrieve_full_tracker(
 
     async def stream_events(self, tracker: DialogueStateTracker) -> None:
         """Streams events to a message broker."""
+        offset = await self.number_of_existing_events(tracker.sender_id)
```
losterloh, line 292: I'd suggest still doing the `self.event_broker is None` check at the very beginning here like before, instead of in `stream_new_events()`.

rasa/core/tracker_store.py
```diff
@@ -289,13 +289,19 @@ async def retrieve_full_tracker(
...
+        events = tracker.events
+        new_events = list(itertools.islice(events, offset, len(events)))
+
+        await self.stream_new_events(new_events, tracker.sender_id)
+
+    async def stream_new_events(self, new_events: List[Event], sender_id: Text) -> None:
```
losterloh, line 298: Should we mark this as an internal method by prefixing the method name with a `_`?

losterloh (commented):
Do we not need changelog entries for pure refactorings like this one?

rasa/core/tracker_store.py
ancalita: Oh I moved that because mypy check would fail with a `union-attr` error type and it didn't make sense to have this duplicated in 2 places: 
```
rasa/core/tracker_store.py:306: error: Item "None" of "Optional[EventBroker]" has no attribute "publish"  [union-attr]
```

losterloh: Then one way to keep the original flow would be to refactor this so we pass the broker as an argument in `stream_new_events()` instead of accessing the instance field - then we can do the check at the beginning but mypy should still be happy that way.

ancalita: I'll try that 👍🏼

ancalita: Seems I cannot pass `self.event_broker` as argument, therefore I have to assign this attribute or copy it to a different local variable, something in this seems wrong although cannot say for sure why 🤔

losterloh: what's the error when you try passing it as an argument?

losterloh: Interesting, my pycharm seems perfectly happy with this solution. But in any case I'd say our linter should be the authoritative source on what's allowed and what's not 🙂 static method works too of course 👍

losterloh (approved):

PR #11102, opened 2022-04-25
Add optional `--endpoints` argument to `rasa train`

melindaloubser1:
**Proposed changes**:
- Add the `--endpoints` argument to rasa train to make endpoints available during training if needed.

### DIFF
--- a/changelog/11102.improvement.md
+++ b/changelog/11102.improvement.md
@@ -1,0 +1,1 @@
+Add `--endpoints` command line parameter to `rasa train` parser.
--- a/tests/cli/test_rasa_train.py
+++ b/tests/cli/test_rasa_train.py
@@ -5,7 +5,7 @@ from pathlib import Path
 
 from _pytest.capture import CaptureFixture
 import pytest
-from typing import Callable
+from typing import Callable, List
 from _pytest.pytester import RunResult
 from _pytest.tmpdir import TempPathFactory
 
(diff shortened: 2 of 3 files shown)


tests/cli/test_rasa_train.py
```diff
@@ -64,6 +64,43 @@ def test_train(run_in_simple_project: Callable[..., RunResult], tmp_path: Path):
...
+        "-d",
+        "domain.yml",
+        "--data",
+        "data",
+        "--out",
+        "train_models",
+        "--fixed-model-name",
+        "test-model",
+        "--endpoints",
+        "endpoints.yml",
```
ancalita, line 85: Should we parameterize this when endpoints are given as argument, and when endpoints are not given as input to the command line and then the default file is read? Also is there a way to assert that the contents of `endpoints.yml` is correctly processed during training?

changelog/11102.improvement.md
```diff
@@ -0,0 +1 @@
+Add `--endpoints` command line parameter to `rasa train` parser.
```
ancalita, line 1: We probably need to add this new info to the CLI docs too?

tests/cli/test_rasa_train.py
melindaloubser1: Yes to the parametrization, good point.

Re. processing: It's not processed at all during training at the moment. If there was a reason to do so in the future, we could add it.

changelog/11102.improvement.md
melindaloubser1: Luckily the docs list the possible `train` parameters by calling `rasa train --help` so it will list the new param automatically.

tests/cli/test_rasa_train.py
```diff
@@ -31,7 +31,19 @@
 import rasa.utils.io
 
 
-def test_train(run_in_simple_project: Callable[..., RunResult], tmp_path: Path):
+@pytest.mark.parametrize(
+    "optional_arguments",
+    [
+        ["--endpoints", "endpoints.yml"],
+        ["--endpoints", "non_existent_endpoints.yml"],
```
ancalita, line 38: Shouldn't this raise some error if the file doesn't exist? 🤔
I think the fixture `run_in_simple_project` creates an `init` project which will contain by default `endpoints.yml`

melindaloubser1: It doesn't raise an error because the `train` process doesn't process the file at all.  For this test, I wanted to make sure that if in the future endpoints are processed during training, we would notice immediately that this test fails because it now cares about existent/non-existent endpoints.

ancalita (approved):

PR #11114, opened 2022-05-09
Update readerwriter.py

duanzhihua:
Aggregated_entities [entity['start']][0] : represents the 0 entity element in aggregated_entities whose entity['start'] is the key.

Aggregated_entities [entity['start']][-1] : indicates the last entity element in aggregated_entities with entity['start'] as the key.

The log prompt here should be to compare the end position of the current entity with the end position of the 0th entity, and raise an exception if they are inconsistent, so change -1 to 0.

### DIFF
--- a/rasa/shared/nlu/training_data/formats/readerwriter.py
+++ b/rasa/shared/nlu/training_data/formats/readerwriter.py
@@ -33,17 +33,17 @@ def _raise_on_same_start_and_different_end_positions(
     Args:
         aggregated_entities: Entities for each start position
     """
-     for entity_list in aggregated_entities.values():
-            end = entity_list[0].get("end")
-            for entity in entity_list[1:]:
-                # By construction, start positions of all entities in `entity_list` are
-                # identical
-                if entity.get("end") != end:
-                    raise ValueError(
-                        f"Entities '{entity}' and "
-                        f"'{entity_list[0]}' have identical "
-                        f"start but different end positions"
-                    )
+    for entity_list in aggregated_entities.values():
+        end = entity_list[0].get("end")
+        for entity in entity_list[1:]:
+            # By construction, start positions of all entities in `entity_list` are
+            # identical
+            if entity.get("end") != end:
+                raise ValueError(
+                    f"Entities '{entity}' and "
+                    f"'{entity_list[0]}' have identical "
+                    f"start but different end positions"
+                )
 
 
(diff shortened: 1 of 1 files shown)


rasa/shared/nlu/training_data/formats/readerwriter.py
```diff
@@ -33,17 +33,18 @@ def _raise_on_same_start_and_different_end_positions(
...
-            # By construction, start positions of all entities in `entity_list` are
-            # identical
-            if entity["end"] != end:
-                raise ValueError(
-                    f"Entities '{entity}' and "
-                    f"'{aggregated_entities[entity['start']][0]}' have identical "
-                    f"start but different end positions"
-                )
+     for entity_list in aggregated_entities.values():
+        if len(entity_list) > 0:
```
melindaloubser1, line 37: Did you run into an error where the list was empty? As far as I can see it wasn't previously necessary to add this check

rasa/shared/nlu/training_data/formats/readerwriter.py
```diff
@@ -33,17 +33,18 @@ def _raise_on_same_start_and_different_end_positions(
...
+            end = entity_list[0].get("end", None)
```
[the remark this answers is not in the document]
duanzhihua: Thank you. It has been revised and submitted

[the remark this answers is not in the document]
melindaloubser1: I think you may have introduced some extra whitespace - you can run `make formatter` to ensure correct formatting

rasa/shared/nlu/training_data/formats/readerwriter.py
```diff
@@ -34,14 +34,14 @@ def _raise_on_same_start_and_different_end_positions(
         aggregated_entities: Entities for each start position
     """
     for entity_list in aggregated_entities.values():
-        end = entity_list[0]["end"]
+        end = entity_list[0].get("end")
```
melindaloubser1, line 37: final question: Did you switch to `get` because of an error you  encountered, or for another reason? An entity without an `end` would not be a valid entity

duanzhihua: The exception of dictionary key **end** was not encountered during the run of the production project. Modify the code just to pass the code climate Test

melindaloubser1 (approved):

PR #11123, opened 2022-05-16
make action server connection error message more descriptive

melindaloubser1:
**Proposed changes**:
- Make the exception raised when the action server cannot be reached more descriptive

### DIFF
--- a/tests/core/test_actions.py
+++ b/tests/core/test_actions.py
@@ -548,7 +548,7 @@ async def test_remote_action_endpoint_not_running(
 
     with pytest.raises(Exception) as execinfo:
         await remote_action.run(default_channel, default_nlg, default_tracker, domain)
-    assert "Failed to execute custom action." in str(execinfo.value)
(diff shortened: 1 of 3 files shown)


tests/core/test_actions.py
```diff
@@ -548,7 +548,7 @@ async def test_remote_action_endpoint_not_running(
 
     with pytest.raises(Exception) as execinfo:
         await remote_action.run(default_channel, default_nlg, default_tracker, domain)
-    assert "Failed to execute custom action." in str(execinfo.value)
+    assert "Failed to execute custom action" in str(execinfo.value)
```
losterloh, line 551: Why are we now omitting the final period `.`? Do we have a policy for formatting log messages?

losterloh (approved):

tests/core/test_actions.py
melindaloubser1: Omitting here because in the messages the test checks, it is now followed by other words - it's no longer a standalone sentence. Otherwise log messages should end with a period.

PR #11129, opened 2022-05-18
Validate incoming JWT tokens from the bot framework

losterloh:
**Proposed changes**:
This PR enables JSON Web Token validation for the JWTs coming in from the Azure botframework channel. This is achieved by periodically updating the keys that Microsoft uses to sign the tokens, saving those in memory, and then validating the tokens and verifying their signature for incoming requests from the bot framework. 
For details on how the botframework does authentication, see [here](https://docs.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-authentication?view=azure-bot-service-4.0#connector-to-bot).

### DIFF
--- a/rasa/core/channels/botframework.py
+++ b/rasa/core/channels/botframework.py
@@ -1,13 +1,26 @@
 import datetime
 import json
 import logging
-import requests
-from sanic import Blueprint, response
-from sanic.request import Request
+import re
+from http import HTTPStatus
 from typing import Text, Dict, Any, List, Iterable, Callable, Awaitable, Optional
 
-from rasa.core.channels.channel import UserMessage, OutputChannel, InputChannel
+import jwt
+import requests
+from jwt import InvalidKeyError, PyJWTError
+from jwt.algorithms import RSAAlgorithm
+from requests import HTTPError
+from sanic import Blueprint, response
+from sanic.request import Request
 from sanic.response import HTTPResponse
+
+from rasa.core.channels.channel import UserMessage, OutputChannel, InputChannel
+
+MICROSOFT_OPEN_ID_URI = (
+    "https://login.botframework.com/v1/.well-known/openidconfiguration"
+)
+
+BEARER_REGEX = re.compile(r"Bearer\s+(.*)")
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


rasa/core/channels/botframework.py
```diff
@@ -188,6 +200,37 @@ def __init__(self, app_id: Text, app_password: Text) -> None:
         self.app_id = app_id
         self.app_password = app_password
 
+        self.jwt_keys: Dict[Text, Any] = {}
+        self.jwt_update_time = datetime.datetime.fromtimestamp(0)
+
+        self._update_cached_jwk_keys()
+
+    def _update_cached_jwk_keys(self) -> None:
+        if datetime.datetime.now() - self.jwt_update_time < datetime.timedelta(days=1):
```
losterloh, line 209: For now, I'd say a daily update of the JWT keys is plenty, and also currently no need to make this configurable.

rasa/core/channels/botframework.py
```diff
@@ -215,6 +262,36 @@ async def health(request: Request) -> HTTPResponse:
 
         @botframework_webhook.route("/webhook", methods=["POST"])
         async def webhook(request: Request) -> HTTPResponse:
+            # Update the JWT keys daily
```
ancalita, line 265: I suggest extracting this new code (L265-294) into a public method that should be unit tested too.

rasa/core/channels/botframework.py
```diff
@@ -188,6 +201,40 @@ def __init__(self, app_id: Text, app_password: Text) -> None:
...
         self.app_password = app_password
 
+        self.jwt_keys: Dict[Text, Any] = {}
+        self.jwt_update_time = datetime.datetime.fromtimestamp(0)
+
+        self._update_cached_jwk_keys()
+
+    def _update_cached_jwk_keys(self) -> None:
+        logger.debug("Updating JWT keys for the Botframework.")
+        req = requests.get(MICROSOFT_OPEN_ID_URI)
```
ancalita, line 211: Nitpicking, I would actually label this variable `response`, since this would hold a Response object.

tests/core/channels/test_botframework.py
```diff
@@ -0,0 +1,155 @@
...
+@pytest.fixture(scope="function")
+def rsa_private_key() -> RSAPrivateKey:
+    return rsa.generate_private_key(
+        public_exponent=65537,
+        key_size=2048,
+    )
+
+
+@pytest.fixture(scope="function")
+def rsa_key_pair(rsa_private_key: RSAPrivateKey) -> Tuple[bytes, bytes]:
```
ancalita, line 38: Where is this fixture used? Maybe best to remove if not in use.

rasa/core/channels/botframework.py
```diff
@@ -188,25 +201,93 @@ def __init__(self, app_id: Text, app_password: Text) -> None:
...
+            jwt_token,
+            key=public_key,
+            audience=self.app_id,
+            algorithms=jwt_header["alg"],
+        )
+
+    def _validate_auth(self, auth_header: Optional[Text]) -> Optional[HTTPResponse]:
+        if not auth_header:
+            return response.text(
+                "No authorization header provided", status=HTTPStatus.UNAUTHORIZED
```
ancalita, line 241: Does this mean that this channel must provide an authorization header now? I'd check if docs need to specify this.
Also not sure if this would be an unauthorized or a bad request type of status? 🤔

rasa/core/channels/botframework.py
```diff
@@ -188,25 +201,93 @@ def __init__(self, app_id: Text, app_password: Text) -> None:
...
+                logger.warning(
+                    f"Could not update JWT keys from {MICROSOFT_OPEN_ID_URI}."
+                )
+                logger.exception(error, exc_info=True)
+
+        auth_match = BEARER_REGEX.match(auth_header)
+        if not auth_match:
+            return response.text(
+                "No Bearer token provided in Authorization header",
+                status=HTTPStatus.UNAUTHORIZED,
```
ancalita, line 258: Should this be a `BadRequest` instead of `Unauthorized`?

ancalita (commented):
Great work 💯 
Left some existential questions about the type of status to be returned in some cases 😄 .
Also please look at the additional comments in the unit tests to strengthen the existing assertions 💪🏼

rasa/core/channels/botframework.py
losterloh: I don't think we need to do anything there, since the header should always be provided by Azure, which is who calls this channel.

Opinion not too strong about the HTTP status code, however, `bad request` is more generic and providing no auth header to me perfectly fits the definition of `unauthorized`:
```

File: rasa/core/channels/botframework.py (hunk 6 above)
Comment (losterloh) [reply rasa/core/channels/botframework.py:258]: I don't think so, see above.

File: tests/core/channels/test_botframework.py (hunk 4 above)
Comment (losterloh) [reply tests/core/channels/test_botframework.py:38]: Thanks, good catch, that was left over from a previous implementation of the test.

File: tests/core/channels/test_botframework.py [hunk 7]
```diff
@@ -0,0 +1,155 @@
...
+            "exp": int(time.time()) + DAY_IN_SECONDS,
+            "iss": "https://api.botframework.com",
+            "aud": "app_id",
+        },
+        rsa_private_key,
+        algorithm="RS256",
+        headers={"kid": "key_id", "alg": "RS256"},
+    )
+
+    resp = bot_framework_input._validate_auth(f"Bearer {encoded}")
```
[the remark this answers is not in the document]
losterloh: Can you elaborate on the advantage of doing this? A server handling a request with insufficient authorization is definitely a normal, non-exceptional case, and asserting that something does _not_ happen can often make tests hard to understand.

tests/core/channels/test_botframework.py
```diff
@@ -0,0 +1,155 @@
...
+            "exp": int(time.time()) + 2 * DAY_IN_SECONDS,
+            "iss": "https://api.botframework.com",
+            "aud": "app_id",
+        },
+        rsa_private_key,
+        algorithm="RS256",
+        headers={"kid": "key_id", "alg": "RS256"},
+    )
+
+    resp = bot_framework_input._validate_auth(f"Bearer {encoded}")
```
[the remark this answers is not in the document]
losterloh: Not sure I understand - the method does not raise a `PyJWTError`, so we can't use `pytest.raises`, right?

tests/core/channels/test_botframework.py
```diff
@@ -0,0 +1,155 @@
...
+    assert resp is not None
+    assert resp.status == HTTPStatus.UNAUTHORIZED
+
+
+def test_validate_auth_returns_unauthorized_for_absent_header(
+    bot_framework_input: BotFrameworkInput,
+):
+    resp = bot_framework_input._validate_auth(None)
+    assert resp is not None
```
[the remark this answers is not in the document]
losterloh: I don't think asserting the content of these texts provides any value, since they would always be discarded by the Azure bot framework, which is what calls these in the end. But no strong opinion, can add those if you think it makes sense 🙂

tests/core/channels/test_botframework.py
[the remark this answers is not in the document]
ancalita: Oh yes, you're right, I didn't pay attention that it's a logging exception, in that case I'd test the logs are correct 😄

tests/core/channels/test_botframework.py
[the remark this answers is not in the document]
ancalita: It's not immediately clear that if the response is `None` it means everything executed without issues. 
I thought it's better to make it explicit in the unit test that no warning was raised to deem the validation successful - based on unit testing practices I've seen across the rasa repo.

tests/core/channels/test_botframework.py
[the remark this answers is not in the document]
ancalita: I should have said these are for unit test coverage, rather than the correctness of the code.

ancalita (approved):
🎉 Thank you for addressing those extra assertions 💯

losterloh:
@Maxinho96 thank you for pointing the`jwt.PyJWKClient` out, did not know about that one before and looks great! We'll consider refactoring this again to use that.

PR #11130, opened 2022-05-18
Correct & clarify parts of API spec

melindaloubser1:
**Proposed changes**:

GET  /status endpoint: Correct response schema for model_id -  a string, not an object.

GET /conversations/{conversation_id}/tracker : Describe each of the enum options for include_events query parameter

POST & PUT /conversations/{conversation_id}/tracker/events : Events schema added for each event type

GET /conversations/{conversation_id}/story: Clarified  the `all_sessions` query parameter and default behaviour.
Also fixed the story returned when a session is ended by a `restarted` event, so that the story is not blank.

POST /model/test/intents :  Remove JSON option, since the server was interpreting all input as YAML regardless of content-type header.

POST /model/parse: Explain what emulation_mode is and how it affects response results


docs/docs/http-api.mdx
```diff
@@ -50,54 +50,44 @@ The [SocketIO channel](./connectors/your-own-website.mdx#websocket-channel) does
 
 ## Security Considerations
 
-We recommend to not expose the Rasa Server to the outside world, but
-rather connect to it from your backend over a private connection (e.g.
-between docker containers).
+We recommend that you don't expose the Rasa Server to the outside world directly, but
+rather connect to it from via e.g. Nginx as a reverse proxy.
```
melindaloubser1, line 54:
```suggestion
rather connect to it via e.g. Nginx.
```

rasa/shared/core/trackers.py
```diff
@@ -940,3 +940,33 @@ def get_trackers_for_conversation_sessions(
...
+
+    Returns:
+        The tracker up until the first restart event.
+    """
+    split_conversations = events.split_events(
+        tracker.events,
+        Restarted,
+        include_splitting_event=False,
+    )
+    if len(split_conversations) == 0:
```
ancalita, line 963: We need some unit test coverage for this new function - I wasn't familiar with this `split_events` function, but it seems that it will always return a `List[List[Event]]` if there are events in the tracker, regardless whether there is a restart or not. The only time it will return an empty list is when the tracker has no events 🤔

rasa/server.py
```diff
@@ -1152,13 +1158,11 @@ async def evaluate_intents(
...
-            if cross_validation_folds:
-                raise ErrorResponse(
-                    HTTPStatus.BAD_REQUEST,
-                    "TestingError",
-                    "Cross-validation is only supported for YAML data.",
-                )
+            raise ErrorResponse(
+                HTTPStatus.BAD_REQUEST,
+                "TestingError",
+                "NLU evaluation is only supported for YAML data.",
```
ancalita, line 1164: Why is JSON removed as option?

rasa/server.py
```diff
@@ -317,6 +321,8 @@ async def get_test_stories(
     else:
         trackers = [await processor.get_tracker(conversation_id)]
 
+    trackers = [get_events_before_restart(tracker) for tracker in trackers]
```
ancalita, line 324: Why is this required, does `Restarted` presence in `tracker.events` cause a bug?

ancalita (commented):
Great stuff 💯 Left a few clarifying questions 😄 
Also worth looking into the JSON removal and why it's causing the `test_evaluate_invalid_intent_model_file` to fail.

rasa/server.py
melindaloubser1: I'm not sure why, but it was removed already previously - it just wasn't documented that it was. Even passing a JSON body, it would get interpreted as YAML and throw an error.

rasa/server.py
melindaloubser1: The presence of `Restarted` means the session is translated into a blank story  - due to using the applied events as defined [here](https://github.com/RasaHQ/rasa/blob/dc4930d50b800c6a059f87b3f92f27468d995aad/rasa/shared/core/trackers.py#L490-L491). Tbh, I think the more correct thing to do is to change the definition of how sessions are split to include restarted events at the beginning of each session instead of leaving them at the end - but I'm concerned this would result in changes elsewhere that we don't want.

rasa/shared/core/trackers.py
melindaloubser1: I'll take another look and add a test

rasa/server.py
ancalita: I see, so this becomes a problem further below in L340 because `tracker.as_story()` makes a call to `applied_events` which will be an empty list if the last event was `Restarted`?

ancalita: What would happen if other events follow the `Restarted` event?

docs/docs/http-api.mdx
```diff
@@ -50,54 +50,44 @@ The [SocketIO channel](./connectors/your-own-website.mdx#websocket-channel) does
...
-between docker containers).
+We recommend that you don't expose the Rasa Server to the outside world directly, but
+rather connect to it via e.g. Nginx.
 
 Nevertheless, there are two authentication methods built in:
 
 ### Token Based Auth
 
-Pass in the token using `--auth-token thisismysecret` when starting
+To use a plaintext token to secure your server, specify the token in the argument `--auth-token thisismysecret` when starting
```
ancalita, line 60: These are great additions, however I just realised that I didn't clarify in my ticket that the page that is missing instructions is the [API spec page ](https://rasa.com/docs/rasa/pages/http-api)🙈

rasa/server.py
ancalita: That's strange, according to this [deprecation warning](https://rasa.com/docs/rasa/next/migration-guide/#nlu-json-format) in the migration guide, the JSON nlu format should only be removed in 4.0 🤔

docs/docs/http-api.mdx
melindaloubser1: ahhh I see I see, will further update the spec

rasa/server.py
melindaloubser1: > so this becomes a problem further below in L340 because tracker.as_story() makes a call to applied_events which will be an empty list if the last event was Restarted?


Exactly.


> What would happen if other events follow the Restarted event?


That should not be possible in this context because `Restarted` is always followed directly by `SessionStarted`, and the conversation has already been split on `SessionStarted` events.

rasa/server.py
melindaloubser1: Ah right I don't mean it was removed everywhere, but it was removed for this endpoint specifically.

melindaloubser1: If you think it should still be supported, then I'll revert this and fix the handling of JSON to work

ancalita: I think we should be consistent and support it here too, maybe worth adding a deprecation warning that this format will be removed from this endpoint in 4.0.

rasa/shared/core/trackers.py
melindaloubser1: After looking at it again, I'm removing the fix. I think it requires more thought about what constitutes a "session" in the context of test stories. For example, should we be adding a record of the user's `/restart` utterance in test stories? What about `action_restart`? I think it's better to open a separate ticket and make decisions about these explicitly.

docs/docs/http-api.mdx
melindaloubser1: I added some description to the spec page, lmk if it serves the purpose you had in mind

rasa/server.py
melindaloubser1: have added a method for it.

docs/docs/http-api.mdx
ancalita: Looks great 🙌🏼

rasa/server.py
```diff
@@ -1484,6 +1518,18 @@ def _validate_yaml_training_payload(yaml_text: Text) -> None:
         )
 
 
+def _validate_json_training_payload(json_text: Text) -> None:
+    try:
+        RasaYAMLReader
```
ancalita, line 1523: This appears incomplete 😄 also probably not needed, since you used `rasa.shared.utils.validation.validate_training_data` function

rasa/server.py
```diff
@@ -1472,6 +1479,33 @@ def _training_payload_from_yaml(
     )
 
 
+def _nlu_training_payload_from_json(
```
ancalita, line 1482: Could we have a unit test covering this new function? 🙏🏼

ancalita (commented):
Almost there 💯 Only asked for one unit test and probably some code can be removed.

rasa/server.py
melindaloubser1: ope this is entirely irrelevant sorry

melindaloubser1:
Follow up issue is here https://github.com/RasaHQ/rasa/issues/11151

rasa/server.py
```diff
@@ -1152,7 +1157,9 @@ async def evaluate_intents(
                     test_data, config_file, int(cross_validation_folds)
                 )
         else:
-            test_data = _test_data_file_from_payload(request, temporary_directory)
+            payload = _nlu_training_payload_from_json(request, temporary_directory)
```
ancalita, line 1160: Should we add the deprecation warning here too?

melindaloubser1: Fortunately it appears automatically because the method calls other methods that contain it

ancalita (approved):
🎉

Issue #4895, opened 2019-12-04
Text with two line breaks  (\n\n) results in the message being broken into multiple parts.


thusithaC:
**Rasa version**: rasa==1.4.3

**Rasa SDK version** (if used & relevant): rasa-sdk==1.4.0

**Python version**: 3.6.8

**Operating system** (windows, osx, ...): ubuntu 18.04, linux

**Issue**:

I have text content which has multiple newline components. This formatting is important because it seems like this is the only way I can make a specific Front-End (Cannot change or edit the FE) parse Markdown formatted text. 

Example sentence text : 
```
A **fund** is a form of collective investment plan managed by professionals. \n\nA Fund can include multiple types of assets like Stocks\n- Bonds\n- Money Market instruments, etc. 
```

My issue is Rasa seems to be breaking down this sentence into multiple messages when I use 
```
dispatcher.utter_message(output)
```

example: 
```
[
    {
        "recipient_id": "fadsf",
        "text": "A **fund** is a form of collective investment plan managed by professionals. "
    },
    {
        "recipient_id": "fadsf",
        "text": "**Assets in a fund could be**: "
    },
    {
        "recipient_id": "fadsf",
        "text": "- Stocks\n- Bonds\n- Money Market instruments, etc. "
    },
]
```  

However, weird thing is if a buttons are present, this gets send correctly. 
i.e. the code is: 
```
dispatcher.utter_message(output, buttons=buttons)
```

The return will be: 
```
[
    {
        "recipient_id": "fadsf",
        "text": "A **fund** is a form of collective investment plan managed by professionals. \n\n**Assets in a fund could be**: \n\n- Stocks\n- Bonds\n- Money Market instruments, etc. \n\nWhen investors subscribe for funds, they will be given a certain number of fund units or shares whose value will change with the fluctuation of the net asset value of the funds.",
        "buttons": [
            {
                "choice": "single",
                "payload": "What is the types of fund offered by bank",
                "title": "Types of funds"
            },
            {
                "choice": "single",
                "payload": "What is the major risk of fund",
                "title": "Fund risks"
            },
            {
                "choice": "single",
                "payload": "How can i invest to fund",
                "title": "Fund subscription"
            }
        ]
    },
]
```

The second form of representation is the desired because I do not want Rasa to break down text into multiple parts based on any internal characters. 

[...]

PR #11131, opened 2022-05-19
Bugfix splitting of multiline messages containing buttons

mrkaaa:
**Proposed changes**:
Currently multiline messages don't get split into separate messages on two newlines if they contain buttons. And this seems to be a bug.
The change proposed is to split a message on separate messages on two newlines and include buttons with the last of them.

Some discussion on this problem I found here: https://github.com/RasaHQ/rasa/issues/4895

### DIFF
--- a/rasa/core/channels/channel.py
+++ b/rasa/core/channels/channel.py
@@ -393,9 +393,17 @@ class CollectingOutputChannel(OutputChannel):
         buttons: List[Dict[Text, Any]],
         **kwargs: Any,
     ) -> None:
-        await self._persist_message(
-            self._message(recipient_id, text=text, buttons=buttons)
-        )
+        message_parts: List[Text] = text.strip().split("\n\n")
+        last_message_id: int = len(message_parts)
+
+        for i, message_part in enumerate(message_parts, start=1):
+            await self._persist_message(
+                self._message(
+                    recipient_id,
+                    text=message_part,
+                    buttons=buttons if i == last_message_id else None,
+                )
+            )
 
     async def send_custom_json(
         self, recipient_id: Text, json_message: Dict[Text, Any], **kwargs: Any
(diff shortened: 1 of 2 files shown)


mrkaaa:
Not sure whether I need to change anything in documentation.

mrkaaa:
@ancalita could you please have a look?
It seems like there is a flaky test in CI as it was successful once though I didn't change anything that could cause it fail.

rasa/core/channels/channel.py
```diff
@@ -393,11 +393,21 @@ async def send_text_with_buttons(
         buttons: List[Dict[Text, Any]],
         **kwargs: Any,
     ) -> None:
-        await self._persist_message(
-            self._message(recipient_id, text=text, buttons=buttons)
-        )
+        """Sends text message with buttons."""
+        message_parts: List[Text] = text.strip().split("\n\n")
+        last_message_id: int = len(message_parts)
```
losterloh, line 398: This variable name is slightly misleading:
- it's not the ID of the last message, but the index of the last message
- However, it's also not really the index of the last message either, and only becomes that index because of the `start=1` below.

This is more of a nitpick, but I'd suggest to remove the `start=1`, changing the arithmetic accordingly, and renaming this variable.

rasa/core/channels/channel.py
```diff
@@ -393,11 +393,21 @@ async def send_text_with_buttons(
...
+
+        for i, message_part in enumerate(message_parts, start=1):
```
losterloh, line 400: Nit: Give the loop index a slightly more descriptive name like `idx` or `index`.

losterloh (approved):
Looks good 🚀 I left two small nits that I suggest changing. Thanks for pointing out the flaky CI, something we need to look into, for now I'll re-trigger that CI job.

rasa/core/channels/channel.py
mrkaaa: Fixed this. It looks better now, thank you!

anaszil:
I am working with RASA 3.6.14, and I just discovered that this bugfix is no longer in the code. Is there any reason for this ? 
Thank you.

PR #11149, opened 2022-05-26
Extract slots with mapping conditions during form activation

ancalita:
**Proposed changes**:
- Fixes the extraction of a slot with mapping conditions that should be filled with information from the user message that triggers the form.


rasa/core/actions/forms.py
```diff
@@ -593,10 +616,21 @@ async def activate(
...
 
+        validated_slot_names = [
+            event.key for event in validate_events if isinstance(event, SlotSet)
+        ]
+
+        return validate_events + [
+            event
+            for event in extraction_events
+            if isinstance(event, SlotSet) and event.key not in validated_slot_names
+        ]
```
m-vdb, line 632: would be rad to be able to do compare events with `!=` or `==` 🤔 I thought this was possible?

m-vdb: 🤔 I see that `__eq__` raises `NotImplemented`, which is really not helpful ahah

rasa/core/actions/forms.py
```diff
@@ -584,6 +585,28 @@ async def activate(
         # collect values of required slots filled before activation
         prefilled_slots = {}
 
+        # run slot extraction the second time during form activation
+        # for slots with mapping conditions
```
m-vdb, line 589: should we update the docstring of the function?

tests/core/actions/test_forms.py
```diff
@@ -90,10 +92,14 @@ async def test_activate_with_prefilled_slot():
...
         - type: from_entity
           entity: {slot_name}
+          conditions:
+          - active_loop: {form_name}
       {next_slot_to_request}:
         type: text
         mappings:
         - type: from_text
+          conditions:
+          - active_loop: {form_name}
```
m-vdb, line 102: should the test be duplicated instead?

rasa/core/actions/forms.py
```diff
@@ -656,3 +691,21 @@ async def deactivate(self, *args: Any, **kwargs: Any) -> List[Event]:
...
+    ) -> List[Event]:
+        temp_tracker = tracker.copy()
+
+        if not await self.is_activated(output_channel, nlg, tracker, domain):
+            default_activation_events = self._default_activation_events()
+            temp_tracker.update_with_events(default_activation_events, domain)
+            events += default_activation_events
+            events += await self.activate(output_channel, nlg, temp_tracker, domain)
+
+        return events
```
m-vdb, line 711: I am not sure what this function does, especially the operation on the temp tracker. Can you explain?

m-vdb (commented):
Nice turnaround @ancalita 🚀 I've left a few comments

rasa/core/actions/forms.py
```diff
@@ -584,6 +585,28 @@ async def activate(
...
+
+        extraction_events = await action_extract_slots.run(
+            output_channel, nlg, tracker, domain
+        )
+
+        events_as_str = "\n".join([str(e) for e in extraction_events])
+        logger.debug(
+            f"The execution of '{ACTION_EXTRACT_SLOTS}' resulted in "
+            f"these events: {events_as_str}."
+        )
```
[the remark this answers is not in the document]
ancalita: Since the tracker that's used effectively is a temporary tracker during `action.run()`, it doesn't matter much that the log gets issued after this update. I will explain in your other comment the decision behind this temp tracker.

rasa/core/actions/forms.py
```diff
@@ -584,6 +585,28 @@ async def activate(
...
+
+        tracker.update_with_events(extraction_events, domain)
```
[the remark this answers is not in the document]
ancalita: I have to update the tracker immediately in order for the tracker to update its `tracker.slots` attribute, which gets used in the for loop below (L610-612) to extract the prefilled slots. Because the tracker is temporary, it won't duplicate the events in the **actual** tracker that the processor updates.

rasa/core/actions/forms.py
ancalita: Yup, I'll see how I can add more context to the docstring about this new code.

tests/core/actions/test_forms.py
ancalita: Yep, good point 👍🏼

rasa/core/actions/forms.py
ancalita: If you look into the `LoopAction.run()` which gets inherited by `FormAction`, this method does three things:
1. activates the form
2. allows the form to collect info (or `do`)
3. deactivates the form once conditions for deactivation are met

It is only after all these steps are done, does the MessageProcessor update the tracker with events (which includes the vital `ActiveLoop` event which sets the `tracker.active_loop_name` to the form name). This is what gets checked by the default action `action_extract_slots` [here](https://github.com/RasaHQ/rasa/blob/e8c0f8495a3960a0feac64fbf3f2fb1c75cc49ce/rasa/core/actions/action.py#L1028) when coming across slots with mapping conditions.

Because this update of the tracker by the MessageProcessor happens **too late**, I had to override a part of the superclass `run` method, namely the `activation` part which I extracted and used a temp tracker instead that gets updated with `ActiveLoop` event in L707 - this is a default activation event. This way ACTION_EXTRACT_SLOTS can fill a slot with a mapping condition during activation, because the `temp_tracker.active_loop_name` will be set to the current form name and will pass the check in the default action `_verify_mapping_conditions()`.

Let me know if any of this is not clear enough 🙏🏼

rasa/core/actions/forms.py
ancalita: I think implementing `__eq__` in events is out of scope here 😬 Happy to submit a ticket though for a clean-up?

tests/core/actions/test_forms.py
ancalita: done: https://github.com/RasaHQ/rasa/pull/11149/commits/8c87ff1eac90aca80a73d27c3d56afbb8e9bae74

rasa/core/actions/forms.py
ancalita: Done: https://github.com/RasaHQ/rasa/pull/11149/commits/8c87ff1eac90aca80a73d27c3d56afbb8e9bae74

rasa/core/actions/forms.py
m-vdb: yeah of course. I am not sure how much it's worth it and how much it would simplify the code, maybe that's worth having a technical spike for that 🤔

ancalita: Opened a [spike ticket](https://rasahq.atlassian.net/browse/ATO-128)

rasa/core/actions/forms.py
m-vdb: OK I think that I understand, thanks a lot for detailing this. Something that was confusing to was the use of a temp tracker. Basically it is mainly used here as a way to compute events that need to actually be set in the real tracker.

rasa/core/actions/loops.py
```diff
@@ -92,3 +90,17 @@ async def deactivate(
...
+
+    async def _activate_loop(
+        self,
+        output_channel: "OutputChannel",
+        nlg: "NaturalLanguageGenerator",
+        tracker: "DialogueStateTracker",
+        domain: "Domain",
+        events: List[Event],
+    ) -> List[Event]:
+        if not await self.is_activated(output_channel, nlg, tracker, domain):
```
m-vdb, line 102: can this line actually stay in `def run()`?

ancalita: Yup, of course, less duplication 😄

m-vdb:
@ancalita given Arjaan response on the ticket, I think this should wait a review from Lukas and/or Melinda. I initially started giving a review in order to go faster but now we're waiting on customer feedback.

joeri-jongen-kbc-be:
> @ancalita given Arjaan response on the ticket, I think this should wait a review from Lukas and/or Melinda. I initially started giving a review in order to go faster but now we're waiting on customer feedback.

@m-vdb @ancalita 
I'm working for the customer, we validated the fix and it seems to work 🥳

losterloh (approved):
Very nice 🚀  Just left some small comments and will leave it up to you whether to tackle them or not.

rasa/core/actions/forms.py
```diff
@@ -593,10 +615,21 @@ async def activate(
             return []
 
         logger.debug(f"Validating pre-filled required slots: {prefilled_slots}")
-        return await self.validate_slots(
+
+        validate_events = await self.validate_slots(
```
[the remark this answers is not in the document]
ancalita: Done: https://github.com/RasaHQ/rasa/pull/11149/commits/fbe279cc616fed9887016b21a33f374da7f1519b

rasa/core/actions/forms.py
```diff
@@ -584,6 +586,26 @@ async def activate(
...
+
+        logger.debug(
+            f"Executing default action '{ACTION_EXTRACT_SLOTS}' at form activation."
+        )
+
+        extraction_events = await action_extract_slots.run(
+            output_channel, nlg, tracker, domain
+        )
+
+        events_as_str = "\n".join([str(e) for e in extraction_events])
```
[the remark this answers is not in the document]
ancalita: Done: https://github.com/RasaHQ/rasa/pull/11149/commits/fbe279cc616fed9887016b21a33f374da7f1519b

rasa/core/actions/forms.py
```diff
@@ -656,3 +690,20 @@ async def deactivate(self, *args: Any, **kwargs: Any) -> List[Event]:
...
+    async def _activate_loop(
+        self,
+        output_channel: "OutputChannel",
+        nlg: "NaturalLanguageGenerator",
+        tracker: "DialogueStateTracker",
+        domain: "Domain",
+        events: List[Event],
+    ) -> List[Event]:
+        default_activation_events = self._default_activation_events()
+        events += default_activation_events
```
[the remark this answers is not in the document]
ancalita: In the end I opted for a slightly different solution by removing the events as a parameter but I think it addresses your concerns, let me know wdyt: https://github.com/RasaHQ/rasa/pull/11149/commits/fbe279cc616fed9887016b21a33f374da7f1519b

losterloh (approved):

PR #11153, opened 2022-05-29
arm64

neto007:
**Proposed changes**:
- ...


RASADSA:
Good Day @neto007,

thanks for the PR on arm64 support but we need to push back for the time being.

We have constraints in RASA OSS with current running tensorflow version - it would make RASA OSS unstable in ARM64 environments. ( it will start but crash later if you put it up for work due missing CPU instructions )
more infos here -> https://github.com/tensorflow/tensorflow/issues/52845

Its planed to bump tensorflow version - but no ETA currently.
With the tensorflow version bump docker multiarch containers are possible for ARM64 support in RASA OSS.

PR #11164, opened 2022-06-02
[ATO-89] Add tests to validate rasa data validate works without errors

sanchariGr:
**Proposed changes**:
- Internal [ticket](https://rasahq.atlassian.net/browse/ATO-89), added some tests to reduce manual effort in checking cli commands.


tests/integration_tests/core/test_cli_response.py
```diff
@@ -0,0 +1,42 @@
+test_cli_response.py
```
m-vdb, line 1: oups?

m-vdb (changes requested):

tests/integration_tests/core/test_cli_response.py
m-vdb: @sanchariGr this prevents the test from running at all, doesn't it?

tests/integration_tests/core/test_cli_response.py
```diff
@@ -0,0 +1,42 @@
...
+
+# NOTE this test will be extended to capture cli logs during process run and validate log level
+
+def test_rasa_validate_debug_no_errors():
+    # Test captures the subprocess output for the command run
+    # and validates that the bot data in 'data/test/test_integration' throws no cli errors in 'debug' mode
+    command = ["rasa data validate --data data/test/test_integration -d data/test/test_integration/domain.yml", "--debug"]
+    out, exitcode = capture(command)
+    assert exitcode == 0
+    assert out is ''
```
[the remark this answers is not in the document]
sanchariGr: @m-vdb fixed all the lint errors

tests/integration_tests/core/test_cli_response.py
```diff
@@ -0,0 +1,41 @@
+import subprocess
+
+VALIDATE_COMMAND = "rasa data validate --data data/test/test_integration \
+\-d data/test/test_integration/domain.yml"
```
m-vdb, line 4: in python, it's usually preferred to not use an escape character for strings, because there are other ways to do this more conveniently. For instance you can break your string into multiple lines like this (notice there is no `,` so it's not a tuple)
```suggestion
VALIDATE_COMMAND = (
    "rasa data validate --data data/test/test_integration "
    "-d data/test/test_integration/domain.yml"
)
```

m-vdb (approved):
All good! one minor comment that will fix the code quality error

ancalita (commented):
Nice to see this follow-up 💯 
Are you planning a separate PR for testing this command when errors are expected?

tests/integration_tests/core/test_cli_response.py
```diff
@@ -0,0 +1,43 @@
...
+
+VALIDATE_COMMAND = (
+    "rasa data validate --data data/test/test_integration "
+    "-d data/test/test_integration/domain.yml"
+)
+
+
+# NOTE this will be extended to test cli logs at process run to validate log level
+def capture(command):
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE, shell=True)
```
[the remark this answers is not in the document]
sanchariGr: Will add the pytester fixture since like you mentioned, it causes unwanted side effects like the initialise_telemetry stuff.

sanchariGr:
> Nice to see this follow-up 💯 Are you planning a separate PR for testing this command when errors are expected?

good idea, I can add a couple of tests with incorrect data files.

tests/integration_tests/core/test_cli_response.py
```diff
@@ -0,0 +1,107 @@
...
+    result = run(
+        "data",
+        "validate",
+        "--data",
+        str(source_file),
+        "-d",
+        str(domain_file),
+        "--debug",
+    )
+    assert result.ret == 0
```
ancalita, line 38: Could you also please add an assertion to check `result.outlines` is as expected?

tests/integration_tests/core/test_cli_response.py
```diff
@@ -0,0 +1,107 @@
...
+    result = run(
+        "data",
+        "validate",
+        "--data",
+        str(source_file),
+        "-d",
+        str(domain_file),
+        "--verbose",
+    )
+    assert result.ret == 0
```
ancalita, line 86: Same comment for  adding assertion to check`result.outlines`

tests/integration_tests/core/test_cli_response.py
```diff
@@ -0,0 +1,107 @@
...
+    result = run(
+        "data",
+        "validate",
+        "--data",
+        str(source_file),
+        "-d",
+        str(domain_file),
+        "--quiet",
+    )
+    assert result.ret == 0
```
ancalita, line 107: Similar comment to the ones above about `result.outlines`.

ancalita (commented):
Looks great 💯 
I suggested to add the assertion about `result.outlines` to explicitly check what gets printed to the user in each case.

tests/integration_tests/core/test_cli_response.py
sanchariGr: Done @ancalita can you take another look 🙏

ancalita (approved):
Whoop Whoop 🎉

PR #11171, opened 2022-06-03
[ATO-114]Add nightly workflows and creation scripts

sanchariGr:
**Proposed changes**:
- [ticket](https://rasahq.atlassian.net/browse/ATO-114)
- Added a workflow and script for nightly releases (Internal)
- The build will be released in the same rasa pypi as oss
- The name should be tagged with "<latest_rasa_version>.dev.date"

Tags created :
Screenshot 2022-08-05 at 18 05 21

Pypi publish also works :
[rasa-nightly](https://pypi.rasa.com/#/package/rasa-nightly)


.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.RASABOT_GITHUB_TOKEN }}
+          script: |
+            github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: "refs/tags/${{ steps.set_tagname.outputs.tag }}",
+              sha: context.sha
+            })
```
m-vdb, line 49: won't pushing the tag trigger the [`continuous-integration` workflow](https://github.com/RasaHQ/rasa/blob/44e85366b6410a9b420e12f08b2babd149b10aa9/.github/workflows/continous-integration.yml#L7-L8), especially [the docker image push and pypi deploy](https://github.com/RasaHQ/rasa/blob/main/.github/workflows/continous-integration.yml#L692-L851)?

m-vdb: (it's possible to avoid the recursion of workflow when using [the `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow), but I'm not sure it'll have the necessary permissions)

scripts/prepare_nightly_release.py
```diff
@@ -0,0 +1,134 @@
...
+    elif version == "micro":
+        return Version(get_current_version()).next_micro()
+    elif version == "alpha":
+        return Version(get_current_version()).next_alpha()
+    elif version == "rc":
+        return Version(get_current_version()).next_release_candidate()
+    elif is_valid_version(version):
+        return Version(version)
+    else:
+        raise Exception(f"Invalid version number '{cmdline_args.next_version}'.")
```
m-vdb, line 100: 🤔 this seems like it's a wrong copy paste

.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,93 @@
...
+        id: set_tagname
+        run: |
+          DATE=$(date +'%Y%m%d')
+
+          # Find latest rasa-oss version
+          echo "Trying to find the latest rasa-oss version..."
+          LATEST_RASA_VERSION=$(python -c "import sys; import os; sys.path.append('src/rasa/rasa'); from version import __version__; print(__version__)")
+          echo "Current RASA version: ${LATEST_RASA_VERSION}"
+
+          LATEST_NIGHTLY_VERSION=$(echo ${LATEST_RASA_VERSION} | awk -F. '{$(NF-1) = $(NF-1) + 1;} {$(NF) = 0;} 1' | sed 's/ /./g')
```
m-vdb, line 30: this is rather `NEXT_MINOR_VERSION`, isn't it?

.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,93 @@
...
+
+          echo "New nightly release version: ${LATEST_NIGHTLY_VERSION}"
+
+          echo "Composing nightly build tag name..."
+          GH_TAG=${LATEST_NIGHTLY_VERSION}-dev${DATE}
```
m-vdb, line 35: can we follow [pep440](https://peps.python.org/pep-0440/#developmental-releases) here and use a `.`? AFAIK our tag names are exactly our python version names.
```suggestion
          GH_TAG=${LATEST_NIGHTLY_VERSION}.dev${DATE}
```

m-vdb (changes requested):

.github/workflows/nightly_release.yml
losterloh: This is actually a very interesting rabbit hole to get into, I was wondering if semvar has any opinion on the matter and found this: https://github.com/semver/semver/issues/200

But I agree with @m-vdb , let's go with the `.`.

.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,93 @@
...
+          echo ${GH_TAG}
+          echo "::set-output name=tag::${GH_TAG}"
+
+      - name: Tag latest main commit as nightly
+        uses: actions/github-script@v5
+        with:
+          github-token: ${{ secrets.RASABOT_GITHUB_TOKEN }}
+          script: |
+            github.rest.git.createRef({
```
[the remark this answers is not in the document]
losterloh: Git tags can have messages (and are then called annotated tags), so hopefully that should be possible here.

.github/workflows/nightly_release.yml
losterloh: We're about to change that so tags do _not_ trigger the CI workflow, as on tags, we'll trigger the release workflows instead, which will contain the CI workflows.

.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: actions/checkout@v3
+
+      - name: Compose tag name
+        id: set_tagname
+        run: |
+          DATE=$(date +'%Y%m%d')
+
+          # Find latest rasa-oss version
+          echo "Trying to find the latest rasa-oss version..."
+          LATEST_RASA_VERSION=$(python -c "import sys; import os; sys.path.append('src/rasa/rasa'); from version import __version__; print(__version__)")
```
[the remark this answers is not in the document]
losterloh: Wouldn't it need to be `${{ github.workspace }}/rasa`?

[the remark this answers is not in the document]
m-vdb: yeah, I even suggest to keep `${{ github.workspace }}` and do `from rasa.version import __version__` but that's really nitpicking

.github/workflows/nightly_release.yml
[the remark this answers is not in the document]
sanchariGr: Turns out with the `github.rest.git.createRef` script i cannot add a tag. I can move this tagging and pushing into the nightly script if we need to annotate it. @losterloh let me know if I should do that.

.github/workflows/nightly_release.yml
sanchariGr: I didnot understand this, I am using the here, 
` with:
          github-token: ${{ secrets.RASABOT_GITHUB_TOKEN }}`

.github/workflows/nightly_release.yml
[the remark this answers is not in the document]
losterloh: sounds good to me @sanchariGr! Generally, I'd say something like "the more complex the logic, the more we should see about putting it in scripts instead of directly in the github actions workflows".

sanchariGr:
@losterloh all the review changes have been made, please take a look 🙏

scripts/prepare_nightly_release.py
```diff
@@ -0,0 +1,123 @@
...
+
+    return parser
+
+
+def project_root() -> Path:
+    """Root directory of the project."""
+    return Path(os.path.dirname(__file__)).parent
+
+
+def get_current_version() -> Text:
```
ancalita, line 34: Where is this function being used? I wasn't able to find where it's being called 🙈

ancalita (commented):
LGTM overall, great stuff 💯 
Left some smallish comments.
Also another question that might be silly: how will the usual micro/minor/major release script behave when reading a nightly version from the `version.py` module? Will this affect it in any way?

.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,116 @@
...
+          DATE=$(date +'%Y%m%d')
+
+          # Find latest rasa-oss version
+          echo "Trying to find the latest rasa-oss version..."
+          LATEST_RASA_MINOR=$(python -c "import sys; import os; sys.path.append('${{ github.workspace }}/rasa'); from rasa.version import __version__; print(__version__)")
+          echo "Current RASA version: ${LATEST_RASA_MINOR}"
+
+          LATEST_NIGHTLY_VERSION=$(echo ${LATEST_RASA_MINOR})
+
+          echo "New nightly release version: ${LATEST_NIGHTLY_VERSION}"
```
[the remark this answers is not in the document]
sanchariGr: Sounds right!

scripts/prepare_nightly_release.py
sanchariGr: Very good catch, thanks, forgot to remove this after i removed the function that was calling it.

sanchariGr:
> LGTM overall, great stuff 💯 Left some smallish comments. Also another question that might be silly: how will the usual micro/minor/major release script behave when reading a nightly version from the `version.py` module? Will this affect it in any way?

@ancalita very good question, i did run the `make release` script locally while the version.py and pyproject package name was changed to the nightly version name and it seems to work fine(added a screenshot of the diff from the `prepare-release-3.2.7` branch). So i don't anticipate any issues but I'll keep an eye out on the next release.
Screenshot 2022-08-31 at 10 43 13

.github/workflows/nightly_release.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Update version (nightly releases) 🚀
+        run: |
+          poetry run pip install toml pep440_version_utils
+          poetry run python ./scripts/prepare_nightly_release.py --next_version "${{ needs.run_script_and_tag_nightly_release.outputs.tag_name }}"
+
+      - name: Build ⚒️ Distributions
+        run: |
+          poetry build
+
+      - name: Publish to PyPI 📦
```
ancalita, line 90: Last minute silly questions 🤪 
If we push to the private GCP artifact registry, why do we publish on PyPI as well?

sanchariGr: This is the same workflow as in rasa-plus, which Jaime had implemented. I will ask him if he plans on getting rid of the pypi and just keep GCP.

ancalita (approved):
🎉

PR #11193, opened 2022-06-15
Fix `rasa data validate` for rules containing default actions as `active_loop`

### LINKED ISSUE #10430: rasa data validate causes error with "action_two_stage_fallback"
### Rasa Open Source version

3.0.0

### Rasa SDK version

 _No response_ 

### Rasa X version

 _No response_ 

### Python version

3.8

OSX

### What happened?

When using the two-stage fallback, I get the following error when running `rasa data validate`

UserWarning: The form 'action_two_stage_fallback' is used in the 'Activate two stage fallback' block, but it is not listed in the domain file. You should add it to your domain file!

I have this in my rules:

version: "3.0" rules:

*   rule: Activate two stage fallback steps:
*   intent: nlu_fallback
*   action: action_two_stage_fallback
*   active_loop: action_two_stage_fallback

### Command / Request

<pre>rasa data validate</pre>

### Relevant log output

 _No response_

ancalita:
**Proposed changes**:
- Fixes #10430 

### DIFF
--- a/changelog/10430.bugfix.md
+++ b/changelog/10430.bugfix.md
@@ -1,0 +1,1 @@
+Fix failed check in `rasa data validate` that verifies forms in rules or stories are consistent with the domain when the rule or story contains a default action as `active_loop` step.
--- a/rasa/validator.py
+++ b/rasa/validator.py
@@ -234,7 +234,7 @@ class Validator:
                     # To support setting `active_loop` to `null`
                     continue
 
-                if event.name not in self.domain.form_names:
+                if event.name not in self.domain.action_names_or_texts:
                     rasa.shared.utils.io.raise_warning(
                         f"The form '{event.name}' is used in the "
                         f"'{story.block_name}' block, but it "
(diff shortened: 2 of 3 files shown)


melindaloubser1 (approved):

losterloh (changes requested):
It seems like the test that was added does not actually test the bug / changed behavior, since the test runs green when the change in `validator.py` is reverted.

ancalita:
@losterloh Good catch! For some reason, the rule that I add is not detected: it gets transformed in a `StoryGraph` with 0 steps, which is why the test runs green since the whole for loop in the verify method gets bypassed. I'll investigate further 👍🏼

ancalita:
> the test that was added does not actually test the bug / changed behavior, since the test runs green when the change in validator.py is reverted.

Found the culprit - I was modifying by mistake the string parameter `file_name` to a Path, then the importer didn't recognise the file as a stories/rules file and therefore initialised a StoryGraph with 0 steps.  Fixed now, if you revert the change and run the test, it fails.

losterloh (approved):
🚀 💯

PR #11286, opened 2022-06-29
Remove dependabot targeting previous releases

virtualroot:
**Proposed changes**:

We can't configure dependabot for targeting upgrades related to security vulnerabilities only.
As it is, it creates too much noise.


losterloh (approved):
🚀 💜 💯

Issue #11013, opened 2022-03-21
Wrong indentation of YAML data written by rasa x 


raoulvm:
### Rasa Open Source version

2.8.24

### Rasa SDK version

2.8.4

### Rasa X version

1.0.1

### Python version

3.8

### What operating system are you using?

Linux, OSX, Windows, Self hosted Cloud (openshift)

### What happened?

When Rasa X writes YAML, its indentations for list items is wrong:

```YAML
key: 
- entry1
- entry2
```
it should be
```YAML
key:
  - entry1
  - entry2
```
The important difference, which can be hardly seen here, are the two extra spaces required for list entries. In the [Rasa Forum the formatter immediately formats the wrong format red](https://forum.rasa.com/t/wrong-indentation-of-yaml-data-written-by-rasa-x/52060/3) though.

That seems to be a recent (like 5 yers ago) issue with ruamel.yaml:
https://stackoverflow.com/questions/44388701/how-get-the-sequences-properly-indented-with-ruamel-yaml

### Command / Request

```shell
Any change in Rasa X that will write back files, no matter if used locally or in Rasa X Enterprise to create a new git branch.
```

ceromero05:
testing comment section to remove <p>

alexweidauer:
Thank you for raising this issue about Rasa X. We decided to stop supporting the Community Edition (free version) of ‘Rasa X’ (see more info here: https://rasa.com/blog/rasa-x-community-edition-changes/). That’s why we’ve closed your issue. We suggest that Rasa Enterprise customers raise the issue with our customer support team if they haven’t done so already.

raoulvm:
Thanks @alexweidauer and @amn41 this has been stale in Zendesk for 24 days now.

amn41:
thanks @raoulvm  - will respond there!

PR #11293, opened 2022-07-04
Make YAML sequence indent configurable

melindaloubser1:
**Proposed changes**:
- Make YAML sequence indent configurable via env var. 
- Once a patch is released for Rasa X including the Rasa version this goes into , it will address https://github.com/RasaHQ/rasa/issues/11013 for Rasa X <1.1.0 > 1.0.0.
See [this commit](https://github.com/RasaHQ/rasa-x/pull/7523/commits/df17e76af786e8dea07f591916ed62b999d95327) for the equivalent logic changes applied in Rasa X 1.1.x.


rasa/shared/constants.py
```diff
@@ -94,3 +94,5 @@
 
 RESPONSE_CONDITION = "condition"
 CHANNEL = "channel"
+DEFAULT_YAML_SEQUENCE_INDENT_ENV_VAR = "DEFAULT_YAML_SEQUENCE_INDENT"
```
ancalita, line 97: I looked at the GH issue and Jira ticket but I couldn't find the request to make this configurable, isn't it better to enforce an indentation of 2 at all times instead? That seems to have been the observation in the GH issue too: `The important difference, which can be hardly seen here, are the two extra spaces required for list entries. ` 🤔

ancalita: Otherwise looks great 👍🏻

melindaloubser1: The implementation in Rasa X made it configurable, so I followed the same pattern. The request wants an indentation of 4 (which, when the dash is added, looks like two), but the default in Rasa up till now was 2. We don't want to change the behaviour for all users, only the ones who use internal yaml formatters that require specific indentation.

ancalita (approved):

PR #11305, opened 2022-07-06
Slack channel fixes

Maxinho96:
**Proposed changes**:
- Fixed response when slack sends retries (None was giving error)
- Running on_new_message as an asyncio background task instead of a blocking await (so that we immediately return 200 to Slack and we don't reach the 3 seconds timeout if the action execution is slow)

### DIFF
--- a/changelog/11305.bugfix.md
+++ b/changelog/11305.bugfix.md
@@ -1,1 +1,2 @@
-Fixed a problem that caused Slack to go in timeout if the action took more than 3 seconds.
+- Fixed error in creating response when slack sends retry messages. Assigning `None` to `response.text` caused `TypeError: Bad body type. Expected str, got NoneType`.
+- Fixed Slack triggering timeout after 3 seconds if the action execution is too slow. Running `on_new_message` as an asyncio background task instead of a blocking await fixes this by immediately returning a response with code 200.
--- a/changelog/11305.bugfix.md
+++ b/changelog/11305.bugfix.md
@@ -1,0 +1,1 @@
+Fixed a problem that caused Slack to go in timeout if the action took more than 3 seconds.
--- a/tests/core/channels/test_slack.py
+++ b/tests/core/channels/test_slack.py
@@ -1,3 +1,4 @@
+from http import HTTPStatus
 import json
 import logging
 import time
@@ -11,6 +12,7 @@ from sanic.compat import Header
 from sanic.request import Request
 
 from rasa.core.channels import SlackInput
+from rasa.core.channels.channel import UserMessage
 from rasa.shared.exceptions import InvalidConfigException
 from tests.utilities import json_of_latest_request, latest_request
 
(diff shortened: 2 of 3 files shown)


ancalita (changes requested):
@Maxinho96 Thank you for your contribution 🙌🏻 
Looks great 💯  Before I approve this PR, I would like you to:
- add a unit test that fails when you revert your changes and passes with your fixes (the unit test should be placed under `tests/core/channels/test_slack.py`)
- add a changelog entry in the changelog directory, you can use the PR id as changelog entry ID, please use type `bugfix`. 
 
Also please note that since your PR is targeting `main`, this change won't be released until the next minor and that is tentatively scheduled towards the end of Q3. Alternatively, you can recreate your PR to target `3.2.x` and we can release a micro much earlier.

ancalita:
@Maxinho96 To double-check did you also rebase your branch locally and then forced push? It seems the IO now contains a lot of extra commits 😬 This will have to be fixed before I can review again.

Maxinho96:
@ancalita Sorry, I have fixed with a rebase :)

tests/core/channels/test_slack.py
```diff
@@ -901,3 +915,64 @@ def test_slack_verify_signature_missing_headers():
...
+
+@pytest.mark.asyncio
+async def test_slack_process_message_retry():
+    input_channel = SlackInput(
+        slack_token="YOUR_SLACK_TOKEN",
+        slack_channel="YOUR_SLACK_CHANNEL",
+        slack_signing_secret="foobar",
+    )
+
+    request = Mock()
```
ancalita, line 932: I have a suggestion here to use `responses.RequestsMock()` context manager instead of `Mock()`. Here is the [documentation link](https://github.com/getsentry/responses#responses-as-a-context-manager), you can pass the localhost URL and the headers to this object via `add` method. Then you can access the request from the context manager like so `request_mock.calls[0].request`. 

Have a try in a time-boxed effort but if it doesn't work out, keeping the tests as they are also works, just let me know. We use this context manager in a few other tests in `test_telemetry` and `test_duckling_entity_extractor` modules if you want to have a look.

ancalita (commented):
Thanks for the speedy implementations, looks great 💯 
I just have two more suggestions for expanding a bit on the issue and solution in the changelog entry & for refactoring a bit the tests. Let me know how you get on.

changelog/11305.bugfix.md
```diff
@@ -0,0 +1 @@
+Fixed a problem that caused Slack to go in timeout if the action took more than 3 seconds.
```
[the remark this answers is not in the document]
Maxinho96: > I would suggest adding more detail in here, similar to what you wrote in the PR description with the addition to document the exact error caused by using `None` as argument (seems to be `TypeError: Bad body type. Expected str, got NoneType)`).

Done

> Related question to the timeout of 3 seconds, out of curiosity how do you know that's the Slack timeout limit?

It was shown in the Slack chat when the timeout occurred, and it is also documented [here](https://api.slack.com/apis/connections/events-api#failure_conditions)

tests/core/channels/test_slack.py
Maxinho96: I am not sure if that can be used, because here I don't need to mock an api call like a `requests.post`, I just need to mock a `Request` object. I took inspiration from the above code [here](https://github.com/RasaHQ/rasa/blob/4d18c81fd159b93d157d40a4c709c4abfbb2cc64/tests/core/channels/test_slack.py#L93)

ancalita: No problem, thanks for looking into this 👍🏻

ancalita (approved):
🎉

PR #11333, opened 2022-07-13
ATO-218 Always set slot values

losterloh:
This fixes a bug where slot values vanish when using the augmented
memoization policy. If in the past, the slot is set to a specific value,
and then later the slot being set to the same value, no new SlotSet
event is emitted. However, this leads to the augmented memoiziation
policy wrongly assuming this slot to be unset, as the older, original
SlotSet event was pruned already.

### DIFF
--- a/data/test_action_extract_slots_11333/config.yml
+++ b/data/test_action_extract_slots_11333/config.yml
@@ -1,0 +1,46 @@
+# The config recipe.
+# https://rasa.com/docs/rasa/model-configuration/
+recipe: default.v1
+
+# Configuration for Rasa NLU.
+# https://rasa.com/docs/rasa/nlu/components/
+language: en
+
+pipeline:
+# No configuration for the NLU pipeline was provided. The following default pipeline was used to train your model.
+# If you'd like to customize it, uncomment and adjust the pipeline.
+# See https://rasa.com/docs/rasa/tuning-your-model for more information.
+  - name: WhitespaceTokenizer
+  - name: RegexFeaturizer
+  - name: LexicalSyntacticFeaturizer
+  - name: CountVectorsFeaturizer
+  - name: CountVectorsFeaturizer
+    analyzer: char_wb
+    min_ngram: 1
+    max_ngram: 4
+  - name: DIETClassifier
+    epochs: 100
+    constrain_similarities: true
+  - name: EntitySynonymMapper
+  # - name: ResponseSelector
+  #   epochs: 100
+  #   constrain_similarities: true
+  # - name: FallbackClassifier
+  #   threshold: 0.3
+  #   ambiguity_threshold: 0.1
+
+# Configuration for Rasa Core.
+# https://rasa.com/docs/rasa/core/policies/
+policies:
+# No configuration for policies was provided. The following default policies were used to train your model.
+# If you'd like to customize them, uncomment and adjust the policies.
+# See https://rasa.com/docs/rasa/policies for more information.
+  - name: AugmentedMemoizationPolicy
+  - name: RulePolicy
+  # - name: UnexpecTEDIntentPolicy
+  #   max_history: 5
+  #   epochs: 100
+  # - name: TEDPolicy
+  #   max_history: 5
+  #   epochs: 100
+  #   constrain_similarities: true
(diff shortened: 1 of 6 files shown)


losterloh:
done @sanchariGr , see current changes - CI should hopefully run green now.

tests/core/test_actions.py
```diff
@@ -2665,3 +2664,51 @@ async def test_action_extract_slots_non_required_form_slot_with_from_entity_mapp
         domain,
     )
     assert events == [SlotSet("form1_info1", "info1"), SlotSet("form1_slot1", "Filled")]
+
+
+async def test_action_extract_slots_returns_slot_set_even_if_slot_value_is_unchanged():
+    event_with_slot_entity = UserUttered(
+        text="I am a text",
+        intent={"name": "intent_with_entity"},
+        entities=[{"entity": "entity", "value": "value"}],
```
melindaloubser1, line 2673: suggest to set the name & value as variables so there's no chance of mismatch lower down in the test

changelog/11333.bugfix.md
```diff
@@ -0,0 +1,5 @@
+`SlotSet` events will be emitted even if slot value remains unchanged.
+
+Previously, `ActionExtractSlots` would not emit any `SlotSet` events if the value remains
+unchanged. However, this breaks the augmented memoization policy cause it will lose these
+slot values when truncating the tracker.
```
melindaloubser1, line 5:
```suggestion
Previously, `ActionExtractSlots` would not emit any `SlotSet` events if the new value was the same as the existing one. This caused the augmented memoization policy to lose these
slot values when truncating the tracker. 
```

melindaloubser1 (changes requested):

rasa/core/actions/action.py
```diff
@@ -1234,7 +1234,7 @@ async def run(
                     if not isinstance(slot, ListSlot):
                         value = value[-1]
 
-                    if tracker.get_slot(slot.name) != value:
+                    if value is not None or slot.value is not None:
```
ancalita, line 1237: I don't remember all the unit test cases, but I'd add this test if it doesn't exist, where the tracker slot has a value, but the new value is `None`.

ancalita (commented):
Looks good 😎 
Left a few suggestions, let's wait for the added AugMemo test cases examples.

tests/core/test_actions.py
```diff
@@ -2665,3 +2664,55 @@ async def test_action_extract_slots_non_required_form_slot_with_from_entity_mapp
...
+        ],
+    )
+
+    tracker._set_slot(entity_name, value)
+
+    action = ActionExtractSlots(None)
+
+    events = await action.run(
+        output_channel=CollectingOutputChannel(),
+        nlg=Mock(),
```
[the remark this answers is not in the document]
losterloh: What's the advantage of that? As far as I can see, the nlg is not used / needed for anything in this test, and using a Mock makes that clearer to the reader of the test I think.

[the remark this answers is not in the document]
ancalita: I see, the suggestion was only to be consistent with the test cases in the module, but not a requirement.

data/test_action_extract_slots_11333/config.yml
```diff
@@ -0,0 +1,46 @@
...
+  - name: WhitespaceTokenizer
+  - name: RegexFeaturizer
+  - name: LexicalSyntacticFeaturizer
+  - name: CountVectorsFeaturizer
+  - name: CountVectorsFeaturizer
+    analyzer: char_wb
+    min_ngram: 1
+    max_ngram: 4
+  - name: DIETClassifier
+    epochs: 100
```
ancalita, line 22: A suggestion for making this run faster would be to lower this to 10 epochs maybe, but if we include this into a cron job, then fine to have it at the same level.

ancalita (approved):
Amazing ⚡ 
Final thing, you might want to place this file into `.gitignore`? It's showing up in the committed files:
`data/test_action_extract_slots_11333/models/.gitkeep`

losterloh:
@ancalita The purpose of the `.gitkeep` file is so that the `models` directory always exists, since git only knows files and not directories. So it's there on purpose. Good idea with reducing the epochs, will do that still.

ancalita:
> @ancalita reducing the epochs to 10 is actually breaking the test 😬 So I won't do that after all.

No worries, yeah there's a sweet spot with these epochs, if you go below a level it just breaks.

PR #11394, opened 2022-07-28
Fix `ValidationAction` dispatching of `BotUttered` messages

ancalita:
**Proposed changes**:
- Fixes https://github.com/RasaHQ/rasa-sdk/issues/612 (more context in the discussion [here](https://rasahq.atlassian.net/browse/ATO-245))

### DIFF
--- a/changelog/11394.bugfix.md
+++ b/changelog/11394.bugfix.md
@@ -1,0 +1,1 @@
+Enables the execution of side effects of running custom validation actions such as dispatching bot messages.
--- a/changelog/11394.bugfix.md
+++ b/changelog/11394.bugfix.md
@@ -1,1 +1,1 @@
-Enables the execution of side effects of running custom validation actions such as dispatching bot messages.
+Enables the dispatching of bot messages returned as events by slot validation actions.
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -180,6 +180,8 @@ class MessageProcessor:
             f"Default action '{ACTION_EXTRACT_SLOTS}' was executed, "
             f"resulting in {len(extraction_events)} events: {events_as_str}"
         )
+        await self.execute_side_effects(extraction_events, tracker, output_channel)
+
         return tracker
 
     async def predict_next_for_sender_id(
--- a/rasa/core/processor.py
+++ b/rasa/core/processor.py
@@ -173,6 +173,9 @@ class MessageProcessor:
         extraction_events = await action_extract_slots.run(
             output_channel, self.nlg, tracker, self.domain
         )
+
+        await self._send_bot_messages(extraction_events, tracker, output_channel)
+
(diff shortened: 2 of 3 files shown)


rasa/core/processor.py
```diff
@@ -180,6 +180,8 @@ async def run_action_extract_slots(
             f"Default action '{ACTION_EXTRACT_SLOTS}' was executed, "
             f"resulting in {len(extraction_events)} events: {events_as_str}"
         )
+        await self.execute_side_effects(extraction_events, tracker, output_channel)
```
melindaloubser1, line 183: I see that the other side effects are scheduling / cancelling reminders - do you know if the behaviour for either of those now changes?
Is it possible to include unit tests for the expected behaviour of those two, once we know what it is?

tests/core/test_processor.py
```diff
@@ -1599,3 +1599,66 @@ async def test_custom_action_triggers_action_extract_slots(
...
+    sender_id = uuid.uuid4().hex
+    message = UserMessage(
+        text="I would like to travel to Lisbon",
+        output_channel=CollectingOutputChannel(),
+        sender_id=sender_id,
+        parse_data={
+            "intent": {"name": "inform", "confidence": 1},
+            "entities": [{"entity": slot_name, "value": "Lisbon"}],
+        },
+    )
```
melindaloubser1, line 1646: I think this can be a plain text message so it's clear it has no bearing on the test

melindaloubser1 (changes requested):

rasa/core/processor.py
ancalita: Good question, I should only apply the first side-effect (sending bot messages), since the other events are disallowed during validation (only `SlotSet` and `BotUttered` are allowed event types to be returned during slot validation, see [here](await self._send_bot_messages(events, tracker, output_channel))).

melindaloubser1: Gotcha, that makes sense. 
One thing I am confused by, why do users want to return `BotUttered` events instead of returning `responses` directly? Or is this different for slot validation than for other custom actions?

ancalita: > why do users want to return BotUttered events instead of returning responses directly?

You mean why wouldn't they use `dispatcher.utter_message` instead of returning the `BotUttered` event? Not sure 😄  It seems both approaches are valid, so we wanted to enable that.

changelog/11394.bugfix.md
```diff
@@ -0,0 +1 @@
+Enables the execution of side effects of running custom validation actions such as dispatching bot messages.
```
melindaloubser1, line 1:
```suggestion
Enables the dispatching of bot messages returned as events by custom actions.
```

melindaloubser1 (approved):
just a small suggestion for the changelog since it's only the messages that are being executed and not other side effects

PR #11409, opened 2022-08-02
Persist ConveRT model locally

### LINKED ISSUE #8229: Persist ConveRT model in a local persistent cache

bielrossi15:
**Proposed changes**:
- Resolves RasaHQ/rasa/issues/8229


tmbo:
@twerkmeister do you mind taking a look at this

changelog/8229.enhancement.md
```diff
@@ -0,0 +1 @@
+Persist ConveRT model in a local persistent cache
```
twerkmeister, line 1:
```suggestion
Added configuration parameter `model_cache_dir` to `ConveRTFeaturizer` to define the local model persistence location
```

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import os
 from typing import Any, Dict, List, Optional, Text, Tuple, Type
+from pathlib import Path
```
twerkmeister, line 5: remove unused import

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -32,9 +33,7 @@
 
 # URL to the old remote location of the model which
 # users might use. The model is no longer hosted here.
-ORIGINAL_TF_HUB_MODULE_URL = (
-    "https://github.com/PolyAI-LDN/polyai-models/releases/download/v1.0/model.tar.gz"
-)
+ORIGINAL_TF_HUB_MODULE_URL = "https://github.com/davidalami/ConveRT/releases/download/1.0/nocontext_tf_model.tar.gz"
```
twerkmeister, line 36: this must not be changed, the value is used in the component to inform people to not use this model url anymore

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -50,7 +49,7 @@
 class ConveRTFeaturizer(DenseFeaturizer, GraphComponent):
     """Featurizer using ConveRT model.
 
-    Loads the ConveRT(https://github.com/PolyAI-LDN/polyai-models#convert)
+    Loads the ConveRT("https://github.com/davidalami/ConveRT/releases/download/1.0/nocontext_tf_model.tar.gz")
```
twerkmeister, line 52: this should not be a model download location but a place where to find information about the model type. It should be updated, but probably to this link: https://github.com/davidalami/ConveRT

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -89,8 +88,11 @@ def create(
         model_storage: ModelStorage,
         resource: Resource,
         execution_context: ExecutionContext,
+        cache_dir: str = None,
```
twerkmeister, line 91: this needs to be removed and instead the `cache_dir` should be added to the config in `get_default_config`

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -89,8 +88,11 @@ def create(
...
     ) -> ConveRTFeaturizer:
         """Creates a new component (see parent class for full docstring)."""
+        if cache_dir:
```
twerkmeister, line 94: this should happen in the `__init__` method and be accessed through the config. This way it will always be applied no matter in which way the Featurizer instance is created

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
```diff
@@ -89,8 +88,11 @@ def create(
...
+            os.environ["TFHUB_CACHE_DIR"] = cache_dir
```
twerkmeister, line 95: this should happen in the `__init__` method and be accessed through the config. This way it will always be applied no matter in which way the Featurizer instance is created

twerkmeister (changes requested):
Thanks for the submission! Unfortunately, quite a few things need changing, but it's doable. See comments for requested changes.

Also, the documentation needs to be updated. The new parameter needs to be explained for the ConveRTFeaturizer in the `components.mdx` file in the docs

bielrossi15:
Ok, i'll be working on these changes!

rasa/nlu/featurizers/dense_featurizer/convert_featurizer.py
bielrossi15: Ok, it's done!

Issue #11061, opened 2022-04-05
Setting domain session_expiration_time has no effect


dgradwell-ams:
### Rasa Open Source version

3.1.0

### Python version

3.9

### What operating system are you using?

Linux

### What happened?

Since version 3.1.0, my preference for `session_expiration_time` (`0`) under `session_config` in the domain doesn't seem to be working anymore.  The value is always being set to `60`, the default.

I've confirmed that this is the actual value in three ways,
1. Inspecting the result of the API call to [get the domain](https://rasa.com/docs/rasa/pages/http-api#operation/getDomain) of the running bot
2. Extracting the trained model archive and inspecting the merged domain therein, and
3. Every `60` minutes of actual inactivity on my session, `action_session_start` is invoked 

This happens whether my domain is contained in a single file or split into multiple files in a directory.  I've tried deleting the trained models and the `.rasa` cache directory and retraining the model to no avail.

Thing is, we have overridden the `action_session_start` in our project, and it has considerable side effects in our app that are only meant to be run once for each conversation. We cannot call it again -- we don't want sessions to expire.

### Reproduction repository

https://github.com/dgradwell-ams/rasa-issue-11061

sara-tagger:
Thanks for raising this issue, @mvielkind will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

rgstephens:
Hi @dgradwell-ams, I can't duplicate this. Just tested `session_expiration_time` with a value of `3` and the session restarted when I was idle past the three minutes.

dgradwell-ams:
Try setting it to zero.

rgstephens:
I tried zero and it worked for me. Started a session, waited more the 60 minutes, entered an utterance and session and slots were still set, no `action_session_start`.

I wonder if this could be channel specific. What channel are you using?  I did my testing with rasa shell.

dgradwell-ams:
@rgstephens Hey, you might be on to something.  We are using custom connectors, but perhaps I missed something when updating.

dgradwell-ams:
@rgstephens OK, so upon further investigation, we cannot see anything in our custom connector that would indicate how or why the domain would not be properly setup.  It's nearly identical to the example here: https://rasa.com/docs/rasa/connectors/custom-connectors

dgradwell-ams:
I'll try to setup a reproduction repo.

[...]

PR #11433, opened 2022-08-11
Follow-up bugfix to merging separate domain files with custom session config

ancalita:
**Proposed changes**:
- Fixes a particular bug that emerges when at least two separate domain files are being merged, where one contains a custom session config and the other files contain no session config. The latter ends up overriding the custom config because it is loaded as a Domain with a default session config since it contains no session config specification and the checks in `merge_domain_dicts` did not protect against this case.
- This is a follow-up to the comments here: https://github.com/RasaHQ/rasa/issues/11061#issuecomment-1210292972


losterloh (approved):
🔥 🚀 💯 and all the approval emojis! Great work!

